PR #424, opened 2012-07-31
Bogus gateway to test PayPal

sideshowcoder:
PayPal has some methods untestable via the normal Bogus Gateway
like setup_authorization, and handles authorization differently, the
PaypalBogusGateway behaves more like the actual PayPal Gateway to enable
simple tests without remote interaction for setup_authorization as well as authorize.


CloCkWeRX:
Any chance you'd package this as a separate gem @sideshowcoder ?

sideshowcoder:
@CloCkWeRX Good idea, haven't touched that in a while so wonder if it still works.

CloCkWeRX:
Friom unit tests, it seems to.

sideshowcoder:
Good idea, I extracted it https://github.com/sideshowcoder/active_merchant-paypal-bogus-gateway seems to work find.

PR #953, opened 2013-12-12
Update Ogone gateway : allow d3d for payment with alias

pwoestelandt:
This allow d3d usage for payment with alias


marchyoung:
@pwoestelandt Hi! Any chance of seeing a test for PR? I understand the need as Ogone allows both CC and alias (token) payments, but the entire ogone_test.rb uses CC exclusively for purchase/authorize and there are no tokenized payment tests. Also, I assume the code is expecting the alias to be a string, not a Billing::CreditCard instance, so that could also be very helpful in exposing any other issues.

PR #1059, opened 2014-03-04
Adds preliminary support for Klarna

edward:
@bslobodin @louiskearns Here’s my first draft for Klarna support.

Plan:
- [x] Klarna ActionView helper with knowns and existing Shopify API
- [x] Add support in ActionView helper that knows how to produce the fields that Klarna needs given cart items
- [x] Get confirmation on what we should do when we have more than one tax rate for a single cart line item on the Shopify side as Klarna only supports a single tax rate
- [x] Test change to `PaymentProvider`-related stuff for change around `requires_cart_items`
- [x] Add tests in Shopify that make a taxed line item and assert that it produces the Klarna-ok object
- [x] Add a transformer in Shopify that turns a Cart::LineItem into an object compatible with what the Klarna gateway wants
- [x] Add support/tests in Shopify for passing in cart items with the data that Klarna needs*
- [x] Get confirmation on what testing and production service_url is
- [x] Confirm that merchant_terms_uri, merchant_checkout_uri, merchant_base_uri, merchant_confirmation_uri can all be the same URL to the cancel page
- [x] Confirm that it’s cool to do locale-guessing from just the shop country
- [x] Test via a remote/integration test that this thing works for real with test credentials
- [x] Get real, non-staging endpoints
- [x] Figure out how return-url and notification-url works
- [x] Add missing shipping-line information

*Cart item data that Klarna needs: 'type', 'reference', 'quantity', 'unit_price', 'tax_rate'

### DIFF
--- a/lib/active_merchant/billing/integrations/klarna/helper.rb
+++ b/lib/active_merchant/billing/integrations/klarna/helper.rb
@@ -88,7 +88,7 @@ module ActiveMerchant #:nodoc:
             subtotal_price = subtotal_price.to_f
             tax_amount = tax_amount.to_f
 
-            tax_rate = tax_amount / (subtotal_price + tax_amount)
+            tax_rate = tax_amount / subtotal_price
             tax_rate = tax_rate.round(4)
 
             percentage_to_two_decimal_precision_whole_number(tax_rate)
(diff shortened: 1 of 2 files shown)


beranPro:
anything new on this PR ?
Are there any plans to integrate Klarna ?

PR #1178, opened 2014-05-07
Deprecate recurring functionality

ntalbott:
There is too much variance between the recurring functionality provided by gateways for ActiveMerchant to do a good job of abstracting it. Thus this deprecates all such functionality with the intention of stripping it out completely at some future point.

Note that this does _not_ refer to tokenization support, but rather the ability to set up and manage the ability some gateways support to automatically charge a card on a recurring basis.

@bslobodin @odorcicd @jduff do you concur?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@
 * Beanstream: Add support for Legato single use tokens [tylerrooney]
 * PayPal Digital Goods: Allow mobile [tomprats]
 * Maxipago: Add installment support [alexandremcosta]
+* Deprecate recurring API support [ntalbott]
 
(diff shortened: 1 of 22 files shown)


dscataglini:
What is the timeline for the removal of the recurring api?

ntalbott:
@dscataglini when I get a Round TUIT.

dscataglini:
@ntalbott LMFAO I've never seen this before.

timdown:
Thanks for not having got around to it yet. The application I'm working on has hundreds of customers using recurring PayPal payments and I really don't want to have to rip out ActiveMerchant and find something else.

PR #1399, opened 2014-09-29
Mercury gateway updates

shulmang:
These are changes to the Mercury Gateway needed to pass a recent Mercury integration certification.   The Mercury Developer Integrations representative I worked with for the certification was Michael Mantineo, mmantineo@mercurypay.com.  The changes included in this PR are:

    * when parsing mercury response if ref_no is not present use invoice_no for ref_no

    * include AcqRefData for PreAuthCapture as well as reversal

    * default to partial auth deny


lib/active_merchant/billing/gateways/mercury.rb
```diff
@@ -90,7 +90,7 @@ def build_non_authorized_request(action, money, credit_card, options)
           xml.tag! "Transaction" do
             xml.tag! 'TranType', 'Credit'
             xml.tag! 'TranCode', action
-            if action == 'PreAuth' || action == 'Sale'
+            if (action == 'PreAuth' || action == 'Sale') and options[:allow_partial_auth]
```
wvanbergen: Code style nitpick: `&&` instead of `and`

lib/active_merchant/billing/gateways/mercury.rb
```diff
@@ -113,7 +113,7 @@ def build_authorized_request(action, money, authorization, credit_card, options)
         xml.tag! "TStream" do
           xml.tag! "Transaction" do
             xml.tag! 'TranType', 'Credit'
-            if action == 'PreAuthCapture'
+            if action == 'PreAuthCapture' and options[:allow_partial_auth]
```
wvanbergen: `&&` instead of `and`

ntalbott:
@Shopify/active-merchant thoughts on this in general? Spreedly+Shopify are running a lot of successful Mercury transactions (http://gatewayindex.spreedly.com/gateways/mercury), so I'm really concerned/dubious about making these changes.

andrewpaliga:
I share with the concern, however partial auth deny does logically make sense for me. It would be good if the latest comments/questions on the PR could be addressed.

shulmang:
My wife recently had a baby and I'm currently really preoccupied with other demands at office.  Will address comments/questions on the PR ASAP.

ntalbott:
Thanks @shulmang!

ntalbott:
I wonder if it would help, too, to split this into multiple distinct PR's. While each of these isn't much code, they're also all meaty in potential impact so would be nice to be considering them separately.

shulmang:
I have factored out the mercury partial auth default deny logic into it's own PR: https://github.com/Shopify/active_merchant/pull/1612
I haven't addressed any issues the default auth deny logic causes in the test suite in that PR yet.
I believe that the AcqRefData for PreAuthCapture has already been added to the mercury integration so is not needed from this PR.
I believe that the invoice_no for ref_no logic was really only needed for the mercury certification scripts and seems to interfere with current test suite so am not bothering to pull to it's own PR.

PR #1401, opened 2014-09-30
Add Axcessms gateway

timtait:
http://www.axcessms.com/

### DIFF
--- a/test/remote/gateways/remote_axcessms_test.rb
+++ b/test/remote/gateways/remote_axcessms_test.rb
@@ -1,4 +1,4 @@
-require_relative '../../test_helper'
+require 'test_helper'
 
 class RemoteAxcessmsTest < Test::Unit::TestCase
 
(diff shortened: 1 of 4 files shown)


ntalbott:
Whichever maintainer gets a chance to work on this next (which might be me): I have (uncommittable) test credentials from @timtait for running remote tests.

test/remote/gateways/remote_axcessms_test.rb
```diff
@@ -0,0 +1,165 @@
+require_relative '../../test_helper'
```
wvanbergen: Please use `require 'test_helper'`

test/unit/gateways/axcessms_test.rb
```diff
@@ -0,0 +1,405 @@
+require_relative '../../test_helper'
```
wvanbergen: `require 'test_helper'`

PR #1416, opened 2014-10-10
Use a more performant implementation of luhn check

vkmita:
```
>> require 'active_merchant'
=> true
>> require 'luhn_checksum'
=> true
>> class Luhner
>>   include ActiveMerchant::Billing::CreditCardMethods
>> end
=> Luhner
>> Benchmark.realtime { 1_000_000.times { LuhnChecksum.valid?("4111111111111111") } }
=> 3.122008
>> Benchmark.realtime { 1_000_000.times { Luhner.send(:valid_checksum?, "4111111111111111") } }
=> 6.562471
>> Benchmark.realtime { 1_000_000.times { LuhnChecksum.valid?("4111111111111111") } }
=> 2.980835
>> Benchmark.realtime { 1_000_000.times { Luhner.send(:valid_checksum?, "4111111111111111") } }
=> 6.498597
```

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -143,16 +143,51 @@ module ActiveMerchant #:nodoc:
             %w[1 2 3 success failure error].include?(number.to_s)
         end
 
+        ODD_LUHN_VALUE = {
+          48 => 0,
+          49 => 1,
+          50 => 2,
+          51 => 3,
+          52 => 4,
+          53 => 5,
+          54 => 6,
+          55 => 7,
+          56 => 8,
+          57 => 9,
+          nil => 0
+        }.freeze
+
+        EVEN_LUHN_VALUE = {
+          48 => 0, # 0 * 2
+          49 => 2, # 1 * 2
+          50 => 4, # 2 * 2
+          51 => 6, # 3 * 2
+          52 => 8, # 4 * 2
+          53 => 1, # 5 * 2 - 9
+          54 => 3, # 6 * 2 - 9
+          55 => 5, # etc ...
+          56 => 7,
+          57 => 9,
+        }.freeze
+
         # Checks the validity of a card number by use of the Luhn Algorithm.
         # Please see http://en.wikipedia.org/wiki/Luhn_algorithm for details.
-        def valid_checksum?(number) #:nodoc:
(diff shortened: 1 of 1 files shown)


vkmita:
@j-mutter @ntalbott @girasquid I inlined the implementation. Looks good?

girasquid:
Looks good - thanks!

jamie:
So, for a 50% speedup in *luhn checks*, this code now raises an exception (rather than simply returning false) if the card number includes non-digit characters. :-1:

And don't get me started on https://github.com/Shopify/active_merchant/blob/7c505d4a03a09bfc495c40b946f0fff877f409f8/lib/active_merchant/billing/credit_card_methods.rb#L157

girasquid:
This raises a bigger question: if you're trying to do a luhn checksum on something that's clearly not a credit card number, **should it work**?

I'm inclined to say no, and that raising an exception is the right thing to have happen here (although maybe it should be a particular exception so that you can catch and handle it).

I wrote this test script to do luhn checks on some made-up numbers using the old check - it looks like a lot of things with non-digits in them actually passed:
```ruby
#!/usr/bin/env ruby

def valid_checksum?(number) #:nodoc:
  sum = 0
  for i in 0..number.length
    weight = number[-1 * (i + 2), 1].to_i * (2 - (i % 2))
    sum += (weight < 10) ? weight : weight - 9
  end

  (number[-1,1].to_i == (10 - sum % 10) % 10)
end

NUMBERS = [
  "4242424242424242",
  "4242hi",
  "4242hi4242",
  "4hi",
  "4242424242424242hi",
  "42424242hi424242",
]

NUMBERS.each do |number|
  p "#{number}: #{valid_checksum?(number)}"
end
```
```
"4242424242424242: true"
"4242hi: true"
"4242hi4242: true"
"4hi: false"
"4242424242424242hi: true"
"42424242hi424242: true"
```

jamie:
Oh, indeed. As they say, "be permissive on your inputs and strict on your outputs" - it's an easy change for my app to explicitly filter out non-digits, and would probably be equally easy here to `return false if number =~ /[^0-9]/` or similar.

Finding the introduced exception was an unwelcome surprise is all, especially in a section of code that isn't likely to be a performance bottleneck anyways.

girasquid:
Sorry for the surprise!

PR #1454, opened 2014-11-11
Improvements to Orbital data formatting.

boone:
Per the Orbital spec and responses from Orbital support.

* Truncates field values by byte count instead of character count
* Adds filtering to addresses for `%` `|` `^` `\` `/`
* Adds truncation to state names
* Limits countries by the supported AVS country codes
* Adds `&` as an acceptable order ID character
* Strips leading spaces from the order ID


boone:
@ntalbott anything else I should do here? Thanks!

boone:
@ntalbott thanks for your help!

PR #1463, opened 2014-11-17
Add void_release method to usaepay transaction gateway.

dppcode:
Also adds a security hash to usaepay transaction gateway api requests.

The void:release method indicates to USAEpay that when the transaction is voided it should immediately send an instruction to the institution to release the funds. With a plain void, the instruction is not sent until the batch closes.


dppcode:
Something not noted in my original comment that developers will want to be aware of - the "advanced" (aka SOAP) API does not support the immediate release, only the "transaction" API does. So, in my implementation, I use the advanced API for all transactions except void, where I instantiate a new "transaction" API gateway object and call void (with immediate release) using that. 

Anyway, happy to go either way here, but would be nice to get this merged so I can get off my fork before I implement line_items for UsaEpay.

Shall I make immediate release the default? I could put the above note in the change log and/or document it somewhere else.

PR #1469, opened 2014-11-26
SecurePayAu: Implement support for FraudGuard anti-fraud check

lathiat:
Implement support for SecurePayAu's FraudGuard.  Uses a different API endpoint, and requires a few extra fields.

Followed convention for options in all cases except the 'fraud' option flag which controls whether the fraud check is performed or not.  The existing fraud_review flag is a response from the gateway, so we don't re-use that.

Most other gateways default to making the fraud check if the required information is present, I didn't want to do that as it may cause unexpected changes to existing installations and requires setup with the merchant to be useful anyway.

I welcome feedback on the above.

Created a unit and remote test for the new functionality.  All unit tests pass, the following remote tests are currently failing but not as a result of this changeset:
test_successful_purchase_with_custom_credit_card_class (error)
test_successful_void (failure)

### DIFF
--- a/lib/active_merchant/billing/gateways/secure_pay_au.rb
+++ b/lib/active_merchant/billing/gateways/secure_pay_au.rb
@@ -2,17 +2,42 @@ require 'rexml/document'
 
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
+    # This module supports the SecurePay (Australia) gateway www.securepay.com.au.  For more information on the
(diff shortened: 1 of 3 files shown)


wvanbergen:
@lathiat Are you still working on addressing the outstanding questions?

lathiat:
Sorry yep I'll pick up on this shortly.

ntalbott:
@lathiat going to close this out since it's 30+ days old; feel free to reopen when you get a chance to address the above. Thanks!

PR #1480, opened 2014-12-04
QuickBooks Payments 

ivanfer:
Adds QuickBooks Payments gateway integration. 

Increments of work done: 
- [x] OAuth signature is generated by lazy-included `oauth` gem
- [x] Remote and unit tests included
- [x] GUID method is custom, but required in some way for QBP integration.

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -1,3 +1,5 @@
+require 'oauth'
+
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class QuickBooksGateway < Gateway
@@ -11,11 +13,22 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'http://payments.intuit.com'
       self.display_name = 'QuickBooks Payments'
       ENDPOINT =  "/quickbooks/v4/payments/charges"
+      OAUTH_ENDPOINTS = {
+        site: 'https://oauth.intuit.com',
+        request_token_path: '/oauth/v1/get_request_token',
+        authorize_url: 'https://appcenter.intuit.com/Connect/Begin',
+        access_token_path: '/oauth/v1/get_access_token'
+      }
 
       STANDARD_ERROR_CODE_MAPPING = {}
 
       def initialize(options={})
         requires!(options, :credential1, :credential2, :credential3, :credential4, :option1)
+        @consumer_key = options[:credential1]
+        @consumer_secret = options[:credential2]
+        @access_token = option[:credential3]
+        @token_secret = option[:credential4]
+        @realm_id = option[:option1]
         super
       end
 
@@ -31,7 +44,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         charge_data(post: post, options: options)
         post[:capture] = "false"
-        
+
         commit(uri: ENDPOINT, body: post)
       end
 
@@ -116,20 +129,21 @@ module ActiveMerchant #:nodoc:
 
       def amount_to_void(authorization:)
         response = parse(ssl_request(:get, "#{gateway_url}#{ENDPOINT}/#{authorization}", post_data, headers))
-        reponse[:amount]
+        response[:amount]
       end
 
       def authorization_object_from_endpoint(authorization:)
 
       end
-        
+
       def parse(body:)
         JSON.parse(body)
       end
 
       def commit(uri:, body:)
-        
-        response = parse(ssl_post(gateway_url + uri, post_data(body: body), headers))
+        endpoint = gateway_url + uri
+        request = ssl_post(endpoint, post_data(body: body), headers(method: :post, uri: endpoint))
+        response = parse(request)
 
         Response.new(
           success_from(response: response),
@@ -159,11 +173,36 @@ module ActiveMerchant #:nodoc:
         data.to_json
       end
 
-      def headers
+      def headers(method:, uri:)
+        return unless method && uri
+        request_uri = URI.parse(uri)
+
+        request_class = case method
+        when :post
+          Net::HTTP::Post
+        when :get
+          Net::HTTP::Get
+        end
+        consumer = OAuth::Consumer.new(
+          @consumer_key,
+          @consumer_secret,
+          OAUTH_ENDPOINTS)
+        access_token = OAuth::AccessToken.new(
+          consumer,
+          @access_token,
+          @token_secret)
+        client_helper = OAuth::Client::Helper.new(
+          request_class.new(request_uri),
+          consumer: consumer,
+          token: access_token,
+          realm: @realm_id,
+          request_uri: request_uri)
+        oauth_header = client_helper.header
+
         {
-          "Content-type"=> "application/json",
-          "Request-Id"=> @request_id.to_s,
-          
+          "Content-type" => "application/json",
+          "Request-Id" => @request_id.to_s,
+          "Authorization" => oauth_header
         }
       end
     end
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -2,7 +2,7 @@ require 'oauth'
 
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
-    class QuickBooksGateway < Gateway
+    class QuickbooksGateway < Gateway
       self.test_url = 'https://sandbox.api.intuit.com'
       self.live_url = 'https://api.intuit.com'
 
@@ -23,12 +23,12 @@ module ActiveMerchant #:nodoc:
       STANDARD_ERROR_CODE_MAPPING = {}
 
       def initialize(options={})
-        requires!(options, :credential1, :credential2, :credential3, :credential4, :option1)
-        @consumer_key = options[:credential1]
-        @consumer_secret = options[:credential2]
-        @access_token = option[:credential3]
-        @token_secret = option[:credential4]
-        @realm_id = option[:option1]
+        requires!(options, :consumer_key, :consumer_secret, :access_token, :token_secret, :realm_id)
+        @consumer_key = options[:consumer_key]
+        @consumer_secret = options[:consumer_secret]
+        @access_token = options[:access_token]
+        @token_secret = options[:token_secret]
+        @realm_id = options[:realm_id]
         super
       end
 
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -23,42 +23,43 @@ module ActiveMerchant #:nodoc:
       STANDARD_ERROR_CODE_MAPPING = {}
 
       def initialize(options={})
-        requires!(options, :consumer_key, :consumer_secret, :access_token, :token_secret, :realm_id)
+        requires!(options, :consumer_key, :consumer_secret, :access_token, :token_secret, :realm)
         @consumer_key = options[:consumer_key]
         @consumer_secret = options[:consumer_secret]
         @access_token = options[:access_token]
         @token_secret = options[:token_secret]
-        @realm_id = options[:realm_id]
+        @realm = options[:realm]
         super
       end
 
       def purchase(money, payment, options = {})
         post = {}
-        charge_data(post: post, options: options)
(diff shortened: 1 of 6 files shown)


lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,283 @@
+begin
+  require "oauth"
```
bizla: This is done.

bizla:
@ntalbott @girasquid @j-mutter This is ready for initial review. Ruthlessness in review is encouraged. :wink:

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+rescue LoadError
+  raise "OAuth is required to use QuickBooks Payments. Please run `gem install oauth`."
```
ntalbott: I'm actually against the prescriptive "Please run `gem install oauth`.", since in many cases one would be actually adding to their Gemfile. Maybe "You need to have the oauth gem installed in order to use the ActiveMerchant QuickbooksGateway adapter."?

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+
+        # System Error
+        'PMT-6000' => STANDARD_ERROR_CODE[:processing_error],   # A temporary Issue prevented this request from being processed.
+      }
+
+      FRAUD_WARNING_CODES = ['PMT-1000','PMT-1001','PMT-1002','PMT-1003']
+
+      def initialize(options = {})
+        requires!(options, :consumer_key, :consumer_secret, :access_token, :token_secret, :realm)
+        @consumer_key = options[:consumer_key]
```
ntalbott: I actually prefer just accessing these values straight out of `@options` vs assigning everything to instance variables.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+
+      def capture(money, authorization, options = {})
+        capture_uri = "#{ENDPOINT}/#{CGI.escape(authorization)}/capture"
+
+        commit(capture_uri)
+      end
+
+      def refund(money, authorization, options = {})
+        post = {}
+        post[:amount] = money.is_a?(String) ? money : amount(money)
```
ntalbott: Money should never be a `String`, if it is it's an error on the caller's part. Can we kill this check?

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+        refund_uri = "#{ENDPOINT}/#{CGI.escape(authorization)}/refunds"
+
+        commit(refund_uri, post)
+      end
+
+      def void(authorization, options = {})
```
ntalbott: I'm not really a fan of this `void` implementation - seems super complicated and error prone. I'd probably prefer to just not support `void` rather than trying to hack it as a refund.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((realm=\")\w+), '\1[FILTERED]').
```
ntalbott: Are you sure none of these values can have parts that don't match `\w`?

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+          gsub(%r((\"card\":{\"number\":\")\d+), '\1[FILTERED]').
+          gsub(%r((\"cvc\":\")\d+), '\1[FILTERED]')
+      end
+
+      private
+
+      def add_charge_data(post, payment, options = {})
+        add_payment(post, payment, options)
+        add_address(post, options)
+        add_context(options[:context]) if options[:context]
```
ntalbott: Not a fan of having a gateway-specific, nested `:context` option. Seems like it takes `:tax` and `:recurring`? I'd put them both at the top level of `options` if we want to support them, vs. nesting them.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+      def commit(uri, body = {}, method = :post)
+        endpoint = gateway_url + uri
+        begin
+          response = case method
+          when :post
+            ssl_post(endpoint, post_data(body), headers(method: :post, uri: endpoint))
+          when :get
+            ssl_request(:get, endpoint, nil, headers(method: :get, uri: endpoint))
+          end
+        rescue ResponseError => e
```
ntalbott: I'm not sure we want to rescue `ResponseError`, since the caller probably expects to handle it themselves (I know Spreedly does).

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+          when :post
+            Net::HTTP::Post
+          when :get
+           Net::HTTP::Get
+        end
+
+        # required for Ruby 1.9 backwards compatibilty
+        begin
+          request_object = request_class.new(request_uri)
+        rescue
```
ntalbott: Can we rescue the exact exception and check the message? Otherwise we risk masking/obscuring other errors.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+
+      def gateway_url
+        test? ? test_url : live_url
+      end
+
+      def post_data(data = {})
+        data.to_json
+      end
+
+      def headers(options)
```
ntalbott: Rather than take in `options`, can we take in `method` and `uri` to make the call signature more explicit?

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+
+        method = options[:method]
+        uri = options[:uri]
+        request_uri = URI.parse(uri)
+
+        request_class = case method
+          when :post
+            Net::HTTP::Post
+          when :get
+           Net::HTTP::Get
```
ntalbott: I know you and I would never pass in anything besides `:post` or `:get`, but I always like to add a default `else` clause to a `case` that raises an exception with useful debugging information.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+          end
+        end
+
+        response_object(response)
+      end
+
+      def response_object(raw_response)
+        parsed_response = parse(raw_response)
+
+        Response.new(
```
ntalbott: Can we also set the `fraud_review` flag?

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -0,0 +1,312 @@
...
+          success?(parsed_response),
+          message_from(parsed_response),
+          parsed_response,
+          authorization: authorization_from(parsed_response),
+          test: test?,
+          cvv_result: cvv_code_from(parsed_response),
```
ntalbott: Does it provide an `avs_result`?

ntalbott:
@bizla lots of comments inline.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Yep, that's better. Changing it up.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Agreed. It was in here entirely because of some awful `void` scenarios. I'll kill this and the void implementation altogether.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Agreed. We've asked them to provide a sane endpoint for voids instead of having to do this.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: According to their dev contact, we should be good here.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Cool, I'll remove it.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: I think we're safe here since we're in a private helper method. The exceptions that flare up are confusing, so to keep things clean, I'd prefer to keep this the way it is -- since this is just about constructing a request object appropriately for the Ruby being used. The Oauth `headers` method will barf if the URI is malformed anyway.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Can do.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Good call.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: @ntalbott does Spreedly rely on `verify` being available?

ntalbott: @bizla we like to be able to offer it, but no. That said, in this case I'd like to just leave `verify` as a $1 auth with no void; then we can come back and make it more robust later (and Spreedly customers may put pressure on Quickbooks to provide a proper authorization void if they're annoyed by the auths sitting there until expiration).

bizla: Certainly works for me. Let's do that.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: We're not supporting this as of yet either, I'll remove it entirely.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Currently, no.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Yes, done.

bizla:
@ntalbott Thanks for the solid review. Ready for another look-over; voids have been removed until we have a better mechanism from Intuit, `verify` does $1 a auth only, defaults are in place for `case` statements, and the `fraud_review` is included in responses.

test/fixtures.yml
```diff
@@ -599,6 +599,15 @@ quantum:
   login: X
   password: Y
 
+# You will need to create a developer sandbox at https://developer.intuit.com/ and
+# successfully generate an OAuth 1.0a access token and token secret.
```
j-mutter: Did anyone ever do this?

test/remote/gateways/remote_quickbooks_test.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    assert refund = @gateway.refund(@amount-1, purchase.authorization)
+    assert_success refund
+  end
+
+  def test_failed_refund
+    response = @gateway.refund(nil, '')
+    assert_failure response
+  end
+
```
j-mutter: Extra line :trollface:

test/unit/gateways/quickbooks_test.rb
```diff
@@ -0,0 +1,375 @@
...
+  def successful_purchase_response
+    <<-RESPONSE
+    {
+      "created": "2014-11-27T22:09:01Z",
+      "status": "CAPTURED",
+      "amount": "20.00",
+      "currency": "USD",
+      "card": {
+        "number": "xxxxxxxxxxxx1111",
+        "name": "alicks profit",
```
j-mutter: lol

j-mutter:
:dancer:

Issue #1489, opened 2014-12-12
NullObject Gateway


kikito:
Hi there,

While doing tests with ActiveMerchant I often needed a [Null Object](http://en.wikipedia.org/wiki/Null_Object_pattern) gateway. This is, a gateway which implemented all the methods a gateway needs, always succeeded (with no side effects), and never raised an error.

The existing [Bogus Gateway](https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/bogus.rb) is somewhere in that alley, but not exactly: It still has some logic, i.e. it raises an exception if the reference doesn't end in "1" or "2". Changing the parameters for these calls in order to comply with this was sometimes difficult and made the tests larger than they needed.

If you think this addition would be worth it, I think I could submit a pull request for this.

PR #1491, opened 2014-12-12
Null gateway

kikito:
Related with #1489 

A Null object never fails, and returns "safe" values with no side-effects.

In this case I think it was fine returning "real" responses instead of creating a "NullResponse" object, since responses are treated as "data" - they don't change the state of anything by themselves.


ntalbott:
Closing this since it's been two months with no movement. Feel free to re-open if you get a chance to work on it further. Thanks!

PR #1501, opened 2014-12-23
SagePay: make `VPSProtocol` user-configurable

boxofrad:
This makes the `VPSProtocol` parameter in the SagePay gateway configurable like so:

```ruby
ActiveMerchant::Billing::SagePayGateway.protocol_version = '2.23'
```

This is desirable due to the fact that SagePay do not intend to update the simulator to support the latest protocol, but getting live & test accounts can be a lengthy / difficult process.

So it's helpful to be able to build against "something" while you're waiting.

Have at it! :sparkling_heart:


ntalbott:
Great PR - thanks!

PR #1502, opened 2014-12-24
Add PCI Compliant Repeat Billing and Cardholder Information storage for ...

cshepherd:
From Netbilling documentation:

  This feature enables you to perform a future repeat-billing or up-sell for the same consumer within the same gateway account, rather than storing and specifying the card number and billing address again. Simply check
  "Enable PCI Compliant Repeat Billing, Up-selling and Cross-selling" in Step 6 of the Credit Cards setup page. When the feature is enabled, you would send "CS:123456789012" as the card_number, where 123456789012 is
  the trans_id of the original transaction.

Per above, netbilling.rb has been modified to accept a string ("123456789012") in lieu of a credit_card for submission.

Furthermore, Netbilling Direct Mode also supports submission of billing information to PCI compliant storage without performing any credit gateway operations with the 'Quasi' (tran_type 'Q') mode. This is implemented in the 'store' method.

### DIFF
--- a/lib/active_merchant/billing/gateways/netbilling.rb
+++ b/lib/active_merchant/billing/gateways/netbilling.rb
@@ -1,5 +1,15 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
+    # To perform PCI Compliant Repeat Billing
+    #
+    #   Ensure that PCI Compliant Repeat Billing is enabled on your merchant account:
+    #    "Enable PCI Compliant Repeat Billing, Up-selling and Cross-selling" in Step 6 of the Credit Cards setup page
+    #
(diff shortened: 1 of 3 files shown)


wvanbergen:
This looks good to me. @Shopify/payments ?

girasquid:
:+1:

ntalbott:
@wvanbergen remote tests pass OK? If so, :+1: from me.

j-mutter:
Code looks fine - haven't personally run the tests though

wvanbergen:
@cshepherd Any way you get get us credentials so we can run the remote tests?

cshepherd:
Remote tests appear to work fine using the default remote test credential account ID '104901072025'.. I've no access to any other credentials that aren't in production, or that are in test mode. If necessary, I can inquire about getting another set of credentials for testing.

PR #1503, opened 2014-12-26
Fix duplicate hash key in braintree blue gateway

grk:
This was producing a warning when running on 2.2


j-mutter:
Nice find, thanks :)

PR #1504, opened 2014-12-30
Add FirstData E4 Support for Multiple Mastercard String Formats

jcbantuelle:
The response back from FirstData uses the string 'mastercard'. When using the FirstData Response to reconstruct the credit card info, it needs this key in the BRANDS constant to convert to a valid card.

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4.rb
@@ -264,7 +264,7 @@ module ActiveMerchant #:nodoc:
 
         Response.new(successful?(response), message_from(response), response,
           :test => test?,
-          :authorization => response_authorization(action, response, credit_card),
+          :authorization => successful?(response) ? response_authorization(action, response, credit_card) : nil,
           :avs_result => {:code => response[:avs]},
           :cvv_result => response[:cvv2]
         )
--- a/lib/active_merchant/billing/gateways/firstdata_e4.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4.rb
@@ -264,7 +264,7 @@ module ActiveMerchant #:nodoc:
 
         Response.new(successful?(response), message_from(response), response,
           :test => test?,
-          :authorization => successful?(response) ? response_authorization(action, response, credit_card) : nil,
+          :authorization => successful?(response) ? response_authorization(action, response, credit_card) : '',
           :avs_result => {:code => response[:avs]},
           :cvv_result => response[:cvv2]
         )
--- a/lib/active_merchant/billing/gateways/firstdata_e4.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4.rb
@@ -27,11 +27,12 @@ module ActiveMerchant #:nodoc:
       BRANDS = {
         :visa => 'Visa',
         :master => "Mastercard",
-        :mastercard => "Mastercard",
         :american_express => "American Express",
         :jcb => "JCB",
         :discover => "Discover"
       }
+
+      E4_BRANDS = BRANDS.merge({:mastercard => "Mastercard"})
 
       self.supported_cardtypes = BRANDS.keys
       self.supported_countries = ["CA", "US"]
@@ -251,7 +252,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def card_type(credit_card_brand)
-        BRANDS[credit_card_brand.to_sym] if credit_card_brand
+        E4_BRANDS[credit_card_brand.to_sym] if credit_card_brand
       end
 
       def commit(action, request, credit_card = nil)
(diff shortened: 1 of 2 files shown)


test/unit/gateways/firstdata_e4_test.rb
```diff
@@ -32,7 +36,7 @@ def test_successful_purchase
     assert response.test?
     assert_equal 'Transaction Normal - Approved', response.message
 
-    ExactGateway::SENSITIVE_FIELDS.each{|f| assert !response.params.has_key?(f.to_s)}
```
j-mutter: Oof, nice catch

lib/active_merchant/billing/gateways/firstdata_e4.rb
```diff
@@ -263,7 +264,7 @@ def commit(action, request, credit_card = nil)
 
         Response.new(successful?(response), message_from(response), response,
           :test => test?,
-          :authorization => response_authorization(action, response, credit_card),
+          :authorization => successful?(response) ? response_authorization(action, response, credit_card) : nil,
```
j-mutter: I'm a little concerned about this change. When `#response_authorization` falls through to `#authorization_from`,with what appears to be an unsuccessful transaction, it [returns an empty string](https://github.com/jcbantuelle/active_merchant/blob/master/lib/active_merchant/billing/gateways/firstdata_e4.rb#L292-L294) not `nil`, so I'm worried that this change in behaviour may break some integrations that expect to always have a string here.

ntalbott: @j-mutter but AM makes no guarantees around `nil` vs. `""` for `Response#authorization`, so I don't think we need to maintain this behavior. It's certainly not maintained on other adapters.

j-mutter: Yeah I know, just might be a nasty surprise for someone

ntalbott: That'll teach 'em! :wink:

I don't really care either way in this specific case, just want to be cautious in general about making anyone think we support their broken calling code. :fearful:

jcbantuelle: I've updated the fallback value to be an empty string instead of nil for consistency.

jcbantuelle:
Just checking in, is there anything else you need from me to get this merged in?

jcbantuelle:
ping @ntalbott @j-mutter

lib/active_merchant/billing/gateways/firstdata_e4.rb
```diff
@@ -263,7 +264,7 @@ def commit(action, request, credit_card = nil)
 
         Response.new(successful?(response), message_from(response), response,
           :test => test?,
-          :authorization => response_authorization(action, response, credit_card),
+          :authorization => successful?(response) ? response_authorization(action, response, credit_card) : '',
```
j-mutter: Can you add an assertion for this change to your unsuccessful test?

jcbantuelle: @j-mutter Thanks for the feedback! I've added an assertion for this, please let me know if you need any further additions.

jcbantuelle:
ping @ntalbott @j-mutter

ntalbott:
This PR confuses me; does AM ever use `:mastercard` as the card type? That seems like a bug, if so.

jcbantuelle:
@ntalbott The issue isn't with AM, it's with how FirstData E4 tokenizes credit card information. If you look at https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/firstdata_e4.rb#L200, you'll see that it calls credit_card.brand. When using the FirstData token response from their server, that brand value is 'mastercard'. The card_type function at https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/firstdata_e4.rb#L262-L264 uses the BRANDS constant to convert the brand value, and without :mastercard in BRANDS, this conversion fails.

ntalbott:
@jcbantuelle OK, got it. My fear is that we're overloading `supported_cardtypes` with a value that will confuse any code using it as metadata. Can we do this without leaking the `:mastercard` key outside this adapter?

jcbantuelle:
@ntalbott Thank you for the feedback, that makes a lot of sense. I've pulled the mastercard value out into a separate hash that is used exclusively in the card_type function to keep this behavior encapsulated in the FirstData E4 adapter. Please let me know if you need any further changes.

jcbantuelle:
ping @ntalbott @j-mutter

j-mutter:
Looks fine to me. @ntalbott @wvanbergen @girasquid anyone want to give this a second +1 before I merge it?

ntalbott:
:shipit:

PR #1511, opened 2015-01-06
CyberSource: Make :ignore_avs and :ignore_cvv settable on a per-transaction basis

rud:
This small change makes it possible to create a Gateway instance with one `ignore_avs` or `:ignore_cvv` setting, and then override it in a single payment invocation. 

I hope this is a useful feature. I would however love feedback on the tests as this is my first `active_merchant` contribution. Is the testing at an appropriate level?

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -242,7 +242,7 @@ module ActiveMerchant #:nodoc:
         xml = Builder::XmlMarkup.new :indent => 2
         add_payment_method_or_subscription(xml, money, creditcard_or_reference, options)
         add_auth_service(xml)
-        add_business_rules_data(xml)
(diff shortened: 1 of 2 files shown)


ntalbott:
@rud the tests look great, but with your current code, if `@options[:ignore_avs]` is `true`, and I set `options[:ignore_avs]` to `false`, then it will still end up as `true`, right? I think you need to tweak the logic, and add a test to cover the case.

rud:
@ntalbott do'h, an excellent point - I'll get that sorted right away

rud:
@ntalbott fixed in 86f9605, thank you for the review

wvanbergen:
Looks good. @ntalbott is this ready to merge?

ntalbott:
@wvanbergen :+1:.

wvanbergen:
Merged!

rud:
Awesome, thanks!

PR #1515, opened 2015-01-07
Remove active_utils from dependency list

wvanbergen:
This removes `active_utils` from the gem dependencies. All the code we use from it should be in ActiveMerchant itself by now.

I also removed the `json` gem as a dependency; we dropped support for Ruby 1.8, and in Ruby 1.9+ it is part of the standard library.

@ntalbott @RichardBlair @karlhungus for review


ntalbott:
@wvanbergen I haven't moved these bits over yet:

```
require 'active_utils/common/network_connection_retries'
silence_warnings{require 'active_utils/common/connection'}
require 'active_utils/common/post_data'
require 'active_utils/common/posts_data'
```

wvanbergen:
@ntalbott Hmm, I was wondering why CI was succeeding, but it turns out that the `offsite_payments` dependency has a `active_utils` dependency, so it is still available in the CI environment. Is the plan to also remove `offsite_payments` as a dependency?

ntalbott:
@wvanbergen yes, I'd like to see `offsite_payments` removed as well.

One possibility is to actually hold off on removing both, and start working on a full major version bump that does it along with removing various deprecations and such.

wvanbergen:
@ntalbott The problem is we cannot load two versions of ActiveUtils at the same time. So I rather get it over with if it is feasible without a migration period. 

The offsite_gateway dependency doesn't really do anything for active merchant anyway, except making the offsite gateways available. SO removing that dependency is not a big deal; people can just add the gem to their Gemfile manually if they aren't already doing so.

wvanbergen:
@girasquid FYI.

wvanbergen:
Changes in offsite_gateways: https://github.com/Shopify/offsite_payments/pull/103

karlhungus:
@girasquid @j-mutter you guys mind having a look see too?

wvanbergen:
Still todo: copy the remaining parts activeutils that AM uses.
Some background: https://github.com/Shopify/active_merchant/pull/1492

ntalbott:
@wvanbergen sounds good to me. If you're going to do that you should rip the shims out as well.

wvanbergen:
I have copied in the necessary stuff from ActiveUtils (Connection, PostData, PostsData, and NetworkConnectionRetries), and the associated tests. 

I also had to remove the offsite payments shim. If we want to keep it around, that's possible but I will have to do a prerelease of that gem, so we can depend on a version that depends on ActiveUtils 3 instead of 2.

wvanbergen:
@bizla @girasquid @j-mutter @aprofeit  It's ready for reviewing bow!

girasquid:
:+1:

lib/active_merchant.rb
```diff
@@ -59,10 +59,10 @@ def Base64.strict_encode64(v)
 require 'timeout'
 require 'socket'
 
-require 'active_utils/common/network_connection_retries'
-silence_warnings{require 'active_utils/common/connection'}
-require 'active_utils/common/post_data'
-require 'active_utils/common/posts_data'
+require 'active_merchant/network_connection_retries'
+silence_warnings { require 'active_merchant/connection' }
```
ntalbott: Is this still needed?

ntalbott:
This is awesome. :heart: :+1: :heart: :+1: :heart:

lib/active_merchant.rb
wvanbergen: I tried but was getting a bunch of warnings. They can potentially be fixed.

ntalbott: I'd prefer fixing them; I can't even remember what they are?

RichardBlair:
lgtm

lib/active_merchant.rb
wvanbergen: I fixed the warnings - they were caused by Connection, and NetworkConnectionRetries both defining a `retry_safe` accessor.

wvanbergen:
This is basically good to go now. Before I can merge:

- Do we want to keep the `offsite_payment` shim?
- offsite_payments needs to be fixed first: https://github.com/Shopify/offsite_payments/pull/103 and https://github.com/Shopify/offsite_payments/pull/106.

@Shopify/payments @ntalbott

ntalbott:
I say go ahead and ditch `offsite_payments`. It'd be ideal to do so at a major version bump, but it's not as widely used, it's an easy switch for those that are using it, and we've been throwing deprecation warnings for months now.

lib/active_merchant.rb
```diff
@@ -59,50 +59,13 @@ def Base64.strict_encode64(v)
...
-          require "active_merchant/offsite_payments_shim"
-          ActiveMerchant::OffsitePaymentsShim
-        else
-          super
-        end
-      end
-    end
-  end
-
-  def self.deprecated(message, caller=Kernel.caller[1])
```
ntalbott: So `deprecated` isn't used anywhere else?

ntalbott:
There's an empty `lib/post_data.rb` file in this changeset?

wvanbergen:
whoops. deleting.

wvanbergen:
Anther small fix: https://github.com/Shopify/active_merchant/pull/1524

lib/active_merchant.rb
wvanbergen: apparently not, and if so, we should use `ActiveSupport::Deprecation` instead of inventing our own IMHO.

wvanbergen: Actually I think it is used elsewhere. I'll leave it for now and remove it in a separate PR.

ntalbott: @wvanbergen except my end goal is remove the activesupport dependency altogether, so I'd vote for keeping it. I also prefer it's output to the one in AS.

j-mutter:
:+1:

ntalbott:
@wvanbergen so I've given https://github.com/Shopify/active_utils/compare/v2.2.3...13b4a70 a review, and here are the bits that are missing:

* https://github.com/Shopify/active_utils/commit/eb090db626487cf17d98b9bf0f292e4e5ab7808e
* https://github.com/Shopify/active_utils/commit/f634423d1f562b29b71d91cab215108d8e83f936

Definitely want to make sure those get in before this is merged.

wvanbergen:
Ported those commits.

Today I will try if I can get all of the updated gems to work inside Shopify.

PR #1516, opened 2015-01-08
Update Travis CI build matrix

wvanbergen:
This updates the Travis CI build matrix.

- Also test against Ruby 2.2
- Tests more explicitly specified versions of Rails / ActiveSupport.

CircleCI will test against the latest dependency versions that meet the gemspec. We should see if we can open source this project there; I think this is now supported.

@j-mutter @girasquid @aprofeit @ntalbott for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,20 @@
+language: ruby
+script: "bundle exec rake test:units"
+sudo: false
+
 rvm:
-  - 2.1.0
-  - 2.0.0
-  - 1.9.3
+- "2.2"
+- "2.1"
+- "2.0"
+- "1.9"
 
 gemfile:
-  - Gemfile
-  - Gemfile_rails40
-
-install: "bundle install"
-script: "bundle exec rake test:units"
+- Gemfile.rails32
+- Gemfile.rails40
+- Gemfile.rails41
+- Gemfile.rails42
 
 notifications:
   email:
-    - payments-dev@shopify.com
-    - nathaniel@talbott.ws
+  - payments-dev@shopify.com
+  - nathaniel@talbott.ws
--- a/Gemfile.rails32
+++ b/Gemfile.rails32
@@ -1,10 +1,12 @@
-group :test do
-  gem 'json-jruby', :platforms => :jruby
-  gem 'jruby-openssl', :platforms => :jruby
-end
+source 'https://rubygems.org'
+gemspec
+
+gem 'jruby-openssl', :platforms => :jruby
 
 group :test, :remote_test do
   # gateway-specific dependencies, keeping these gems out of the gemspec
   gem 'braintree', '>= 2.0.0'
   gem 'oauth', '0.4.7'
 end
+
+gem 'rails', '~> 3.2.0'
--- a/Gemfile.rails42
+++ b/Gemfile.rails42
@@ -8,3 +8,5 @@ group :test, :remote_test do
   gem 'braintree', '>= 2.0.0'
   gem 'oauth', '0.4.7'
 end
+
+gem 'rails', '~> 4.2.0'
--- a/activemerchant.gemspec
+++ b/activemerchant.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency("offsite_payments", "~> 2.0.0")
 
   s.add_development_dependency('rake')
+  s.add_development_dependency('test-unit', '~> 1.2.3')
   s.add_development_dependency('mocha', '~> 0.13.0')
   s.add_development_dependency('rails', '>= 3.2.14')
   s.add_development_dependency('thor')
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,22 +1,13 @@
 #!/usr/bin/env ruby
 $:.unshift File.expand_path('../../lib', __FILE__)
 
-begin
-  require 'rubygems'
-  require 'bundler'
-  Bundler.setup
-rescue LoadError => e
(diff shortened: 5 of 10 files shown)


.travis.yml
```diff
@@ -1,16 +1,20 @@
+language: ruby
+script: "bundle exec rake test:units"
+sudo: false
```
wvanbergen: This enables the new infrastructure on Travis, which much faster build initialization times.

Gemfile.rails42
```diff
@@ -0,0 +1,12 @@
+source 'https://rubygems.org'
+gemspec
+
+gem 'jruby-openssl', :platforms => :jruby
+
+group :test, :remote_test do
+  # gateway-specific dependencies, keeping these gems out of the gemspec
+  gem 'braintree', '>= 2.0.0'
+  gem 'oauth', '0.4.7'
+end
```
wvanbergen: Should we set these as development dependencies instead? 

(Side note: the oauth gem is outdated and generates a lot of warnings.)

Gemfile.rails32
```diff
@@ -1,10 +1,12 @@
-group :test do
-  gem 'json-jruby', :platforms => :jruby
```
wvanbergen: I removed this because it's part of Ruby's standard library, and the `json` gem is supported directly in JRuby these days.

Gemfile.rails42
girasquid: Technically you only need them if you're working on the braintree or QBP gateways - or running their tests, so I think it makes sense to leave these as test dependencies for now.

test/test_helper.rb
```diff
@@ -1,22 +1,13 @@
...
 
 require 'test/unit'
 
 require 'mocha/version'
-if(Mocha::VERSION.split(".")[1].to_i < 12)
-  require 'mocha'
-else
-  require 'mocha/setup'
-end
+require 'mocha/setup'
```
girasquid: Why this change? If it's needed - do you still need to require `mocha/version` in the line above?

wvanbergen: The gemspec requires `mocha ~> 0.13.0`, so `(Mocha::VERSION.split(".")[1].to_i` will always be 13.

Forgot remove the version line.

girasquid: Oh ok - :cool:

girasquid:
The Circle failures look like something is missing:
```
NoMethodError: undefined method `refute' for test_supports_scrubbing?(GatewayTest):GatewayTest

NoMethodError: undefined method `refute' for test_should_not_allow_scrubbing_if_unsupported(GatewayTest):GatewayTest
```

wvanbergen:
I should have noted this PR is a WIP :)

girasquid:
Oh ok - let me know when you're ready for review.

Gemfile.rails42
wvanbergen: In that case we should add them as development dependencies in the gemspec. Development dependencies will not be installed when doing `gem install active_merchant`, so they will only be installed for active_merchant development.

activemerchant.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency("offsite_payments", "~> 2.0.0")
 
   s.add_development_dependency('rake')
+  s.add_development_dependency('test-unit', '~> 1.2.3')
```
wvanbergen: I added this as an explicit dependency. 
- 1.2.3 is outdated, but was the version that was bundled with Ruby 1.8
- Ruby 1.9 bundles Minitest instead, but includes a shim so Test::Unit tests still work.
- Ruby 2.2 drops the Test:Unit shim.

So, in order to reliably test against different Ruby versions, we have to explicitly specify either `test-unit` or `minitest`. Because we are using `Test::Unit::TestCase` subclasses for tests, and Minitest has version conflicts with different activesupport versions, I decided on using test-unit.

wvanbergen:
This is ready for review now! @Shopify/payments @ntalbott 

Once this is merged, I want to see if I can bump the mocha and test-unit versions because both are quite outdated.

Gemfile.rails42
ntalbott: @wvanbergen I think it's best to keep them in the `Gemfile`, not add them as dev dependencies. It makes it easier to play around with versions, especially since we still have a Rails dependency and could run into compatibility issues. I think of dev dependencies being for dev tools, whereas this is a special case of a library that's only needed if you want to use a particular gateway.

activemerchant.gemspec
ntalbott: We don't support Ruby 1.8 anymore, right? Not saying that changes the decision, but maybe it means we can use a newer version of the `test-unit` gem?

.travis.yml
ntalbott: :cool:

activemerchant.gemspec
wvanbergen: We could, I first one to merge this, and try upgrading test-unit and mocha (also outdated) afterwards.

ntalbott: @wvanbergen :+1:

wvanbergen:
Going to merge this tomorrow morning unless I hear objections before that time :P

Gemfile.rails42
wvanbergen: Cool, I'll leave it like this.

I'd like to get rid of the braintree gem, and maybe standadize on a shared oauth gem that is not shitty. But not now.

ntalbott: I hate the braintree gem with the :fire: passion of a thousand :sunny:. So I'd be all for that.

PR #1517, opened 2015-01-08
Clean up some tests

wvanbergen:
- When using mocha's `with` parameter matcher, you should not use assertions inside the block, but return true or false.
- Using uninitialized instance variables has different effect in different ruby versions. Make sure `@options` is set before using it in `@gateway.verify(@credit_card, @options)` calls.

Currently, these issues do not make the test suit fail due to the magic Ruby + mocha + test-unit/minitest version combination, but it fails when I change those.

@Shopify/payments @ntalbott for review


j-mutter:
:hatched_chick:

PR #1518, opened 2015-01-08
Drop init.rb

wvanbergen:
`init.rb` is only needed for people that have installed activemerchant in Rails 2 as a rails plugin, and are not using bundler.

@ntalbott @Shopify/payments

### DIFF
--- a/init.rb
+++ b/init.rb
@@ -1,1 +1,0 @@
-require 'active_merchant'


girasquid:
:+1:

j-mutter:
:hatched_chick:

PR #1520, opened 2015-01-08
[QuickBooks Payments] remove oauth gem, implement OAuth request signing in QuickbooksGateway

bizla:
Upon inspection of the `oauth` gem at https://github.com/oauth-xx/oauth-ruby, we noticed [the default HMAC-SHA1 mechanism](https://github.com/oauth-xx/oauth-ruby/blob/master/lib/digest/hmac.rb) used to sign requests is [officially discouraged](http://www.ruby-doc.org/stdlib-2.0/libdoc/digest/rdoc/Digest/HMAC.html) and thus unacceptable for deployment in a PCI environment. 

This removes the gem dependency and isolates OAuth 1.0a request signing into the QuickBooks Payments adapter, and uses the OpenSSL implementation of HMAC instead, as suggested by the official Ruby documentation.

@girasquid @ntalbott /cc @Shopify/active-merchant

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -1,9 +1,3 @@
-begin
-  require "oauth"
-rescue LoadError
-  raise "You need to have the oauth gem installed in order to use the ActiveMerchant QuickbooksGateway adapter."
-end
-
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class QuickbooksGateway < Gateway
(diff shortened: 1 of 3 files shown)


bizla:
@ntalbott This should also get rid of that massive `rake test` noise we both saw.

jnormore:
Nice catch!

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -204,44 +198,36 @@ def post_data(data = {})
...
+        oauth_parameters = {
+          oauth_nonce: generate_unique_id,
+          oauth_timestamp: Time.now.to_i.to_s,
+          oauth_signature_method: 'HMAC-SHA1',
+          oauth_version: "1.0",
+          oauth_consumer_key: @options[:consumer_key],
+          oauth_token: @options[:access_token]
+        }
 
-        # required for Ruby 1.9 backwards compatibilty
```
girasquid: Why did this get removed?

wvanbergen:
:+1: I fixed all the warnings that would show up in test runs, except for the hundreds of warnings that are due to this gem. :fire:

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: This was needed to satisfy the gem's client helper interface. It's needed for the Signature Base String. Since we have the `method` param already, we can just use it directly and not have to construct any `URI`s.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -204,44 +198,36 @@ def post_data(data = {})
...
-          request_object,
-          consumer: consumer,
-          token: access_token,
-          realm: @options[:realm],
-          request_uri: request_uri)
-        oauth_header = client_helper.header
+        # prepare Authorization header string
+        oauth_parameters = Hash[oauth_parameters.sort_by {|k, _| k}]
+        oauth_header_string = "OAuth realm=\"#{@options[:realm]}\", "
+        oauth_parameters.each_key {|k| oauth_header_string << "#{k}=\"#{oauth_parameters[k]}\", "}
```
girasquid: Instead of doing this, would it be clearer if this used an array and `join`?
```ruby
oauth_headers = ["OAuth realm=\"#{@options[:realm]}\""]
oauth_headers += oauth_parameters.map {|k, v| "#{k}=\"#{v}\""}
oauth_header_string = oauth_headers.join(", ")
```

Doing it that way means you don't need the `chomp` below, as well.

lib/active_merchant/billing/gateways/quickbooks.rb
girasquid: Nevermind, I see now

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -204,44 +198,36 @@ def post_data(data = {})
...
 
-        # required for Ruby 1.9 backwards compatibilty
-        begin
-          request_object = request_class.new(request_uri)
-        rescue
-          request_object = request_class.new(uri)
-        end
+        # prepare components for signature
+        oauth_signature_base_string = [method.to_s.upcase, request_uri.to_s, oauth_parameters.to_param].map{|v| CGI.escape(v) }.join('&')
+        oauth_signing_key = "#{CGI.escape(@options[:consumer_secret])}&#{CGI.escape(@options[:token_secret])}"
```
girasquid: Can you change this to be an array that you join with `&`, to be consistent with the line above?

lib/active_merchant/billing/gateways/quickbooks.rb
girasquid: :cool:

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: :+1: Switching it up.

lib/active_merchant/billing/gateways/quickbooks.rb
bizla: Doh, I had that but didn't commit it. I'll amend.

girasquid: :heart_eyes:

wvanbergen:
Needs a rebase, but LGTM!

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -204,44 +198,36 @@ def post_data(data = {})
...
-        client_helper = OAuth::Client::Helper.new(
-          request_object,
-          consumer: consumer,
-          token: access_token,
-          realm: @options[:realm],
-          request_uri: request_uri)
-        oauth_header = client_helper.header
+        # prepare Authorization header string
+        oauth_parameters = Hash[oauth_parameters.sort_by {|k, _| k}]
+        oauth_headers = ["OAuth realm=\"#{@options[:realm]}\", "]
```
girasquid: I don't think you want the trailing `, ` here - when you join it with `, ` aren't you going to get duplicates?

bizla: Hmm, you're right, but the remote tests passed? heh I'll fix it.

girasquid:
:+1:

PR #1521, opened 2015-01-09
Remove duplicate hash keys that were causing warnings

wvanbergen:
When hash keys are redefined, the last occurrence will be kept.

- For skipjack, I kept the one that I could find in their docs: http://imgserver.skipjack.com/imgServer/5293710/AVS%20and%20CVV2.pdf
- WireCard and HPS were just duplicates, so I removed the last occurrence.

@ntalbott @Shopify/payments for review


bizla:
:sparkles: :+1: :sparkles:

PR #1522, opened 2015-01-09
Bump test/unit and mocha

wvanbergen:
Bump test-unit and mocha versions, as discussed in #1516.

@ntalbott @Shopify/payments


test/unit/gateways/jetpay_test.rb
```diff
@@ -12,7 +12,6 @@ def setup
     @amount = 100
 
     @options = {
-      :order_id => '1',
```
wvanbergen: This was also a duplicate hash key (see below on line 19).

bizla:
:+1:

PR #1523, opened 2015-01-09
Remove base64 strict backport as it is no longer needed

wvanbergen:
It ships with Ruby 1.9 and up.

@ntalbott @Shopify/payments

### DIFF
--- a/lib/active_merchant.rb
+++ b/lib/active_merchant.rb
@@ -36,22 +36,7 @@ end
 require 'active_support/core_ext/class/delegating_attributes'
 require 'active_support/core_ext/module/attribute_accessors'
 
-begin
-  require 'active_support/base64'
-
-  unless defined?(Base64)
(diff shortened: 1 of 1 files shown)


lib/active_merchant.rb
```diff
@@ -36,28 +36,13 @@
...
-end
-
+require 'base64'
 require 'securerandom'
-require 'builder'
 require 'cgi'
 require 'rexml/document'
 require 'timeout'
 require 'socket'
+require 'builder'
```
girasquid: Why move this here?

girasquid:
:+1:

lib/active_merchant.rb
wvanbergen: This is me not stashing my changes properly.

wvanbergen: Fixed.

ntalbott:
:cool: :+1:

PR #1524, opened 2015-01-09
Use class_attribute instead of superclass_delegating_accessor in posts_data

wvanbergen:
It has been deprecated, in lieu of `class_attribute`.

@ntalbott @Shopify/payments

### DIFF
--- a/lib/active_merchant/posts_data.rb
+++ b/lib/active_merchant/posts_data.rb
@@ -2,26 +2,26 @@ module ActiveMerchant #:nodoc:
   module PostsData  #:nodoc:
 
     def self.included(base)
-      base.superclass_delegating_accessor :ssl_strict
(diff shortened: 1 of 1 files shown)


wvanbergen:
This is already fixed in ActiveUtils 3, but I copied an old release of active_utils (2.x) so the copy in AM still is generating deprecating warnings when using activesupport 4.2

ntalbott:
@wvanbergen have you checked to make sure we didn't miss any other fixes in active_utils master?

j-mutter:
Nice catch, I thought these were all gone :+1:

wvanbergen:
@ntalbott May be worth it to go through https://github.com/Shopify/active_utils/compare/v2.2.3...13b4a70

PR #1525, opened 2015-01-09
[authorize.net] Fix ambiguous use of parenthesis in authorize.net gateway

wvanbergen:
- When generating XML, use parentheses when calling tag methods
- Do not use parentheses around if clauses.

In certain Ruby versions, some of these lines were causing warnings because the interpretation is ambiguous without parentheses.

@ntalbott @Shopify/payments


j-mutter:
:bow:

girasquid:
:heart_eyes: :+1:

PR #1530, opened 2015-01-14
Add integration of Robokassa into ActiveMerchant

fdezign:
Robokassa is the most popular and biggest payment gateway in Russia. We use Shopify for our shop and need to add Robokassa as a possible payment gateway. We've found an article of integration it: 
http://sysmagazine.com/posts/147435/
But Shopify is a CMS where I cannot manually add and integrate it. Could you integrate it?
I'm sure that all of russian users of Shopify will appreciate this innovation and you'll have much more customers!


ntalbott:
@fdezign we'd be willing to consider a PR to add it, but this PR seems to be pulling in massive amounts of non-related commits? Going to close; please feel free to re-open with a PR just adding Robokassa and we'll consider it for inclusion.

andrewpaliga:
@fdezign please direct integration requests to payment-integrations@shopify.com.

Also, please note that [Active Merchant](https://github.com/Shopify/active_merchant) only supports credit card integrations. To offer additional payments types such as wallets / direct banking integrationS you will need to integrate to Shopify using out [offsite gateway integration](https://github.com/Shopify/offsite-gateway-sim).

PR #1531, opened 2015-01-14
Don't use relative requires

wvanbergen:
A library shouldn't use relative requires. You can/should assume that the dependency manager (e.g. bundler) has set the `$LOAD_PATH` correctly to include `lib` (and `test` in test mode). Using relative requires can cause double loading of files in some cases, which can be a subtle source of bugs and warnings. 

If you really need to require a file relative to the current file, use `require_relative`, but we don't need it in this gem.


wvanbergen:
@ntalbott @Shopify/payments

wvanbergen:
For more info, see http://weblog.rubyonrails.org/2009/9/1/gem-packaging-best-practices/

girasquid:
lgtm

ntalbott:
Guess who hates relative requires?

This guy :arrow_right: @ntalbott

:heart:

PR #1532, opened 2015-01-14
Remove rails/init.rb that is no longer relevant for Rails 3+

wvanbergen:
I removed `init.rb` (see #1518) but forgot this ancient relic of the past.

@ntalbott @Shopify/payments


girasquid:
:+1:

PR #1533, opened 2015-01-14
Remove executable flags from files that do not ened it

wvanbergen:
A bunch of files had the executable bit set, which is not needed. Now only `script/generate` is executable.

@ntalbott @Shopify/payments


fw42:
:+1:

PR #1534, opened 2015-01-14
Orbital: Handle incorrectly encoded response

denis:
When an order contains non-standard characters (normally in the
customer's name and is normally accented characters) and is submitted
to Orbital, they will send back an XML response flagged as UTF-8 but is
actually ISO-8859-1. This causes parsers like REXML to raise an
exception while parsing what it thinks is UTF-8.

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -427,7 +427,7 @@ module ActiveMerchant #:nodoc:
 
       def parse(body)
         response = {}
-        xml = REXML::Document.new(body)
+        xml = REXML::Document.new(body.encoding.name == "UTF-8" ? body : body.encode("UTF-8", "ISO-8859-1"))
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -427,7 +427,7 @@ def add_managed_billing(xml, options)
 
       def parse(body)
         response = {}
-        xml = REXML::Document.new(body)
+        xml = REXML::Document.new(body.encoding.name == "UTF-8" ? body : body.encode("UTF-8", "ISO-8859-1"))
```
wvanbergen: Are you certain `ISO-8859-1` is always the other encoding? Why not just `body.encode("UTF-8")`?

ntalbott:
Closing this since it's been a month with no movement. Feel free to re-open if you get a chance to work on it further. Thanks!

PR #1535, opened 2015-01-15
Autoload gateways with require paths relative to load path

wvanbergen:
This registers the autoload paths of gateways relative to the load path, instead of as an absolute path.
E.g. `active_merchant/billing/gateways/viaklix` vs `/Users/.../active_merchant/lib/active_merchant/billing/gateways/viaklix.rb`

I am not a big fan of the autoloading setup, but this should be 100% compatible.

I also always require `Gateway` instead of autoloading it, because literally every gateway needs it.

@ntalbott @Shopify/active-merchant for review.


j-mutter:
:hatched_chick:

ntalbott:
:egg:-cellent.

wvanbergen:
Merged, and released as 1.46.0

PR #1536, opened 2015-01-15
Update compatibility.rb

stephenbaldwin:
This is an instance variable - variable being used for rails compatibility is a class instance variable. No need for this


ntalbott:
@stephenbaldwin both are class instance variables...

This code is going away soon in v2 regardless, so not going to worry about it at this point.

PR #1537, opened 2015-01-15
Add in method to calculate tax

stephenbaldwin:
This isn't a standard ActiveMerchant method, but should be included to allow for preview of tax charges (Vindicia automatically appends tax to orders)


ntalbott:
@stephenbaldwin this just isn't a fit for ActiveMerchant - it's totally out of scope. Probably best as a fully standalone gem, or a gem that monkey-patches AM to add it in for those that want it.

PR #1540, opened 2015-01-16
add support for partial refunds + captures

ivanfer:
#### QBP updates

This PR adds support for partial captures and refunds, and makes remote tests a little more explicit.

Thought we would be able to add in voids for Quickbooks Payments in this PR, but without passing remote tests for them, it's a feature we'll add later.

for review: @bizla @girasquid @ntalbott

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -75,16 +75,18 @@ module ActiveMerchant #:nodoc:
       end
 
       def capture(money, authorization, options = {})
+        post = {}
         capture_uri = "#{ENDPOINT}/#{CGI.escape(authorization)}/capture"
+        post[:amount] = localized_amount(money, currency(money))
 
-        commit(capture_uri)
+        commit(capture_uri, post)
       end
 
       def refund(money, authorization, options = {})
         post = {}
-        refund_uri = "#{ENDPOINT}/#{CGI.escape(authorization)}/refunds"
+        post[:amount] = localized_amount(money, currency(money))
 
-        commit(refund_uri, post)
+        commit(refund_uri(authorization), post)
       end
 
       def verify(credit_card, options = {})
@@ -234,7 +236,7 @@ module ActiveMerchant #:nodoc:
       def cvv_code_from(response)
         if response['errors'].present?
           FRAUD_WARNING_CODES.include?(response['errors'].first['code']) ? 'I' : ''
-        elsif
+        else
           success?(response) ? 'M' : ''
         end
       end
@@ -266,6 +268,10 @@ module ActiveMerchant #:nodoc:
           raise response_error
         end
         response_error.response.body
+      end
+
+      def refund_uri(authorization)
+        "#{ENDPOINT}/#{CGI.escape(authorization)}/refunds"
       end
     end
   end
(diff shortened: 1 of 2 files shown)


test/remote/gateways/remote_quickbooks_test.rb
```diff
@@ -66,7 +70,9 @@ def test_partial_refund
     purchase = @gateway.purchase(@amount, @credit_card, @options)
     assert_success purchase
 
-    assert refund = @gateway.refund(@amount-1, purchase.authorization)
+    assert refund = @gateway.refund(@partial_amount, purchase.authorization)
+    binding.pry
```
ntalbott: Probably don't want that in there anymore :grin:

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -75,16 +75,22 @@ def authorize(money, payment, options = {})
...
       def refund(money, authorization, options = {})
         post = {}
-        refund_uri = "#{ENDPOINT}/#{CGI.escape(authorization)}/refunds"
+        post[:amount] = localized_amount(money, currency(money))
+
+        commit(refund_or_void_uri(authorization), post)
+      end
 
-        commit(refund_uri, post)
+      def void(authorization, options = {})
```
ntalbott: Thought you weren't adding in `void`?

ivanfer: indeed :sweat:

test/remote/gateways/remote_quickbooks_test.rb
```diff
@@ -7,6 +7,8 @@ def setup
     @credit_card = credit_card('4000100011112224')
     @declined_card = credit_card('4000000000000001')
 
+    @partial_amount = @amount-1
```
girasquid: spaces around `-` please

ivanfer:
Updated with your comments - good to go?

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -267,6 +269,10 @@ def extract_response_body_or_raise(response_error)
         end
         response_error.response.body
       end
+
+      def refund_or_void_uri(authorization)
```
girasquid: You're only using this for refunds - why pull it out into a message?

ivanfer: voids and refunds hit the same endpoint - once we get confirmation from Intuit that they work in sandbox, it will look like this. YAGNI?

girasquid: YAGNI for now, imo

ivanfer: cool, done.

girasquid:
:+1:

ivanfer:
Thanks for the review!

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -267,6 +269,10 @@ def extract_response_body_or_raise(response_error)
         end
         response_error.response.body
       end
+
+      def refund_uri(authorization)
```
j-mutter: Is there a reason this was extracted out into a method but `capture_uri` isn't?

ivanfer: only that  ``` capture ``` alone uses this uri but the refund uri is to be shared with voids, which is coming later. I can factor this out then too

PR #1541, opened 2015-01-19
Fixed typo in Redsys gateway's error messages mapping

agseco:
Just found the error message in our application's logs and realised the typo; I decided to fix it straight away.


girasquid:
Thanks!

PR #1545, opened 2015-01-26
Add finalize option to quickpay at capture

askehansen:
Add the ability to finalize the transaction when doing a partial capture so the transaction doesn't remain open

http://doc.quickpay.dk/api/messagetypes.html#index6h2

### DIFF
--- a/lib/active_merchant/billing/gateways/quickpay.rb
+++ b/lib/active_merchant/billing/gateways/quickpay.rb
@@ -19,7 +19,7 @@ module ActiveMerchant #:nodoc:
                            currency autocapture cardnumber expirationdate
                            cvd cardtypelock testmode),
 
-          :capture   => %w(protocol msgtype merchant amount transaction),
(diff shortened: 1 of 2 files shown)


duff:
Hi @askehansen,

I'm thinking a unit test verifying that we're passing the right value for each of the cases would be helpful for this pull request.  Many of the gateways have unit tests using `CommStub` to assert such things.

Thanks!

askehansen:
Im not quite sure what you mean?

duff:
Here's an example in the Stripe unit tests verifying that the right things get passed to the gateway:

https://github.com/Shopify/active_merchant/blob/f8998d6747a9d99fc727731d4f1c9e9690001a3e/test/unit/gateways/stripe_test.rb#L506-L512

Does that help explain?

askehansen:
Yea thanks, i'll do that

askehansen:
I just added some tests

duff:
Looks great @askehansen!  Just merged it.  Thanks!

PR #1546, opened 2015-01-26
Implement transaction scrubbing for Authorize.Net

girasquid:
@duff @j-mutter for review.

I'd like some feedback on what to leave inside the remote tests - right now I just have the commented out `dump_transcript_and_fail` because I'd like it to be easy to regenerate a transcript when you need one, but I'm open to suggestions.

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -123,6 +123,16 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def supports_scrubbing?
+        true
+      end
+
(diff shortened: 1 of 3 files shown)


j-mutter:
Code looks good.

For the transcript dump, could we make it a rake task or something maybe?

girasquid:
I did some checking into making a rake task, and it turned out to be pretty non-trivial. For now I think I'm just going to stick to having it live in the remote tests, which already have the right credentials + options for the call.

j-mutter:
okay.jpg

girasquid:
I merged this but forgot to update the commit message - closing.

PR #1549, opened 2015-01-28
Add proxy address and port configuration

arkes:
Add access to "proxy_address" and "proxy_port"


j-mutter:
LGTM. The functionality is already there in the `Connection` class. @ntalbott @girasquid any objections?

girasquid:
None here.

ntalbott:
I'm fine with it.

PR #1550, opened 2015-01-30
Translate 'Unsupported' in AVS responses to S

girasquid:
@j-mutter @ivanfer for review.

MiGS returns the string "Unsupported" for their AVSResultCode when they don't support AVS. That's not even a code!

This change makes it so that the code gets converted into 'S', which is the code for "Bank does not support AVS" according to [the wikipedia page](http://en.wikipedia.org/wiki/Address_Verification_System).

### DIFF
--- a/lib/active_merchant/billing/gateways/migs.rb
+++ b/lib/active_merchant/billing/gateways/migs.rb
@@ -220,11 +220,13 @@ module ActiveMerchant #:nodoc:
       end
 
       def response_object(response)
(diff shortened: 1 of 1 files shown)


ivanfer:
Someone did a bad.
:palm_tree:

j-mutter:
Ugh, really? ಠ_ಠ

j-mutter:
But yeah, :shipit:

PR #1552, opened 2015-02-02
Pay gate ip and email update

StephanAtG2:
Client requires that payprotect be used in the authtx. I have added this and updated the test cases to include this. Will confirm with paygate on what the behavior is when the email and ip is sent to paygate but not activated on the account.

### DIFF
--- a/lib/active_merchant/billing/gateways/pay_gate_xml.rb
+++ b/lib/active_merchant/billing/gateways/pay_gate_xml.rb
@@ -194,8 +194,6 @@ module ActiveMerchant #:nodoc:
         SUCCESS_CODES.include?(response[:res])
       end
 
-
-
       def build_request(action, options={})
         xml = Builder::XmlMarkup.new
         xml.instruct!
--- a/lib/active_merchant/billing/gateways/pay_gate_xml.rb
+++ b/lib/active_merchant/billing/gateways/pay_gate_xml.rb
@@ -229,7 +229,9 @@ module ActiveMerchant #:nodoc:
           :budp  => 0,
           :amt   => amount(money),
           :cur   => (options[:currency] || currency(money)),
-          :cvv   => creditcard.verification_value
+          :cvv   => creditcard.verification_value,
+          :email => options[:email],
+          :ip    => options[:ip]
         }
       end
 
--- a/lib/active_merchant/billing/gateways/pay_gate_xml.rb
+++ b/lib/active_merchant/billing/gateways/pay_gate_xml.rb
@@ -193,7 +193,9 @@ module ActiveMerchant #:nodoc:
       def successful?(response)
         SUCCESS_CODES.include?(response[:res])
       end
-      
+
+
+
       def build_request(action, options={})
         xml = Builder::XmlMarkup.new
         xml.instruct!
--- a/lib/active_merchant/billing/gateways/pay_gate_xml.rb
+++ b/lib/active_merchant/billing/gateways/pay_gate_xml.rb
@@ -181,11 +181,20 @@ module ActiveMerchant #:nodoc:
         commit(action, build_request(action, options))
       end
 
+      def refund(money, authorization, options={})
+        action = 'refundtx'
+
+        options.merge!(:money => money, :authorization => authorization)
+        commit(action, build_request(action, options))
+      end
+
       private
 
       def successful?(response)
         SUCCESS_CODES.include?(response[:res])
       end
+
+
 
       def build_request(action, options={})
         xml = Builder::XmlMarkup.new
@@ -201,6 +210,10 @@ module ActiveMerchant #:nodoc:
               money           = options.delete(:money)
               authorization   = options.delete(:authorization)
               build_capture(protocol, money, authorization, options)
+            when 'refundtx'
+              money           = options.delete(:money)
+              authorization   = options.delete(:authorization)
+              build_refund(protocol, money, authorization, options)
           else
             raise "no action specified for build_request"
           end
@@ -225,6 +238,12 @@ module ActiveMerchant #:nodoc:
       def build_capture(xml, money, authorization, options={})
         xml.tag! 'settletx', {
           :tid => authorization
+        }
+      end
+      def build_refund(xml, money, authorization, options={})
+        xml.tag! 'refundtx', {
+          :tid => authorization,
+          :amt => amount(money)
         }
       end
 
--- a/lib/active_merchant/billing/gateways/pay_gate_xml.rb
+++ b/lib/active_merchant/billing/gateways/pay_gate_xml.rb
@@ -178,7 +178,7 @@ module ActiveMerchant #:nodoc:
         action = 'settletx'
 
         options.merge!(:money => money, :authorization => authorization)
-        commit_capture(action, build_request(action, options), authorization)
+        commit(action, build_request(action, options), authorization)
       end
 
       def refund(money, authorization, options={})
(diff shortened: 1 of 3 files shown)


StephanAtG2:
Contacted pay gate. Adding ip and email always will not affect the api behavior

test/remote/gateways/remote_pay_gate_xml_test.rb
```diff
@@ -52,4 +54,13 @@ def test_invalid_login
     assert_failure response
     assert_equal 'Incorrect Credentials Supplied', response.message
   end
+
+   def test_purchase_and_full_credit
```
bizla: Please fix indentation here.

test/unit/gateways/pay_gate_xml_test.rb
```diff
@@ -38,6 +41,15 @@ def test_successful_settlement
     assert response.test?
   end
 
+  def test_successful_refund
+    @gateway.expects(:ssl_post).returns(successful_refund_response)
+
+    assert response = @gateway.refund(@amount, '16996548', @options)
+    assert_instance_of Response, response
```
bizla: This isn't required on account of the subsequent `assert_success`.

lib/active_merchant/billing/gateways/pay_gate_xml.rb
```diff
@@ -227,6 +240,12 @@ def build_capture(xml, money, authorization, options={})
           :tid => authorization
         }
       end
+      def build_refund(xml, money, authorization, options={})
```
bizla: Add a newline above this.

lib/active_merchant/billing/gateways/pay_gate_xml.rb
```diff
@@ -251,6 +270,13 @@ def commit(action, request)
           :authorization  => response[:tid]
         )
       end
+      def commit_capture(action, authorization, request)
```
bizla: Add a newline above this.

bizla: This method is largely a duplication of `#commit`. If you want to add the ability to pass in authorizations, augment the `#commit` method rather than duplicating it. See other integrations for examples of this.

lib/active_merchant/billing/gateways/pay_gate_xml.rb
bizla: Add a newline above this.

lib/active_merchant/billing/gateways/pay_gate_xml.rb
```diff
@@ -201,6 +208,10 @@ def build_request(action, options={})
               money           = options.delete(:money)
               authorization   = options.delete(:authorization)
               build_capture(protocol, money, authorization, options)
+            when 'refundtx'
+              money           = options.delete(:money)
+              authorization   = options.delete(:authorization)
+              build_refund(protocol, money, authorization, options)
```
bizla: The block
```ruby 
money = options.delete(:money)
authorization = options.delete(:authorization)
``` 
is repeated unnecessarily inside each case block. I'd suggest you take the opportunity to clean this up, i.e. move these lines above the `case` statement such that it looks like: 
```ruby 
when 'authtx'
  build_authorization(...) 
when 'settletx' 
  build_capture(...)
when 'refundtx' 
  build_refund(...) 
```

bizla:
See inline comments. I'd prefer to a bit more polish and a commit squash before we merge upstream.

test/remote/gateways/remote_pay_gate_xml_test.rb
ivanfer: Also, this remote test is doing a purchase and refund, renaming this to ``` test_successful_purchase_and_refund ``` would bring the wording in line with the expectation of the test a little more

StephanAtG2:
Hi thanks. 

Heck of alota comments for only 2 lines of code changed. 

Anyways expect a commit soon

AndyOliverWpg:
@StephanAtG2 Your PR shows 62 lines changed. Does this Pull Request include more code than you wanted to add to the repo?

StephanAtG2:
I think its because i continued on my previous branch and just submitted an new pull request for the last changes.

Should i have created a new branch for these changes?

bizla:
Yeah. Pull requests are based off the diff between branches, so if you intended on a 2-line change, then your diff between master and your branch should be 2 lines.

StephanAtG2:
Its fine. Most of these changes is still valid thou. Im just going to fix the ones mentioned above and recommit.

lib/active_merchant/billing/gateways/pay_gate_xml.rb
StephanAtG2: One question.

The first commit that runs gets the auth code that is needed in the capture request call. how do i successfully merge this 2 functions into one?

StephanAtG2: Another note. Only reason that function was split up like that is because the refundtx request requires the tid from the authrx response and not the capturerx response. So i had to make sure that the purchase function returns the authrx's tid instead of the capturerx because the responses tid are infact not the same at paygate

bizla:
Sounds good -- ping when you'd like a review.

StephanAtG2:
Please note. I have added comments to a request you made. The one about the duplicate function

lib/active_merchant/billing/gateways/pay_gate_xml.rb
ivanfer: You can simply have an extra param in ``` commit ``` called auth, if present use that instead of ``` response[:tid] ```

bizla: I'd suggest you look through a mature integration like https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb for ideas on how to structure your code to cleanly handle many cases.

StephanAtG2:
Hi i have made the update. Can you guys please review

lib/active_merchant/billing/gateways/pay_gate_xml.rb
```diff
@@ -244,14 +261,14 @@ def parse(action, body)
         hash
       end
 
-      def commit(action, request)
+      def commit(action, request, authorization = nil)
         response = parse(action, ssl_post(self.live_url, request))
         Response.new(successful?(response), message_from(response), response,
           :test           => test?,
-          :authorization  => response[:tid]
+          :authorization  => authorization ? authorization : response[:tid]
```
bizla: :+1: sweet

lib/active_merchant/billing/gateways/pay_gate_xml.rb
```diff
@@ -192,15 +199,16 @@ def build_request(action, options={})
         xml.instruct!
 
         xml.tag! 'protocol', :ver => API_VERSION, :pgid => (test? ? TEST_ID : @options[:login]), :pwd => @options[:password] do |protocol|
+          money       = options.delete(:money)
+          authorization   = options.delete(:authorization)
+          creditcard  = options.delete(:creditcard)
```
bizla: I think you intended to line up the `=` here? Sorry to be so picky about whitespace. ;)

bizla:
Other than the whitespace comment, this looks okay to me. 
Could you [squash this down to 1 commit](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) with a descriptive commit message? Something like `[PayGate] Add refund processing, accept :email and :ip to authorization parameters`.

StephanAtG2:
I cannot for the life of me get these commits to squash. Its like it squashes but resets back to the original commits. its cool thou gonna fork master again and add changes to a new fork and submit a new pull request. is that fine

girasquid:
Yup, that works. Going to close this one and wait for your squashed one.

PR #1555, opened 2015-02-03
[MiGS] Return P for CVV results when Unsupported

girasquid:
@j-mutter @ivanfer for review.

MiGS returns the string "Unsupported" for their CSCResultCode (CVV check) when they don't support doing the check. That's not even a code!

This change makes it so that in that situation the code is translated to be 'P', which according to [this page](http://www.emsecommerce.net/avs_cvv2_response_codes.htm) means "Not Processed".


ivanfer:
:+1:

j-mutter:
Ughghghghghghg

:shipit:

PR #1556, opened 2015-02-05
Implementation of Monei.net gateway for AM

leolara:
This an implementation of the Monei.net gateway for active merchant.

The goal is to be a Shopify gateway for Spain.

@alexandresaiz (Alexandre Saiz) has been talking about this from the business said with Louis Kearns of Shopify.

Please, have a look and let me know if there are any problems with the code.

### DIFF
--- a/lib/active_merchant/billing/gateways/monei.rb
+++ b/lib/active_merchant/billing/gateways/monei.rb
@@ -226,7 +226,7 @@ module ActiveMerchant #:nodoc:
             xml.Street address[:address1].to_s
             xml.Zip address[:zip].to_s
             xml.City address[:city].to_s
-            xml.State address[:state] if address.has_key? :state
+            xml.State address[:state].to_s if address.has_key? :state
             xml.Country address[:country].to_s
           end
           xml.Contact do
--- a/lib/active_merchant/billing/gateways/monei.rb
+++ b/lib/active_merchant/billing/gateways/monei.rb
@@ -190,7 +190,7 @@ module ActiveMerchant #:nodoc:
 
       # Private: Add payment part to XML
       def add_payment(xml, action, money, options)
-        code = tanslate_payment_code action
+        code = tanslate_payment_code(action)
 
         xml.Payment(:code => code) do
           xml.Presentation do
@@ -243,8 +243,8 @@ module ActiveMerchant #:nodoc:
 
         {
             :unique_id => xml.xpath("//Response/Transaction/Identification/UniqueID").text,
-            :status => (translate_status_code xml.xpath("//Response/Transaction/Processing/Status/@code").text),
-            :reason => (translate_status_code xml.xpath("//Response/Transaction/Processing/Reason/@code").text),
+            :status => translate_status_code(xml.xpath("//Response/Transaction/Processing/Status/@code").text),
+            :reason => translate_status_code(xml.xpath("//Response/Transaction/Processing/Reason/@code").text),
             :message => xml.xpath("//Response/Transaction/Processing/Return").text
         }
       end
--- a/lib/active_merchant/billing/gateways/monei.rb
+++ b/lib/active_merchant/billing/gateways/monei.rb
@@ -190,7 +190,7 @@ module ActiveMerchant #:nodoc:
 
       # Private: Add payment part to XML
       def add_payment(xml, action, money, options)
-        code = tr_payment_code action
+        code = tanslate_payment_code action
 
         xml.Payment(:code => code) do
           xml.Presentation do
@@ -243,8 +243,8 @@ module ActiveMerchant #:nodoc:
 
         {
             :unique_id => xml.xpath("//Response/Transaction/Identification/UniqueID").text,
-            :status => (tr_status_code xml.xpath("//Response/Transaction/Processing/Status/@code").text),
-            :reason => (tr_status_code xml.xpath("//Response/Transaction/Processing/Reason/@code").text),
+            :status => (translate_status_code xml.xpath("//Response/Transaction/Processing/Status/@code").text),
+            :reason => (translate_status_code xml.xpath("//Response/Transaction/Processing/Reason/@code").text),
             :message => xml.xpath("//Response/Transaction/Processing/Return").text
         }
       end
@@ -285,7 +285,7 @@ module ActiveMerchant #:nodoc:
       end
 
       # Private: Translate Monei status code to native ruby symbols
-      def tr_status_code(code)
+      def translate_status_code(code)
         {
             '00' => :success,
             '40' => :neutral,
@@ -300,7 +300,7 @@ module ActiveMerchant #:nodoc:
       end
 
       # Private: Translate AM operations to Monei operations codes
-      def tr_payment_code(action)
+      def tanslate_payment_code(action)
         {
             :purchase => 'CC.DB',
             :authorize => 'CC.PA',
--- a/lib/active_merchant/billing/gateways/monei.rb
+++ b/lib/active_merchant/billing/gateways/monei.rb
@@ -43,8 +43,8 @@ module ActiveMerchant #:nodoc:
       #               :billing_address  Hash with billing address information
       #               :description      Merchant created purchase description (optional)
       #               :currency         Sale currency to override money object or default (optional)
-      #               :given            Customer given name to override card holder's (optional)
-      #               :family           Customer family name to override card holder's (optional)
+      #               :first_name       Customer first name to override card holder's (optional)
+      #               :last_name        Customer family name to override card holder's (optional)
       #
       # Returns Active Merchant response object
       def purchase(money, credit_card, options)
@@ -60,8 +60,8 @@ module ActiveMerchant #:nodoc:
       #               :billing_address  Hash with billing address information
       #               :description      Merchant created authorization description (optional)
       #               :currency         Sale currency to override money object or default (optional)
-      #               :given            Customer given name to override card holder's (optional)
-      #               :family           Customer family name to override card holder's (optional)
+      #               :first_name       Customer first name to override card holder's (optional)
+      #               :last_name        Customer family name to override card holder's (optional)
       #
       # Returns Active Merchant response object
       def authorize(money, credit_card, options={})
@@ -119,8 +119,8 @@ module ActiveMerchant #:nodoc:
       #               :billing_address  Hash with billing address information
       #               :description      Merchant created authorization description (optional)
       #               :currency         Sale currency to override money object or default (optional)
-      #               :given            Customer given name to override card holder's (optional)
-      #               :family           Customer family name to override card holder's (optional)
+      #               :first_name       Customer first name to override card holder's (optional)
+      #               :last_name        Customer family name to override card holder's (optional)
       #
       # Returns Active Merchant response object of Authorization operation
       def verify(credit_card, options={})
@@ -219,8 +219,8 @@ module ActiveMerchant #:nodoc:
         xml.Customer do
 
           xml.Name do
-            xml.Given options[:given] || credit_card.first_name
-            xml.Family options[:family] || credit_card.last_name
+            xml.Given options[:first_name] || credit_card.first_name
+            xml.Family options[:last_name] || credit_card.last_name
           end
           xml.Address do
             xml.Street address[:address1].to_s
--- a/lib/active_merchant/billing/gateways/monei.rb
+++ b/lib/active_merchant/billing/gateways/monei.rb
@@ -266,7 +266,7 @@ module ActiveMerchant #:nodoc:
 
       # Private: Decide success from servers response
       def success_from(response)
-        response[:status] == :success or response[:status] == :new
+        response[:status] == :success || response[:status] == :new
       end
 
       # Private: Get message from servers response
--- a/lib/active_merchant/billing/gateways/monei.rb
+++ b/lib/active_merchant/billing/gateways/monei.rb
@@ -81,7 +81,7 @@ module ActiveMerchant #:nodoc:
       #
       # Returns Active Merchant response object
       def capture(money, authorization, options={})
-        execute_authorization(:capture, money, authorization, options)
+        execute_dependant(:capture, money, authorization, options)
       end
 
       # Public: Refunds from previous purchase
@@ -97,7 +97,7 @@ module ActiveMerchant #:nodoc:
       #
       # Returns Active Merchant response object
       def refund(money, authorization, options={})
-        execute_authorization(:refund, money, authorization, options)
+        execute_dependant(:refund, money, authorization, options)
       end
 
       # Public: Voids previous authorization
@@ -108,7 +108,7 @@ module ActiveMerchant #:nodoc:
       #
       # Returns Active Merchant response object
       def void(authorization, options={})
-        execute_authorization(:void, nil, authorization, options)
+        execute_dependant(:void, nil, authorization, options)
       end
 
       # Public: Verifies credit card. Does this by doing a authorization of 1.00 Euro and then voiding it.
@@ -145,7 +145,7 @@ module ActiveMerchant #:nodoc:
       end
 
       # Private: Execute operation that depends on athorization code from previous purchase or authorize operation
-      def execute_authorization(action, money, authorization, options)
+      def execute_dependant(action, money, authorization, options)
         request = build_request do |xml|
           add_identification_authorization(xml, authorization, options)
           add_payment(xml, action, money, options)
--- a/lib/active_merchant/billing/gateways/monei.rb
+++ b/lib/active_merchant/billing/gateways/monei.rb
@@ -43,8 +43,6 @@ module ActiveMerchant #:nodoc:
       #               :billing_address  Hash with billing address information
       #               :description      Merchant created purchase description (optional)
       #               :currency         Sale currency to override money object or default (optional)
-      #               :first_name       Customer first name to override card holder's (optional)
-      #               :last_name        Customer family name to override card holder's (optional)
       #
       # Returns Active Merchant response object
       def purchase(money, credit_card, options)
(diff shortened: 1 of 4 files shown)


leolara:
@ntalbott hey, could you, please, have a look at this gateway?

ntalbott:
@leolara to set expectations: I'm slammed, and the maintainers will get to it when we can. Could be a few weeks at this point.

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      # Constructor
+      #
+      # options - Hash containing the gateway credentials, ALL MANDATORY
+      #           :sender_id  Sender ID
+      #           :channel_id Channel ID
+      #           :login      User login
+      #           :pwd        User password
+      #
+      def initialize(options={})
+        requires!(options, :sender_id, :channel_id, :login, :pwd)
```
andrewpaliga: Are `:sender_id` and `:channel_id` supplied by the merchant?

leolara: All four parameters are supplied from Monei to the merchant, who have to user them to configure the gateway. All four parameters are required and different for every merchant.

andrewpaliga:
@j-mutter I :tophat:'s this locally -- looks good. Can take a quick look at the code?

leolara:
Hi @j-mutter 

I do not understand the expression, "I :tophat:'d this locally".

Do you want me to look at the code?

Regards,
Leo

andrewpaliga:
@leolara that just means tested and the comment was directed at @j-mutter he will take a look when he gets a chance

leolara:
@andrewpaliga Awesome!! Sorry for the confusion and thank you for the attention.

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      # Public: Performs purchase operation
+      #
+      # money       - Amount of purchase
+      # credit_card - Credit card
+      # options     - Hash containing purchase options
+      #               :order_id         Merchant created id for the purchase
+      #               :billing_address  Hash with billing address information
+      #               :description      Merchant created purchase description (optional)
+      #               :currency         Sale currency to override money object or default (optional)
+      #               :given            Customer given name to override card holder's (optional)
```
j-mutter: Please use the ActiveMerchant standard [`:name`](https://github.com/Shopify/active_merchant/wiki/contributing#address-hash) key in an address hash, or `:first_name`, `:last_name` from a [`CreditCard`](https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/credit_card.rb#L107-L115) to get the customer's name

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      # options       - Hash containing capture options
+      #                 :order_id         Merchant created id for the authorization (optional)
+      #                 :description      Merchant created authorization description (optional)
+      #                 :currency         Sale currency to override money object or default (optional)
+      #
+      # Note: you should pass either order_id or description
+      #
+      # Returns Active Merchant response object
+      def capture(money, authorization, options={})
+        execute_authorization(:capture, money, authorization, options)
```
j-mutter: These method names are a bit weird, `#execute_authorization` sounds like you're doing the initial auth, not acting on one.

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+
+          xml.Name do
+            xml.Given options[:given] || credit_card.first_name
+            xml.Family options[:family] || credit_card.last_name
+          end
+          xml.Address do
+            xml.Street address[:address1].to_s
+            xml.Zip address[:zip].to_s
+            xml.City address[:city].to_s
+            xml.State address[:state] if address.has_key? :state
```
j-mutter: Is there a reason you don't `to_s` the state, but you do for everything else?

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+        response[:unique_id]
+      end
+
+      # Private: Encode POST parameters
+      def post_data(xml)
+        "load=#{CGI.escape(xml)}"
+      end
+
+      # Private: Translate Monei status code to native ruby symbols
+      def tr_status_code(code)
```
j-mutter: Can you please map the error codes here to their respective [`STANDARD_ERROR_CODES`](https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L80-L93), and include them in the response from your `#commit` method above? A good example of this is in the Stripe gateway. [Here is the response](https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L395) and [here is the mapping](https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L35-L46)

lib/active_merchant/billing/gateways/monei.rb
leolara: @j-mutter the first_name and last_name are taken from the `CreditCard` already. This is a way to override that. Should I delete this way to override, or change the options to `:first_name` and `:last_name`?

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+          message_from(response),
+          response,
+          authorization: authorization_from(response),
+          test: test?
+        )
+      end
+
+      # Private: Decide success from servers response
+      def success_from(response)
+        response[:status] == :success or response[:status] == :new
```
j-mutter: Please prefer `||` over `or` for logic checks.

test/remote/gateways/remote_monei_test.rb
```diff
@@ -0,0 +1,117 @@
...
+    response = @gateway.purchase(@amount, @declined_card, @options)
+    assert_failure response
+    assert_equal 'invalid cc number/brand combination', response.message
+  end
+
+  def test_successful_authorize_and_capture
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount, auth.authorization)
```
j-mutter: The newlines/lack of are all over the place in these tests, but I can clean that up in the merge

test/remote/gateways/remote_monei_test.rb
```diff
@@ -0,0 +1,117 @@
...
+    response = @gateway.authorize(@amount, @declined_card, @options)
+    assert_failure response
+  end
+
+  def test_partial_capture
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount-1, auth.authorization)
+    assert_success capture
```
j-mutter: Is multi capture supported? Could you add a line here to capture the outstanding balance?

test/remote/gateways/remote_monei_test.rb
```diff
@@ -0,0 +1,117 @@
...
+    assert refund = @gateway.refund(@amount, purchase.authorization)
+    assert_success refund
+  end
+
+  def test_partial_refund
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success purchase
+
+    assert refund = @gateway.refund(@amount-1, purchase.authorization)
+    assert_success refund
```
j-mutter: Same question as above re: multi-refund

lib/active_merchant/billing/gateways/monei.rb
leolara: @j-mutter Done

lib/active_merchant/billing/gateways/monei.rb
leolara: @j-mutter Done

lib/active_merchant/billing/gateways/monei.rb
leolara: @j-mutter Done

lib/active_merchant/billing/gateways/monei.rb
leolara: @j-mutter I could translate all the rejections into `:card_declined` but the other values are more status that error codes.

j-mutter:
No major gripes with anything, I'll circle back to this in a bit to look at the changes. Remote tests passed for me locally.

@ntalbott @girasquid would appreciate some more :eyes: if you guys have some spare cycles

test/remote/gateways/remote_monei_test.rb
leolara: @j-mutter Do you think I should add more new lines inside the test methods?

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      end
+
+      # Private: Build XML wrapping code yielding to code to fill the transaction information
+      def build_request
+        builder = Nokogiri::XML::Builder.new(:encoding => 'UTF-8') do |xml|
+          xml.Request(:version => "1.0") do
+
+            xml.Header { xml.Security(:sender => @options[:sender_id]) }
+
+            xml.Transaction(:mode => test? ? 'CONNECTOR_TEST' : 'LIVE', :response => 'SYNC', :channel => @options[:channel_id]) do
```
girasquid: If I'm following this correctly - when you're in test mode, you need something different in the XML you're sending as well as hitting a different URL?

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      # Private: Add identification part to XML for new orders
+      def add_identification_new_order(xml, options)
+        requires!(options, :order_id)
+        xml.Identification do
+          xml.TransactionID options[:order_id]
+        end
+      end
+
+      # Private: Add identification part to XML for orders that depend on authorization from previous operation
+      def add_identification_authorization(xml, authorization, options)
```
girasquid: It seems like this and `add_identification_new_order` above could be wrapped up into one function, that took an optional `ReferenceID`:
```ruby
def add_order_id(xml, options)
  requires!(options, :order_id)
  xml.Identification do
    xml.ReferenceID options[:authorization] if options[:authorization]
    xml.TransactionID options[:order_id]
  end
end
```

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      # Private: Add payment part to XML
+      def add_payment(xml, action, money, options)
+        code = tr_payment_code action
+
+        xml.Payment(:code => code) do
+          xml.Presentation do
+            xml.Amount amount(money)
+            xml.Currency options[:currency] || currency(money)
+            xml.Usage options[:description] || options[:order_id]
+          end unless money.nil?
```
girasquid: When will you be doing a transaction for a `nil` amount?

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+            '64' => :waiting_risk,
+            '65' => :rejected_risk,
+            '70' => :rejected_validation,
+            '80' => :waiting,
+            '90' => :new
+        }[code]
+      end
+
+      # Private: Translate AM operations to Monei operations codes
+      def tr_payment_code(action)
```
girasquid: The `tr` abbreviation isn't necessary. Please name this `translate`, if that's what it is doing.

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      def add_identification_authorization(xml, authorization, options)
+        xml.Identification do
+          xml.ReferenceID authorization
+          xml.TransactionID options[:order_id]
+        end
+      end
+
+      # Private: Add payment part to XML
+      def add_payment(xml, action, money, options)
+        code = tr_payment_code action
```
girasquid: Can you add brackets to this, to make it clear what's function-call and what's arguments?

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+        end
+      end
+
+      # Private: Parse XML response from Monei servers
+      def parse(body)
+        xml = Nokogiri::XML(body)
+
+        {
+            :unique_id => xml.xpath("//Response/Transaction/Identification/UniqueID").text,
+            :status => (tr_status_code xml.xpath("//Response/Transaction/Processing/Status/@code").text),
```
girasquid: Brackets around arguments please, here and line 247

lib/active_merchant/billing/gateways/monei.rb
leolara: of course, will do

lib/active_merchant/billing/gateways/monei.rb
leolara: @girasquid In `void` operation. You can see it calls `execute_dependant` with `amount=nil`which gets passed along to this method `add_payment`.

lib/active_merchant/billing/gateways/monei.rb
leolara: @girasquid If an user of the library would call `purchase` with an `options[:authorization]` value that would cause a problem. I can assume the user will not do that, or check in the new_order operations whether `options[:authorization]` is empty. What would you prefer?

lib/active_merchant/billing/gateways/monei.rb
leolara: @girasquid The system is very flexible. There are two test modes plus live, we are only using the test mode that test the connector. Additionally, it is possible to do test requests to the live system, the only difference is that in the live system there is a transaction fee for test requests and in the test server they are for free. We do not want to do the test request to the live server, because if someone runs the remote test many times will cost money.

lib/active_merchant/billing/gateways/monei.rb
leolara: Done

lib/active_merchant/billing/gateways/monei.rb
leolara: Done

lib/active_merchant/billing/gateways/monei.rb
leolara: Done

lib/active_merchant/billing/gateways/monei.rb
leolara: Please, @j-mutter let me know the best way to solve this

lib/active_merchant/billing/gateways/monei.rb
leolara: @j-mutter Done

test/remote/gateways/remote_monei_test.rb
leolara: @j-mutter It is not implemented by the gateway. I event added a test that tests the failure of multi-capture and multi-refund

leolara:
@j-mutter @andrewpaliga @girasquid 

I think everything is done of what has been requested up to this point, with the exception of the option for first_name and last_name where I asked for feedback. What is the order of overriding first_name and last_name, like is first credit_card.first_name and if nil then options.address.first_name or inverse, or what?

Cheers,
Leo

lib/active_merchant/billing/gateways/monei.rb
andrewpaliga: @leolara I would just swap `:given` and `:family` for `:first_name` and `:last_name` here

ntalbott: I dunno, this extra override seems unnecessary. I'd vote for dropping it altogether.

leolara: I did as @andrewpaliga said, tested and was pushing when I saw @ntalbott message.

What do you prefer guys?

cc/ @ntalbott @j-mutter

andrewpaliga: Drop it :+1:

leolara: ok, I understand that I should remove the options to override first_name and last_name, both will be taken only from the `credit_card` object.

leolara:
@j-mutter @andrewpaliga @girasquid I think everything you have said is done. Please, have a look.

leolara:
I forgot to mention @ntalbott

leolara:
@j-mutter @andrewpaliga @girasquid @ntalbott So, it seems it is everything, isn't?

j-mutter:
I'll make another pass at this this afternoon and merge it if it looks good :+1:

lib/active_merchant/billing/gateways/monei.rb
```diff
@@ -0,0 +1,314 @@
...
+      end
+
+      # Private: Get message from servers response
+      def message_from(response)
+        response[:message]
+      end
+
+      # Private: Get error code from servers response
+      def error_code_from(response)
+        success_from(response) ? nil : STANDARD_ERROR_CODE[:card_declined]
```
j-mutter: I just noticed this, does Monei not provider actual error codes we could map to?

j-mutter: The `translate_` methods below seem to imply yes, so why do all failures simply get the `card_declined` error code?

j-mutter:
One last question about error codes, otherwise LGTM

lib/active_merchant/billing/gateways/monei.rb
leolara: Hi @j-mutter,

The only codes that we get from the API are:

NEW 90
WAITING 80
REJECTED_VALIDATION 70
REJECTED_RISK ￼65
￼REJECTED_BANK 60
￼NEUTRAL ￼40
￼SUCCESS 00

I translated all the rejections to card_declined because I do not get the level of detail in the response that STANDARD_ERROR_CODE has. What do you suggest?

j-mutter: Ah ok, if that's all we get then I guess we can't do much about it.

j-mutter:
Merged, thanks!

leolara:
@j-mutter @ntalbott @andrewpaliga @girasquid Thank you all!!

cc/ @alexandresaiz

PR #1557, opened 2015-02-08
Add new gateway   rficb

romanof76:


girasquid:
@romanof76 ActiveMerchant doesn't support offsite gateways. All the offsite gateway integrations are in https://github.com/Shopify/offsite_payments, so you'll want to submit this there.

PR #1560, opened 2015-02-09
Authorize.Net: properly send shipping address when provided

girasquid:
The previous version of the code would treat the payment_source as the best source for names - which meant that your credit card's billing name would override your shipping address name.

@aprofeit @ivanfer

### DIFF
--- a/test/unit/gateways/authorize_net_test.rb
+++ b/test/unit/gateways/authorize_net_test.rb
@@ -513,6 +513,54 @@ class AuthorizeNetTest < Test::Unit::TestCase
     end.respond_with(successful_authorize_response)
   end
 
+  def test_includes_shipping_name_when_different_from_billing_name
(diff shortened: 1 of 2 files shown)


test/unit/gateways/authorize_net_test.rb
```diff
@@ -523,6 +546,7 @@ def test_truncation
       order_id: "a" * 21,
       description: "a" * 256,
       billing_address: address(
+        name: ["a" * 51, "a" * 51].join(" "),
```
girasquid: I had to make this change in order to give the addresses being used in these tests actual values; by default they're set to "Jim Smith" but were falling through to the billing address of the credit card.

aprofeit:
:+1:

ivanfer:
:palm_tree:

PR #1561, opened 2015-02-10
Fix paymill refunds

andrewpaliga:
Paymill returns a string `response_code` for refunds in production environment which causes refunds to be categorized as unsuccessful. This PR fixes this and ensures a string "20000" code still returns successful.

@girasquid @aprofeit

### DIFF
--- a/lib/active_merchant/billing/gateways/paymill.rb
+++ b/lib/active_merchant/billing/gateways/paymill.rb
@@ -85,7 +85,7 @@ module ActiveMerchant #:nodoc:
           :test => (parsed['mode'] == 'test'),
         }
 
-        succeeded = (parsed['data'] == []) || (parsed['data']['response_code'] == 20000)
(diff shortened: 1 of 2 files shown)


aprofeit:
:+1:

lib/active_merchant/billing/gateways/paymill.rb
```diff
@@ -85,7 +85,7 @@ def response_from(raw_response)
           :test => (parsed['mode'] == 'test'),
         }
 
-        succeeded = (parsed['data'] == []) || (parsed['data']['response_code'] == 20000)
+        succeeded = (parsed['data'] == []) || (parsed['data']['response_code'].to_i == 20000)
```
aprofeit: When status code `200` is not enough, add 2 zeroes.

girasquid:
:+1:

PR #1564, opened 2015-02-11
IPP storage/tokenization

jscheid:
This adds support for storage (tokenization) to the IPP gateway implementation. It also fixes a warning in the IPP remote test and sets the IPP default currency to AUD.

### DIFF
--- a/lib/active_merchant/billing/gateways/ipp.rb
+++ b/lib/active_merchant/billing/gateways/ipp.rb
@@ -34,7 +34,12 @@ module ActiveMerchant #:nodoc:
             xml.CustRef options[:order_id]
             add_amount(xml, money)
             xml.TrnType "1"
-            add_credit_card(xml, payment)
+            if payment.respond_to?(:number)
+              add_credit_card(xml, payment)
+            else
+              # triggered payment for an tokenised credit card
+              add_token_details(xml, payment)
+            end
             add_credentials(xml)
             xml.TrnSource options[:ip]
           end
@@ -47,7 +52,12 @@ module ActiveMerchant #:nodoc:
             xml.CustRef options[:order_id]
             add_amount(xml, money)
             xml.TrnType "2"
-            add_credit_card(xml, payment)
+            if payment.respond_to?(:number)
+              add_credit_card(xml, payment)
+            else
+              # triggered payment for an tokenised credit card
+              add_token_details(xml, payment)
+            end
             add_credentials(xml)
             xml.TrnSource options[:ip]
           end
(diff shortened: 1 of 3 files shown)


j-mutter:
LGTM. Remote tests are green except the transcript dump that's a skip. 

@girasquid @ntalbott any objections?

ntalbott:
Why is `store` not using the existing `commit` method? I'm not keen on adding another place that's doing request/response handling.

j-mutter:
Derp, totally missed that. #mondays. :coffee: :coffee:

jscheid:
Hi, thanks for taking a look at this!

In order to reuse the `commit` method here I'd have to make the value for `SOAPAction` (more) configurable,  the call to `new_submit_xml` more configurable (i.e. use `new_submit_xml_with_type` instead), allow passing in a custom `message_from` method and allow passing in a custom `authorization_from` method.

I'm happy to do all that, I just figured the two API invocations and their request/response formats are sufficiently different from each other that it's cleaner to just introduce a new method.

jscheid:
When I say "introduce a new method" I mean "borrow the relevant code from the `commit` method for the `store` method" :-)

Again, happy to restructure this if it's not DRY enough for you guys.

ntalbott:
@jscheid yah, we don't like to go crazy with DRY, but I think it's appropriate here to parameterize things and share the code. I'd also recommend changing up the urls such that there's no need for the extra constants; just back up the class level urls a level, and have the additional path added as appropriate.

jscheid:
Hi guys, is this more like it?

bruno:
:+1: on getting this merged if it's ok ...

aspett:
`SingleSubmitVOID` causes a 500 error on demo IP Payment servers. Changing this to `SingleSubmitVoid` fixes the issue.

ntalbott:
@jscheid sorry I'm getting back to this so late! Can you verify and add tests that verify that the `Response#authorization` returns a value that can be passed directly to `purchase` and `authorize` as a payment method? That's an implicit assumption I've just started trying to document more clearly here: https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards#store. Thanks!

jscheid:
No problem. @demillir, can I leave this with you?

bruno:
@ntalbott I've just made some changes to add support for `purchase` and `authorize` using the `tokens`. It was a good catch before prior to this, you could not have used the `token` you get back from the `store()` method in those two operations as they both expected a credit card.

jasonwebster:

bruno:
I don't seem to be able to re-open this PR ... there were no _outstanding_ issues per-se, just applied the changes as requested and was just waiting for a further review from the latest changes from the core/committers ...

PR #1566, opened 2015-02-13
Handle network tokenization (Apple Pay) options for CyberSource

jnormore:
Updates cybersource gateway to handle network tokenization options (apple pay).

Docs: http://apps.cybersource.com/library/documentation/dev_guides/tokenization_SCMP_API/Tokenization_SCMP_API.pdf

CyberSource API version has to be bumped in order to use this.

More gateways coming after this, this one is one of the only gateways so far that actually has sandbox environment support for this so remote tests work.

@ntalbott @girasquid what are your thoughts on the approach? We need to lock down the interface to this before I go to far with other gateways.


girasquid:
Network tokenization information looks like it always comes along with a plain-old `CreditCard` - does it make sense to create an augmented `CreditCard` class that can handle storing the original properties and also the network tokenization ones? By having the object live somewhere that wasn't in a particular gateway's integration code you'd also be able to lock it down better, I think.

rwdaigle:
Hi @jnormore. I've been doing a fair but of work at Spreedly looking into Apple Pay and network tokenization. It's great to see this becoming a first class citizen in AM. Few comments, if I may?

> More gateways coming after this

I would love to know what other gateways support a tokenization API. Do you have a canonical list anywhere?

It seems that the current landscape is quite varied, with some gateways supporting the ability to pass in encrypted (Apple Pay specific) payment tokens and others (like CS here) letting you pass in the decrypted components. It would be good to understand the range of options possible to know what's the right level of abstraction.

Without seeing many other examples, I tend to agree with the direction @girasquid advocates for here:

> Network tokenization information looks like it always comes along with a plain-old CreditCard - does it make sense to create an augmented CreditCard class that can handle storing the original properties and also the network tokenization ones?

My sense is there are enough different "levels" of network tokenization support that it will become necessary to have a separate entity from `CreditCard` to represent tokenized payment data (though it will share a lot of functionality with a card). While most underlying tokenization schemes adhere to a PAN and exp date format, making it look and feel very much like a traditional credit card, there seem to already be enough departures from this that forcing it into a card entity might not work.

However, I'm probably getting ahead of myself. While I think we might get here eventually, at the very least I do think we should start pulling tokenization info out of the `options` hash and into properties of a card, as much as they can be shoe-horned in for now as @girasquid suggests.

Otherwise, @jnormore, the implementation looks to accurately adhere to the CS spec :+1:

jnormore:
> I would love to know what other gateways support a tokenization API. Do you have a canonical list anywhere?

We've been working with some on this, I'll add to the list as they make their APIs public for it. If you want to follow up over email, Nathaniel and I had a thread going a while back related to apple pay that you might find useful.

> It would be good to understand the range of options possible to know what's the right level of abstraction.

Essentially, all network tokenization implementations will have to adhere to the spec anyway so they will be similar in most respects, the difference will come in how much extra processing the gateway or processor will do on their end (especially with respect to handling the different card types).

Thanks for the feedback, I agree with what you both are saying about the need for the new object type. I'll make those updates and ping back when ready :)

rwdaigle:
@jnormore 

> Essentially, all network tokenization implementations will have to adhere to the spec

Well, there's the network tokenization spec that deals mostly with the internals of the payment network and then there's Apple Pay, which adds a bunch of stuff *above* the gateway (it's brand of encryption, the payment token format etc...)

I guess I'm wondering where the distinction in AM lies between Apple Pay specific support and network tokenization support.

For instance, Auth.net, Litle and others have an API that lets you pass in an encrypted Apple Pay payment token. Wrapped inside is a card token and cryptogram conforming to the network tokenization spec, but it's obscured. And here Cybersource has an API that lets you pass in the card token and cryptogram directly.

It seems to me that these will need to have different representations in AM even though they are both, under the covers, powered by network tokenization.

Hopefully I'm not talking past you, just want to understand what the approach is for these different gateway API scenarios!

jnormore:
@rwdaigle I see what you mean. We already have support for the PKPaymentToken passthrough directly in AM, https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/apple_pay_payment_token.rb

What I'm trying to do here is support the decrypted token, still a token but on the payment network side and looks like a normal PAN and requires the extra verification cryptogram passed through some existing fields (3DS usually).

rwdaigle:
> We already have support for the PKPaymentToken

I see that now - apologies for not doing some more digging before jumping in here.

> What I'm trying to do here is support the decrypted token

Understood.

jnormore:
So I added the ``NetworkTokenizationCreditCard`` here, it does simplify a few things, good call.

Can I get a review when you're ready?

cc @bizla for review too please, thanks!

lib/active_merchant/billing/network_tokenization_credit_card.rb
```diff
@@ -0,0 +1,16 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class NetworkTokenizationCreditCard < CreditCard
```
bizla: We have the `PaymentToken` class that intends to abstractly represent tokenized data. If we have a better understanding on how to model this kind of tokenized data, then I believe we should extend `PaymentToken` rather than introduce a new `CreditCard` subclass.

lib/active_merchant/billing/network_tokenization_credit_card.rb
```diff
@@ -0,0 +1,16 @@
...
+      # A +NetworkTokenizationCreditCard+ object represents a tokenized credit card 
+      # using the EMV Network Tokenization specification, http://www.emvco.com/specifications.aspx?id=263.
+      #
+      # It includes all fields of the +CreditCard+ class with additional fields for 
+      # verification data that must be given to gateways through existing fields (3DS / EMV).
+      #
+      # The only tested usage of this at the moment is with an Apple Pay decrupted PKPaymentToken,
```
bizla: Typo: "decrupted"

lib/active_merchant/billing/network_tokenization_credit_card.rb
jnormore: It's essentially a credit card though, it shares all the attributes of a credit card, it just has the extra verification fields

lib/active_merchant/billing/network_tokenization_credit_card.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Billing #:nodoc:
+    class NetworkTokenizationCreditCard < CreditCard
+      # A +NetworkTokenizationCreditCard+ object represents a tokenized credit card 
+      # using the EMV Network Tokenization specification, http://www.emvco.com/specifications.aspx?id=263.
+      #
+      # It includes all fields of the +CreditCard+ class with additional fields for 
+      # verification data that must be given to gateways through existing fields (3DS / EMV).
+      #
+      # The only tested usage of this at the moment is with an Apple Pay decrupted PKPaymentToken, 
+      # https://developer.apple.com/library/prerelease/ios/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.html
```
bizla: Minor: Probably want to link to https://developer.apple.com/library/ios/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.html (i.e. the non-prerelease docs)

jnormore: google fail, thanks ;)

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -468,8 +472,48 @@ def add_tax_service(xml)
...
+          xml.tag! 'ccAuthService', {'run' => 'true'}
+        end
+      end
+
+      def network_tokenization?(payment_method)
+        payment_method.is_a?(NetworkTokenizationCreditCard)
+      end
+
+      def add_network_tokenization(xml, payment_method, options)
+        return unless network_tokenization?(payment_method)
```
girasquid: Do you need this check? It seems like you're only calling `add_network_tokenization` in situations where you already know that the `payment_method` is a `NetworkTokenizationCreditCard`.

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -286,4 +286,19 @@ def test_successful_validate_pinless_debit_card
     assert_equal 'Y', response.params["status"]
     assert_equal true,  response.success?
   end
+
+  def test_network_tokenization_authorize_and_capture
+    credit_card = network_tokenization_credit_card('4111111111111111',
+      :brand              => 'visa',
+      :eci                => "05",
+      :payment_cryptogram => "EHuWW9PiBkWvqE5juRwDzAUFBAk="
```
girasquid: Any docs on how to generate this for someone writing a new test?

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -286,4 +286,19 @@ def test_successful_validate_pinless_debit_card
     assert_equal 'Y', response.params["status"]
     assert_equal true,  response.success?
   end
+
+  def test_network_tokenization_authorize_and_capture
```
girasquid: Is there ever a network tokenized purchase, or is it always auth+capture? Do refund semantics change?

girasquid:
Couple questions, looks ok though.

lib/active_merchant/billing/gateways/cyber_source.rb
bizla: I like the fool-proofing here actually. :)

test/remote/gateways/remote_cyber_source_test.rb
jnormore: this is from the cybersource docs. there's no way to generate this without initiating an actual apple pay payment

test/remote/gateways/remote_cyber_source_test.rb
jnormore: always auth+capture

jnormore: refunds are no different

lib/active_merchant/billing/network_tokenization_credit_card.rb
bizla: Right, I'm seeing that now. :cool: (I'm cheating, we totally discussed this offline. ;))

lib/active_merchant/billing/network_tokenization_credit_card.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Billing #:nodoc:
+    class NetworkTokenizationCreditCard < CreditCard
+      # A +NetworkTokenizationCreditCard+ object represents a tokenized credit card 
+      # using the EMV Network Tokenization specification, http://www.emvco.com/specifications.aspx?id=263.
+      #
+      # It includes all fields of the +CreditCard+ class with additional fields for 
+      # verification data that must be given to gateways through existing fields (3DS / EMV).
+      #
+      # The only tested usage of this at the moment is with an Apple Pay decrypted PKPaymentToken, 
+      # https://developer.apple.com/library/ios/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.html
```
bizla: Can you add some clarification commentary here and in `PaymentToken`? 
That is -- if I'm correct -- some language to describe that: 
- `NetworkTokenizationCreditCard` is exactly like a credit card but with EMV/3DS standard payment network tokenization data
- `PaymentToken` is meant for modeling proprietary payment token structures (like `stripe` and `apple_pay`)

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -27,11 +27,13 @@ module Billing #:nodoc:
     # * To process pinless debit cards through the pinless debit card
     #   network, your Cybersource merchant account must accept pinless
     #   debit card payments.
+    # * The order of the XML elements does matter, make sure to follow the order in 
+    #   the documentation exactly.
```
bizla: :disappointed:

bizla:
Nice to see that this is coming one big learning exercise. :smiley: I feel like once we have a better overall grip on the tokenization landscape, we'll be able to iron out our data modelling a bit better.

:ship: once the comments reflect the intent of `PaymentToken` vs `NetworkTokenizationCreditCard`

test/remote/gateways/remote_cyber_source_test.rb
rwdaigle: What is the experience like when somebody passed in a `NetworkTokenizationCreditCard` for a purchase? Looks like the tokenization parts just get ignored and its processed as though it were a credit card which, I believe, will result in a decline.

Now that we have a growing body of payment types, should we have a standard exception thrown to denote when a certain payment type is not valid for the given operation?

jnormore: This is a little unclear, I have a feeling that some gateways may do and auth+capture on their end in this case but in general yeah it will just be declined.

I'm on the fence about whether we should handle this or let the gateway handle it cause most integrations that use AM would handle the decline properly but if we started to introduce another exception type it might cause some problems (not really an excuse I guess either)

lib/active_merchant/billing/network_tokenization_credit_card.rb
jnormore:
```
It includes all fields of the +CreditCard+ class with additional fields for 
verification data that must be given to gateways through existing fields (3DS / EMV).
```

Good for the first point?

I'll add something to ``PaymentToken`` :)

jnormore: I actually think what you already have in ``PaymentToken`` makes a lot of sense already:

```
   # Base class representation of cryptographic payment data tokens that may be used for EMV-style transactions
    # like Apple Pay. Payment data may be transmitted via any data type, and may also be padded
    # with metadata specific to the cryptographer. This metadata should be parsed and interpreted in concrete
    # implementations of your given cryptographer. Like credit cards, you must also return a string representing
    # the token's type, like 'apple_pay' or 'stripe' should your target payment gateway process these tokens.
```

what do you think?

test/remote/gateways/remote_cyber_source_test.rb
rwdaigle: > whether we should handle this or let the gateway handle it

Mmmm, yes, I do suppose the general approach of AM is to not introduce too much between the caller and the gateway. Ok!

jnormore:
I think I addressed all issues, I'm going to merge this unless anyone has any objections?

lib/active_merchant/billing/network_tokenization_credit_card.rb
bizla: Yeah, now that the context is a little clearer, the comments seem good enough.

bizla:
:+1: This is cool.

girasquid:
:+1:

PR #1568, opened 2015-02-16
Add apple pay support for braintree blue

jnormore:
Continuation of https://github.com/Shopify/active_merchant/pull/1566

This one is real simple, they've adding most of the extra logic to their gem so it's just a matter of passing the new structure through.

I updated ``Gemfile`` for now to include their working branch, this is just to get the PR open and some :eyes: on it. I won't merge until they have that in master and a new gem version released.

NOTE: You can use any sandbox credentials for this but the guys at braintree have to enable the ability to use apple pay for your account before you can test it (remote tests work for me right now). They are adding the ability to programmatically enable an account for apple pay though, coming soon (this week).

cc @girasquid @bizla @ntalbott @rwdaigle @markabe


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -537,12 +537,22 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
             :first_name => credit_card_or_vault_id.first_name,
             :last_name => credit_card_or_vault_id.last_name
           )
-          parameters[:credit_card] = {
-            :number => credit_card_or_vault_id.number,
-            :cvv => credit_card_or_vault_id.verification_value,
-            :expiration_month => credit_card_or_vault_id.month.to_s.rjust(2, "0"),
-            :expiration_year => credit_card_or_vault_id.year.to_s
-          }
+          if credit_card_or_vault_id.is_a?(NetworkTokenizationCreditCard)
```
rwdaigle: @jnormore I've seen something like this `is_a?` check in all the adapters so far. Wondering if it shouldn't be the responsibility of the credit card/payment method to know what its type is? Something like:

```ruby
payment_method.tokenized?
```

Also, I feel like we need to decide if we're going to be generic and call this type of support "network tokenization" support or be specific and call it "Apple Pay" support. Some gateways (Cybersource) seem to implement true generic network tokenization support whereas others, like Braintree here, are specific to Apple Pay as evidence by the `apple_pay_card` parameter.

You've seen more of these than I, how are gateways implementing this feature – In true generic fashion, or specific to AP?

jnormore: Ideally we could do something like ``payment_method.tokenized?`` but this breaks when you have a string or a hash at the payment method (for example stripe token). I think we're stuck with the ``is_a?`` for the time being, unless you have other suggestions?

Braintree is the only one I know of that specifically calls it apple pay, the rest I've seen are generic. I think even the braintree solution would work for other non-apple solutions if we had one to try. I'm sure one will come along soon anyway ;)

rwdaigle: > but this breaks when you have a string or a hash at the payment method (for example stripe token)

I don't quite follow here, but I'm the new kid on the block so that's to be expected.



> Braintree is the only one I know of that specifically calls it apple pay

Cool. I'm fine leaving it as the generic "network tokenization" until a real-world scenario forces us to do otherwise.

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -537,12 +537,22 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
...
+              :cryptogram => credit_card_or_vault_id.payment_cryptogram
+            }
+          else
+            parameters[:credit_card] = {
+              :number => credit_card_or_vault_id.number,
+              :cvv => credit_card_or_vault_id.verification_value,
+              :expiration_month => credit_card_or_vault_id.month.to_s.rjust(2, "0"),
+              :expiration_year => credit_card_or_vault_id.year.to_s
+            }
+          end
```
bizla: :cry: this is a thing of beauty

bizla:
:+1:

girasquid:
:+1:

Gemfile
```diff
@@ -6,4 +6,5 @@ gem 'jruby-openssl', :platforms => :jruby
 group :test, :remote_test do
   # gateway-specific dependencies, keeping these gems out of the gemspec
   gem 'braintree', '>= 2.0.0'
+  # gem 'braintree', github: "braintree/braintree_ruby", branch: "raw_apple_pay_preview" # for apple pay support in braintree
```
jnormore: I added this gem line as comment here instead since it's not a production release. How do you guys feel about this?

girasquid: Are your tests going to pass if you only installed the things in the test group?

jnormore: not the apple pay tests for braintree

girasquid: If the braintree gem is included so that the braintree tests pass - it seems like if that's what you need for those tests to pass, that should be what's in the gemfile (and then this can get updated to go back to the regular gem once it's been released).

jnormore: actually (as we discussed offline), going to wait to merge this until we have the ability to enable apple pay for a merchant ID through their API as well.

girasquid: :+1:

j-mutter:
@jnormore I think you need a rebase :\

rwdaigle:
@jnormore also, here's another PR that's going to the `network-tokenization` branch instead of `master`, if that's not intentional.

j-mutter:
Closing in favour of https://github.com/Shopify/active_merchant/pull/1629

PR #1572, opened 2015-02-17
Flo2Cash: Map Reference->:order_id (not :invoice)

ntalbott:
:invoice is actually rarely used, and :order_id is preferred when
possible.

This will break anyone who is using :invoice with Flo2Cash since
:invoice wouldn't get mapped anymore, but since the adapter is new
seems worth the possible breakage.

@duff @markabe please review.


girasquid:
:+1:

duff:
yacht_luxury_-_google_search

markabe:
Like a band-aid.  :boom: :+1:

PR #1575, opened 2015-02-18
Flo2Cash: Fix card brand handling

ntalbott:
Also improves error handling so that it returns responses instead of
raising exceptions.

@duff @markabe please review.


markabe:
I see, lot's of cleanup too.  Cool. :ship:

duff:
That does whack a good bit of duplication, eh?  :clap:

PR #1578, opened 2015-02-23
Remove the vindicia gateway

mutemule:
Vindicia is configured with a plaintext endpoint, which is bad.  However, this plaintext endpoint does not respond to TCP SYN packets, so I guess that's not so bad.  Well, except that it means that Vindicia is broken and does not work.

Given that Vindicia is broken, just remove it.

@ntalbott @girasquid @Shopify/payments


ntalbott:
:+1:

girasquid:
:+1:

j-mutter:
:fire:

newton10471:
I presume the plaintext endpoint we're talking about is this? 

self.live_url = "http://soap.vindicia.com/soap.pl"

Which yes, is bad, and doesn't work, but it works with https.  Any chance we could add this gateway back using https instead?  Without going into detail, people actually do use this particular gateway :-D.

ntalbott:
@newton10471 if someone wanted to submit a PR adding it back that has fully passing remote tests, and can provide us maintainers with credentials to run those tests, we'd be willing to add it back. Beware, though, that the recurring functionality is deprecated in AM, and we won't bring it back in.

newton10471:
@ntalbott ok sounds good thanks

PR #1579, opened 2015-02-25
Disallow production gateway to alter endpoints for Checkout

markabe:
@ntalbott Just noticed this, and what looks like a security hole to me. Please ignore if I'm off base.

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout.rb
+++ b/lib/active_merchant/billing/gateways/checkout.rb
@@ -26,8 +26,6 @@ module ActiveMerchant #:nodoc:
       }
 
       def initialize(options = {})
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/gateways/checkout.rb
```diff
@@ -26,7 +26,7 @@ class CheckoutGateway < Gateway
       }
 
       def initialize(options = {})
-        @url = (options[:gateway_url] || self.live_url)
+        @url = self.live_url
```
ntalbott: Can we just drop the `@url` ivar and use `self.live_url` where appropriate?

markabe:
@ntalbott True. Updated.

PR #1582, opened 2015-02-25
WorldPay: Remove country overlap with WorldPay US

markabe:
@ntalbott As we discussed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ The [ActiveMerchant Wiki](http://github.com/Shopify/active_merchant/wikis) conta
 * [WebPay](https://webpay.jp/) - JP
 * [WePay](https://www.wepay.com/) - US
 * [Wirecard](http://www.wirecard.com) - AD, CY, GI, IM, MT, RO, CH, AT, DK, GR, IT, MC, SM, TR, BE, EE, HU, LV, NL, SK, GB, BG, FI, IS, LI, NO, SI, VA, FR, IL, LT, PL, ES, CZ, DE, IE, LU, PT, SE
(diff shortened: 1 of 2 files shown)


j-mutter:
Um, what's the reasoning for this? Does this product/endpoint not actually support US?

ntalbott:
@j-mutter so the deal is, Worldpay has a bunch of different products, and they don't provision the product that ActiveMerchant calls "Worldpay" for US merchants. The country list is supposed to indicate where a gateway account is acquirable, not where the gateway can be used for processing, so this change (better) reflects reality.

@markabe I'd also like us to change the display name of the `Worldpay` gateway to `Worldpay Global` to help reflect this distinction.

Overall, Worldpay is a mess, and after lots of merchant confusion, this is one piece of what we're doing to try to help future merchants be less confused.

markabe:
@ntalbott Updated display name.

ntalbott:
@j-mutter any concerns remain before I merge? We could do this just on the Spreedly side, but our thinking was that it wasn't just an issue for our customers.

j-mutter:
As long as it's just the display name that's changed, I think we're ok. We'll need to update some stuff on our end, but otherwise I think this is ok.  @girasquid thoughts?

ntalbott:
@j-mutter so this is going to change the display name, and the supported countries for the `worldpay` adapter. No change to the class name and such, though - this is all metadata tweaks.

girasquid:
> No change to the class name and such, though - this is all metadata tweaks.

In light of that, I don't think we have any issues here. Thanks for doing this!

PR #1583, opened 2015-02-25
Remove Ruby 1.9 from travis.yml as it is no longer supported

j-mutter:
Alright, we're no longer building for ruby 1.9 on Travis. Opinions on *enforcing* 2.0+?

@girasquid @ntalbott @wvanbergen

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ rvm:
 - "2.2"
 - "2.1"
 - "2.0"
-- "1.9"
 
 gemfile:
 - Gemfile.rails32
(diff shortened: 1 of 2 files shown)


ntalbott:
@j-mutter I'm anti-enforcement. I'd say we just only test on 2.0+, and document that as what we target for contributions. If someone can still hack it to work, that's up to them. We haven't had very many Ruby version compatibility support issues come up since dropping 1.8, so I'm not really worried about it.

j-mutter:
Same, I'll add a note to the GettingStarted or something about ruby versions.

wvanbergen:
Does that mean we actively should avoid Ruby features we know that won't
work in Ruby 1.9? (Module#prepend, named parameters, etc)

ntalbott:
@wvanbergen no, we're OK breaking backwards compatibility, but we'll let the language do it vs. forcing it via an ActiveMerchant error.

girasquid:
:+1: for not enforcing

PR #1588, opened 2015-02-27
Making error message human readable for shopify customers

MauricioMurga:
Making error message human readable for shopify customers.

### DIFF
--- a/lib/active_merchant/billing/gateways/conekta.rb
+++ b/lib/active_merchant/billing/gateways/conekta.rb
@@ -69,7 +69,7 @@ module ActiveMerchant #:nodoc:
         details[:name] = options[:customer] if options[:customer]
         details[:email] = options[:email] if options[:email]
         details[:phone] = options[:phone] if options[:phone]
-        details[:device_fingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
+        post[:device_fingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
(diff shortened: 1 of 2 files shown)


j-mutter:
@MauricioMurga is there documentation somewhere I can see to confirm this is the correct field to use? I don't see it anywhere in [these docs](https://www.conekta.io/en/docs/api).

lib/active_merchant/billing/gateways/conekta.rb
```diff
@@ -197,8 +197,9 @@ def response_error(raw_response)
       end
 
       def json_error(raw_response)
-        msg = 'Invalid response received from the Conekta API.'
-        msg += "  (The raw response returned by the API was #{raw_response.inspect})"
+        #msg = 'Invalid response received from the Conekta API.'
+        #msg += "  (The raw response returned by the API was #{raw_response.inspect})"
```
j-mutter: These can be deleted. Please don't submit PRs with commented out code, it just means we have to go in and remove it when we merge.

MauricioMurga:
@j-mutter, message_to_purchaser documentation can be found here: https://www.conekta.io/en/docs/api#errors

j-mutter:
Oh weird, I must have typo'd my search when I first looked. Thanks :)

MauricioMurga:
@j-mutter, no problem :)

ntalbott:
Looks like this since got merged elsewhere.

PR #1589, opened 2015-02-28
Add Qvalent gateway

markabe:


markabe:
@duff @ntalbott please review.

test/remote/gateways/remote_qvalent_test.rb
```diff
@@ -0,0 +1,106 @@
+require "test_helper"
+
+class RemoteQvalentTest < Test::Unit::TestCase
+  def setup
+    Base.mode = :test
```
ntalbott: FYI, this is done by `test_helper.rb`, so not necessary here. I'll remove when I merge.

test/remote/gateways/remote_qvalent_test.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    assert refund = @gateway.refund(@amount, response.authorization)
+    assert_success refund
+    assert_equal "Succeeded", refund.message
+  end
+
+  def test_failed_refund
+    response = @gateway.refund(nil, "")
+    assert_failure response
+    assert response.message =~ /Invalid card number/
```
ntalbott: Prefer `assert_match %r{regexp}, message` since it gives better output. Will fix when I merge.

PR #1594, opened 2015-03-03
Improve Quickbooks transcript scrubbing regex

girasquid:
The original regex was a bit too firm about the structure of the fields; this updates it to handle a card number being immediately after the word "number", and a CVC being immediately after the word "cvc".

@j-mutter @jnormore for review.

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -104,8 +104,8 @@ module ActiveMerchant #:nodoc:
           gsub(%r((oauth_nonce=\")\w+), '\1[FILTERED]').
           gsub(%r((oauth_signature=\")[a-zA-Z%0-9]+), '\1[FILTERED]').
           gsub(%r((oauth_token=\")\w+), '\1[FILTERED]').
-          gsub(%r((\"card\":{\"number\":\")\d+), '\1[FILTERED]').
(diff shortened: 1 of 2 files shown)


jnormore:
:+1:

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -104,8 +104,8 @@ def scrub(transcript)
           gsub(%r((oauth_nonce=\")\w+), '\1[FILTERED]').
           gsub(%r((oauth_signature=\")[a-zA-Z%0-9]+), '\1[FILTERED]').
           gsub(%r((oauth_token=\")\w+), '\1[FILTERED]').
-          gsub(%r((\"card\":{\"number\":\")\d+), '\1[FILTERED]').
-          gsub(%r((\"cvc\":\")\d+), '\1[FILTERED]')
+          gsub(%r((number\D+)\d{16}), '\1[FILTERED]').
+          gsub(%r((cvc\D+)\d{3}([^,]+)), '\1[FILTERED]\2')
```
j-mutter: any reason these are different, considering the patterns are the same other than the length of the number to filter?

girasquid: Whoops - I don't need the trailing part. Removed.

j-mutter:
:hatched_chick:

girasquid:
Forgot to close this in the merge.

PR #1596, opened 2015-03-04
Name/verification fields are optional with network tokenization

rwdaigle:
Some network tokenization implementations, specifically Apple Pay, do not
provide, or make optional, the name and verification value fields.
NetworkTokenizationCreditCard should not require these fields.

Does this look sane, @jnormore?

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -49,7 +49,13 @@ module ActiveMerchant #:nodoc:
     class CreditCard < Model
       include CreditCardMethods
 
-      cattr_accessor :require_verification_value
+      class << self
+        # Inherited, but can be overridden w/o changing parent's value
+        attr_accessor :require_verification_value
+        attr_accessor :require_name
+      end
+
+      self.require_name = true
       self.require_verification_value = true
 
       # Returns or sets the credit card number.
@@ -243,13 +249,19 @@ module ActiveMerchant #:nodoc:
         require_verification_value
       end
 
+      def self.requires_name?
+        require_name
+      end
+
       private
 
       def validate_essential_attributes #:nodoc:
         errors = []
 
-        errors << [:first_name, "cannot be empty"] if first_name.blank?
-        errors << [:last_name,  "cannot be empty"] if last_name.blank?
+        if self.class.requires_name?
+          errors << [:first_name, "cannot be empty"] if first_name.blank?
+          errors << [:last_name,  "cannot be empty"] if last_name.blank?
+        end
 
         if(empty?(month) || empty?(year))
           errors << [:month, "is required"] if empty?(month)
@@ -294,7 +306,7 @@ module ActiveMerchant #:nodoc:
           unless valid_card_verification_value?(verification_value, brand)
             errors << [:verification_value, "should be #{card_verification_value_length(brand)} digits"]
           end
-        elsif CreditCard.requires_verification_value?
+        elsif self.class.requires_verification_value?
           errors << [:verification_value, "is required"]
         end
         errors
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/credit_card.rb
```diff
@@ -49,8 +49,10 @@ module Billing #:nodoc:
     class CreditCard < Model
       include CreditCardMethods
 
-      cattr_accessor :require_verification_value
+      class_attribute :require_verification_value
```
jnormore: why class_attribute?

jnormore: This won't work well in multi threaded environments :(

rwdaigle: @jnormore this was necessary so sub-classes (i.e. `NetworkTokenizedCreditCard` could define their own required fields without changing their parent class's settings. Without this, when `NetworkTokenizedCreditCard` sets `self.require_verification_value = false` it would cause `CreditCard` to not require a CVV as well.

Regarding multi-threaded environments, would an instance variable based approach be preferred?

```ruby
class CreditCard
  self << class
    attr_accessor :require_verification_value
  end
end
```

I believe this is thread-safe while accomplishing the [intended function](https://coderwall.com/p/xhcmbg/cattr_accessor-vs-class_attribute).

jnormore: yeah you're right, I mis-interpreted it :)

jnormore:
You're right that apple pay doesn't provide the name in the decrypted blob (most of the time) but some gateways will still require it and it's available to the device when it's generating the payment token request (``requiredBillingAddressFields`` from https://developer.apple.com/library/ios/ApplePay_Guide/CreateRequest.html#//apple_ref/doc/uid/TP40014764-CH3-SW2)

I guess providing the ability to skip validation here would make sense though in case the gateway doesn't care.

rwdaigle:
> it (customer name) is available to the device when it's generating the payment token request

Actually, it looks like name is [not part](https://developer.apple.com/library/ios/documentation/PassKit/Reference/PKPaymentRequest_Ref/index.html#//apple_ref/c/tdef/PKAddressField) of `PKAddressField` (which is what's used to specify required billing or shipping data). So... I do still think name should be treated as a non-required field.

@jnormore If I fix the multi-threaded issue, are we :+1:?

jnormore:
absolutely, I think this is necessary anyway :)

rwdaigle:
@jnormore inheritable class attribute logic has [been updated](https://github.com/Shopify/active_merchant/commit/924d53de48ed98e6a218800f463c303c7b9ca271#diff-5f60353ee3917b9a473e169bde83d99dR52)

lib/active_merchant/billing/credit_card.rb
```diff
@@ -243,13 +249,19 @@ def self.requires_verification_value?
...
+      end
+
       private
 
       def validate_essential_attributes #:nodoc:
         errors = []
 
-        errors << [:first_name, "cannot be empty"] if first_name.blank?
-        errors << [:last_name,  "cannot be empty"] if last_name.blank?
+        if(self.class.requires_name?)
```
jnormore: super picky of me but can you remove the brackets? :)

jnormore:
Picky comment but LGTM

jnormore:
@bizla can you review too please?

lib/active_merchant/billing/credit_card.rb
```diff
@@ -49,7 +49,13 @@ module Billing #:nodoc:
     class CreditCard < Model
       include CreditCardMethods
 
-      cattr_accessor :require_verification_value
+      class << self
+        # Inheritable, but over-ridable
```
bizla: `but can be overridden`

lib/active_merchant/billing/credit_card.rb
bizla: +1

bizla:
Minor nitpick here too, otherwise looks good.

rwdaigle:
Nits addressed (and happy to do so)! Ready to merge.

jnormore:
:ship: when ready

bizla:
Merging.

PR #1597, opened 2015-03-04
check status field on responses from the gateway

ivanfer:
A declined transaction on `QuickBooks` gateway assumed that there would always be an error object in the response - sometimes there are not. 

@girasquid

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -242,11 +242,13 @@ module ActiveMerchant #:nodoc:
       end
 
       def success?(response)
-        response['errors'].present? ? FRAUD_WARNING_CODES.concat(['0']).include?(response['errors'].first['code']) : true
+      return FRAUD_WARNING_CODES.concat(['0']).include?(response['errors'].first['code']) if response['errors']
+        
+      !['DECLINED', 'CANCELLED'].include?(response['status'])
       end
 
       def message_from(response)
-        response['errors'].present? ? response["errors"].map {|error_hash| error_hash["message"] }.join(" ") : "Transaction Approved"
+        response['errors'].present? ? response["errors"].map {|error_hash| error_hash["message"] }.join(" ") : response['status']
       end
 
       def errors_from(response)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -242,11 +242,13 @@ def cvv_code_from(response)
...
       def success?(response)
-        response['errors'].present? ? FRAUD_WARNING_CODES.concat(['0']).include?(response['errors'].first['code']) : true
+      return FRAUD_WARNING_CODES.concat(['0']).include?(response['errors'].first['code']) if response['errors']
+        
+      !['DECLINED', 'CANCELLED'].include?(response['status'])
       end
 
       def message_from(response)
-        response['errors'].present? ? response["errors"].map {|error_hash| error_hash["message"] }.join(" ") : "Transaction Approved"
+        response['errors'].present? ? response["errors"].map {|error_hash| error_hash["message"] }.join(" ") : response['status']
```
girasquid: What is the message going to be, when it comes from the status? I can see that for declines/cancels it will be "DECLINED" or "CANCELLED".

test/remote/gateways/remote_quickbooks_test.rb
```diff
@@ -22,7 +22,7 @@ def setup
   def test_successful_purchase
     response = @gateway.purchase(@amount, @credit_card, @options)
     assert_success response
-    assert_equal 'Transaction Approved', response.message
+    assert_equal 'CAPTURED', response.message
```
girasquid: Ah - I see now. Do we want this to be the message, or do we still want "Transaction Approved"?

girasquid:
:+1:

lib/active_merchant/billing/gateways/quickbooks.rb
ivanfer: For charges - ` AUTHORIZED, DECLINED, CAPTURED, CANCELLED, SETTLED, REFUNDED
`
For refunds - `ISSUED, DECLINED, SETTLED`

test/remote/gateways/remote_quickbooks_test.rb
ivanfer: I think it's more descriptive this way

lib/active_merchant/billing/gateways/quickbooks.rb
girasquid: :cool: - this works.

ivanfer:
Closed by https://github.com/Shopify/active_merchant/commit/abe433f6ef8c3370eeb1a6da2fb2f08bed102baf

PR #1598, opened 2015-03-04
Fix a simple argument typo in Qvalent add_reference

bruno:
The method had an argument that contained a simple typo, `opions` instead of `options`.


j-mutter:
Nice catch! Thanks

PR #1599, opened 2015-03-05
Remove integration_mode getter and setter.

mattfawcett:
The integration_mode getter/setter are no longer used, and the deprecation error it would fire results in an error due to the missing constant `OFFSITE_PAYMENT_EXTRACTION_MESSAGE`


ntalbott:
Thanks so much for this!

PR #1603, opened 2015-03-06
PayConex: Add gateway

duff:
https://www.bluefin.com/


duff:
Got the encoding issue to work in Ruby 2.0 locally.  Let's see if it's fixed in Travis.

PR #1605, opened 2015-03-09
send billing address with PayPal reference transaction for successful AV...

pledgemusic:
We've been advised by PayPal that sending the billing address with a reference transaction will avoid issues with AVS check failing.  So here we're adding the address is if supplied and if performing a DoReferenceTransaction.


ntalbott:
@pledgemusic can you add a unit and/or remote test for this? Thanks!

jasonwebster:

PR #1607, opened 2015-03-11
JetPay REVERSEAUTH functionality

brokenbeatnik:
Added proposed reverseauth options to void request for JetPay, which distinguishes between voiding an auth and voiding a capture. Fixed previously broken test case. Closes shopify/active_merchant#1606.

This was spurred by an customer using JetPay with a workflow that required them to be able to void an auth. The void method as implemented voided a capture but not an auth, and would return an error if you attempted to void the auth. Per the XML Specification at https://www.jetpay.com/resources/library/pdf/xml/JetPay_XML_Specification-20120216.pdf, a REVERSEAUTH request (section 3.6, p. 16) must be submitted to complete this. Since there is only one void method, the reverseauth flag and the credit card number are sent in through the options hash. If the flag :reverseauth is set to true, the credit card is used and a REVERSEAUTH is submitted. If not, the VOID request is submitted as before. 

I added a test case for the remote_jetpay_test and ensured it passed. It appears the VOID test was failing, so I made it capture before attempting a proper VOID.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,7 @@
 * Qvalent: Fix scrub replacement, it was too greedy [markabe]
 * PayConex: Add gateway [duff]
 * AuthorizeNet: Add credit support [duff]
+* JetPay: Add REVERSEAUTH support to void [brokenbeatnik]
 
 == Version 1.47.0 (February 25, 2015)
 
--- a/lib/active_merchant/billing/gateways/jetpay.rb
+++ b/lib/active_merchant/billing/gateways/jetpay.rb
@@ -81,7 +81,12 @@ module ActiveMerchant #:nodoc:
 
       def void(reference, options = {})
         transaction_id, approval, amount = reference.split(";")
-        commit(amount.to_i, build_void_request(amount.to_i, transaction_id, approval))
+        if (options[:reverseauth] == true)
+          request = build_reverseauth_request(amount.to_i, transaction_id, approval, options[:credit_card]) 
+        else
+          request = build_void_request(amount.to_i, transaction_id, approval)
+        end
+        commit(amount.to_i, request)
       end
 
       def credit(money, transaction_id_or_card, options = {})
@@ -151,6 +156,16 @@ module ActiveMerchant #:nodoc:
           xml.tag! 'Approval', approval
           xml.tag! 'TotalAmount', amount(money)
 
+          xml.target!
+        end
+      end
+      
+      def build_reverseauth_request(money, transaction_id, approval, card)
+        build_xml_request('REVERSEAUTH', transaction_id) do |xml|
+          add_credit_card(xml, card)
+          xml.tag! 'Approval', approval
+          xml.tag! 'TotalAmount', amount(money)
+          
           xml.target!
         end
       end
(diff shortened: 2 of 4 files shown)


ntalbott:
@brokenbeatnik I like the spirit & functionality of this change, but I think the implementation needs to change.

The issue is that requiring a gateway specific option in order for a core API method to work is a non-starter. What should be done instead is to encode the type of the "authorization" into the authorization itself. So when `authorize` and `capture` are called, and they return their respective `Response#authorization` value, that value should contain a field (as per other gateways that have multi-part authorization strings) indicating the type of reference contained in the `authorization`. `void` would then unpack the authorization string and select the right JetPay call without the caller having to know which it was originally.

For authorization strings from pre-existing transactions that don't have the type encoded, you would just continue to default to the capture type, which would be broken, but it has been for years now.

Does this make sense? Anything I can clarify?

brokenbeatnik:
Thanks for the insight and feedback @ntalbott . So are you saying I should modify authorization_from to also return whether it was an auth or capture? I would add that to the end of the auth string (which for this gateway currently contains 3 fields) as an optional field. If not present, void would assume capture. 

Would it still be okay to send the card in via the options hash? The REVERSEAUTH seems to require a card number, even though VOID doesn't seem to.

ntalbott:
:+1: on your interpretation.

Unfortunately, requiring a card number for `void` is just broken in general; most systems built on AM, such as Spreedly, won't support doing that. Can you verify with JetPay that that's required, and that it's the full card number and not just a non-PCI relevant subset of the card number like the last four?

brokenbeatnik:
I'm waiting on a response from JetPay engineering. They're discussing it, and once they let me know, I'll make the changes and probably submit a new pull request and close this one.

zelch:
Having reviewed the matter, we concur that the current functionality is not ideal and should be corrected.

In the long term, we intend to modify VOID so that it functions properly for transactions in both the captured and non-captured states, however we do not expect this to be available until some time after spring release.

In the meantime me recommend the somewhat awkward steps of doing a CAPTURE followed by a VOID, both of which have options that do not require a card number.

We apologize for the inconvenience.

Regards,
Zephaniah E. Loss-Cutler-Hull.
Sr. Software Engineer.
JetPay.

brokenbeatnik:
Given JetPay's comments, I will close this pull request soon, but I have one more question. @zelch - does the customer incur more transaction fees for the CAPTURE/VOID than for the REVERSEAUTH? If so, then I'll "manually" implement the REVERSEAUTH outside of ActiveMerchant, which is less desirable. If not, I'll implement the recommended CAPTURE/VOID flow.

brokenbeatnik:
Final word: the client cannot allow a capture/void, as to the customer it will be a flag as a false-positive fraud scenario (imagine how you'd feel if you saw $200 taken off your card by some party, even if it were refunded immediately. We'll have to run on the forked version until the VOID performs REVERSEAUTH for incomplete transactions. I'll close the pull request since in any case this code won't make it in to the main line, and JetPay users will wait for JetPay central to make the changes to how VOID works behind the scenes.

Thanks all!

girasquid:
Thanks for letting us know!

PR #1611, opened 2015-03-12
Cenpos

markabe:
@ntalbott Please review.


markabe:
Might as well flag you too @duff :smile:

duff:
Does CenPOS offer any avs_response and/or cvv_response information?

markabe:
There did appear to be some information in the response.  However I tried to altering the requests in order to affect the response and it always came back Match.  I didn't have much confidence that it did what it should so I left it out.

duff:
I don't see much info at all in their docs about it either. :(

duff:
Closed by 94b7645fe93446fde3173de1e0f5a28dc25d31d6.

PR #1612, opened 2015-03-12
mercury gateway default to partial auth deny

shulmang:
mercury gateway default to partial auth deny

### DIFF
--- a/lib/active_merchant/billing/gateways/mercury.rb
+++ b/lib/active_merchant/billing/gateways/mercury.rb
@@ -90,7 +90,7 @@ module ActiveMerchant #:nodoc:
           xml.tag! "Transaction" do
             xml.tag! 'TranType', 'Credit'
             xml.tag! 'TranCode', action
-            if action == 'PreAuth' || action == 'Sale'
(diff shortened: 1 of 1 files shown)


andrewpaliga:
Speaking with Mercury support I can verify that this is desired behaviour. Enabling Partial Auth will authorize the transaction for the remaining balance on the card. 

ex. You are buying something for $100 and there is only $50 left on the card. The transaction would be approved for $50 with Partial Auth tag passed, and declined with Partial Auth tag left out. 

I think the latter is a more sensible default.

Note that Partial Auth only applies to card present transactions. When Partial Auth is supported there are additional requirements on the POS software to handle these cases (prompt for payment completion etc.). 

@j-mutter @ntalbott do these changes look good?

ntalbott:
Yup, in a CNP context, can't ever imagining wanting that to happen. :+1: for this change.

j-mutter:
Tests? :innocent:

ntalbott:
Implemented in https://github.com/activemerchant/active_merchant/commit/ebf1293228555c0e4d2512f61e1e89d6e44b59d2.

PR #1613, opened 2015-03-12
Braintree gem update

matdevdug:
Updated the Braintree gem to the latest version.


j-mutter:
We're currently point to a side branch of the Braintree gem to support ApplePay transactions, which is actually built on top of 2.40.0.

PR #1619, opened 2015-03-17
Add new gateway - WorldpayOnlinePayments

ao:


lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+
+      self.supported_countries = %w(HK US GB AU AD BE CH CY CZ DE DK ES FI FR GI GR HU IE IL IT LI LU MC MT NL NO NZ PL PT SE SG SI SM TR UM VA)
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover, :jcb, :maestro, :laser, :switch]
+
+      self.homepage_url = 'http://online.worldpay.com'
+      self.display_name = 'Worldpay Online Payments'
+
+      def initialize(options={})
+        requires!(options, :client_key)
+        requires!(options, :service_key)
```
j-mutter: These can be combined into a single `requires!` call

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+        @client_key = options[:client_key]
+        @service_key = options[:service_key]
+        super
+      end
+
+      def authorize(money, creditcard, options={})
```
j-mutter: Please rename to `credit_card` to be consistent with all the other gateways.

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+        # create token
+        response = create_token(true, creditcard.first_name+' '+creditcard.last_name, creditcard.month, creditcard.year, creditcard.number, creditcard.verification_value)
```
j-mutter: This method signature is a little awkward - why not just pass in the `credit_card` and let the method sort it out?

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+        if response.success?
+          # Create Authorize Only Order
+          options[:authorizeOnly] = true
+          post = create_post_for_auth_or_purchase(response.authorization, money, options)
+          response = commit(:post, 'orders', post)
+        end
+        response
+      end
+
+      def capture(money, orderCode, options={})
```
j-mutter: `authorization` instead of `orderCode` please :)

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+        if (orderCode)
```
j-mutter: This param is required, so you don't need to check for it.

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
j-mutter: Also, can you use a `MultiResponse` here since you're making two requests?

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+            :test => @service_key[0]=="T" ? true : false,
+            :authorization => false,
+            :avs_result => {},
+            :cvv_result => {},
+            :error_code => false
+          )
+        end
+      end
+
+      def purchase(money, creditcard, options={})
```
j-mutter: Same comment as above re: `credit_card`

j-mutter: And using a `MultiResponse`

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+          "countryCode"=>options[:billing_address]&&options[:billing_address][:country] ? options[:billing_address][:country] : ''
+          },
+          "customerOrderCode" => options[:order_id],
+          "orderType" => "ECOM",
+          "authorizeOnly" => options[:authorizeOnly] ? true : false
+        }
+      end
+
+      def parse(body)
+        if (!body)
```
j-mutter: `def parse(body={})` will do this for you, but when would it ever be `nil`?

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+        end
+        headers
+      end
+
+      def commit(method, url, parameters=nil, options = {})
+
+        raw_response = response = nil
+        success = false
+        begin
+          if parameters == nil
```
j-mutter: `json = parameters ? parameters.to_json : nil`

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+            json = parameters.to_json
+          end
+
+          raw_response = ssl_request(method, self.live_url + url, json, headers(options))
+
+          if (raw_response != '')
+            response = parse(raw_response)
+            success = !response.key?("httpStatusCode")
+          else
+            success = true
```
j-mutter: So, an empty response is deemed a success? That seems a little backwards, but I'm not familiar with this API

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,216 @@
...
+        elsif response["token"]
+          authorization = response["token"]
+        else
+          authorization = response["message"]
+        end
+
+        Response.new(success,
+          success ? "SUCCESS" : response["message"],
+          response,
+          :test => @service_key[0]=="T" ? true : false,
```
j-mutter: Can you move this check into `test?` method?

test/unit/gateways/worldpay_online_payments_test.rb
```diff
@@ -0,0 +1,252 @@
+require 'test_helper'
+
+class WorldpayOnlinePaymentsTest < Test::Unit::TestCase
+  def setup
+    @gateway = WorldpayOnlinePaymentsGateway.new(
```
j-mutter: Why not just use the fixture?

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
ao: 200 is deemed successful in this case

j-mutter: But the response is completely empty? Just seems a little odd.

ao: This is how the API responds in this case instance.

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -0,0 +1,212 @@
...
+
+      def refund(money, orderCode, options={})
+        obj = money ? {"refundAmount" => money} : {}
+        commit(:post, "orders/#{CGI.escape(orderCode)}/refund", obj, options)
+      end
+
+      def void(orderCode, options={})
+        response = commit(:delete, "orders/#{CGI.escape(orderCode)}", nil, options)
+        puts response
+        exit
```
j-mutter: This is… not good :innocent:

PR #1624, opened 2015-03-19
Nmi vault gateway

chrisblatchley:
Added NMI Vault Gateway


chrisblatchley:
This PR is redundant.

PR #1626, opened 2015-03-20
New Gateway: SecurionPay

szajbus:
[Review on Reviewable](https://reviewable.io/reviews/shopify/active_merchant/1626)

### DIFF
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -79,16 +79,8 @@ module ActiveMerchant #:nodoc:
 
         post[:description] = options[:description] if options[:description]
         post[:email] = options[:email] if options[:email]
-        if options[:customer]
-          update_customer(options[:customer], post, options)
-        else
-          MultiResponse.run(:first) do |r|
-            r.process { verify(payment, options) }
-            if r.params["id"].present?
-              r.process { commit(:post, "customers", post, options) }
-            end
-          end
-        end
+
+        commit(:post, 'customers', post, options)
       end
 
       def update_customer(customer_id, post, options = {})
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -4,7 +4,7 @@ module ActiveMerchant #:nodoc:
       self.test_url = 'https://api.securionpay.com/'
       self.live_url = 'https://api.securionpay.com/'
 
-      self.supported_countries = ['US']
+      self.supported_countries = Country::COUNTRIES.map { |country| country[:alpha2] }
       self.default_currency = 'USD'
       self.money_format = :cents
       self.supported_cardtypes = [:visa, :master, :american_express, :discover, :jcb, :diners_club]
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -145,9 +145,10 @@ module ActiveMerchant #:nodoc:
         add_customer(post, payment, options)
         add_customer_data(post, options)
         post[:description] = options[:description]
-        post[:metadata] = {}
-        post[:metadata][:email] = options[:email] if options[:email]
-        post.delete(:metadata) if post[:metadata].empty?
+        if options[:email]
+          post[:metadata] = {}
+          post[:metadata][:email] = options[:email]
+        end
         post
       end
 
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -83,7 +83,13 @@ module ActiveMerchant #:nodoc:
         commit(:post, 'customers', post, options)
       end
 
-      def update_customer(customer_id, post, options = {})
+      def update_customer(customer_id, payment, options = {})
+        post = {}
+        add_creditcard(post, payment, options)
+
+        post[:description] = options[:description] if options[:description]
+        post[:email] = options[:email] if options[:email]
+
         commit(:post, "customers/#{CGI.escape(customer_id)}", post, options)
       end
 
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -31,7 +31,7 @@ module ActiveMerchant #:nodoc:
 
       def initialize(options={})
         requires!(options, :secret_key)
-        @secret_key = options[:secret_key]
+        @options = options
 
         super
       end
@@ -199,7 +199,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def headers(options = {})
-        secret_key = options[:secret_key] || @secret_key
+        secret_key = options[:secret_key] || @options[:secret_key]
 
         headers = {
           "Authorization" => "Basic " + Base64.encode64(secret_key.to_s + ":").strip,
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -187,17 +187,11 @@ module ActiveMerchant #:nodoc:
         response = api_request(method, url, parameters, options)
         success = !response.key?("error")
 
-        # card = response["card"] || response["active_card"] || {}
-        # avs_code = AVS_CODE_TRANSLATOR["line1: #{card["address_line1_check"]}, zip: #{card["address_zip_check"]}"]
-        # cvc_code = CVC_CODE_TRANSLATOR[card["cvc_check"]]
         Response.new(success,
           success ? "Transaction approved" : response["error"]["message"],
           response,
           test: test?,
           authorization: success ? response["id"] : response["error"]["charge"],
-          # avs_result: { :code => avs_code },
-          # cvv_result: cvc_code,
-          # error_code:  success ? nil : STANDARD_ERROR_CODE_MAPPING[response["error"]["code"]]
           error_code: success ? nil : STANDARD_ERROR_CODE_MAPPING[response["error"]["code"]]
         )
       end
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -130,8 +130,12 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_customer_data(post, options)
-        metadata_options = [:description, :ip, :user_agent, :referrer]
-        post.update(options.slice(*metadata_options))
+        post.update(
+          description: options[:description],
+          ip:          options[:ip],
+          user_agent:  options[:user_agent],
+          referrer:    options[:referrer]
+        )
       end
 
       def create_post_for_auth_or_purchase(money, payment, options)
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -197,15 +197,11 @@ module ActiveMerchant #:nodoc:
 
       def headers(options = {})
         secret_key = options[:secret_key] || @secret_key
-        # version = options[:version] || @version
 
         headers = {
           "Authorization" => "Basic " + Base64.encode64(secret_key.to_s + ":").strip,
-          # "User-Agent" => "Stripe/v1 ActiveMerchantBindings/#{ActiveMerchant::VERSION}",
-          # "X-Stripe-Client-User-Agent" => user_agent,
-          # "X-Stripe-Client-User-Metadata" => {:ip => options[:ip]}.to_json
+          "User-Agent" => "SecurionPay/v1 ActiveMerchantBindings/#{ActiveMerchant::VERSION}"
         }
-        # headers.merge!("Stripe-Version" => version) if version
         headers
       end
 
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -1,0 +1,268 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class SecurionPayGateway < Gateway
+      self.test_url = 'https://api.securionpay.com/'
+      self.live_url = 'https://api.securionpay.com/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class SecurionPayGateway < Gateway
+      self.test_url = 'https://api.securionpay.com/'
+      self.live_url = 'https://api.securionpay.com/'
+
+      self.supported_countries = Country::COUNTRIES.map { |country| country[:alpha2] }
```
ntalbott: Please list out countries individually; this should reflect countries that merchants can get a SecurionPay account in.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+        'card_declined' => STANDARD_ERROR_CODE[:card_declined],
+        'processing_error' => STANDARD_ERROR_CODE[:processing_error],
+        'lost_or_stolen' => STANDARD_ERROR_CODE[:card_declined],
+        'suspected_fraud' => STANDARD_ERROR_CODE[:card_declined],
+        'expired_token' => STANDARD_ERROR_CODE[:card_declined]
+      }
+
+      def initialize(options={})
+        requires!(options, :secret_key)
+        @secret_key = options[:secret_key]
```
ntalbott: Don't save options off, just use them via `@options`.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+        post = {}
+        add_creditcard(post, payment, options)
+
+        post[:description] = options[:description] if options[:description]
+        post[:email] = options[:email] if options[:email]
+        if options[:customer]
+          update_customer(options[:customer], post, options)
+        else
+          MultiResponse.run(:first) do |r|
+            r.process { verify(payment, options) }
```
ntalbott: Why is `store` calling `verify`? Can you ditch the MultiResponse and just make a single call to store a card?

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+            if r.params["id"].present?
+              r.process { commit(:post, "customers", post, options) }
+            end
+          end
+        end
+      end
+
+      def update_customer(customer_id, post, options = {})
```
ntalbott: This shouldn't be a public method.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+        end
+      end
+
+      def store(payment, options = {})
+        post = {}
+        add_creditcard(post, payment, options)
+
+        post[:description] = options[:description] if options[:description]
+        post[:email] = options[:email] if options[:email]
+        if options[:customer]
```
ntalbott: Why is `store` doing an update? Simplify to just doing a store please.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+
+      def unstore(customer_id, options = {})
+        if options[:card_id]
+          commit(:delete, "customers/#{CGI.escape(customer_id)}/cards/#{CGI.escape(options[:card_id])}", nil, options)
+        else
+          commit(:delete, "customers/#{CGI.escape(customer_id)}", nil, options)
+        end
+      end
+
+      def create_token(payment, options = {})
```
ntalbott: Maybe this is really what `store` should be? Regardless, this should not be a public method.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+        transcript.
+          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
+          gsub(%r((card\[number\]=)\d+), '\1[FILTERED]').
+          gsub(%r((card\[cvc\]=)\d+), '\1[FILTERED]')
+      end
+
+      private
+
+      def add_customer(post, payment, options)
+        post[:customer] = options[:customer] if options[:customer] && !payment.respond_to?(:number)
```
ntalbott: The `!payment.respond_to?(:number)`.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+      end
+
+      def add_customer_data(post, options)
+        metadata_options = [:description, :ip, :user_agent, :referrer]
+        post.update(options.slice(*metadata_options))
```
ntalbott: Lets just expand this out; we prefer explicit to implicit. So assign each thing vs. using `update`/`slice`.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,269 @@
...
+      def create_post_for_auth_or_purchase(money, payment, options)
+        post = {}
+        add_amount(post, money, options, true)
+        add_creditcard(post, payment, options)
+        add_customer(post, payment, options)
+        add_customer_data(post, options)
+        post[:description] = options[:description]
+        post[:metadata] = {}
+        post[:metadata][:email] = options[:email] if options[:email]
+        post.delete(:metadata) if post[:metadata].empty?
```
ntalbott: Not a fan of adding speculatively and then deleting. Should only get added if needed.

test/remote/gateways/remote_securion_pay_test.rb
```diff
@@ -0,0 +1,341 @@
...
+    transcript = @gateway.scrub(transcript)
+    assert_scrubbed(@credit_card.number, transcript)
+    assert_scrubbed(@credit_card.verification_value, transcript)
+    assert_scrubbed(@gateway.options[:secret_key], transcript)
+  end
+
+  def test_successful_purchase
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal "charge", response.params["objectType"]
```
ntalbott: Tests should only rarely be checking `Response#params`. Rather, check the common attributes on `Response` - `authorization`, `message`, `avs_code`, etc.

test/remote/gateways/remote_securion_pay_test.rb
```diff
@@ -0,0 +1,341 @@
...
+  def test_unsuccessful_purchase
+    assert response = @gateway.purchase(@amount, @declined_card, @options)
+    assert_failure response
+    assert_match %r{The card was declined for other reason.}, response.message
+    assert_match Gateway::STANDARD_ERROR_CODE[:card_declined], response.error_code
+  end
+
+  def test_successful_purchase_with_token
+    # Create token
+    assert response = @gateway.create_token(@credit_card)
```
ntalbott: As per above, `create_token` should not even be a public method. Though I'm wondering if it's properly `store`, and the current `store` is out of scope for ActiveMerchant.

test/remote/gateways/remote_securion_pay_test.rb
```diff
@@ -0,0 +1,341 @@
...
+    assert response = @gateway.authorize(@amount, @declined_card, @options)
+    assert_failure response
+  end
+
+  def test_partial_capture
+    assert auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount-1, auth.authorization)
+    assert_failure capture
```
ntalbott: Partial capture doesn't work?

test/remote/gateways/remote_securion_pay_test.rb
```diff
@@ -0,0 +1,341 @@
...
+    assert_success purchase
+    assert purchase.authorization
+
+    assert refund = @gateway.refund(@amount + 1, purchase.authorization, @options)
+    assert_failure refund
+    assert_match %r{Wrong Refund data}, refund.message
+  end
+
+  def test_successful_void
+    assert purchase = @gateway.purchase(@amount, @credit_card, @options)
```
ntalbott: The primary purpose of `void` in ActiveMerchant is to cancel authorizations - does it work for that? Can you test that scenario, please?

test/remote/gateways/remote_securion_pay_test.rb
```diff
@@ -0,0 +1,341 @@
...
+    assert_success response
+    assert_equal card_id, response.params['id']
+
+    # Unstore the customer
+    assert response = @gateway.unstore(customer_id)
+    assert_success response
+    assert_equal customer_id, response.params['id']
+  end
+
+  def test_successful_update
```
ntalbott: `update` in ActiveMerchant is for changing card details stored under a token. It's totally optional, so if SecurionPay doesn't support that, then I'd just drop it.

test/unit/gateways/securion_pay_test.rb
```diff
@@ -0,0 +1,858 @@
...
+    assert_equal @options[:billing_address][:address2], post[:card][:addressLine2]
+    assert_equal @options[:billing_address][:country], post[:card][:addressCountry]
+    assert_equal @options[:billing_address][:city], post[:card][:addressCity]
+  end
+
+  def test_ensure_does_not_respond_to_credit
+    assert !@gateway.respond_to?(:credit)
+  end
+
+  def test_gateway_without_credentials
```
ntalbott: This is overkill.

test/unit/gateways/securion_pay_test.rb
```diff
@@ -0,0 +1,858 @@
...
+
+  def test_address_is_included_with_card_data
+    stub_comms(@gateway, :ssl_request) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end.check_request do |method, endpoint, data, headers|
+      assert data =~ /card\[addressLine1\]/
+    end.respond_with(successful_purchase_response)
+  end
+
+  def generate_options_should_allow_key
```
ntalbott: I have no clue what this is, but doesn't look relevant.

test/unit/gateways/securion_pay_test.rb
```diff
@@ -0,0 +1,858 @@
...
+      -> "Content-Type: application/json;charset=UTF-8\r\n"
+      -> "Transfer-Encoding: chunked\r\n"
+      -> "Connection: close\r\n"
+      -> "Set-Cookie: __cfduid=d2825af5ba35f4e682da0c51abb33a5d91426807018; expires=Fri, 18-Mar-16 23:16:58 GMT; path=/; domain=.securionpay.com; HttpOnly\r\n"
+      -> "CF-RAY: 1c9ce35b76a805d5-WAW\r\n"
+      -> "Content-Encoding: gzip\r\n"
+      -> "\r\n"
+      -> "154\r\n"
+      reading 340 bytes...
+      -> "[GZIP_ENCODED_CONTENT_REMOVED_INTENTIONALLY]"
```
ntalbott: Is it possible to have responses never be gzipped? Perhaps via an `Accept` header?

ntalbott:
@szajbus left a bunch of comments; biggest issues are with matching the proper ActiveMerchant public API.

test/remote/gateways/remote_securion_pay_test.rb
szajbus: That's right. Partial captures are not supported by the gateway.

lib/active_merchant/billing/gateways/securion_pay.rb
szajbus: Actually `create_token` API method is meant for use from JavaScript or mobile applications, so this seems to be out of the scope of ActiveMerchant. Removed it.

test/unit/gateways/securion_pay_test.rb
szajbus: Yes, it is. Should it be a default header sent with all requests to the API? At the moment I just recorded the transcript with gzip disabled and replaced it here, but did not add the header to the requests.

szajbus:
@ntalbott Thank you for the review. We implemented the changes you suggested.

test/remote/gateways/remote_securion_pay_test.rb
ntalbott: @szajbus any feedback on my comment RE `void`-ing authorizations?

test/remote/gateways/remote_securion_pay_test.rb
ntalbott: @szajbus per my comment above, `update` as implemented does not match the correct API. Either remove or fix please.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -0,0 +1,276 @@
...
+        post = {}
+        add_creditcard(post, payment, options)
+
+        post[:description] = options[:description] if options[:description]
+        post[:email] = options[:email] if options[:email]
+
+        commit(:post, 'customers', post, options)
+      end
+
+      def update_customer(customer_id, payment, options = {})
```
ntalbott: This is not a supported ActiveMerchant public API. Remove.

ntalbott:
@szajbus I added a few more comments; also looks like you missed a few of my comments?

test/remote/gateways/remote_securion_pay_test.rb
szajbus: @ntalbott updated this test case to make `authorize` instead of `purchase`.

test/remote/gateways/remote_securion_pay_test.rb
szajbus: @ntalbott Removed completely.

ntalbott:
@szajbus you'll either need to remove `store`/`unstore`, or have them conform to the API. In particular, the `Response#authorization` after a `store` should contain a string with everything necessary to charge against the stored card, `purchase` and `authorize` should allow taking that returned string in place of a credit card and initiating a charge, and `unstore` should take just that string to remove the card from the vault.

A good example in ActiveMerchant of the correct API is the Authorize.Net adapter, which was recently enhanced to support the correct API for charging against Auth.Net's CIM.

ntalbott:
I went ahead and merged this with edits; in particular, I just stripped out all tokenization in the interest of getting it merged. If you'd still like to have tokenization supported, please do open a new PR that just does that and we can discuss there. Thanks!

szajbus:
@ntalbott thanks!

PR #1632, opened 2015-03-26
MonerisGateway: Added ResAddToken functionality

tzumby:
Hi guys,

This adds a new method called add_token to convert a temporary cc token into a permanent one. The temporary token is issued when using the Tokenized Moneris transaction (embedding their iframe to collect the credit card number). You can use that temporary token for 15 minutes but after that it expires. This is useful for customers who opt in to save their payment method for future purchases.

I added the failed and success responses to the spec but unfortunately it's not possible to test this. I could get a temporary token and add it to the tests but since it expires in 15 minutes it will keep failing. I reached out to Moneris to see if there's an easy way to get a temp token, or at least a "permanent" temporary token that can be used for tests but it doesn't look like it's possible.


j-mutter:
I'm not a fan of adding one-off public methods to gateways, as it defeats the point of using the AM abstractions. Is there a way this could just leverage the [`store`](https://github.com/tzumby/active_merchant/blob/master/lib/active_merchant/billing/gateways/moneris.rb#L134-L140) method maybe, which is the accepted API for saving tokenized cards.

ntalbott:
Yup, :-1: on a new public API method. Feels like this should just be a tweak to `store` that lets it take a temporary token (in addition to taking a full card).

tzumby:
Agreed, I will update the store method to check if the argument is a credit card or a token and call either res_add_cc or res_add_token.

tzumby:
I ended up just updating the store method and checking to see if the passed argument is a CreditCard object. 

I thought about maybe using the ActiveMerchant::Billing::NetworkTokenizationCreditCard object and have that one respond to "number", "year", "month" but since adding a credit card vs a token is a separate API call on the Moneris side we would still have to check weather the object is a NetworkTokenizationCreditCard vs CreditCard (so duck typing this would serve no purpose). 

Let me know what you guys think, I don't like the if statement there but I can't think of an alternative without touching too many other parts.

ntalbott:
@tzumby this looks generally OK, but will need a unit and remote test before we can merge it. Thanks!

tzumby:
@ntalbott I added the unit tests for storing and unstoring using a token and will push that commit. For the remote test the problem is Moneris doesn't provide a temporary token that doesn't expire. This means every time we run the remote tests we would have to use a fresh token. In my app I get around this by just recording the request with VCR (but at that point I'm really just using the cassette like a mock).

Let me know if just the unit test is fine and I'll push that, otherwise we can discuss about the remote test. I'm curious what you think.

ntalbott:
@tzumby it's typical to add a token each time the remote test is run, vs. reusing an existing one. Is that what you mean?

tzumby:
Yep, pretty much. We can't reuse the same token and there's no API we can use to generate those tokens. The only way is to use Moneris' Hosted Tokenization iframe and pass a bogus CC number. 
The flow is:

CC (via iframe) -> temp token (15 mins expiry) -> permanent token (no expiry). 

There other way to get a token is via the already existing method using the CC number.

Hope this explains is better.

ntalbott:
@tzumby oh, yuck. In that case, yes, I think just a unit test is fine. Thanks!

tzumby:
Done. 

This is kind of ugly, I didn't create the pull request off a branch so I'm not sure I can rebase this into one commit. If that's a problem I can close this one and create a new pull request.

ntalbott:
@tzumby I can easily rebase when I merge; so long as this PR only contains changes that should be merged, should be fine.

tzumby:
Thanks @ntalbott, double checked the changes and everything is good to go.

jasonwebster:

PR #1633, opened 2015-03-30
Send amount in Pin Payments capture requests

ab9:
## Summary

This changes the Pin Payments gateway class so that capture requests will send the amount. Pin Payments does not support partial captures, so currently if an Active Merchant user attempts to capture less than the originally-authorised amount, Pin Payments will capture the _full_ amount and report success. With the changes in this pull request, attempts to capture less than the originally-authorised amount will fail.

## Details

Active Merchant includes a gateway class for Pin Payments (ActiveMerchant::Billing::PinGateway). When it makes a capture request to Pin Payments, it does not include an amount parameter in the request. It was programmed this way because the Pin Payments API would simply ignore the amount parameter, just as it would any other unrecognised parameter (because Pin Payments does not do partial captures; [a capture is always of the full authorised amount](https://pin.net.au/docs/api/charges#put-charges)).

We recently modified the Pin Payments API so that users can include an amount parameter in capture requests. It works like a programming assertion: If it is included, the server will check that the amount parameter matches the originally authorised amount. If the amounts match, the full capture will be attempted; if the amounts do not match, no capture will be attempted and the Pin Payments API will return an error response to the user.

The changes in this pull request make Active Merchant take advantage of this new feature. We think this will be helpful for Active Merchant users: if an Active Merchant user attempts a partial capture and gets an erroneous response, he is less likely to be misled into thinking that Pin Payments performed a partial capture.

### DIFF
--- a/lib/active_merchant/billing/gateways/pin.rb
+++ b/lib/active_merchant/billing/gateways/pin.rb
@@ -62,7 +62,7 @@ module ActiveMerchant #:nodoc:
       # Captures a previously authorized charge. Capturing only part of the original
       # authorization is currently not supported.
       def capture(money, token, options = {})
-        commit(:put, "charges/#{CGI.escape(token)}/capture", {}, options)
+        commit(:put, "charges/#{CGI.escape(token)}/capture", { :amount => amount(money) }, options)
       end
 
       # Updates the credit card for the customer.
--- a/lib/active_merchant/billing/gateways/pin.rb
+++ b/lib/active_merchant/billing/gateways/pin.rb
@@ -59,7 +59,7 @@ module ActiveMerchant #:nodoc:
         purchase(money, creditcard, options)
       end
 
-      # Captures a previously authorized charge. Capturing a certin amount of the original
+      # Captures a previously authorized charge. Capturing only part of the original
       # authorization is currently not supported.
       def capture(money, token, options = {})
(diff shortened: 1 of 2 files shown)


ntalbott:
@ab9 generally a big :+1: from me, but what will this do if `nil` is passed for the amount? Just keep the old behavior?

andrewpaliga:
:+1: from me -- @j-mutter ?

j-mutter:
:+1: Same question re: `nil` amounts as @ntalbott

ab9:
@ntalbott @j-mutter Yeah, if the amount is nil, you'll get the old behavior: the Pin Payments app will attempt to capture the full amount originally authorised.

j-mutter:
Remotes look good.  Merging.
```bash
vagrant@vagrant:~/src/active_merchant$ bundle exec ruby -I"lib:test" test/remote/gateways/remote_pin_test.rb
Loaded suite test/remote/gateways/remote_pin_test
Started
.............

Finished in 35.820351471 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
13 tests, 39 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
0.36 tests/s, 1.09 assertions/s
```

PR #1634, opened 2015-03-30
Enhance Fat Zebra gateway with multi-currency and auth/capture

nagash:
Adds some extra features to the Fat Zebra gateway:

* Multi-currency support
* Authorize and capture support

### DIFF
--- a/lib/active_merchant/billing/gateways/fat_zebra.rb
+++ b/lib/active_merchant/billing/gateways/fat_zebra.rb
@@ -38,6 +38,7 @@ module ActiveMerchant #:nodoc:
 
         add_amount(post, money, options)
         add_creditcard(post, creditcard, options)
+        add_extra_options(post, options)
         post[:reference] = options[:order_id]
         post[:customer_ip] = options[:ip]
 
@@ -49,6 +50,7 @@ module ActiveMerchant #:nodoc:
 
         add_amount(post, money, options)
         add_creditcard(post, creditcard, options)
+        add_extra_options(post, options)
         post[:capture] = false
         post[:reference] = options[:order_id]
         post[:customer_ip] = options[:ip]
@@ -57,8 +59,10 @@ module ActiveMerchant #:nodoc:
       end
 
       def capture(money, authorization, options = {})
-        post = options
+        post = {}
         add_amount(post, money, options)
+        add_extra_options(post, options)
+
 
         commit(:post, "purchases/#{CGI.escape(authorization)}/capture", post)
       end
@@ -71,6 +75,7 @@ module ActiveMerchant #:nodoc:
       def refund(money, txn_id, reference)
         post = {}
 
+        add_extra_options(post, options)
         post[:amount] = money
         post[:transaction_id] = txn_id
         post[:reference] = reference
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/fat_zebra.rb
```diff
@@ -44,6 +44,36 @@ def purchase(money, creditcard, options = {})
...
+        post[:customer_ip] = options[:ip]
+
+        commit(:post, 'purchases', post)
+      end
+
+      # Capture a previous authorization
+      #
+      # money - Integer - the amount to capture
+      # txn_id - String - the original transaction to be captured
+      def capture(money, txn_id, options = {})
```
j-mutter: Minor nitpick, but can you use `authorization` instead of `txn_id` for the param name? And there's no reason to comment on what these methods do, they're part of the core ActiveMerchant API and should be well understood by anyone using them.

j-mutter:
LGTM. Do you have a link to the API documentation for this?

nagash:
@j-mutter Docs are here:
Multi currency: https://confluence.fatzebra.com.au/display/DOC/Purchases
Auth/Capture: https://confluence.fatzebra.com.au/display/DOC/Pre+Authorization+and+Capture

I've removed the comments and updated the txn_id param to be authorization.

j-mutter:
Awesome, thanks. I'll get this cleaned up and merged in. I checked out the docs, and it doesn't look like there is a `void` concept with this gateway, which usually goes along with an authorization (as opposed to a `refund` which is on an already-captured purchase). Do you know how Fat Zebra handles this? Can we just use the `refund` endpoint and just alias `void` to it?

nagash:
@j-mutter This lack of a void method seems to be pretty common across Australian gateways, presumably due to limitations in the underlying payments infrastructure. Authorizations will automatically be released after X days (determined by the issuing bank), there is no way for it to be controlled from the merchant/gateway side. Here's a snippet from a different Australian gateway (SecurePay):

> A preauthorisation allows reserve funds on a customer's credit card for five to 10 days, depending on the cardholder's bank. During that time, your can complete the transaction to capture the funds. The customer's bank will normally free up preauthorised funds that have not been completed within three days, making these funds available to the customer, and no longer reserved.

> Only a cardholder can release funds reserved on their card. They must call their issuing bank and request release of the preauthorisation. SecurePay and merchants do not have the authority to release funds reserved on a customer's bank account.

It's also the same for NAB Transact. The NAB Transact gateway within AM also has an authorize method with no corresponding void.

I will get in touch with my contact at Fat Zebra to confirm, but I assume this will be the case.

nagash:
Sorry for the last minute addition - just received word from Fat Zebra that this feature was ready to go this morning.

There is no official docs for the soft descriptor support yet, however some of the other extra fields are documented here - https://confluence.fatzebra.com.au/display/DOC/Extra+and+Extended+Fields

Here's the contents of the email received from Fat Zebra regarding this addition:

> Soft descriptors are ready to go and require the below simple change to the API:
> 
> The request payload needs to be updated to include the extra field and the name and location like below:
> ```{
>     "card_holder": "Jim Smith",
>     "card_number": "5123456789012346",
>     "card_expiry": "05/2023",
>     "cvv": "987",
>     "amount": 1000,
>     "reference": "ORD98976",
>     "customer_ip": "111.222.111.123",
>     "extra": {
>       "name": "Card Acceptor Name",
>       "location": "Sydney"
>     }
> }```
> 
> Both Name and Location are writable fields.

PR #1635, opened 2015-03-30
Use options metadata in create_post for Stripe

kitt:
Stripe has asked for specific, currently non-standard, metadata to be included on purchase posts.

Allow said metadata to be passed in via an options field.

@j-mutter

### DIFF
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -439,6 +439,17 @@ class StripeTest < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
+  def test_client_data_submitted_with_metadata_in_options
+    stub_comms(@gateway, :ssl_request) do
+      updated_options = @options.merge({:metadata => {:option1 => 'optional', :option2 => 'not required'}, :order_id => "42", :email => "foo@wonderfullyfakedomain.com"})
+      @gateway.purchase(@amount,@credit_card,updated_options)
+    end.check_request do |method, endpoint, data, headers|
+      assert_match(/metadata\[option1\]=optional/, data)
+      assert_match(/metadata\[option2\]=not\+required/, data)
+      assert_match(/metadata\[email\]=foo\%40wonderfullyfakedomain\.com/, data)
+      assert_match(/metadata\[order_id\]=42/, data)
+    end.respond_with(successful_purchase_response)
+  end
 
   def test_add_address
     post = {:card => {}}
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -441,11 +441,11 @@ class StripeTest < Test::Unit::TestCase
 
   def test_client_data_submitted_with_metadata_in_options
     stub_comms(@gateway, :ssl_request) do
-      updated_options = @options.merge({:metadata => {:option1 => 'optional', :option2 => 'not required'}, :order_id => "42", :email => "foo@wonderfullyfakedomain.com"})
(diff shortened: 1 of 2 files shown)


test/unit/gateways/stripe_test.rb
```diff
@@ -439,6 +439,17 @@ def test_client_data_submitted_with_purchase_without_email_or_order
...
+  def test_client_data_submitted_with_metadata_in_options
+    stub_comms(@gateway, :ssl_request) do
+      updated_options = @options.merge({:metadata => {:option1 => 'optional', :option2 => 'not required'}, :order_id => "42", :email => "foo@wonderfullyfakedomain.com"})
+      @gateway.purchase(@amount,@credit_card,updated_options)
+    end.check_request do |method, endpoint, data, headers|
+      assert_match(/metadata\[option1\]=optional/, data)
+      assert_match(/metadata\[option2\]=not\+required/, data)
+      assert_match(/metadata\[email\]=foo\%40wonderfullyfakedomain\.com/, data)
+      assert_match(/metadata\[order_id\]=42/, data)
+    end.respond_with(successful_purchase_response)
```
j-mutter: Don't really need the `respond_with`

j-mutter:
:hatched_chick: 

[API docs for reference](https://stripe.com/docs/api#metadata)

test/unit/gateways/stripe_test.rb
```diff
@@ -439,6 +439,17 @@ def test_client_data_submitted_with_purchase_without_email_or_order
     end.respond_with(successful_purchase_response)
   end
 
+  def test_client_data_submitted_with_metadata_in_options
+    stub_comms(@gateway, :ssl_request) do
+      updated_options = @options.merge({:metadata => {:option1 => 'optional', :option2 => 'not required'}, :order_id => "42", :email => "foo@wonderfullyfakedomain.com"})
```
girasquid: Could you name these keys something that's super obviously a test key/not expected in any way?

girasquid:
One question, otherwise :+1:

test/unit/gateways/stripe_test.rb
kitt: Removing it results in:

```
..........F
==========
Failure: <false> is not true.
test_client_data_submitted_with_metadata_in_options(StripeTest)
```

j-mutter: Huh, ok. I think I've only ever used it when I've needed the response, so just assumed you didn't need it here.

test/unit/gateways/stripe_test.rb
kitt: yep.

test/unit/gateways/stripe_test.rb
```diff
@@ -439,6 +439,17 @@ def test_client_data_submitted_with_purchase_without_email_or_order
     end.respond_with(successful_purchase_response)
   end
 
+  def test_client_data_submitted_with_metadata_in_options
+    stub_comms(@gateway, :ssl_request) do
+      updated_options = @options.merge({:metadata => {:this_is_a_random_key_name => 'with a random value', :i_made_up_this_key_too => 'canyoutell'}, :order_id => "42", :email => "foo@wonderfullyfakedomain.com"})
+      @gateway.purchase(@amount,@credit_card,updated_options)
+    end.check_request do |method, endpoint, data, headers|
+      assert_match(/metadata\[this_is_a_random_key_name\]=with\+a\+random\+value/, data)
+      assert_match(/metadata\[i_made_up_this_key_too\]=canyoutell/, data)
```
girasquid: :heart_eyes:

PR #1636, opened 2015-03-30
Ignore files generated by rdoc

aschrab:
Add /doc/ to .gitignore file so that using rdoc to generate docs doesn't
cause git to list untracked files.


girasquid:
Thanks!

PR #1638, opened 2015-04-03
Add Komoju gateway

k2nr:
Komoju is a Japanese payment gateway which supports all major payment methods in Japan(JPY). Currently we support five payment methods: Domestic Credit Card, WebMoney, Bank transfer, Konbini (Convenience Store), and PayEasy payments.

Website: https://www.komoju.com
API docs: https://docs.komoju.com


ntalbott:
@k2nr just got this merged - thanks! FYI, I removed Konbini and Bank Transfer, as they're out of scope for ActiveMerchant today.

PR #1641, opened 2015-04-07
paypal error fixed #1640

### LINKED ISSUE #1640: IndexError while parsing a cc payment using paypal
https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb#L435

This error is raised because xml tag value is 'SOAP_ENV:Envelope' and not 'Envelope'.
Due to this, cc payment always fails.

gauravsaini23:
error is raised because xml tag value is 'SOAP_ENV:Envelope' and not 'Envelope'.
Due to this, cc payment always fails.


girasquid:
We have a fairly large number of credit card payments going through PayPal that are successful - do you have a test case that can reproduce this?

girasquid:
Going to close this out for now - feel free to reopen when this has a test. :)

PR #1645, opened 2015-04-09
[Braintree] Add errored avs codes

ivanfer:
## AVS Error Codes
Braintree has some AVS codes corresponding to [errors](https://developers.braintreepayments.com/ios+ruby/reference/general/processor-responses/avs-cvv-responses); these are E (AVS system error) and S(Issuing bank does not support AVS). Since these match to what's available in `avs_result` we can just return these if present. This should be returned before any other mappings, since `avs_street_address_response_code` and `avs_postal_code_response_code` according to [Braintree docs](https://developers.braintreepayments.com/ios+ruby/reference/response/transaction#avs_error_response_code).

@girasquid @j-mutter

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -335,7 +335,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def avs_code_from(transaction)
-        avs_mapping["street: #{transaction.avs_street_address_response_code}, zip: #{transaction.avs_postal_code_response_code}"]
+        transaction.avs_error_response_code ||
+          avs_mapping["street: #{transaction.avs_street_address_response_code}, zip: #{transaction.avs_postal_code_response_code}"]
       end
 
       def avs_mapping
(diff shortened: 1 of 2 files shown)


girasquid:
Can you add some tests for this? It looks like postal code `30000` will give you the `E` response and `30001` will give you `S` (https://developers.braintreepayments.com/ios+ruby/reference/general/testing#avs-responses)

j-mutter:
:+1: for tests

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -335,6 +335,7 @@ def response_options(result)
       end
 
       def avs_code_from(transaction)
+        transaction.avs_error_response_code ||
         avs_mapping["street: #{transaction.avs_street_address_response_code}, zip: #{transaction.avs_postal_code_response_code}"]
```
j-mutter: Can you indent this so it's more obvious that it's logically linked to the line above?

ivanfer:
Thanks for finding the zip codes that trigger that response - remote tests and indents added.

girasquid:
:+1:

j-mutter:
:hatched_chick:

Issue #1642, opened 2015-04-07
Can not make a 'live' transaction with authorize.net's test servers.


jcavalieri:
I'm integrating testing of authorizing and capturing using the authorize.net gateway and found what I believe to be a bug in the authorization.net gateway.

Essentially in order to test authorization and capture with authorize.net's test servers the transactions to those test servers need to be 'live' transactions.

I believe the bug exists here:
https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L367

Specifically, whether or not to use test servers and whether or not mark the transaction as a test transaction is using the same "test?" method.

```
url = (test? ? test_url : live_url)
```
and

```
Response.new(
            success_from(response),
            message_from(response, avs_result, cvv_result),
            response,
            authorization: authorization_from(response),
            test: test?,
```
Therefore truly testing authorization and capture on authorize.net's test severs is impossible because the authorization call returns a bogus transaction id which is then rejected when the capture call is made with that id.

I'd be glad to help with a PR but would like some guidance on how best to handle this.

-John

bacrossland:
:+1:

girasquid:
> the authorization call returns a bogus transaction id which is then rejected when the capture call is made with that id.

Can you share a test case that reproduces this? [This test](https://github.com/Shopify/active_merchant/blob/ab46446aebb18eaa44ed1d92e11b5bedf94b03d0/test/remote/gateways/remote_authorize_net_test.rb#L68-L75) does a capture after an authorization against Authorize.Net [with the gateway in test mode](https://github.com/Shopify/active_merchant/blob/ab46446aebb18eaa44ed1d92e11b5bedf94b03d0/test/test_helper.rb#L18), and seems to work. What does the bogus authorization look like? What error do you get?

jcavalieri:
Let me fork and look into it more.

girasquid:
Ok - going to close this for now. Re-open it when you've got a test case for me. :)

jcavalieri:
The remote test does indeed work.

I figured what is going on. It is the existence of option[:test] key when initializing the gateway. When it is set to false the gateway uses neither the test servers nor test requests. But setting option[:test] to true uses both the test servers and test requests.

What works is setting the base mode to test and not even having a option[:test] key when initializing the gateway. This doesn't seem intuitive to me but I don't know if this counts as a an issue anymore.

girasquid:
[...]

PR #1648, opened 2015-04-15
Fixes Issue #1642, test_request for auth net is now a separate option and can be sent to live servers

jcavalieri:


girasquid:
@duff As I recall it was you that added `using_live_gateway_in_test_mode?` - how do you feel about this change?

test/remote/gateways/remote_authorize_net_test.rb
```diff
@@ -74,6 +74,48 @@ def test_successful_authorize_and_capture
...
+
+  # you can send test requests to live servers
+  def test_setting_test_request_does_not_set_test_mode
+    original_base_mode = ActiveMerchant::Billing::Base.mode
+    ActiveMerchant::Billing::Base.mode = :production
+
+    gateway = AuthorizeNetGateway.new(fixtures(:authorize_net).merge({:test_request => true}))
+
+    gateway.expects(:ssl_post).with(gateway.live_url, any_parameters, any_parameters)
+    gateway.purchase(@amount, @credit_card, @options)
```
girasquid: Where is the assertion that this didn't set test mode?

jcavalieri: It is that it is using the live url instead of the test url.

girasquid: Ah, ok - this test name looks a little unclear. Would it be better as `setting_test_request_goes_to_live_url` ?

jcavalieri: Good recommendation, change made.

girasquid:
This looks ok to me - I'd like a comment from @duff to make sure he's cool before I merge it.

duff:
Thanks for checking @girasquid!

I'm not sure if I'm fully getting the change.  I attempted to explain the reasoning here in this commit like you mentioned:

https://github.com/Shopify/active_merchant/commit/6977a6e96ebd9b0a338ed7f13503705f169831ea

We basically had a customer who went for a long time connecting to a live_url without realizing that the account was actually in test mode.  I think that test for `using_live_gateway_in_test_mode?` is important, no?  It looks like that test is no longer there.

I don't think I fully understand the intent behind the change.  Might you help me understand more?

I think the overall goal is that a test transaction should never be able to be run against the live_url.  And

```
response[:test_request] == "1"
```

is the only way I know of to check it.

jcavalieri:
The test "using_live_gateway_in_test_mode" was the problem.

With Authorize.net You can send test requests to live servers. So "test mode" was overloaded and not sufficient and needed to be split out into the following:
* "Test mode" means that you are using test servers.
* "Test Request" means you send test requests whether it is live or test servers.

duff:
I think that `using_live_gateway_in_test_mode?` is an important test for most folks.  In speaking to @ntalbott here, he suggested an option that could be passed to `initialize` to allow you to turn off the `using_live_gateway_in_test_mode?` check if you really need it turned off.  But the default should be such that running test transactions against a live gateway should fail with the appropriate error message.

Does that help explain?

ntalbott:
Yup, strong opinion on our side around the default being to error on test transactions to a live server, which comes out of our experience with a customer of a customer who ran test transactions without realizing it for almost a year.

Totally cool with a configuration option at gateway creation that overrides the check, though.

cipater:
*bump* 

I've been banging my head against the wall trying to figure out why Authorize.Net was telling me 'Using a live Authorize.net account in Test Mode is not permitted.' when that seems to be entirely the point of putting the account in Test Mode. 

To my dismay, I've discovered that ActiveMerchant is the one generating this error, thus preventing me from testing a live account that's set to Test Mode. 

I fully understand the reasoning behind the default behavior. But with no alternative way around it, I'm left at a frustrating impasse. Any updates on if this PR will be merged?

jcavalieri:
yeah, @cipater, I don't know what the hold up is. There seems to be confusion between test mode and test environments. My fix definitely addresses this and could save many people troubleshooting time.

duff:
Hi @jcavalieri and @cipater, I'd recommend checking out @ntalbott's comment about it:

https://github.com/activemerchant/active_merchant/pull/1648#issuecomment-95944831

If there was a PR with a configuration to override the check, that could work.  This PR though changes the default behavior.  I'm going to close this one for now, knowing you can make another PR if you'd like with the config option.  Does that help explain?

jcavalieri:
I'm confused @duff , this doesn't mess with test mode, it only allows for test requests whether in test mode or live.

jcavalieri:
The over cautiousness is what caused the bug in the first place.

ntalbott:
@jcavalieri any change in this area **must** retain the default behavior of raising an exception if a test request is run against a production gateway. We consider that over cautiousness to be a feature, not a bug. Specifically, this removal:

https://github.com/activemerchant/active_merchant/pull/1648/files#diff-32b8ad7eb3755b05bb1a03b6e1f9a3cdL373

Will not be merged. But a PR that allows disabling that check would be accepted.

PR #1649, opened 2015-04-16
Quickpay API v10 support

dinesh:


ntalbott:
This is a... large change. Can you summarize the v10 differences?

At a minimum, all recurring support is deprecated in ActiveMerchant and will need to be stripped out of this.

dinesh:
ok, I guess I need to adjust more stuff.  You can reject this PR. thx

ta:
@ntalbott: We (QuickPay) have launched a new platform (v10) with a completely new API - the old (v7) and new platforms will co-exist for some (longer) period of time. This is why the change is substantial.

ntalbott:
@ta can you summarize the changes between v7 and v10?

ta:
@ntalbott: of course :)

What kind of changes would you like me to describe? The change should be 100% transparent for the AM user (in a new PR that we'll create soon).

As I mentioned v10 is a completely new platform and API, and for some time we need to support both platforms (v4-v7 and v10) at the same time.

Please let me know if you need more information

ntalbott:
@ta do you have a summary you're giving to your customers on what's changed with the new API, why they should upgrade, etc.? Would be really helpful to see something like that to help us prioritize.

PR #1650, opened 2015-04-17
Stripe: Add support for idempotency keys

michaelherold:
Stripe [has the ability to use an idempotency key][docs] to ensure that
a request is idempotent. This is very useful for things like retrying
a charge when Stripe returns an intermittent failure, as the request
will only be processed _once_ on Stripe's end even if the first time the
request was attempted it was actually processed.

[docs]: https://stripe.com/docs/api#idempotent_requests


girasquid:
Thanks!

girasquid:
Thanks!

PR #1651, opened 2015-04-18
Only define constant once

mhenrixon:
Gets rid of warnings about previous definitions was here.

> /Users/mhenrixon/.rvm/gems/ruby-2.2.2/gems/activemerchant-1.45.0/lib/active_merchant/country.rb:67: warning: previous definition of COUNTRIES was here


girasquid:
This has been fixed in a new version of ActiveMerchant/Utils; please upgrade.

PR #1653, opened 2015-04-20
Add verify to moneris us

mrezentes:


test/unit/gateways/moneris_us_test.rb
```diff
@@ -22,6 +22,21 @@ def test_default_options
     assert_equal "qatoken", @gateway.options[:password]
   end
 
+  def test_successful_verify
+    @gateway.expects(:ssl_post).twice.returns(successful_purchase_response)
```
duff: I would have expected to see successful_authorize_response here since under the covers, it's doing an authorize rather than a purchase.  Perhaps something along the lines of:

https://github.com/Shopify/active_merchant/blob/master/test/unit/gateways/bridge_pay_test.rb#L105-L127

In Bridgepay's case, it's actually doing a void so in Moneris US, perhaps it could stub out `successful_authorize` followed by a successful capture of $0, since that seems to be the bizarro way to void an auth for Moneris US.

mrezentes: I discussed with @ntalbott.  I had originally made an auth response, but it was identical to the purchase response so it seemed unnecessary.

duff: Even though there'd be duplication, I'd still go for having a `successful_authorization_response` if for nothing else than to indicate intent.  When I look at the `test_successful_verify` method, my first thought is that there must something amiss since it's stubbing out the purchase call twice rather than stubbing out the `authorize` and the `capture` for $0.  

I'm also in the camp that there's value in having a unit test for `test_successful_verify_with_failed_void` to ensure that `ignore_result` is hooked up appropriately.

I'll mess with it here for a few minutes and see what I come up with.

duff:
Closed by 385ec491625faf4bcbe67a8b5c8e730b0ce85c3d

PR #1654, opened 2015-04-20
added exception Errno::EHOSTUNREACH to retry

randito78:
We noticed we had this exception raised and it was not caught by active_merchant


joevandyk:
+1

girasquid:
Thanks!

PR #1656, opened 2015-04-21
Use constants in `Connection` to replace magic numbers

larrylv:


ntalbott:
@larrylv this seems like a good idea, but so far as I can tell is redundant, since those constants are already used as the defaults for those values. Would just not setting those values here be better?

larrylv:
@ntalbott Since `PostsData` module will override `ActiveModule::Connection`'s default `open_timeout` and `read_timeout` here https://github.com/Shopify/active_merchant/blob/df777e116704704a5eda7f6de76eebba637f3ae2/lib/active_merchant/posts_data.rb#L47-L48, I think the `class_attribute` setting in `PostsData` is not redundant.

I mean, every time we try to issue a `raw_ssl_request`, we will create a new connection and then override the timeout variables with class_attribute value, so without the defaults they will be nil.

What do you think?

larrylv:
@ntalbott Sorry to ping you, but do you think my reply reasonable?

PR #1659, opened 2015-04-22
[Stripe] Add EMV "chip & sign", "chip & offline PIN" and Maestro support

bizla:
We've worked with Stripe to establish an EMV "Chip & offline PIN", and "chip & sign" API. This augments the Stripe integration to pass a chip card reader's TLV-BER strings through to Stripe. This also adds Maestro as a supported card type. 

A couple small things were added/changed to "core" ActiveMerchant too: 
- `CreditCard` now includes `icc_data` and `fallback_reason` fields, i.e. we've added an EMV chip to the credit card we're modelling. 
- VISA cards can have 19 digit PANs in the UK, so the regex has been modified.

Please review @j-mutter @girasquid @ntalbott


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -90,10 +90,16 @@ def purchase(money, payment, options = {})
...
       def capture(money, authorization, options = {})
         post = {}
+
         add_amount(post, money, options)
         add_application_fee(post, options)
 
-        commit(:post, "charges/#{CGI.escape(authorization)}/capture", post, options)
+        if emv_tc_response = options.delete(:icc_data)
+          post[:card] = {emv_approval_data: emv_tc_response}
+          commit(:post, "charges/#{CGI.escape(authorization)}", post, options)
```
girasquid: This line is duplicated with the `else` below - maybe add `:card` to `post` and then have the `commit` outside?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -283,9 +288,12 @@ def add_address(post, options)
 
       def add_creditcard(post, creditcard, options)
         card = {}
-        if creditcard.respond_to?(:number)
+        if creditcard.respond_to?(:icc_data) && creditcard.icc_data.present?
+          add_emv_creditcard(post, creditcard.icc_data)
+        elsif creditcard.respond_to?(:number)
           if creditcard.respond_to?(:track_data) && creditcard.track_data.present?
             card[:swipe_data] = creditcard.track_data
+            post[:card] = {:fallback_reason => creditcard.fallback_reason} if creditcard.fallback_reason
```
girasquid: Can you use a 1.9 style hash here?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -314,6 +321,12 @@ def add_creditcard(post, creditcard, options)
         end
       end
 
+      def add_emv_creditcard(post, icc_data, options = {})
+        card = {}
+        card[:emv_auth_data] = icc_data
+        post[:card] = card
```
girasquid: Why split this into three lines, instead of just 
```ruby
post[:card] = {emv_auth_data: icc_data}
```
?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -327,6 +340,13 @@ def add_flags(post, options)
         post[:recurring] = true if (options[:eci] == 'recurring' || options[:recurring])
       end
 
+      def add_metadata(post, options = {})
```
girasquid: :heart_eyes:

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -401,6 +421,7 @@ def commit(method, url, parameters = nil, options = {})
           :authorization => success ? response["id"] : response["error"]["charge"],
           :avs_result => { :code => avs_code },
           :cvv_result => cvc_code,
+          :emv_authorization => card["emv_auth_data"],
```
girasquid: Is it ok that you might be passing a `nil` `emv_authorization` here?

test/remote/gateways/remote_stripe_test.rb
```diff
@@ -43,6 +48,25 @@ def test_successful_purchase
     assert response.params["paid"]
     assert_equal "ActiveMerchant Test Purchase", response.params["description"]
     assert_equal "wow@example.com", response.params["metadata"]["email"]
+  end
+
+  # this actually shouldn't be legal
```
girasquid: I don't get it. It shouldn't be legal, but the test passes? Why?

test/remote/gateways/remote_stripe_test.rb
```diff
@@ -85,6 +109,30 @@ def test_authorization_and_capture
     assert_success capture
   end
 
+  def test_authorization_and_capture_with_emv_credit_card_in_uk
+    @gateway = StripeGateway.new(fixtures(:stripe_emv_uk))
+    assert authorization = @gateway.authorize(@amount, @emv_credit_cards[:uk], @options)
+    assert_success authorization
+    assert authorization.emv_authorization, "Authorization should contain emv_authorization containing the EMV ARPC"
+    assert !authorization.params["captured"]
```
girasquid: `refute`?

test/remote/gateways/remote_stripe_test.rb
```diff
@@ -85,6 +109,30 @@ def test_authorization_and_capture
...
+    assert_success capture
+    assert capture.emv_authorization, "Capture should contain emv_authorization containing the EMV TC"
+  end
+
+  def test_authorization_and_capture_with_emv_credit_card_in_us
+    @gateway = StripeGateway.new(fixtures(:stripe_emv_us))
+    assert authorization = @gateway.authorize(@amount, @emv_credit_cards[:us], @options)
+    assert_success authorization
+    assert authorization.emv_authorization, "Authorization should contain emv_authorization containing the EMV ARPC"
+    assert !authorization.params["captured"]
```
girasquid: `refute`

test/remote/gateways/remote_stripe_test.rb
```diff
@@ -105,6 +153,28 @@ def test_authorization_and_void
     assert_success void
   end
 
+  def test_authorization_and_void_with_emv_credit_card_in_us
+    @gateway = StripeGateway.new(fixtures(:stripe_emv_us))
+    assert authorization = @gateway.authorize(@amount, @emv_credit_cards[:us], @options)
+    assert_success authorization
+    assert authorization.emv_authorization, "Authorization should contain emv_authorization containing the EMV ARPC"
+    assert !authorization.params["captured"]
```
girasquid: `refute`

test/remote/gateways/remote_stripe_test.rb
```diff
@@ -105,6 +153,28 @@ def test_authorization_and_void
...
+    assert void = @gateway.void(authorization.authorization)
+    assert_success void
+  end
+
+  def test_authorization_and_void_with_emv_credit_card_in_uk
+    @gateway = StripeGateway.new(fixtures(:stripe_emv_uk))
+    assert authorization = @gateway.authorize(@amount, @emv_credit_cards[:uk], @options)
+    assert_success authorization
+    assert authorization.emv_authorization, "Authorization should contain emv_authorization containing the EMV ARPC"
+    assert !authorization.params["captured"]
```
girasquid: `refute`

test/test_helper.rb
```diff
@@ -180,6 +180,16 @@ def network_tokenization_credit_card(number = '4242424242424242', options = {})
       Billing::NetworkTokenizationCreditCard.new(defaults)
     end
 
+    # this mock is only useful with unit tests, as cryptograms generated by an EMV terminal
```
girasquid: :heart_eyes:

test/test_helper.rb
```diff
@@ -180,6 +180,16 @@ def network_tokenization_credit_card(number = '4242424242424242', options = {})
...
+    # are specific to the target acquirer, so remote tests using this mock will fail.
+    def credit_card_with_icc_data(options = {})
+      defaults = {
```
girasquid: It kind of seems like this method:
1) Doesn't do anything with `options`
2) Has a var called `defaults` that is actually options and never changes.

Do you need options to be passable? Or can you remove the argument and rename `defaults`?

girasquid: Also, with this only being used for stripe unit tests right now - do you want to put it in the stripe test? Or is the hope that it's useful for other EMV implementation unit tests also?

lib/active_merchant/billing/gateways/stripe.rb
j-mutter: :+1:

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -226,23 +232,22 @@ def type
...
-        add_customer(post, payment, options)
-        add_customer_data(post, options)
-        post[:description] = options[:description]
-        post[:statement_description] = options[:statement_description]
-
-        post[:metadata] = options[:metadata] || {}
-        post[:metadata][:email] = options[:email] if options[:email]
-        post[:metadata][:order_id] = options[:order_id] if options[:order_id]
-        post.delete(:metadata) if post[:metadata].empty?
+        unless payment.respond_to?(:icc_data) && payment.icc_data.present?
```
j-mutter: You've done this check a few times, worth making it a method?

lib/active_merchant/billing/gateways/stripe.rb
j-mutter: Agreed, time for my macro:
New hash syntax? :innocent:

test/remote/gateways/remote_stripe_test.rb
j-mutter: Because we're spoofing ICC data or something? ¯\(°_o)/¯

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -226,23 +232,22 @@ def type
...
+          add_amount(post, money, options, true)
```
j-mutter: You don't need the amount?

j-mutter: Is it contained in the payload?

test/remote/gateways/remote_stripe_test.rb
```diff
@@ -43,6 +48,25 @@ def test_successful_purchase
...
+  def test_successful_purchase_with_emv_credit_card_in_uk
+    @gateway = StripeGateway.new(fixtures(:stripe_emv_uk))
+    assert response = @gateway.purchase(@amount, @emv_credit_cards[:uk], @options)
+    assert_success response
+    assert_equal "charge", response.params["object"]
+    assert response.params["paid"]
+    assert_match CHARGE_ID_REGEX, response.authorization
+  end
+
+  # this shouldn't work technically
```
j-mutter: Why?

lib/active_merchant/billing/gateways/stripe.rb
bizla: Subtle difference, we're supposed to post to `charges/:authorization` in this case and to `charges/:authorization/capture` in the `else`.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -90,10 +90,16 @@ def purchase(money, payment, options = {})
...
+
         add_amount(post, money, options)
         add_application_fee(post, options)
 
-        commit(:post, "charges/#{CGI.escape(authorization)}/capture", post, options)
+        if emv_tc_response = options.delete(:icc_data)
+          post[:card] = {emv_approval_data: emv_tc_response}
+          commit(:post, "charges/#{CGI.escape(authorization)}", post, options)
+        else
+          commit(:post, "charges/#{CGI.escape(authorization)}/capture", post, options)
```
bizla: Note the `/capture` here that isn't in the initial `if`'s `commit` call.

lib/active_merchant/billing/gateways/stripe.rb
bizla: Oh yeah, probably. I'll add it.

lib/active_merchant/billing/gateways/stripe.rb
bizla: Yeah, everything is in the passed ICC data.

lib/active_merchant/billing/gateways/stripe.rb
bizla: Yup. Changing it up.

lib/active_merchant/billing/gateways/stripe.rb
girasquid: Oh, yup - I see now. Thanks!

lib/active_merchant/billing/gateways/stripe.rb
bizla: A lot was happening in this method before, so this is the residue I guess. ;) I'll change it.

lib/active_merchant/billing/gateways/stripe.rb
bizla: Yes, it'd make the EMV chip unhappy and then we'd know to stop the transaction.

test/remote/gateways/remote_stripe_test.rb
bizla: Because Stripe technically has everything they need to capture funds, but it's ill-advised to do in actual practice, since we need a message back to the chip to actually finalize the transaction on the card level. I put the comment there so we know to never use a `purchase` in the EMV case.

lib/active_merchant/billing/gateways/stripe.rb
girasquid: :cool:

test/remote/gateways/remote_stripe_test.rb
bizla: I'll clarify the comments.

test/test_helper.rb
bizla: Yeah, this should probably be isolated to the Stripe tests. I'll move it.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -426,6 +445,10 @@ def json_error(raw_response)
       def non_fractional_currency?(currency)
         CURRENCIES_WITHOUT_FRACTIONS.include?(currency.to_s)
       end
+
+      def emv_payment?(payment)
```
j-mutter: :heart_eyes:

j-mutter: Though I feel like this belongs on the `CreditCard` model

bizla: Though about that, but then I'd have to still ask if `payment.respond_to?(:emv_payment?)` since `payment` isn't guaranteed to be a `CreditCard`.

girasquid: :+1: to putting it on `CreditCard`

girasquid: Maybe a method on `CreditCard` that can tell you if it's an EMV card, and then this method to do the checking?

j-mutter: Still seems like the Right Thing To Do™

bizla: So you'd prefer we go from: 
```ruby
class StripeGateway
...
def emv_payment?(payment)
  payment.respond_to?(:icc_data) && payment.icc_data.present?
end
```
to
```ruby
class StripeGateway
...
def emv_payment?(payment)
  payment.respond_to(:emv?) && payment.emv?
end
```
where
```ruby
class CreditCard
...
def emv? # also needs tests
  icc_data.present? 
end
```
? 
Doesn't seem better, but I might just have cabin fever. :disappointed_relieved: :

girasquid: Yeah - that way as EMV support gets added to other gateways (that only support credit cards getting passed in), they can just ask the credit card if it needs to be handled through EMV flows.

bizla: :ok: Done!

test/unit/credit_card_test.rb
```diff
@@ -432,4 +432,12 @@ def test_month_and_year_are_immediately_converted_to_integers
     card.start_year = "1"
     assert_equal 1, card.start_year
   end
+
+  def test_should_report_as_emv_if_icc_data_present
```
girasquid: :heart_eyes:

girasquid:
Could use a squash - :+1: from me.

bizla:
38 -> 9 commits. Merging on :green_apple: if there are no objections.

PR #1663, opened 2015-04-24
Add Omise Payment Gateway Support [TH]

zdk:
This pull request adds support for Omise payment gateway from Thailand. (https://www.omise.co)
We support Visa and MasterCard, and most debit cards from Thailand.
We are fully PCI-DSS 3.0 compliant. Our certificate can be found here: https://www.omise.co/blog/2015-02-18-how-secure-is-omise.html 

Our restful json API supports authorize, capture, partial or full refund, and saving the card as customer profile for later charging.
We use the PATCH method in the API to update the card token, converting it to a customer profile. We've added it on the pull request, hope that's ok. 

Discussion on this pull request are welcome!
API Reference: https://docs.omise.co 
Thanks,
Omise Team

### DIFF
--- a/lib/active_merchant/billing/gateways/omise.rb
+++ b/lib/active_merchant/billing/gateways/omise.rb
@@ -1,0 +1,319 @@
+require 'active_merchant/billing/rails'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class OmiseGateway < Gateway
+      API_VERSION = '1.0'
+      API_URL     = 'https://api.omise.co/'
+      VAULT_URL   = 'https://vault.omise.co/'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        'invalid_security_code' => STANDARD_ERROR_CODE[:invalid_cvc],
+        'failed_capture'        => STANDARD_ERROR_CODE[:card_declined]
+      }
+
+      self.live_url = self.test_url = API_URL
+
+      # Currency supported by Omise
+      # * Thai Baht with Satang, ie. 9000 => 90 THB
+      self.default_currency = 'THB'
+      self.money_format     = :cents
+
+      #Country supported by Omise
+      # * Thailand
+      self.supported_countries = %w( TH )
+
+      # Credit cards supported by Omise
+      # * VISA
+      # * MasterCard
+      self.supported_cardtypes = [:visa, :master]
+
+      # Omise main page
+      self.homepage_url = 'https://www.omise.co/'
+      self.display_name = 'Omise'
+
+      # Creates a new OmiseGateway.
+      #
+      # Omise requires public_key for token creation.
+      # And it requires secret_key for other transactions.
+      # These keys can be found in https://dashboard.omise.co/test/api-keys
+      #
+      # ==== Options
+      #
+      # * <tt>:public_key</tt> -- Omise's public key (REQUIRED).
+      # * <tt>:secret_key</tt> -- Omise's secret key (REQUIRED).
+
+      def initialize(options={})
+        requires!(options, :public_key, :secret_key)
+        @public_key = options[:public_key]
+        @secret_key = options[:secret_key]
+        super
+      end
+
+      # Perform a purchase (with auto capture)
+      #
+      # ==== Parameters
+      #
+      # * <tt>money</tt>          -- The purchasing amount in Thai Baht Satang
+      # * <tt>payment_method</tt> -- The CreditCard object
+      # * <tt>options</tt>        -- An optional parameters, such as token from Omise.js
+      #
+      # ==== Options
+      # * <tt>token_id</tt> -- token id, use Omise.js library to retrieve a token id
+      # if this is passed as an option, it will ignore tokenizing via Omisevaultgateway object
+      #
+      # === Example
+      #  To create a charge on a card
+      #
+      #   purchase(money, Creditcard_object)
+      #
+      #  To create a charge on a token
+      #
+      #   purchase(money, nil, { :token_id => token_id, ... })
+      #
+      #  To create a charge on a customer
+      #
+      #   purchase(money, nil, { :customer_id => customer_id })
+
+      def purchase(money, payment_method, options={})
+        create_charge(money, payment_method, options)
+      end
+
+      # Authorize a charge.
+      #
+      # ==== Parameters
+      #
+      # * <tt>money</tt>          -- The purchasing amount in Thai Baht Satang
+      # * <tt>payment_method</tt> -- The CreditCard object
+      # * <tt>options</tt>        -- An optional parameters, such as token or capture
+
+      def authorize(money, payment_method, options={})
+        options[:capture] = 'false'
+        create_charge(money, payment_method, options)
+      end
+
+      # Capture an authorized charge.
+      #
+      # ==== Parameters
+      #
+      # * <tt>money</tt>     -- An amount in Thai Baht Satang
+      # * <tt>charge_id</tt> -- The CreditCard object
+      # * <tt>options</tt>   -- An optional parameters, such as token or capture
+
+      def capture(money, charge_id, options={})
+        post = {}
+        add_amount(post, money, options)
+        commit(:post, "charges/#{CGI.escape(charge_id)}/capture", post, options)
+      end
+
+      # Refund a charge.
+      #
+      # ==== Parameters
+      #
+      # * <tt>money</tt>     -- An amount of money to charge in Satang.
+      # * <tt>charge_id</tt> -- The CreditCard object
+      # * <tt>options</tt>   -- An optional parameters, such as token or capture
+
+      def refund(money, charge_id, options={})
+        options[:amount] = money if money
+        commit(:post, "charges/#{CGI.escape(charge_id)}/refunds", options)
+      end
+
+      # Store a card details as customer
+      #
+      # ==== Parameters
+      #
+      # * <tt>payment_method</tt> -- The CreditCard.
+      # * <tt>options</tt>        -- Optional Customer information:
+      #     'email'       (A customer email)
+      #     'description' (A customer description)
+
+      def store(payment_method, options={})
+        post, card_params = {}, {}
+        add_customer_data(post, options)
+        add_token(card_params, payment_method, options)
+        commit(:post, 'customers', post.merge(card_params), options)
+      end
+
+      # Delete a customer and all associated credit cards.
+      #
+      # ==== Parameters
+      #
+      # * <tt>customer_id</tt> -- The Customer identifier (REQUIRED).
+
+      def unstore(customer_id, options={})
+        commit(:delete, "customers/#{CGI.escape(customer_id)}")
+      end
+
+      # Enable scrubbling sensitive information
+      def supports_scrubbing?
+        true
+      end
+
+      # Scrub sensitive information out of HTTP transcripts
+      #
+      # ==== Parameters
+      #
+      # * <tt>transcript</tt> -- The HTTP transcripts
+
+      def scrub(transcript)
+        transcript.
+          gsub(/(Authorization: Basic )\w+/i, '\1[FILTERED]').
+          gsub(/(\\"number\\":)\\"\d+\\"/, '\1[FILTERED]').
+          gsub(/(\\"security_code\\":)\\"\d+\\"/,'\1[FILTERED]')
+      end
+
+      private
+
+      def create_charge(money, payment_method, options)
+        post = {}
+        add_token(post, payment_method, options)
+        add_amount(post, money, options)
+        add_customer(post, options)
+        post[:capture] = options[:capture] if options[:capture]
+        commit(:post, 'charges', post, options)
+      end
+
+      def headers(options={})
+        key = options[:key] || @secret_key
+        {
+          'Content-Type'    => 'application/json;utf-8',
+          'User-Agent'      => "Omise/v#{API_VERSION} ActiveMerchantBindings/#{ActiveMerchant::VERSION}",
+          'Authorization'   => 'Basic ' + Base64.encode64(key.to_s + ':').strip,
+          'Accept-Encoding' => 'utf-8'
+        }
+      end
+
+      def url_for(endpoint)
+        (endpoint == 'tokens' ? VAULT_URL : API_URL) + endpoint
+      end
+
+      def post_data(parameters)
+        parameters.present? ? parameters.to_json : nil
+      end
+
+      def https_request(method, endpoint, parameters=nil, options={})
+        raw_response = response = nil
+        begin
+          raw_response = ssl_request(method, url_for(endpoint), post_data(parameters), headers(options))
+          response = parse(raw_response)
+        rescue ResponseError => e
+          raw_response = e.response.body
+          response = parse(raw_response)
+        rescue JSON::ParserError
+          response = json_error(raw_response)
+        end
+        response
+      end
+
+      def parse(body)
+        JSON.parse(body)
+      end
+
+      def json_error(raw_response)
+        msg  = "Invalid response received from Omise API. Please contact support@omise.co if you continue to receive this message."
+        msg += "The raw response returned by the API was #{raw_response.inspect})"
+        { message: msg }
+      end
+
+      def commit(method, endpoint, params=nil, options={})
+        response = https_request(method, endpoint, params, options)
+        Response.new(
+          successful?(response),
+          message_from(response),
+          response,
+          {
+            authorization: authorization_from(response),
+            test: test?,
+            error_code: successful?(response) ? nil : standard_error_code_mapping(response)
+          }
+        )
+      end
+
+      def standard_error_code_mapping(response)
+        STANDARD_ERROR_CODE_MAPPING[error_code_from(response)] || message_to_standard_error_code_from(response)
+      end
+
+      def error_code_from(response)
+        error?(response) ? response['code'] : response['failure_code']
+      end
+
+      def message_to_standard_error_code_from(response)
+        message = response['message'] if response['code'] == 'invalid_card'
+        case message
+          when /brand not supported/
+            STANDARD_ERROR_CODE[:invalid_number]
+          when /number is invalid/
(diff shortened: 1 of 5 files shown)


zdk:
cc @fred

lib/active_merchant/billing/gateways/omise/omise_vault.rb
```diff
@@ -0,0 +1,37 @@
+require 'active_merchant/billing/gateways/omise/omise_core'
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class OmiseVaultGateway < Gateway
```
ntalbott: No need for another gateway here; this should be implemented on the main gateway as `store`.

lib/active_merchant/billing/gateways/omise/omise_core.rb
```diff
@@ -0,0 +1,142 @@
+require 'json'
+require 'active_merchant/billing/gateway'
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    module OmiseCore
```
ntalbott: Please inline this in the `omise.rb` file.

test/remote/gateways/remote_omise_test.rb
```diff
@@ -0,0 +1,121 @@
...
+    token = @gateway.send(:omise_tokenize, @credit_card)
+    assert token.is_a?(OmisePaymentToken)
+    @options[:token_id] = token.payment_data.params['token']['id']
+    response = @gateway.store(nil, @options)
+    assert_success response
+    customer = @gateway.unstore(response.params['id'])
+    assert_equal true, customer.params['deleted']
+  end
+
+  def test_authorize_and_capture
```
ntalbott: I don't see a `capture` call? Is that tested elsewhere?

ntalbott:
@zdk I'm excited to see a solid gateway in Thailand! The adapter needs some work - it's over-factored at this point - but hit me up with any questions about my above comments. I'm psyched to help you get this merged.

fred:
Thanks for the comments @ntalbott 

Regarding ```capture``` , in this case it's auto capture unless specified otherwise. Should we add a test to only authorize and not capture?

ok, we will under-factor it a  bit :)

ntalbott:
@fred so the ActiveMerchant expectation is manual capture - can you set it up that way?

zdk:
@ntalbott Thanks, we'll change regarding the review.

zdk:
@ntalbott anything you would like to add more?
I'll check for your feedback tomorrow. 
( Bangkok time :smile: )
Thanks!

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,371 @@
...
+      # Refund a charge.
+      #
+      # ==== Parameters
+      #
+      # * <tt>money</tt>     -- An amount of money to charge in Satang.
+      # * <tt>charge_id</tt> -- The Charge object identifier.
+      # * <tt>options</tt>   -- A standard hash options.
+
+      def refund(money, charge_id, options={})
+        options[:amount] = money if money
```
girasquid: It looks like `money` is a required parameter here - why the `if money`?

test/unit/gateways/omise_test.rb
```diff
@@ -0,0 +1,646 @@
...
+    }
+
+    @card_token = {
+      object: 'token',
+      id: 'tokn_test_4zgf1crg50rdb68xlk5'
+    };
+  end
+
+  def test_supported_countries
+    assert @gateway.supported_countries == %w( TH )
```
girasquid: You can probably use `assert_equal` here.

test/unit/gateways/omise_test.rb
```diff
@@ -0,0 +1,646 @@
...
+  end
+
+  def test_supports_scrubbing
+    assert @gateway.supports_scrubbing? == true
```
girasquid: `assert` is checking that `@gateway.supports_scrubbing?` returns `true` here - you don't need the `== true`.

test/unit/gateways/omise_test.rb
```diff
@@ -0,0 +1,646 @@
...
+  end
+
+  def test_scrub
+    assert @gateway.scrub(
```
girasquid: This test doesn't do any checks for card number or CVC scrubbing - can you add them?

girasquid: If you'd like to stick with small responses like this - you could loop over the short strings and match them individually.

lib/active_merchant/billing/gateways/omise.rb
ntalbott: @girasquid well, with many adapters you can pass in `nil` as the amount and it'll just refund the whole transaction. That said, unless having a blank `amount` blows up transactions, you still shouldn't need the check.

girasquid: Oh! Thanks for the explanation, I had no idea.

girasquid:
I have some concerns about the scrubbing tests, but other than that this looks great.

lib/active_merchant/billing/gateways/omise.rb
zdk: @ntalbott  Thanks for the clarification! :+1:

zdk:
@girasquid Thanks for your feedback! 
I just making the change accordingly your comments. Hope this looks sane. :octocat:

girasquid:
Looks good!

fred:
Thanks guys for the helpful comments, we are very excited to contribute and have our awesome gateway supported in ActiveMerchant. 
We have many Shopify customers in Thailand requesting our gateway support. The lack of a proper gateway here has put back many merchants from using Shopify.
We will send a request to Shopify to have our gateway supported, so, is there any changes needed that we can contribute before sending a request?
Thanks for the great support.
Cheers.

girasquid:
I'm good to merge this once @ntalbott confirms he's happy with the changes. For integration into Shopify - send an email to payment-integrations@shopify.com.

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,376 @@
...
+        commit(:post, "charges/#{CGI.escape(charge_id)}/refunds", options)
+      end
+
+      # Backward compatible method for refunding a charge.
```
ntalbott: Drop this; no one should be coding against deprecated API's on new adapters.

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,376 @@
...
+      end
+
+      # Void a charge.
+      #
+      # ==== Parameters
+      #
+      # * <tt>charge_id</tt> -- The Charge object identifier.
+      # * <tt>options</tt>   -- A standard hash options.
+
+      def void(charge_id, options={})
```
ntalbott: The main point of `void` in ActiveMerchant is to cancel an authorization, will that work as implemented here?

I'm not a big fan of having to `MultiResponse` here - no way to do it in one call?

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,376 @@
...
+      # Create a Customer and associated Card.
+      #
+      # ==== Parameters
+      #
+      # * <tt>payment</tt> -- The CreditCard.
+      # * <tt>options</tt> -- A standard hash options.
+      # * use options set_default_card: true to set the default card
+
+      def store(payment, options={})
+        return Response.new(false, "#{payment.errors.full_messages.join('. ')}") if payment.is_a?(CreditCard) and !payment.valid?
```
ntalbott: Don't do/check validation here - that should be handled by the calling code, and/or be seen as an error coming back from the API.

test/remote/gateways/remote_omise_test.rb
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  def test_failed_purchase
+    response = @gateway.purchase(@amount, @invalid_cvc)
+    assert_failure response
+    assert_equal Gateway::STANDARD_ERROR_CODE[:invalid_cvc], response.error_code
+  end
+
+  def test_successful_purchase_with_token
+    token = @gateway.send(:omise_tokenize, @credit_card)
```
ntalbott: Don't test via hitting `private` methods. Instead, hit the public `store` method, and interact with it.

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,376 @@
...
+
+      def add_amount(post, money, options)
+        post[:amount]      = amount(money)
+        post[:currency]    = options[:currency] || currency(money)
+        post[:description] = options[:description] if options.key?(:description)
+      end
+
+    end #OmiseGateway
+
+    class OmisePaymentToken < PaymentToken
```
ntalbott: While there are plans to return a proper token object from `store`, we don't today, so I don't think this adds value. Just return the actual token string in the `Response#authorization`.

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,376 @@
...
+        token =
+          if successful?(response)
+            Response.new(true, nil, token: response)
+          else
+            Response.new(false, message_from(response), {}, error_code: STANDARD_ERROR_CODE_MAPPING[error_code_from(response)])
+          end
+        OmisePaymentToken.new(token)
+      end
+
+      def add_or_create_token(post, payment, options={})
```
ntalbott: Can you split this into two methods, one for adding to a request, and one for creating a token? I think that'll make the code clearer.

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,376 @@
...
+        return 'Success' if successful?(response)
+        (response['message'] ? response['message'] : response['failure_message'])
+      end
+
+      def authorization_from(response)
+        response['id'] if successful?(response)
+      end
+
+      def successful?(response)
+        ( response.key?('object') and response['object'] != 'error' ) and response['failure_code'].nil?
```
ntalbott: I see use of `and` and `or` a lot, which often doesn't do what you want. Instead use `&&` and `||` exclusively.

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,376 @@
...
+          :name             => creditcard.name,
+          :security_code    => creditcard.verification_value,
+          :expiration_month => creditcard.month,
+          :expiration_year  => creditcard.year
+        }
+        post[:card] = card
+      end
+
+      def add_customer(post, options={})
+        return if post[:card].nil? or post[:card].match(/tokn(_test)?_[1-9a-z]+/)
```
ntalbott: Is `tokn` a typo? Or that's really the match?

ntalbott:
Dropped a bunch more comments. Making progress.

lib/active_merchant/billing/gateways/omise.rb
fred: @ntalbott no, it's a 4 letter identifier part of our UID strings.

lib/active_merchant/billing/gateways/omise.rb
fred: yep, that's should be fixed.

test/remote/gateways/remote_omise_test.rb
fred: will do

lib/active_merchant/billing/gateways/omise.rb
fred: ok, let's remove this.

lib/active_merchant/billing/gateways/omise.rb
fred: @ntalbott currently the settlement time is very short in the bank here, so can't do a void most of the time. Thus we dont expose an external API for void.
For now it must be implemented this way. but we will improve soon when we have access to more banks.

test/remote/gateways/remote_omise_test.rb
fred: will fix this :+1:

lib/active_merchant/billing/gateways/omise.rb
ntalbott: @fred not sure what you mean, since an `authorize` shouldn't settle/capture until `capture` is explicitly called, right? The goal of `void` in ActiveMerchant is to "cancel" an `authorize` that hasn't yet been captured - can Omise do that?

robinclart: Void is used by some banks to refer to the cancelling of a capture before settlement. The bank we're using calls it "reverse" to cancel an authorization.

But anyhow we currently don't support either of them because of the added complexity this feature added to our charge object representation. Furthermore since the issuing bank will reverse the auth by themselves after some time we decided to keep go without those.

girasquid: It sounds like void is not supported - this can be removed, in that case.

ntalbott: As @girasquid says, just drop `void` if you don't support canceling auths.

FWIW at @spreedly we've seen that some merchants get complaints if auths aren't cleaned up when possible, since customers (especially those with debit cards) aren't clear on why there's a hold on $X on their card for a while.

lib/active_merchant/billing/gateways/omise.rb
zdk: The way we save the card is attaching to a customer like so:

```
curl https://vault.omise.co/tokens \
  -X POST \
  -u pkey_test_4zt0fss8gs0z6b4zlsq: \
  -d "card[name]=JOHN DOE" \
  -d "card[city]=Bangkok" \
  -d "card[postal_code]=10320" \
  -d "card[number]=4242424242424242" \
  -d "card[expiration_month]=4" \
  -d "card[expiration_year]=2017"
```
  
```
curl https://api.omise.co/customers \
  -X POST \
  -u skey_test_4zt0fss8eklsj88dx9l: \
  -d "description=John Doe (id: 30)" \
  -d "email=john.doe@example.com" \
  -d "card=token_id"
```

and return customer json representation (https://docs.omise.co/api/customers/)

similar to https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L173

Is this ok to support as a `store` method?

ntalbott: ActiveMerchant is intended as an abstraction across gateways; the `Response#authorization` returned from a `store` call should contain a token (String) that can be passed to `purchase` (and `authorize`) in place of a `CreditCard` for future charges. ActiveMerchant does not care or have any first-class representation of a customer record; it only deals in specific payment methods (credit cards, tokens) and transactions against those payment methods.

You can also return whatever you'd like in the `Response#params` Hash, so if there's a customer id or similar you want it should go in there for anyone using ActiveMerchant who doesn't mind gateway-specific entanglements.

ntalbott: Also, FWIW, I wouldn't use the Stripe adapter as an example of good ActiveMerchant practice. It's kind of a bloated mess these days. I'd look at gateways like NMI, eWay Rapid, and Balanced (even though it's going away, :cry:).

zdk: @ntalbott haha, great thanks for the explanation! :+1:  I'll fix for the omise :D

zdk:
@ntalbott here they are.
Thanks! :dancers:

fred:
Guys, we did some major cleanup. looking good?
@ntalbott @girasquid , do you guys need a squash? 
let us know!

cheers.

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -0,0 +1,298 @@
+require 'active_merchant/billing/rails'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class OmiseGateway < Gateway
+      API_VERSION = '1.0'
+      API_URL     = 'https://api.omise.co/'
+      VAULT_URL   = 'https://vault.omise.co/'
+
+      STANDARD_ERROR_CODE_MAPPING = {
```
girasquid: Are there any other standard errors that come back? I noticed that the only error being changed is this one - do the other errors all map to ActiveMerchant's standard error codes?

girasquid:
Question about error codes and could use a squash, but other than that this looks good to me.

lib/active_merchant/billing/gateways/omise.rb
zdk: atm, this is the only code I could map the code directly to ActiveMerchant's standard error codes.
unless it needed to parse a message.
```
{
  "object": "error",
  "location": "https://docs.omise.co/api/errors#invalid-card",
  "code": "invalid_card",
  "message": "number is invalid and brand not supported (unknown)"
}
```
or
```
{
  "object": "error",
  "location": "https://docs.omise.co/api/errors#invalid-card",
  "code": "invalid_card",
  "message": "expiration date cannot be in the past"
}
```
I'm uncertain about parsing a message, is it good idea to do so? or no need to parse?
@girasquid  what do you think?

girasquid: If your messages are consistent, I think parsing them is fine - it would be great to have more standard error codes coming out of this integration if we can.

zdk:
@girasquid  just pushed the change regarding error codes, so now there are more standard error codes. Thanks!

girasquid:
Awesome, thanks! After a squash I think this is good to go. @ntalbott final thoughts?

zdk:
you're welcome, here is a squash @girasquid.
@ntalbott how do you feel now? :smile:

zdk:
@ntalbott Have you had any chance to review this one?  Is it in ActiveMerchant standard/expectation now or need some more change? We're looking forward to have this merged.
Thanks!

fred:
@girasquid @ntalbott do you guys have any updates for this PR? we're eager to move forward.
Cheers.

girasquid:
Merged - thanks!

fred:
@girasquid Thank you!

robinclart:
:+1:

jun-omise:
Awesome! :+1:

PR #1665, opened 2015-04-24
Move device fingerptint to root (conekta)

MauricioMurga:
Conekta needs device fingerprint in the root of the request body.


girasquid:
Thanks!

PR #1669, opened 2015-04-29
Add verification value to request only if it's valid

tjstankus:


tjstankus:
@duff please review :)

test/unit/gateways/authorize_net_test.rb
```diff
@@ -626,6 +626,18 @@ def test_truncation
...
   end
 
+  def test_invalid_cvv
+    invalid_cvvs = ['47', '12345', '']
+    invalid_cvvs.each do |cvv|
+      card = credit_card(@credit_card.number, { verification_value: cvv })
+      stub_comms do
+        @gateway.purchase(@amount, card)
+      end.check_request do |endpoint, data, headers|
+        parse(data) { |doc| assert_nil doc.at_xpath('//cardCode') }
```
duff: Small little nit - might just checking that the data doesn't include `cardCode` be sufficient?  Or is that included elsewhere in data?

duff: I see now that that's how the other tests are working too with `parse`.  Perhaps consistency is a good thing here.  Disregard my previous nit. :)

PR #1670, opened 2015-04-30
[Stripe] Bug fix: add amounts only on non-EMV transactions, temporarily omit EMV testcases

bizla:
This partially addresses https://github.com/Shopify/active_merchant/issues/1664 and adds a few small bugfixes: 
- Don't attach transaction amount if sending an EMV transaction. 
- Attach the EMV fallback reason to the correct node of the request. 

@j-mutter @girasquid /cc @duff @ntalbott

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -91,13 +91,13 @@ module ActiveMerchant #:nodoc:
       def capture(money, authorization, options = {})
         post = {}
 
-        add_amount(post, money, options)
         add_application_fee(post, options)
 
         if emv_tc_response = options.delete(:icc_data)
           post[:card] = { emv_approval_data: emv_tc_response }
           commit(:post, "charges/#{CGI.escape(authorization)}", post, options)
         else
+          add_amount(post, money, options)
           commit(:post, "charges/#{CGI.escape(authorization)}/capture", post, options)
         end
       end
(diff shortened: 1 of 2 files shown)


duff:
:+1:

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -91,13 +91,13 @@ def purchase(money, payment, options = {})
...
         post = {}
 
-        add_amount(post, money, options)
         add_application_fee(post, options)
 
         if emv_tc_response = options.delete(:icc_data)
           post[:card] = { emv_approval_data: emv_tc_response }
           commit(:post, "charges/#{CGI.escape(authorization)}", post, options)
         else
+          add_amount(post, money, options)
```
girasquid: This is because the `icc_data` will include the amount, right?

girasquid:
:+1:

lib/active_merchant/billing/gateways/stripe.rb
bizla: Yep, the `icc_data` essentially replaces the transaction JSON. It's the JSON of like... 1981.

girasquid: JSON before it was :cool:

j-mutter: I'm having total deja vu here - didn't I ask this exact question about this exact line somewhere? Is this real life? THERE IS NO SPOON

girasquid: [You totally did!](https://github.com/Shopify/active_merchant/pull/1659#discussion_r28890463)

j-mutter: Phew, ok.

bizla: There is spoon.

PR #1672, opened 2015-05-01
Truncate card number

tjstankus:
This avoids an XSD related error. The authorize.net XSD stipulates:

```xml
<xs:element name="cardNumber">
  <xs:simpleType>
    <xs:restriction base="xs:string">
      <xs:minLength value="4"/>
      <xs:maxLength value="16"/>
    </xs:restriction>
  </xs:simpleType>
</xs:element>
```

Also, I decided to use a constant for the max card number length value. Other truncation values in this class are hardcoded/magic numbers. Normally, I'd feel strongly against magic numbers, but if a magic number would be better here for consistency, that seems reasonable to me.

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -51,6 +51,8 @@ module ActiveMerchant #:nodoc:
 
       APPLE_PAY_DATA_DESCRIPTOR = "COMMON.APPLE.INAPP.PAYMENT"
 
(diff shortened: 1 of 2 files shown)


tjstankus:
Ping @duff

duff:
@tjstankus I hear you re: magic numbers.  I went with your statement about choosing consistency in this case.

If a number such as this is referenced once, I _personally_ prefer to just reference the number rather than having a named value.  I also know others will likely disagree. :smile:

tjstankus:
All good, @duff, thanks for merging!

PR #1673, opened 2015-05-01
[fix] WorldpayOnlinePayments - Description can't be null

ao:


andrewpaliga:
@ao this change looks good but we're still getting an error in our remote tests. Refund method needs to support passing in an amount as a parameter. When I switch `nil` to `@amount` [in the remote refund test](https://github.com/Shopify/active_merchant/blob/master/test/remote/gateways/remote_worldpay_online_payments_test.rb#L69) I get an error.

ao:
@andrewpaliga right now our API processes a full refund if a refundAmount is not set or is set to `0` (hence `nil`). Otherwise it attempts to process the refund as a `partialRefund`.
Therefore it is attempting to process a `partialRefund` of the total amount, which is not partial. That is why it fails.

Is it possible to move forward with this implementation without the API changes as they may take some time to get to production?

andrewpaliga:
@ao unfortunately we cannot move forward with Shopify integration without API changes as all full refunds will fail with the current implementation.

ao:
@andrewpaliga API has been updated and have amended the relevant test.

andrewpaliga:
@ao :+1:

PR #1674, opened 2015-05-06
S5: Add gateway

davidsantoso:
@duff this is the S5 gateway. Let me know if you have any comments/questions/concerns!

### DIFF
--- a/lib/active_merchant/billing/gateways/s5.rb
+++ b/lib/active_merchant/billing/gateways/s5.rb
@@ -1,0 +1,210 @@
+require 'nokogiri'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class S5Gateway < Gateway
+      self.test_url = 'https://test.ctpe.io/payment/ctpe'
+      self.live_url = 'https://ctpe.io/payment/ctpe'
+
+      self.supported_countries = ['DK']
+      self.default_currency = 'EUR'
+      self.supported_cardtypes = [:visa, :master, :maestro]
+
+      self.homepage_url = 'http://www.s5.dk/'
+      self.display_name = 'S5'
+
+      SUPPORTED_TRANSACTIONS = {
+        'sale'      => 'CC.DB',
+        'authonly'  => 'CC.PA',
+        'capture'   => 'CC.CP',
+        'refund'    => 'CC.RF',
+        'void'      => 'CC.RV',
+      }
+
+      def initialize(options={})
+        requires!(options, :sender, :channel, :login, :password)
+        super
+      end
+
+      def purchase(money, payment, options={})
+        request = build_xml_request do |xml|
+          add_payment(xml, money, 'sale', options)
+          add_account(xml, payment)
+          add_customer(xml, payment, options)
+        end
+
+        commit(request)
+      end
+
+      def refund(money, authorization, options={})
+        request = build_xml_request do |xml|
+          add_identification(xml, authorization)
+          add_payment(xml, money, 'refund', options)
+        end
+
+        commit(request)
+      end
+
+      def authorize(money, payment, options={})
+        request = build_xml_request do |xml|
+          add_payment(xml, money, 'authonly', options)
+          add_account(xml, payment)
+          add_customer(xml, payment, options)
+        end
+
+        commit(request)
+      end
+
+      def capture(money, authorization, options={})
+        request = build_xml_request do |xml|
+          add_identification(xml, authorization)
+          add_payment(xml, money, 'capture', options)
+        end
+
+        commit(request)
+      end
+
+      def void(authorization, options={})
+        request = build_xml_request do |xml|
+          add_identification(xml, authorization)
+          add_payment(xml, nil, 'void', options)
+        end
+
+        commit(request)
+      end
+
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r(<Security.+), '\1[FILTERED]\2').
+          gsub(%r(<User login.+), '\1[FILTERED]\2').
+          gsub(%r((<Number>).+(</Number>)), '\1[FILTERED]\2').
+          gsub(%r((<Verification>).+(</Verification>)), '\1[FILTERED]\2')
+      end
+
+      private
+
+      def add_identification(xml, authorization)
+        xml.Identification do
+          xml.ReferenceID authorization
+        end
+      end
+
+      def add_payment(xml, money, action, options)
+        xml.Payment(code: SUPPORTED_TRANSACTIONS[action]) do
+          xml.Memo         "return_code=#{options[:memo]}" if options[:memo]
+          xml.Presentation do
+            xml.Amount     amount(money)
+            xml.Currency   options[:currency] || currency(money)
+            xml.Usage      options[:description]
+          end
+        end
+      end
+
+      def add_account(xml, creditcard)
+        xml.Account do
+          xml.Number        creditcard.number
+          xml.Holder        "#{creditcard.first_name} #{creditcard.last_name}"
+          xml.Expiry(year: creditcard.year, month: creditcard.month)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/s5.rb
```diff
@@ -0,0 +1,236 @@
...
+
+      def add_address(post, creditcard, options)
+        if address = options[:billing_address] || options[:address]
+          post[:address] = "#{address[:address1]} #{address[:address2]}"
+          post[:company] = address[:company].to_s
+          post[:phone]   = address[:phone].to_s.gsub(/[^0-9]/, '') || "0000000"
+          post[:zip]     = address[:zip].to_s
+          post[:city]    = address[:city].to_s
+          post[:country] = address[:country].to_s
+          post[:state]   = address[:state].blank?  ? 'n/a' : address[:state]
```
duff: I'd recommend using 

https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/empty.rb

since we're trying to eliminate the dependency on ActiveSupport.  Like this:

https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L6
https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L307

test/remote/gateways/remote_s5_test.rb
```diff
@@ -0,0 +1,136 @@
...
+    @declined_card = credit_card('4000300011112220')
+
+    @options = {
+      order_id: '1',
+      billing_address: address,
+      description: 'Store Purchase'
+    }
+  end
+
+  def test_dump_transcript
```
duff: I'm not sure how others feel about this, but I've been deleting this test method after it's served its purpose and generated a transcript I can use.  I'd whack it.

lib/active_merchant/billing/gateways/s5.rb
```diff
@@ -0,0 +1,236 @@
...
+        add_memo(post, options) if options[:memo]
+
+        commit('sale', post)
+      end
+
+      def refund(money, authorization, options={})
+        post = {}
+        post[:authorization] = authorization
+        add_invoice(post, money, options)
+        add_memo(post, options) if options[:memo]
```
duff: I'm thinking you can put the conditional of `if options[:memo]` in the `add_memo` method to simplify.  Then you can remove the other few times the conditional is here.

lib/active_merchant/billing/gateways/s5.rb
davidsantoso: Makes sense. I'll swap that out.

test/remote/gateways/remote_s5_test.rb
girasquid: I'm cool with that.

lib/active_merchant/billing/gateways/s5.rb
```diff
@@ -0,0 +1,236 @@
...
+        post[:email] = options[:email]
+        post[:ip] = options[:ip]
+      end
+
+      def add_address(post, creditcard, options)
+        if address = options[:billing_address] || options[:address]
+          post[:address] = "#{address[:address1]} #{address[:address2]}"
+          post[:company] = address[:company].to_s
+          post[:phone]   = address[:phone].to_s.gsub(/[^0-9]/, '') || "0000000"
+          post[:zip]     = address[:zip].to_s
```
duff: I'm thinking you probably don't need to `to_s` these.

test/remote/gateways/remote_s5_test.rb
davidsantoso: Yeah, definitely doesn't seem clean to have a method with just comments in it. I wasn't sure if it was supposed to be kept around in case someone wanted to be able to quickly generate a transcript again, but, if there isn't a formal convention around it, I'd say whack as well.

lib/active_merchant/billing/gateways/s5.rb
duff: Also curious if the gateway blows up if you give it an empty state.

test/unit/gateways/s5_test.rb
```diff
@@ -0,0 +1,413 @@
...
+
+  def test_failed_void
+    @gateway.expects(:ssl_post).returns(failed_void_response)
+
+    response = @gateway.void('')
+    assert_failure response
+  end
+
+  def test_successful_verify
+    @gateway.expects(:ssl_post).twice.returns(successful_authorize_response)
```
duff: Thinking you want:

```
@gateway.expects(:ssl_post).times(2).returns(successful_authorize_response).then.returns(successful_void_response)
```

test/unit/gateways/s5_test.rb
```diff
@@ -0,0 +1,413 @@
...
+
+    response = @gateway.verify(@credit_card, @options)
+    assert_success response
+  end
+
+  def test_successful_verify_with_failed_void
+    @gateway.expects(:ssl_post).returns(failed_void_response)
+
+    response = @gateway.verify(@credit_card, @options)
+    assert_failure response
```
duff: Thinking you want this to be `assert_success`.

test/unit/gateways/s5_test.rb
```diff
@@ -0,0 +1,413 @@
...
+
+  def test_successful_verify
+    @gateway.expects(:ssl_post).twice.returns(successful_authorize_response)
+
+    response = @gateway.verify(@credit_card, @options)
+    assert_success response
+  end
+
+  def test_successful_verify_with_failed_void
+    @gateway.expects(:ssl_post).returns(failed_void_response)
```
duff: I'm thinking 

```
@gateway.expects(:ssl_post).times(2).returns(successful_authorize_response).then.returns(failed_void_response)
```

lib/active_merchant/billing/gateways/s5.rb
davidsantoso: Just tested to confirm and it doesn't blow up if state is empty. Originally I thought it'd be better to have something there ("n/a") than nothing at all, but S5 documentation doesn't even mention addresses so maybe it's just better to remove it.

Thoughts?

duff:
@davidsantoso I made a few little inline comments.  I know that a lot of gateway implementations implement their `purchase`, `authorize`, etc by filling up a hash of params and then calling `commit` with that hash.  And in `commit`, the hash is turned into JSON, a form-encoded string, etc.  I think the approach makes sense in these 2 cases because it's a line or two of code to convert the hash to the needed format.

When the payload is xml, I'm growing skeptical of the approach of filling up a hash first since it eventually has to be mapped in the `build_xml_request` method.  I'm thinking it's simpler for xml payloads if the `purchase` is creating an xml request and then passing that request to `commit`.  Then `purchase` and `refund` and `void` can have very different xml requests and things like the authorization xml element doesn't get passed to `purchase` but it does get passed to `void`.  It just seems like extra work and code to fill a hash from a card and options rather than just mapping it to the xml request.

Here are some examples:

https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/litle.rb#L29-L46
https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/vanco.rb#L115-L143

lib/active_merchant/billing/gateways/s5.rb
duff: :cool:  Then I wouldn't fill it in with anything.  Same goes for phone number if that works.

test/unit/gateways/s5_test.rb
davidsantoso: Specifically with `test_successful_verify_with_failed_void`?

duff: Yup.  Since if the auth works and the void fails, the verify should be successful.

davidsantoso: Ahh gotcha. Sounds good, thanks!

test/unit/gateways/s5_test.rb
davidsantoso: Fixed!

davidsantoso:
Thanks for the great feedback @duff. Made all of the small tweaks then refactored the building of XML to not fill up a hash then map it in `build_xml_request`. Now it just takes a block which cleans up the request body quite a bit.

Let me know if you have any additional thoughts and/or tweaks.

duff:
I'm digging it @davidsantoso.  Looks beautiful to me.

PR #1678, opened 2015-05-08
[POSSIBLE BREAKAGE] Stripe: Make version explicit

duff:
Stripe has many API versions, some of which are breaking changes:

https://stripe.com/docs/upgrades#api-changelog

If someone today created a new Stripe account, the remote tests would
not pass because Stripe has changed some of the attribute responses around
`cards` (used by our `store` method):

https://stripe.com/docs/upgrades#2015-02-18

In addition, the `statement_description` no longer works since it's been
replaced by `statement_descriptor`.  Clients of Active Merchant may be
using `statement_description` so we're now mapping that to
`statement_descriptor` under the covers.

So how do we solve these issues?  This commit makes the specification of
the Stripe API version explicit in the headers.  Previously, you could
specify the version you wanted to use for a particular transaction.  Or,
you could specify it when you instantiated the gateway.  In this commit,
we still allow such customizations.  If no version is specified though,
we explicitly use the current latest version (2015-04-07).

This improves things such that if Stripe releases a new, breaking API
version, clients using Active Merchant won't be broken.  If we want to
upgrade to the default version Active Merchant uses, we can do so
explicitly and change the underlying options we're passing accordingly.

If you're already specifying the Stripe API version, then this commit
shouldn't break you.  If you're relying on the default API version of
the underlying Stripe account, there may be issues which we'll fix as we
learn of them.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -49,7 +49,6 @@ module ActiveMerchant #:nodoc:
         requires!(options, :login)
         @api_key = options[:login]
         @fee_refund_api_key = options[:fee_refund_login]
-        @version = options[:version]
 
         super
       end
@@ -243,7 +242,7 @@ module ActiveMerchant #:nodoc:
           add_customer_data(post, options)
           add_metadata(post, options)
           post[:description] = options[:description]
-          post[:statement_description] = options[:statement_description]
+          post[:statement_descriptor] = options[:statement_description]
           add_customer(post, payment, options)
           add_flags(post, options)
         end
@@ -381,18 +380,21 @@ module ActiveMerchant #:nodoc:
 
       def headers(options = {})
         key     = options[:key] || @api_key
-        version = options[:version] || @version
         idempotency_key = options[:idempotency_key]
 
         headers = {
           "Authorization" => "Basic " + Base64.encode64(key.to_s + ":").strip,
           "User-Agent" => "Stripe/v1 ActiveMerchantBindings/#{ActiveMerchant::VERSION}",
+          "Stripe-Version" => api_version(options),
           "X-Stripe-Client-User-Agent" => user_agent,
           "X-Stripe-Client-User-Metadata" => {:ip => options[:ip]}.to_json
         }
-        headers.merge!("Stripe-Version" => version) if version
         headers.merge!("Idempotency-Key" => idempotency_key) if idempotency_key
         headers
+      end
+
+      def api_version(options)
+        options[:version] || @options[:version] || "2015-04-07"
       end
 
       def api_request(method, endpoint, parameters = nil, options = {})
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * Vanco: Allow specification of fund_id [duff]
 * S5: Add gateway [davidsantoso]
 * SecureNet: Truncate order_id [duff]
+* [POSSIBLE BREAKAGE] Stripe: Be explicit about API version [duff]
 
 
 == Version 1.49.0 (May 1, 2015)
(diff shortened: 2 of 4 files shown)


duff:
@Shopify/active-merchant I'd love any feedback you might have about this PR.  I suspect you may be explicitly specifying the Stripe API version that you're using, but I'm not sure of that.

duff:
The cool part is that I have 2 personal Stripe accounts.  One is a brand new account at the latest API version while the other one is really old using a very old API version.  With this commit, the remote tests now work for both accounts.  The thing I don't know is whether there are some underlying options being passed to Stripe for which we don't have remote tests.

ntalbott:
This looks awesome. Big :+1: :+1: from me.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -243,7 +242,7 @@ def create_post_for_auth_or_purchase(money, payment, options)
           add_customer_data(post, options)
           add_metadata(post, options)
           post[:description] = options[:description]
-          post[:statement_description] = options[:statement_description]
+          post[:statement_descriptor] = options[:statement_description]
```
girasquid: I'm a little nervous about this change, but the other changes in here look ok - doing a little more testing to confirm.

ntalbott: This is actually the change that precipitated the need for more explicit version handling @girasquid; customers using freshly created Stripe accounts are broken today on Spreedly when passing a statement descriptor due to the floating version requirement. These changes should make this a non-breaking change.

girasquid: Oh - maybe I'm missing something here. How does this work for old accounts? Does the descriptor still come through properly?

ntalbott: @girasquid yes, since the version is specified/locked in the request, the request is handled on the latest version of the API.

girasquid: Right - I mean accounts from versions that used `statement_description`, like 2014-08-20. Does it still come through ok there?

duff: :cool:

You narrowed in on the exact option that triggered this pull request.  :+1: 

For old Stripe API versions, `statement_description` worked great.  For newer versions, it bombs with `Received unknown parameter: statement_description`.

So with this PR, we explicitly specify the API version which means we don't really care what the underlying Stripe account is specifying version wise.  We're saying what version we're wanting.

girasquid: I'm following so far, but what I'm wondering is: when you use an API version that *doesn't* have the `statement_descriptor` (and instead expects `statement_description`) - what happens?

girasquid: Aaand that's when he realized that he could run the remote tests and specify a version there, thanks to the statement descriptor test. Looks good.

duff: I'm thinking that it's the responsibility of active merchant to make sure things are in sync.  So the default version specified here:

https://github.com/Shopify/active_merchant/pull/1678/files#diff-8299be227de5ba32c34c30c42a828c9cR397

should be in sync with any options that are passed through (like `statement_descriptor`, `application_fee`, `external_id`, etc).

If, in the future, Stripe adds a new API version wherein `application_fee` is renamed to `super_application_fee`, this PR would still work since we wouldn't be using that latest version, we'd still be using `2015-04-07`.  At some point someone could update Active Merchant to the latest, but then they'd need to use `super_application_fee`.

I think remote tests for Stripe are really important here.  As more options are passed, it'll be more important that we have remote tests for them.

I'd also note that overriding the `version` when you're using an `ActiveMerchant::StripeGateway` is risky since you could specify a version where the options we're passing don't match up.

PR #1680, opened 2015-05-12
USA ePay gateway will now indicate a manual entry

AnotherJoSmith:
Based on the requirements outlined in this email:

> From: Vlad Galyuz <vladg@usaepay.com>
> ...
> 2) You can set the card present flag and place the card number in the credit card field instead of the swipe field and that will indicate that it was manual entry.

@andrewpaliga @girasquid

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -138,6 +138,15 @@ module ActiveMerchant #:nodoc:
       # @return [String]
       attr_accessor :track_data
 
+      # Returns or sets whether a card has been processed using manual entry.
+      #
+      # This attribute is optional and is only used by gateways who use this information in their transaction risk
+      # calculations. See {this page on 'card not present' transactions}[http://en.wikipedia.org/wiki/Card_not_present_transaction]
+      # for further explanation and examples of this kind of transaction.
+      #
+      # @return [true, false]
+      attr_accessor :manual_entry
+
       # Returns or sets the ICC/ASN1 credit card data for a EMV transaction, typically this is a BER-encoded TLV string.
       #
       # @return [String]
--- a/lib/active_merchant/billing/gateways/usa_epay_transaction.rb
+++ b/lib/active_merchant/billing/gateways/usa_epay_transaction.rb
@@ -178,7 +178,8 @@ module ActiveMerchant #:nodoc:
           post[:card]   = credit_card.number
           post[:cvv2]   = credit_card.verification_value if credit_card.verification_value?
           post[:expir]  = expdate(credit_card)
-          post[:name]   = credit_card.name
+          post[:name]   = credit_card.name unless credit_card.name.blank?
+          post[:cardpresent] = true if credit_card.manual_entry
         end
       end
 
@@ -256,4 +257,3 @@ module ActiveMerchant #:nodoc:
     end
   end
 end
-
--- a/test/remote/gateways/remote_usa_epay_transaction_test.rb
+++ b/test/remote/gateways/remote_usa_epay_transaction_test.rb
@@ -3,15 +3,15 @@ require 'test_helper'
 class RemoteUsaEpayTransactionTest < Test::Unit::TestCase
   def setup
     @gateway = UsaEpayTransactionGateway.new(fixtures(:usa_epay))
-    @creditcard = credit_card('4000100011112224')
+    @credit_card = credit_card('4000100011112224')
     @declined_card = credit_card('4000300011112220')
     @credit_card_with_track_data = credit_card_with_track_data('4000100011112224')
-    @options = { :billing_address => address(:zip => "27614", :state => "NC") }
+    @options = { :billing_address => address(:zip => "27614", :state => "NC")}
     @amount = 100
   end
 
   def test_successful_purchase
-    assert response = @gateway.purchase(@amount, @creditcard, @options)
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
     assert_equal 'Success', response.message
     assert_success response
   end
@@ -22,8 +22,22 @@ class RemoteUsaEpayTransactionTest < Test::Unit::TestCase
     assert_success response
   end
 
+  def test_successful_authorization_with_manual_entry
+    @credit_card.manual_entry = true
+    assert response = @gateway.authorize(@amount, @credit_card, @options)
+    assert_equal 'Success', response.message
+    assert_success response
+  end
+
+   def test_successful_purchase_with_manual_entry
+    @credit_card.manual_entry = true
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_equal 'Success', response.message
+    assert_success response
+   end
+
   def test_successful_purchase_with_extra_details
-    assert response = @gateway.purchase(@amount, @creditcard, @options.merge(:order_id => generate_unique_id, :description => "socool"))
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(:order_id => generate_unique_id, :description => "socool"))
     assert_equal 'Success', response.message
     assert_success response
   end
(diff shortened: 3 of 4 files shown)


test/remote/gateways/remote_usa_epay_transaction_test.rb
```diff
@@ -6,7 +6,7 @@ def setup
     @creditcard = credit_card('4000100011112224')
     @declined_card = credit_card('4000300011112220')
     @credit_card_with_track_data = credit_card_with_track_data('4000100011112224')
-    @options = { :billing_address => address(:zip => "27614", :state => "NC") }
+    @options = { :billing_address => address(:zip => "27614", :state => "NC"), :manual_entry => false }
```
andrewpaliga: I would leave the default as not passing a `:manual_entry` option

test/remote/gateways/remote_usa_epay_transaction_test.rb
```diff
@@ -22,6 +22,12 @@ def test_successful_purchase_with_track_data
     assert_success response
   end
 
+   def test_successful_purchase_with_manual_entry
+    assert response = @gateway.purchase(@amount, @creditcard, @options.merge(:manual_entry => true))
+    assert_equal 'Success', response.message
+    assert_success response
+   end
+
```
andrewpaliga: test for authorizations?

andrewpaliga:
@ntalbott fyi

ntalbott:
I think this is fine, and so long as the default behavior stays the same I don't think we'll have any issue with it.

The only thing I'd say is that "manual entry" feels like a `CreditCard` property, so I wonder if it should be a settable flag at the top level. Do any other adapters have a similar flag?

test/remote/gateways/remote_usa_epay_transaction_test.rb
girasquid: :+1: for this

girasquid:
> "manual entry" feels like a CreditCard property, so I wonder if it should be a settable flag at the top level.

:+1: from me for this; it seems like something a `CreditCard` should know about itself.

> Do any other adapters have a similar flag?

I think so, but I'll let @andrewpaliga confirm.

Overall I'm :+1: on this PR once a test for authorize gets added, but I think when we update another integration to support manual entry we should add the property to `CreditCard` as well.

test/remote/gateways/remote_usa_epay_transaction_test.rb
AnotherJoSmith: Will add one

AnotherJoSmith:
I'll be adding a similar check to Authorization.net, but those are the only two that I was told about.

andrewpaliga:
Yeah I agree it makes more sense on the `CreditCard`, let's make that change. Default behaviour will stay the same. I haven't seen the flag on any other adapters.

AnotherJoSmith:
I added a property to CreditCard instead of checking for `:manual_entry` in options. Let me know if this is alright.

lib/active_merchant/billing/gateways/usa_epay_transaction.rb
```diff
@@ -170,7 +170,7 @@ def add_invoice(post, options)
         post[:description]  = options[:description]
       end
 
-      def add_credit_card(post, credit_card)
+      def add_credit_card(post, credit_card, options)
```
ntalbott: Adding `options` to this method signature is no longer necessary, right?

lib/active_merchant/billing/credit_card.rb
```diff
@@ -138,6 +138,15 @@ def brand=(value)
...
       attr_accessor :track_data
 
+      # Returns or sets whether a card has been processed using manual entry.
+      #
+      # This attribute is optional and is only used by gateways who use this information in their transaction risk
+      # calculations. See {this page on 'card not present' transactions}[http://en.wikipedia.org/wiki/Card_not_present_transaction]
+      # for further explanation and examples of this kind of transactions.
+      #
+      # @return [true, false]
+      attr_accessor :manual_entry
```
ntalbott: :+1:

ntalbott:
Some changes you can probably back out, but in general LGTM.

lib/active_merchant/billing/gateways/usa_epay_transaction.rb
AnotherJoSmith: True, will remove this

lib/active_merchant/billing/credit_card.rb
```diff
@@ -138,6 +138,15 @@ def brand=(value)
       # @return [String]
       attr_accessor :track_data
 
+      # Returns or sets whether a card has been processed using manual entry.
+      #
+      # This attribute is optional and is only used by gateways who use this information in their transaction risk
+      # calculations. See {this page on 'card not present' transactions}[http://en.wikipedia.org/wiki/Card_not_present_transaction]
+      # for further explanation and examples of this kind of transactions.
```
girasquid: I think transaction should be singular, here

test/remote/gateways/remote_usa_epay_transaction_test.rb
```diff
@@ -22,6 +22,20 @@ def test_successful_purchase_with_track_data
     assert_success response
   end
 
+  def test_successful_authorization_with_manual_entry
+    @credit_card.manual_entry = true
```
girasquid: You're setting `manual_entry` on `@credit_card`, but the setup method is setting up `@creditcard`. Does this test pass? Should it?

test/remote/gateways/remote_usa_epay_transaction_test.rb
```diff
@@ -22,6 +22,20 @@ def test_successful_purchase_with_track_data
...
+    assert response = @gateway.authorize(@amount, @creditcard, @options)
+    assert_equal 'Success', response.message
+    assert_success response
+  end
+
+   def test_successful_purchase_with_manual_entry
+    @credit_card.manual_entry = true
```
girasquid: Same question here

test/remote/gateways/remote_usa_epay_transaction_test.rb
AnotherJoSmith: Hmmm they pass but I guess that is because the manual_entry parameter is optional. Since credit card is two words should I refactor the test to use `@credit_card`?

girasquid: I'd lean towards using the `@creditcard` variable for this, but if you'd rather rename it I'm cool with that too.

AnotherJoSmith: `@creditcard` is used 31 times in the test directory vs 2893 times for `@credit_card` so i think I will change it.

girasquid: :cool:

lib/active_merchant/billing/gateways/usa_epay_transaction.rb
```diff
@@ -178,7 +178,8 @@ def add_credit_card(post, credit_card)
           post[:card]   = credit_card.number
           post[:cvv2]   = credit_card.verification_value if credit_card.verification_value?
           post[:expir]  = expdate(credit_card)
-          post[:name]   = credit_card.name
+          post[:name]   = credit_card.name unless credit_card.name.blank?
+          post[:cardpresent] = true if credit_card.manual_entry
```
girasquid: Is there any repercussion to passing this in as `false`? If there isn't, this could be `post[:cardpresent] = credit_card.manual_entry`

girasquid: Thinking about it more - I think what you have here reads clearer right now, especially if this isn't required for all transactions.

AnotherJoSmith: Could be nil though

girasquid: Yeah, fair point. Stick with what you have here.

PR #1683, opened 2015-05-13
Ogone: Pass along a timeout value on each request

tomhipkin:
Hey,

I have come across the need to specify a timeout in requests to Ogone. This feature is supported in their API and aims to ensure that a response is always returned. In the case **they** are having trouble acquiring funds, our application will receive and know to start polling for the end result.

Sadly I am unable to write the full test I would like (with actual response) because I cannot simulate the gateway's time out response. There is no way I can see to mark a specific request to behave like a timed out request and respect the `RTIMEOUT` value provided.

This change conforms to their documentation found at the top of page 9 here: http://www.barclaycard.co.uk/business/files/ePDQ_Direct_Link_Guide.pdf

I will be using this change in the wild to see if it works, I'd appreciate any help in creating a valid test response for this.

Cheers.

ActiveMerchant: 1.49.0 
Ruby: 2.1.2p95
ActiveSupport: 4.2.1
Nokogiri: 1.6.6.2

### DIFF
--- a/lib/active_merchant/billing/gateways/ogone.rb
+++ b/lib/active_merchant/billing/gateways/ogone.rb
@@ -335,6 +335,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def commit(action, parameters)
+        add_pair parameters, 'RTIMEOUT', @options[:timeout] if @options[:timeout]
         add_pair parameters, 'PSPID',  @options[:login]
         add_pair parameters, 'USERID', @options[:user]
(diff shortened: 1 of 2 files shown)


tomhipkin:
I've been using this in production for a few weeks now and this change has made a massive difference for our application.

When Ogone takes too long, it now responds with an incomplete response (92) and ensures the rest of the request executes timeout free.

Hoping we can get this accepted now. Thoughts?

girasquid:
Thanks!

PR #1687, opened 2015-05-13
add fraud review to cyber source

### LINKED ISSUE #1681: CyberSource: Response fails if the decision is REVIEW
When using CyberSource's Decision manager, a response of REVIEW can be sent back, depending on your settings, currently `success = response[:decision] == "ACCEPT"` on the [commit method](https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L658) makes it fail since it doesn't say ACCEPT.

Any objections to changing that line to maybe `success = response[:decision] != "REJECT"` ?

geermc4:
Fix #1681


girasquid:
Can you add tests for this? Remote would be ideal if possible, as well as a unit test that stubs the response.

geermc4:
Kind of a late response, but back to this again, however, I need to make a few changes to it, just ran into this `:decision=>"REJECT", :reasonCode=>"100"` so this PR as is doesn't really work. Decision Manager rejects, but the reason code says "success".... because [CyberSource](https://support.cybersource.com/cybskb/index?page=content&id=C156)

geermc4:
@ntalbott @girasquid either of you have any ideas on how to handle this?

I changed my previous commit to the following...

```
       def success?(response)
-        response[:decision] == "ACCEPT" || response[:reasonCode] == '100'
+        accepted?(response) || fraud_review?(response)
+      end
+
+      def accepted?(response)
+        response[:decision] == "ACCEPT"
       end
```

This now does the right thing, however, the reason code from CyberSource is still 100 and so the "error" I get when the transaction fails is [Successful transaction](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L55)

This is part of the `ActiveMerchant::Billing::Response` .... `@message="Successful transaction", @success=false`

CyberSource is broken with it's reason code, but we cant fix that, so what would be a good way to address this?

ntalbott:
@geermc4 heads up, my goal is to have ActiveMerchant consistently **fail** fraud review transactions. The reasoning is that success means "I've got money", and we don't want people running transactions, having them flagged, and not realizing that they need to manually approve them. Of course those who are intentionally using fraud review would just check that flag alongside of success to ensure they're handling it right.

Now, to your actual question: I'm not sure I actually understand the issue. Is it just that the message ends up being confusing?

geermc4:
@ntalbott yes, It fails the transaction with a reason of "Successful transaction", kind of confusing :)
Just got off the phone with CyberSource's support and they will look into why their API returns a 100 for this transaction.

As to failing fraud review, with this (once it's working properly), the response will be review if you have Decision Manager running, so why fail them? I would assume that people who turn it on would know to check right?

ntalbott:
@geermc4 our experience at Spreedly is that often these things are on and people don't know. Then there's massive confusion when they've failed to collect a bunch of dough. Thus the preference to force devs to deal directly with the fraud review flag, vs. being able to ignore it.

geermc4:
@ntalbott Still think it's kind of weird to strip out features for those that do want them. Decision manager wasn't on by default for us, someone had to set it up.

Anyways, incase you dont remove the feature :)...

Took a look at the unparsed XML comming from CyberSource, and it seems that the parsing of that XML is incorrect


```xml
<soap:Envelope>
  <soap:Header>
    <wsse:Security>
      <wsu:Timestamp wsu:Id="Timestamp-1069448404">
        <wsu:Created>2015-06-25T18:00:11.161Z</wsu:Created>
      </wsu:Timestamp>
    </wsse:Security>
  </soap:Header>
  <soap:Body>
    <c:replyMessage>
      <c:merchantReferenceCode>R908752766-P7X28EIZ</c:merchantReferenceCode>
      <c:requestID>4352552107335000001514</c:requestID>
      <c:decision>REJECT</c:decision>
      <c:reasonCode>481</c:reasonCode>
      <c:requestToken>Ahj/7wSR1+eW0Z+hIBfUakGLBwyas2LeZCa2Z7CZKT98GaQyQBT98GaQyTSCFZYKHC8OkmW6QHkrtwJyOvzy2jP0JAL6gAAAwf/o</c:requestToken>
      <c:purchaseTotals>
        <c:currency>USD</c:currency>
      </c:purchaseTotals>
      <c:ccAuthReply>
        <c:reasonCode>100</c:reasonCode>
        <c:amount>442.08</c:amount>
        <c:authorizationCode>888888</c:authorizationCode>
        <c:avsCode>X</c:avsCode>
        <c:avsCodeRaw>I1</c:avsCodeRaw>
        <c:cvCode>M</c:cvCode>
        <c:cvCodeRaw>M</c:cvCodeRaw>
        <c:authorizedDateTime>2015-06-25T18:00:11Z</c:authorizedDateTime>
        <c:processorResponse>100</c:processorResponse>
        <c:reconciliationID>10825317LB5YO0LJ</c:reconciliationID>
      </c:ccAuthReply>
      <c:afsReply>
        <c:reasonCode>100</c:reasonCode>
        <c:afsResult>0</c:afsResult>
        <c:hostSeverity>1</c:hostSeverity>
        <c:consumerLocalTime>11:00:10</c:consumerLocalTime>
        <c:scoreModelUsed>default</c:scoreModelUsed>
        <c:binCountry>us</c:binCountry>
        <c:cardScheme>Visa Credit</c:cardScheme>
        <c:cardIssuer>JPMORGAN CHASE BANK,N.A.</c:cardIssuer>
      </c:afsReply>
      <c:decisionReply>
        <c:casePriority>3</c:casePriority>
        <c:activeProfileReply/>
      </c:decisionReply>
    </c:replyMessage>
  </soap:Body>
</soap:Envelope>
```

So there is a REJECT with a reasonCode of 481 ( [The order has been rejected by Decision Manager](https://support.cybersource.com/cybskb/index?page=content&id=C156) )

But there's also a Successful transaction within the authorization of the card, which according to CyberSource, happens before Decision manager kicks in. So the issue here is really to go and post back to CyberSource to void that authorization then right? Even if you do reject all Decision Manager review/reject, the auth stays there I believe.

ntalbott:
@geermc4 sorry I wasn't clear: I'm very :+1: on properly implementing the `fraud_review?` flag. The distinction is that when a transaction comes back with the `fraud_review?` flag set, `succeeded` should be `false`. Ideally, we'd have a `Response#state`, but instead we have two flags. So:

* Transaction failed to process: `succeeded = false, fraud_review = false`
* Transaction successful, not flagged for fraud review: `succeeded = true, fraud_review = false`
* Transaction successful, but flagged for fraud review: `succeeded = false, fraud_review = true`

Thus if you are using a fraud review tool, you need to check both flags to determine the right course of action.

Now, to your issue - from what I can tell, the problem is multiple `reasonCode` fields. Seems like you'll need a PR that distinguishes and sets the `fraud_review` flag based on the right one.

As far as `void`-ing the auth if it's in review, I think that'd be up to the caller to do as a follow-on step if they don't want to review and approve the transaction, right?

geermc4:
@ntalbott :+1: we are on the same page. Yes I agree with the way you outlined the flags.

On the void topic. The problem is not with review, and yes, up to them if it's up for review, but the current problem I see is that if Decision Manager has rules to *Reject* certain transactions, succeeded will be false, fraud_review would be false, but the Auth could have gone through, and... it stays there.

The XML I pasted above shows exactly that actually...

```
<c:decision>REJECT</c:decision>
....
<c:ccAuthReply>
<c:reasonCode>100</c:reasonCode>
```

Attached is not the same transaction, but the same case, I had to manually go in and do the Authorization Reversal

screen shot 2015-06-25 at 6 48 46 pm

ntalbott:
@geermc4 since auths eventually expire, I'm not sure if that's a situation that's worth handling in AM. Those who are concerned (you?) could maybe check `Response#params` so you know to go back and run the `void`?

geermc4:
"Eventually" is like a month though, I dont know about you but a hold on my card for a significant amount of money is no good, like, from the customers perspective

fastjames:
Just finding this since I ran into it yesterday. The on-site CS rep explained that the auth occurs before the transaction is sent to Decision Manager, which is why it appears to be both successful and not successful. The first thing I noticed yesterday was that reason code 481 isn't in the reason codes table in the code; would adding r481 (and r480) get us a better message? I'll test that out later today.

Second, do you plan to add any additional metadata fields to this PR or should they go on a separate PR? I was asked to pass IP address, user ID, and device fingerprint, none of which appear to be supported. These are all for decision manager so they're at least tangentially related to what's going on here.

ntalbott:
@geermc4 I agree RE needing to run a `void` in that case, then, since there's no real way to know the auth went through. If you can do that and line the flags up with what we agreed on above, I'll work to get this merged.

Looks like some other changes also got smooshed in that should get dropped from this PR?

geermc4:
Ah yes, some new stuff was added to the fork, I'll send a clean one when I get a chance

jasonwebster:

PR #1689, opened 2015-05-14
S5: Remove address requirement for purchase and authorize

davidsantoso:
@duff quick fix for the S5 gateway. Add the customer info if it's passed in the options hash via `billing_address` or just `address`. Otherwise, don't add it to the XML.

### DIFF
--- a/lib/active_merchant/billing/gateways/s5.rb
+++ b/lib/active_merchant/billing/gateways/s5.rb
@@ -126,7 +126,7 @@ module ActiveMerchant #:nodoc:
           xml.Contact do
             xml.Email      options[:email]
             xml.Ip         options[:ip]
-            xml.Phone      address[:phone]
+            xml.Phone      address[:phone] if address
           end
           xml.Address do
             xml.Street     "#{address[:address1]} #{address[:address2]}"
@@ -134,7 +134,7 @@ module ActiveMerchant #:nodoc:
             xml.City       address[:city]
             xml.State      address[:state]
             xml.Country    address[:country]
-          end
+          end if address
           xml.Name do
             xml.Given      creditcard.first_name
             xml.Family     creditcard.last_name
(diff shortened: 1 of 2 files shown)


duff:
@davidsantoso I'm thinking we'd still want the `email` and `ip` to get sent, no?

j-mutter:
Wouldn't you want this [a few lines up](https://github.com/davidsantoso/active_merchant/blob/s5-remove-address-requirement/lib/active_merchant/billing/gateways/s5.rb#L137) instead? Or do you really want to omit *everything*

j-mutter:
Oh hai @duff, haha.

j-mutter:
@davidsantoso Can you add a remote test that confirms the lack of address is acceptable?

davidsantoso:
@duff @j-mutter good call. Should there be a guard statement on `Phone` as well since that's part of the address hash?

@j-mutter yep I'll add one now.

duff:
Good timing @j-mutter !

It is interesting that the generator includes `billing_address` by default in the `@options` in the `setup` method of the remote tests.  I do wonder if the default set of options ought to be more minimalist and then have separate remote tests where options like `billing_address` are included.

duff:
@davidsantoso :+1: on guards where needed there.

j-mutter:
We've been trying to move from the generic `address` key to `billing_address` for a while now, and it should probably stay. I would imagine most gateways will want it, and the ones where it's optional are the outliers.

davidsantoso:
It is helpful when testing for adding addresses to a transaction :wink: This was probably more my issue in not having robust enough of tests.

davidsantoso:
@duff @j-mutter added two additional tests for transactions without addresses and refactored the guard clause on the customer info XML generation.

Let me know if you guys have anymore thoughts/concerns!

PR #1692, opened 2015-05-15
Remove Adyen adapter

ntalbott:
Adyen does not allow any access to their third party API, so having an adapter for them is just misleading. If they eventually open up their API, this adapter did work against their test environment and could be refreshed to add support back in.

However, be 100% sure Adyen has actually changed their policy before adding back; individual Adyen reps may not know policy, and could give bad information regarding accessibility of the direct API.

@girasquid can you :+1: this for Shopify?


girasquid:
:+1:

j-mutter:
:fire:

PR #1696, opened 2015-05-15
CASHnet: uri encode the merchant gateway name

mrezentes:


ntalbott:
Not sure how to test this, so :+1:.

mrezentes:
I felt the same way.

PR #1697, opened 2015-05-17
Litle Gateway: Enhancements to error handling and tokenization

jasonbosco:
- When a Litle request fails due to an XML schema validation error, currently the error message is silently lost and not surfaced in the response object returned. I've fixed this issue by surfacing the `response` and `message` attributes of the top level `litleOnlineResponse` node in the response object.

- Litle allows passing the credit card verification value with tokenization requests as per their documentation (Litle_XML_Reference_Guide_XML8.14_V2.27.pdf: Pg 189):

 > When you submit the CVV2/CVC2/CID in a registerTokenRequest, the Litle platform encrypts and stores the value on a temporary basis for later use in a tokenized Auth/Sale transaction submitted without the value. This is done to accommodate merchant systems/workflows where the security code is available at the time of token registration, but not at the time of the Auth/Sale.

I've added support for this.

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -264,6 +264,12 @@ module ActiveMerchant #:nodoc:
           end
         end
 
+        if parsed.empty?
+          %w(response message).each do |attribute|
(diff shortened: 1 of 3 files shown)


girasquid:
Thanks!

PR #1698, opened 2015-05-18
Indicate a recurring billing method to indicate CVV is not expected

walkertexascoder:
From NMI: "It looks like there is an indication that can be submitted
with the transaction that will indicate it is a recurring charge (not
a first time charge) and it should not expect the CVV to be passed.
The “billing-method” will need to be indicated as “recurring” when the
sale is passed through the API"

This is not a problem for most customers but fails for Optimal
Payments, in particular.


walkertexascoder:
@bvandenbos :eyes: please.

PR #1699, opened 2015-05-18
BPoint gateway - add all response data to params, rename transaction_number => authorization

tjstankus:


duff:
@tjstankus I'm wondering whether an approach like:

https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/cenpos.rb#L229

or

https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/litle.rb#L252

or

https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/bridge_pay.rb#L121

might be simpler than explicitly listing the elements.

tjstankus:
The one caveat with handling the response nodes like that is there's one duplicate node name. A simplified example:

```xml
<ProcessPaymentResult>
  <ResponseCode>string</ResponseCode>
</ProcessPaymentResult>
<response>
  <ResponseCode>string</ResponseCode>
</response>
```

So, both `<response>` and `<ProcessPaymentResult>` have `<ResponseCode>` nodes. This could be handled by using hash keys built from two node names, e.g. `response_response_code`, but that looks kinda gross. Are there any objections to using a nested hash, e.g.:

```ruby
{ process_payment_result: { response_code: 'abc' }, response: { response_code: '123' } }
```

tjstankus:
On an unrelated note, interesting that the Travis build would fail on a Stripe-related error for this PR.

tjstankus:
@duff Ping ^^ - sorry forgot to @ you in previous comments

ntalbott:
@tjstankus general rule of thumb for the `Response#params` hash is that it's allowed and expected to be ugly. API consumers that want beauty should use the specific accessors, and should not depend on the contents of `params`, and so `params` is a quick and dirty, as automatic as possible mapping of the response into a hash. It's sort of "feels right if you're doing it right" thing - the recommended usage path is nice, the not recommended path is a bit grungy.

For this specific case, we generally solve it by using prefixes as you outlined.

duff:
:+1: for ugly prefixes!

tjstankus:
@duff refactored response parsing logic now in place

tjstankus:
Removed `transaction_id` from options per conversation with @duff

PR #1701, opened 2015-05-19
Sagepay: Add Verify

anellis:


test/unit/gateways/sage_pay_test.rb
```diff
@@ -186,6 +186,20 @@ def test_successful_store
     assert_equal '1', response.authorization
   end
 
+  def test_successful_verify
+    response = stub_comms do
+      @gateway.verify(@credit_card, @options)
+    end.respond_with(successful_purchase_response)
```
duff: Since `verify` is using `authorize` and `void` under the covers, I'm thinking you'll want to use those rather than `successful_purchase_response`.  Similar to this:

https://github.com/Shopify/active_merchant/blob/master/test/unit/gateways/merchant_partners_test.rb#L201-L215

anellis: Makes sense! In the existing tests they were using the
successful/unsuccessful purchase stubs for everything (purchase, store, avs
results, cvv results, etc) so I just continued that pattern but I can make
a separate stub for authorize and void.

PR #1702, opened 2015-05-19
Litle Gateway: Ability to optionally use paypage_registration_id 

dontmatta:
Ability to optionally use Litle's paypage registration id obtained via iframe, to make authorization call for token


dontmatta:
Sorry, double clicked.  Let me close this and fix the issues.

PR #1704, opened 2015-05-20
Litle Gateway: Ability to store credit card using paypage

dontmatta:
This allows users to store the credit card using the paypage registration id rather than the credit card number.  This allows users to use either the Litle JS or iframe implementation.


dontmatta:
Can someone please take a look at this?  Thanks!

PR #1695, opened 2015-05-15
Add contactless flag to credit card model. Add contactless support to Stripe.

davidseal:
### Description
Adds flag for tracking whether a swiped or EMV transaction has been done using a contactless method (tapped card or mobile device). While EMV provides some contactless tags, these can't always be relied on since different card and different kernels won't necessarily produce these tags. It's up to the integration solution to track how the card data was read.

cc/ @bizla @ryanbalsdon @abecevello

PR #1705, opened 2015-05-20
Feature/encrypted pin stripe support

ryanbalsdon:
This relies on https://github.com/Shopify/active_merchant/pull/1695

It adds encrypted/Online PIN support to the Stripe EMV endpoint.

cc/ @bizla @davidseal @abecevello

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -159,6 +159,21 @@ module ActiveMerchant #:nodoc:
       # @return [String]
       attr_accessor :fallback_reason
 
+      # Returns or sets whether card-present card data has been read contactlessly.
+      #
+      # @return [true, false]
+      attr_accessor :contactless
+
+      # Returns the ciphertext of the card's encrypted PIN.
+      #
+      # @return [String]
+      attr_accessor :encrypted_pin_cryptogram
+
+      # Returns the Key Serial Number (KSN) of the card's encrypted PIN.
+      #
+      # @return [String]
+      attr_accessor :encrypted_pin_ksn
+
       def type
         ActiveMerchant.deprecated "CreditCard#type is deprecated and will be removed from a future release of ActiveMerchant. Please use CreditCard#brand instead."
         brand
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/credit_card.rb
```diff
@@ -158,6 +158,21 @@ def brand=(value)
...
+
+      # Returns the ciphertext of the card's encrypted PIN.
+      #
+      # @return [String]
+      attr_accessor :encrypted_pin
+
+      # Returns the Key Serial Number (KSN) of the card's encrypted PIN.
+      #
+      # @return [String]
+      attr_accessor :encrypted_pin_key_id
```
bizla: `encrypted_pin_ksn` instead? `id` suggests this is an index of some kind.

davidseal: +1, I think using ksn would be more clear.

ryanbalsdon: It is an index. The table just isn't stored in memory and that was the big innovation of DUKPT.

bizla: So a credit card carries on it an ID rather than a KSN? The acquiring bank knows what to do with it?

ryanbalsdon: KSN has 3 parts: BDK ID, Reader Serial Number and Key ID.

It's the Reader that encrypting the PIN, not the Card.

encrypted_pin_key_id is the terminology used by Stripe so I assume it's what First Data is calling it too.

bizla: Sure, ultimately we want to keep this as generic as possible and make the gateway adapters worry about their field name. Since this is a KSN, the field should be called `key_serial_number` or `encrypted_pin_ksn` -- something that makes it obvious for use in other adapters.

ryanbalsdon:
:pager: @bizla :pizza::question:

bizla:
LGTM. Don't merge using the green button though. I'll merge this after #1695 merges.

PR #1707, opened 2015-05-25
Add verification_value to example usage

camelmasa:
I couldn't purchase if `verification_value` was blank.
I think the example is better if we'll add `verigication_value`.


girasquid:
Thanks!

girasquid:
Thanks!

camelmasa:
@girasquid Thank you for merging :smile:

PR #1708, opened 2015-05-25
Adding Tokenization to Quickbooks Gateway

forrest:
Without PCI compliance, the Quickbooks gateway needs to tokenize the CC number in the client JS before sending to the server (https://developer.intuit.com/docs/0150_payments/0300_developer_guides/tokens). This adds support for dealing with the supplied token.

### DIFF
--- a/test/unit/gateways/quickbooks_test.rb
+++ b/test/unit/gateways/quickbooks_test.rb
@@ -43,7 +43,16 @@ class QuickBooksTest < Test::Unit::TestCase
 
   def test_successful_authorize
     @gateway.expects(:ssl_post).returns(successful_authorize_response)
-    response = @gateway.authorize(@amount, @credit_card, @options)
+    response = @gateway.authorize(@amount, @credit_card, @options.merge({token: "1234"}))
+    assert_success response
+
+    assert_equal @authorization, response.authorization
+    assert response.test?
+  end
+
+  def test_successful_authorize_with_token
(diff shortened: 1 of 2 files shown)


test/unit/gateways/quickbooks_test.rb
```diff
@@ -43,7 +43,16 @@ def test_failed_purchase
 
   def test_successful_authorize
     @gateway.expects(:ssl_post).returns(successful_authorize_response)
-    response = @gateway.authorize(@amount, @credit_card, @options)
+    response = @gateway.authorize(@amount, @credit_card, @options.merge({token: "1234"}))
```
ivanfer: Did you want this in the other test?

forrest: :facepalm: thanks for pointing this out. Will have it fixed in a jiffy

ivanfer:
Do you think you could add a remote test ([found here](https://github.com/Shopify/active_merchant/blob/master/test/remote/gateways/remote_quickbooks_test.rb)) that uses a token for purchase?

forrest:
@ivanfer I'm on a pretty tight timeline, so won't be able to get to this in the short-term. Perhaps someone else will be able to carry this PR the last piece of journey.

ntalbott:
@forrest I'm closing this since it needs a remote test and since it needs to be changed to conform to the ActiveMerchant API (which takes tokens in the same argument as credit cards, vs. as an option). Feel free to re-open or re-submit if/when you want to take this further.

PR #1709, opened 2015-05-25
S5: Set regex closure on scrubbing method

davidsantoso:
@duff quick change to add limits to the regex matchers for scrubbing. Before everything after `<Secruity...` was set to be replaced with `[FILTERED]` but now it should be just the fields we want to be scrubbed.

Also, I decided to filter out the `channel` value. I think it's actually part of the credentials that a user would receive so it should probably also be scrubbed.


UI7777 (commented):
ه

PR #1712, opened 2015-05-28
S5: Include Recurrence tag to enable recurring transactions without CVV

davidsantoso:
@duff quick update to S5 to include the `Recurrence` tag and appropriate `mode` for future recurring transactions on a credit card that has been previously verified. Only question I have is whether or not it should also be included on the `authorize` action? I'm not 100% sure.

Also, as always with S5, the remote test is a bit "meh". I'm having to manually send what error I want back instead of S5 actually doing the validations in `TEST` mode. I've included it, but unfortunately it's very mildly helpful.


duff:
:+1: on including it in the `authorization`.

PR #1713, opened 2015-05-28
Illegal character '&' parsing

masaruhoshi:
__Issue__
Garanti gateway seems to be returning XML responses with invalid characters (non-html entities).

__Solution__
Check if the raw response from Garanti is parseable. If it's not, try to clear it removing occurrences of '&' characters. Perhaps this may be happening due to a badly generated response exclusively from Garanti, but I decided placing the method `strip_invalid_xml_chars` in `Gateway` as it may be useful in other gateways.

__Review__
@girasquid @aprofeit cc: @Shopify/payments

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -212,16 +212,6 @@ module ActiveMerchant #:nodoc:
         })
       end
 
-      def strip_invalid_xml_chars(xml)
-        begin
-          REXML::Document.new(xml)
-        rescue REXML::ParseException
-          xml = xml.gsub(/&(?!(?:[a-z]+|#[0-9]+|x[a-zA-Z0-9]+);)/, '&amp;')
-        end
-
-        xml
-      end
-
       private # :nodoc: all
 
       def name
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -212,6 +212,16 @@ module ActiveMerchant #:nodoc:
         })
       end
 
+      def strip_invalid_xml_chars(xml)
+        begin
+          REXML::Document.new(xml)
+        rescue REXML::ParseException
+          xml = xml.gsub(/&(?!(?:[a-z]+|#[0-9]+|x[a-zA-Z0-9]+);)/, '&amp;')
+        end
+
+        xml
+      end
+
       private # :nodoc: all
 
       def name
--- a/lib/active_merchant/billing/gateways/garanti.rb
+++ b/lib/active_merchant/billing/gateways/garanti.rb
@@ -230,7 +230,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def parse(body)
-        xml = REXML::Document.new(body)
+        xml = REXML::Document.new(strip_invalid_xml_chars(body))
 
         response = {}
         xml.root.elements.to_a.each do |node|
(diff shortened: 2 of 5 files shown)


lib/active_merchant/billing/gateway.rb
```diff
@@ -212,6 +212,16 @@ def user_agent
         })
       end
 
+      def strip_invalid_xml_chars(xml)
+        begin
+          REXML::Document.new(xml)
+        rescue REXML::ParseException
+          xml = xml.gsub(/&(?!(?:[a-z]+|#[0-9]+|x[a-zA-Z0-9]+);)/, '&amp;')
```
girasquid: If I'm following this correctly - if you have valid XML, you'll return a `REXML::Document` - but if you have invalid XML that raises a `REXML::ParseException`, you'll get a string back? Can you make the return values consistent?

This should also be going through Nokogiri; we'd like to use that over REXML going forward.

girasquid:
Can you keep `strip_invalid_xml_chars` on the Garanti gateway for now? We can move it to `Gateway` when another gateway needs it, but for now we only know that Garanti does.

masaruhoshi:
np! :ok_hand:

aprofeit:
Can you link the original exception/issue? Just curious on what it looks like.

lib/active_merchant/billing/gateways/garanti.rb
```diff
@@ -230,7 +230,7 @@ def commit(money,request)
       end
 
       def parse(body)
-        xml = REXML::Document.new(body)
+        xml = REXML::Document.new(strip_invalid_xml_chars(body))
```
j-mutter: Why rebuild the REXML doc if it's being done in the method? Maybe the strip method should just run the gsub no matter what?

masaruhoshi:
@aprofeit, they are [here] (https://errbit-virginia.cs.shopify.com/apps/54b810d6f9fec1380b000003/problems/553a3737f9fec10bf600005f) and [here] (https://errbit-virginia.cs.shopify.com/apps/54b810d6f9fec1380b000003/problems/555253a2f9fec163c400023a)

lib/active_merchant/billing/gateways/garanti.rb
masaruhoshi: That's actually a good point.

masaruhoshi:
@girasquid , @aprofeit , @j-mutter , any more comments ?

lib/active_merchant/billing/gateways/garanti.rb
```diff
@@ -251,6 +251,10 @@ def success?(response)
         response[:message] == "Approved"
       end
 
+      def strip_invalid_xml_chars(xml)
+        xml.gsub(/&(?!(?:[a-z]+|#[0-9]+|x[a-zA-Z0-9]+);)/, '&amp;')
```
girasquid: This regex is a bit hard to parse - would the regex [here](http://stackoverflow.com/a/17237801/240418) do the same job but be more clear?

masaruhoshi: That regex covers basically '&', '<', '>', double quote and apostrophe. Any other html entity (such as &atilde; &ccedil; etc) will have wrongfully converted their '&' with '&amp;'.

www.debuggex.com is a really option to understand complex regex.

girasquid: Fair enough!

girasquid:
:+1:

aprofeit:
:+1:

PR #1714, opened 2015-05-28
viaklix: add sales tax, payjunction: support for swiped cards

waelchatila:
payjunction support for physical present cards (swiped cards)

viaklix: if sales tax field not being sent


ntalbott:
@waelchatila these would need to be split into separate PR's and have tests to be merged. Also note that track data implementations should closely follow the conventions used in other gateways.

Feel free to open new PR's with these separated and tested if you want to carry them forward - thanks!

PR #1717, opened 2015-05-31
Give travis another break

byroot:
Ref: https://github.com/Shopify/active_merchant/pull/1226

The idea here is that there is little value to test all the ruby + rails combinaisons. We only need to ensure the code works with all supported Rails version and all Ruby versions independently.

This will reduce the number of jobs per build to 6 from 12 currently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ script: "bundle exec rake test:units"
 sudo: false
 
 rvm:
-- "2.2"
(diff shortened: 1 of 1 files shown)


ntalbott:
Closed by https://github.com/activemerchant/active_merchant/commit/662a1dcbe1b59b1fd4d3aa2d035f3ce513e57b3a.

PR #1720, opened 2015-06-02
Add Verify to Cardstream

anellis:


anellis:
@duff Adding Verify to Cardstream.

As mentioned, the remote tests are wonky for our account right now.

PR #1728, opened 2015-06-05
CheckoutV2: Add Gateway

anellis:
@duff Here's checkout. As I mentioned, I just changed purchase to multiresponse so that it works with void and refund

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -1,0 +1,198 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CheckoutV2Gateway < Gateway
+      self.display_name = "Checkout.com V2 Gateway"
+      self.homepage_url = "https://www.checkout.com/"
+      self.test_url = "https://api2.checkout.com/v2/"
+      self.live_url = "http://sandbox.checkout.com/api2/v2/"
+
+      self.supported_countries = ['AT', 'BE', 'BG', 'CY', 'CZ', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GR', 'HR', 'HU', 'IE', 'IS', 'IT', 'LI', 'LT', 'LU', 'LV', 'MT', 'MU', 'NL', 'NO', 'PL', 'PT', 'RO', 'SE', 'SI', 'SK', 'US']
+      self.default_currency = "USD"
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club]
+
+      def initialize(options={})
+        requires!(options, :secret_key)
+        super
+      end
+
+      def purchase(amount, payment_method, options={})
+        MultiResponse.run do |r|
+          r.process { authorize(amount, payment_method, options) }
+          r.process { capture(amount, r.authorization, options) }
+        end
+      end
+
+      def authorize(amount, payment_method, options={})
+        post = {}
+        post[:autoCapture] = "n"
+        add_invoice(post, amount, options)
+        add_payment_method(post, payment_method)
+        add_customer_data(post, options)
+
+        commit(:authorize, post)
+      end
+
+      def capture(amount, authorization, options={})
+        post = {}
+        add_invoice(post, amount, options)
+        add_customer_data(post, options)
+
+        commit(:capture, post, authorization)
+      end
+
+      def void(authorization, options={})
+        post = {}
+        commit(:void, post, authorization)
+      end
+
+      def refund(amount, authorization, options={})
+        post = {}
+        add_invoice(post, amount, options)
+        add_customer_data(post, options)
+
+        commit(:refund, post, authorization)
+      end
+
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: )[a-zA-Z0-9._-]+)i, '\1[FILTERED]').
+          gsub(%r((\\",\\"number\\":\\")\d+), '\1[FILTERED]').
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -0,0 +1,212 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CheckoutV2Gateway < Gateway
+      self.display_name = "Checkout.com V2 Gateway"
+      self.homepage_url = "https://www.checkout.com/"
+
+      self.live_url = "http://sandbox.checkout.com/api2/v2/"
```
duff: This looks a bit :fish:y.  Do they have a non-sandox url?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -0,0 +1,212 @@
...
+      end
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        "20014" => STANDARD_ERROR_CODE[:invalid_number],
+        "20100" => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        "20054" => STANDARD_ERROR_CODE[:expired_card],
+        "40104" => STANDARD_ERROR_CODE[:incorrect_cvc],
+        "40108" => STANDARD_ERROR_CODE[:incorrect_zip],
+        "40111" => STANDARD_ERROR_CODE[:incorrect_address],
+        "20014" => STANDARD_ERROR_CODE[:card_declined],
```
duff: I'm getting a warning running remote tests:

```
gateways/checkout_v2.rb:191: warning: duplicated key at line 197 ignored: "20014"
```

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -0,0 +1,212 @@
...
+        post[:card][:name] = payment_method.name
+        post[:card][:number] = payment_method.number
+        post[:card][:cvv] = payment_method.verification_value
+        post[:card][:expiryYear] = format(payment_method.year, :four_digits)
+        post[:card][:expiryMonth] = format(payment_method.month, :two_digits)
+      end
+
+      def add_customer_data(post, options)
+        post[:email] = options[:email] || "unspecified@example.com"
+        if(billing_address = (options[:billing_address] || options[:address]) && post[:card])
```
duff: I don't think we need the `options[:address` part for new gateways.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -0,0 +1,212 @@
...
+          post[:card][:billingDetails][:city] = billing_address[:city]
+          post[:card][:billingDetails][:state] = billing_address[:state]
+          post[:card][:billingDetails][:country] = billing_address[:country]
+          post[:card][:billingDetails][:postcode]    = billing_address[:zip]
+          post[:card][:billingDetails][:phone] ={}
+          post[:card][:billingDetails][:phone][:number] = billing_address[:phone]
+        end
+      end
+
+      def add_reference(post, authorization)
```
duff: I don't see where this method gets called.  Perhaps it could be :hocho:.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -0,0 +1,212 @@
...
+      def parse_element(response, node)
+        if node.has_elements?
+          node.elements.each{|element| parse_element(response, element) }
+        else
+          response[node.name.underscore.to_sym] = node.text
+        end
+      end
+
+      def success_from(response)
+        (response["responseCode"] == ("10000" || "10100")) || response["cards"]
```
duff: Perhaps a bit of refactoring here could make it more intention revealing.  (I wouldn't have expected `response["cards"]` to indicate success).

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -0,0 +1,212 @@
...
+        return {} unless body
+        JSON.parse(body)
+        rescue JSON::ParserError
+          {
+            "message" => "Invalid response received.",
+            "raw_response" => scrub(body)
+          }
+      end
+
+      def parse_element(response, node)
```
duff: I'm not seeing where this method gets called.  Could be another :hocho: candidate.

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -0,0 +1,113 @@
...
+    assert_scrubbed(@declined_card.verification_value, transcript)
+    assert_scrubbed(@gateway.options[:secret_key], transcript)
+  end
+
+  def test_successful_purchase
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+  end
+
```
duff: I'm thinking it would be useful to have a `test_successful_purchase_with_minimal_options` to make sure we don't need `order_id`, `email`, or `description` to run run a purchase.

test/unit/gateways/checkout_v2_test.rb
```diff
@@ -0,0 +1,330 @@
...
+  end
+
+  def test_failed_verify
+    response = stub_comms do
+      @gateway.verify(@credit_card)
+    end.respond_with(failed_authorize_response, successful_void_response)
+    assert_failure response
+    assert_equal "Invalid Card Number", response.message
+  end
+
```
duff: It would cool to have a unit test checking the behavior of us getting back invalid json from the gateway.

duff:
@anellis I gave this a quick look over and made a few inline comments.  I'm thinking we're pretty close on this one.

anellis:
@duff Per their API documentation it looks like the sandbox url is their only one and you just specify a test charge id and secret key if you want it to be a test, but I agree that it looks strange so I sent Ravish a message just to clarify. I've made all the other changes and I'm just waiting to hear back on that.

anellis:
@duff updated! some of the weirdness (like the response["cards"]) was left over from when I was trying to implement store before realizing that their store wasn't the same as how we think of it (for them you have to store specific customers and then store cards attached to those customers, which is why it was requiring a unique email for it) so I took that stuff out

PR #1729, opened 2015-06-05
Add scrubbing to cybersource

jnormore:
@andrewpaliga @j-mutter @ntalbott @duff a couple reviews please? thanks :)


test/unit/gateways/authorize_net_test.rb
```diff
@@ -1,3 +1,4 @@
+# encoding: utf-8
```
jnormore: was causing an error in the ruby 1.9.3 tests

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -6,7 +6,7 @@ def setup
 
     @gateway = CyberSourceGateway.new({nexus: "NC"}.merge(fixtures(:cyber_source)))
 
-    @credit_card = credit_card('4111111111111111')
+    @credit_card = credit_card('4111111111111111', verification_value: '321')
```
jnormore: ``assert_scrubbed`` finds ``123`` from the email if we don't use a different cvv here

test/unit/gateways/authorize_net_test.rb
j-mutter: Not supported anymore? :)

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -52,6 +52,36 @@ def setup
...
+    credit_card = network_tokenization_credit_card('4111111111111111',
+      :brand              => 'visa',
+      :eci                => "05",
+      :payment_cryptogram => "EHuWW9PiBkWvqE5juRwDzAUFBAk="
+    )
+
+    transcript = capture_transcript(@gateway) do
+      assert auth = @gateway.authorize(@amount, credit_card, @options)
+      assert_success auth
+      assert_equal 'Successful transaction', auth.message
```
j-mutter: Any reason to assert these here and not in the test above?

ntalbott:
Not sure if it's helpful, but we've used this code in Spreedly for a while:

```ruby
  def scrub(transcript)
    transcript.
      gsub(%r((<wsse:Password [^>]*>)[^<]*(</wsse:Password>))i, '\1[FILTERED]\2').
      gsub(%r((<accountNumber>)[^<]*(</accountNumber>))i, '\1[FILTERED]\2').
      gsub(%r((<cvNumber>)[^<]*(</cvNumber>))i, '\1[FILTERED]\2').
      gsub(%r((<cavv>)[^<]*(</cavv>))i, '\1[FILTERED]\2').
      gsub(%r((<xid>)[^<]*(</xid>))i, '\1[FILTERED]\2').
      gsub(%r((<authenticationData>)[^<]*(</authenticationData>))i, '\1[FILTERED]\2')
  end
```

jnormore:
thanks @ntalbott, I've replace with your version, I'd rather use something you've been using for a while :)

j-mutter:
@ntalbott @jnormore  cool - I was just about to ask if the username/merchant ID should be filtered out. I didn't think they would be considered sensitive, and could be useful when debugging.

test/remote/gateways/remote_cyber_source_test.rb
jnormore: nope, removed :)

j-mutter:
:hatched_chick: once :green_heart:

test/unit/gateways/authorize_net_test.rb
jnormore: riiiigghhttt, removed :)

andrewpaliga:
:+1:

PR #1731, opened 2015-06-05
Sage: Credit really is credit not refund

duff:
Revert "Sage gateway: Deprecate #credit call"

This reverts commit 425ffea0837ef592949d809735c855b647c22131.

I had deprecated the `credit` call 2 years ago thinking at the time that
the `credit` was really a refund.  In fact it was not.  It's a general
credit that isn't based on a previous transaction id.

I don't see an ability to do a standard `refund` in the Sage docs:

http://www.coderxo.com/docs/BankcardHTTPS.pdf

For now, this commit removes `refund` since the interface was broken.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
 * Banwire: Add default email [anellis]
 * PayU India: Handle bad JSON [ntalbott]
 * Dibs: Pass CVC param only if there's a value [bruno]
+* Sage: Credit really is credit not refund [duff]
 
 
 == Version 1.50.0 (June 1, 2015)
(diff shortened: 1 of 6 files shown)


ntalbott:
Bummer that you haven't found a proper `refund`, but :+1: on fixing this. :shipit:

duff:
I've emailed our contacts at Sage about it.  We'll add refund if they let us know of something we're missing.

duff:
I heard back from Sage re: a standard refund.  I'll be implementing that forthwith.

duff:
Added refund: d854072b17b9cdc8ff296b5445d6931663fd190d

PR #1733, opened 2015-06-07
Add error_code and error_code_from to gateway generator

jnormore:
We added ``STANDARD_ERROR_CODE_MAPPING = {}`` to the generated gateway in https://github.com/Shopify/active_merchant/commit/1a0c650f10e711f7f5a21954192fda3057c5e10d, we should also add the method that should use this map.

@jduff @j-mutter @andrewpaliga


andrewpaliga:
LGTM

j-mutter:
:cool:

PR #1734, opened 2015-06-10
filter more invalid characters in epay url

dwradcliffe:
Problem: epay is now sending < and > in the url also

Example: `https://ssl.ditonlinebetalingssystem.dk/auth/default.aspx?decline=1&error=209&orderid=xxxxxxxxxxx&errortext=The%20payment%20was%20declined%20of%20unknown%20reasons.%20For%20more%20information%20contact%20the%20bank.%20E.g.%20try%20with%20another%20credit%20card.<br%20/>Denied%20-%20Call%20your%20bank%20for%20information`

Solution: This patch replaces those characters with the appropriate escape codes.

r: @mutemule @Shopify/active-merchant

### DIFF
--- a/lib/active_merchant/billing/gateways/epay.rb
+++ b/lib/active_merchant/billing/gateways/epay.rb
@@ -186,7 +186,7 @@ module ActiveMerchant #:nodoc:
         # Authorize gives the response back by redirecting with the values in
         # the URL query
         if location = response['Location']
-          query = CGI::parse(URI.parse(location.gsub(' ', '%20')).query)
+          query = CGI::parse(URI.parse(location.gsub(' ', '%20').gsub('<', '%3C').gsub('>', '%3E')).query)
         else
           return {
             'accept' => '0',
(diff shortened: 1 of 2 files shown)


eapache:
From the looks of it they are actually just sending raw html in the parameter; should we be preemptively escaping any other characters?

lib/active_merchant/billing/gateways/epay.rb
```diff
@@ -186,7 +186,7 @@ def do_authorize(params)
         # Authorize gives the response back by redirecting with the values in
         # the URL query
         if location = response['Location']
-          query = CGI::parse(URI.parse(location.gsub(' ', '%20')).query)
+          query = CGI::parse(URI.parse(location.gsub(' ', '%20').gsub('<', '%3C').gsub('>', '%3E')).query)
```
eapache: `gsub` takes a map as its second parameter IIRC, so you should be able to do this with a single call rather than chaining them together

j-mutter:
Test? :innocent:

dwradcliffe:
@eapache I'd rather not guess at what else they are going to break and add unnecessary code. I attempted using a hash but it didn't seem to work and the code was much more obtuse, IMO.

@j-mutter test added!

dwradcliffe:
ping @j-mutter @eapache @ntalbott for review, please :)

eapache:
LGTM

PR #1735, opened 2015-06-10
Introduce PayeezyGateway

huoxito:
Hi all,

I believe [this gateway](https://developer.payeezy.com/) is somehow related to the existing FirstdataE4 gateway here. I took / copied the test examples from that one to write the tests for this one.

For the record, the PayeezyGateway class here was originally created by Sachin Shetty. I only removed dependencies and added test examples. Please let me know if there's something wrong regarding active_merchant compatibility when you have a chance.

cheers

### DIFF
--- a/lib/active_merchant/billing/gateways/payeezy.rb
+++ b/lib/active_merchant/billing/gateways/payeezy.rb
@@ -1,0 +1,256 @@
+module ActiveMerchant
+  module Billing
+    class PayeezyGateway < Gateway
+      class_attribute :integration_url
+
+      self.test_url = 'https://api-cert.payeezy.com/v1/transactions'
+      self.integration_url = 'https://api-cat.payeezy.com/v1/transactions'
+      self.live_url = 'https://api.payeezy.com/v1/transactions'
+
+      self.default_currency = 'USD'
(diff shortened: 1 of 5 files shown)


JDutil:
@huoxito should probably add to the README as well.

huoxito:
just amended the commit with README change and a couple other fixes after running this in production, pls let me know if I could help with anything else to get this merged

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -0,0 +1,256 @@
...
+
+      def authorize(money, creditcard, options = {})
+        commit(:authorize, purchase_or_auth_request(money, creditcard, options))
+      end
+
+      def capture(money, authorization, options = {})
+        commit(:capture, capture_or_credit_request(money, authorization, options))
+      end
+
+      def credit(money, authorization, options = {})
```
ntalbott: This should be `refund` not `credit`. `credit` does an unreferenced push of funds to a card.

huoxito:
thanks @ntalbott, I've rebased on top of master again and renamed the method properly

PR #1736, opened 2015-06-11
Support for iATS ACH

rwdaigle:
@duff @ntalbott a few notes. This PR...

* Modifies the authorization ref returned for `purchase` transactions from the pure ref returned by the gateway, e.g., `A7F8B8B3` to a composite, e.g., `A7F8B8B3|check`. Non check purchases still return the raw ref, e.g., `A7F8B8B3`. I don't believe this to be a breaking change, but could use a :+1: 
* Due to the way iATS handles test ACH transactions, it's not possible to test the failure of an ACH refund transaction in real-time (even their test ACH transactions are time-delayed). I've done my best, but not sure how else to test this aspect.


duff:
:+1:

PR #1738, opened 2015-06-12
ogone: refactor url path generation to allow for new endpoints

tahb:
To send a request to a different endpoint the code needed to become more flexible to accommodate this and more. I've refactored it and it's now extendable whilst maintaining the previous logic (presence of payid dictating one of two endpoints, hmm!)

### DIFF
--- a/lib/active_merchant/billing/gateways/ogone.rb
+++ b/lib/active_merchant/billing/gateways/ogone.rb
@@ -356,13 +356,13 @@ module ActiveMerchant #:nodoc:
         response
       end
 
-      def commit(action, parameters)
-        add_pair parameters, 'RTIMEOUT', @options[:timeout] if @options[:timeout]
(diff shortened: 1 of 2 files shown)


girasquid:
Can you add tests for this?

tomhipkin:
@girasquid Added these now, and some more thorough request specs that didn't exist beforehand!

ntalbott:
@tomhipkin it looks like `:query` isn't used anywhere in the code? So I don't see the need for the added complexity here.

tomhipkin:
@ntalbott Ah yes, I have another feature branch ready to PR which has these refactorings as a dependency. 

You can see it here: https://github.com/tomhipkin/active_merchant/commit/9f7180188363382d80fd15738738fd5ce67c7126

tomhipkin:
@ntalbott @girasquid is there anything further I can do to progress this change?

PR #1742, opened 2015-06-15
iATS Gateway: Fix nil error on empty authorization result

tjstankus:
Meh. I started to go down the path of forcing the authorization result to be an
empty string instead of nil, but then noticed that there are other values in the
response hash that can be nil. So, I opted for consistency and modified the code
to handle nils. I avoided the use of `try`, however.

### DIFF
--- a/lib/active_merchant/billing/gateways/iats_payments.rb
+++ b/lib/active_merchant/billing/gateways/iats_payments.rb
@@ -212,7 +212,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def successful_result_message?(response)
-        response[:authorization_result].start_with?('OK')
+        result = response[:authorization_result]
+        return false if result.nil?
+        result.start_with?('OK')
       end
 
       def success_from(response)
@@ -221,7 +223,7 @@ module ActiveMerchant #:nodoc:
 
       def message_from(response)
         if(!successful_result_message?(response))
-          return response[:authorization_result].strip
+          return (response[:authorization_result] || '').strip
         elsif(response[:status] == 'Failure')
           return response[:errors]
         else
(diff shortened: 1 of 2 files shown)


duff:
Mulling on this a bit.  It does seem odd that we got a nil for `response[:authorization_result]`.  Some gateways get their responses differently when the transaction succeeded verses when it failed:

https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L426
https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/pay_conex.rb#L230
https://github.com/Shopify/active_merchant/blob/master/lib/active_merchant/billing/gateways/dibs.rb#L187

I think that this PR is an improvement over what's out there.  That being said, the original logic that's been out there around `success_from` being based on `successful_result_message?` seems a big squirrelly to me.  When I look at the docs for iATS, I see this:

normal_and_iats_webservices_processlink_4_3_pdf__page_13_of_72_

Those docs help me understand why things are squirrelly.  Whew!

The thing I'm concerned about is that according to our stack trace, `response[:status] == "Success"` evaluated to true but with this logic around `response[:authorization_result]` nil checking, the transaction would be marked as failed.  I'm wondering what iATS support might say.  If we get `response[:status] == "Success"` and `response[:authorization_result]` is `nil`, is that a successful transaction?  Their docs list 2 possible results for  `response[:authorization_result]` -> "OK" and "Error".  But in this case, we got neither. :sweat:

tjstankus:
@duff Agreed, this is confusing. So is the other iATS PR I just submitted. I'll follow up with iATS support on this one.

tjstankus:
Investigating whether #1743 affects response data, which would affect this issue as well.

tjstankus:
Taking a wait-and-see approach to this issue. It only surfaced twice in quick succession and hasn't surfaced since. Closing this PR.

PR #1745, opened 2015-06-17
Authorize.Net - Add invoice to refund

marquisong:
This PR adds a small update to Authorize.Net, specifically the refund method, to allow the passing of invoice information in the options over to Authorize.Net during refund action.

As a result, the invoice information will be displayed in Authorize.Net's dashboard under transaction detail and report page.
I have tested this change using Authorize.Net's sandbox testing.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,7 @@
 * S5: Make scrubbing regex less greedy [duff]
 * CardStream: Add support for verify [anellis]
 * Authorize.net: UTF-8 encode requests [duff]
+* Authorize.net: Add invoice information to refund [marquisong]
 * Banwire: Add default email [anellis]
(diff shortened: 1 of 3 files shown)


aprofeit:
Thanks for the PR. Do you have a link to documentation which shows these are valid fields in a refund? The reference at http://developer.authorize.net/api/reference/ does not include `invoiceNumber` or `description` as valid options, although they are present on other requests.

marquisong:
I actually got this working with the support from Authorize.Net's support team.
Also, the following post in Authorize.Net's forum has mentioned about the invoiceNumber field for refund (by Authorize.Net's moderator Joy).
Btw, I will try to find out further documentation on this from Authorize.Net.
Thanks @aprofeit

aprofeit:
I was able to verify this works, thanks!

PR #1748, opened 2015-06-17
Support mapping Stripe advanced decline codes to standard codes

abecevello:
Adds support for mapping Stripe advanced decline codes to the standard error codes.

Also adds a mapping for the 'call_issuer' advanced decline code to the standardized 'call_issuer' error code.

@bizla @girasquid

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -42,6 +42,7 @@ module ActiveMerchant #:nodoc:
         'incorrect_cvc' => STANDARD_ERROR_CODE[:incorrect_cvc],
         'incorrect_zip' => STANDARD_ERROR_CODE[:incorrect_zip],
         'card_declined' => STANDARD_ERROR_CODE[:card_declined],
+        'call_issuer' => STANDARD_ERROR_CODE[:call_issuer],
         'processing_error' => STANDARD_ERROR_CODE[:processing_error]
       }
 
@@ -430,7 +431,7 @@ module ActiveMerchant #:nodoc:
           :avs_result => { :code => avs_code },
           :cvv_result => cvc_code,
           :emv_authorization => card["emv_auth_data"],
-          :error_code => success ? nil : STANDARD_ERROR_CODE_MAPPING[response["error"]["code"]]
+          :error_code => success ? nil : error_code_from(response)
         )
       end
 
@@ -458,6 +459,15 @@ module ActiveMerchant #:nodoc:
 
       def emv_payment?(payment)
         payment.respond_to?(:emv?) && payment.emv?
+      end
+
+      def error_code_from(response)
+        code = response['error']['code'] 
+        decline_code = response['error']['decline_code'] if code == 'card_declined'
+
+        error_code = STANDARD_ERROR_CODE_MAPPING[decline_code]
+        error_code ||= STANDARD_ERROR_CODE_MAPPING[code]
+        error_code
       end
     end
   end
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -419,6 +419,28 @@ class StripeTest < Test::Unit::TestCase
     assert_failure response
 
     assert_equal Gateway::STANDARD_ERROR_CODE[:card_declined], response.error_code
+    refute response.test? # unsuccessful request defaults to live
+    assert_equal 'ch_test_charge', response.authorization
+  end
+
+  def test_declined_request_advanced_decline_codes
+    @gateway.expects(:ssl_request).returns(declined_call_issuer_purchase_response)
+
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_failure response
+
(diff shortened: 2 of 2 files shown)


girasquid:
@duff I'd be curious to get some Spreedly thoughts on this as well

girasquid:
Tests?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -459,6 +460,20 @@ def non_fractional_currency?(currency)
       def emv_payment?(payment)
         payment.respond_to?(:emv?) && payment.emv?
       end
+
+      def error_code_from(response)
```
bizla: You'll want to add some unit tests for this.

bizla: This can be simplified as: 

```ruby
code = response['error']['code'] 
decline_code = response['error']['decline_code'] 

if decline_code && code == 'card_declined'
  STANDARD_ERROR_CODE_MAPPING[decline_code] 
else
  STANDARD_ERROR_CODE_MAPPING[code] 
end
```

abecevello: @bizla that isn't equivalent to what I have. In my code if `STANDARD_ERROR_CODE_MAPPING[decline_code]` doesn't have a mapping, the code falls back to `STANDARD_ERROR_CODE_MAPPING[code]`. Your version doesn't do that.

abecevello: I can use parts of your suggestion to simplify what I have though. Change coming shortly once I test it.

bizla:
```ruby
code = response['error']['code'] 
decline_code = response['error']['decline_code'] 

error_code = if decline_code && code == 'card_declined'
  STANDARD_ERROR_CODE_MAPPING[decline_code] 
else
  STANDARD_ERROR_CODE_MAPPING[code] 
end

error_code
```

duff:
@girasquid, we're not using the `STANDARD_ERROR_CODE_MAPPING` stuff at Spreedly yet.  I haven't yet formed any strong opinions wrt that area. :)

abecevello:
Updated to have tests and simplify the `error_code_from` method.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -459,6 +460,15 @@ def non_fractional_currency?(currency)
...
+        code = response['error']['code'] 
+        decline_code = response['error']['decline_code'] if code == 'card_declined'
```
girasquid: You can use `STANDARD_ERROR_CODE[:card_declined]` here instead of a magic string

test/unit/gateways/stripe_test.rb
```diff
@@ -423,6 +423,28 @@ def test_declined_request
...
   end
 
+  def test_declined_request_advanced_decline_codes
+    @gateway.expects(:ssl_request).returns(declined_call_issuer_purchase_response)
+
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_failure response
+
+    assert_equal Gateway::STANDARD_ERROR_CODE[:call_issuer], response.error_code
+    assert !response.test? # unsuccessful request defaults to live
```
girasquid: `refute` ?

girasquid:
:+1:

lib/active_merchant/billing/gateways/stripe.rb
bizla: I don't think you want to do that actually, since `card_declined` just happens to be our mapping string as well as Stripe's. Stripe could change that string and this snippet would suggest they're supposed to be the same.

test/unit/gateways/stripe_test.rb
bizla: Probably can omit this assertion altogether.

lib/active_merchant/billing/gateways/stripe.rb
girasquid: You're right - keep the string!

abecevello: Although they are the same string in this case, technically they are in different domains. The string here is part of the Stripe API domain. `STANDARD_ERROR_CODE[:card_declined]` is in the standard error code domain. There is nothing stopping these values from being different, although convention will probably have them be the same value forever.

Up to you if you want me to use the standard one instead.

abecevello: And @bizla beat me to the explanation.

girasquid: I think when we originally built these we copied them off of what Stripe had; the odds of this ever going out of sync with what they have are pretty low. That said - good arguments here for keeping it as a separate string, so I say keep it.

bizla:
:pizza: I'll merge this in.

aprofeit:
@bizla you gonna merge this in?

bizla:
Yup. Just running through all the other PRs in order.

PR #1753, opened 2015-06-23
Redsys: Corrected AUD currency code

agseco:


aprofeit:
Thanks for the find. Do you have a link to the documentation that shows this is the correct code? I'm curious if `ARS` and `BOB` should get the same treatment, stripping the leading `0`.

agseco:
I'm currently awaiting confirmation from CatalunyaCaixa/Redsys for ```ARS``` and ```BOB```currencies in this regard.

Once I get the confirmation I can fix the currency codes if it's required and send a new PR.

In terms of documentation, I'm afraid unfortunately there are no public manuals hosted. I'll enquire just in case.

agseco:
Our CatalunyaCaixa agent confirmed leading ```0``` should be stripped from currency codes. I'll send another PR with this changes.

PR #1757, opened 2015-06-24
Authorize.net: Add store ability

duff:
This allows us to use the Authorize.net gateway like we always have.  It
also allows us to store cards in CIM and run purchases, auths, voids,
captures, and refunds against those stored cards.

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -62,82 +62,50 @@ module ActiveMerchant #:nodoc:
       end
 
       def purchase(amount, payment, options = {})
-        commit("AUTH_CAPTURE") do |xml|
-          add_order_id(xml, options)
-          xml.transactionRequest do
-            xml.transactionType('authCaptureTransaction')
-            xml.amount(amount(amount))
-
-            add_payment_source(xml, payment)
-            add_invoice(xml, options)
-            add_customer_data(xml, payment, options)
-            add_market_type(xml, payment)
-            add_settings(xml, payment, options)
-            add_user_fields(xml, amount, options)
+        if payment.is_a?(String)
+          commit(:cim_purchase) do |xml|
+            add_cim_auth_purchase(xml, "profileTransAuthCapture", amount, payment, options)
+          end
+        else
+          commit(:purchase) do |xml|
+            add_auth_purchase(xml, "authCaptureTransaction", amount, payment, options)
           end
         end
       end
 
       def authorize(amount, payment, options={})
-        commit("AUTH_ONLY") do |xml|
-          add_order_id(xml, options)
-          xml.transactionRequest do
-            xml.transactionType('authOnlyTransaction')
-            xml.amount(amount(amount))
-
-            add_payment_source(xml, payment)
-            add_invoice(xml, options)
-            add_customer_data(xml, payment, options)
-            add_market_type(xml, payment)
-            add_settings(xml, payment, options)
-            add_user_fields(xml, amount, options)
+        if payment.is_a?(String)
+          commit(:cim_authorize) do |xml|
+            add_cim_auth_purchase(xml, "profileTransAuthOnly", amount, payment, options)
+          end
+        else
+          commit(:authorize) do |xml|
+            add_auth_purchase(xml, "authOnlyTransaction", amount, payment, options)
           end
         end
       end
 
       def capture(amount, authorization, options={})
-        commit("PRIOR_AUTH_CAPTURE") do |xml|
-          add_order_id(xml, options)
-          xml.transactionRequest do
-            xml.transactionType('priorAuthCaptureTransaction')
-            xml.amount(amount(amount))
-            xml.refTransId(split_authorization(authorization)[0])
-
-            add_invoice(xml, options)
-            add_user_fields(xml, amount, options)
-          end
+        if auth_was_for_cim?(authorization)
+          cim_capture(amount, authorization, options)
+        else
+          normal_capture(amount, authorization, options)
         end
       end
 
       def refund(amount, authorization, options={})
-        transaction_id, card_number = split_authorization(authorization)
-        commit("CREDIT") do |xml|
-          xml.transactionRequest do
-            xml.transactionType('refundTransaction')
-            xml.amount(amount.nil? ? 0 : amount(amount))
-            xml.payment do
-              xml.creditCard do
-                xml.cardNumber(card_number || options[:card_number])
-                xml.expirationDate('XXXX')
-              end
-            end
-            xml.refTransId(transaction_id)
-
-            add_customer_data(xml, nil, options)
-            add_user_fields(xml, amount, options)
-          end
+        if auth_was_for_cim?(authorization)
+          cim_refund(amount, authorization, options)
+        else
+          normal_refund(amount, authorization, options)
         end
       end
 
       def void(authorization, options={})
-        commit("VOID") do |xml|
-          add_order_id(xml, options)
-          xml.transactionRequest do
-            xml.transactionType('voidTransaction')
-            xml.refTransId(split_authorization(authorization)[0])
-
-            add_user_fields(xml, nil, options)
-          end
+        if auth_was_for_cim?(authorization)
+          cim_void(authorization, options)
+        else
+          normal_void(authorization, options)
         end
       end
 
@@ -146,7 +114,7 @@ module ActiveMerchant #:nodoc:
           raise ArgumentError, "Reference credits are not supported. Please supply the original credit card or use the #refund method."
         end
 
-        commit("CREDIT") do |xml|
+        commit(:credit) do |xml|
           add_order_id(xml, options)
           xml.transactionRequest do
             xml.transactionType('refundTransaction')
@@ -168,6 +136,29 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def store(credit_card, options = {})
+        commit(:cim_store) do |xml|
+          xml.profile do
+            xml.merchantCustomerId(truncate(options[:merchant_customer_id], 20) || SecureRandom.hex(10))
+            xml.description(truncate(options[:description], 255)) unless empty?(options[:description])
+            xml.email(options[:email]) unless empty?(options[:email])
+
+            xml.paymentProfiles do
+              xml.customerType("individual")
+              add_billing_address(xml, credit_card, options)
+              add_shipping_address(xml, options, "shipToList")
+              xml.payment do
+                xml.creditCard do
+                  xml.cardNumber(truncate(credit_card.number, 16))
+                  xml.expirationDate(format(credit_card.year, :four_digits) + '-' + format(credit_card.month, :two_digits))
+                  xml.cardCode(credit_card.verification_value) if credit_card.verification_value
+                end
+              end
+            end
+          end
+        end
+      end
+
       def supports_scrubbing?
         true
       end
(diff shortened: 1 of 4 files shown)


duff:
@ntalbott (and anyone else who'd like to take a look),

I'd love a second set of :eyes: on this one.  It's somewhat meaty.  :meat_on_bone:

I don't think this changes any existing behavior; it just adds functionality.  You'll notice that testing wise, the changes there are almost entirely additive.

duff:
In theory, this could mean that we could someday move 

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net_cim.rb

to its own gem.  (not sure how many folks are actually using it).

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -62,82 +62,50 @@ def initialize(options={})
...
-          add_order_id(xml, options)
-          xml.transactionRequest do
-            xml.transactionType('priorAuthCaptureTransaction')
-            xml.amount(amount(amount))
-            xml.refTransId(split_authorization(authorization)[0])
-
-            add_invoice(xml, options)
-            add_user_fields(xml, amount, options)
-          end
+        if (auth_was_for_cim?(authorization))
```
masaruhoshi: You can remove the external parenthesis

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -62,82 +62,50 @@ def initialize(options={})
...
-                xml.cardNumber(card_number || options[:card_number])
-                xml.expirationDate('XXXX')
-              end
-            end
-            xml.refTransId(transaction_id)
-
-            add_customer_data(xml, nil, options)
-            add_user_fields(xml, amount, options)
-          end
+        if (auth_was_for_cim?(authorization))
```
masaruhoshi: Remove the external parenthesis

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -62,82 +62,50 @@ def initialize(options={})
...
       def void(authorization, options={})
-        commit("VOID") do |xml|
-          add_order_id(xml, options)
-          xml.transactionRequest do
-            xml.transactionType('voidTransaction')
-            xml.refTransId(split_authorization(authorization)[0])
-
-            add_user_fields(xml, nil, options)
-          end
+        if (auth_was_for_cim?(authorization))
```
masaruhoshi: The outside parenthesis (sorry for being annoying :smile: )

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -168,6 +136,29 @@ def verify(credit_card, options = {})
         end
       end
 
+      def store(credit_card, options = {})
+        commit(:cim_store) do |xml|
+          xml.profile do
+            xml.merchantCustomerId(truncate(options[:merchant_customer_id], 20) || SecureRandom.hex(10))
```
masaruhoshi: So we just create a random customer id if none is provided?

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -180,20 +171,139 @@ def scrub(transcript)
 
       private
 
+      def add_auth_purchase(xml, transaction_type, amount, payment, options)
+        add_order_id(xml, options)
+        xml.transactionRequest do
+          xml.transactionType(transaction_type)
+          xml.amount(amount(amount))
```
masaruhoshi: This make you wonder if you're seeing in double. Or triple.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -180,20 +171,139 @@ def scrub(transcript)
...
+            xml.profileTransPriorAuthCapture do
+              xml.amount(amount(amount))
+              xml.transId(transaction_id)
+            end
+          end
+        end
+      end
+
+      def normal_capture(amount, authorization, options)
+        transaction_id = split_authorization(authorization)[0]
```
masaruhoshi: change it to
`transaction_id, _ = split_authorization(authorization)`

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -180,20 +171,139 @@ def scrub(transcript)
...
+          end
+        end
+      end
+
+      def cim_void(authorization, options)
+        commit(:cim_void) do |xml|
+          add_order_id(xml, options)
+          xml.transaction do
+            xml.profileTransVoid do
+              xml.transId(split_authorization(authorization)[0])
```
masaruhoshi: Why did you stop getting the `transaction_id` at the beginning of the method?

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -180,20 +171,139 @@ def scrub(transcript)
...
+          end
+        end
+      end
+
+      def normal_void(authorization, options)
+        commit(:void) do |xml|
+          add_order_id(xml, options)
+          xml.transactionRequest do
+            xml.transactionType('voidTransaction')
+            xml.refTransId(split_authorization(authorization)[0])
```
masaruhoshi: You should get this `transaction_id` before building the xml, as you did on `cim_refund` and the other methods.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -305,54 +420,63 @@ def add_check(xml, check)
...
-        xml.customerIP(options[:ip]) unless empty?(options[:ip])
+      def add_shipping_address(xml, options, root_node="shipTo")
+        address = options[:shipping_address] || options[:address]
+        return unless address
 
-        xml.cardholderAuthentication do
-          xml.authenticationIndicator(options[:authentication_indicator])
-          xml.cardholderAuthenticationValue(options[:cardholder_authentication_value])
+        xml.send(root_node) do
+          (first_name, last_name) = if address[:name]
```
masaruhoshi: You don't need the parenthesis here.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -465,14 +623,50 @@ def parse(action, body)
...
-        (
-          response[:response_code] == APPROVED &&
-          TRANSACTION_ALREADY_ACTIONED.exclude?(response[:response_reason_code])
-        )
+      def parse_cim(body)
+        response = {}
+
+        doc = Nokogiri::XML(body).remove_namespaces!
+
+        if (element = doc.at_xpath("//messages/message"))
```
masaruhoshi: Remove external parenthesis.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -465,14 +623,50 @@ def parse(action, body)
...
+          response[:message_code] = element.at_xpath("code").content[/0*(\d+)$/, 1]
+          response[:message_text] = element.at_xpath("text").content.chomp('.')
+        end
+
+        response[:result_code] = if(element = doc.at_xpath("//messages/resultCode"))
```
masaruhoshi: What about this:
```
element = doc.at_xpath("//messages/resultCode")
response[:result_code] = element.content if element && !empty(element.content)
```

lib/active_merchant/billing/gateways/authorize_net.rb
ntalbott: @masaruhoshi is there a particular style guide you're targeting? In general we haven't enforced a style guide on AM, so I don't think these tweaks are necessarily germane.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -62,82 +62,50 @@ def initialize(options={})
...
-            xml.transactionType('authCaptureTransaction')
-            xml.amount(amount(amount))
-
-            add_payment_source(xml, payment)
-            add_invoice(xml, options)
-            add_customer_data(xml, payment, options)
-            add_market_type(xml, payment)
-            add_settings(xml, payment, options)
-            add_user_fields(xml, amount, options)
+        if payment.is_a?(String)
```
ntalbott: @duff generally we've reversed this, and first checked `respond_to?(:number)` to see if it's a credit card, then assumed it's a token otherwise. Not sure if you care :grin:

lib/active_merchant/billing/gateways/authorize_net.rb
ntalbott: @duff so annoying that a CIM capture is different!

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -180,20 +171,139 @@ def scrub(transcript)
...
+      end
+
       def add_payment_source(xml, source)
         return unless source
-        if card_brand(source) == 'check'
+
+        case source
+        when String
+          add_token_payment_method(xml, source)
+        when Check
```
ntalbott: `is_a?` checks are problematic, since it's common to define your own "quacks like" payment methods. Can you use a `respond_to?` instead?

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -305,54 +420,63 @@ def add_check(xml, check)
...
+      def add_shipping_address(xml, options, root_node="shipTo")
+        address = options[:shipping_address] || options[:address]
+        return unless address
 
-        xml.cardholderAuthentication do
-          xml.authenticationIndicator(options[:authentication_indicator])
-          xml.cardholderAuthenticationValue(options[:cardholder_authentication_value])
+        xml.send(root_node) do
+          (first_name, last_name) = if address[:name]
+            address[:name].split
```
ntalbott: I dunno, but bare `split` always feels risky to me.

lib/active_merchant/billing/gateways/authorize_net.rb
ntalbott: @duff all these lookups look the same - maybe a little helper method?

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -481,30 +675,92 @@ def message_from(response, avs_result, cvv_result)
...
+          [response[:customer_profile_id], response[:customer_payment_profile_id], action].join("#")
+        else
+          [response[:transaction_id], response[:account_number], action].join("#")
+        end
       end
 
       def split_authorization(authorization)
-        transaction_id, card_number = authorization.split("#")
-        [transaction_id, card_number]
+        authorization.split("#")
```
ntalbott: So the fact that `Response#authorization` now has three elements won't break existing stored authorizations?

ntalbott:
@duff some inline suggestions, but overall :+1:.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Shouldn't break 'em since for the existing behavior we're only accessing the first two.  And if the last element is empty, we use the current behavior.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Agreed.  Will do.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Was thinking I don't need to, since it's only referenced once in this method.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Hmmm... not sure why it would be better to be before.

Talking to @ntalbott here, I'm gonna make a helper method for this `split` to be a bit more explicit.

lib/active_merchant/billing/gateways/authorize_net.rb
masaruhoshi: Good point.

lib/active_merchant/billing/gateways/authorize_net.rb
masaruhoshi: Thanks for the heads up, @ntalbott. :+1: 
I mentioned the parenthesis as they're not mandatory in ruby and to be consistent with the other places he's using ifs.

duff: They're quite different!  Same goes for `refund`, `void`, etc.

@masaruhoshi I'll look at removing some parens.  They can be a bit noisy.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Agreed.  Will do.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Yup.  Most folks probably don't care about `merchant_customer_id`.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Will use a helper method for this.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: I'll look into this too.

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Good catch.  I think this one has been here for awhile but I'll look at tightening it up.

masaruhoshi:
Good job, @duff. :tanabata_tree:

lib/active_merchant/billing/gateways/authorize_net.rb
duff: Looked at this more and I'm thinking the existing may be simpler since in this gateway's case, there's one path that cares if it's CIM, and another path that cares about the others which aren't just cards (there's eChecks, credit cards, apple pay stuff).

duff:
Implemented changes!  Thanks for the :eyes: !

PR #1758, opened 2015-06-25
Leading zeroes stripped from currency codes

agseco:
As opposed to what is specified in the CatalunyaCaixa/Redsys documentation, our agent has confirmed us that currency codes must be specified without any leading ```0```.

Otherwise transactions will fail with ```SIS0042```error.


aprofeit:
Thanks!

PR #1762, opened 2015-06-28
Allow to pass currency to MiGS gateway

alovak:
MiGS accepts vpc_Currency field. If more than one currency is configured for the merchant then this field is required.


girasquid:
Can you add tests for this?

alovak:
@girasquid I have added currency to remote test. Also fixed offsite_purchase test.

But,

with my MID refund_test fails (because it works only for purchase mode)
and test_server_purchase_url works for Visa and MC as other brands are not enabled for my MID.

girasquid:
Thanks!

PR #1764, opened 2015-06-29
Send contactless read_method whenever it is set

abecevello:
Changes stripe.rb to always pass the `read_method` parameter of `contactless` if the credit card says it is a contactless card. This is needed to support contactless magswipe cards where they have track data but we also need to specify that they are contactless.

Please review @bizla @girasquid

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -305,6 +305,7 @@ module ActiveMerchant #:nodoc:
           if creditcard.respond_to?(:track_data) && creditcard.track_data.present?
             card[:swipe_data] = creditcard.track_data
             card[:fallback_reason] = creditcard.fallback_reason if creditcard.fallback_reason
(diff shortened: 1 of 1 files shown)


girasquid:
:+1:

girasquid:
Test fails look legit though - you'll need to handle the card being something that's not quite a credit card

abecevello:
@girasquid Should be fixed.

bizla:
:pizza: I'll commit this on green CI.

abecevello:
@bizla @girasquid I had to modify this to set the contactless flag when we know we have a card. With the previous code of setting it at the beginning, the `post[:card]` object is nil which then crashes. Setting it to an empty hash wouldn't work since some of the add_ methods just set the hash, which would overwrite the contactless flag.

This introduced a small amount of duplication, but I'm not sure what the active merchant convention is for these constants to avoid the duplication.

bizla:
LGTM.

PR #1766, opened 2015-06-30
Add supports_network_tokenization to gateways

jnormore:
This verifies that a gateway supports network tokenization, all but auth.net use a true/false response right now.

@girasquid @bizla thoughts?


lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -173,6 +173,25 @@ def scrub(transcript)
...
 
+      def supports_network_tokenization?
+        card = Billing::NetworkTokenizationCreditCard.new({
+          :number => "4111111111111111",
+          :month => 12,
+          :year => 20,
+          :first_name => 'John',
+          :last_name => 'Smith',
+          :brand => 'visa',
+          :payment_cryptogram => 'EHuWW9PiBkWvqE5juRwDzAUFBAk='
```
girasquid: Where does this value come from?

jnormore: test cryptogram, used in all remote tests too

girasquid:
Change looks ok - @ntalbott / @duff, any thoughts?

lib/active_merchant/billing/gateways/authorize_net.rb
girasquid: :cool:

bizla: A comment next to that line would be helpful IMO. 
(Or instead of `new` would a a class method on `NetworkTokenizationCreditCard` like `test_credit_card` be useful here? Is that cryptogram reusable?)

jnormore: > Is that cryptogram reusable?

Doesn't matter, we don't care about the success/failure response, we only care that the '155' error code wasn't returned (which means the processor doesn't support network tokenization)

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -173,6 +173,25 @@ def scrub(transcript)
...
+        })
+
+        request = post_data(:authorize) do |xml|
+          add_auth_purchase(xml, "authOnlyTransaction", 1, card, {})
+        end
+        raw_response = ssl_post(url, request, headers)
+        response = parse(:authorize, raw_response)
+        response[:response_reason_code].to_s != "155"
```
bizla: Constantize `"155"` to give it some semantic value?

test/unit/gateways/braintree_blue_test.rb
```diff
@@ -579,6 +579,10 @@ def test_apple_pay_card
     assert_equal "transaction_id", response.authorization
   end
 
+  def test_supports_network_tokenization
+    assert_instance_of TrueClass, @gateway.supports_network_tokenization?
```
bizla: why not a simple `assert`?

jnormore: I want it to be explicit that it's ``true`` and not another response that passes the assertion

bizla: :cool:

bizla:
One outstanding comment but otherwise LGTM.

lib/active_merchant/billing/gateways/authorize_net.rb
girasquid: :+1: for this

test/unit/gateways/braintree_blue_test.rb
duff: One other possibility:

```
assert_equal true, @gateway.supports_network_tokenization?
```

duff:
@girasquid LGTM

PR #1769, opened 2015-07-02
Add GMO Gateway

chrisblatchley:
This enables the use of the GMO-PG gateway, used in Japan. It been tested with live money successfully.


CaDs:
I wonder why GMO-PG gateway is no longer in the current maser branch.

PR #1772, opened 2015-07-07
Authorize.net: Add allow_partial_auth field

anellis:
@duff Here is adding the partial auth gateway specific field. The only slightly strange thing about it is that, when doing a partial auth, until the balance is paid in full the transactions are held instead of approved so it doesn't look like a success.  The message is pretty clear though.

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -327,6 +327,12 @@ module ActiveMerchant #:nodoc:
               xml.settingValue("true")
             end
           end
+          if options[:disable_partial_auth] == "true"
+            xml.setting do
+              xml.settingName("allowPartialAuth")
+              xml.settingValue("false")
+            end
+          end
           if options[:duplicate_window]
             set_duplicate_window(xml, options[:duplicate_window])
           elsif self.class.duplicate_window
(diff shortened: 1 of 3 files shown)


duff:
@anellis I've been mulling on whether it makes sense to simplify a bit.  This authorize.net feature sure is one I haven't seen on other gateways.  I'm wondering if, rather than give our customers the full ability to mess around with the setting (and all of the ramifications of that setting around `splitTenderId`s and the adjusted handling of the authorization), we instead only give them ability to turn off the feature explicitly.  

So perhaps there could be a `options[:disable_partial_auth]` thing that gets passed to ActiveMerchant which would simply set ` <allowPartialAuth>false</allowPartialAuth>`

I'm also not sure of all of the ramifications of this part of the docs:

> If successfully authorized, all transactions in the group are held until the final transaction of the group is successfully authorized.

All in all, since our customer was bitten by the fact that the underlying auth.net account had the feature _enabled_ and because this customer wants the ability to explicitly disable it, perhaps giving active merchant that ability may suffice for now.  Whatcha think?

anellis:
Hmm that's true, I was thinking since we're interacting with it at all we may as well go ahead and make it fully available now in case anyone does want it enabled, but it does make things much more complex and this probably falls along the lines of that YAGNI acronym I keep seeing you guys throw around :)

anellis:
@duff updated to be infinitely simpler

duff:
@anellis :+1: 

I just tweaked it to use `true` rather than `"true"` as the indicator.

PR #1773, opened 2015-07-07
Merchant Ware: Update credentials for remote tests

davidsantoso:
@duff: Quick update to Merchant Ware (version 4) adapter to use active/valid credentials. Remote tests should now pass again.


davidsantoso:
Awesome thanks!

PR #1774, opened 2015-07-07
S5: Add Store

anellis:


anellis:
@duff addresses spreedly/core#1231

duff:
@anellis In theory, we should also be able to run an auth/purchase using a card stored in the S5 vault.  Example: 

https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_pay_conex_test.rb#L148

anellis:
@duff ah, sorry! Updated.

duff:
:+1: :+1:

PR #1776, opened 2015-07-08
Mercury: No longer default to allow partial auth

duff:
Most customers don't want to allow partial auths.  The mercury docs
state:

Supporting partial authorizations is required for all POS systems.

Mercury tech support reported that for an eCommerce transaction, that
allow partial auth flag should not be passed.  Then when a customer uses
a gift card with insufficient funds, they'd get an expected decline
response.

### DIFF
--- a/lib/active_merchant/billing/gateways/mercury.rb
+++ b/lib/active_merchant/billing/gateways/mercury.rb
@@ -90,7 +90,7 @@ module ActiveMerchant #:nodoc:
           xml.tag! "Transaction" do
             xml.tag! 'TranType', 'Credit'
             xml.tag! 'TranCode', action
-            if action == 'PreAuth' || action == 'Sale'
+            if options[:allow_partial_auth] && (action == 'PreAuth' || action == 'Sale')
               xml.tag! "PartialAuth", "Allow"
             end
             add_invoice(xml, options[:order_id], nil, options)
@@ -113,7 +113,7 @@ module ActiveMerchant #:nodoc:
         xml.tag! "TStream" do
           xml.tag! "Transaction" do
             xml.tag! 'TranType', 'Credit'
-            if action == 'PreAuthCapture'
+            if options[:allow_partial_auth] && (action == 'PreAuthCapture')
               xml.tag! "PartialAuth", "Allow"
             end
             xml.tag! 'TranCode', (@use_tokenization ? (action + "ByRecordNo") : action)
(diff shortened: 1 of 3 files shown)


duff:
@ntalbott This one is a bit funky so I'd dig a second set of :eyeglasses: on it.  I was going to base it on whether the card had track_data present, but we don't get a card for `capture`.  I'd also love a :+1: or a :-1: from @activemerchant/developers in case Shopify is offended by having the extra option to pass for POS transactions.

The alternative is to pass allow partial auth by default and allow users of active merchant to turn that off with a `disable_partial_auth` option.  Now that I'm writing this comment, I'm leaning that way.

duff:
Perhaps it should be implemented similar to this: https://github.com/activemerchant/active_merchant/commit/bccb5002229d142f2da6fe368d3729d8d6c2cb4a

lib/active_merchant/billing/gateways/mercury.rb
```diff
@@ -90,7 +90,7 @@ def build_non_authorized_request(action, money, credit_card, options)
           xml.tag! "Transaction" do
             xml.tag! 'TranType', 'Credit'
             xml.tag! 'TranCode', action
-            if action == 'PreAuth' || action == 'Sale'
+            if (options[:allow_partial_auth] == true) && (action == 'PreAuth' || action == 'Sale')
```
ntalbott: Why `== true`? I'd expect a simple truthy check would suffice?

lib/active_merchant/billing/gateways/mercury.rb
```diff
@@ -113,7 +113,7 @@ def build_authorized_request(action, money, authorization, credit_card, options)
         xml.tag! "TStream" do
           xml.tag! "Transaction" do
             xml.tag! 'TranType', 'Credit'
-            if action == 'PreAuthCapture'
+            if (options[:allow_partial_auth] == true) && (action == 'PreAuthCapture')
```
ntalbott: Same.

ntalbott:
@duff so it's pretty clear there's not an AM standard for partial auth today. If there was, I'd like that standard to be "partial auth is disabled unless explicitly enabled". Thus, I like your approach here. :+1:

aprofeit:
> I'd like that standard to be "partial auth is disabled unless explicitly enabled".

That makes sense. This seems like the right approach.

andrewpaliga:
LGTM as well

lib/active_merchant/billing/gateways/mercury.rb
duff: I agree.  Not sure what I was thinking.  

I had thougts about people passing in "false", but they shouldn't do that.  All in all, not a very vulcan moment.  Truthiness is better and simpler.  Will fix.

PR #1778, opened 2015-07-09
Adding network tokenization to Moneris

andrewpaliga:
@jnormore @girasquid for review please

### DIFF
--- a/test/remote/gateways/remote_moneris_test.rb
+++ b/test/remote/gateways/remote_moneris_test.rb
@@ -21,6 +21,17 @@ class MonerisRemoteTest < Test::Unit::TestCase
     assert_false response.authorization.blank?
   end
 
+  def test_successful_purchase_with_network_tokenization
+    @credit_card = network_tokenization_credit_card('4242424242424242',
+      payment_cryptogram: "BwABB4JRdgAAAAAAiFF2AAAAAAA=",
+      verification_value: nil
+    )
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_moneris_test.rb
```diff
@@ -156,13 +178,6 @@ def test_cvv_match_when_enabled
     assert_equal({'code' => 'M', 'message' => 'CVV matches'}, response.cvv_result)
   end
 
-  def test_cvv_no_match_when_enabled
```
girasquid: Why did this get removed?

girasquid:
One question about a test disappearing, but looks ok overall.

test/remote/gateways/remote_moneris_test.rb
andrewpaliga: The test failed and I don't understand why it wouldn't be expected to fail. The test above is the exact same (other than the amount) and expects a success response... (same response the deleted test was actually getting)

girasquid: Is it possible that that amount is supposed to trigger the CVV response?

jnormore:
LGTM, same question about the test though. i don't see anything obvious in why it stopped working.

test/remote/gateways/remote_moneris_test.rb
andrewpaliga: Not sure, but the test is also failing on master so...

girasquid: I guess if it's always been failing it makes sense to remove it.

PR #1779, opened 2015-07-09
Merchant Ware v4: Add support for verify

davidsantoso:
@duff: quick addition to the Merchant Ware version 4 adapter to add support for the verify action.


duff:
@davidsantoso We usually have a unit test or two as well.

duff:
The unit test's value can seem dubious, but when other folks don't have remote credentials, perhaps they provide value.

davidsantoso:
Ah right. Alrighty, adding!

davidsantoso:
@duff alrighty, added `verify` unit tests. Let me know if there's anything else missing.

duff:
@davidsantoso :+1:

PR #1781, opened 2015-07-10
Stripe: Make purchase via vaulted card consistent

duff:
In particular, the way most gateways are using `store` is like this:

```
store = @gateway.store(@credit_card)
response = @gateway.purchase(@amount, store.authorization)
```

This commit adjusts Stripe to behave in this way, such that we don't
need to grab specific parts of the response.params in order to run
purchases against vaulted cards.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 * Merchant Ware v4: Add support for verify [davidsantoso]
 * Mercury: No longer default to allow partial auth [duff]
 * PayPal: Fix soft_descriptor and support soft_descriptor_city [duff]
+* Stripe: Make purchase via vaulted card consistent [duff]
 
 
 == Version 1.51.0 (July 2, 2015)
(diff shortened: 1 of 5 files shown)


duff:
@activemerchant/developers I'd :heart: a 2nd set of :eyes: on this one.  The easiest place to spot the adjustment is in the remote tests.  Thanks!

duff:
@ntalbott expressed a concern about killing the existing API for stripe using the `options` rather than deprecating it.  I updated this PR to address those concerns such that now things ought to be deprecated appropriately.

I can't wait for us to remove all of the deprecations!

aprofeit:
Little late, but :+1:

duff:
@aprofeit Thanks!  I got a :+1: from @ntalbott in another channel and thought it would ok to merge.

aprofeit:
Was my bad taking so long :p

duff:
HA!  I have no expectation time wise in terms of getting feedback. :)  We're all volunteers here in active merchant land.

PR #1785, opened 2015-07-14
Update authorize.net device type to wireless pos

abecevello:
This updates the device type parameter for Authorize.net card present transactions to be wireless pos (code 7). This was suggested to us to use instead of the unknown type.

I'm not sure if there is a way to make this a parameter so we don't need active merchant to hardcode a specific device type. @girasquid do you have any ideas?

Please review @girasquid @bizla

### DIFF
--- a/test/unit/gateways/authorize_net_test.rb
+++ b/test/unit/gateways/authorize_net_test.rb
@@ -69,7 +69,7 @@ class AuthorizeNetTest < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
-  def test_retail_market_type_included_in_swipe_transactions_with_valid_track_data
+  def test_retail_market_type_device_type_included_in_swipe_transactions_with_valid_track_data
     [BAD_TRACK_DATA, nil].each do |track|
       @credit_card.track_data = track
       stub_comms do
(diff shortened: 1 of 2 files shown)


bizla:
Can you provide some context as to why this has to change? A test for this expectation would be appreciated -- even if it's just an extra assertion somewhere in Auth.Net card present tests.

girasquid:
@abecevello you could make it an option that gets passed in, and then fall down to the default if the option isn't provided.

abecevello:
Updated to have tests and allow the default to be overwritten using the options hash.

@bizla This is the recommended device type for Shopify POS from Authorize.net.

test/unit/gateways/authorize_net_test.rb
```diff
@@ -89,6 +89,22 @@ def test_retail_market_type_included_in_swipe_transactions_with_valid_track_data
         parse(data) do |doc|
           assert_not_nil doc.at_xpath('//retail')
           assert_equal "2", doc.at_xpath('//retail/marketType').content
+          assert_equal "7", doc.at_xpath('//retail/deviceType').content
+        end
+      end.respond_with(successful_purchase_response)
+    end
+  end
+
+  def test_device_type_used_from_options_if_included_with_valid_track_data
```
girasquid: :heart_eyes: for adding this test too!

girasquid:
:+1:

bizla:
:pizza: Merging.

bizla:
Closed via 2a466b86aea8730b6499df88b56f3cc01f27ffa6

PR #1787, opened 2015-07-15
Akamai URL update

taf2:
Authorize.NET sent out email to notify customers that they are switching to Akamai and that the new URL changes are api2 instead of api


andrewpaliga:
:+1:

girasquid:
Thanks!

PR #1788, opened 2015-07-15
Add decrypted apple_pay to litle

anellis:


anellis:
@duff would you mind taking a look at this apple pay addition?

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -196,6 +196,9 @@ def add_payment_method(doc, payment_method)
             doc.expDate(exp_date(payment_method))
             doc.cardValidationNum(payment_method.verification_value)
           end
+          doc.cardholderAuthentication do
```
rwdaigle: Should we wrap the whole `cardholderAuthentication` block in the if statement so it doesn't even appear unless we have a cryptogram? Otherwise there will always be an empty element.

rwdaigle:
@duff we should be sure to note the associated API version bump in the changelog in case there are unexpected upgrade issues.

rwdaigle:
@anellis Seeing this test error from Travis?

```
================================================================================
Error: test_failed_authorize(LitleTest): NoMethodError: undefined method `payment_cryptogram' for #<ActiveMerchant::Billing::CreditCard:0x000000058f4828>
/home/travis/.rvm/gems/ruby-2.2.0/gems/activesupport-3.2.22/lib/active_support/core_ext/object/try.rb:36:in `try'
/home/travis/build/activemerchant/active_merchant/lib/active_merchant/billing/gateways/litle.rb:200:in `block in add_payment_method'
```

Other `NetworkTokenizationCreditCard` implementations test for type [like this](https://github.com/activemerchant/active_merchant/blob/dc705d60decc12e9c942f07c4937ecd09dbee69d/lib/active_merchant/billing/gateways/stripe.rb#L324-L329)?

anellis:
@rwdaigle yeah I was discussing that in the dev slack since it's weird that it only failed some of them. Apparently there was a bug with try in Rails 3.2. @duff updated!

rwdaigle:
@anellis I don't think we should require the user to pass in `order_source: "applepay"` in the tx options. Instead, if the payment method is a `NetworkTokenizationCreditCard`, we should automatically set that value [within the adapter](https://github.com/anellis/active_merchant/blob/0904f45c58c3b3ae906d1fdf28d3609098b17503/lib/active_merchant/billing/gateways/litle.rb#L239-L247)?

anellis:
@rwdaigle you're right! That makes much more sense. Updated.

rwdaigle:
@duff this gets my :+1:

duff:
@anellis - Looks great. :+1: 

@rwdaigle Thanks for the help!

PR #1790, opened 2015-07-16
[Bugfix] Stripe: Allow purchases with tokens without customer specification

bizla:
This is a small fix to the changes introduced in https://github.com/activemerchant/active_merchant/commit/808a142691a3f452b07f275aeb1f96e4efe607cd, whereby if the payment was specified by a flat string without a `|`, it would be interpreted as the customer ID rather than the card token. 

This simply restores the previous default but includes the condition from https://github.com/activemerchant/active_merchant/commit/808a142691a3f452b07f275aeb1f96e4efe607cd.

@duff @aprofeit for review please.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -17,6 +17,7 @@
 * Authorize.Net: Allow passing device type through options, make wireless POS the default [abecevello]
 * Authorize.Net: Update to new Akamai URL [taf2]
 * Braintree: Add hold_in_escrow [anellis]
+* Stripe: Allow purchases with tokens without customer specification [bizla]
 
 == Version 1.51.0 (July 2, 2015)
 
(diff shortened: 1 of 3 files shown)


aprofeit:
Any way to add a test that would have caught this?

bizla:
Looks like the original case was overwritten: https://github.com/activemerchant/active_merchant/commit/808a142691a3f452b07f275aeb1f96e4efe607cd#diff-b6ba33f7d7a676c8e0a4911fb561d9f6R490
Remote testing with both cases works though (as it did before) -- I'll add the original test back.

bizla:
Restored the card-token-only tests, all pass locally. I'll merge on your blessings. :rose:

aprofeit:
:+1:

duff:
@bizla Thanks for the fix!

bizla:
:grinning:

PR #1804, opened 2015-07-23
Add scrubbing to blue pay

anellis:
And remove strange windows carriage returns in blue_pay files


anellis:
@duff take two! All weird carriage returns are removed so hopefully that fixes it

PR #1805, opened 2015-07-24
QuickPay v10: support 2-letter country codes

girasquid:
QuickPay v10 requires 3-letter country codes, but [the documentation on gateway.rb calls for 2-letter](https://github.com/activemerchant/active_merchant/blob/18987d4d599cf3fd7a7e781399d561d110150a5a/lib/active_merchant/billing/gateway.rb#L47); this removes a bunch of extra whitespace and also adds a mapping for that so the interface is consistent with the other gateways.

@aprofeit @ntalbott for review - open to suggestions on more tests to add.

I also left this potentially raising an exception for a country code being passed in that can't be found, vs. [what jetpay does](https://github.com/activemerchant/active_merchant/blob/18987d4d599cf3fd7a7e781399d561d110150a5a/lib/active_merchant/billing/gateways/jetpay.rb#L279) - thoughts?

### DIFF
--- a/test/remote/gateways/remote_quickpay_v10_test.rb
+++ b/test/remote/gateways/remote_quickpay_v10_test.rb
@@ -1,7 +1,7 @@
 require 'test_helper'
 
 class RemoteQuickPayV10Test < Test::Unit::TestCase
-  
+
   def setup
     @gateway = QuickpayGateway.new(fixtures(:quickpay_v10_api_key))
     @amount = 100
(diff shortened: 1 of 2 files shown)


test/remote/gateways/remote_quickpay_v10_test.rb
```diff
@@ -9,34 +9,44 @@ def setup
...
   end
-  
+
   def card_brand response
     response.params['metadata']['brand']
   end
-  
+
+  def test_successful_purchase_with_short_country
+    options = @options.merge({billing_address: address(country: 'CA')})
```
j-mutter: This feels weird. `DK` maybe?

girasquid: Changed

aprofeit:
I'm fine with this, I think it's better to raise if it can't find rather than fail silently.

j-mutter:
Agreed. :+1:

PR #1806, opened 2015-07-27
Gateway CenPOS: parse cvs_result and avs_result from response

tjstankus:


lib/active_merchant/billing/gateways/cenpos.rb
```diff
@@ -166,7 +166,9 @@ def commit(action, post)
           raw,
           authorization: authorization_from(post, raw),
           error_code: error_code_from(succeeded, raw),
-          test: test?
+          test: test?,
+          cvv_result: @cvv_result,
```
duff: @tjstankus Might it be possible to avoid using the instance vars?

tjstankus:
@duff Agreed that setting instance vars in the parse method is side-effect'y and not
ideal. Ideally, I'd simply add the CVV and AVS response parsing to the existing
parse code. However, the ParameterValidationResultList that needs to be parsed
to get the CVV and AVS results is more deeply nested than the current xml
parsing code takes into consideration. Adding it would involve rewriting the
parsing code or injecting new behavior via nested conditionals; both approaches
seem error-prone and undesirable. 

So, I chose to split the CVV and AVS result parsing out separately into code
that's not dependent on nodes being only 2 deep. Of course the XML data is
required to parse it and the `parse` method has that data. But, it's pretty
simple to cache that data in a local variable and pass that to a new method that
returns the validation results - so that's what I did. :)

Does this approach seem preferable to you?

duff:
For some reason a bunch of remote tests seem to be failing for me.  Not sure if I'm having personal problems or not.  I'm seeing this:

```
22 tests, 37 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
45.4545% passed
```

tjstankus:
@duff Remote CenPOS tests are all passing for me both before and after rebasing with latest master. Perhaps it's a credentials issue? I'll send the creds I'm using separately.

duff:
Weird!  They're all passing for me now.  I must have been having personal issues.

duff:
@tjstankus :+1: 

I added just a few little tweaks here: 824d1a34bea6939f65011423cb516c646a43fc48

PR #1807, opened 2015-07-27
Stripe: Add validate:false field to store

anellis:


anellis:
@duff Adding the validate:false field

duff:
@anellis Looks like a unit test is failing

duff:
@anellis Looks good.  Only thing I added was a unit test to ensure the param is getting passed.

PR #1808, opened 2015-07-27
Forte: Add gateway

davidsantoso:
@duff adding the Forte gateway. It has a few quirks, so let me know if you have any thoughts on these:

1. All failures respond with a 4xx HTTP response code, so I did an override of the `handle_response` method originally in `posts_data.rb`. I'm not sure if this was the best way to go, but it seemed as though errors like a declined card shouldn't raise an `ActiveMerchant::ResponseError`? It didn't seem like a response error to me and more of a "processing error"?

2. If you run the tests, unfortunately sometimes you may get a `401 Unauthorized` on some requests. But then other times it'll pass. While testing, I noticed that adding a `sleep(5)` helped with the remote tests passing more often so that's why those are there. Really meh but there's something strange going on with their API that most of the time you'll get failures unless you wait a few seconds between some API calls- like an authorize and capture.

3. There's no refund. Just credit which requires the credit card to be submitted in the request.

4. I symbolized the response keys, but I'm not sure if that's generally not good practice. Not a quirk, but something I did that I thought I'd see if it was not really inline with how most AM adapters work.

### DIFF
--- a/lib/active_merchant/billing/gateways/forte.rb
+++ b/lib/active_merchant/billing/gateways/forte.rb
@@ -1,0 +1,245 @@
+require 'json'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class ForteGateway < Gateway
+      self.test_url = 'https://sandbox.forte.net/api/v2'
+      self.live_url = 'https://api.forte.net/v2'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'https://www.forte.net'
+      self.display_name = 'Forte'
+
+      def initialize(options={})
+        requires!(options, :api_key, :secret, :location_id, :account_id)
+        super
+      end
+
+      def purchase(money, payment_method, options={})
+        post = {}
+        add_amount(post, money, options)
+        add_payment_method(post, payment_method)
+        add_billing_address(post, options)
+        add_shipping_address(post, options)
+        post[:action] = "sale"
+
+        commit(:post, post)
+      end
+
+      def authorize(money, payment_method, options={})
+        post = {}
+        add_amount(post, money, options)
+        add_payment_method(post, payment_method)
+        add_billing_address(post, options)
+        add_shipping_address(post, options)
+        post[:action] = "authorize"
+
+        commit(:post, post)
+      end
+
+      def capture(money, authorization, options={})
+        post = {}
+        post[:transaction_id] = transaction_id_from(authorization)
+        post[:authorization_code] = authorization_code_from(authorization)
+        post[:action] = "capture"
+
+        commit(:put, post)
+      end
+
+      def credit(money, payment_method, options={})
+        post = {}
+        add_amount(post, money, options)
+        add_payment_method(post, payment_method)
+        add_billing_address(post, options)
+        post[:action] = "disburse"
+
+        commit(:post, post)
+      end
+
+      def void(authorization, options={})
+        post = {}
+        post[:transaction_id] = transaction_id_from(authorization)
+        post[:authorization_code] = authorization_code_from(authorization)
+        post[:action] = "void"
+
+        commit(:put, post)
+      end
+
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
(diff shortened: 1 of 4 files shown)


davidsantoso:
@duff just a bump to see if you've had a chance to look at this today?

test/remote/gateways/remote_forte_test.rb
```diff
@@ -0,0 +1,150 @@
...
+    assert_equal 'INVALID CREDIT CARD NUMBER', response.message
+  end
+
+  def test_successful_authorize_and_capture
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    sleep(7)
+
+    assert capture = @gateway.capture(@amount, auth.authorization, { transaction_id: auth.params["transaction_id"] })
```
duff: Looks like `transaction_id` is required?  If so, perhaps the authorization's Response.authorization should be a multi-part string.

lib/active_merchant/billing/gateways/forte.rb
```diff
@@ -0,0 +1,246 @@
...
+          avs_result: AVSResult.new(code: response[:response][:avs_result]),
+          cvv_result: CVVResult.new(response[:response][:cvv_code]),
+          test: test?,
+          error_code: error_code_from(response)
+        )
+      end
+
+      def handle_resp(response)
+        case response.code.to_i
+        when 200..400
```
duff: I would think you'd want `200..499` to get things like their funky 401 `This user does not have permission to access one or more of the requested Ids/Tokens.`

duff: And it'll also get this test, which I think ought to be added to the remote tests:

```
  def test_invalid_login
    gateway = ForteGateway.new(api_key: "InvalidKey", secret: "InvalidSecret", location_id: "11", account_id: "323")
    assert response = gateway.purchase(@amount, @credit_card, @options)
    assert_failure response
    assert_equal "UserName and Password combination not found.", response.message
  end
```

test/remote/gateways/remote_forte_test.rb
davidsantoso: Ah ok. Is there a convention used in AM? Like separate by `:` or spaces?

lib/active_merchant/billing/gateways/forte.rb
davidsantoso: :+1: :+1:

test/remote/gateways/remote_forte_test.rb
duff: I'd recommend searching for references to `authorization_from`.  You'll see many gateways where we're joining and splitting.  Not sure what the overall consensus is.  Example:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L731

duff:
I spotted your question about symbolizing the keys.  I'm not seeing many places where we're doing it.  I'm inclined to not do it and just leave 'em as strings.  Less code :+1:

duff:
I've also contacted Forte about our needing the `sleep` calls in remote tests.  That's pretty funky! :smile:

davidsantoso:
Ha right? I don't know what the deal is...

And, I'll remove the key symbolization. I've done it before with other gateway/seen a few other gateways do it so I wasn't sure if it was something we were trying to do by convention with AM adapters. Happy to have less code :wink:

davidsantoso:
@duff alrighty, changes have been made. Let me know if you have any other thoughts.

duff:
@davidsantoso Looks like some failing Travis tests

davidsantoso:
Forgot to update unit test. Should pass this time...

davidsantoso:
@duff all green!

duff:
Do they allow an eCheck `authorization`?  If so, a remote test for that would be cool.  Also a remote test for a failed eCheck purchase would be :egg:cellent.  Also thinking some eCheck unit tests would be helpful.

davidsantoso:
@duff docs are pretty vague around echeck support. I'll write some remote tests and see what happens. You can send a `card` or `echeck` in with the `transaction` so it's entirely possible!

duff:
I'm retracting my question about echeck authorizations.  I saw them in the Authorize.net implementation, but the more I think about it, the more I'm wondering about what their actual value is.

davidsantoso:
@duff just pushed an update to add more tests for echecks!

PR #1809, opened 2015-07-28
QuickPay: use the value of the alpha3 country code, instead of trying to include the full object

girasquid:
@aprofeit @ivanfer 

Open to suggestions on how best to test this - as far as I can tell the remote test that I added in [the previous PR](https://github.com/activemerchant/active_merchant/pull/1805) shouldn't have been passing, because when I dumped the request the country code was being passed in as a string representation of the entire [CountryCode object](https://github.com/activemerchant/active_merchant/blob/fdace518c3e30ee9e248219b926f2cf26f2ed4f5/lib/active_merchant/country.rb#L10).

### DIFF
--- a/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
+++ b/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
@@ -201,7 +201,7 @@ module ActiveMerchant
             :city         => address[:city],
             :region       => address[:address2],
             :zip_code     => address[:zip],
-            :country_code => country.code(:alpha3)
+            :country_code => country.code(:alpha3).value
           }
           mapped
         end


aprofeit:
What do you mean when you dumped the request? Is it possible it's implicitly calling `value` on that object when it gets sent?

girasquid:
It doesn't appear to be - this is the output of what was POST'd to QuickPay:
```<- "{\"currency\":\"CAD\",\"order_id\":\"c1.10\",\"invoice_address\":{\"name\":\"Luke Hutscal\",\"street\":\"22 Murray Street\",\"city\":\"Ottawa\",\"region\":\"Apartment 212\",\"zip_code\":\"K1N1G9\",\"country_code\":{\"value\":\"DEU\",\"format\":\"alpha3\"}},\"metadata\":{\"manual_entry\":false}}"```

You can see the `country_code` key has the entire `CountryCode` object, instead of just the string "DEU".

aprofeit:
Looks like it called `to_json` on it. Very strange the test still works :\

aprofeit:
In any case, change makes sense :+1:

PR #1811, opened 2015-07-29
USA ePay advanced: added quickUpdateCustomer API call

kimikaza:
Added a missing call to "quickUpdateCustomer" on USA ePay Advanced gateway, since it was missing in original implementation of usa_epay_advanced.rb. Unit/remote tests added to corresponding files.

### DIFF
--- a/lib/active_merchant/billing/gateways/usa_epay_advanced.rb
+++ b/lib/active_merchant/billing/gateways/usa_epay_advanced.rb
@@ -348,27 +348,69 @@ module ActiveMerchant #:nodoc:
       #  * Same as add_customer
       #
       def update_customer(options={})
+        ActiveMerchant.deprecated "update_customer is deprecated and will be removed from a future release of ActiveMerchant. Please use update instead."
         requires! options, :customer_number
 
         request = build_request(__method__, options)
         commit(__method__, request)
       end
 
-      # Update a customer by changing only a few fields.
+      # Update a customer by replacing either all of the customer details or changing only a few fields
+      #
       #
       # ==== Required
       # * <tt>:customer_number</tt> -- customer to update
       #
+      # 
+      # To update only a few fields use this set of options
+      #
       # ==== Options
       # * <tt>:fields</tt> -- array of 2 element long arrays representing FieldValue object
       #   * <tt>:Field</tt> -- name of the field
-      #   * <tt>:Value</tt> -- field value
+      #   * <tt>:Value</tt> -- field value 
       #
-      def quick_update_customer(options={})
-        requires! options, :customer_number
+      # To replace customer details use this set of options
+      #
+      # ==== Options
+      # * <tt>:id</tt> -- merchant assigned id
+      # * <tt>:notes</tt> -- notes about customer
+      # * <tt>:data</tt> -- base64 data about customer
+      # * <tt>:url</tt> -- customer website
+      # * <tt>:billing_address</tt> -- usual options
+      # * <tt>:payment_methods</tt> -- array of payment method hashes.
+      #   * <tt>:method</tt> -- credit_card or check
+      #   * <tt>:name</tt> -- optional name/label for the method
+      #   * <tt>:sort</tt> -- optional integer value specifying the backup sort order, 0 is default
+      #
+      # ==== Recurring Options
+      # * <tt>:enabled</tt> -- +true+ enables recurring
+      # * <tt>:schedule</tt> -- daily, weekly, bi-weekly (every two weeks), monthly, bi-monthly (every two months), quarterly, bi-annually (every six months), annually, first of month, last day of month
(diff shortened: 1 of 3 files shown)


ntalbott:
@kimikaza current ActiveMerchant policy is to conform to a limited set of public API's; in this case, `update_customer` should be renamed to just `update`, and the `quick_update_customer` functionality should be encapsulated in that method as well if desired. So while I'm OK conceptually with supporting this functionality as part of `update`, this won't be merged as-is since it adds a non-conforming public API method.

kimikaza:
@ntalbott thanks for the info. Is it ok, though, to make it into one "update" method since usa-epay's updateCustomer accepts a customer object, and quickUpdateCustomer accepts an array of FieldValue objects instead? Also if I rename the old update_customer into "update" we'd loose backwards compatibility. Sorry, I'm not sure I understood correctly.

ntalbott:
@kimikaza to answer your second question first: we would need to properly deprecate `update_customer`, throwing a message that warns it will be removed.

As to having a single `update` call, the goal is to match the standard ActiveMerchant API, which probably looks more like `quickUpdateCustomer`. The thing is AM doesn't usually support the full range of possibilities with tokenization API's, since AM focuses on just storing/managing cards, and does its best to hide any necessary customer management to that end. For those that want full, robust integration with customer management API's, we recommend going directly against the gateway and not using AM.

kimikaza:
@ntalbott 
As you explained to me, I changed the "update_customer" so it is deprecated, and I encapsulated both "update_customer" and "quick_update_customer" into one "update" method. I changed the remote/unit tests to support this now.
I hope I understood you correctly and this complies with AM policies now.

lib/active_merchant/billing/gateways/usa_epay_advanced.rb
```diff
@@ -348,12 +348,71 @@ def add_customer(options={})
...
+      #
+      # ==== Point of Sale Options
+      # * <tt>:price_tier</tt> -- name of customer price tier
+      # * <tt>:tax_class</tt> -- tax class
+      # * <tt>:lookup_code</tt> -- lookup code from customer/member id card; barcode or magnetic stripe; can be assigned by merchant; defaults to system assigned if blank
+      #
+      # ==== Response
+      # * <tt>#message</tt> -- customer number assigned by gateway
+
+      def update(options={})
```
ntalbott: Per AM's standard API, this should be `update(identifier, options={})`. I'm guessing `identifier` would be `customer_number`?

ntalbott:
@kimikaza this looks like a big improvement. I've added a note RE the API.

kimikaza:
@ntalbott Thanks. This gateway was programmed this way when I arrived at it so I just concluded to leave it that way. Quick update would be called like this: 
  GATEWAY.update(customer_number: some_identifier, fields: [['Email', 'newemail@example.com'],['Address','newaddress']])

Do you want me to extract identifier as a separate argument, or can I just leave this as it is?

Oh, and I'm a non-native english speaker, what does RE mean?

ntalbott:
RE is "regarding", sorry for that.

Yes, the customer number should be pulled out into another argument. Do a quick grep of the codebase for `def update(` to see other examples of the signature for that method.

kimikaza:
@ntalbott I've got this now changed so update method looks like this: def update(identifier, options={})
Tests updated too, everything is tested and working on my project so I hope this is my final post here :)
Thanks for helping me.

kimikaza:
@ntalbott Do I need to do anything more to get this pull request accepted?

kimikaza:
@ntalbott I just ping to see what is the status of this PR and is anything more expected of me regarding this? I hope I'm not sounding too eager.

ntalbott:
@kimikaza this still isn't conforming to `update` expectations, but that's mostly our fault for not clearly laying out what those are. I've tried to lay those out more clearly here: https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards#storeunstoreupdate

Basically this needs to be massively simplified, and/or if you want the more complicated version, that's fine but it's not a fit for ActiveMerchant. Curious to hear your thoughts.

kimikaza:
@ntalbott I actually did those changes because I was dropped into a legacy code that used activemerchant gem for USAePAY recurring payment. I thought at first that USAePAY simply omited some of their methods, mostly for recurring payment.
Anyway, I need things like customer history, list of transactions due in recurring payment etc. and I realized that it probably doesn't fit activemerchant philosophy at all, so I continued working on my fork of the library, and I don't have resources to make my own gem so I'll just keep it as it is on my profile for now.

The only thing that I can contribute with is that I actually found a bug in existing add_customer method and I'm thinking of making pull request for that, as that method is already a part of activemerchant so I could, as well, fix a bug that's in it.

I don't think any more that this branch should enter activemerchant.
Thanks for your help!

ntalbott:
@kimikaza thanks for the follow-up!

PR #1812, opened 2015-07-29
USAePay: Add test mode setting via options

marquisong:
This PR adds the ability to set test_mode flag via options for USAePay Transaction. As a result, it allows the passing of test mode flag over to USAePay for the purpose of simulating a transaction without actually processing the transaction.
Documentation on this field(UMtestmode) can be found at https://wiki.usaepay.com/merchant/epaymentform, under Developer Reference > Form Fields.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -37,6 +37,7 @@
 * Braintree: Update country list [duff]
 * NMI: Don't include dup_seconds if nil [rwdaigle]
 * QuickPay: Make all operations to v10 platform synchronous [ta]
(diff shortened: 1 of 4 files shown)


ntalbott:
@marquisong what's the advantage of this instead of just using the already supported `test: true` option at initialization?

ntalbott:
Ah, I think I get it - this lets setting testmode to true in production. Seems reasonable, I'll work to get it merged.

marquisong:
@ntalbott That is correct. Thanks for looking into the PRs.

PR #1813, opened 2015-07-29
non-reference credit support for Network Merchants

barooo:
This adds a #credit method to the NetworkMerchantsGateway for non-reference credits.  Since the API didn't previously have a #credit method I didn't add a deprecation warning for reference credits (refunds).

I Verified this works against my merchant account, and verified that the remote test still passes after my addition.


ntalbott:
Implemented in https://github.com/activemerchant/active_merchant/commit/b261824f450f3cf845bd90715424bf83b182b1f2.

PR #1815, opened 2015-07-30
QuickPay: strip # from Order IDs

girasquid:
@aprofeit @ivanfer 

Discovered while testing that QuickPay doesn't support "#" in order IDs.


aprofeit:
:+1:

PR #1816, opened 2015-07-30
Update paypal api version

anellis:


anellis:
@duff here is the updated paypal api version. I ran all of the remote and unit tests for paypal (remote and unit paypal_express, paypal_common_api, paypal_digital_goods, and the regular remote and unit paypal_test) and there were no new issues.

I did however make a small change in the remote_paypal_test to make the new order id unique because otherwise it says you've already captured funds for that id. I did check before and after I updated the version though and that was NOT caused by the new version. Even in the old version the test didn't pass.

PR #1817, opened 2015-07-30
QuickPay: fix method signature on #void

girasquid:
@aprofeit @ivanfer 

The QuickPay v10 integration's `void` method doesn't accept `options`, which all other integrations do - this brings it into sync with the others.

### DIFF
--- a/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
+++ b/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
@@ -35,7 +35,7 @@ module ActiveMerchant
         end
       end
 
-      def void(identification)
+      def void(identification, _options = {})
         commit(synchronized_path "/payments/#{identification}/cancel")
       end
 


lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
```diff
@@ -35,7 +35,7 @@ def authorize(money, credit_card, options = {})
         end
       end
 
-      def void(identification)
+      def void(identification, options = {})
```
aprofeit: Since it's not actually passed to anything, should it be `_` or `_options`?

girasquid: Fair point; changed it to `_options`

aprofeit:
:+1:

aprofeit:
Wonder if we can tests to ensure all the gateways have the same method signatures.

PR #1818, opened 2015-07-30
Stripe: return refund id for refund authorization

anellis:


anellis:
@duff this returns the correct refund id for authorization

PR #1820, opened 2015-07-31
TNS: Translate countries to alpha3 codes

anellis:


anellis:
@duff this translates the country codes to alpha 3 and handles 400 errors better.

duff:
@anellis This is looking great.  Only thing I was thinking about was having a test for the 400 handling.  My initial look here:

https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_tns_test.rb#L78

makes me think that the handling of that 401 is different than how we're handling other gateways.  It doesn't look like we typically raise `ActiveMerchant::ResponseError` for these types of things.   Looks like most `test_invalid_login` remote tests do things differently than we're doing for TNS.

The exploration made me think that perhaps the country stuff could be one commit and the 400 handling could be another commit / pr.  So perhaps this one could be modified to be just the country stuff and then another one for the 4xx?

lib/active_merchant/billing/gateways/tns.rb
```diff
@@ -150,7 +157,12 @@ def commit(action, post)
           'Content-Type' => 'application/json',
         }
         post[:apiOperation] = action.upcase
-        raw = parse(ssl_request(:put, url, build_request(post), headers))
+        begin
+          raw = parse(ssl_request(:put, url, build_request(post), headers))
+        rescue ResponseError => e
+          raise unless(e.response.code.to_s =~ /400/)
```
duff: In looking at the `test_invalid_login` remote test... if that gets modified to no longer raise the exception, then I think this line won't be needed.

anellis:
@duff okay! I was thinking that but then I always get nervous about changing documented expected behavior but I suppose no one gets mad if you stop exceptions and give error messages instead :)

This commit is now just the country stuff and I'll follow up with another pr for the 400 stuff

duff:
@anellis :+1: 

I just added one little commit 517eaa4fade41e0f90f93b8f273a48d4b9260e63

PR #1823, opened 2015-07-31
TNS: Rescue Errors

anellis:


anellis:
@duff so then this is just the 4xx stuff by itself! I figure that the the login test now acts as a test for exception handling as well. If you'd like me to add another one I can, I just couldn't figure out how to trigger another exception now that I fixed the country code thing. I'm sure I can keep playing with it and break it somehow though ;)

duff:
Looks great to me!

PR #1826, opened 2015-08-05
Elavon: Supports capture via CCComplete

marquisong:
This PR adds the support to capture payments using CCCOMPLETE when no credit card details being provided in the options.
Currently, in order to capture a payment, the credit card information is required to be supplied in the options.
As a result of this update, if no credit card information being supplied in the capture request, it will use the CCCOMPLETE transaction type instead. I have added and updated all relevant unit and remote tests.

Further details can be found in https://www.myvirtualmerchant.com/VirtualMerchant/download/developerGuide.pdf (Under Credit Card Completion (cccomplete) section).


ntalbott:
This is an *awesome* improvement - thanks!

PR #1829, opened 2015-08-05
NMI gateway (non-Auth.net mode) support

rwdaigle:
This is a rewrite of the NMI adapter to support their native API and not
the previous Auth.net emulation mode API. Every attempt has been made to
retain backwards compatibility. However, there may be small discrepancies
depending on your usage of the API.

What is known is that there are a few fields without a sensible mapping
to the new API. These request params include:

* cardholder_authentication_value and authentication_indicator which are
3D Secure authentication fields. The NMI adapter doesn't yet support 3DS.
* solution_id (application_id), which identifies the agent making the API
call, has no equal in the native API.
* cust_id has no mapping to the native API

Also, the native API does not support the notion of a fraud_review, so that
functionality is not available in the rewritten implementation.

The advantages of using the NMI native API include being on the most
current version which will make future development easier, as well as
support for features such as ACH/echeck payment methods (which this
rewrite does include).

@ntalbott does this pass your sniff test?

### DIFF
--- a/test/unit/gateways/nmi_test.rb
+++ b/test/unit/gateways/nmi_test.rb
@@ -1,237 +1,326 @@
 require 'test_helper'
-require 'unit/gateways/authorize_net_test'
 
 class NmiTest < Test::Unit::TestCase
   include CommStub
 
   def setup
-    @gateway = NmiGateway.new(:login => 'X', :password => 'Y')
+    @gateway = NmiGateway.new(fixtures(:nmi))
 
     @amount = 100
     @credit_card = credit_card
+    @check = check
     @options = {}
   end
 
-  def test_credit_card_purchase_no_recurring_flag
+  def test_successful_purchase
     response = stub_comms do
       @gateway.purchase(@amount, @credit_card)
     end.check_request do |endpoint, data, headers|
-      assert_no_match(/x_recurring_billing/, data)
+      assert_match(/username=#{@gateway.options[:login]}/, data)
+      assert_match(/password=#{@gateway.options[:password]}/, data)
+      assert_match(/type=sale/, data)
+      assert_match(/amount=1.00/, data)
+      assert_match(/payment=creditcard/, data)
+      assert_match(/ccnumber=#{@credit_card.number}/, data)
+      assert_match(/cvv=#{@credit_card.verification_value}/, data)
+      assert_match(/ccexp=#{sprintf("%.2i", @credit_card.month)}#{@credit_card.year.to_s[-2..-1]}/, data)
     end.respond_with(successful_purchase_response)
 
     assert_success response
+    assert response.test?
+    assert_equal "2762757839", response.authorization
   end
 
-  def test_credit_card_purchase_with_recurring_flag
+  def test_purchase_with_options
     response = stub_comms do
-      @gateway.purchase(@amount, @credit_card, :recurring => true)
+      @gateway.purchase(@amount, @credit_card,
+        :recurring => true, :order_id => "#1001", :description => "AM test",
+        :currency => "GBP", :dup_seconds => 15, :customer => "123")
     end.check_request do |endpoint, data, headers|
-      assert_match(/x_recurring_billing=TRUE/, data)
+      assert_match(/billing_method=recurring/, data)
+      assert_match(/orderid=#{CGI.escape("#1001")}/, data)
+      assert_match(/orderdescription=AM\+test/, data)
+      assert_match(/currency=GBP/, data)
+      assert_match(/dup_seconds=15/, data)
+      assert_match(/customer_id=123/, data)
     end.respond_with(successful_purchase_response)
 
     assert_success response
   end
 
-  def test_successful_authorization
-    @gateway.expects(:ssl_post).returns(successful_authorization_response)
+  def test_failed_purchase
+    response = stub_comms do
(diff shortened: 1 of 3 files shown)


ntalbott:
@rwdaigle only question I have: is `Response#authorization` handling backwards compatible?

rwdaigle:
@ntalbott

> is `Response#authorization` handling backwards compatible?

It is simply the transaction id in both API responses, so there's definitely no field ambiguity.

ntalbott:
@rwdaigle great!

Looks like NMI replied with mappings for most of the fields you weren't able to map previously? Are you going to apply those?

rwdaigle:
@ntalbott I'm still waiting on some more details regarding the 3DS fields. There is no `solution_id` equivalent. And I did make the `customer_id` field mapping.

I'll post back here when I've reached a conclusion about the 3DS stuff.

ntalbott:
Yup, pretty sure `solution_id` was just a copy/paste from when NMI was split from Auth.Net: https://github.com/activemerchant/active_merchant/commit/a6cfae5e77ed249de84dbe46a4a91e3a0903a11d#diff-570504655ed8be17318b9c2f7080bb5eR3

rwdaigle:
@ntalbott the NMI 3DS story is a bit muddled, and there's certainly no mapping from the auth.net to native APIs. From NMI:

> With 3DS there are two options:
> * We have our own Payer Authentication service. In order to work, the merchant has to use the three-step redirect or Quickclick (a hosted checkout page) so that we have control over the customer's browser. We then re-direct the customer to the service and back.
> * Some merchants or software providers will integrate into CardinalCommerce Centinel on their own. We have the variables there to pass them through, but the instructions on how to set that up would come from the 3D Secure provider. I do know if you run a sale with the 3D Secure variables through the gateway, and use the Query API on that transaction you'll get a "cavv_result". If that result is blank or null, that means the sale was run on an unsupported processor. 

The current version of the adapter only passes two variables (`cardholder_authentication_value` and `authentication_indicator`) whereas NMI's native 3DS support has four:

> These are related to 3D Secure (Verified By Visa, Mastercard SecureCode). I'm not sure how these map one to one, but these are the variables for the direct post:
> * cardholder_auth: Set 3D Secure condition. Values: 'verified' or 'attempted'
> * eci: E-commerce indicator.
> * cavv: Cardholder authentication verification value.
> * xid: Cardholder authentication transaction id.

It seems that there is no way to migrate these fields. Should we move forward as is, @ntalbott?

ntalbott:
@rwdaigle I think those fields are only on the NMI adapter because it was copied from the Authorize.Net adapter when the latter went XML. So I'd say we just drop them with a "POSSIBLE BREAKAGE" changelog note.

rwdaigle:
@ntalbott :+1: 

You write the changelog entry, not me, right?

duff:
@rwdaigle I see a few warnings when running `rake`.

test/unit/gateways/nmi_test.rb
```diff
@@ -243,141 +332,105 @@ def test_supported_card_types
...
-  def already_actioned_capture_response
-    '$1$,$2$,$311$,$This transaction has already been captured.$,$$,$P$,$0$,$$,$$,$1.00$,$CC$,$credit$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$39265D8BA0CDD4F045B5F4129B2AAA01$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$'
+  def successful_echeck_store_response
+    'response=1&responsetext=Customer Added&authcode=&transactionid=&avsresponse=&cvvresponse=&orderid=35b5500a13d23a7e9706fdf3518556b3&type=&response_code=100&customer_vault_id=1910603011'
   end
 
-  def fraud_review_response
-    "$4$,$$,$253$,$Thank you! For security reasons your order is currently being reviewed.$,$$,$X$,$0$,$$,$$,$1.00$,$$,$auth_capture$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$207BCBBF78E85CF174C87AE286B472D2$,$M$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$,$$"
+  def transcript
+    <<-TRANSCRIPT
```
duff: I think using `%q` is a good habit for transcript scrubbing tests, even if it doesn't matter for transcripts like NMI's.

A normal heredoc has troubles for some transcripts (like https://github.com/activemerchant/active_merchant/commit/0fd556480f1e3fdd547a7b0779834863e9f60504 and https://github.com/activemerchant/active_merchant/commit/47348e7bc003477dd24d491d1ef8aa174e504f85 )

rwdaigle:
@duff fixed

duff:
Closed by b261824f450f3cf845bd90715424bf83b182b1f2

PR #1830, opened 2015-08-06
added payu latam for colombia gateway

damiangalindo:


damiangalindo:
This is to add the functionality for PayU Latam for Colombia, but also gives the opportunity to add the other Latinamerican Gateways for PayU

lib/active_merchant/billing/gateways/payu_co.rb
```diff
@@ -0,0 +1,18 @@
+module ActiveMerchant #:nodoc:
```
ntalbott: This doesn't seem to add value, since it's exactly the same as the values in PayuLatamGateway. Further, what is the difference besides currencies and countries between PayU Colombia and PayU Latam in general?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,311 @@
...
+
+        options.merge!({type: "REFUND"})
+
+        add_transaction(post, options)
+        
+        commit("refund", post)
+      end
+
+      # options = {order_id: "order_id", reason: "reson for the cancelation", transaction_id: "id of the transaction"}
+      def void(options={})
```
ntalbott: `void` in ActiveMerchant is intended to reverse an `authorize`, so this doesn't seem useful without `authorize` support.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,311 @@
...
+            orderId: order_id
+          }
+        }
+
+        add_merchant(post)
+
+        commit("info", post)
+      end
+
+      def verify(credit_card, options={})
```
ntalbott: I don't think this will work since there is no `authorize` support?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,311 @@
...
+
+        options.merge!({type: "VOID"})
+
+        add_transaction(post, options)
+        
+        commit("void", post)
+      end
+
+      #id for the order, integer value
+      def order_status(order_id)
```
ntalbott: This is outside the scope of ActiveMerchant's public API. Please remove.

ntalbott:
This is going to need some work before merging; see comments above. I'd love to see PayU Latam supported, though, so I'll try to stay on top of this!

juankhazbun:
PayU Latam has support for Argentina, Colombia, México, Panamá, Perú and Brasil. Each of them have different currrency, so the idea of the PayuLatamGateway class is provide an easy way to extend the base functionality to the other supported countries.

Currently PayU Brasil is the only one that supports Authorize and Capture separately, as you can see in [Transaction Processing](http://developers.payulatam.com/en/sdk/payments.html) in PayU Latam page. Authorize and Capture together is the most used transaction type.  That's why we opted to extend a base PayuLatamGateway class so that we could reimplement any specific functionality based on the country.

jasonwebster:

PR #1832, opened 2015-08-06
Stripe: don't send blank, non-nil values

girasquid:
@anellis I noticed for https://github.com/activemerchant/active_merchant/pull/1807/files that the way that values are passed in to Stripe was changed - when we went to deploy that a lot of requests started failing with the error `You passed an empty string for 'referrer'. We assume empty values are an attempt to unset a parameter; however 'referrer' cannot be unset. You should remove 'referrer' from your request or supply a non-empty value` - so this restores the old logic and includes a change to also let `false` get in. Open to suggestions on a cleaner way to do this.

@aprofeit @anellis for review please.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -383,7 +383,7 @@ module ActiveMerchant #:nodoc:
         return nil unless params
 
         params.map do |key, value|
-          next if value.nil?
+          next if value != false && value.blank?
           if value.is_a?(Hash)
             h = {}
             value.each do |k, v|
@@ -391,7 +391,7 @@ module ActiveMerchant #:nodoc:
             end
             post_data(h)
           elsif value.is_a?(Array)
-            value.map { |v| "#{key}[]=#{CGI.escape(v.to_s)}" }.join("&") unless value.blank?
+            value.map { |v| "#{key}[]=#{CGI.escape(v.to_s)}" }.join("&")
           else
             "#{key}=#{CGI.escape(value.to_s)}"
           end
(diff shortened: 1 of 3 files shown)


j-mutter:
Maybe add a unit test that does a `stub_comms...check_request` to ensure the empty strings aren't being sent?

Otherwise LGTM

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -383,7 +383,7 @@ def post_data(params)
         return nil unless params
 
         params.map do |key, value|
-          next if value.nil?
+          next if value != false && value.blank?
```
aprofeit: Won't this call `next` on all `values` except `false` now? How can any other value be included in the request now other than `false`?

j-mutter: It's skipped if it's not `false` AND `blank`... It's a little counter intuitive, but the logic works:
```ruby
[1] pry(main)> value = "hello"
=> "hello"
[2] pry(main)> value != false && value.blank?
=> false
[3] pry(main)> value = false
=> false
[4] pry(main)> value != false && value.blank?
=> false
[5] pry(main)> value = ""
=> ""
[6] pry(main)> value != false && value.blank?
=> true
[7] pry(main)> value = nil
=> nil
[8] pry(main)> value != false && value.blank?
=> true
```

aprofeit: Ah right :+1:

aprofeit:
So we just need automated remote tests that test every possible permutation of all the options :)

But this lgtm for now.

anellis:
ah sorry about that! This lgtm as well, seems like a good way to handle it.

girasquid:
All good - thanks for the review!

PR #1835, opened 2015-08-06
Paystation: Map order_id to non-unique merchant reference field

anellis:


anellis:
@duff for paystation! the `test_capture_without_cvv` and `test_successful_purchase_in_gbp" remote tests were failing before with our account.

PR #1837, opened 2015-08-07
Handle Cardstream nil addresses

rwdaigle:
Not just nil street address (follow-up to #1834)

/cc @duff


duff:
@rwdaigle I merged it and tweaked the unit test just a bit to keep around the `test_successful_purchase_without_street_address` since it was added to address the bug on this line:

https://github.com/activemerchant/active_merchant/pull/1834/files#diff-19fe6e8fd626bdd5ca775c40f7812b27L149

that blew up when there was some address but a nil street.

PR #1842, opened 2015-08-13
make sure operation(s) exists

ta:
Fixes bug when no operations exists on the payment

### DIFF
--- a/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
+++ b/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
@@ -214,9 +214,7 @@ module ActiveMerchant
         end
 
         def invalid_operation_message(response)
-          if response['operations']
-            response['operations'].last['qp_status_msg']
-          end
+          response['operations'] && response['operations'].last['qp_status_msg']
         end
 
         def map_address(address)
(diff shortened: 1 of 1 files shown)


girasquid:
Can you add a test?

ta:
I'd be happy to but it would be somewhat difficult to test via remote as some tests requires an active acquirer agreement, and the bug I've fixed originates from a situation where no acquirers are active. I would break the other tests.

A unit test would really just test if `nil || nil || true => true`.

If I've missed something - please enlighten me :)

girasquid:
Would it be possible to add some test credentials for an account with no acquirers that would be able to go through this test? Or not worth it?

ta:
It can be done - IMO its really not worth it.

girasquid:
Fair enough, let's not bother.

lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
```diff
@@ -214,7 +214,9 @@ def invalid_operation_code(response)
         end
 
         def invalid_operation_message(response)
-          response['operations'].last['qp_status_msg']
+          if response['operations']
```
girasquid: This could be `response['operations'] && response['operations'].last['qp_status_msg']`, right?

ta: Hmm.. no, then it would return a boolean. It should return the error message from the last operation (if present)

girasquid:
```ruby
irb(main):004:0> foo['operations'] && foo['operations'].last['msg']
=> "hello"```

ta: My bad - I stand corrected :)

PR #1843, opened 2015-08-13
QuickpayV10: Add Verify

anellis:


anellis:
@duff would you mind giving this a :eyes:

PR #1845, opened 2015-08-14
QuickpayV10: Add scrubbing

anellis:


anellis:
@duff one more, didn't realize it didn't have scrubbing either

test/unit/gateways/quickpay_v10_test.rb
```diff
@@ -268,4 +272,14 @@ def expected_expiration_date
     '%02d%02d' % [@credit_card.year.to_s[2..4], @credit_card.month]
   end
 
+  def transcript
+   'POST /payments/7452604/authorize?synchronized HTTP/1.1\r\nContent-Type: application/json\r\nAuthorization: Basic OjAzNTA4ZTc3OTFiYTZjOWQwZTY4MzA3MTZlNjUwZjM1YzQzNDJjNGIzNTc2NzIzYWQ1NTZlMjM2Y2E0Yzc3ODg=\r\nUser-Agent: Quickpay-v10 ActiveMerchantBindings/1.52.0\r\nAccept: application/json\r\nAccept-Version: v10\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nConnection: close\r\nHost: api.quickpay.net\r\nContent-Length: 136'
```
duff: @anellis Looks like this test will only be using the final string in the method.  Using `%q` like this: https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/nmi_test.rb#L421 would be :egg: cellent

anellis:
@duff Updated!

duff:
@anellis :cool: 

One more thang. :)  When I run the remote test `test_successful_purchase` with `DEBUG_ACTIVE_MERCHANT=true`, I see the line you've got in the test for scrubbing containing the card number and cvv.  But I also see one other line containing the same info but in a different format.  In particular, I see:
 
```
D, [2015-08-17T11:18:34.178127 #20295] DEBUG -- : {"amount":"100","card":{"number":"1000000000000008","cvd":"123","expiration":"1609","issued_to":"Longbob Longsen"},"auto_capture":false}
```

We should have that extra line in the test too.  (And update the `scrub` call to handle both)

anellis:
@duff I added those to the scrubbing. Out of curiosity, I thought those debug lines were only available in the active merchant tests (and therefore not an issue). Is that something that customers (anyone's customers, not just ours) can turn on externally as well?

duff:
My understanding is that anything in that shows up with the `DEBUG_ACTIVE_MERCHANT` flag would show in a transcript.

anellis:
when i add and run the `dump_transcript_and_fail` method to get the transcript.log file the `DEBUG` lines don't show up in that. I was just wondering since I usually work off of that log file and wanted to make sure I wasn't missing things by not doing the DEBUG instead. I suppose either way it doesn't hurt to be safe.

duff:
Interesting!  I'm not sure.  I haven't used the `dump_transcript_and_fail` method.  I've always gone old school with DEBUG_ACTIVE_MERCHANT=true.  @ntalbott might know the difference off the top of his head.  I haven't compared the two.

PR #1846, opened 2015-08-14
Add Creditcall

davidsantoso:
@duff adding the [Creditcall](https://www.creditcall.com) gateway. Let me know if you have and question/comments/concerns.

### DIFF
--- a/lib/active_merchant/billing/gateways/creditcall.rb
+++ b/lib/active_merchant/billing/gateways/creditcall.rb
@@ -1,0 +1,210 @@
+require 'nokogiri'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CreditcallGateway < Gateway
+      self.test_url = 'https://test.cardeasexml.com/generic.cex'
+      self.live_url = 'https://live.cardeasexml.com/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'https://www.creditcall.com'
+      self.display_name = 'Creditcall'
+
+      def initialize(options={})
+        requires!(options, :terminal_id, :transaction_key)
+        super
+      end
+
+      def purchase(money, payment_method, options={})
+        multi_response = MultiResponse.run do |r|
+          r.process { authorize(money, payment_method, options) }
+          r.process { capture(money, r.authorization, options) }
+        end
+
+        Response.new(
+          multi_response.primary_response.success?,
+          multi_response.primary_response.message,
+          multi_response.primary_response.params,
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/creditcall.rb
```diff
@@ -0,0 +1,204 @@
...
+      self.homepage_url = 'https://www.creditcall.com'
+      self.display_name = 'Creditcall'
+
+      def initialize(options={})
+        requires!(options, :terminal_id, :transaction_key)
+        super
+      end
+
+      def purchase(money, payment_method, options={})
+        MultiResponse.run(:use_first_response) do |r|
```
duff: I don't think you want to use `: use_first_response` here.  For `verify` that makes sense since we don't care about the `void`.

lib/active_merchant/billing/gateways/creditcall.rb
```diff
@@ -0,0 +1,204 @@
...
+          if address = options[:billing_address] || options[:address]
+            xml.AdditionalVerification do
+              xml.Address address[:address1]
+              xml.Zip address[:zip]
+            end
+          end
+        end
+      end
+
+      def expiration_date(payment_method)
```
duff: Can we take advantage of https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card_formatting.rb#L13 here?

test/remote/gateways/remote_creditcall_test.rb
```diff
@@ -0,0 +1,147 @@
...
+      order_id: '1',
+      ip: "127.0.0.1",
+      email: "joe@example.com"
+    }
+
+    response = @gateway.purchase(@amount, @credit_card, options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+  end
+
```
duff: Could be :cool: to have a test to run a purchase without a `billing_address` and `description`.  Perhaps a `test_successful_purchase_sans_options`.

lib/active_merchant/billing/gateways/creditcall.rb
davidsantoso: We definitely can. Didn't know that existed :wink:

lib/active_merchant/billing/gateways/creditcall.rb
davidsantoso: :+1:

test/remote/gateways/remote_creditcall_test.rb
davidsantoso: :+1:

davidsantoso:
@duff implemented the change you suggested. Probably want to double check the `purchase` action as we figured out that Creditcall requires the transaction reference from the authorization instead of the capture, but we want to use the capture response to determine if it was a success, etc.

Issue #1827, opened 2015-08-05
SecureNet: using production env and TEST parameter set to true 


wedstar:
hi, 
currently in activemerchant securenet, 
if options[:test] = false passed, it forces us to use 
live_url and TEST element tag set to FALSE ( by default if we dont send / specify this tag, default is FALSE)

and if options[:test] = true, it will use test_url and TEST element tag set to TRUE

according to SecureNet / WorldPay support, we can run a transaction in prod environment (live_url) with the TEST parameter set to TRUE, this will prevent any changes from actually being made, but penalties can be incurred from the card associations if gets run too many of these types of transactions.

what you guys think any thoughts? 
i was thinking perhaps we should have another key called 'test_mode'
if test_mode presence, we set ``XML.tag! 'TEST' to test_mode`` otherwise if test? we set to 'TRUE' otherwise 'FALSE' , please see below

in https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/secure_net.rb:210 
```
 def add_more_required_params(xml, options)
   
        xml.tag! 'RETAIL_LANENUM', '0'
        xml.tag! 'TEST', options[:test_mode] || (test? ? 'TRUE' : 'FALSE')
        xml.tag! 'TOTAL_INSTALLMENTCOUNT', 0
        xml.tag! 'TRANSACTION_SERVICE', 0
      end
``
thank you

girasquid:
> this will prevent any changes from actually being made, but penalties can be incurred from the card associations if gets run too many of these types of transactions.

What's the benefit of doing this vs. just using the test mode/test URL? Having a test mode that might incur penalties doesn't seem great, so I'm not clear on what you would gain here.

wedstar:
hi @girasquid , the securnet support team recommends to only use the TEST parameter to verification purposes in the production environment, i guess doesnt have many benefits 

Securenet allows us to set TEST = true in production mode, and i think the code should allow us to do so, regardless penalties or no penalties, thats my opinion.

ntalbott:
Closed in preference for the PR #1848.

PR #1848, opened 2015-08-18
Securenet: add test_mode parameters for TEST element tag.

wedstar:
Securenet allows us to set TEST = true in production mode, and i think the code should allow us to do so.
please see:  https://github.com/activemerchant/active_merchant/issues/1827

### DIFF
--- a/lib/active_merchant/billing/gateways/secure_net.rb
+++ b/lib/active_merchant/billing/gateways/secure_net.rb
@@ -208,8 +208,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_more_required_params(xml, options)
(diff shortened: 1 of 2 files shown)


ntalbott:
@wedstar I *think* this is changing default behavior, since setting just `:test` to `true` will now result in the `TEST` param being set to `FALSE` (though the right url will be used). Seems like if `:test` is true, then `TEST` should always be `TRUE` as well, right?

wedstar:
Hi @ntalbott  , i dont think it will change the default behaviour.
if `options[:test_mode]` is nil, we will use `test?` (which this is the default behaviour)
if `options[:test_mode]` is `true` or `false` we will use this test_mode value instead of `test?`
thank you

PR #1849, opened 2015-08-18
USAePay: Use names from the given billing and shipping address

marquisong:
This PR updates the add_address_for_type() to use the names from billing and shipping address, instead of credit card's name. It will fall back to current behaviour if names in billing and shipping address are blank.

### DIFF
--- a/lib/active_merchant/billing/gateways/usa_epay_transaction.rb
+++ b/lib/active_merchant/billing/gateways/usa_epay_transaction.rb
@@ -141,9 +141,10 @@ module ActiveMerchant #:nodoc:
 
       def add_address_for_type(type, post, credit_card, address)
         prefix = address_key_prefix(type)
+        first_name, last_name = split_names(address[:name])
 
-        post[address_key(prefix, 'fname')]    = credit_card.first_name
-        post[address_key(prefix, 'lname')]    = credit_card.last_name
+        post[address_key(prefix, 'fname')]    = first_name.blank? && last_name.blank? ? credit_card.first_name : first_name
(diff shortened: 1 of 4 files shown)


aprofeit:
Would you be able to add some tests for this? Thinking:
- Shipping and billing address names are sent when present
- Credit card name is sent when shipping and billing address names are absent

marquisong:
Thanks @aprofeit - There's already tests for billing and shipping address names are sent when present but I have added the tests for scenarios where billing and shipping address names are absent now.

lib/active_merchant/billing/gateways/usa_epay_transaction.rb
```diff
@@ -154,6 +155,13 @@ def add_address_for_type(type, post, credit_card, address)
         post[address_key(prefix, 'phone')]    = address[:phone]     unless address[:phone].blank?
       end
 
+      def split_names(full_name)
+        names = (full_name || '').split
```
aprofeit: Should we add a guard to explicitly `return [nil, nil] if names.size <= 1`? Two reasons:
- When `''` or `nil` is passed to `split_names` it returns `["", nil]`. Feel like it should return the same value for both items.
- If name is a single word, it will end up combining the address name and credit card name in the `post`. For example, `split_names('bob') => ['', 'bob']` which would then add `fname = credit_card.first_name` and `lname = address.last_name`. Should probably use _only_ the address name or _only_ the credit card name.

marquisong: @aprofeit Yes, new guards have been added above to make it to use and return a consistent result e.g. `[nil, nil]`.
Thanks again.

PR #1851, opened 2015-08-18
Beanstream: Add Network Tokenization support

girasquid:
@duff @andrewpaliga 

Beanstream tells me that there's no way to do an automated remote test for this, so I haven't (but I'm open to suggestions).


andrewpaliga:
@girasquid do remote tests pass when the tokenization params are passed?

j-mutter:
@andrewpaliga the problem is that you can't really fake the values for those fields. So you could send them, but it won't work.

I haven't read the docs, but this makes sense to me. :ship:

aprofeit:
:+1: test in production I guess.

PR #1852, opened 2015-08-19
QuickPayV10: Change tests to point to proper gateway

anellis:
The QuickPayV10 gateway tests were previously pointing to the old QuickPay Gateway in their setup.


anellis:
@duff

PR #1853, opened 2015-08-19
Remove ActiveMerchant::Billing::Base.integration_mode

### LINKED ISSUE #1661: OFFSITE_PAYMENT_EXTRACTION_MESSAGE removed but still referenced
The constant `OFFSITE_PAYMENT_EXTRACTION_MESSAGE` was removed by d730e442a7aa0bb7f88b705c2a1a811c703e479f , but it is still referenced [here](https://github.com/Shopify/active_merchant/blob/456cf85cab51d161960ebfde1944f0f76f77e54d/lib/active_merchant/billing/base.rb#L13-L18).

Calling `ActiveMerchant::Billing::Base.integration_mode = :test` causes:

```[...] lib/active_merchant/billing/base.rb:13:in `integration_mode=': uninitialized constant ActiveMerchant
::Billing::Base::OFFSITE_PAYMENT_EXTRACTION_MESSAGE (NameError)```

### PR DESCRIPTION
**Pull request by aprofeit**
It doesn't look like `ActiveMerchant::Billing::Base.integration_mode` is used any longer and it errors out if you try to use it since `OFFSITE_PAYMENT_EXTRACTION_MESSAGE` is no longer defined.

`ActiveMerchant::Billing::Base.mode` used to set both `integration_mode` and `gateway_mode`, I changed it to just be an alias for `gateway_mode` now.

Review please @girasquid @ntalbott 

Closes #1661

### DIFF
--- a/lib/active_merchant/billing/base.rb
+++ b/lib/active_merchant/billing/base.rb
@@ -1,32 +1,21 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     module Base
+      GATEWAY_MODE_DEPRECATION_MESSAGE = 'Base#gateway_mode is deprecated in favor of Base#mode and will be removed in a future version'
+
       # Set ActiveMerchant gateways in test mode.
       #
-      #   ActiveMerchant::Billing::Base.gateway_mode = :test
-      mattr_accessor :gateway_mode
+      #   ActiveMerchant::Billing::Base.mode = :test
+      mattr_accessor :mode
 
-      # Set ActiveMerchant integrations in test mode.
-      #
-      #   ActiveMerchant::Billing::Base.integration_mode = :test
-      def self.integration_mode=(mode)
-        ActiveMerchant.deprecated(OFFSITE_PAYMENT_EXTRACTION_MESSAGE)
-        @@integration_mode = mode
-      end
-
-      def self.integration_mode
-        ActiveMerchant.deprecated(OFFSITE_PAYMENT_EXTRACTION_MESSAGE)
-        @@integration_mode
-      end
-
-      # Set both the mode of both the gateways and integrations
-      # at once
-      mattr_reader :mode
-
-      def self.mode=(mode)
+      def self.gateway_mode=(mode)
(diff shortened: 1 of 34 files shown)

### CODE REVIEW

**Comment by aprofeit** [comment]
Er, sorry, didn't realize we have tests to explicitly allow `mode` and `gateway_mode` to be different. I'll fix.

**Comment by aprofeit** [comment]
Do we actually need `gateway_mode` and `mode` anymore? Can we use _just_ mode and kill `gateway_mode`?

**Comment by j-mutter** [comment]
@aprofeit was just about to ask the same thing. Should just have `mode` imo.

**Comment by girasquid** [comment]
Test failures look legit

**Comment by aprofeit** [comment]
Since we no longer need to differentiate between `gateway_mode` and `integration_mode` I deprecated `gateway_mode` and removed `integration_mode` in favor of just using `mode`. `gateway_mode` will continue to function as an alias for `mode` with a deprecation notice.

Thoughts? @girasquid @ntalbott

File: lib/active_merchant/billing/base.rb [hunk 1]
```diff
@@ -1,32 +1,21 @@
...
-      #
-      #   ActiveMerchant::Billing::Base.gateway_mode = :test
-      mattr_accessor :gateway_mode
+      GATEWAY_MODE_DEPRECATION_MESSAGE = 'Base#gateway_mode is deprecated in favor of Base#mode and will be removed in a future version'
 
-      # Set ActiveMerchant integrations in test mode.
+      # Set ActiveMerchant gateways in test mode.
       #
-      #   ActiveMerchant::Billing::Base.integration_mode = :test
-      def self.integration_mode=(mode)
```
girasquid: This is completely removing `integration_mode` - has it been long enough that we're fine doing that?

aprofeit: The method doesn't work anyways. `OFFSITE_PAYMENT_EXTRACTION_MESSAGE` is undefined.

girasquid: Oh! Fair enough, then.

j-mutter:
:+1:

girasquid:
:+1:

PR #1854, opened 2015-08-20
Ogone: Add additional 3d-secure parameters

ntalbott:
Also: Refactor signature calculation

@rwdaigle can I get a review here?

### DIFF
--- a/test/remote/gateways/remote_ogone_test.rb
+++ b/test/remote/gateways/remote_ogone_test.rb
@@ -39,23 +39,25 @@ class RemoteOgoneTest < Test::Unit::TestCase
     assert_equal OgoneGateway::SUCCESS_MESSAGE, response.message
   end
 
+  # This test is commented out since it is mutually exclusive with the other signature tests.
   # NOTE: You have to set the "Hash algorithm" to "SHA-1" in the "Technical information"->"Global security parameters"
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_ogone_test.rb
```diff
@@ -41,21 +41,21 @@ def test_successful_purchase_with_utf8_encoding_2
...
-    assert response = gateway.purchase(@amount, @credit_card, @options)
-    assert_success response
-    assert_equal OgoneGateway::SUCCESS_MESSAGE, response.message
-  end
+  # def test_successful_purchase_with_signature_encryptor_to_sha256
+  #   gateway = OgoneGateway.new(fixtures(:ogone).merge(:signature_encryptor => 'sha256'))
+  #   assert response = gateway.purchase(@amount, @credit_card, @options)
+  #   assert_success response
+  #   assert_equal OgoneGateway::SUCCESS_MESSAGE, response.message
+  # end
```
rwdaigle: What's with the removal of these tests and no new ones?

rwdaigle: Reading more I believe it's because you need to toggle a setting in the control panel to get either to pass...?

ntalbott: @rwdaigle correct, I don't like remote tests that always fail. I'll clarify the comment.

PR #1856, opened 2015-08-21
New openpay specific fields (Cabify requirements)

darkaz:
- Added fraud_safe.- true if the charge would be free of fraud system
- Creation of the customer when "name" and "email" are in options
http://www.openpay.mx/en/docs/api/#charge-a-new-card
- If shipping_address is in options, it will send as customer address


rwdaigle:
@duff was wondering if I could get your take on this. I'm most interested in how, if a `:customer` option is present, it routes it to a different resource. For some reason I'm uneasy with the constant resource forking.

Docs are here: http://www.openpay.mx/en/docs/api/?undefined#introduction

rwdaigle:
I believe this can be closed due to changed customer requirements.

PR #1858, opened 2015-08-24
Adding CCAvenue Payment gateway

avenuestest:
CCAvenue Payment Gateway Integration

### DIFF
--- a/lib/active_merchant/billing/gateways/ccavenue_mcpg.rb
+++ b/lib/active_merchant/billing/gateways/ccavenue_mcpg.rb
@@ -1,0 +1,164 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CcavenueMcpgGateway < Gateway
+      self.test_url = 'https://test.ccavenue.com/transaction/transaction.do?command=initiateTransaction'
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/ccavenue_mcpg.rb
```diff
@@ -0,0 +1,164 @@
...
+	  
+	  # code for creating encrypted request
+      def post_data(parameters = {})
+		  plain_text = create_plain_text(parameters);
+		  key = parameters[:working_key]
+		  secret_key =  [Digest::MD5.hexdigest(key)].pack("H*") 
+		  cipher = OpenSSL::Cipher::Cipher.new('aes-128-cbc')
+		  cipher.encrypt
+		  cipher.key = secret_key
+		  cipher.iv  = INIT_VECTOR
```
ntalbott: @avenuestest do you work for or with CCAvenue? If so, you should let them know that reusing the same IV makes this encryption very vulnerable: https://en.wikipedia.org/wiki/Initialization_vector#SSL_2.0_IV.

On the plus side, it's useless extra work anyhow since the communication is over TLS, so I'll still consider inclusion of CCAvenue in ActiveMerchant.

ntalbott:
Does CCAvenue only allow `purchase`?

We will need credentials in order to run the remote tests - can you provide those?

PR #1862, opened 2015-08-31
Provide better insight to CVV usage in requests

davidsantoso:
Sometimes, the CVV isn't required for a transaction so it's helpful to know in the transcript whether or not it was actually sent with a request. With the current scrubbing logic, "[FILTERED]" will always be used even if the CVV was not included. This replaces "[FILTERED]" with "[BLANK]" when the CVV is nil, an empty string, or just whitespace, so it's explicit about what was sent.

On the other hand, some gateways do not require a CVV on every request, in those cases, a guard clause is being added to not add the CVV tag/param/key to the request at all.

This PR includes changes for the following gateways, but more will be added over time.

- AlliedWallet
- Paypal
- Openpay
- Redsys
- NMI

Overall, this should help with visibility into the transactions as the transcript captures them.

cc/ @duff


davidsantoso:
@duff one thing to note- the scrubbing regex might a bit too strict, like using `\d+` for non blank values. I wonder if that might be an issue when it comes to someone accidentally entering something that isn't a digit and then it "leaks" in the sense that the CVV will be present, albeit a little incorrect. Thoughts?

duff:
@davidsantoso This is looking great.  I hadn't thought of skipping CVV tags if the CVV is blank.  That's a good simplification I think and it does accomplish the goal of helping us know by looking at a transcript if a CVV was actually sent.

Only little adjustment I made was in fd93daba9b021ded79ad0cec3a507027a9b09a49 to make it work for both `nil` and the empty string.  And I added a little test to verify that we are / aren't sending the CVV info at the right time for those 2 gateways since that'll be our indicator.

:+1:

PR #1865, opened 2015-09-04
Send line item data to Cybersource on auth/capture

fastjames:
Cybersource can use line item data when evaluating Decision Manager
rules, so we should pass that data to them on auth and capture requests
if it's available.

I'm in a quandry over how to unit test this change. The current tests don't really expose the request structure, and the test responses shouldn't change since there's no use of Decision Manager. I have several additional fields that I'd like to add to the request body, so I am open to suggestions about how to test all of them.


fastjames:
Closing until I can verify proper operation.

PR #1866, opened 2015-09-08
Send customer IP address to CyberSource

fastjames:
Allow passing customer IP address to Cybersource via options[:ip], which
appears to be the active_merchant convention.


fastjames:
Closing while I work on an odd response from CyberSource.

PR #1867, opened 2015-09-09
Add beanstream verify method, unit tests and remote tests

mrezentes:


duff:
@mrezentes I just spotted this one.  @mentioning in the GH issue is probably a good idea if you're waiting on PR.

All merged in!!

PR #1868, opened 2015-09-11
Set `receipt_email` to Stripe request if email is available.

miccheng:
To ensure that payer gets an email receipt from Stripe.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -368,7 +368,7 @@ module ActiveMerchant #:nodoc:
 
       def add_metadata(post, options = {})
         post[:metadata] = options[:metadata] || {}
-        post[:metadata][:email] = options[:email] if options[:email]
+        post[:metadata][:email] = post[:receipt_email] = options[:email] if options[:email]
         post[:metadata][:order_id] = options[:order_id] if options[:order_id]
         post.delete(:metadata) if post[:metadata].empty?
       end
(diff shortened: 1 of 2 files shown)


ntalbott:
@miccheng will this cause an email receipt to always be sent if the `:email` option is set? That would be undesirable - it should require explicit action to trigger a receipt.

miccheng:
Emails are never sent in test mode on Stripe. And it feels more proper to
always inform the payer through an email receipt in production.

On Saturday, September 12, 2015, Nathaniel Talbott <notifications@github.com>
wrote:

> @miccheng <https://github.com/miccheng> will this cause an email receipt
> to always be sent if the :email option is set? That would be undesirable
> - it should require explicit action to trigger a receipt.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/activemerchant/active_merchant/pull/1868#issuecomment-139609819>
> .
>


-- 
Regards,

Michael Cheng CM
Mobile: +65 9185 5166
LinkedIn: http://sg.linkedin.com/in/miccheng

girasquid:
The option needs to be explicit - if you have your own receipt sending code, having this also send a receipt means that your customers are going to get two receipts. Not great.

ntalbott:
@miccheng as @girasquid says, this conflicts with orgs that have their own receipt sending code, and breaks the current expectation. Happy to see a different PR that makes this an explicit option - thanks!

miccheng:
No worries. I'll do another PR.

PR #1869, opened 2015-09-11
Map standard error codes for paypal gateway

JakeCataford:
This does not map all of the error codes for paypal, but it provides a framework for providing mappings in the future. Right now, this commit includes mappings for a bunch of paypal error codes which translate to `:card_declined`.

If there is no mapping for the error code, ~~I return the original error code instead~~ I return a processing error.

@girasquid @etiennebarrie

### DIFF
--- a/lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
+++ b/lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
@@ -43,17 +43,16 @@ module ActiveMerchant #:nodoc:
       STANDARD_ERROR_CODE_MAPPING = {
         '15005' => :card_declined,
         '10754' => :card_declined,
-        '10756' => :card_declined,
         '10752' => :card_declined,
         '10759' => :card_declined,
         '10761' => :card_declined,
-        '10762' => :card_declined,
         '15002' => :card_declined,
-        '15004' => :card_declined,
-        '11084' => :card_declined
+        '11084' => :card_declined,
+        '15004' => :incorrect_cvc,
+        '10762' => :invalid_cvc,
       }
 
-      DEFAULT_ERROR_CODE = :processing_error
+      STANDARD_ERROR_CODE_MAPPING.default = :processing_error
 
       def self.included(base)
         base.default_currency = 'USD'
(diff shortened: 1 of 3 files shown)


JakeCataford:
@jpcaissy

lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
```diff
@@ -637,11 +650,16 @@ def commit(action, request)
...
           :authorization => authorization_from(response),
           :fraud_review => fraud_review?(response),
+          :error_code => error_code(response),
           :avs_result => { :code => response[:avs_code] },
           :cvv_result => response[:cvv2_code]
         )
       end
 
+      def error_code(response)
+        STANDARD_ERROR_CODE_MAPPING[response[:error_codes]] || response[:error_codes]
```
jpcaissy: `error_code` should only return a standardized error code. Instead of returning Paypal's original code, you could default to `processing_error`

lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
```diff
@@ -40,6 +40,21 @@ module PaypalCommonAPI
...
+      STANDARD_ERROR_CODE_MAPPING = {
+        '15005' => :card_declined,
+        '10754' => :card_declined,
+        '10756' => :card_declined,
+        '10752' => :card_declined,
+        '10759' => :card_declined,
+        '10761' => :card_declined,
+        '10762' => :card_declined,
+        '15002' => :card_declined,
+        '15004' => :card_declined,
```
girasquid: This should be `incorrect_cvc`

lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
```diff
@@ -40,6 +40,21 @@ module PaypalCommonAPI
 
       FRAUD_REVIEW_CODE = "11610"
 
+      STANDARD_ERROR_CODE_MAPPING = {
+        '15005' => :card_declined,
+        '10754' => :card_declined,
+        '10756' => :card_declined,
```
girasquid: This should be `processing_error`

lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
```diff
@@ -40,6 +40,21 @@ module PaypalCommonAPI
...
+        '10752' => :card_declined,
+        '10759' => :card_declined,
+        '10761' => :card_declined,
+        '10762' => :card_declined,
```
girasquid: This should be `invalid_cvc`

JakeCataford:
This pr shouldn't break any existing code, as the `:error_code` parameter was non existent before. Is this good to go?

girasquid:
:+1:

PR #1871, opened 2015-09-14
Fix checkout_v2 integration for an empty phone number

AnotherJoSmith:
Fixes an issue where sending JSON like this
```json
              "phone" : {
                  "number" : ""
              }
```

with the phone element but with an empty number would fail. The API would return `Validation error: Invalid length for 'phone number'`. Now if the phone number is empty, it will not send the top level "phone" element.

@girasquid @ivanfer for review

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -100,7 +100,7 @@ module ActiveMerchant #:nodoc:
           post[:card][:billingDetails][:state] = address[:state]
           post[:card][:billingDetails][:country] = address[:country]
           post[:card][:billingDetails][:postcode] = address[:zip]
-          post[:card][:billingDetails][:phone] = { number: address[:phone] }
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -100,7 +100,7 @@ def add_customer_data(post, options)
           post[:card][:billingDetails][:state] = address[:state]
           post[:card][:billingDetails][:country] = address[:country]
           post[:card][:billingDetails][:postcode] = address[:zip]
-          post[:card][:billingDetails][:phone] = { number: address[:phone] }
+          post[:card][:billingDetails][:phone] = { number: address[:phone] } unless address[:phone].empty?
```
ivanfer: use `blank?` instead of empty because `empty?` is not defined on nil

AnotherJoSmith: done! :+1:

girasquid:
:+1:

ivanfer:
:palm_tree:

PR #1873, opened 2015-09-15
Quickpay: edit store and add ability to purchase with stored card

anellis:


anellis:
@duff finally got this quickpay store working correctly! Would you mind taking a look? You'll have to use a different api key then the one in fixtures, so I've added it to the 1password

duff:
@anellis The `unstore` remote test seems to be failing for me.  Thought I'd check with you to see if it was working for you.

anellis:
@duff alright! Sorry about that. Now just the acquirer test should fail with our api key (since the account I  made has different acquirers than the other)

wsmoak:
May I ask why the use of the `/subscriptions` endpoint was removed?

The `/cards` endpoint only returns a single-use token, which cannot be stored and used for recurring payments.

wsmoak:
Thanks! More here: https://github.com/activemerchant/active_merchant/issues/2172
tl;dr: `/subscriptions` isn't necessary, but the 'token' call needs to move from 'store' to 'purchase'.

duff:
Hi @wsmoak, does that mean that every purchase call would store another card in the Quickpay vault?  Is there any way to make it such that purchase just runs a purchase (using either a stored reference or a passed in card) and make `store` give the stored reference that isn't just one-time?  Is that what the `/subscriptions` endpoint does?  

I think @anellis was in communication with Quickpay support who was suggesting the `token` call for `store`.  Does the `/subscriptions` endpoint hook up auto charging based on a schedule or is it like the `token` call, only more than one-use?

Thanks for looking into this!  The fact that the 2nd attempt to use the token failed indicates that the `token` call is not the thing we need. 😭

wsmoak:
No, every purchase would not store another card in the QuickPay vault.

I don't think we need to go back to using `/subscriptions` (but from my brief look it didn't look automatic/scheduled.)

I haven't tried this, but it appears that `/payments/:id/authorize` takes _either_ a token or the full card info:  http://tech.quickpay.net/api/services/?scope=merchant#POST-payments--id-authorize---format-

duff:
Coolo!  Sounds like it has potential.

mroach:
I've been looking at the way this has been implemented and I haven't been able to figure out how to use a card after it has been used for the first purchase. It doesn't help that QuickPay hardly has any integration documentation, only generated API docs.

From what I understand...
When you store a card for the first time you're given a `token` and an `id`. The `token` *can only be used once for a purchase* and the `id` is the card's ID that you want to store for future charges, updates, or to delete the card. Then, every transaction requires that you first generate a new token to then be sent with you request (e.g. authorize, purchase).

Are my understandings of this process correct? If so, it doesn't seem that the current version of the implementation allows for the creation of a `token` against a card since the `create_token` method is private. That's probably fine, but I'm thinking there should be modifications to call that internally when a card `id` is passed instead of a credit card object. For example if you call `authorize`, you should only have to call it with the card's `id` and internally the `token` should be created for you.

PR #1875, opened 2015-09-16
Send customer IP address to CyberSource

fastjames:
Allow passing customer IP address to Cybersource via options[:ip], which
appears to be the active_merchant convention.


fastjames:
I have run this code change against the CyberSource test business center, and it performed as expected.

fastjames:
I added a unit test for the presence of the ipAddress element, similar to my other PR.

girasquid:
Thanks!

PR #1876, opened 2015-09-16
FirstData_e4: add level_3 data

mrezentes:


test/remote/gateways/remote_firstdata_e4_test.rb
```diff
@@ -31,6 +31,14 @@ def test_successful_purchase_with_track_data
     assert_success response
   end
 
+  def test_successful_purchase_with_invoice_detail
+    test_level_3 = "<TaxAmount>8.84</TaxAmount><TaxRate>7.25</TaxRate>"
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(level_3: test_level_3))
```
rwdaigle: This seems to imply the level 3 options will be passed in not as a hash, but as an XML string? If so, I believe that's non-standard and the input options should be hash-based:

```ruby
test_level_3 = { tax_amount: 8.84, tax_rate: 7.25 }
response = @gateway.purchase(@amount, @credit_card, @options.merge(level_3: test_level_3))
```

ntalbott: @rwdaigle L3 data is NUTSO, so what we did with CenPOS and are proposing with this gateway is to allow the customer to pass it through in a blob vs. us trying to parse it and support it as distinct fields.

PR #1877, opened 2015-09-16
Send line item data to Cybersource on auth/capture

fastjames:
Cybersource can use line item data when evaluating Decision Manager
rules, so we should pass that data to them on auth and capture requests
if it's available.

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -636,6 +636,7 @@ module ActiveMerchant #:nodoc:
         else
           add_address(xml, payment_method_or_reference, options[:billing_address], options)
           add_address(xml, payment_method_or_reference, options[:shipping_address], options, true)
+          add_line_item_data(xml, options) if options[:line_items]
           add_purchase_data(xml, money, true, options)
           add_creditcard(xml, payment_method_or_reference)
         end
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -73,6 +73,16 @@ class CyberSourceTest < Test::Unit::TestCase
     assert response.test?
   end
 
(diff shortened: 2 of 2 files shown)


fastjames:
I was reluctant to modify the existing ```add_line_item_data``` method so I overlaid what I needed to make it optional. I can fold those two together if it's OK for the tax calculation request (that seemed to be the only way the request enforced a requirement for line item data).

fastjames:
I have tested this change against the Cybersource sandbox, and it worked as expected.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -636,6 +642,7 @@ def add_payment_method_or_subscription(xml, money, payment_method_or_reference,
         else
           add_address(xml, payment_method_or_reference, options[:billing_address], options)
           add_address(xml, payment_method_or_reference, options[:shipping_address], options, true)
+          optionally_add_line_item_data(xml, options)
```
j-mutter: Why not just `add_line_item_data(xml, options) if options[:line_items]`

j-mutter:
Are there any cases where anyone *wouldn't* want to send this data along? I'm not super familiar with the CyberSource API.

Also should be tested.

lib/active_merchant/billing/gateways/cyber_source.rb
fastjames: Correct, I think the previous method was an artifact of an earlier approach. Changed.

fastjames:
@j-mutter I'm not aware of any cases when someone would pass the data to the call but not want it to be sent to CyberSource.
I'll add tests to this PR.

fastjames:
@j-mutter I have added tests for auth and purchase. I borrowed the pattern from the payflow unit test, since I did not see any other tests here that inspected the request structure.

jasonwebster:

PR #1878, opened 2015-09-18
added nab velocity payment gateway

ashishCetuG:
North American Bancard (NAB) is a credit card processing company that offers solutions for credit, debit, EBT, check conversion and guarantee, gift & loyalty cards and more. 
North American Bancard (NAB) provides a ruby gem ruby-sdk in https://github.com/nab-velocity/ruby-sdk. 
We are adding  ruby-sdk gem to active merchant as a payment gateway.


ntalbott:
@ashishCetuG this doesn't seem to follow any ActiveMerchant conventions, and so won't be merged. If you want us to support Velocity, you'll need to closely read existing ActiveMerchant adapters and re-write this adapter to use the conventions of the library.

ashishCetuG:
@ntalbott can you please tell me to which portions in particular that i need to re-write and  is not following conventions.

ntalbott:
@ashishCetuG I'd start with working on matching the public API of other adapters - `purchase`/`authorize`/`void`/etc. all have a very specific set of arguments they take.

Again, your best bet is to closely read and follow the example of other ActiveMerchant adapters, like Litle, HPS, Ezic, etc.

PR #1882, opened 2015-09-24
Orbital customer profiles is confusing

nathany:
`@options` for the gateway vs. local `options` to purchase/authorize is pretty confusing.

not sure if adding a comment is enough to make this clear for people trying to figure out usage.

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -180,69 +180,74 @@ module ActiveMerchant #:nodoc:
       def initialize(options = {})
         requires!(options, :merchant_id)
         requires!(options, :login, :password) unless options[:ip_authentication]
-        super
+        # optional options:
+        # :customer_profiles
+        # :retry_logic
+        # :ip_authentication
+        # :bin
+        super # super sets @options
       end
 
       # A – Authorization request
-      def authorize(money, creditcard, options = {})
-        order = build_new_order_xml(AUTH_ONLY, money, options) do |xml|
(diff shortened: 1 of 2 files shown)


girasquid:
I'm not sold on the comment clearing things up that much - maybe instead, have initialize set an ivar and check that? That way you'd at least know where the value was coming from

nathany:
I agree regarding this comment.

Another option would be to use `opts` for local arguments rather than `options`. It doesn't make it clear where @options comes from, but at least it's clear that it isn't a typo.

I could also add comments to `initialize` to say which arguments it takes.

nathany:
please take another look

girasquid:
I'd rather stick with `options` as the name of what's getting passed to the methods, to stay consistent with other implementations. Is there an issue with the ivar approach? It looks like [that's how stripe.rb handles it](https://github.com/activemerchant/active_merchant/blob/0dc1e5091a57d479fcc78cc9ce1640a559b9f1fd/lib/active_merchant/billing/gateways/stripe.rb#L52-L53).

nathany:
Oh, I see what you mean.

nathany:
Unfortunately this is a potentially breaking change. The current unit tests are directly modifying the @options hash:

```ruby
@gateway.options[:customer_profiles] = true
```

Now, I could fix the tests, but if anyone else happens to be doing something like this, I've just broken their code by depending on new ivars instead of the @options hash.

I'd rather not add additional hacks like overriding `options[]=`.

I think I'm going to give up on this PR. It's just not _that_ important to me. Thanks for taking a look at it and providing feedback.

PR #1886, opened 2015-10-01
Adding basic EMV support to the bogus gateway

ryanbalsdon:
We don't have access to the PAN when testing EMV cards so the success/failure switch has to be done based on the amount instead. This patch keeps support for deciding on PAN for non-emv payments.

Review please @davidseal, @abecevello, @bizla

### DIFF
--- a/lib/active_merchant/billing/gateways/bogus.rb
+++ b/lib/active_merchant/billing/gateways/bogus.rb
@@ -4,9 +4,13 @@ module ActiveMerchant #:nodoc:
     class BogusGateway < Gateway
       AUTHORIZATION = '53433'
 
+      AUTHORIZATION_EMV_SUCCESS = '8A023030'
+      AUTHORIZATION_EMV_DECLINE = '8A023035'
+
       SUCCESS_MESSAGE = "Bogus Gateway: Forced success"
       FAILURE_MESSAGE = "Bogus Gateway: Forced failure"
-      ERROR_MESSAGE = "Bogus Gateway: Use CreditCard number ending in 1 for success, 2 for exception and anything else for error"
+      NUMBER_ERROR_MESSAGE = "Bogus Gateway: Use CreditCard number ending in 1 for success, 2 for exception and anything else for error"
+      AMOUNT_ERROR_MESSAGE = "Bogus Gateway: Use amount ending in 00 for success, 05 for failure and anything else for exception"
       UNSTORE_ERROR_MESSAGE = "Bogus Gateway: Use trans_id ending in 1 for success, 2 for exception and anything else for error"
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/bogus.rb
```diff
@@ -19,26 +20,18 @@ class BogusGateway < Gateway
...
       end
 
       def purchase(money, paysource, options = {})
-        money = amount(money)
-        case normalize(paysource)
-        when /1$/, AUTHORIZATION
-          Response.new(true, SUCCESS_MESSAGE, {:paid_amount => money}, :test => true, :authorization => AUTHORIZATION)
-        when /2$/
-          Response.new(false, FAILURE_MESSAGE, {:paid_amount => money, :error => FAILURE_MESSAGE }, :test => true, :error_code => STANDARD_ERROR_CODE[:processing_error])
+        if paysource.respond_to?(:emv?) && paysource.emv?
```
bizla: You can do `paysource.try(:emv?)` here.

ryanbalsdon: `#try` is a Rails thing? It makes tests fail.

ryanbalsdon:
I think this is ready to ship. Thoughts? @bizla @abecevello @girasquid

lib/active_merchant/billing/gateways/bogus.rb
```diff
@@ -118,6 +111,54 @@ def unstore(reference, options = {})
 
       private
 
+      def authorize_emv(money, paysource, options = {})
+        money = amount(money)
+        case money
+        when /00$/
+          Response.new(true, SUCCESS_MESSAGE, {:authorized_amount => money}, :test => true, :authorization => AUTHORIZATION, :emv_authorization => "8A023030" )
```
bizla: `"8A023030"` should be a constant so it reads more easily to those not familiar with EMV.

lib/active_merchant/billing/gateways/bogus.rb
```diff
@@ -118,6 +111,54 @@ def unstore(reference, options = {})
...
+        when /05$/
+          Response.new(false, FAILURE_MESSAGE, {:authorized_amount => money, :error => FAILURE_MESSAGE }, :test => true, :error_code => STANDARD_ERROR_CODE[:processing_error], :emv_authorization => "8A023035" )
```
bizla: Same here re: constant.

bizla:
Constantize the `emv_authorization` default strings and then it looks good to merge to me.

abecevello:
I have the same comment as @bizla. Fine to merge from me when that constant is created.

ryanbalsdon:
Added the constants and squashed to a single commit.

bizla:
Looks good. I'll merge this in.

PR #1887, opened 2015-10-01
[Paybox] set currency from request options

ivanfer:
## Issue
Paybox Integration doesn't respect currency because it's drawing from the credentials `options` hash instead of the request `options`

## Fix
extract money parameters, and test that it works

@aprofeit @girasquid

### DIFF
--- a/lib/active_merchant/billing/gateways/paybox_direct.rb
+++ b/lib/active_merchant/billing/gateways/paybox_direct.rb
@@ -67,6 +67,8 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_invoice(post, options)
         add_creditcard(post, creditcard)
+        add_amount(post, money, options)
+
         commit('authorization', money, post)
       end
 
@@ -74,6 +76,8 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_invoice(post, options)
         add_creditcard(post, creditcard)
+        add_amount(post, money, options)
+
         commit('purchase', money, post)
       end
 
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/paybox_direct.rb
```diff
@@ -125,6 +133,11 @@ def add_reference(post, identification)
         post[:numtrans] = identification[10,10]
       end
 
+      def add_amount(post, money, options)
+        post[:montant] = ('0000000000' + (money ? amount(money) : ''))[-10..-1]
```
aprofeit: What does this change have to do with the currency? And what are all these `0`s for?

aprofeit: Oh I see, just moved code.

ivanfer: i refactored out the money-type fields into it's own method, from its previous home on `commit`. It looks like this gateway needs a specific length for the `montant` field. Dunno.

lib/active_merchant/billing/gateways/paybox_direct.rb
```diff
@@ -135,8 +148,6 @@ def parse(body)
       end
 
       def commit(action, money = nil, parameters = nil)
-        parameters[:montant] = ('0000000000' + (money ? amount(money) : ''))[-10..-1]
-        parameters[:devise] = CURRENCY_CODES[options[:currency] || currency(money)]
```
aprofeit: Where did this `options` come from before?

aprofeit: Wondering why this isn't just blowing up on `undefined local variable or method`

ivanfer: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L136

aprofeit: Inheritance, always making things _sooo_ easy to understand.

aprofeit:
Nice tests :+1:

PR #1891, opened 2015-10-06
JetPay: rewrite gateway for api version 2

mrezentes:
Still need to iron out the credential issues.  JetPay Support hasn't responded with how the Device and DeveloperID works for all customers.


mrezentes:
PR is updated with device and developer id as Jetpay has instructed.  Still waiting on test scripts from Jetpay.

PR #1894, opened 2015-10-09
Authentication for HTTP requests through proxy

cristianstanescu:
Allow passing a user and a password along address and port for proxies.

Remove previous :proxy_address and :proxy_port accessors and replace them with :proxy, a Ruby URI String to be parsed for proxy information.

### DIFF
--- a/lib/active_merchant/connection.rb
+++ b/lib/active_merchant/connection.rb
@@ -30,6 +30,8 @@ module ActiveMerchant
     attr_accessor :tag
     attr_accessor :ignore_http_status
     attr_accessor :max_retries
+    attr_accessor :proxy_address
+    attr_accessor :proxy_port
     attr_reader :proxy
 
     def initialize(endpoint)
@@ -43,6 +45,8 @@ module ActiveMerchant
       @max_retries  = MAX_RETRIES
       @ignore_http_status = false
       @ssl_version = nil
+      @proxy_address = nil
+      @proxy_port = nil
       @proxy = nil
     end
 
@@ -96,7 +100,7 @@ module ActiveMerchant
 
     private
     def http
-      http = new_http_agent(endpoint, proxy)
+      http = new_http_agent
       configure_debugging(http)
       configure_timeouts(http)
       configure_ssl(http)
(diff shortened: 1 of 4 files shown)


ntalbott:
This seems reasonable, but would need to be implemented in a backwards-compatible way. Thanks!

cristianstanescu:
Thank you for your quick review.

I see what you mean about backwards-compatibility. I thought of it that way initially, but there were two reasons I figured for doing it this way in the end:

1. Laziness, as it seemed to me useless duplication, a little too much too write and work with something like this:

  ```ruby
  proxy_uri = URI.parse(ENV['PAYMENT_GATEWAY_PROXY'])

  gateway = ActiveMerchant::Billing::SagePayGateway.new
  gateway.proxy_address = proxy_uri.host
  gateway.proxy_port = proxy_uri.port
  gateway.proxy_user = proxy_uri.user
  gateway.proxy_password = proxy_uri.password
  ```

  vs.

  ```ruby
  gateway = ActiveMerchant::Billing::SagePayGateway.new
  gateway.proxy = ENV['PAYMENT_GATEWAY_PROXY']
  ```

2. I saw that `@endpoint` used in https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/connection.rb#L37 is also looking for a String/URI and seemed more consistent passing the `proxy` this way also.

---

These being the reasons, and not being familiar with the code and recommendations in Activemerchant, for backward compatibility, should I favor:

A. Keep `:proxy_address`, `:proxy_port` and add `:proxy_user` and `:proxy_password` as in the first example

B. Keep `:proxy_address`, `:proxy_port` and add `:proxy` as a String to be parsed as an URI and extract the data from there if it's given rather than `:proxy_address` and `:proxy_port`.

I would go for **B**, but I guess it's better to ask first.

cristianstanescu:
@ntalbott this should now be backward compatible.

I've seen just a small thing after pushing. I've modified  the whitespace on this line by accident: https://github.com/activemerchant/active_merchant/pull/1894/files#diff-9f78979036d6e1787f93f58cf3ea6e34R65 Is it ok or should I commit a change to take it back? Thanks.

cristianstanescu:
@ntalbott Could you please have another look at this one? Thank you.

jasonwebster:

PR #1896, opened 2015-10-14
Micropayment: upgrade to new api

mrezentes:


lib/active_merchant/billing/gateways/micropayment.rb
```diff
@@ -81,11 +84,12 @@ def add_invoice(post, money, options)
...
       end
 
-      def add_payment_method(post, payment_method)
+      def add_payment_method(post, payment_method, options={})
         post[:firstname] = payment_method.first_name
         post[:surname] = payment_method.last_name
         post[:number] = payment_method.number
-        post[:cvc2] = payment_method.verification_value
+        post[:recurring] = options[:recurring] == true ? 1 : 0
+        post[:cvc2] = payment_method.verification_value unless post[:recurring] == 1
```
duff: Would things work with the `unless` clause?  So we always pass the `verification_value`?

lib/active_merchant/billing/gateways/micropayment.rb
```diff
@@ -81,11 +84,12 @@ def add_invoice(post, money, options)
         post[:project] = options[:project] || "sprdly"
       end
 
-      def add_payment_method(post, payment_method)
+      def add_payment_method(post, payment_method, options={})
         post[:firstname] = payment_method.first_name
         post[:surname] = payment_method.last_name
         post[:number] = payment_method.number
-        post[:cvc2] = payment_method.verification_value
+        post[:recurring] = options[:recurring] == true ? 1 : 0
```
duff: I'm wondering if the `recurring` flag is required.  For some gateways, I think we've passed the `recurring` flag if it's been specified and done nothing if it isn't specified.

lib/active_merchant/billing/gateways/micropayment.rb
```diff
@@ -103,9 +107,9 @@ def add_reference(post, authorization)
       end
 
       def commit(action, params)
-
         params[:testMode] = 1 if test?
         params[:accessKey] = @options[:access_key]
+        params[:apiKey] = @options[:api_key] || @am_api_key
```
duff: I'm thinking this could be simplified to:

```
params[:apiKey] = @options[:api_key] || "af1fd841af792f4c50131414ff76e004"
```

And then the creation of the `@am_api_key` on line 18 won't be needed.

lib/active_merchant/billing/gateways/micropayment.rb
```diff
@@ -159,8 +163,9 @@ def split_authorization(authorization)
         authorization.split("|")
       end
 
-      def authorization_from(response)
-        "#{response["sessionId"]}|#{response["transactionId"]}"
+      def authorization_from(response, request_params)
+        session_id = response["sessionId"] ? response["sessionId"] : request_params[:sessionId]
```
duff: Looks like there are some responses that don't have a `session_id`, eh?  This makes me :sob:.

duff:
@mrezentes - this is looking good.  I made a few comments; I'm thinking we're close!

lib/active_merchant/billing/gateways/micropayment.rb
mrezentes: Yes.  The capture uses the session id of the auth.

PR #1897, opened 2015-10-17
Added support for CUSTOM_ID2 field to BluePay transactions

ajporterfield:
Currently, the blue_pay gateway included in this activemerchant gem supports passing customer in the options hash, which maps to BluePay's CUSTOM_ID field.  BluePay has a second custom id that can be passed, but there's no current support for it in the gem.  This pull request adds support for it.  I manually verified that this new field does, in fact, display when passed in BluePay's backend.

I also fix the indentation of the add_customer_data method.

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_pay.rb
+++ b/lib/active_merchant/billing/gateways/blue_pay.rb
@@ -430,8 +430,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_customer_data(post, options)
-          post[:EMAIL]     = options[:email]
-          post[:CUSTOM_ID] = options[:customer]
+        post[:EMAIL]     = options[:email]
+        post[:CUSTOM_ID] = options[:customer]
+        post[:CUSTOM_ID2] = options[:custom_id2]
       end
 
       def add_duplicate_override(post, options)
--- a/test/remote/gateways/remote_blue_pay_test.rb
+++ b/test/remote/gateways/remote_blue_pay_test.rb
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class BluePayTest < Test::Unit::TestCase
+  include CommStub
+
   def setup
     Base.mode = :test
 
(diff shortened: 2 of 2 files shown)


girasquid:
Can you add a test that confirms the data gets sent over correctly? Do you want to include the key when options[:custom_id2] is `nil`, or should it be left out in that case?

ajporterfield:
Thanks for the review, @girasquid.  I'll work on the test.  Yes, I think including the key is fine even if options[:custom_id2] is nil.  The same would apply to options[:customer] and other optional options.

ajporterfield:
@girasquid, I added a test to ensure CUSTOM_ID2 is in the params returns by post_data.  Is this what you wanted confirmation on?  I had to call two private methods, so it's kind of ugly.  If there's a better way to get the request params from the response object please let me know.  Thanks!

girasquid:
Use CommStub instead: https://github.com/activemerchant/active_merchant/blob/fd505fe4708b4bc56622e7e02390665839af0d31/test/unit/gateways/stripe_test.rb#L183-L187

ajporterfield:
So I gave CommStub a shot.  Here's what I have locally...

    def test_custom_id2_option
      response = stub_comms(@gateway, :ssl_request) do
        @gateway.purchase(@amount, @credit_card, @options.merge(custom_id2: 'ABC123'))
      end.check_request do |method, endpoint, data, headers|
        assert_match("CUSTOM_ID2=ABC123", data)
      end
    end

But I'm getting this error when I run tests for this file (rake test TEST=test/remote/gateways/remote_blue_pay_test.rb).  Do you have any words of wisdom?

    Failure: <false> is not true.
    test_custom_id2_option(BluePayTest)
    active_merchant/test/comm_stub.rb:49:in `teardown'

Another thing I noticed is that all tests pass when I just run rake test - even if I change me test_custom_id2_option to check something that doesn't match.  Why is that?

girasquid:
Add a `respond_with` at the end, like in the Stripe example (just make it return a success response or something)

ajporterfield:
Thanks @girasquid.  I think they test is in good shape now.

ajporterfield:
Hey @girasquid, I wanted to follow up with you on this to see if you needed anything else from me on this PR?  Please let me know.  Thanks!

girasquid:
Looks good - will merge soon, sorry for the delay.

PR #1898, opened 2015-10-20
Netbilling: add order_id to user_info

mrezentes:


mrezentes:
@duff, looks like there are multiple ways to do this.  I went with user_data because they specifically mention adding key value pairs to user data.  It is one of the 3 options presented in the HS thread and the gateway documentation.  We will definitely need them to confirm it is getting to its destination.

PR #1900, opened 2015-10-21
Authorize.net CIM: adding avs_code to gateway response

feigner:
Updated the Authorize.net CIM gateway to include the avs result in the response.


jsdalton:
:+1: This fix would be great for us. We're using this library as part of a Spree application, and right now the AVS code is not showing up in the admin. It's definitely a bummer for our client to not be able to troubleshoot authorization failures within our application.

jasonwebster:

PR #1902, opened 2015-10-21
Braintree: bump gem to an official release (2.50.0)

girasquid:
@jnormore @ntalbott for review.

`braintree` 2.50.0 includes the apple pay functions that we're using, so we don't need to run off of a side branch anymore.


j-mutter:
:tada:

PR #1906, opened 2015-10-23
Worldpay commit checks update

ao:


lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -141,7 +141,23 @@ def commit(method, url, parameters=nil, options = {})
 
           if (raw_response != '')
             response = parse(raw_response)
-            success = !response.key?("httpStatusCode")
+            if type=='token'
+              success = response.key?("token")
+            elsif
```
ivanfer: prefer `else` here and below

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -141,7 +141,23 @@ def commit(method, url, parameters=nil, options = {})
...
+              if response.key?("httpStatusCode")
+                success = false
+              elsif
+                if type=='authorize'
+                  success = response["paymentStatus"]=="AUTHORIZED" ? true: false
```
ivanfer: don't need the ternary here

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -141,7 +141,23 @@ def commit(method, url, parameters=nil, options = {})
...
+                elsif type=='purchase'
+                  success = response["paymentStatus"]=="SUCCESS" ? true: false
+                elsif type=='capture' || type=='refund' || type=='void'
+                  success = true
```
ivanfer: These types of operations can never fail?

ao: They don't contain a response body, merely a 200 and skip the httpStatusCode check above

aprofeit:
@ao could you provide a little description of what the intention of these changes are? Do you have any docs to show why these needed to change?

ao:
@aprofeit These changes are being made because the current implementation does not check for full SUCCESS and if a payment fails during security checking/fraud, then it is still considered a SUCCESS.
The paymentStatus needs to be included to make sure it is 100% SUCCESS.

ao:
@aprofeit Any reason this was closed?

aprofeit:
Was merged in https://github.com/activemerchant/active_merchant/commit/86431f5ae539c594fa702431f63d5524f8f15476 :)

ivanfer:
Thanks @ao !

test/unit/gateways/worldpay_online_payments_test.rb
```diff
@@ -22,8 +22,8 @@ def setup
   end
 
   def test_successful_purchase
-    @gateway.expects(:ssl_request).returns(successful_token_response)
     @gateway.expects(:ssl_request).returns(successful_purchase_response)
+    @gateway.expects(:ssl_request).returns(successful_token_response)
```
duff: @mrezentes mentioned to me about the order of these lines not seeming to be the same as the order of the implementation, meaning that it looks like the code creates a token first.

I'm also wondering about how `expects` works and whether we need something like this:

https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/redsys_test.rb#L168

PR #1910, opened 2015-10-27
Enabling daily pay period to match docs

jamesprior:
:daily was listed as an available transaction period but not actually submitted in the request

### DIFF
--- a/lib/active_merchant/billing/gateways/payflow.rb
+++ b/lib/active_merchant/billing/gateways/payflow.rb
@@ -293,6 +293,7 @@ module ActiveMerchant #:nodoc:
       def get_pay_period(options)
         requires!(options, [:periodicity, :bimonthly, :monthly, :biweekly, :weekly, :yearly, :daily, :semimonthly, :quadweekly, :quarterly, :semiyearly])
         case options[:periodicity]
+          when :daily then 'Daily'
           when :weekly then 'Weekly'
           when :biweekly then 'Bi-weekly'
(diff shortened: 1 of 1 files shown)


aprofeit:
According to [these docs on p114](https://www.paypalobjects.com/webstatic/en_US/developer/docs/pdf/pp_payflowpro_xmlpay_guide.pdf), daily doesn't seem to be a valid option.

screen shot 2015-11-23 at 3 55 27 pm

Should we instead remove daily as an available `periodicity` option?

jamesprior:
Right - I was looking at https://developer.paypal.com/webapps/developer/docs/classic/payflow/recurring-billing/#using-the-inquiry-action-to-view-the-status-of-payments which does list daily as an option.  I tested it out and FWIW it does seem to accept Daily and charge on a daily basis.

We could also just go the other way and remove daily as a valid option

jasonwebster:

jamesprior:
I don't object to closing out of staleness but I'm not clear on what the outstanding issues are.  One set of docs didn't list it and another did.  As of a year ago it would work with daily as an option.  It sounds like this isn't affecting anyone, but if it sounds useful can you point me towards the next step to work on?

jasonwebster:
Hi! Thanks for commenting. I automatically closed all open PRs that were older than 14 months, so perhaps the message didn't suit them all. I'm glad it started the conversation up again though. I'll actually take a look at this, and see what's up.

jamesprior:
Sounds great, thanks!

PR #1911, opened 2015-10-28
Bug Fix: valid_checksum? crash on non-digits

PatrickTulskie:
When upgrading ActiveMerchant on a few different Rails applications I found that the new `CreditCard.valid_checksum?` method crashes when you pass it obfuscated or otherwise funky card numbers with non-digits in them.  The older 1.4.x version of this method would just return false.

Anyhow this fixes this the issue by just checking to see if there are non-digit characters in the card number before trying to compute the checksum.


aprofeit:
Thanks! I merged your changes, but just moved the check of non-digit characters to its own method so we can keep `valid_checksum?` the same as [where it came from](https://github.com/zendesk/luhn_checksum/blob/636efa92415ad7c47bec91062c149ba4dc9f5c9d/lib/luhn_checksum.rb#L32).

PatrickTulskie:
Makes total sense.  Thanks a bunch!

PR #1919, opened 2015-11-02
Stripe: Migrate from /refund to /refunds

matthelm:
This PR updates the Stripe endpoint we use for creating refunds. The previous endpoint returned the Charge object, while the new endpoint returns the Refund object, so as a result the test and fixtures have been updated. Will this change be a problem for clients using Active Merchant?

This PR also includes the metadata object in the `/refunds` request, which is added in this new endpoint.

*Before*
`https://api.stripe.com/v1/charges/{CHARGE_ID}/refund`

*After*
`https://api.stripe.com/v1/charges/{CHARGE_ID}/refunds`

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -115,9 +115,10 @@ module ActiveMerchant #:nodoc:
         add_amount(post, money, options)
         post[:refund_application_fee] = true if options[:refund_application_fee]
         post[:reverse_transfer] = options[:reverse_transfer] if options[:reverse_transfer]
+        post[:metadata] = options[:metadata] if options[:metadata]
 
         MultiResponse.run(:first) do |r|
-          r.process { commit(:post, "charges/#{CGI.escape(identification)}/refund", post, options) }
+          r.process { commit(:post, "charges/#{CGI.escape(identification)}/refunds", post, options) }
 
           return r unless options[:refund_fee_amount]
 
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -107,7 +107,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def void(identification, options = {})
-        commit(:post, "charges/#{CGI.escape(identification)}/refund", {}, options)
+        commit(:post, "charges/#{CGI.escape(identification)}/refunds", {}, options)
       end
 
       def refund(money, identification, options = {})
@@ -470,8 +470,6 @@ module ActiveMerchant #:nodoc:
           [response["id"], response["sources"]["data"].first["id"]].join("|")
         elsif method == :post && url.match(/customers\/.*\/cards/)
           [response["customer"], response["id"]].join("|")
-        elsif url.include?("refund") && response["refunds"]
-          response["refunds"]["data"].first["id"]
         else
           response["id"]
         end
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -1396,44 +1396,16 @@ class StripeTest < Test::Unit::TestCase
   def successful_void_response
     <<-RESPONSE
     {
-      "id": "ch_4IrhQMqukqu7C2",
-      "object": "charge",
-      "created": 1403816613,
-      "livemode": false,
-      "paid": true,
-      "amount": 50,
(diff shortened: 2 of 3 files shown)


matthelm:
@AnotherJoSmith mind giving this an initial review?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -115,9 +115,12 @@ def refund(money, identification, options = {})
         add_amount(post, money, options)
         post[:refund_application_fee] = true if options[:refund_application_fee]
         post[:reverse_transfer] = options[:reverse_transfer] if options[:reverse_transfer]
+        post[:metadata] = options[:metadata]
+
+        options.delete(:metadata)
```
AnotherJoSmith: Why do you need to delete the metadata from the options map here?

AnotherJoSmith:
Looks good :monorail:

matthelm:
Please review @j-mutter @aprofeit

test/unit/gateways/stripe_test.rb
```diff
@@ -342,8 +342,7 @@ def test_successful_refund
     assert response = @gateway.refund(@refund_amount, 'ch_test_charge')
     assert_success response
 
-    assert_equal 'ch_test_charge', response.authorization
-    assert response.test?
```
j-mutter: Eh, not a `test` anymore? Or you don't feel this is necessary?

j-mutter:
Remote tests?

test/unit/gateways/stripe_test.rb
matthelm: Stripe's Refund object doesn't have a `livemode` flag to use when populating `test?`. Suggestions?
https://stripe.com/docs/api#refund_object

j-mutter: Oh ok - TIL. I guess it just uses the associated charge_id to figure it out. This is fine, then

matthelm:
@j-mutter I ran the remote tests each on their own. The output is [here](https://gist.github.com/matthelm/0bbc060740577d009c0c). Many tests failed in the first three scenarios, but they were also failing on master. All the tests in `remote_stripe_test.rb` were green. Is there anything else I need to test?

```
remote_stripe_apple_pay_test.rb
remote_stripe_connect_test.rb
remote_stripe_emv_test.rb
remote_stripe_test.rb
```

j-mutter:
:+1:

ntalbott:
@matthelm @j-mutter I was surprised this didn't require a change/bump to the default Stripe API version? https://github.com/activemerchant/active_merchant/blob/e28815158657c312bd6de1d08867f4bd22dc0f59/lib/active_merchant/billing/gateways/stripe.rb#L426

j-mutter:
I'm pretty sure this endpoint predates the current minimum. @matthelm can you confirm?

matthelm:
@ntalbott I ran the remote tests, ensuring the default value `2015-04-07` was used, and all the tests were green. The last file in the gist is the test output. https://gist.github.com/matthelm/0bbc060740577d009c0c#file-remote_stripe_test-rb

Is there anything else you'd like me to do to verify this is sound?

ntalbott:
Nope, sounds good. So `2015-04-07` must allow either/or, eh?

matthelm:
Exactly

test/unit/gateways/stripe_test.rb
```diff
@@ -1396,44 +1396,16 @@ def successful_partially_refunded_response(options = {})
   def successful_void_response
     <<-RESPONSE
     {
-      "id": "ch_4IrhQMqukqu7C2",
-      "object": "charge",
-      "created": 1403816613,
-      "livemode": false,
-      "paid": true,
-      "amount": 50,
+      "id": "re_173VMpAWOtgoysogOSE7Hzss",
```
girasquid: voids create refund objects?

girasquid:
One question, but looks ok.

test/unit/gateways/stripe_test.rb
matthelm: Stripe's AM adapter implements the `void` method using `/refund`, so wouldn't it be creating refunds?

girasquid: Yup, you're right. Carry on.

matthelm:
I'm ready to merge, but am not an "Owner". Is somebody able to merge for me?

girasquid:
Done.

Issue #1153, opened 2014-04-30
Internalize handling of currency fractions


bslobodin:
Initial discussion started in https://github.com/Shopify/active_merchant/pull/1148.

Kagetsuki:
Just checking in here. I'll be monitoring this and will try to catch updates. I have two apps using AM with JPY, one using AM directly and hooked to WebPay and one Spree app with the Stripe JPY beta. I'll test both and submit patches if either gateway behaves irregularly with the fixes.

Kagetsuki:
I hate to bump this but I'm still kludging AM and it's a mess. Can we please get some official movement on this at some point in the near future?

trungdq88:
Please put a warning on the README to state clear that the currency support is limited. Or at least put in validations.

PR #1920, opened 2015-11-03
Stripe: Allow custom zero decimal currencies

duff:
The Stripe adapter has been overriding which currencies are zero decimal
currencies for awhile.  The parent Gateway class defines a default list
but Stripe has its own list:

https://support.stripe.com/questions/which-zero-decimal-currencies-does-stripe-support

The ability to override was broken in this commit:
https://github.com/activemerchant/active_merchant/commit/2a35543b4a8348cd115f4b2f138731e5f3c4a080

since the `non_fractional_currency?` method was changed to a class
method.

This commit allows us to override it again.  It also adds the VND
currency to Stripe's list since they now list that currency in their
docs.

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -57,7 +57,7 @@ module ActiveMerchant #:nodoc:
       include CreditCardFormatting
 
       DEBIT_CARDS = [ :switch, :solo ]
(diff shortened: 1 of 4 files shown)


duff:
@aprofeit, @girasquid, @bslobodin,

I know Shopify is using Stripe and is concerned about currencies ( https://github.com/activemerchant/active_merchant/issues/1153 ).

I'd sure love another set of :eyes: on this PR since I suspect you may be relying on the ability of the Stripe adapter to override the zero decimal currency list.

Thanks!

duff:
I decided to merge this one in to fix a customer attempting to use Stripe with the `TWD` currency.  We can certainly discuss more if needed.  Or modify as needed.

PR #1922, opened 2015-11-04
Fix variables in generated remote gateway tests

sdball:
From the generator the remote test has some `response.message` assertions that run into errors because those tests don't have a response variable set.


aprofeit:
Merged, thanks for fixing this!

PR #1927, opened 2015-11-11
Redsys: Support new SHA256 authentication method

davidsantoso:
@duff and @rwdaigle : This updates Redsys to support both the current SHA1 and the new SHA256 authentication methods. Since the credentials are the same (minus using a newly generated secret key) you'll need to add the key/value pair `:sha256 => true` during initialization to use the SHA256 method, otherwise it defaults to SHA1.

Hopefully this allows people to do a gradual update and quickly revert if something seems to be going wrong with their newly generated keys, etc. If Redsys really does completely stop SHA1 support on 11/23, then we can probably just remove all of the legacy SHA1 auth code and forgo the `:sha256 => true` during initialization.

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys.rb
+++ b/lib/active_merchant/billing/gateways/redsys.rb
@@ -24,6 +24,16 @@ module ActiveMerchant #:nodoc:
     #
     # Written by Samuel Lown for Cabify. For implementation questions, or
     # test access details please get in touch: sam@cabify.com.
+    #
+    # *** SHA256 Authentication Update ***
+    #
+    # Redsys is dropping support for the SHA1 authentication method. This
+    # adapter has been updated to work with the new SHA256 authentication
+    # method, however in your initialization options hash you will need to
+    # specify the key/value :signature_algorithm => "sha256" to use the
+    # SHA256 method. Otherwise it will default to using the SHA1.
+    #
+    #
     class RedsysGateway < Gateway
       self.live_url = "https://sis.sermepa.es/sis/operaciones"
       self.test_url = "https://sis-t.redsys.es:25443/sis/operaciones"
@@ -36,6 +46,8 @@ module ActiveMerchant #:nodoc:
       self.supported_cardtypes = [:visa, :master, :american_express, :jcb, :diners_club]
       self.homepage_url        = "http://www.redsys.es/"
       self.display_name        = "Redsys"
+
+      SHA256_SIGNATURE_VERSION = 'HMAC_SHA256_V1'
 
       CURRENCY_CODES = {
         "AED" => '784',
@@ -161,9 +173,11 @@ module ActiveMerchant #:nodoc:
       # * <tt>:secret_key</tt> -- The Redsys Secret Key. (REQUIRED)
       # * <tt>:terminal</tt> -- The Redsys Terminal. Defaults to 1. (OPTIONAL)
       # * <tt>:test</tt> -- +true+ or +false+. Defaults to +false+. (OPTIONAL)
+      # * <tt>:signature_algorithm</tt> -- +"sha256"+ Defaults to +"sha1"+. (OPTIONAL)
       def initialize(options = {})
         requires!(options, :login, :secret_key)
         options[:terminal] ||= 1
+        options[:signature_algorithm] ||= "sha1"
         super
       end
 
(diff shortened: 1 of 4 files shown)


ntalbott:
@davidsantoso instead of a boolean, I'd use a `:signature_algorithm` option that defaults to `:sha1`, similar to how Ogone handles it.

duff:
This is looking great @davidsantoso.  @ntalbott has a pretty good idea re: `signature_algorithm`.  

I'd also dig seeing remote tests running when using sha_256.

davidsantoso:
@ntalbott @duff right on. I'll push an update.

duff:
Perhaps we could leave our existing `RemoteRedsysTest` for the `sha1`.  And then have another remote test file for `sha_256`.  Just brainstorming. :smile:

davidsantoso:
@duff we certainly could. Although, they'd be exactly the same with the exception being during initialization. Let me know, happy to add another if that's better.

duff:
I'm not certain of the best way to handle it, but since the requests are so different depending on the `signature_algorithm`, it seems logical to me to have a set of tests for each.  Eventually when Redsys really removes support for sha_1, we'll be able to clean things up and remove the old tests.

davidsantoso:
@duff and @ntalbott: Alrighty made the updates and added separate unit/remote tests as well.

test/remote/gateways/remote_redsys_sha256_test.rb
```diff
@@ -0,0 +1,186 @@
+require 'test_helper'
+
+class RemoteRedsysTest < Test::Unit::TestCase
```
duff: I'm thinking this ought to be named more in line with the file.

test/unit/gateways/redsys_sha256_test.rb
```diff
@@ -0,0 +1,383 @@
+require 'test_helper'
+
+class RedsysTest < Test::Unit::TestCase
```
duff: I'm thinking this ought to be named more in line with the file too.

test/unit/gateways/redsys_sha256_test.rb
```diff
@@ -0,0 +1,383 @@
...
+  include CommStub
+
+  def setup
+    Base.mode = :test
+    @credentials = {
+      :login      => '091952713',
+      :secret_key => "qwertyasdf0123456789",
+      :terminal   => '1',
```
duff: I was thinking we'd see a `signature_algorithm`, no?

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -426,6 +472,48 @@ def clean_order_id(order_id)
           "%04d%s" % [rand(0..9999), cleansed[0...8]]
         end
       end
+
+      # *****************************************************
+      # SHA256 Specific Authentication/Validation Helpers
+      # *****************************************************
+
+      def sha256_authentication
+        @sig_algorithm_cache ||= @options[:signature_algorithm] == "sha256"
```
duff: I'm thinking we don't need to cache it.

duff:
@davidsantoso I made a few little comments.  I think this is close!

I'm psyched that you and Google translate were able to work together to push this one through.

test/remote/gateways/remote_redsys_sha256_test.rb
davidsantoso: :+1:

test/unit/gateways/redsys_sha256_test.rb
davidsantoso: :+1:

test/unit/gateways/redsys_sha256_test.rb
davidsantoso: So, I'm now realizing that I might have accidentally overwritten some changes I made to the unit tests. :disappointed: But yes, there should be a `signature_algorithm`.

lib/active_merchant/billing/gateways/redsys.rb
davidsantoso: Sounds good.

davidsantoso:
Thanks for the feedback @duff. I'll get those updated. Also, I realized that I actually accidentally overwrote some changes I made to the SHA256 unit tests, so it might take me a bit to get those request/response strings updated again.

duff:
:sweet_potato: !

davidsantoso:
@duff ready for a look again when you have a minute. Fixed up the issues you brought up has in the test suites.

duff:
@davidsantoso :+1: :+1:

PR #1929, opened 2015-11-12
Adding 3DSecure to Worldpay Online Payments gateway

ao:
Attempting to add 3DSecure to Worldpay Online Payments gateway.
Was not sure how to implement this as couldn't find any docs about it.

Please advise if this is wrong.


ntalbott:
@ao I don't see the `authorise3DSOrder` method being called anywhere?

ActiveMerchant in general doesn't directly support 3DS, but adding passthrough fields to help with it is OK.

test/remote/gateways/remote_worldpay_online_payments_test.rb
```diff
@@ -160,4 +160,38 @@ def test_invalid_login
...
+            <input type="submit" />
+          </form>
+        </body>
+        </html>
+      )
+    end
+    puts "Test 3D-secure via `open 3ds.html`"
+    puts "View results at http://requestb.in/#{bin}?inspect"
+    puts "Finalize with: `curl -v -d PaRes='' '#{response.params["redirectURL"]}'`"
+=end
```
ao: @ntalbott `authorise3DSOrder` would be called right at the end, but I'm not sure how/where to add it as I don't think this process can be automated, hence the manual section above.

Thoughts?

ntalbott: @ao 3DS is out of scope for ActiveMerchant, so it should probably be in your own code or a separate library.

ao: @ntalbott How is it being done for these? https://docs.shopify.com/support/settings/checkout-and-payment/does-shopify-support-3d-secure

How could we set this up as well?

girasquid: These are integrated on the Shopify side using https://github.com/activemerchant/offsite_payments; customers are redirected to the (offsite) gateway screen to complete their transaction and do the 3D secure challenge. 3DS support is out of scope for ActiveMerchant (which deals with direct connections to credit card gateways), but if you'd like to integrate it you could make a PR on `offsite_payments`.

PR #1931, opened 2015-11-17
CashNet: Fix URI parse on 302 response

davidsantoso:
@duff and/or @rwdaigle: This fixes the exception we've noticed a few times when CashNet returns a 302 response and the adapter follows it. Not standard AM so let me know if you guys have other thoughts on how you'd like to handle a 302 in this case. But, encoding the `location` should fix any bogus characters in that URL.


rwdaigle:
@davidsantoso no tests so not sure what case this is fixing?

If this is intended to fix bogus UTF-8 characters, does it do that? It seems like URI encoding might obscure the URL in some other ways?

PR #1932, opened 2015-11-18
Omise: Add a new optional api_version config.

zdk:
Hi,

Regarding new Omise API version, I have created the relevant update in this PR.

The main change is adding the configurable `api_version` for the Omise gateway.
So, `api_version` can be set in activemerchant if it's needed.

Following are changes made in this PR:
* Add `Omise-Version` header if `api_version` is set.
* Remove unused API_VERSION variable.
* Add Ruby/#{RUBY_VERSION} to User-Agent.
* Modify remote test and refactor.

All unit tests:
```
active_merchant|omise-v2 ⇒ bundle exec rake test:units TEST=test/unit/gateways/omise_test.rb
/Users/di/.rbenv/versions/2.2.2/bin/ruby -I"lib:test" -rubygems -w -I"/Users/di/.rbenv/versions/2.2.2/lib/ruby/2.2.0" "/Users/di/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rake/rake_test_loader.rb" "test/unit/gateways/omise_test.rb"
Loaded suite /Users/di/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rake/rake_test_loader
Started
................................

Finished in 0.022835 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
32 tests, 53 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1401.36 tests/s, 2321.00 assertions/s
```

Omise remote tests:

```
active_merchant|omise-v2 ⇒ bundle exec rake test:remote TEST=test/remote/gateways/remote_omise_test.rb
/Users/di/.rbenv/versions/2.2.2/bin/ruby -I"lib:test" -rubygems -w -I"/Users/di/.rbenv/versions/2.2.2/lib/ruby/2.2.0" "/Users/di/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rake/rake_test_loader.rb" "test/remote/gateways/remote_omise_test.rb"
Loaded suite /Users/di/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rake/rake_test_loader
Started
............

Finished in 12.280202 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
12 tests, 29 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
0.98 tests/s, 2.36 assertions/s
```

Hope this makes sense.
Thanks!

### DIFF
--- a/lib/active_merchant/billing/gateways/omise.rb
+++ b/lib/active_merchant/billing/gateways/omise.rb
@@ -176,12 +176,14 @@ module ActiveMerchant #:nodoc:
 
       def headers(options={})
         key = options[:key] || @secret_key
-        {
+        h = {
           'Content-Type'    => 'application/json;utf-8',
           'User-Agent'      => "ActiveMerchantBindings/#{ActiveMerchant::VERSION} Ruby/#{RUBY_VERSION}",
           'Authorization'   => 'Basic ' + Base64.encode64(key.to_s + ':').strip,
           'Accept-Encoding' => 'utf-8'
         }
+        h['Omise-Version'] = @api_version if @api_version
+        h
       end
 
       def url_for(endpoint)
@@ -194,10 +196,8 @@ module ActiveMerchant #:nodoc:
 
       def https_request(method, endpoint, parameters=nil, options={})
         raw_response = response = nil
-        req_headers  = headers(options)
         begin
-          req_headers['Omise-Version'] = @api_version if @api_version
-          raw_response = ssl_request(method, url_for(endpoint), post_data(parameters), req_headers)
+          raw_response = ssl_request(method, url_for(endpoint), post_data(parameters), headers(options))
           response = parse(raw_response)
         rescue ResponseError => e
           raw_response = e.response.body
--- a/lib/active_merchant/billing/gateways/omise.rb
+++ b/lib/active_merchant/billing/gateways/omise.rb
@@ -176,14 +176,13 @@ module ActiveMerchant #:nodoc:
 
       def headers(options={})
         key = options[:key] || @secret_key
-        h = {
+        {
           'Content-Type'    => 'application/json;utf-8',
+          'Omise-Version'   => @api_version || "2014-07-27",
           'User-Agent'      => "ActiveMerchantBindings/#{ActiveMerchant::VERSION} Ruby/#{RUBY_VERSION}",
           'Authorization'   => 'Basic ' + Base64.encode64(key.to_s + ':').strip,
           'Accept-Encoding' => 'utf-8'
         }
-        h['Omise-Version'] = @api_version if @api_version
-        h
       end
 
       def url_for(endpoint)
--- a/lib/active_merchant/billing/gateways/omise.rb
+++ b/lib/active_merchant/billing/gateways/omise.rb
@@ -3,7 +3,6 @@ require 'active_merchant/billing/rails'
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class OmiseGateway < Gateway
-      API_VERSION = '1.0'
       API_URL     = 'https://api.omise.co/'
       VAULT_URL   = 'https://vault.omise.co/'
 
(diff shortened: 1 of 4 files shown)


zdk:
cc @girasquid

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -194,8 +194,10 @@ def post_data(parameters)
 
       def https_request(method, endpoint, parameters=nil, options={})
         raw_response = response = nil
+        req_headers  = headers(options)
         begin
-          raw_response = ssl_request(method, url_for(endpoint), post_data(parameters), headers(options))
+          req_headers['Omise-Version'] = @api_version if @api_version
```
girasquid: What happens when this doesn't get sent? Would you automatically get the newest version, or 1.0, or .. ?

lib/active_merchant/billing/gateways/omise.rb
```diff
@@ -194,8 +194,10 @@ def post_data(parameters)
 
       def https_request(method, endpoint, parameters=nil, options={})
         raw_response = response = nil
+        req_headers  = headers(options)
```
girasquid: Instead of doing this here, why not modify the `headers` method to add the `Omise-Version` header for you (if available)? You'd be able to get away without modifying the body of this method at all, in that case

girasquid:
I think you're doing the work to modify headers in a bit of a strange place - but once that's sorted out, this looks good.

lib/active_merchant/billing/gateways/omise.rb
j-mutter: :+1: I would just add it to the `headers` method and just set a default value if needed as @girasquid mentioned below

lib/active_merchant/billing/gateways/omise.rb
zdk: The Omise api expects `Omise-Version` header only when user explicitly set this header.
So, when this doesn't get sent, Omise users expect the api version that has been set in their dashboard ( the setting in their account ).

lib/active_merchant/billing/gateways/omise.rb
zdk: I think that's a good option too, if so, I may have to write something like this.

```
def headers(options={})
   key = options[:key] || @secret_key
   h = {
     'Content-Type'    => 'application/json;utf-8',
     'User-Agent'      => "ActiveMerchantBindings/#{ActiveMerchant::VERSION} Ruby/#{RUBY_VERSION}",
     'Authorization'   => 'Basic ' + Base64.encode64(key.to_s + ':').strip,
     'Accept-Encoding' => 'utf-8'
   }
   h['Omise-Version'] = @api_version if @api_version
   h
 end
```
Does this look ok?

j-mutter: Yeah something like that would work

zdk: Ok, thanks. I just made the fix. 
Let me know if you want me to squash.

ntalbott:
Hey folks, hate to jump in so late, but I'd prefer ActiveMerchant always specified the version to Omise, and then allow overriding it via the `@options` hash. So the default is to lock it to a known-good version.

See the Stripe gateway for an example of this; it was changed to work that way in the last year since a floating version causes surprising compatibility problems.

zdk:
@ntalbott that's also nice idea, to have a known-good state. I'll change it, so it will have a default version.

zdk:
Hi guys,
Does everything look OK now?

girasquid:
Looks good - thanks!

zdk:
@girasquid Awesome, thanks!

PR #1934, opened 2015-11-18
Bridgepay: Add echeck support

davidsantoso:
@duff and/or @rwdaigle: This adds eCheck support to the Bridgepay gateway.

Let me know if you have any thoughts/questions/concerns.

### DIFF
--- a/lib/active_merchant/billing/gateways/bridge_pay.rb
+++ b/lib/active_merchant/billing/gateways/bridge_pay.rb
@@ -6,8 +6,8 @@ module ActiveMerchant #:nodoc:
       self.display_name = "BridgePay"
       self.homepage_url = "http://www.bridgepaynetwork.com/"
 
-      self.test_url = "https://gatewaystage.itstgate.com/SmartPayments/transact.asmx/ProcessCreditCard"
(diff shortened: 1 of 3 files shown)


rwdaigle:
All around looks great! Few comments:

* I'd use this as an opportunity to add a `scrub` method. Add account/routing number to that scrubbing if that's how other adapters work too?
* Does Bridgepay support storing the card? If so, we should use this as an opp. to add `store` as well.

davidsantoso:
Ahhh good catch with the scrub- definitely needs to be updated (I think)!

I just checked the PDF docs and they don't currently support `store` so I think we're ok on that end.

davidsantoso:
@rwdaigle ready for round 2.

rwdaigle:
Unit tests:

```
Finished in 14.523776 seconds.
--------------------------------------------------------------------------------
3041 tests, 14123 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------
209.38 tests/s, 972.41 assertions/s
```

Remote tests:

```
Finished in 37.505004 seconds.
--------------------------------------------------------------------------------
100% passed
--------------------------------------------------------------------------------
0.43 tests/s, 0.99 assertions/s
```

PR #1935, opened 2015-11-18
Allow to use this gem with Rails 5

rafaelfranca:
The only required changes were to remove an unused require that was removed in Rails 5 and to change the gemspec to allow this gem to be installed with that version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ gemfile:
 - Gemfile.rails40
 - Gemfile.rails41
 - Gemfile.rails42
+- Gemfile.rails5
 
 matrix:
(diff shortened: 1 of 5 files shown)


rafaelfranca:
cc @duff

ntalbott:
@rafaelfranca is Rails 5 released yet? Usually we don't add a new Rails version until it's in the wild.

ntalbott:
@rafaelfranca if you're comfortable supporting it, :+1:.

rafaelfranca:
Sure! I can support it.

j-mutter:
:heart_eyes:

PR #1936, opened 2015-11-19
Do not allow to install the gem with Ruby 1.9

rafaelfranca:
The support to Ruby 1.9 was removed at 8a6c6ff626ebd292dc50e8e1c074aab49956a42d so we should not allow users to install the gem with that version.


girasquid:
I forgot the '#' on the commit message :disappointed: - this has been merged. Closing.

PR #1937, opened 2015-11-23
Fix detecting test refunds for Stripe

aprofeit:
#### Problem
Currently all Stripe refund responses return `test? => false`. This changed [when we switched](https://github.com/activemerchant/active_merchant/commit/f388b95d3495bbbddbcf98dcb3e60624339bf2a3) from using the `/charge/refund` endpoint to the `/charge/refunds` endpoint.

The reason why `test?` always returns `false` is the refund object that's returned by Stripe does not have a `livemode` key which we use to determine whether a response is test or not.

#### Changes
This PR changes the refund action to always include the charge in the refund response. Then, when checking whether a response is test or not, if there's no `livemode` key, check if there's a `charge` object we can look for the `livemode` key in.

#### Review
@berkcaputcu @ivanfer

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -115,6 +115,7 @@ module ActiveMerchant #:nodoc:
         post[:refund_application_fee] = true if options[:refund_application_fee]
         post[:reverse_transfer] = options[:reverse_transfer] if options[:reverse_transfer]
         post[:metadata] = options[:metadata] if options[:metadata]
+        post[:expand] = [:charge]
 
         MultiResponse.run(:first) do |r|
           r.process { commit(:post, "charges/#{CGI.escape(identification)}/refunds", post, options) }
@@ -283,7 +284,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_expand_parameters(post, options)
-        post[:expand] = Array.wrap(options[:expand])
+        post[:expand] ||= []
+        post[:expand].concat(Array.wrap(options[:expand]).map(&:to_sym)).uniq!
       end
 
       def add_customer_data(post, options)
@@ -453,7 +455,7 @@ module ActiveMerchant #:nodoc:
         Response.new(success,
           success ? "Transaction approved" : response["error"]["message"],
           response,
-          :test => response.has_key?("livemode") ? !response["livemode"] : false,
+          :test => response_is_test?(response),
           :authorization => authorization_from(success, url, method, response),
           :avs_result => { :code => avs_code },
           :cvv_result => cvc_code,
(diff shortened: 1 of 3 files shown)


aprofeit:
To explain a little more, the Stripe API supports [expanding params](https://stripe.com/docs/api#expanding_objects), which allows an associated object to be included in the response. `charge` is a expandable on [`refund` responses](https://stripe.com/docs/api#refunds).

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -283,7 +284,8 @@ def add_destination(post, options)
       end
 
       def add_expand_parameters(post, options)
-        post[:expand] = Array.wrap(options[:expand])
+        post[:expand] ||= []
+        post[:expand].concat(Array.wrap(options[:expand]).map(&:to_sym)).uniq!
```
berkcaputcu: not sure where this is used but this can be a `Set` instead of `Array`

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -492,6 +494,16 @@ def json_error(raw_response)
         }
       end
 
+      def response_is_test?(response)
+        if response.has_key?('livemode')
+          !response['livemode']
+        elsif response['charge'].is_a?(Hash) && response['charge'].has_key?('livemode')
+          !response['charge']['livemode']
+        else
+          false
```
berkcaputcu: should we log something if it falls back to this case? is it expected that it falls back to `else`?

test/remote/gateways/remote_stripe_test.rb
```diff
@@ -105,6 +105,7 @@ def test_successful_refund
     assert_success response
     assert response.authorization
     assert refund = @gateway.refund(@amount - 20, response.authorization)
+    assert refund.test?
```
berkcaputcu: :heart:

lib/active_merchant/billing/gateways/stripe.rb
aprofeit: I see what you mean, unfortunately `Array.wrap` doesn't play nicely with `Set`s, ends up building an array with the `Set` as its element. Would rather keep using this nice `Array.wrap` to handle `options[:expand]` being `nil`, `'charge'`, `['charge', 'somethig_else']`.
```
irb(main):001:0> Array.wrap(Set.new(['a']))
=> [#<Set: {"a"}>]
```

berkcaputcu: I see, `Array.wrap` is a tricky one. This comment is not in the scope of this PR anyway, so :+1:

lib/active_merchant/billing/gateways/stripe.rb
aprofeit: Logging would be nice, but there's no existing logging currently. Gonna keep this as is to avoid changing behaviour.

ivanfer:
:rocket:

girasquid:
:+1:

PR #1940, opened 2015-11-24
Fix detecting test voids for Stripe

methodmissing:
#### Problem

Currently all Stripe void responses return test? => false. This changed when we switched from using the `/charge/refund` endpoint to the `/charge/refunds` endpoint. This PR is a superset of https://github.com/activemerchant/active_merchant/pull/1937 and the semantics for the charge  [expansion](https://stripe.com/docs/api#refunds) is similar given both `refund` and `void` actions call into the same endpoint and receives a `refund object` as response.

The reason why test? always returns false is the refund object that's returned by Stripe does not have a `livemode` key which we use to determine whether a response is test or not.

#### Changes

This PR changes the `void` action to always include the charge in the refund response. Then, when checking whether a response is test or not, if there's no `livemode` key, check if there's a charge object we can look for the `livemode` key in.

#### Review

@aprofeit @girasquid

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -106,7 +106,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def void(identification, options = {})
-        commit(:post, "charges/#{CGI.escape(identification)}/refunds", {}, options)
+        post = {}
+        post[:expand] = [:charge]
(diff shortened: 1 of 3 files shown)


aprofeit:
:+1:

girasquid:
:+1:

aprofeit:
:balloon:

PR #1941, opened 2015-11-25
Add token support for KomojuGateway

shioyama:
This adds a `store` method to the Komoju gateway (see #1638) so that merchants can enable payment        profiles. It also allows that stored token string to be passed in place of payment details when purchasing.

For more information see the Komoju docs on Tokens: https://docs.komoju.com/api/overview


jasonwebster:

Issue #1822, opened 2015-07-31
Elavon - Send cardholder's IP on purchase and authorization


cjoudrey:
We are currently not sending the cardholder's IP when performing an auth or sale.

[According to the docs](https://www.myvirtualmerchant.com/VirtualMerchant/download/developerGuide.pdf) this can be done by passing `ssl_cardholder_ip`.

It should be an easy fix that we can do in [`def purchase`](https://github.com/activemerchant/active_merchant/blob/2897332fc998309683558874840767253389db68/lib/active_merchant/billing/gateways/elavon.rb#L72) and [`def authorize`](https://github.com/activemerchant/active_merchant/blob/2897332fc998309683558874840767253389db68/lib/active_merchant/billing/gateways/elavon.rb#L94).

Example of another gateway that passes IP: https://github.com/activemerchant/active_merchant/blob/2897332fc998309683558874840767253389db68/lib/active_merchant/billing/gateways/iats_payments.rb#L80-L83

cc @shawnfrancis

shawnfrancis:
My client wrote me an email on the Evalon integration:

Shawn, also noticed one of their other Payment Gateway filters won't work - looks like Billing Address Country doesn't flow through from Shopify to Elavon on transactions (Country looks like it's left blank), so any fraud filter people would setup to restrict customers with certain Billing Countries will automatically decline every credit card transaction that comes through.

The last part to verify is if CVV2 and AVS codes pass through correctly from Shopify.

Sorry about all this - it looks like Elavon isn't setup to be a suitable Payment Gateway for Shopify.

jordanliddle:
@cjoudrey I'll look at setting this up today

PR #1942, opened 2015-11-26
Elavon: Include IP address

aprofeit:
#### Changes
Includes the IP in the request to Elavon. Closes https://github.com/activemerchant/active_merchant/issues/1822

#### Testing
I was able to get some credentials from Elavon to run the remote tests, but am waiting on confirmation they're cool with us committing the credentials to AM before I add them.

From my remote test runs I was able to confirm that IP shows up in Elavon.

elavon

#### Review
@cjoudrey @girasquid

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -82,6 +82,7 @@ module ActiveMerchant #:nodoc:
         add_address(form, options)
         add_customer_data(form, options)
         add_test_mode(form, options)
+        add_ip(form, options)
         commit(:purchase, money, form)
       end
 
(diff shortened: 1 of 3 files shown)


cjoudrey:
Is there anything we should assert in the remote test?

aprofeit:
@cjoudrey don't think so, the IP isn't included in the response, so no way to assert that. I think it's enough that we just include the IP in the existing requests and assert they continue to function as expected. The unit tests cover asserting that the param is present in the request as we expect.

girasquid:
Makes sense :+1:

cjoudrey:
Cool! :shipit:

j-mutter:
:+1:

PR #1944, opened 2015-11-30
TransFirst: Add support for ACH and more operations

davidsantoso:
@duff and/or @rwdaigle this updates the TransFirst Transaction Central adapter. Could things to note:

1. There's no `authorize` with the Transaction Central API- all "Sale" actions are first authorizations, then at midnight TransFirst will automatically settle them.
1. Since there is so `authorize` there's also no `capture`.
1. Unfortunately, TransFirst accounts (test or prod) are unable to immediately settle credit card sale transactions, so I can't write a remote refund tests because all test `purchases` aren't settled and they'll always fail. But, I was able to get those responses for the unit tests so at least they're there. Remote tests are just a bit lacking.

Let me know if you have any other question/concerns/comments.

### DIFF
--- a/lib/active_merchant/billing/gateways/trans_first.rb
+++ b/lib/active_merchant/billing/gateways/trans_first.rb
@@ -1,7 +1,8 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class TransFirstGateway < Gateway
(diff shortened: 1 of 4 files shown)


rwdaigle:
Merged

PR #1948, opened 2015-12-01
Element: Add gateway

davidsantoso:
@duff and/or @rwdaigle: Adding the Element gateway. Pretty standard, nothing really to point out.


lib/active_merchant/billing/gateways/element.rb
```diff
@@ -0,0 +1,267 @@
...
+        response["expressresponsemessage"]
+      end
+
+      def authorization_from(response)
+        response["transaction"]["transactionid"]
+      end
+
+      def error_code_from(response)
+        unless success_from(response)
+          # TODO: lookup error code for this response
```
rwdaigle: Should this be filled in?

davidsantoso: Ahh, good catch. Actually not supposed to be filled in. I'm just going to :hocho: it.

lib/active_merchant/billing/gateways/element.rb
```diff
@@ -0,0 +1,267 @@
...
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((<AccountID>).+?(</AccountID>))i, '\1[FILTERED]\2').
```
davidsantoso: I don't think it is come to think of it. Updating...

davidsantoso:
@rwdaigle it does. Should `store` be done by default in addition to the other actions if it's available?

lib/active_merchant/billing/gateways/element.rb
girasquid: :+1: for not scrubbing if it's not actually PCI-level sensitive - accountID can be really handy for debugging if you need to look something up in logs and send to support staff somewhere

davidsantoso: :+1:

rwdaigle:
> Should store be done by default in addition to the other actions if it's available?

I believe that is Spreedly's current approach, yes (let me know if we need to update our internal docs to reflect this if it's unclear)?

davidsantoso:
@rwdaigle updated with `store` and added the ability to charge via a an Element `PaymentAccountID`. Let me know if you have any other feedback!

lib/active_merchant/billing/gateways/element.rb
```diff
@@ -0,0 +1,327 @@
...
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((<AccountToken>).+?(</AccountToken>))i, '\1[FILTERED]\2').
+          gsub(%r((<CardNumber>).+?(</CardNumber>))i, '\1[FILTERED]\2').
+          gsub(%r((<CVV>).+?(</CVV>))i, '\1[FILTERED]\2')
```
rwdaigle: I believe we need `RoutingNumber` and `AccountNumber` here as well?

davidsantoso:
@rwdaigle wow good catch! Definitely need to scrub those. I just updated again!

rwdaigle:
> wow good catch!

10/10 would PR again?

rwdaigle:
Closed w/ https://github.com/activemerchant/active_merchant/commit/7519bc780e704e8e801a5483cb4a235d6b9e55f5

PR #1953, opened 2015-12-07
Clearhaus: Add tests for signed requests

anellis:


anellis:
@duff these are just a few tests for the whole purchase without cvv thing

PR #1956, opened 2015-12-08
Clearhaus: Add text_on_statement option

anellis:
@duff and please take a look at this one as well!


duff:
@anellis This is looking good.  I noticed one little thing.  I think this line:

https://github.com/anellis/active_merchant/blob/clearhaus_text_on_statement/lib/active_merchant/billing/gateways/clearhaus.rb#L150

should be removed since `description` shouldn't be mapped to that field.  Looks like that's been there since the gateway was added.

anellis:
@duff oh good catch! maybe it actually should be mapped to description? There doesn't seem to be another "description" type option so I can't think of why not?

anellis:
@duff updated!

PR #1958, opened 2015-12-09
Bridgepay: Add ability to store cards and pay with token

anellis:


anellis:
@duff Would you mind taking a look? I did have to upgrade to the new bridgepay endpoint to get the store functionality but it seems compatible

duff:
@anellis Looks great!  Think it's worth having any unit tests for it?

anellis:
@duff good point! Unit tests are now included.

PR #1959, opened 2015-12-09
Payeezy: Add support for echecks

davidsantoso:
@duff and/or @rwdaigle: Adding eCheck support to the Payeezy gateway.


rwdaigle:
@davidsantoso need to update `scrub`

davidsantoso:
@rwdaigle ahh gets me every time. Alrighty, just pushed an update to include scrubbing.

PR #1960, opened 2015-12-10
AndroidPay: Initial support

mrezentes:
@rwdaigle  This adds initial support for android pay.  Includes adjustments to the gateways currently using the NetworkTokenizationCreditCard.


rwdaigle:
@jnormore @ntalbott would love a second set of :eyes: on this.

Basically, an Android Pay produced `NetworkTokenizationCreditCard` has all the same values of an Apple Pay one. However, some gateways have apple or android specific field names, so we need to know what type of tokenized card we are dealing with. To do this, we add a simple, and explicit, `source` attribute to the `NetworkTokenizationCreditCard` that each gateway adapter can use to determine how to send the payment.

Seems simple and straight forward?

girasquid:
Makes sense to me.

jnormore:
You read my mind ;)

:shipit:

PR #1961, opened 2015-12-10
Authorize.net: add eCheck credit

davidsantoso:
@rwdaigle adding eCheck credit support to Authorize.net. Two things:

1. I talked to Authorize.net support and they have limited support for eChecks in sandbox. So even though the remote test returns an error, according to Authorize.net support, if we get back the message, "The transaction is currently under review" then it should process in production. So pretty janky but I don't think there's much we can do about that.
1. If you run the full remote tests, the Apple Pay tests fail due to an expired credit card. It shouldn't cause any issues with this AFAICT. Not quite sure if that's something that needs to be fixed with this.


rwdaigle:
@davidsantoso so it doesn't look like there was any added functionality, just some tests to confirm the behavior?

davidsantoso:
@rwdaigle correct. It was actually already supported but the remote test is more or less confirmation.

PR #1963, opened 2015-12-11
Add scrubbing support to fat zebra

bruno:
This pull request adds support for scrubbing on Fat Zebra.


bruno:
I messed up the merge from upstream, so removing this and replaced it with #2037 (from a clean `master`).

PR #1964, opened 2015-12-11
Encodes Cashnet's redirect_to response before parsing it

hoenth:
In some cases, the redirect to url contained non-parsable characters. By encoding them before attempting to parse, we will reduce the likelihood that the get request fails after a successful post

### DIFF
--- a/lib/active_merchant/billing/gateways/cashnet.rb
+++ b/lib/active_merchant/billing/gateways/cashnet.rb
@@ -132,7 +132,8 @@ module ActiveMerchant #:nodoc:
         if (200...300).include?(response.code.to_i)
           return response.body
         elsif 302 == response.code.to_i
-          return ssl_get(URI.parse(response['location']))
+          encoded_url = URI.encode(response['location'])
+          return ssl_get(URI.parse(encoded_url))
         end
         raise ResponseError.new(response)
       end
--- a/lib/active_merchant/billing/gateways/cashnet.rb
+++ b/lib/active_merchant/billing/gateways/cashnet.rb
@@ -132,7 +132,7 @@ module ActiveMerchant #:nodoc:
         if (200...300).include?(response.code.to_i)
           return response.body
         elsif 302 == response.code.to_i
-          encoded_url = URI.encode(response['location'])
+          encoded_url = URI::Parser.new.escape(response['location'])
           return ssl_get(URI.parse(encoded_url))
         end
         raise ResponseError.new(response)
--- a/lib/active_merchant/billing/gateways/cashnet.rb
+++ b/lib/active_merchant/billing/gateways/cashnet.rb
@@ -33,6 +33,7 @@ module ActiveMerchant #:nodoc:
           :merchant_gateway_name
         )
         options[:default_item_code] ||= "FEE"
+        options[:default_custcode] ||= "ActiveMerchant/#{ActiveMerchant::VERSION}"
         super
       end
 
@@ -80,7 +81,8 @@ module ActiveMerchant #:nodoc:
         post[:operator]       = @options[:operator]
         post[:password]       = @options[:password]
         post[:station]        = (@options[:station] || "WEB")
-        post[:custcode]       = (@options[:custcode] || "ActiveMerchant/#{ActiveMerchant::VERSION}")
+        post[:itemcode]       = (@options[:item_code] || @options[:default_item_code])
+        post[:custcode]       = (@options[:custcode] || @options[:default_custcode])
         post.merge(parameters).collect { |key, value| "#{key}=#{CGI.escape(value.to_s)}" }.join("&")
       end
 
--- a/test/unit/gateways/cashnet_test.rb
+++ b/test/unit/gateways/cashnet_test.rb
@@ -141,6 +141,15 @@ class Cashnet < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
+  def test_location_with_non_parsable_characters_gets_encoded_before_parsing
+    gateway = CashnetGateway.new(merchant: 'X', operator: 'X', password: 'test123', merchant_gateway_name: 'X', custcode: "TheCustCode")
+    stub_comms(gateway, :raw_ssl_request) do
+      stub_comms(gateway, :ssl_get) do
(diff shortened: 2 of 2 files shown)


duff:
Hi @hoenth, I'm thinking a test demonstrating the issue would be pretty helpful.

rwdaigle:
@hoenth if this is still an issue, would you mind submitting a corresponding test or close the issue if it's not a priority for you? Thanks!

hoenth:
@rwdaigle Sorry for the delay. Took me a bit to understand how to best stub it. Let me know if you need anything further

rwdaigle:
@hoenth do you mind rebasing your commits on master, they seem of date?

hoenth:
OK, all done

rwdaigle:
Sorry for the back and forth on this, @hoenth, but any idea how to fix this? This is on Ruby 2.1.

```
/Users/ryan/dev/projects/active_merchant/lib/active_merchant/billing/gateways/cashnet.rb:135:in `handle_response': warning: URI.escape is obsolete
```

rwdaigle:
@hoenth got it, looks like `CGI::escape` will suffice.

rwdaigle:
@hoenth not sure if you want to look into this or not, but it looks like the mocking approach we took has a blind spot. We had to revert this PR. See the [stack trace here](https://github.com/activemerchant/active_merchant/pull/2015#issuecomment-181553418)

hoenth:
@rwdaigle - It seems my fix was the same as @davidsantoso  (https://github.com/activemerchant/active_merchant/pull/1931).  The problem was with the commit to change from `URI.encode(response['location'])` to `CGI.escape(response['location'])`. I'm not seeing a separate commit for that to reference, but this PR no longer includes that. 

In essence, all this pull request does now, in comparison to 1931 is add a test for url encoding. 

When I run the remote tests in ruby 2.1.6, I'm not seeing any `URI.escape is obsolete` warnings. In what context were you seeing them?

So if we use this original PR, we should be fine. Tests will be added and we won't get the `request_uri' for #<URI::Generic:0x007f6ee35b36d8>` error that we get with `CGI::escape`

rwdaigle:
@hoenth 

> When I run the remote tests in ruby 2.1.6, I'm not seeing any URI.escape is obsolete warnings. In what context were you seeing them?

Can you run the whole suite with `rake` - most warnings show up then.

hoenth:
@rwdaigle, it's true I see the warnings there. But according to this: http://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape/13059657#13059657, the only better solution is to use the Addressable gem, which I would guess is not something that AM wants to include just to fix this one issue. According to the same SO answer (which I don't have enough knowledge on the inner workings of URI or CGI to comment on), CGI.escape should only be used for the parameters, not the entire url. 

If we can't have this change, then transactions processed through Cashnet are at risk of raising an error (from a bad URI) and our systems wont pick up that the transaction went through. Unfortunately, I have no input into Cashnet to have them change their process, we are only a consumer of it for our clients. 

Perhaps others can suggest another library that can be used to encode the URI and parameters that Cashnet sends back, that won't raise the warning or the error we saw with CGI.escape.

rwdaigle:
@hoenth I dug into this a little more and I actually don't think the issue is `URI` vs. `CGI`.

If you look at the exception, the URI type that gets passed to the AM connection is `URI::Generic`, not `URI::HTTPS`, which is what it should be if the location were a valid URL.

Is it possible Cashnet is returning a relative `Location` path here? Can we get a sample of what those location header values look like? I suspect we're not getting full URLs back from Cashnet...

hoenth:
@rwdaigle - This is a sample headers, but likely is not one to cause the invalid URI error:
````
HTTP/1.1 302 Found
Cache-Control: private
Content-Length: 603
Content-Type: text/html; charset=utf-8
Location: https://commerce.cashnet.com/cashnetu/Gateway/htmlgw.aspx?client=EMARKETVENDOR_DEMO&merchant=YYYYYYYYYYY&command=SALE&pmtcode=givecorpcc&station=YYYYYYYYYY&custcode=TheCustCode&operator=YYYYYYYYYY&password=YYYYYYYYYYYY&itemcode=GiveCorps&name_g=Sally+Test&addr_g=1234+Anystreet&city_g=Anytown&state_g=CA&zip_g=94601&amount=5.00&cardno=5454545454545454&expdate=1215&desc=&fname=&lname=&order_number=&email_g=
Set-Cookie: ASP.NET_SessionId=; path=/
P3P: CP="NOI DSP COR NID NOR"
Date: Thu, 11 Feb 2016 19:45:51 GMT
Set-Cookie: TS01b3ea04=; Path=/
````

I have removed the sensitive information. 

However, there are fields that don't have values in this redirect that might, based on user entry. Or they may have different values, and perhaps they include characters that are not valid. 

Unfortunately, I can't get a sample of a URI that actually causes the error, because we don't communicate to Cashnet directly, but through Spreedly, and they aren't saving the redirect location. It happens only intermittently.

hoenth:
@rwdaigle Can we get a resolution on this? Unfortunately, if we get a URI parse error now, it may mean that a transaction has been successfully completed at Cashnet but we don't get record of it in our system, as AM never returns an appropriate response object.

rwdaigle:
@ntalbott wondering how to solve this.

Basically, Cashnet returns a `Location` header that may contain URL unsafe characters. So we need to encode that value. However, the `CGI.escape` method is deprecated in Ruby (and there doesn't seem to be a viable replacement). What to do?

ntalbott:
@rwdaigle how much example data do we have? I'm wondering how wide the set of URL unsafe data is, and whether it's practical to explicitly escape what's there vs. grabbing a generic utility. Since Cashnet is returning the unsafe value, I also wonder about asking them what the total set of unsafe characters are that we should be escaping.

There's also some precedent for this in ActiveMerchant; you might search Issue and even code history for uses of `CGI.escape` to see what we've used in the past to replace its use.

rwdaigle:
@hoenth looks like `URI::Parser.new.escape` will do the trick w/o warnings. Mind updating the PR?

hoenth:
@rwdaigle Done.

PR #1966, opened 2015-12-12
Add scrubbing to NAB Transact

bruno:
I've also added a remote test (using the Braintree Gateway as an inspiration) however lots of the other tests seem to fail on that remote file ... that one passes.


bruno:
I have messed up the merge from upstream on our fork, thus the extra commit, which I want to avoid introducing into `activemerchant/active_merchant` ... fixing and re-submitting in a separate PR.

PR #1969, opened 2015-12-15
Bpoint: Handle message for invalid login

anellis:


anellis:
@duff this is for the bpoint issue of not returning an error message when there's an invalid login

PR #1970, opened 2015-12-15
Element: Parse responses from unexpected API errors

davidsantoso:
@duff: quick update to Element to parse responses that for whatever reason, doesn't follow their documented format.


duff:
@davidsantoso All merged!  Only thing I removed was the line or 2 around `error_code_from`.  Looking at the other gatways, that method looks all wrapped into supporting a `STANDARD_ERROR_CODE_MAPPING` for the gateway.

davidsantoso:
@duff gotcha. Thanks!

PR #1974, opened 2015-12-18
Add store support to Realex

bruno:
This adds support for storing/tokenising customers and payment methods on the Realex gateway using the RealVault product from Realex.

Note:
- Realex now publishes the passing and failing test cards, so I added those to the fixtured (also adding `brand: master` for the mastercard test numbers as the test helper defaults to visa)
- Updated the remote tests to reflect the expected results from Realex
- Realex tokenises cards in two steps, first you need to create a customer and then add a payment method, so it requires two options and those have some specific formatting requirements, i.e. only accepts a set types of characters, etc. Thus why I added the `requires!` call to
the `store` method.

Still TODO: modify the implementation to add support for purchase, authorize and capture using the tokenised card.


bruno:
I seem to have messed up the upstream merge ... fixing on our fork and will re-submit this pull request with a clean history.

PR #1978, opened 2015-12-21
Sage: Refactor bankcard and virtual check into one class

anellis:


anellis:
@duff or @ryan can you give me your thoughts on this while we're working out what we want to do with vault?

anellis:
@duff alright, I now have the vault as a nested class, sending the ssl_post through the primary gateway. This should fix the transcript issues.

lib/active_merchant/billing/gateways/sage.rb
```diff
@@ -1,172 +1,422 @@
...
-      # * <tt>reference</tt> - The authorization reference string returned by the original transaction's Response#authorization.
       def void(reference, options = {})
-        if reference.split(";").last == "virtual_check"
-          virtual_check.void(reference, options)
-        else
-          bankcard.void(reference, options)
-        end
+        post = {}
+        add_reference(post, reference)
+        source = reference.split(";").last == "virtual_check" ? "virtual_check" : "bankcard"
```
rwdaigle: I see the string `virtual_check` used in a lot of places - perhaps we could set it as a const?

rwdaigle: (same for `bankcard`?)

rwdaigle:
@anellis am I correct in saying that this is a backwards-incompatible change (in that there are no longer `SageBankcardGateway`, or vault classes)?

anellis:
@rwdaigle Yes, you are correct in that respect. Someone could be using the SageBankCardGateway directly instead of using it through SageGateway. Maybe I should keep those classes and put a deprecation warning?

rwdaigle:
Ah, I didn't notice those weren't meant to be externally accessible classes. I think we're ok on that front.

anellis:
@rwdaigle is this what you had in mind for using a constant with the source?

rwdaigle:
What do you think about just having two `SOURCE_CARD` and `SOURCE_ECHECK` constants instead of going through a hash? If there are just two values a hash is probably overkill?

anellis:
@rwdaigle updated!

PR #1985, opened 2015-12-31
Cardstream: Adjust authorize and capture

anellis:


anellis:
@duff this remaps our authorize to cardstreams SALE with a captureDelay and our capture to their CAPTURE. This should fix the 3ds problems

lib/active_merchant/billing/gateways/card_stream.rb
```diff
@@ -199,11 +200,10 @@ def parse(body)
       end
 
       def commit(action, parameters)
-
+        parameters.update(:countryCode => self.supported_countries[0]) unless action == 'CAPTURE'
```
rwdaigle: @anellis can you explain this hard-coded `self.supported_countries[0]` part?

rwdaigle: ah, I see it's just a carryover value...

PR #1990, opened 2016-01-11
iyzico payment gateway integration

mddharmesh:
iyzico(https://iyzico.com) payment gateway integration.


lib/active_merchant/billing/gateways/iyzico.rb
```diff
@@ -0,0 +1,322 @@
...
+        true
+      end
+
+      def scrub(transcript)
+        transcript
+      end
+
+      private
+
+      class RandomStringGenerator
```
ntalbott: Please use a proper random value generator; in Ruby, use `SecureRandom`.

ntalbott:
* This is missing most standard remote and unit tests. Please fill them out.
* Adapters should not require non-standard options like the "basket" parameters you're passing. Please conform to the standard options used by other adapters. Extra options are fine, but they must be optional.
* The PKI method seems overly complex for what it does. Can you please try to simplify it? I'd start by killing most/all the type checks.

mddharmesh:
Thank you for the feedback.
We will do the changes and we will make new pull request.

timbeiko:
@mddharmesh I will close this for now. Please submit your changes as a new PR, link back to this and we'll review them.

PR #1992, opened 2016-01-11
Use In-Context Express redirect url for PaypalExpressGateway

xuorig:
### Context
We want to enable the In-Context Checkout in Express Checkout (https://developer.paypal.com/docs/classic/express-checkout/in-context/) on Shopify checkout. To do so, our base url, live and sandbox, must change to `https://www.paypal.com/checkoutnow` and `http://www.sandbox.paypal.com/checkoutnow` respectively.

The in-context checkout needs a paypal js library to work, but the new url doesn't affect going through a standard checkout without the library for the popup. This shouldn't affect other flows.

We also can't send the `&cmd=_express-checkout`, which kicks in the old express checkout flow when using In-Context. We therefore don't need a mobile / non-mobile url for paypal_express only.

### Solution proposed
We only need `paypal_express` to change. Digital goods need to stay as they are, same with payflow, and paypal.

I've changed `paypal_express_common` with a new method `base_redirect_url_for` which returns the basic url for all express gateways. 

Each gateway then implements his own `redirect_url_for` method, that can add query params or change options to the base redirect url.

I need to test this extensively to see if the new URL does indeed work in all flows, but im creating this PR to get some feedback meanwhile.

@girasquid 
cc @Krystosterone

### DIFF
--- a/lib/active_merchant/billing/gateways/payflow_express.rb
+++ b/lib/active_merchant/billing/gateways/payflow_express.rb
@@ -69,6 +69,11 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'https://www.paypal.com/cgi-bin/webscr?cmd=xpt/merchant/ExpressCheckoutIntro-outside'
       self.display_name = 'PayPal Express Checkout'
 
+      def redirect_url_for(token, options = {})
+        cmd = options[:mobile] ? '_express-checkout-mobile' : '_express-checkout'
+        build_redirect_url_for(token, options: options, cmd: cmd)
+      end
+
       def authorize(money, options = {})
         requires!(options, :token, :payer_id)
         request = build_sale_or_authorization_request('Authorization', money, options)
--- a/lib/active_merchant/billing/gateways/paypal_express_common.rb
+++ b/lib/active_merchant/billing/gateways/paypal_express_common.rb
@@ -16,13 +16,11 @@ module ActiveMerchant
         test? ? test_redirect_url : live_redirect_url
       end
 
-      def redirect_url_for(token, options = {})
-        options = {:review => true, :mobile => false}.update(options)
-
-        cmd  = options[:mobile] ? '_express-checkout-mobile' : '_express-checkout'
-        url  = "#{redirect_url}?cmd=#{cmd}&token=#{token}"
+      def build_redirect_url_for(token, options: {}, cmd: nil)
+        options = {review: true}.update(options)
+        url = "#{redirect_url}?token=#{token}"
         url += '&useraction=commit' unless options[:review]
-
+        url += "&cmd=#{cmd}" if cmd
         url
       end
     end
--- a/test/unit/gateways/paypal_express_test.rb
+++ b/test/unit/gateways/paypal_express_test.rb
@@ -2,15 +2,11 @@ require 'test_helper'
 require 'nokogiri'
 
 class PaypalExpressTest < Test::Unit::TestCase
-  TEST_REDIRECT_URL        = 'https://www.sandbox.paypal.com/cgi-bin/webscr?cmd=_express-checkout&token=1234567890'
-  TEST_REDIRECT_URL_MOBILE = 'https://www.sandbox.paypal.com/cgi-bin/webscr?cmd=_express-checkout-mobile&token=1234567890'
-  LIVE_REDIRECT_URL        = 'https://www.paypal.com/cgi-bin/webscr?cmd=_express-checkout&token=1234567890'
-  LIVE_REDIRECT_URL_MOBILE = 'https://www.paypal.com/cgi-bin/webscr?cmd=_express-checkout-mobile&token=1234567890'
+  TEST_REDIRECT_URL        = 'https://www.sandbox.paypal.com/checkoutnow?token=1234567890'
+  LIVE_REDIRECT_URL        = 'https://www.paypal.com/checkoutnow?token=1234567890'
 
   TEST_REDIRECT_URL_WITHOUT_REVIEW = "#{TEST_REDIRECT_URL}&useraction=commit"
   LIVE_REDIRECT_URL_WITHOUT_REVIEW = "#{LIVE_REDIRECT_URL}&useraction=commit"
-  TEST_REDIRECT_URL_MOBILE_WITHOUT_REVIEW = "#{TEST_REDIRECT_URL_MOBILE}&useraction=commit"
-  LIVE_REDIRECT_URL_MOBILE_WITHOUT_REVIEW = "#{LIVE_REDIRECT_URL_MOBILE}&useraction=commit"
 
   def setup
     @gateway = PaypalExpressGateway.new(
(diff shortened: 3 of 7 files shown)


ntalbott:
@xuorig so this will not affect any current users of any existing functionality of the PayPal Express adapter?

xuorig:
@ntalbott that's right, it should go back to the "classic" existing express flow when the In-Context checkout is not available and I confirmed that in my own tests. We are reaching out with paypal to confirm this.

lib/active_merchant/billing/gateways/payflow_express.rb
```diff
@@ -69,6 +69,11 @@ class PayflowExpressGateway < Gateway
       self.homepage_url = 'https://www.paypal.com/cgi-bin/webscr?cmd=xpt/merchant/ExpressCheckoutIntro-outside'
       self.display_name = 'PayPal Express Checkout'
 
+      def redirect_url_for(token, options = {})
+        cmd = options[:mobile] ? '_express-checkout-mobile' : '_express-checkout'
+        "#{base_redirect_url_for(token, options)}&cmd=#{cmd}"
```
girasquid: Instead of doing this, why not have `base_redirect_url_for` accept a `cmd` argument (that could be nil)?

girasquid: So these calls become:
```ruby
def redirect_url_for(token, options = {})
  cmd = options[:mobile] ? '_express-checkout-mobile' : '_express-checkout'
  base_redirect_url_for(token, options, cmd)
end
```

And the one below is just:
```ruby
def redirect_url_for(token, options = {})
  base_redirect_url_for(token, options)
end
```

test/unit/gateways/paypal_express_test.rb
```diff
@@ -39,13 +35,11 @@ def teardown
   def test_live_redirect_url
     Base.mode = :production
     assert_equal LIVE_REDIRECT_URL, @gateway.redirect_url_for('1234567890')
-    assert_equal LIVE_REDIRECT_URL_MOBILE, @gateway.redirect_url_for('1234567890', :mobile => true)
```
girasquid: This goes away because with express in context, there is no mobile-specific URL, right?

xuorig: Yes exactly!

lib/active_merchant/billing/gateways/payflow_express.rb
xuorig: That's definitely cleaner :+1:

xuorig:
@girasquid Addressed your comments! There's a chance the new url might work with payflow and all other express checkouts, I'm confirming that tomorrow morning by testing and also our paypal contact.

In this case, the PR would change to be even simpler! Will update once I know.

Krystosterone:
I heard back from our contact over at Paypal and sadly, this new URL is only supported by Paypal Express.

xuorig:
😿 alright so this PR stays the same then!

lib/active_merchant/billing/gateways/paypal_express_common.rb
```diff
@@ -16,13 +16,11 @@ def redirect_url
         test? ? test_redirect_url : live_redirect_url
       end
 
-      def redirect_url_for(token, options = {})
+      def base_redirect_url_for(token, options: {}, cmd: nil)
         options = {:review => true, :mobile => false}.update(options)
```
Krystosterone: Why have a combination of named parameter and an options hash? Let's change it all to named parameters. 

Also, `mobile => false` has no more sense anymore in this context. 

I'm not fond of classes or methods names `base_*`. WDYT of naming it `build_redirect_url(`

xuorig: `mobile: false` does have sense when this is `payflow_express` or `digital_goods`. I can change it to `build_redirect_url`.

xuorig: @Krystosterone changing to an options named parameter will break all existing implementations :/, I do agree it's weird to have both.

Krystosterone: What I am saying is that `:mobile => false` doesn't make sense here because it isn't even used :arrow_down: . It was bumped up to the two other functions and now pass in `cmd`.

xuorig: Oh I get what you meant now, you're right!

xuorig: removed mobile

lib/active_merchant/billing/gateways/paypal_express_common.rb
```diff
@@ -16,13 +16,11 @@ def redirect_url
         test? ? test_redirect_url : live_redirect_url
       end
 
-      def redirect_url_for(token, options = {})
-        options = {:review => true, :mobile => false}.update(options)
-
-        cmd  = options[:mobile] ? '_express-checkout-mobile' : '_express-checkout'
-        url  = "#{redirect_url}?cmd=#{cmd}&token=#{token}"
+      def build_redirect_url_for(token, options: {}, cmd: nil)
+        options = {:review => true}.update(options)
```
girasquid: 1.9 style hashes? :innocent:

xuorig: Wasn't sure if you wanted to keep those that way :+1:

girasquid:
One style niggle, but otherwise this is looking good.

xuorig:
Addressed your style comment. 

To recap: 
  - this changes only the `paypal_express` gateway, leaving `payflow_express` and `digital_goods` the same.
  - The new URL works only with paypal_express, and does not need the `cmd` param anymore.
  - The `cmd` param is still needed for `payflow_express` and `digital_goods`, it is appended in the `redirect_url_for` method.

@Krystosterone @ntalbott for a final check ?

ntalbott:
So does this only affect things if the `:mobile` option is passed?

Also, I'm surprised to only see assertions that were deleted? Are there tests that cover the new/changed functionality?

xuorig:
@ntalbott 
> So does this only affect things if the :mobile option is passed?
For paypal express, we just dont need the mobile option anymore. Every other express gateway stays the same and uses the mobile option as usual. Passing mobile to paypal express will just be ignored, so it is backward compatible.

> I'm surprised to only see assertions that were deleted?
The only thing changing from before is the URL returned for paypal express, which is changed in tests.

ntalbott:
@xuorig got it. So, just to reiterate, the only thing changing is the url, and that url will produce the same results for existing users of the adapter?

xuorig:
@ntalbott exactly.

@Krystosterone: Can we confirm one last time with amazon that no matter what the merchant's config is, or any edge case, the new URL will indeed be backwards compatible ?

Krystosterone:
> amazon => paypal
:laughing: 

The URL is not fully backwards compatible. For more context, you can look at their documentation that lists the caveats here: https://developer.paypal.com/docs/classic/express-checkout/in-context/#before-you-get-started

For example, setting `options[:funding_sources][:source]` to `BML` would not work (https://github.com/activemerchant/active_merchant/blob/5bb9d2c16c18afde89da20a07c3f25c7002312fe/lib/active_merchant/billing/gateways/paypal_express.rb#L172).

I wonder if we better leave everything as is, except for:
```
cmd  = options[:mobile] ? '_express-checkout-mobile' : '_express-checkout'
```
that would become
```
cmd = nil
cmd = options[:mobile] ? '_express-checkout-mobile' : '_express-checkout' if options.key?(:mobile)
```
Or having a new `PaypalInContext` class, so not to break every existing implementation that uses this gem

girasquid:
> Or having a new PaypalInContext class, so not to break every existing implementation that uses this gem

I think this approach makes the most sense here, based on that compatibility list. Upgrading your gem version and losing the ability to do bill me later would be less than ideal.

xuorig:
Closing this in favor of a new `PaypalExpressInContext` gateway

PR #1994, opened 2016-01-13
Return Braintree Blue id for unsuccessful transactions when available

prburke:
The Braintree Blue gateway should return the id of a transaction regardless of whether or not it was successful.  Not having the id in the response makes tracking down issues difficult.  

This is currently how the Braintree Orange gateway works:
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/smart_ps.rb#L235

@girasquid


girasquid:
Makes sense to me - @aprofeit up for second review?

aprofeit:
Weird, I wonder what the original intent of that was.

Changes make sense to me :+1:

girasquid:
@ntalbott / @duff any objections?

duff:
@girasquid Looks great to me!

PR #1996, opened 2016-01-13
Braintree Blue support for transaction option eci: 'recurring'

prburke:
Braintree Blue needs to support the `eci: 'recurring'` option when procession transaction parameters.  This aligns with the current behaviour of other gateways like Stripe e.g.

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L384

@girasquid @aprofeit 

/cc @jeromecornet


ntalbott:
@prburke I'm not aware of any gateways (though I guess Stripe does?) that support this option, nor do I think it should be standard for ActiveMerchant (vs. `recurring: true` as is supported now). Can you point me at a gateway other than Stripe that does it like this?

prburke:
@ntalbott Thanks for the feedback.  We've found a better place to make this change so I'll close this PR.

PR #1997, opened 2016-01-21
Checkout V2: Add Descriptor Name and City Options

anellis:


anellis:
@duff adding a few gateway specific fields

PR #1883, opened 2015-09-25
World Net TPS Gateway

varyonic:
New gateway for World Net TPS, http://worldnettps.com/

PR #1999, opened 2016-01-21
World net

namskap:
Files added  from Worldnet


varyonic:
Looks like this might duplicate #1883?

lib/active_merchant/billing/gateways/pay_secure.rb
```diff
@@ -66,7 +66,7 @@ def add_credit_card(post, credit_card)
 
       def commit(action, money, parameters)
         response = parse( ssl_post(self.live_url, post_data(action, parameters)) )
-
+        puts response
```
bruno: Seems like a debugging statement, this should probably not be here.

Gemfile
```diff
@@ -1,9 +1,13 @@
...
 gemspec
 
+
 gem 'jruby-openssl', :platforms => :jruby
 
 group :test, :remote_test do
   # gateway-specific dependencies, keeping these gems out of the gemspec
-  gem 'braintree', '>= 2.50.0'
+gem 'braintree', '>= 2.50.0'
+gem 'builder', '~> 3.2.2'
```
bruno: Why use `builder`? ActiveMerchant seems to favour Nokogiri for all XML generation/parsing

PR #2001, opened 2016-01-22
Cardstream: Add AVS Code and Message

anellis:


anellis:
@duff so here's my attempt at fixing the Cardstream code and message problem. 

For the messages, they have special messages that don't map to any of their codes, as is standard in AM, so I can't simply use the code to derive the message. Each number in the code actually provides other details, so if I only provided the code and tried to derive one of the standard AM messages from the code, I wouldn't be giving the customer back all the information. For this reason, I decided to just overwrite the AVS class message handling to be able to pass in a custom message. Would love to hear your thoughts!

rwdaigle:
@anellis we should probably strive to maintain a consistent external interface, including the standard AVS messages. Is there no way to map, at the code/message level, from one to another before passing into the `AVSResult` class. Can you give an example of this mis-match?

anellis:
@rwdaigle there are a few instances where the messages say things that the response code does not. For example, one possible message is "security checks not supported", however the response codes don't reflect that, the most it seems that one could interpret from the code is "address and cvv not checked" but you wouldn't get the specific reasoning for why. If I were using the code to derive the message i'd have to consolidate their "data not checked" and "security checks not supported" messages into one AM standard message, and the user would lose some information in the process. 

I could use the cardstream messages to map to standard Active Merchant messages, but I still have to overwrite the AVSResult class, as the standard is to [use the code to map to messages](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/avs_result.rb#L74). However that would solve the problem of providing a consistent external interface.

rwdaigle:
> If I were using the code to derive the message i'd have to consolidate their "data not checked" and "security checks not supported" messages into one AM standard message, and the user would lose some information in the process.

I'd do my best to provide reasonable mapping to `AVSResult` codes. If the caller wishes to dig deeper, keep in mind they can always query the raw result params hash, which retains all the original response values (so no info is actually lost, just slightly obscured for those that want to know). That seems like a reasonable trade off?

anellis:
@rwdaigle alright, I went with mapping the cardstream codes to our standard codes. I just couldn't find a way to return their messages without overwriting the AvsResult class. This is because even if I just pass a regular hash in for `avs_result`, the [Response class will take that and turn it into an AvsResult object](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/response.rb#L32)...and then back to hash :).

rwdaigle:
> I just couldn't find a way to return their messages

I think that's ok, keeping in mind that if a caller wants access to the raw message they can still look at the response params directly to access it.

PR #2002, opened 2016-01-22
Merchant ESolutioins: Truncate order_id

anellis:


anellis:
@duff this is to truncate order_id and avoid getting the "Invoice Number Length Error"

duff:
@anellis I'm thinking this could be useful:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L274

anellis:
@duff that is helpful! updated

PR #2003, opened 2016-01-25
TransFirst Transaction Express: add gateway

sdball:
This is a new gateway from TransFirst. Their SOAP interface allows the following standard actions:

- purchase
- authorize
- capture
- void
- refund
- credit (requires approval from gateway)
- verify
- store
- update

NOTE Transaction Express will automatically do a partial capture if the purchase amount exceeds the available credit limit. For automatic partial captures the rspCode field will be "10" and there will be an "amt" field in the response containing the balance successfully captured.

### DIFF
--- a/lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
+++ b/lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
@@ -285,10 +285,10 @@ module ActiveMerchant #:nodoc:
           customer_id = r.params["custId"]
 
           store_payment_method_request = build_xml_payment_storage_request do |doc|
-            doc.tag! "v1:cust" do
+            doc["v1"].cust do
               add_customer_id(doc, customer_id)
-              doc.tag! "v1:pmt" do
-                doc.tag! "v1:type", 0 # add
+              doc["v1"].pmt do
+                doc["v1"].type 0 # add
                 add_payment_method(doc, payment_method)
               end
             end
@@ -300,9 +300,9 @@ module ActiveMerchant #:nodoc:
 
       def update(token, payment_method, options={})
         wallet_details_request = build_xml_payment_search_request do |doc|
-          doc.tag! "v1:type", 1 # recurring profile
-          doc.tag! "v1:pmtCrta" do
-            doc.tag! "v1:pmtId", token
+          doc["v1"].type 1 # recurring profile
+          doc["v1"].pmtCrta do
+            doc["v1"].pmtId token
           end
         end
 
@@ -311,14 +311,14 @@ module ActiveMerchant #:nodoc:
           return r unless r.success? && r.params["FndRecurrProfResponse"]
 
           update_payment_method_request = build_xml_payment_update_request do |doc|
-            doc.tag! "v1:cust" do
-              doc.tag! "v1:contact" do
-                doc.tag! "v1:id", r.params["contact_id"]
+            doc["v1"].cust do
+              doc["v1"].contact do
+                doc["v1"].id r.params["contact_id"]
               end
 
-              doc.tag! "v1:pmt" do
-                doc.tag! "v1:id", token
-                doc.tag! "v1:type", 1 # update
+              doc["v1"].pmt do
+                doc["v1"].id token
+                doc["v1"].type 1 # update
                 add_payment_method(doc, payment_method)
               end
             end
@@ -441,79 +441,78 @@ module ActiveMerchant #:nodoc:
 
       # -- request methods ---------------------------------------------------
       def build_xml_transaction_request
-        build_xml_request("v1:SendTranRequest") do |doc|
+        build_xml_request("SendTranRequest") do |doc|
           yield doc
         end
       end
 
       def build_xml_payment_storage_request
-        build_xml_request("v1:UpdtRecurrProfRequest") do |doc|
+        build_xml_request("UpdtRecurrProfRequest") do |doc|
           yield doc
         end
       end
 
       def build_xml_payment_update_request
         merchant_product_type = 5 # credit card
-        build_xml_request("v1:UpdtRecurrProfRequest", merchant_product_type) do |doc|
+        build_xml_request("UpdtRecurrProfRequest", merchant_product_type) do |doc|
           yield doc
         end
       end
 
       def build_xml_payment_search_request
-        build_xml_request("v1:FndRecurrProfRequest") do |doc|
+        build_xml_request("FndRecurrProfRequest") do |doc|
           yield doc
         end
       end
 
       def build_xml_request(wrapper, merchant_product_type=nil)
-        xml = Builder::XmlMarkup.new
         soapenv = "http://schemas.xmlsoap.org/soap/envelope/"
         v1 = "http://postilion/realtime/merchantframework/xsd/v1/"
 
-        xml.tag! "soapenv:Envelope", "xmlns:soapenv" => soapenv do
-          xml.tag! "soapenv:Body" do
-            xml.tag! wrapper, "xmlns:v1" => v1 do
-              add_merchant(xml, product_type=merchant_product_type)
-              yield xml
+        Nokogiri::XML::Builder.new(encoding: "UTF-8") do |xml|
+          xml["soapenv"].Envelope("xmlns:soapenv" => soapenv) do
+            xml["soapenv"].Body do
+              xml["v1"].send(wrapper, "xmlns:v1" => v1) do
+                add_merchant(xml)
+                yield(xml)
+              end
             end
           end
-        end
-
-        xml.target!
+        end.doc.root.to_xml
       end
 
 
       def add_merchant(doc, product_type=nil)
-        doc.tag! "v1:merc" do
-          doc.tag! "v1:id", @options[:gateway_id]
-          doc.tag! "v1:regKey", @options[:reg_key]
-          doc.tag! "v1:inType", "1"
-          doc.tag! "v1:prodType", product_type if product_type
+        doc["v1"].merc do
+          doc["v1"].id @options[:gateway_id]
+          doc["v1"].regKey @options[:reg_key]
+          doc["v1"].inType "1"
+          doc["v1"].prodType product_type if product_type
         end
       end
 
       def add_transaction_code(doc, action)
-        doc.tag! "v1:tranCode", TRANSACTION_CODES[action]
+        doc["v1"].tranCode TRANSACTION_CODES[action]
       end
 
       def add_amount(doc, money)
-        doc.tag! "v1:reqAmt", amount(money)
+        doc["v1"].reqAmt amount(money)
       end
 
       def add_order_number(doc, options)
         return unless options[:order_id]
 
-        doc.tag! "v1:authReq" do
-          doc.tag! "v1:ordNr", options[:order_id]
-        end
+        doc["v1"].authReq {
+          doc["v1"].ordNr options[:order_id]
+        }
       end
 
       def add_payment_method(doc, payment_method)
-        doc.tag! "v1:card" do
-          doc.tag! "v1:pan", payment_method.number
-          doc.tag! "v1:sec", payment_method.verification_value
-          doc.tag! "v1:xprDt", expiration_date(payment_method)
-        end
+        doc["v1"].card {
+          doc["v1"].pan payment_method.number
+          doc["v1"].sec payment_method.verification_value
+          doc["v1"].xprDt expiration_date(payment_method)
+        }
       end
 
       def expiration_date(payment_method)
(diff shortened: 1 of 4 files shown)


sdball:
Pinging @rwdaigle for review.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+        wallet_sale: 14,
+      }
+
+      def initialize(options={})
+        requires!(options, :gateway_id, :reg_key)
+        super
+      end
+
+      def purchase(amount, payment_method, options={})
+        if credit_card? payment_method
```
rwdaigle: I believe the AM style is: `if credit_card?(payment_method)`

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+          request = build_xml_transaction_request do |doc|
+            add_transaction_code(doc, :purchase)
```
rwdaigle: I'd follow the pattern of passing the tx code into `commit` instead of adding it a separate step here. Mostly for consistency with most other adapters.

```ruby
commit(:purchase, request)
```

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+        end
+
+        commit(request)
+      end
+
+      # Transaction Express has three types of possible void
+      # - void_authorization
+      # - void_purchase (alias: void_capture)
+      # - void_refund (alias: void_credit)
+      def void(authorization, options={})
```
rwdaigle: This is tricky one but, instead of having the user pass in what type of void it is, what we can do is return an `authorization` from the original tx that has the tx type embedded in it. Something that looks like `purchase:1298shad698a`.

Then, when an authorization is passed to this void method, we know what type the original tx was and can self-contain the logic that determines what void type to invoke here.

If you search AM for `split_authorization` you'll see a lot of other instances where this is required.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: The trouble there is that the transaction needs to be written deep into the XML document. If we build it all up it wouldn't be easy to insert the transaction code.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: Oh hey cool, so we can add in some character separated data into the authorization? The API doesn't have to assume that it's dealing with transactions that started outside the API?

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: :+1:

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+            xml.tag! wrapper, "xmlns:v1" => v1 do
+              add_merchant(xml, product_type=merchant_product_type)
+              yield xml
+            end
+          end
+        end
+
+        xml.target!
+      end
+
```
rwdaigle: extra line

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+                add_payment_method(doc, payment_method)
+              end
+            end
+          end
+
+          r.process { commit(store_payment_method_request) }
+        end
+      end
+
+      def update(token, payment_method, options={})
```
rwdaigle: `update`is not part of the AM API, so we can go ahead and omit this method.

ntalbott: @rwdaigle some gateways do actually support it as part of the public AM API; that said, I think it's hard to support consistently so I'm :+1: on not making it part of the standard suite of AM methods we implement.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+      end
+
+      def error_code_from(succeeded, response)
+        return if succeeded
+        response["errorCode"] || response["rspCode"]
+      end
+
+      def message_from(succeeded, response)
+        if succeeded
+          if response["rspCode"] == "10"
```
rwdaigle: We don't need to branch right here. Just have one "Succeeded" response message, and the caller can look at the `rspCode` field themselves if they to differentiate.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+          code = response["rspCode"]
+          extended_code = response["extRspCode"]
+
+          message = RESPONSE_MESSAGES[code]
+          extended = EXTENDED_RESPONSE_MESSAGES[extended_code]
+
+          [message, extended].compact.join('. ')
+        elsif response["faultstring"]
+          response["faultstring"]
+        elsif response.empty?
```
rwdaigle: Just return `nil` for this case.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,597 @@
...
+          "Empty response received from Transaction Express gateway."
+        else
+          "Unknown response received from Transaction Express gateway."
```
rwdaigle: Same here.

rwdaigle:
Few small things, but great job on your first real adapter! *And* you had to wade through a bunch of nested XML to get there :sob:

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
rwdaigle: Given how nasty this XML is, I buy it :+1:

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
rwdaigle: Correct! Unfortunately it's a pretty common pattern.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: I'm happy to remove it. But I was following the methods described by https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards which says to have an update method if possible

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: :+1:

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: I keep wanting to jump between the gateway and the user :-)

sdball:
Thanks! And wooooo XML!

I'll get the minor fixes all wrapped up after I finish puzzling out a good way to handle passing the action to `commit` and dropping that into the authorization.

I may need to actually bring in nokogiri instead of the super simple builder.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
rwdaigle: Remove it is! Looks like the Wiki could be updated.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
rwdaigle: You have good instincts - the gateway is trying to do such horrible things to the user!

rwdaigle:
>  after I finish puzzling out a good way to handle passing the action to commit

np skipping if this if it's [too much of a hassle](https://github.com/activemerchant/active_merchant/pull/2003#discussion_r50750427)

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: I'd like to have it work the conventional way. Otherwise we'll need to add the transaction code to the XML request and separately pass the action to `commit` so that it can be included in the authorization. Eww.

But if I run into a serious wall I'll get philosophically pragmatic about it.

sdball:
Ok! Revisions done. I was able to bring the addition of the transaction code into the commit method and used that to also store the action in the response authorization.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,574 @@
+require "nokogiri"
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class TransFirstTransactionExpressGateway < Gateway
+      self.display_name = "TransFirst Transaction Express"
+      self.homepage_url = "http://transactionexpress.com/"
+
+      self.test_url = "https://ws.cert.transactionexpress.com/portal/merchantframework/MerchantWebServices-v1?wsdl"
```
rwdaigle: Is the `wsdl` param required? Usually that's what you add to get back a description of the web service instead of the service endpoint itself?

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,574 @@
...
+      self.live_url = "https://ws.transactionexpress.com/portal/merchantframework/MerchantWebServices-v1?wsdl"
+
+      self.supported_countries = ["US"]
+      self.default_currency = "USD"
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club]
```
rwdaigle: Are these confirmed, specifically Diners Club?

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,574 @@
...
+      CURRENCY_CODES = Hash.new{|h,k| raise ArgumentError.new("Unsupported currency: #{k}")}
+      CURRENCY_CODES["USD"] = "840"
+
+      V1_NAMESPACE = "http://postilion/realtime/merchantframework/xsd/v1/"
+      SOAPENV_NAMESPACE = "http://schemas.xmlsoap.org/soap/envelope/"
+      AUTHORIZATION_FIELD_SEPARATOR = "|"
+
+      APPROVAL_CODES = %w(00 10)
+
+      RESPONSE_MESSAGES = {
```
rwdaigle: These are usually at the top of the gateway adapter, despite probably being better declared as private, as you have here.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -0,0 +1,574 @@
...
+        yy + mm
+      end
+
+      def add_pan(doc, payment_method)
+        doc["v1"].card do
+          doc["v1"].pan payment_method.number
+        end
+      end
+
+      def add_contact(doc, fullname, options)
```
rwdaigle: We usually name this method `add_customer` instead of `add_contact`.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: I'll give it a run and find out. Those are just the URLs they had in the documentation.

rwdaigle:
I think this looks great! Few small things to move around and I'll merge.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: Hah no problem. I actually just did a commit to move them down here because it seemed like a lot of noise before the actual code.

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: :+1:

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
rwdaigle: Totally agree on the noise. I kinda liked that I didn't have to wade through it to get to the meat of the adapter, but did note it stood out compared to the other adapters.

sdball: Cool, I've reset that commit and pushed :+1:

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: They provide test card numbers for each in the documentation, although they say "Diners (considered MC)"

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
sdball: Looks like it's not required at all. I'll get that committed.

PR #2010, opened 2016-02-03
TransFirst: Fix missing address and remove CC only fields for ACH

davidsantoso:
@rwdaigle or @duff: 2 fixes for TransFirst: 

* Missing `Address` if no address is included as an option to purchase requests
* Removed some credit card only fields from an ACH transaction.

### DIFF
--- a/lib/active_merchant/billing/gateways/trans_first.rb
+++ b/lib/active_merchant/billing/gateways/trans_first.rb
@@ -38,24 +38,27 @@ module ActiveMerchant #:nodoc:
         post = {}
 
         add_amount(post, money)
-        add_invoice(post, options)
         add_payment(post, payment)
         add_address(post, options)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/trans_first.rb
```diff
@@ -85,13 +88,15 @@ def add_address(post, options)
         address = options[:billing_address] || options[:address]
 
         if address
-          add_pair(post, :Address, address[:address1])
-          add_pair(post, :ZipCode, address[:zip])
+          add_pair(post, :Address, address[:address1], required: true)
+          add_pair(post, :ZipCode, address[:zip], required: true)
+        else
+          add_pair(post, :Address, "", required: true)
+          add_pair(post, :ZipCode, "", required: true)
```
rwdaigle: ohrlly?

PR #2012, opened 2016-02-04
Fixing list of currencies without fractions

Krystosterone:
The current list for currencies without fractions is not correct, as per paypal documentation: https://developer.paypal.com/webapps/developer/docs/classic/api/currency_codes/

Only `%w(HUF JPY TWD)` do not support decimals. As of now, having this list would throw errors as the total amount and the line items would be rounded, resulting in a `The totals of the cart item amounts do not match order amounts.` error from paypal.

I've tested out this fix with sandbox accounts located with these appropriate currencies and confirmed the bug by having sandbox accounts with the remaining currencies, located in the appropriate countries.

@duff @girasquid for review please.

### DIFF
--- a/lib/active_merchant/billing/gateways/paypal_express.rb
+++ b/lib/active_merchant/billing/gateways/paypal_express.rb
@@ -26,7 +26,7 @@ module ActiveMerchant #:nodoc:
         'TW' => 'zh_TW'
       }
 
-      CURRENCIES_WITHOUT_FRACTIONS = %w(BRL HUF JPY MYR TWD TRY)
+      CURRENCIES_WITHOUT_FRACTIONS = %w(HUF JPY TWD)
 
       self.test_redirect_url = 'https://www.sandbox.paypal.com/cgi-bin/webscr'
       self.supported_countries = ['US']


aprofeit:
Kinda curious why these others were added, if docs only say those three this makes sense to me :+1:

ThereExistsX:
:rainbow:

duff:
Not sure why, but that Paypal doc link doesn't seem to work for me.

I noticed that https://github.com/activemerchant/active_merchant/commit/5bb9d2c16c18afde89da20a07c3f25c7002312fe might have some useful background.  I'm not sure how much of that is still relevant.

aprofeit:
@duff the link 404'd for me as well, but weirdly worked on the second time.

duff:
Nice.  :shipit: !

Krystosterone:
I had noticed that PR, but the listing of currencies seems to be erroneous. As of now, paypal documentation lists the currencies that don't support decimals as: `HUF JPY TWD`.

@duff Are we ok to merge this?

duff:
@Krystosterone You're right.  I shoulda been more careful reading through each of the note sections in the docs.

PR #2016, opened 2016-02-09
Add ACH support to the Stripe Gateway

sdball:
Stripe requires that bank accounts be externally verified before they can be used in API transactions. The integration here has the process split into two actions:

1. Store the bank account (kicks off the Stripe verification process)
2. Purchase using the customer id and bank account id

The customer id and bank account id are returned by the `store` method formatted for direct use in `purchase` and `refund` after verification is complete.

Reference:
- https://stripe.com/docs/guides/ach

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -46,6 +46,11 @@ module ActiveMerchant #:nodoc:
         'incorrect_pin' => STANDARD_ERROR_CODE[:incorrect_pin]
       }
 
+      BANK_ACCOUNT_HOLDER_TYPE_MAPPING = {
+        "personal" => "individual",
+        "business" => "company",
+      }
+
       def initialize(options = {})
         requires!(options, :login)
         @api_key = options[:login]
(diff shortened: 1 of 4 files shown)


sdball:
Ping @rwdaigle for review.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -542,6 +552,38 @@ def error_code_from(response)
         error_code ||= STANDARD_ERROR_CODE_MAPPING[code]
         error_code
       end
+
+      def tokenize_bank_account(bank_account, options = {})
+        account_holder_type = bank_account.account_holder_type == "personal" ? "individual" : "company"
```
masaruhoshi: Is it OK if we have constants for those?

sdball: That's a great idea :+1: I'll make a constant lookup and give it the key instead of having a ternary.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -84,6 +89,8 @@ def purchase(money, payment, options = {})
           if payment.is_a?(ApplePayPaymentToken)
             r.process { tokenize_apple_pay_token(payment) }
             payment = StripePaymentToken.new(r.params["token"]) if r.success?
+          elsif ach?(payment)
```
rwdaigle: I wonder if this state should be moved outside the `MultiResponse`?

Also, do we return a Response-based error in other adapters? Seems like a `Response` should be reserved for when an actual call is made. What other immediate error return states can we use?

sdball: Hm, agreed! Keeping that inside the `MultiResponse` is a holdover from when I was trying to handle bank accounts directly in the `purchase` method.

Returning a new Response is something we do elsewhere in the Stripe gateway, e.g. https://github.com/sdball/active_merchant/blob/add-ach-support-to-stripe/lib/active_merchant/billing/gateways/stripe.rb#L154

PR #2019, opened 2016-02-15
Return a failure response instead of raising an exception for invalid Clearhaus signing keys

sdball:
If the signing key is invalid return a failure response instead of raising up the exception from OpenSSL.


sdball:
Taking this back, looks like we could clean up whitespace around signing keys to make it easier to create the gateway.

PR #2021, opened 2016-02-18
Move and improve VISA Electron detection from SagePay to CreditCardMethods

sdball:
The previous regex based Electron ranges were a bit complex to modify and apparently out of date as it was incorrectly identifying some VISA debit cards as VISA Electron.

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -111,6 +111,11 @@ module ActiveMerchant #:nodoc:
         @brand = (value.respond_to?(:downcase) ? value.downcase : value)
       end
 
+      # Returns if the card matches known Electron BINs
+      def electron?
+        self.class.electron?(number)
+      end
+
       # Returns or sets the first name of the card holder.
       #
       # @return [String]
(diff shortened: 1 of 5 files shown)


rwdaigle:
@sdball I wonder if we shouldn't update [`CreditCardMethods#brand?`](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card_methods.rb#L97) to include this logic. Seems a shame to lock it up in a single gateway?

sdball:
Hmm, this logic was in `SagePay` before so I assumed there was a reason it was not in a higher level spot, e.g. maybe some gateways want to see "Visa Electron" as "Visa". As far as I can tell only SagePay is weird and needs a special value for Electron cards.

If we do want to move this to brand, what would we return for visa electron cards? It's not one of the valid types already.

We could perhaps move `electron?` to `CreditCardMethods`.

Looking at `CreditCardMethods` we could also potentially improve the `brand?` logic to be less… regexy. As much as I love regexes they are poor representations of number ranges.

rwdaigle:
> We could perhaps move `electron?` to `CreditCardMethods`

I like this as it keeps the Electron determination close to the existing branding logic while letting individual gateways determine if they want that level of granularity.

sdball:
Sweet, I'll get that moved around in this pull request :+1:

PR #2022, opened 2016-02-19
Use autoCapture flag for Checkout v2 purchase

curiousepic:
Fix for https://github.com/spreedly/core/issues/1610, @rwdaigle to review.


rwdaigle:
Seeing the remote tests fail (as they did before this commit) - want to see if we can resolve it since the failure is pretty isolated (on refund).

PR #2034, opened 2016-03-02
CyberSource: Update test and live URL

marquisong:
This PR updates the test and live URL of CyberSource to point to new URLs provided by CyberSource.
As announced by CyberSource, they are migrating their infrastructure to Akamai-dedicated URLs and requesting their customers to migrate to these new URLs by June 30, 2016.

I have tested this change using unit and remote tests.


erick-mitchells:
Any word on when this will be brought in?  As noted above, the old URLs will no longer be valid as of June 30.

duff:
Thanks @marquisong !

PR #2035, opened 2016-03-06
Basic implementation of `store` for Realex

bruno:
This adds support for storing/tokenising customers and payment methods
on the Realex gateway using the RealVault product from Realex.

Note:
- Realex now publishes the passing and failing test cards, so I added
  those to the fixtured (also adding `brand: master` for the mastercard
  test numbers as the test helper defaults to visa)
- Updated the remote tests to reflect the expected results from Realex
- Realex tokenises cards in two steps, first you need to create a
  customer and then add a payment method, so it requires two options and
  those have some specific formatting requirements, i.e. only accepts a
  set types of characters, etc. Thus why I added the `requires!` call to
  the `store` method.

This replaces #1974

Still TODO: modify the implementation to add support for purchase,
authorize and capture using the tokenised card.

### DIFF
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -83,6 +83,21 @@ module ActiveMerchant
         commit(request)
       end
 
+      def store(creditcard, options = {})
+        # the customer token and payment method token are mandatory in Realex
+        requires!(options, :customer, :cardref)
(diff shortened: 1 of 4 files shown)


davidsantoso:
Hi @bruno - I'm grooming old pull requests and if this is still something you're interested in, please:

1. Reopen the PR
2. Rebase on `master`
3. Squash all commits
4. Tag me for review

Thanks!

PR #2036, opened 2016-03-07
Add scrubbing to NAB Transact

bruno:
This allows scrubbing of NAB transact transcripts.

I've also added a remote test (using the Braintree Gateway as an inspiration) however lots of the other tests seem to fail on that remote file ... that one passes.

Replaces #1966


bruno:
Sorry, pushed again something that's specific to our implementation and mixed up a bunch of other commits. Will try again.

PR #2042, opened 2016-03-10
Sagepay permit cv2 with token payment

rob-anderson:
Currently when posting a purchase transaction using a previously stored SagePay token as the payment method, it is not possible to resupply the CV2 number.

From the SagePay docs:

> The CV2 value is only stored prior to authorisation. After the initial attempted use of the token the CV2 will be deleted to comply with industry regulations. If you have AVS/CV2 checks enabled on your account, you will need to obtain this information from the customer each time the token is used.

So if a token is to be used more than once, it is mandatory to resupply the CV2 number (assuming this is enabled on the account).  The existing gateway does not permit a token to be supplied with a CV2 number.

This patch adds an optional :CV2 field which will be posted with the token, if present.  A remote test is also included.

### DIFF
--- a/lib/active_merchant/billing/gateways/sage_pay.rb
+++ b/lib/active_merchant/billing/gateways/sage_pay.rb
@@ -264,6 +264,7 @@ module ActiveMerchant #:nodoc:
       def add_token_details(post, token, options)
         add_token(post, token)
         add_pair(post, :StoreToken, options[:customer])
+        add_pair(post, :CV2, options[:verification_value])
       end
 
       def add_token(post, token)
--- a/test/remote/gateways/remote_sage_pay_test.rb
+++ b/test/remote/gateways/remote_sage_pay_test.rb
@@ -250,6 +250,15 @@ class RemoteSagePayTest < Test::Unit::TestCase
     assert_success purchase
   end
 
+  def test_successful_store_and_repurchase_with_resupplied_verification_value
+    assert response = @gateway.store(@visa)
+    assert_success response
+    assert !response.authorization.blank?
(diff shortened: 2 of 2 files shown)


rob-anderson:
It's only the bb02f5f commit which is relevant.  Don't know why the pull request has picked up the ancient stuff, sorry.

cristianstanescu:
I'm also interested in this issue with SagePay CV2 verification and I was actually planning to create a PR myself.

There doesn't seem to be a standard method in place for passing CV2 for a purchases against tokens. I've had a look at Braintree, Stripe, Paymill, Wepay and each seem to be implementing "their own token approach".

However I would have a suggestion here: instead of using the hash key `:CV2` when assigning from `options` (https://github.com/activemerchant/active_merchant/pull/2042/files#diff-a3db0f082e46cc719b09a6b8de5d6193R267):

    add_pair(post, :CV2, options[:CV2])

maybe use `:verification_value` to be closer to the ActiveMerchant language (https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card.rb#L44):

    add_pair(post, :CV2, options[:verification_value])

rob-anderson:
OK, rebased/squashed/forced etc. and CV2 changed to verification_value.

Sorry it took so long :)

timbeiko:
Had a look and LGTM. @ntalbott what do you think?

aprofeit:
lgtm assuming it's been manually tested and everything is working as expected

timbeiko:
Just manually tested it and everything still works fine. Merging into master.

PR #2044, opened 2016-03-10
Jetpay: Add origin field

anellis:


lib/active_merchant/billing/gateways/jetpay.rb
```diff
@@ -200,13 +200,18 @@ def scrub(transcript)
...
+      def build_xml_request(transaction_type, options = {}, transaction_id = nil, &block)
         xml = Builder::XmlMarkup.new
         xml.tag! 'JetPay' do
           # The basic values needed for any request
           xml.tag! 'TerminalID', @options[:login]
           xml.tag! 'TransactionType', transaction_type
           xml.tag! 'TransactionID', transaction_id.nil? ? generate_unique_id.slice(0, 18) : transaction_id
+          if options && options[:origin]
+            xml.tag! 'Origin', options[:origin]
+          else
```
duff: @rwdaigle I'm curious about INTERNET being the default.  Do you think we need a default for this field?  I'm afraid this might break folks who don't want the field specified.

rwdaigle: I guess in that the existing behavior is to have nothing specified, we can not have a default. Only if the field were mandatory would `INTERNET` probably be the correct default.

duff: Coolo.  I just pushed the adjustment in 74940a0929a9d188d444c2567f0f13e5a469a6e3

PR #2045, opened 2016-03-11
Ogone: query single payments

tahb:
We found the library lacking the ability to query directly for a single payment, which ogone does support. It's also super handy to be able to search by ORDERID or PAYID.

Been using this in production for around a year now.


davidsantoso:
@tahb thanks for the PR- although I can imagine this is super handy, at the moment the scope of gateway adapters in Active Merchant stops when it comes to querying a transaction/payment. In general (with a few exceptions) gateway adapters are not pure layers on top of a gateway API but more of general interface.

PR #2048, opened 2016-03-18
GlobalCollect: New gateway support

curiousepic:
@rwdaigle to review

### DIFF
--- a/lib/active_merchant/billing/gateways/globalcollect.rb
+++ b/lib/active_merchant/billing/gateways/globalcollect.rb
@@ -1,0 +1,293 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class GlobalcollectGateway < Gateway
+      self.display_name = "GlobalCollect"
+      self.homepage_url = "http://www.globalcollect.com/"
+
+      self.test_url = "https://api-sandbox.globalcollect.com/"
+      self.live_url = "https://api.globalcollect.com/"
+
+      self.supported_countries = %w(AD AE AT AU BD BE BG BN CA CH CY CZ DE DK
+      EE EG ES FI FR GB GI GR HK HU ID IE IL IM IN IS IT JO KW LB LI LK LT LU
+      LV MC MT MU MV MX MY NL NO NZ OM PH PL PT QA RO SA SE SG SI SK SM TR TT
+      UM US VA VN ZA)
+      self.default_currency = "USD"
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      def initialize(options={})
+        requires!(options, :merchant_id, :api_key_id, :secret_api_key)
+        super
+      end
+
+      def purchase(money, payment, options={})
+        MultiResponse.run do |r|
+          r.process { authorize(money, payment, options) }
+          r.process { capture(money, r.authorization, options) }
+        end
+      end
+
+      def authorize(money, payment, options={})
+        post = nestable_hash
+        add_order(post, money, options)
+        add_payment(post, payment)
+        add_customer_data(post, options, payment)
+        add_address(post, payment, options)
+
+        commit(:authorize, post)
+      end
+
+      def capture(money, authorization, options={})
+        post = nestable_hash
+        add_order(post, money, options)
+        add_customer_data(post, options)
+        commit(:capture, post, authorization)
+      end
+
+      def refund(money, authorization, options={})
+        post = nestable_hash
+        add_amount(post, money, options={})
+        add_refund_customer_data(post, options)
+        commit(:refund, post, authorization)
+      end
+
+      def void(authorization, options={})
+        post = nestable_hash
+        commit(:void, post, authorization)
+      end
+
+      def verify(payment, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, payment, options) }
+          r.process { void(r.authorization, options) }
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: )[^\\]*)i, '\1[FILTERED]').
+          gsub(%r(("cardNumber\\":\\")\d+), '\1[FILTERED]').
+          gsub(%r(("cvv\\":\\")\d+), '\1[FILTERED]')
+      end
+
+      private
+
+      BRAND_MAP = {
+        "visa" => "1",
+        "american_express" => "2",
+        "master" => "3",
+        "discover" => "128"
+      }
+
+      def add_order(post, money, options)
+        post["order"]["amountOfMoney"] = {
+          "amount" => amount(money),
+          "currencyCode" => options[:currency] || currency(money)
+        }
+        post["order"]["references"] = {
+          "merchantReference" => options[:order_id],
+          "descriptor" => options[:description] # Max 256 chars
+        }
+        post["order"]["references"]["invoiceData"] = {
+          "invoiceNumber" => options[:invoice]
+        }
+      end
+
+      def add_amount(post, money, options)
+        post["amountOfMoney"] = {
+          "amount" => amount(money),
+          "currencyCode" => options[:currency] || currency(money)
+        }
+      end
+
+      def add_payment(post, payment)
+        year  = format(payment.year, :two_digits)
+        month = format(payment.month, :two_digits)
+        expirydate =   "#{month}#{year}"
+
+        post["cardPaymentMethodSpecificInput"] = {
+            "paymentProductId" => BRAND_MAP[payment.brand],
+            "skipAuthentication" => "true", # refers to 3DSecure
+            "skipFraudService" => "true"
+        }
+        post["cardPaymentMethodSpecificInput"]["card"] = {
+            "cvv" => payment.verification_value,
+            "cardNumber" => payment.number,
+            "expiryDate" => expirydate,
+            "cardholderName" => payment.name
+        }
+      end
+
+      def add_customer_data(post, options, payment = nil)
+        post["order"]["customer"] = {
+          "merchantCustomerId" => options[:customer]
+        }
+        if payment
+          post["order"]["customer"]["personalInformation"] = {
+            "name" => {
+              "firstName" => payment.first_name,
+              "surname" => payment.last_name
+            }
+          }
+        end
+        post["order"]["companyInformation"] = {
+          "name" => options[:company]
+        }
+        post["order"]["contactDetails"] = {
+          "emailAddress" => options[:email]
+        }
+        if address = options[:billing_address] || options[:address]
+          post["order"]["contactDetails"] = {
+            "phoneNumber" => address[:phone]
+          }
+        end
+      end
+
+      def add_refund_customer_data(post, options)
+        if address = options[:billing_address] || options[:address]
+          post["customer"]["address"] = {
+            "countryCode" => address[:country]
+          }
+          post["customer"]["contactDetails"] = {
+            "emailAddress" => options[:email],
+            "phoneNumber" => address[:phone]
+          }
+        end
+      end
+
+      def add_address(post, creditcard, options)
+        billing_address = options[:billing_address] || options[:address]
+        shipping_address = options[:shipping_address]
+        if billing_address = options[:billing_address] || options[:address]
+          post["order"]["customer"]["billingAddress"] = {
+            "street" => billing_address[:address1],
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+      self.supported_countries = %w(AD AE AT AU BD BE BG BN CA CH CY CZ DE DK
+      EE EG ES FI FR GB GI GR HK HU ID IE IL IM IN IS IT JO KW LB LI LK LT LU
+      LV MC MT MU MV MX MY NL NO NZ OM PH PL PT QA RO SA SE SG SI SK SM TR TT
+      UM US VA VN ZA)
+      self.default_currency = "USD"
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      def initialize(options={})
+        requires!(options, :merchant_id, :apiKeyId, :secretApiKey)
```
rwdaigle: Let's go ahead and underscore these options to: `:api_key_id` and `:secret_api_key`

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+        super
+      end
+
+      def purchase(money, payment, options={})
+        MultiResponse.run(:use_first_response) do |r|
```
rwdaigle: I wonder if we should *not* `:use_first_response` since the purchase is not actually successful if the underlying auth isn't successfully captured.

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+      }
+
+      def add_order(post, money, options)
+        post["order"]["amountOfMoney"] = {
+          "amount" => amount(money),
+          "currencyCode" => options[:currency] || currency(money)
+        }
+        post["order"]["references"] = {
+          "merchantOrderId" => options[:order_id], # Max 10 digits
+          "merchantReference" => options[:order_id],
```
rwdaigle: This should probably be `options[:merchant]`?

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+          "descriptor" => options[:description] # Max 256 chars
+        }
+        if options[:invoicedata]
+          post["order"]["reference"]["invoiceData"] = {
+            "invoiceNumber" => options[:invoicenumber]
```
rwdaigle: This should probably be `options[:invoice]` and not inside the `if` block? (`:invoice` being one of the standardized option keys)

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+        }
+      end
+
+      def add_customer_data(post, options)
+        post["order"]["customer"] = {
+          "merchantCustomerId" => options[:customer]
+        }
+        post["order"]["customer"]["personalInformation"] = {
+          "name" => {
+            "firstName" => options[:firstname],
```
rwdaigle: We should get the first/last names [from the payment method](https://github.com/activemerchant/active_merchant/blob/b1c5c11812a2284a2a9c566e5efd34064006040b/lib/active_merchant/billing/credit_card.rb#L117-L122) instead of having the merchant pass in data in the options we already have.

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+            "surname" => options[:lastname]
+          }
+        }
+        post["order"]["companyInformation"] = {
+          "name" => options[:company]
+        }
+        post["order"]["contactDetails"] = {
+          "emailAddress" => options[:email],
+          "phoneNumber" => options[:phone]
```
rwdaigle: We should get the phone number from the existing [address options](https://github.com/activemerchant/active_merchant/wiki/Contributing#address-hash) if they exist

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+        }
+      end
+
+      def add_refund_customer_data(post, options)
+        if address = options[:billing_address] || options[:address]
+          post["customer"]["address"] = {
+            "countryCode" => address[:country]
+          }
+          post["customer"]["address"]["name"] = {
```
rwdaigle: ditto on name from pm here too?

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+        when :refund
+          uri + "payments/#{authorization}/refund"
+        when :void
+          uri + "payments/#{authorization}/cancel"
+        end
+      end
+
+      def commit(action, post, authorization = nil)
+        begin
+        response = parse(ssl_post(url(action, authorization), post.to_json, headers(action, post, authorization)))
```
rwdaigle: should indent under `begin`?

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+        rescue ResponseError => e
+          raise unless(e.response.code.to_s =~ /4\d\d/)
```
rwdaigle: I think a numerical comparison here might be more clear?

```ruby
raise e if e.response.code.to_i >= 500
```

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+      end
+
+      def url(action, authorization)
+        (test? ? test_url : live_url) + uri(action, authorization)
+      end
+
+      def uri(action, authorization)
+        uri = "/v1/#{@options[:merchant_id]}/"
+        case action
+        when :test_connection
```
rwdaigle: Should we remove the test connection, especially since the hard-coded `POST` method in the digest creation would cause it to fail?

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+#{date}
+#{uri(action, authorization)}
+EOS
+        digest = OpenSSL::Digest.new('sha256')
+        key = @options[:secretApiKey]
+        "GCS v1HMAC:#{@options[:apiKeyId]}:#{Base64.strict_encode64(OpenSSL::HMAC.digest(digest, key, data))}"
+      end
+
+      def date
+        @date ||= Time.now.strftime("%a, %d %b %Y %H:%M:%S") + " GMT" # Must be same in digest and HTTP header
```
rwdaigle: Instead of hard-coding the timezone, we should probably do:

```
Time.now.utc.strftime("%a, %d %b %Y %H:%M:%S %Z")
```

Otherwise, the timestamp will actually be off by 4-5 hours (the diff between GMT and EDT - the local timezone)

lib/active_merchant/billing/gateways/globalcollect.rb
```diff
@@ -0,0 +1,294 @@
...
+
+      def authorization_from(succeeded, response)
+        if succeeded
+          response["id"] || response["payment"]["id"] || response["paymentResult"]["payment"]["id"]
+        else
+          response["errorId"]
+        end
+      end
+
+      def error_code_from(succeeded, response)
```
rwdaigle: I wonder if `response[:errorId]` isn't the better error code to use (it's what's used to determine if it was an unsuccessful response)?

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: The purpose of this parameter is a bit vague and meant to be whatever the merchant wants. Passing `merchant` could be a good use, but I think it's meant to still refer to the transaction specifically.  Could concatenate `merchant + orderid`.

> merchantOrderId
> integer (10)
> Your order ID for this transaction that is also returned in our report files
>
> merchantReference	
> string (30)
> Your unique reference of the transaction that is also returned in our report files. This is almost always used for your reconciliation of our report files.

test/remote/gateways/remote_globalcollect_test.rb
```diff
@@ -0,0 +1,160 @@
...
+    @declined_card = credit_card('5424180279791732')
+    @accepted_amount = 4005
+    @rejected_amount = 2997
+    @options = {
+      billing_address: address,
+      description: 'Store Purchase'
+    }
+  end
+
+#   def test_digest
```
rwdaigle: We should remove debug code from the adapter submission.

test/remote/gateways/remote_globalcollect_test.rb
```diff
@@ -0,0 +1,160 @@
...
+    assert_success capture
+  end
+
+  def test_failed_capture
+    response = @gateway.capture(@amount, '123')
+    assert_failure response
+    assert_equal 'INVALID_PAYMENT_ID', response.message
+  end
+
+  # Because payments are not fully authorized immediately, refunds can only be
```
rwdaigle: (but can leave this)

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: Will do.

lib/active_merchant/billing/gateways/globalcollect.rb
rwdaigle: Let's do this - since order id is truncated to a pretty short length (10), let's leave it empty and put the order id only in the merchantReference field? That way there's no loss of data.

This assumes the merchantOrderId isn't required?

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: Good catch, meant to strip out the conditional there.

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: Many other adapters use the regex there, but now that you point it out, this should in fact be 
`raise e if e.response.code.to_i >= 300`

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: Will do.

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: Something made me think any server would be on GMT.  Obviously not, good catch.

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: `errorId` is a unique string that refers to the attempted transaction; `[errors][code]` gives a meaningful response that theoretically aligns with the error message.

lib/active_merchant/billing/gateways/globalcollect.rb
rwdaigle: The 3xx response codes just mean "redirect", so would probably keep it at 400 and above?

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: I'm still fuzzy on MultiResponse and its arguments, but that sounds right. Would the same be true of `verify`?

lib/active_merchant/billing/gateways/globalcollect.rb
curiousepic: It's not, good idea.

lib/active_merchant/billing/gateways/globalcollect.rb
rwdaigle: So MultiResponse is a way to represent more than one request as a single response. By default the status of the last response is proxied as the state of the overall tx. But, if you specify `:user_first_response` instead, it will use the first response as its proxy.

So you want to choose the call that best represents the status of the tx. For a purchase that's really an auth and a capture underneath the covers, you can't say the purchase has succeeded unless the capture has succeeded.

However, for `verify` which is an auth and void, even if the void doesn't succeed the auth response is how we know if the card is valid or not. If the auth succeeds and the void fails, the funds just won't ever be collected, so no harm done. So in that case `:use_first_response` is most accurate.

curiousepic: Got it; that helps.

PR #2050, opened 2016-03-21
Gateway/visanet peru

shasum:
@rwdaigle Visanet Peru Gateway Adapter

### DIFF
--- a/lib/active_merchant/billing/gateways/visanet_peru.rb
+++ b/lib/active_merchant/billing/gateways/visanet_peru.rb
@@ -30,9 +30,7 @@ module ActiveMerchant #:nodoc:
 
         add_invoice(post, amount, options)
         add_payment_method(post, payment_method)
-        add_customer_data(post, options)
-        add_device_fingerprint_data(post, options)
-        add_merchant_define_data(post, options)
+        add_antifraud_data(post, options)
 
         # No vaulting for now
         post[:createAlias] = false
@@ -64,26 +62,12 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      # def refund(amount, authorization, options={})
-      # end
-
-      # def credit(amount, payment_method, options={})
-      # end
-
       def verify(credit_card, options={})
         MultiResponse.run(:use_first_response) do |r|
           r.process { authorize(100, credit_card, options) }
           r.process(:ignore_result) { void(r.authorization, options) }
         end
       end
-
-      # def store(payment_method, options = {})
-      #   post = {}
-      #   add_payment_method(post, payment_method)
-      #   add_customer_data(post, options)
-
-      #   commit("store", post)
-      # end
 
       def supports_scrubbing?
         true
@@ -104,10 +88,10 @@ module ActiveMerchant #:nodoc:
       CURRENCY_CODES["PEN"] = 604
 
       def add_invoice(post, money, options)
-        post[:amount] = amount(money).to_f
         # Visanet Peru expects a 9-digit numeric purchaseNumber
         post[:purchaseNumber] = options[:order_id]
         post[:externalTransactionId] = options[:order_id]
+        post[:amount] = amount(money).to_f
         post[:currencyId] = CURRENCY_CODES[options[:currency] || currency(money)]
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+      self.test_url = "https://devapi.vnforapps.com/api.tokenization/api/v2/merchant"
+      self.live_url = "https://api.vnforapps.com/api.tokenization/api/v2/merchant"
+
+      self.supported_countries = ["US", "PE"]
+      self.default_currency = "PEN"
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      def initialize(options={})
+        requires!(options, :login, :password)
```
rwdaigle: We should use credential names that directly reflect what the gateway calls them, i.e. `:access_key_id` and `:secret_access_key` for VisaNet.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+        puts options
+        case action
+        when "authorize"
+          commit("void", params, options)
+        when "capture"
+          commit("cancel", params, options)
+        end
+      end
+
+      # def refund(amount, authorization, options={})
```
rwdaigle: You can go ahead and remove this dead code.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+      # end
+
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      # def store(payment_method, options = {})
```
rwdaigle: Ditto for removing here.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+        antifraud = {}
+        billing_address = options[:billing_address] || options[:address]
+        if (billing_address)
+          antifraud[:billTo_street1] = billing_address[:address1]
+          antifraud[:billTo_city] = billing_address[:city]
+          antifraud[:billTo_state] = billing_address[:state]
+          antifraud[:billTo_country] = billing_address[:country]
+          antifraud[:billTo_postalCode]    = billing_address[:zip]
+        end
+        post[:antifraud] = antifraud
```
rwdaigle: Instead of initializing an empty `:antifraud` element here, perhaps we should check/initialize where it's needed instead, which I believe is just in `add_device_fingerprint_data` and `add_merchant_define_data`?

And while I'm looking at it, it seems those two methods aren't ever called independent of each other. Can we combine them into one `add_antifraud_data` method?

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+          post[:antifraud][:merchantDefineData] = merchantDefineData
+        end
+      end
+
+      def commit(action, params, options)
+        case action
+        when "authorize"
+          url = base_url() + "/" + options[:merchant_id]
+          method = :post
+        when "capture"
```
rwdaigle: For these three `purchaseNumber` based operations, why don't we just pass in the right string to use in the URL vs. having to map from a "capture" to a "deposit"? So the calls to `commit` will look like:

```ruby
commit("deposit", post, options)
commit("void", post, options)
commit("cancelDeposit", post, options)
```

I guess you can leave "authorize" since it has no direct mapping.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+        # No vaulting for now
+        post[:createAlias] = false
+
+        commit("authorize", post, options)
+      end
+
+      def capture(authorization, options={})
+        params = {}
+        action, merchant_id, purchase_number = split_authorization(authorization)
+        options[:merchant_id] = merchant_id
```
rwdaigle: Usually we treat `options` as a read-only hash. So instead of assigning `merchant_id` to a new option, just add it to the `params` hash.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+        commit("capture", params, options)
+      end
+
+      def void(authorization, options={})
+        params = {}
+        action, merchant_id, purchase_number = split_authorization(authorization)
+        options[:merchant_id] = merchant_id
+        options[:purchaseNumber] = purchase_number
+        params[:externalTransactionId] = purchase_number
+        puts options
```
rwdaigle: Let's get rid of debug output too :)

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+      private
+
+      CURRENCY_CODES = Hash.new{|h,k| raise ArgumentError.new("Unsupported currency: #{k}")}
+      CURRENCY_CODES["USD"] = 840
+      CURRENCY_CODES["PEN"] = 604
+
+      def add_invoice(post, money, options)
+        post[:amount] = amount(money).to_f
+        # Visanet Peru expects a 9-digit numeric purchaseNumber
+        post[:purchaseNumber] = options[:order_id]
```
rwdaigle: Let's go ahead and do this for the purchaseNumber:

```
post[:purchaseNumber] = options[:purchase_number] || rand(999999999)
```

Since it's such a specific format that should probably get truncated, we shouldn't try to conflate it with the "real" order id. Let's give it its own option and generate a default one if not present?

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -0,0 +1,239 @@
...
+          response_error(raw_response)
+        rescue JSON::ParserError
+          unparsable_response(raw_response)
+        else
+          Response.new(
+            success_from(response),
+            message_from(response),
+            response,
+            :test => test?,
+            :authorization => action + "|" + (response["merchantId"] || '') + "|" + (response["externalTransactionId"] || ''),
```
rwdaigle: Can we move this concatenation logic to its own method, just to keep it clearly delineated?

rwdaigle:
Looks great, @shasum! Just a few comments sprinkled here and there.

You can push new commits to your branch to get them reflected here. @-mention me in a comment and I can merge it in!

shasum:
@rwdaigle new commits are in addressing feedback items

PR #2051, opened 2016-03-24
feat/securionpay store

shasum:
@rwdaigle 
added store and customer methods to securion_pay

### DIFF
--- a/lib/active_merchant/billing/gateways/securion_pay.rb
+++ b/lib/active_merchant/billing/gateways/securion_pay.rb
@@ -79,6 +79,29 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def store(credit_card, options = {})
+        if options[:customer_id].blank?
+          MultiResponse.run() do |r|
+            #create charge object
+            r.process { authorize(100, credit_card, options) }
+            #create customer and save card
+            r.process { create_customer_add_card(r.authorization, options) }
+            #void the charge
+            r.process(:ignore_result) { void(r.params["metadata"]["chargeId"], options) }
+          end
+        else
+          verify(credit_card, options)
+        end
+      end
+
+      def customer(options = {})
+        if options[:customer_id].blank?
+          return nil
+        else
+          commit("customers/#{CGI.escape(options[:customer_id])}", nil, options, :get)
+        end
+      end
+
       def supports_scrubbing?
         true
       end
@@ -92,8 +115,18 @@ module ActiveMerchant #:nodoc:
 
       private
 
+      def create_customer_add_card(authorization, options)
+        post = {}
+        post[:email] = options[:email]
+        post[:description] = options[:description]
+        post[:card] = authorization
+        post[:metadata] = {}
+        post[:metadata][:chargeId] = authorization
+        commit('customers', post, options)
+      end
+
       def add_customer(post, payment, options)
-        post[:customer] = options[:customer] if options[:customer]
+        post[:customerId] = options[:customer_id] if options[:customer_id]
       end
 
       def add_customer_data(post, options)
@@ -156,8 +189,8 @@ module ActiveMerchant #:nodoc:
         JSON.parse(body)
       end
 
-      def commit(url, parameters = nil, options = {})
-        response = api_request(url, parameters, options)
+      def commit(url, parameters = nil, options = {}, method = nil)
+        response = api_request(url, parameters, options, method)
         success = !response.key?("error")
 
         Response.new(success,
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -92,8 +115,18 @@ def scrub(transcript)
...
+        post[:description] = options[:description]
+        post[:card] = authorization
+        post[:metadata] = {}
+        post[:metadata][:chargeId] = authorization
+        commit('customers', post, options)
+      end
+
       def add_customer(post, payment, options)
-        post[:customer] = options[:customer] if options[:customer]
+        post[:customerId] = options[:customer_id] if options[:customer_id]
```
rwdaigle: Is the `:customer` field name not supported at all (meaning this never worked)?

If so, we can be a little more liberal in introducing this breaking change.

lib/active_merchant/billing/gateways/securion_pay.rb
```diff
@@ -79,6 +79,29 @@ def verify(credit_card, options={})
...
+          MultiResponse.run() do |r|
+            #create charge object
+            r.process { authorize(100, credit_card, options) }
+            #create customer and save card
+            r.process { create_customer_add_card(r.authorization, options) }
+            #void the charge
+            r.process(:ignore_result) { void(r.params["metadata"]["chargeId"], options) }
+          end
+        else
+          verify(credit_card, options)
```
rwdaigle: I'm curious how doing a `verify` will automatically store the card?

What does the active merchant response look like from verify?

lib/active_merchant/billing/gateways/securion_pay.rb
shasum: As far as I checked, :customer was only being populated in the adapter's auth/purchase methods. Both those methods send out POST requests to create a new Charge. The create new charge documentation lists :customerId as a data input (and not :customer). Also, :customer has never been exercised in any of the existing tests. All this lead to believe :customer has either been unused or erroneous.

lib/active_merchant/billing/gateways/securion_pay.rb
shasum: When provided a valid customerId, the act of creating a new charge automatically adds the card to the customer (relevant paragraph from api doc pasted below). The authorize stage of verify does that job. Although subsequent void refunds the charge, the card stays added to the customer since the authorize step was successful. The 3rd test-case within 'test_successful_store' verifies this. Verify response is pasted further down below. The newly added card's id is available within the response params.

> customerId
> string, optional (either customerId or card is required)
> Identifier of customer that will be associated with this charge.
> This field is required if charge is being created with customer's existing card.
> If specified then successful charge created with new card (either by token or by card details) will add that card to customer's card and is will be set as customer's default card.

`{"id":"char_fIZbI9jfq4t6TeE1n0o0fe6X","created":1458839587,"objectType":"charge","amount":100,"currency":"USD","description":"ActiveMerchant test charge","card":{"id":"card_JmBx1cu2yeFxdk6YIWMrUURd","created":1458839587,"objectType":"card","first6":"424242","last4":"4242","fingerprint":"ecAKhFD1dmDAMKD9","expMonth":"9","expYear":"2017","cardholderName":"Longbob Longsen","brand":"Visa","type":"Credit Card"},"captured":false,"refunded":false,"disputed":false,"metadata":{"email":"foo@example.com"}}`

`   @primary_response=
    #<ActiveMerchant::Billing::Response:0x007ffce4123b38
     @authorization="char_fIZbI9jfq4t6TeE1n0o0fe6X",
     @avs_result=
      {"code"=>nil, "message"=>nil, "street_match"=>nil, "postal_match"=>nil},
     @cvv_result={"code"=>nil, "message"=>nil},
     @emv_authorization=nil,
     @error_code=nil,
     @fraud_review=nil,
     @message="Transaction approved",
     @params=
      {"id"=>"char_fIZbI9jfq4t6TeE1n0o0fe6X",
       "created"=>1458839587,
       "objectType"=>"charge",
       "amount"=>100,
       "currency"=>"USD",
       "description"=>"ActiveMerchant test charge",
       "card"=>
        {"id"=>"card_JmBx1cu2yeFxdk6YIWMrUURd",
         "created"=>1458839587,
         "objectType"=>"card",
         "first6"=>"424242",
         "last4"=>"4242",
         "fingerprint"=>"ecAKhFD1dmDAMKD9",
         "expMonth"=>"9",
         "expYear"=>"2017",
         "cardholderName"=>"Longbob Longsen",
         "brand"=>"Visa",
         "type"=>"Credit Card"},
       "captured"=>false,
       "refunded"=>false,
       "disputed"=>false,
       "metadata"=>{"email"=>"foo@example.com"}},
     @success=true,
     @test=true
`

lib/active_merchant/billing/gateways/securion_pay.rb
rwdaigle: Confirmed that it's nowhere to be found in their docs.

lib/active_merchant/billing/gateways/securion_pay.rb
rwdaigle: Ok, I understand now. Thank you!

PR #2052, opened 2016-03-28
Add all optional fields to SagePay requests

cristianstanescu:
SagePay direct integration allows sending multiple optional fields with requests. This PR add all missing optional fields available with SagePay's protocol version 3.00 as described in the "Sage Pay Direct Integration and Protocol Guidelines 3.00" appendix "A1. You submit your transaction registration POST"

### DIFF
--- a/lib/active_merchant/billing/gateways/sage_pay.rb
+++ b/lib/active_merchant/billing/gateways/sage_pay.rb
@@ -43,6 +43,26 @@ module ActiveMerchant #:nodoc:
         "NOTMATCHED" => 'N'
       }
 
+      OPTIONAL_REQUEST_FIELDS = {
+        paypal_callback_url: :PayPalCallbackURL,
+        basket: :Basket,
+        gift_aid_payment: :GiftAidPayment ,
+        apply_avscv2: :ApplyAVSCV2 ,
+        apply_3d_secure: :Apply3DSecure,
+        account_type: :AccountType,
+        billing_agreement: :BillingAgreement,
+        basket_xml: :BasketXML,
+        customer_xml: :CustomerXML,
+        surcharge_xml: :SurchargeXML,
+        vendor_data: :VendorData,
+        language: :Language,
+        website: :Website,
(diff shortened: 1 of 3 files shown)


timbeiko:
LGTM. @aprofeit can you please have a look?

lib/active_merchant/billing/gateways/sage_pay.rb
```diff
@@ -195,14 +215,11 @@ def add_customer_data(post, options)
       end
 
       def add_optional_data(post, options)
-        add_pair(post, :GiftAidPayment, options[:gift_aid_payment]) unless options[:gift_aid_payment].blank?
```
aprofeit: The logic seems to have slightly changed here, previously it would add nothing if `option[:gift_aid_payment]` was `blank?`, now it looks like it will include `nil` for all keys which aren't present in `options`. Was this intentional / are we sure this is okay?

timbeiko: I had a look and `add_pair` handles this because it will only add a pair if the value isn't blank or if the option is required:
```
def add_pair(post, key, value, options = {})
          post[key] = value if !value.blank? || options[:required]
end 
```

aprofeit: Nice!

aprofeit:
One nice thing would be to get the remote tests using these options just to make sure it's not going to blow up.

timbeiko:
Agreed, @cristianstanescu could you please add them and then I'll merge the PR? Thanks!

cristianstanescu:
Thank you @timbeiko and @aprofeit for picking this up and helping getting it merged. Especially as we've also started working on implementing 3D Secure with SagePay on our system.

I've added the remote tests and rebased. A bit sloppy on my side not to include the remote tests when the PR was made. I hope it's ok now. I've tested using our SagePay account.

timbeiko:
@cristianstanescu thanks! LGTM. @aprofeit what do you think?

aprofeit:
👍

PR #2053, opened 2016-03-29
Add AVS support to Barclaycard Smartpay

curiousepic:
@rwdaigle to review

### DIFF
--- a/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
+++ b/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
@@ -75,8 +75,6 @@ module ActiveMerchant #:nodoc:
       end
 
       def store(creditcard, options = {})
-        # ??? require :email and :customer, :order_id?
-
         post = store_request(options)
         post[:card] = credit_card_hash(creditcard)
         post[:recurring] = {:contract => 'RECURRING'}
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -97,16 +95,43 @@ def scrub(transcript)
...
+        '18' => 'I'	  # Neither postal code nor address were checked
+      }
+
       def commit(action, post)
         request = post_data(flatten_hash(post))
         raw_response = ssl_post(build_url(action), request, headers)
         response = parse(raw_response)
 
+        # AVS only present if enabled for account by gateway
+        avs_code = response["avsResult"][0..1].strip if response["avsResult"]
```
rwdaigle: I'd move this to its own method just to make this a bit tidier?

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -163,7 +188,7 @@ def post_data(data)
 
       def message_from(response)
         return response['resultCode'] if response.has_key?('resultCode') # Payment request
-        return response['response'] if response['response'] # Modification request
+        return response['response'] if response.has_key?('response') # Modification request
```
rwdaigle: Why is this necessary?

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -34,7 +34,7 @@ def authorize(money, creditcard, options = {})
         post[:card] = credit_card_hash(creditcard)
 
         if address = (options[:billing_address] || options[:address])
-          post[:billingAddress] = address_hash(address)
+          post[:card][:billingAddress] = address_hash(address)
```
rwdaigle: So this was a bug in the original version of the adapter?

curiousepic: I had changed this when I began working on the update a month ago and honestly don't know why.  Remote tests pass without this nesting, so I will remove the change.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
curiousepic: Will do.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
curiousepic: It's not, just noticed it and tidied it up; but in fact it probably works without .has_key? which would be even tidier.  But I'll remove it since it's unrelated to the feature.

rwdaigle:
@curiousepic as an FYI, it's good to @-mention the reviewer when updates to the PR are made - I don't get a notification when you push new code to the PR, nor can I know if that resolves all the issues.

I'm assuming this is good to go again?

curiousepic:
@rwdaigle Good to know!

PR #2054, opened 2016-03-30
Add currency and customer in purchase response

darkaz:
Change requested by Spreedly customer (www.agavesanmiguel.com)


duff:
Hi @darkaz,

Thanks for the PR!  I brought in the currency part but thought that the customer part could wait until we have a customer who needs it.  I also wonder if the customer logic could be simplified to be a bit more consistent with how other gateways are handling the `email` and `phone_number` fields.

PR #2056, opened 2016-04-05
[visanet peru] modified to handle empty billing address

shasum:
@rwdaigle 
modified to handle empty billing address
cvv scrub modified to handle string value


lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -77,7 +77,7 @@ def scrub(transcript)
         transcript.
           gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
           gsub(%r((\"cardNumber\\\":\\\")\d+), '\1[FILTERED]').
-          gsub(%r((\"cvv2Code\\\":)\d+), '\1[FILTERED]')
+          gsub(%r((\"cvv2Code\\\":\\\")\d+), '\1[FILTERED]')
```
rwdaigle: I would have expected a new test here to confirm this functionality? Or a modification to an existing test if it wasn't properly testing scrubbing before?

shasum: @rwdaigle 
commit 7fabf38 modified CVV code from integer to string causing existing scrub test to fail. This fixes it.

rwdaigle: @duff FYI you might have committed something w/o running the tests? Would be surprised...

duff: @rwdaigle Hmmm...  All the tests seem to be working on master for me.

shasum: @duff 
does remote test_transcript_scrubbing work for you?
unit may have worked as the request response are mocked

duff: Ahhh..  Good catch @shasum!  10 out of the 11 remote tests for this gateway are failing for me because I don't have credentials. Sorry!

Thanks for the fix!

Any chance we can add test credentials to our fixtures file here?  https://github.com/activemerchant/active_merchant/blob/master/test/fixtures.yml#L1015

duff: I should have said all of the *unit* tests were working for me on master in my initial reply. :)

shasum: @duff my bad for not committing the test credentials. I wasn't sure if it's okay to commit them to the repo. A quick check with @rwdaigle and I can add them.

rwdaigle: @duff they're in our 1P since I believe they're client creds.

duff: :+1: All remote tests working now.  Thanks!

PR #2063, opened 2016-04-07
Fix incorrect handling of response body as params in a rare Paymill error response

methodmissing:
As per gateway issues experienced this morning.

@girasquid

### DIFF
--- a/lib/active_merchant/billing/gateways/paymill.rb
+++ b/lib/active_merchant/billing/gateways/paymill.rb
@@ -154,7 +154,7 @@ module ActiveMerchant #:nodoc:
         begin
           raw_response = ssl_request(:get, "#{save_card_url}?#{post_data(post)}", nil, {})
         rescue ResponseError => e
-          return Response.new(false, e.response.body, e.response.body, {})
+          return Response.new(false, e.response.body)
         end
 
         response_for_save_from(raw_response)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/paymill.rb
```diff
@@ -154,7 +154,7 @@ def save_card(credit_card)
         begin
           raw_response = ssl_request(:get, "#{save_card_url}?#{post_data(post)}", nil, {})
         rescue ResponseError => e
-          return Response.new(false, e.response.body, e.response.body, {})
+          return Response.new(false, e.response.body)
```
girasquid: Should the second `e.response.body` here be `e.response.params`?

girasquid:
One question, but looks ok otherwise.

lib/active_merchant/billing/gateways/paymill.rb
methodmissing: Looked into that earlier, but noticed other gateways return with `Response.new(false, "a message")` for these cases. It requires changes to the `MockResponse` in the test too (doesn't implement `params`) and wanted to keep changes minimal for this edge case.

girasquid: That's fair - this works, then!

methodmissing:
@rwdaigle

PR #2064, opened 2016-04-07
Latitude19 gateway support

shasum:
@rwdaigle 

known issues: no declined card, so missing several fail scenarios. void-a-capture fails, pending clarification from gateway support.

rest ready for review.

### DIFF
--- a/lib/active_merchant/billing/gateways/latitude19.rb
+++ b/lib/active_merchant/billing/gateways/latitude19.rb
@@ -143,8 +143,6 @@ module ActiveMerchant #:nodoc:
 
       def scrub(transcript)
         transcript.
-          gsub(%r((\"pgwAccountNumber\\\":\\\")\d+), '\1[FILTERED]').
-          gsub(%r((\"pgwConfigurationId\\\":\\\")\d+), '\1[FILTERED]').
           gsub(%r((\"cardNumber\\\":\\\")\d+), '\1[FILTERED]').
           gsub(%r((\"cvv\\\":\\\")\d+), '\1[FILTERED]')
       end
--- a/lib/active_merchant/billing/gateways/latitude19.rb
+++ b/lib/active_merchant/billing/gateways/latitude19.rb
@@ -7,7 +7,7 @@ module ActiveMerchant #:nodoc:
       self.live_url = "https://gateway.l19tech.com/payments/"
       self.test_url = "https://gateway-sb.l19tech.com/payments/"
 
-      self.supported_countries = ["US"]
+      self.supported_countries = ["US", "CA"]
       self.default_currency = "USD"
       self.money_format = :cents
       self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb]
@@ -57,8 +57,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def purchase(amount, payment_method, options={})
-        if options[customer_account_number]
-          auth_or_sale("sale", nil, options[customer_account_number], amount, payment_method, options)
+        if options["customer_account_number"]
+          auth_or_sale("sale", nil, options["customer_account_number"], amount, payment_method, options)
         else
           MultiResponse.run() do |r|
             r.process { get_session(options) }
@@ -69,8 +69,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def authorize(amount, payment_method, options={})
-        if options[customer_account_number]
-          auth_or_sale("auth", nil, options[customer_account_number], amount, payment_method, options)
+        if options["customer_account_number"]
+          auth_or_sale("auth", nil, options["customer_account_number"], amount, payment_method, options)
         else
           MultiResponse.run() do |r|
             r.process { get_session(options) }
@@ -81,6 +81,23 @@ module ActiveMerchant #:nodoc:
       end
 
       def capture(amount, authorization, options={})
+        post = {}
+        post[:method] = "deposit"
+
+        add_request_id(post)
+
+        params = {}
+
+        add_invoice(params, amount, options)
+
+        params[:pgwTID] = authorization
+        params[:pgwAccountNumber] = @options[:account_number]
+        params[:pgwConfigurationId] = @options[:configuration_id]
+        params[:pgwHMAC] = OpenSSL::HMAC.hexdigest('sha512', @options[:secret], message(params, post[:method]))
+
+        post[:params] = [params]
+
+        commit("v1/", post)
       end
 
       def void(authorization, options={})
@@ -90,14 +107,33 @@ module ActiveMerchant #:nodoc:
       end
 
       def credit(amount, payment_method, options={})
-      end
-
-      def verify(credit_card, options={})
-        MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
-          r.process(:ignore_result) { void(r.authorization, options) }
+        if options["customer_account_number"]
+          refundWithCard(nil, options["customer_account_number"], amount, payment_method, options)
+        else
+          MultiResponse.run() do |r|
+            r.process { get_session(options) }
+            r.process { get_token(r.authorization, payment_method, options) }
+            r.process { refundWithCard(r.authorization, nil, amount, payment_method, options) }
+          end
         end
       end
+
+      def verify(payment_method, options={})
+        if options["customer_account_number"]
+          verifyOnly(nil, options["customer_account_number"], payment_method, options)
+        else
+          MultiResponse.run() do |r|
+            r.process { get_session(options) }
+            r.process { get_token(r.authorization, payment_method, options) }
+            r.process { verifyOnly(r.authorization, nil, payment_method, options) }
+          end
+        end
+      end
+
+        # MultiResponse.run(:use_first_response) do |r|
+        #   r.process { authorize(100, credit_card, options) }
+        #   r.process(:ignore_result) { void(r.authorization, options) }
+        # end
 
       # def store(payment_method, options = {})
       #   post = {}
@@ -145,7 +181,7 @@ module ActiveMerchant #:nodoc:
         if method == "getSession"
           params[:pgwAccountNumber] + "|" + params[:pgwConfigurationId] + "|" + params[:requestTimeStamp] + "|" + method
         else
-          params[:pgwAccountNumber] + "|" + params[:pgwConfigurationId] + "|" + params[:orderNumber] + "|" + method + "|" + (params[:amount] || "") + "|" + (params[:sessionToken] || "") + "|" + (params[:accountToken] || "")
+          params[:pgwAccountNumber] + "|" + params[:pgwConfigurationId] + "|" + (params[:orderNumber] || "") + "|" + method + "|" + (params[:amount] || "") + "|" + (params[:sessionToken] || "") + "|" + (params[:accountToken] || "")
         end
       end
 
@@ -163,8 +199,6 @@ module ActiveMerchant #:nodoc:
         post[:params] = [params]
 
         commit("session", post)
-        # add_payment_method(post, payment_method)
-        # add_customer_data(post, options)
       end
 
       def get_token(session_id, payment_method, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+        "master" => "MC",
+        "visa" => "VI",
+        "american_express" => "AX",
+        "discover" => "DS",
+        "diners_club" => "DC",
+        "jcb" => "JC"
+      }
+
+      def initialize(options={})
+        requires!(options, :account_number, :configuration_id, :secret)
```
rwdaigle: How is the `configuration_id` described by the gateway? Is it something set once per merchant account, or something that can vary with each transaction?

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+        super
+      end
+
+      def purchase(amount, payment_method, options={})
+        if options[:account_token]
```
rwdaigle: Explain what having `account_token` present means/does? Looks like it bypasses the normal purchase path?

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+        else
+          response["result"]["lastActionSucceeded"] == 1 ? "Succeeded" : "Failed"
+        end
+      end
+
+      def error_from(response)
+        return response["error"] if response["error"]
+
+        if response["result"].key?("pgwResponseCode")
+          "pgwResponseCode|" + response["result"]["pgwResponseCode"] + "|pgwResponseCodeDescription|" + RESPONSE_CODE_MAPPING[response["result"]["pgwResponseCode"]] + "|responseText|" + (response["result"]["responseText"] || "") + "|processorResponseCode|" + (response["result"]["processor"]["responseCode"] || "")
```
rwdaigle: This method is being used to populate the `error_code` attribute of the response. We should keep it simple and just return the `pgwResponseCode` attribute.

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+        end
+      end
+
+      def message_from(response)
+        return response["error"] || "HTTP Response Error" unless response.key?("result")
+
+        return response["error"] if response["error"]
+
+        if response["result"].key?("pgwResponseCode")
+          (RESPONSE_CODE_MAPPING[response["result"]["pgwResponseCode"]] || "") + "|" + (response["result"]["responseText"] || "")
```
rwdaigle: Should keep it simple and just return one error message (not a concatenation)

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+        end
+      end
+
+      def authorization_from(response, method)
+        method + "|" + (
+          response["result"]["sessionId"] ||
+          response["result"]["sessionToken"] ||
+          response["result"]["pgwTID"] ||
+          response["result"]["accountToken"]
+        )
```
rwdaigle: Dear god, are all these really necessary in follow-up calls?

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+
+      def avs_from(response)
+        response["result"].key?("avsResponse") ? AVSResult.new(code: response["result"]["avsResponse"]) : nil
+      end
+
+      def cvv_from(response)
+        response["result"].key?("cvvResponse") ? CVVResult.new(response["result"]["cvvResponse"]) : nil
+      end
+
+      def response_error(raw_response)
```
rwdaigle: Is the error handling here redundant with what's in `commit`?

lib/active_merchant/billing/gateways/latitude19.rb
shasum: >Merchant’s configuration identifier on the L19 Payment Gateway. A merchant may have one, or more than one con- figuration identifier. The configuration identifier represents a mer- chant "division," and it determines how the transaction is routed by the back-end payment processor. The most common reason for having multiple configurations is to support multiple payment currencies. In this case, a merchant will have one configuration identifier per currency. Merchant configurations must be set up by the L19 Payment Gateway in conjunction with the back-end payment processor.

lib/active_merchant/billing/gateways/latitude19.rb
shasum: `account_token` represents a stored card

lib/active_merchant/billing/gateways/latitude19.rb
shasum: `responseText` conveys a meaningful transaction message. Thought it might be useful, in addition to the decoded response code.

lib/active_merchant/billing/gateways/latitude19.rb
shasum: Okay. It has 4 parts. I'll strip out the error descriptions. Does it make sense to leave in `processorResponseCode` in addition to `pgwResponseCode` (the lat19 gateway response code)

lib/active_merchant/billing/gateways/latitude19.rb
shasum: Yes sir. It's a chatty api [in your own words]  :)

Thank god, it's a big OR and not a concatenation, again :)

lib/active_merchant/billing/gateways/latitude19.rb
shasum: @rwdaigle 
I assume you're referring to the JSON error handler within `response_error()` method. It is needed. The one sitting inside `commit` handles JSON errors in a successful http response, this latter one handles it for `ResponseError` object, generated when status code falls outside of 200..300 range. Not sure if I'm addressing your question correctly.

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+
+        params[:requestTimeStamp] = Time.now.getutc.strftime("%Y%m%d%H%M%S") if method == "getSession"
+
+        params[:pgwHMAC] = OpenSSL::HMAC.hexdigest('sha512', @options[:secret], message(params, method))
+      end
+
+      def add_invoice(params, money, options)
+        params[:amount] = amount(money)
+        params[:orderNumber] = options[:order_id]
+        params[:transactionClass] = "eCommerce" || options[:transaction_class]
```
rwdaigle: This will always return `eCommerce`, is that intended?

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,411 @@
...
+        verify(payment_method, options, "store")
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((\"pgwAccountNumber\\\":\\\")\d+), '\1[FILTERED]').
```
rwdaigle: We only need to scrub sensitive values (passwords, secrets, PANs etc...) from the transcript. Are account number and configuration id sensitive/not public?

test/fixtures.yml
```diff
@@ -371,6 +371,11 @@ jetpay:
 komoju:
   login: sk_f1dd75ce3d5cad477eac0c827c1cac8eaa51ede3
 
+latitude19:
```
rwdaigle: If we have a test only creds, we can commit them

lib/active_merchant/billing/gateways/latitude19.rb
shasum: They most likely are not public. Not sure if they fall under the sensitive category. Did it as a precautionary step. I'll remove them from scrub.

lib/active_merchant/billing/gateways/latitude19.rb
shasum: Bug. Thanks for spotting it. Supposed to be override/default.

test/fixtures.yml
shasum: @rwdaigle
 
Not 100% sure about this part. Right now the creds (`account_number, configuration_id, secret`) work only on their sandbox. From the few conversations with Mike, it seemed to me like the same `account_number` and `configuration_id` may get carried over to production. But, probably not the `secret`. Other aspect is that, these values have been setup particularly for Fonteva. So, bit unsure if this can be commiitted to the open repo.

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: Ok, it's a tough call, but let's leave it as an initialize option as it is here :+1:

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: We usually handle the stored card path using this method signature (instead of relying on an `option`):

```ruby
def purchase(amount, payment_method_or_token, options={})
  if payment_method_or_token.respond_to?(:number)
    # ...
  else
    auth_or_sale # ....
  end
end
```

Basically, the stored payment method token is passed in as the the second arg as a first-class arg vs. an option. Many gateway adapters do this if you need a reference.

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: The caller will always have access to all the response params the gateway returns if they want to get that level of info. But for the `error_code` field, we should stick to just the code itself.

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: If they're both always present, I'd choose the one that provides the greatest level of granularity/detail

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: > Thank god, it's a big OR and not a concatenation, again :)

Ah, didn't notice that. That's slightly less repulsive, then :)

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: :+1:

test/fixtures.yml
rwdaigle: If the secret is test-only (and it's really the only thing that's sensitive), then we can commit.

lib/active_merchant/billing/gateways/latitude19.rb
shasum: modified to first class arg

lib/active_merchant/billing/gateways/latitude19.rb
shasum: `account_number, configuration_id` removed from scrub.

@rwdaigle should temp (`sessionToken`) and long-term (`accountToken`) tokens be scrubbed? they represent a valid card.

lib/active_merchant/billing/gateways/latitude19.rb
shasum: Rectified

lib/active_merchant/billing/gateways/latitude19.rb
shasum: modified to return a single message. Either decoded `pgwResponseCode` or `responseText`

lib/active_merchant/billing/gateways/latitude19.rb
shasum: modified to return a single error code, `pgwResponseCode` or `processorResponseCode`

test/fixtures.yml
shasum: test creds committed to fixture

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: > should temp (sessionToken) and long-term (accountToken) tokens be scrubbed? they represent a valid card.

I don't believe so, unless they can be used to transact in isolation from other creds (like secret etc...)

shasum: @rwdaigle Alright.

hmac goes out with every request type except one, and the one without hmac contains PAN that's going to get scrubbed. so, it should be good. leaving tokens un-scrubbed  :+1:

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,416 @@
...
+      def add_request_id(post)
+        post[:id] = SecureRandom.hex(16)
+      end
+
+      def add_timestamp()
+        Time.now.getutc.strftime("%Y%m%d%H%M%S")
+      end
+
+      def add_hmac(params, method)
+        if method == "getSession"
```
rwdaigle: I believe `getSession` is an endpoint we were invoking only for testing? If so, we should remove references to it from the adapter to keep it clean and debug-free.

lib/active_merchant/billing/gateways/latitude19.rb
```diff
@@ -0,0 +1,416 @@
...
+      end
+
+      def purchase(amount, payment_method, options={})
+        if payment_method.is_a?(String)
+          auth_or_sale("sale", payment_method, amount, nil, options)
+        else
+          MultiResponse.run() do |r|
+            r.process { get_session(options) }
+            r.process { get_token(r.authorization, payment_method, options) }
+            r.process { auth_or_sale("sale", r.authorization, amount, payment_method, options) }
```
rwdaigle: Does this `auth_or_sale` call to the gateway really need both the token AND the payment method (which was already used to get the token)?

shasum: If the token is a short term token (`sessionToken` and not `accountToken`), generating the token (`tokenize`) involves only the PAN. The subsequent `purchase`/`authorize` calls need to pass in `card_expiry` and `card_type` required fields.

lib/active_merchant/billing/gateways/latitude19.rb
shasum: `getSession` is the gateway to rest of the gateway calls (excuse the pun). we generate a `session`, based on that a `token`, and based on that a `transaction`

image

rwdaigle: ha, totally missed that. Sorry!

lib/active_merchant/billing/gateways/latitude19.rb
rwdaigle: ok :+1:

PR #2072, opened 2016-04-16
Extra Braintree error messaging

jordan-brough:
Extra Braintree error messaging

If we can't find an error message from the other methods then we can fall back
to this.

The braintree gem has provided `Braintree::ErrorResult#message` since
version 2.4.0 from July 9, 2010:
https://github.com/braintree/braintree_ruby/blob/2.59.0/CHANGELOG.rdoc#240

See also:
https://developers.braintreepayments.com/reference/general/result-objects/ruby#error-results
which says:

> The errors result object will only be populated if the error is due
> to a failed validation.
> ...
> The message on the error result gives a human-readable description
> of what went wrong, regardless of the cause and nature of the error.
> The message can contain multiple error messages.
> Note: This was added in version 2.4.0

FYI: In `lib/active_merchant/billing/gateways/braintree_blue.rb` I required version 2.4.0 because that's where this feature was introduced in the Braintree gem, but version `2.50.0` is already [required by the ActiveMerchant Gemfile](https://github.com/activemerchant/active_merchant/blob/a405e03/Gemfile#L8).

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -6,7 +6,9 @@ rescue LoadError
   raise "Could not load the braintree gem.  Use `gem install braintree` to install it."
 end
 
-raise "Need braintree gem 2.x.y. Run `gem install braintree --version '~>2.0'` to get the correct version." unless Braintree::Version::Major == 2
+unless Braintree::Version::Major == 2 && Braintree::Version::Minor >= 4
+  raise "Need braintree gem >= 2.4.0. Run `gem install braintree --version '~>2.4'` to get the correct version."
+end
 
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -6,7 +6,9 @@
   raise "Could not load the braintree gem.  Use `gem install braintree` to install it."
 end
 
-raise "Need braintree gem 2.x.y. Run `gem install braintree --version '~>2.0'` to get the correct version." unless Braintree::Version::Major == 2
+unless Braintree::Version::Major == 2 && Braintree::Version::Minor >= 4
```
jordan-brough: We could probably require a higher version than `2.4.0` if we wanted. I noticed at least `2.50.0` is [required by the ActiveMerchant Gemfile](https://github.com/activemerchant/active_merchant/blob/a405e03/Gemfile#L8).

jordan-brough:
@girasquid or @duff I see you both have merged Braintree-related PRs in the past. Any thoughts on this?

timbeiko:
LGTM. @ThereExistsX @aprofeit can you have a quick look?

aprofeit:
👍

ThereExistsX:
🌈

Issue #2058, opened 2016-04-06
VisaNet Peru exception


rwdaigle:
@shasum you're probably best positioned to look into this? Not sure under what circumstance the merchant id is not being returned?

```
#<TypeError: no implicit conversion of nil into String>

/srv/core/current/vendor/bundle/ruby/2.1.0/bundler/gems/active_merchant-fc32c3ca8d6a/lib/active_merchant/billing/gateways/visanet_peru.rb:155:in `+'
/srv/core/current/vendor/bundle/ruby/2.1.0/bundler/gems/active_merchant-fc32c3ca8d6a/lib/active_merchant/billing/gateways/visanet_peru.rb:155:in `url'
/srv/core/current/vendor/bundle/ruby/2.1.0/bundler/gems/active_merchant-fc32c3ca8d6a/lib/active_merchant/billing/gateways/visanet_peru.rb:127:in `commit'
/srv/core/current/vendor/bundle/ruby/2.1.0/bundler/gems/active_merchant-fc32c3ca8d6a/lib/active_merchant/billing/gateways/visanet_peru.rb:38:in `authorize'
/srv/core/current/vendor/bundle/ruby/2.1.0/bundler/gems/active_merchant-fc32c3ca8d6a/lib/active_merchant/billing/gateways/visanet_peru.rb:22:in `block (2 levels) in purchase'
/srv/core/current/vendor/bundle/ruby/2.1.0/bundler/gems/active_merchant-fc32c3ca8d6a/lib/active_merchant/billing/response.rb:59:in `process'
```

shasum:
@rwdaigle 
seeing that it has originated from a purchase tx, I'd suspect merchant_id has not been passed in the options object. Like we did with purchase_number, maybe it's good to provide a default in our adapter, which would then cause an error response from the gateway that can be handed back to the user?

rwdaigle:
What is the purpose of the merchant id from their docs? Is it assigned once per merchant account, or is it a value that can change with each transaction?

shasum:
@rwdaigle 
assigned once per merchant account. I believe visanet assigns it to the merchant while setting up an account. It has been shared on our Trello api test credentials section and not generated by any of our invoked api calls.

rwdaigle:
@shasum ah, ok then. I believe we should update this gateway to include `merchant_id` as a required credential that's passed to `initialize`.

And.. if we do that, we should be able to simplify the authorization format as well?

shasum:
[...]

PR #2074, opened 2016-04-19
refactor visanet peru

shasum:
@duff @rwdaigle 

changes to address items discussed in #2058

### DIFF
--- a/lib/active_merchant/billing/gateways/visanet_peru.rb
+++ b/lib/active_merchant/billing/gateways/visanet_peru.rb
@@ -42,6 +42,7 @@ module ActiveMerchant #:nodoc:
         params = {}
         _, purchase_number = split_authorization(authorization)
         params[:purchaseNumber] = purchase_number
+        params[:externalTransactionId] = options[:order_id]
         commit("deposit", params)
       end
 
@@ -49,6 +50,7 @@ module ActiveMerchant #:nodoc:
         params = {}
         action, purchase_number = split_authorization(authorization)
         params[:purchaseNumber] = purchase_number
+        params[:externalTransactionId] = options[:order_id]
 
         case action
         when "authorize"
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -88,10 +84,12 @@ def scrub(transcript)
...
-        purchase_number = options[:purchase_number] || rand(100000000 .. 1000000000).to_s
-        params[:purchaseNumber] = purchase_number
-        params[:externalTransactionId] = purchase_number
-        params[:merchantId] = options[:merchant_id]
+        regex_purchase_number = /^[0-9]{9}$/
+        if options[:order_id] && options[:order_id].to_s =~ regex_purchase_number
+          params[:purchaseNumber] = options[:order_id].to_s
+        else
+          params[:purchaseNumber] = (SecureRandom.random_number(900_000_000) + 100_000_000).to_s
+        end
```
rwdaigle: Two things:

* We should probably not have divergent behavior solely based on an option value's format
* It wasn't clear, but in my mind the right solution is [to do this](https://github.com/activemerchant/active_merchant/issues/2058#issuecomment-208538805) with the external transaction id and purchase number. Is there anything that prevents us from doing so with this new approach?

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -116,8 +114,9 @@ def add_antifraud_data(params, options)
           antifraud[:billTo_postalCode] = billing_address[:zip]
         end
 
-        antifraud[:deviceFingerprintId] = options[:device_fingerprint_id]
-        antifraud[:merchantDefineData] = options[:merchant_define_data]
+        antifraud[:deviceFingerprintId] = options[:device_fingerprint_id] || SecureRandom.hex(16)
```
rwdaigle: Is this a mandatory field?

lib/active_merchant/billing/gateways/visanet_peru.rb
shasum: @rwdaigle 

accepted. `externalTransactionId` was used in the earlier version. got rid of it thinking `order_id` being merchant supplied, will there be value in sending it back along with the response. I see now it will serve good purpose as a response identifier for the merchant. will make both changes.

lib/active_merchant/billing/gateways/visanet_peru.rb
shasum: @rwdaigle 

yes, although I didn't try NOT passing it.

image

PR #2077, opened 2016-04-22
Merchant e-Solutions: pass optional 3Dsecure params

curiousepic:
@duff to review

### DIFF
--- a/lib/active_merchant/billing/gateways/merchant_e_solutions.rb
+++ b/lib/active_merchant/billing/gateways/merchant_e_solutions.rb
@@ -1,6 +1,8 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class MerchantESolutionsGateway < Gateway
+      include Empty
+
       self.test_url = 'https://cert.merchante-solutions.com/mes-api/tridentApi'
       self.live_url = 'https://api.merchante-solutions.com/mes-api/tridentApi'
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/merchant_e_solutions.rb
```diff
@@ -122,6 +125,15 @@ def add_creditcard(post, creditcard, options)
         post[:card_exp_date]  = expdate(creditcard)
       end
 
+      # not a 3dsecure integration - simply passes params to gateway if present,
+      # verified externally
+      def add_3dsecure_params(post, options)
+        post[:xid] = options[:xid] unless options[:xid].blank?
```
duff: We're trying to lessen our dependence on ActiveSupport so we're trying to lessen the use of `blank?`.  Example of the approach we're using:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L463
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L6

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
```diff
@@ -122,6 +125,15 @@ def add_creditcard(post, creditcard, options)
         post[:card_exp_date]  = expdate(creditcard)
       end
 
+      # not a 3dsecure integration - simply passes params to gateway if present,
```
duff: I'd recommend removing the comments.

duff:
@curiousepic Looks good.  After the few little tweaks I mentioned, we should be good to go.

If you force push your branch with the adjustments, the PR should reflect it.

curiousepic:
@duff PR updated with those changes.

PR #2083, opened 2016-05-02
Added contactless-magstripe option to credit_card model

ryanbalsdon:
This is needed to select between the `contactless` and `contactless_magstripe_mode` read methods in the Stripe endpoint.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -38,6 +38,7 @@
 * Remove AN and KV country codes as they're not recognized by ISO-3166-1 [apdunston]
 * Worldpay: Pass unchanged amount with correct currency exponent [curiousepic]
 * Improve our handling of currencies sans fractions [duff]
+* Stripe: Added support for the contactless magstripe entry mode option [rbalsdon]
 
 
 == Version 1.58.0 (March 1, 2016)
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -182,10 +182,15 @@ module ActiveMerchant #:nodoc:
       # @return [String]
       attr_accessor :fallback_reason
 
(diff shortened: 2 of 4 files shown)


methodmissing:
LGTM

ryanbalsdon:
Tagging @girasquid @aprofeit and @rwdaigle as per pre-merge instructions.

rwdaigle:
@ryanbalsdon this is outside my domain, so consider this code-level feedback more than anything else. Is this distinction between contactless and contactless-magstripe only relevant for Stripe? Just wondering if adding it to `CreditCard` is the right place unless it truly is a distinction tied to the card itself and not the transaction/gateway?

ryanbalsdon:
There are currently 5 different possible sources for a credit card:
 - Manual/Web Entry
 - Swipe/Magstripe
 - Chip/EMV
 - Contactless (using EMV data)
 - Contactless (using Magstripe data)

We know that the data is:
 - Swipe by looking for non-nil track data on the credit card.
 - Chip by looking for non-nil EMV data on the credit card.
 - Contactless by looking for non-nil EMV data on the credit card _and_ a contactless flag set.

This patch adds support for knowing that the card was read from a contactless/magstripe interface by splitting the `contactless` flag into `contactless_emv` and `contactless_magstripe`. This turns the above list into:

We know that the credit card is:
 - Swipe by looking for non-nil track data on the credit card.
 - Chip by looking for non-nil EMV data on the credit card.
 - Contactless EMV by looking for non-nil EMV data on the credit card _and_ a contactless_emv flag set.
 - Contactless Magstripe by looking for non-nil EMV data on the credit card _and_ a contactless_magstripe flag set.

@rwdaigle does that help a bit more with context?

rwdaigle:
Thanks @ryanbalsdon! I think this implementation is 👍

PR #2084, opened 2016-05-02
Worldpay: Map amount exponent to currency type

curiousepic:
@duff to confirm and merge.

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -23,6 +23,8 @@ module ActiveMerchant #:nodoc:
         'switch'           => 'MAESTRO-SSL'
       }
 
+      CURRENCIES_WITHOUT_FRACTIONS = %w{HUF IDR ISK JPY KRW}
+
       def initialize(options = {})
         requires!(options, :login, :password)
         super
@@ -181,12 +183,11 @@ module ActiveMerchant #:nodoc:
 
       def add_amount(xml, money, options)
         currency = options[:currency] || currency(money)
-        amount   = localized_amount(money, currency)
 
         amount_hash = {
-          :value => amount,
+          :value => amount(money),
           'currencyCode' => currency,
-          'exponent' => 2
+          'exponent' => non_fractional_currency?(currency) ? 0 : 2
         }
 
         if options[:debit_credit_indicator]
@@ -340,13 +341,6 @@ module ActiveMerchant #:nodoc:
       def encoded_credentials
         credentials = "#{@options[:login]}:#{@options[:password]}"
         "Basic #{[credentials].pack('m').strip}"
-      end
-
-      def localized_amount(money, currency)
-        amount = amount(money)
-        return amount unless CURRENCIES_WITHOUT_FRACTIONS.include?(currency.to_s)
-
-        amount.to_i / 100 * 100
       end
     end
   end
--- a/test/remote/gateways/remote_worldpay_test.rb
+++ b/test/remote/gateways/remote_worldpay_test.rb
@@ -88,34 +88,18 @@ class RemoteWorldpayTest < Test::Unit::TestCase
     assert_equal "Could not find payment for order", response.message
   end
 
-  def test_currency
-    assert_success(result = @gateway.authorize(@amount, @credit_card, @options.merge(:currency => 'USD')))
+  def test_authorize_fractional_currency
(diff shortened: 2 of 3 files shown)


duff:
I like the explicitness of handling the exponents in a way like you've done, but I'm mulling on whether we can tie into 

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L254

I see that Stripe has its own list of currencies without fractions:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L24

Perhaps Worldpay should have its own list (based on the docs you're using -> http://support.worldpay.com/support/kb/ggse/submittingtransactionsdirect/dxmla002.html)

If we go down that path, then we may be able to do something like

```
'exponent' => non_fractional_currency?(currency) ? 0 : 2
```

curiousepic:
@duff ooo, that makes sense.  Will implement.

curiousepic:
Looking at implementing that also exposed some previous handling of non-fractional currencies that I need to reconcile with the new method as well.

curiousepic:
@duff Removed previous handling of non-fractional currencies and related tests, and I'm checking `gateway`'s list of currencies. PR updated.

lib/active_merchant/billing/gateway.rb
```diff
@@ -57,7 +57,7 @@ class Gateway
       include CreditCardFormatting
 
       DEBIT_CARDS = [ :switch, :solo ]
-      CURRENCIES_WITHOUT_FRACTIONS = %w(BIF BYR CLP CVE DJF GNF HUF ISK JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
+      CURRENCIES_WITHOUT_FRACTIONS = %w(BIF BYR CLP CVE DJF GNF HUF IDR ISK JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
```
duff: I'm not sure `IDR` should apply to all gateways.  I wonder if WorldpayGateway should have its own list of `CURRENCIES_WITHOUT_FRACTIONS` like Stripe does.

curiousepic: Hmm.  Stripe's list only has one currency that is not contained in `gateway`'s list. And I did not see `IDR` called out anywhere else. 

Currently, only Sagepay calls `gateway`'s list instead of their own.

In any case, it won't hurt to have a WorldPay-specific list, I'll go ahead and switch back to that and update.

duff: Yeah I'm on the fence about it.  I'm a little cautious about updating the Gateway list since lots of gateways are calling this:

https://github.com/duff/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L241

which is using that list.

And Wikipedia's description of the `sen` is really confusing!  https://en.wikipedia.org/wiki/Indonesian_rupiah

I'm thinking we _know_ from Worldpay's docs how they handle the `IDR` so it may be safer to only mess with Worldpay.

curiousepic:
@duff PR updated with WorldPay-specific list.

jasonwebster:
This actually caused a regression for Active Merchant users that still use Money objects for amounts. Yes, that's depreciated, but it still hasn't been removed yet. By removing the `#localized_amount` method that checked if it was a non-fractional currency, we've delegated to the default `#amount` method which always blindly calls `#cents` on any object that responds to it.

curiousepic:
Thanks for the heads up @jasonwebster; looking into a fix that supports both.

jasonwebster:
No problem! I've been thinking about it too.

I think the crux of the issue is that the assumptions of the tests added in this PR are wrong. Or, at least they're inconsistent with other gateways. Looking at our [base gateway](https://github.com/jasonwebster/active_merchant/blob/07e28c4936c2ae46f0470d878c19f36564c3df6e/test/unit/gateways/gateway_test.rb#L83-L91) and [Stripe tests](https://github.com/jasonwebster/active_merchant/blob/07e28c4936c2ae46f0470d878c19f36564c3df6e/test/unit/gateways/stripe_test.rb#L431-L440) for example, which is another gateway with a `:cents` money format, if you want to charge ¥100, you actually have to pass in `10000`--almost as if the non-fractional currency _was_ fractional. I guess that's the external contract of Active Merchant--all transaction amounts accepted by the public API methods are expected to be in cents, period.

jasonwebster:
Opened https://github.com/activemerchant/active_merchant/pull/2267. I'd appreciate confirmation that my interpretation of the public contract for non-fractional amounts is correct.

PR #2090, opened 2016-05-04
Updated transactionProduct to use invoice number instead of descripti…

merchantwarrior:
Updated transactionProduct to use invoice number instead of description as this will be more relevant to the merchant

@ntalbott @duff for review

### DIFF
--- a/test/unit/gateways/merchant_warrior_test.rb
+++ b/test/unit/gateways/merchant_warrior_test.rb
@@ -90,11 +90,11 @@ class MerchantWarriorTest < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
-  def test_description_truncated
+  def test_orderid_truncated
     stub_comms do
-      @gateway.purchase(@success_amount, @credit_card, description: "ThisIsQuiteALongDescriptionWithLotsOfChars")
+      @gateway.purchase(@success_amount, @credit_card, order_id: "ThisIsQuiteALongDescriptionWithLotsOfChars")
     end.check_request do |endpoint, data, headers|
-      assert_match(/transactionProduct=ThisIsQuiteALongDescriptionWithLot&p/, data)
+      assert_match(/transactionProduct=ThisIsQuiteALongDescriptionWithLot&/, data)
     end.respond_with(successful_purchase_response)
   end
 
--- a/lib/active_merchant/billing/gateways/merchant_warrior.rb
+++ b/lib/active_merchant/billing/gateways/merchant_warrior.rb
@@ -12,9 +12,9 @@ module ActiveMerchant #:nodoc:
 
       self.supported_countries = ['AU']
       self.supported_cardtypes = [:visa, :master, :american_express,
-                                  :diners_club, :discover]
-      self.homepage_url = 'http://www.merchantwarrior.com/'
-      self.display_name = 'MerchantWarrior'
+                                  :diners_club, :discover, :jcb]
+      self.homepage_url = 'https://www.merchantwarrior.com/'
+      self.display_name = 'Merchant Warrior'
 
       self.money_format = :dollars
       self.default_currency = 'AUD'
@@ -27,7 +27,7 @@ module ActiveMerchant #:nodoc:
       def authorize(money, payment_method, options = {})
         post = {}
         add_amount(post, money, options)
-        add_product(post, options)
+        add_order_id(post, options)
         add_address(post, options)
         add_payment_method(post, payment_method)
         commit('processAuth', post)
@@ -36,7 +36,7 @@ module ActiveMerchant #:nodoc:
       def purchase(money, payment_method, options = {})
         post = {}
         add_amount(post, money, options)
-        add_product(post, options)
+        add_order_id(post, options)
         add_address(post, options)
         add_payment_method(post, payment_method)
         commit('processCard', post)
@@ -83,10 +83,13 @@ module ActiveMerchant #:nodoc:
         post['customerCity'] = address[:city]
         post['customerAddress'] = address[:address1]
         post['customerPostCode'] = address[:zip]
+		post['customerIP'] = address[:ip]
+		post['customerPhone'] = address[:phone]
+		post['customerEmail'] = address[:email]
       end
 
-      def add_product(post, options)
-        post['transactionProduct'] = truncate(options[:description], 34)
+      def add_order_id(post, options)
+        post['transactionProduct'] = truncate(options[:order_id], 34) || SecureRandom.hex(15)
       end
 
       def add_payment_method(post, payment_method)


andrewpaliga:
👍

test/unit/gateways/merchant_warrior_test.rb
```diff
@@ -90,11 +90,11 @@ def test_scrub_name
     end.respond_with(successful_purchase_response)
   end
 
-  def test_description_truncated
```
duff: The goal of this test was to verify that we're actually truncating a really long value.

duff:
I see this description in the Merchant Warrior docs for the `transactionProduct` field:

mwe_api_pdf

That seems to correspond to the `description` field, no?

There are some "standard" options many gateways use including `description` and `invoice`:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L28-L33

At Spreedly, we use the `description` option, but we don't currently use the `invoice` option.  @andrewpaliga, does Shopify use the `invoice` field and not the `description` field?

andrewpaliga:
@duff I've spoken to one of our devs and we use `description` too in fact. @merchantwarrior have you verified if Shopify is sending the info you are looking for in the `invoice` field?

merchantwarrior:
@duff This is true. We keep this field open for either an either a product id or description. What we found is that the majority of our merchants used this particular field as a reference id and/or invoice id and left the description side of things to their shopping cart or software provider.

@andrewpaliga In our previous check with the test store we have setup (and the one Shopify allowed us to test) we found that the description was not being passed to the gateway and as such the transactions were being rejected.

@duff @andrewpaliga The use case is that a merchant will login to our administration interface and search for a transaction, they will then see the transaction_product field and use this to match a transaction in shopify. So what we would need is some form of 'order_id' and I note that this is available in ActiveMerchant. Would we be able to use this in this implementation or does the 'description' used by Shopify contain an order id of some sort?

duff:
I see this in the MerchantWarrior docs:

mwe_api_pdf


Perhaps the `order_id` attribute of active merchant could be mapped to `transactionReferenceID`  in the Merchant Warrior adapter.

merchantwarrior:
@duff Because the transactionReferenceID is used as a unique identifier to also stop duplicate transactions this may not work. Eg. order_id of a declined transaction will be the same order_id for a second attempt at the transaction which may then be approved (however the transactionReferenceID would no longer be unique and be rejected). 

It's possible to place the order_id in a custom field e.g. custom1 etc. Is the main suggestion here to leave the tranasctionProduct as the 'description'? Not too sure if our merchants will make use of the Shopify 'description' (unless it is a unique reference number to the transaction). However if this is the recommended approach we're happy to abide by this and just place an order_id or something similar in the custom1 field. Just wondering what the best approach would be. Thanks for your help.

merchantwarrior:
@duff @andrewpaliga  The reason we were looking at potentially changing the description to something else is that we were receiving the following when testing the current implementation of the gateway with Shopify:

screen shot 2016-05-06 at 1 00 45 am

merchantwarrior:
@duff Have left the description alone as suggested and placed the order_id in a custom field. 

@andrewpaliga Could you please double check with your developers what data is sent as 'description'. It looks like the truncate that was added could possibly be failing and because of that shopify is sending through a blank transactionProduct through to our gateway which is then being rejected and transactions cannot be processed successfully.

andrewpaliga:
@merchantwarrior it looks like `nil` is the value of `options["description"]` for Shopify transactions. This would explain the error you're seeing. I'm not sure in what circumstance `options["descriptions"]` in not `nil`. @aprofeit @ThereExistsX ?

duff:
Looks like the tests failed too, btw.

Thanks for the help on this @merchantwarrior!

merchantwarrior:
@andrewpaliga Thanks for chasing this up. 

@duff Based on other payment providers in the activemerchant project (have only checked AU ones so far) it looks like 'order_id' is more commonly used than the 'description'. This is what we would prefer to be seeing in the transactionProduct field as well. This would also resolve the issue that is currently stopping the Shopify implementation from working. 

Our concern is that there are a number of customers of ours using Spreedly and we would not want this change to break their implementation. 

Our proposed change would be: 

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/merchant_warrior.rb#L89

Changing the above line to: 

`post['transactionProduct'] = options[:order_id] || SecureRandom.hex(15)`

This would give us the correct value for the field, fix the issue currently with the Shopify implementation and also ensure a default value is passed across in case the 'order_id' was not available.

Your thoughts? Just want to make sure what we're implementing is not going to be breaking anything or rocking the boat anywhere!

In regards to the test failing - not too sure on this one. The test file was actually reverted to your original test file so the fact it's failing now and not before has me a little lost!?

duff:
I'm gonna leave this one up to the Shopify folks since it's sounding like it's somewhat specific to the integration there.  With regard to the failing test, make sure you're truncating the value with the `truncate` method.

andrewpaliga:
@merchantwarrior `options[:order_id]` will always be populated with something similar to `c12345678` representing the checkout id. I'm not certain it make sense to set `post['transactionProduct'] to this value but I'll let you be the judge of that.

merchantwarrior:
@andrewpaliga Thanks. I guess a better question would be - what value would a customer use to relate back to an order inside Shopify? Eg. If they login to their gateway (merchantwarrior) and saw the 'order_id' attached to the transaction, would they then be able to use this value in an 'order search' inside Shopify to identify the order associated with the payment?

andrewpaliga:
@merchantwarrior yes they'd be able to search with the value and the corresponding order would be retrieved.

merchantwarrior:
@duff @andrewpaliga Thanks very much. Changes made for review. If all is good - it would be great to get this into our test shopify store for testing of transactions.

duff:
@merchantwarrior You'll want to truncate the order_id.  And squash it down to one commit.

merchantwarrior:
@duff Cheers. order_id has been truncated and all commits squashed into one commit. Anything else let me know :)

merchantwarrior:
@duff Just wondering if there is anything that we may have missed, or anything we can do to have this merged? We're just waiting on this new version being used within Shopify so that our clients who are waiting can take advantage of it.

duff:
Only thing I noticed was that the test was removed.  I think that test provides some value.

I'm leaving the merging of this one to Shopify since it's specific to their integration.

andrewpaliga:
@merchantwarrior once you add the test back we can merge this PR. Can you also verify whether the remote tests are passing for your integration?

merchantwarrior:
@andrewpaliga Can confirm the remote tests are functioning correctly as we have test transactions being created successfully via our modified activemerchant module.

The truncate test has been placed back. We look forward to this PR being merged so that we can test our shopify implementation and make this available to our merchants. 

Thanks!

PR #2092, opened 2016-05-05
Removing invalid country codes AN and KV

apdunston:
Removing codes AN and KV from the countries list and SecurionPayGateway as they are not part of the ActiveMerchant-selected standard ISO-3166-1.

AN is a deleted code from a different standard ISO 3166-2. It doesn't refer to a country but to the country-subdivision Netherlands Antilles.

KV appears to refer to Kosovo. Kosovo is a disputed territory not recognized by ISO 3166-1. The code XK is being used by several organizations as a temporary country code for Kosovo, but it is not an accepted standard.

Because these codes aren't part of ISO-3166-1 and because they are only used by SecurionPay, I'm just removing them.


apdunston:
@rwdaigle, please review and merge.

PR #2094, opened 2016-05-09
visanet peru gateway response message

shasum:
@duff 

- meaningful gateway message lookup in multiple response fields
- currency conversion from cents to $$.cc before passing to gateway

### DIFF
--- a/lib/active_merchant/billing/gateways/visanet_peru.rb
+++ b/lib/active_merchant/billing/gateways/visanet_peru.rb
@@ -1,6 +1,7 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class VisanetPeruGateway < Gateway
+      include Empty
       self.display_name = "VisaNet Peru Gateway"
       self.homepage_url = "http://www.visanet.com.pe"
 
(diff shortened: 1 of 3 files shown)


duff:
@shasum I'm thinking it would be great to split up these into 2 PR's - one for the error messages and one for the currency fix.  The error messages part is looking good.  

I'm mulling some on the currency part.  Does their docs say anything about it?  

It sounds like we might want to do something that we do for a number of gateways specifying a different `money_format`.  Example:

https://github.com/duff/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L13

Then the `amount` method uses that:

https://github.com/duff/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L234

shasum:
@duff 

crazy me. I stared straight into the `amount()` method while I was doing this piece and yet failed to realize the `:dollars` money format does exactly what I wanted 😄 . thanks for pointing it out. I'll refactor the currency part and submit separate PRs.

duff:
Excellent.  Thanks @shasum !

shasum:
@duff this PR now fixes only the message part

duff:
@shasum Looking good.  2 little things.  We're trying to lessen our dependence on Active Support by using the `blank?` method less.  Eventually we'd like to eliminate the dependency altogether.  Instead of `blank?`, we've been using `empty?`.  Examples:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L6

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L457

Also, it would be great if you could squash it down to one commit.

shasum:
@duff emptied and squashed

PR #2095, opened 2016-05-09
Added BIN2 regex range to Mastercard brand detection

ryanbalsdon:
Small snippet from the Mastercard Announcement for context:
> *Supporting Economic Growth with a New BIN Range Expansion*
> To help ensure that MasterCard and its customers are best positioned to enable the next generations of payments, MasterCard is expanding the Bank Identification Number (BIN) range by introducing a new range of 2 series BINs (222100–272099) that will be processed in the same manner as the MasterCard 51–55 series BINs are processed today.

A test with the PAN used in the test addition is now part of the M/Chip certification suite (and is currently failing).

Review please @lyverovski and @davidseal or @bizla 
Ping @girasquid @aprofeit and @rwdaigle

### DIFF
--- a/test/unit/credit_card_test.rb
+++ b/test/unit/credit_card_test.rb
@@ -193,6 +193,7 @@ class CreditCardTest < Test::Unit::TestCase
   def test_should_correctly_identify_card_brand
     assert_equal 'visa',             CreditCard.brand?('4242424242424242')
     assert_equal 'american_express', CreditCard.brand?('341111111111111')
+    (222100..272099).each {|bin| assert_equal 'master', CreditCard.brand?(bin.to_s + '1111111111'), "Failed with BIN #{bin}"}
     assert_nil CreditCard.brand?('')
   end
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -43,6 +43,7 @@
 * BlueSnap: Add gateway [duff]
 * VisaNet Peru: Select the most meaningful gateway error message [shasum]
 * SecurionPay: Update country list [duff]
+* Support for BIN 2 MasterCard brand detection [rbalsdon]
 
 
 == Version 1.58.0 (March 1, 2016)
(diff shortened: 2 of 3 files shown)


aprofeit:
Wow that regex is crazy. Hopefully our tests are sufficient because I can't think of how to verify this myself 👍

rwdaigle:
@ryanbalsdon do you have a link to the official MasterCard announcement (for posterity sake)?

ryanbalsdon:
I'm having trouble finding the official announcement on the web.
Here are a couple credible-looking blogs though:
https://www.linkedin.com/pulse/mastercards-new-2-bin-scott-tivey
https://www.forte.net/blog/mastercard-bin-range-coming/

ryanbalsdon:
@aprofeit: I expanded on the test to try every single possible bin because the regex is a bit too ridiculous to trust. It adds about 1.1s (on my machine) to the test suite but did find a bug so is probably worth keeping.

test/unit/credit_card_test.rb
```diff
@@ -193,6 +193,7 @@ def test_should_display_number
   def test_should_correctly_identify_card_brand
     assert_equal 'visa',             CreditCard.brand?('4242424242424242')
     assert_equal 'american_express', CreditCard.brand?('341111111111111')
+    (222100..272099).each {|bin| assert_equal 'master', CreditCard.brand?(bin.to_s + '0089600203'), "Failed with BIN #{bin}"}
```
lyverovski: If we don't need it to pass Luhn here (do we?), then why not just fill with '1111111111'?

ryanbalsdon: The majority of these are going to fail Luhn already so it can't be important. I didn't put any thought into which number to use so just pasted in the one from the certification test. I'll change it to all-ones.

lyverovski:
Other than the comment on replacing with 1's, LGTM. Well done on the exhaustive tests to check the range.

davidseal:
LGTM 💯

PR #2096, opened 2016-05-09
NMI: Pass value of recurring option if present

curiousepic:
@duff to confirm and merge.


curiousepic:
After consideration this is likely not needed, closing.

PR #2098, opened 2016-05-10
Fix regression for storing stripe cards from tokens.

Senjai:
Previously, we could pass in string tokens we received from the frontend
via Stripe.js into the store method. In fact, there is explicit message
handling in `add_creditcard` to handle the case where this is a string. This worked in version 1.48, but does not work in the upgrade to version 1.58. This case is used in https://github.com/solidusio/solidus for the gateway extension.

When `StripePaymentToken` and the `card_brand` method were added, two
things happened:

Firstly, card_brand expects the argument or "payment" to respond to `type` or
`brand` which by default breaks the existing functionality for strings
and raises a `NoMethodError` exception.

Secondly, while a `StripePaymentToken` construct has been provided, it
can't be used as the argument here. While it passes the card_brand
check without throwing an exception the logic falls through to calling
`add_creditcard` where the payment argument is a `StripePaymentToken`.

In this scenario, if it were to work, the argument would have to be
`payment.payment_data` as that's what contains the actual token for
add_creditcard to work.

Additionally, the payment token construct is defined in the private
interace which leads me to believe it should not be directly used.

To ensure backwards compatability, I added the support for string tokens
back, but I wouldn't mind passing an actual object in the future.
However, that sort of change should be reserved for a major release, or
kept backwards compatible.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -167,7 +167,7 @@ module ActiveMerchant #:nodoc:
         params = {}
         post = {}
 
-        if card_brand(payment) == "check"
+        if !payment.is_a?(String) && card_brand(payment) == "check"
           bank_token_response = tokenize_bank_account(payment)
           if bank_token_response.success?
             params = { source: bank_token_response.params["token"]["id"] }
--- a/test/remote/gateways/remote_stripe_test.rb
+++ b/test/remote/gateways/remote_stripe_test.rb
@@ -306,6 +306,24 @@ class RemoteStripeTest < Test::Unit::TestCase
     assert_match "The customer's bank account must be verified", purchase.message
   end
 
+  def test_store_card_from_token
+    test_card = {
+      card: {
+        number: "4242424242424242",
+        exp_month: "12",
+        exp_year: Date.current.year + 5,
+        cvv: "123"
+      }
+    }
+
+    tokenized_card = @gateway.send(:api_request, :post, "tokens?#{test_card.to_query}")
+    assert_equal(tokenized_card["error"], nil)
+    assert(tokenized_card["id"])
+
+    store = @gateway.store(tokenized_card["id"])
+    assert_success(store)
+  end
+
   def test_successful_purchase_from_stored_and_verified_bank_account
     store = @gateway.store(@check)
     assert_success store
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -37,6 +37,15 @@ class StripeTest < Test::Unit::TestCase
     assert response.test?
   end
 
+  def test_successful_store_with_string_card_token
+    @gateway.expects(:ssl_request).returns(successful_new_customer_response)
+    @gateway.expects(:add_creditcard).with({}, "test_token", {})
+
+    assert response = @gateway.store("test_token")
+    assert_instance_of Response, response
+    assert_success response
+  end
+
   def test_successful_new_customer_with_apple_pay_payment_token
     @gateway.expects(:ssl_request).returns(successful_new_customer_response)
     @gateway.expects(:tokenize_apple_pay_token).returns(Response.new(true, nil, token: successful_apple_pay_token_exchange))


brendandeere:
LGTM

forkata:
:+1:

Senjai:
@timbeiko Just poking on this one. Is everything okay? I understand the awkwardness here with the fix but there are a few solidus stores that are eager to get past our locked down active merchant, if there is anything I can do to help you speed this along I am available.

timbeiko:
@Senjai I've been looking into this with the team and we're wondering if there is a way to implement the logic to defend against payment being a string inside the `card_brand` method itself.  And I was trying to see if there was a case where `payement` will be a String and should end up after the first `if`.

Senjai:
@timbeiko I don't know if I follow. The card_brand method expects a token object (as far as I can tell) or an `ActiveMerchant::Billing::Creditcard`. However, using stripejs for token generation, all we have is the token string and it wouldn't make sense to check the brand of a token. Does that help at all?

jordan-brough:
@timbeiko it would be great to get the current PR, or something close to it, merged in since ActiveMerchant v1.58.0 broke stripejs integration as far as I can tell.  Perhaps we could merge the current PR and then work on improving things after that?

bizla:
The proper fix for this problem would be to construct a `StripePaymentToken` and have `store` handle it appropriately rather than guarding explicitly that an arbitrary `String` doesn't pass through. 

The fix suggested in this PR is to ignore a `String` payment in the initial case, but the reasoning for that isn't obvious. Guarding against a primitive type at that level -- where we're still deciding on what how to express the thing we want to store -- is coding a little too close to the implementation.

Ultimately what you're actually after is having `store` handle the condition:
```ruby
if payment.is_a?(StripePaymentToken)
```
whereby, as you said, you want 
```ruby
add_creditcard(payment.payment_data)
```
to be called. That's what the fix should look like. Doing so makes the relationship between payment tokens within this context and the desired effect obvious. 

This means your consuming code would have to change to look something like: 
```ruby
stripe.store(StripeGateway::StripePaymentToken(payment_data: token_string_from_stripe_js))
```
which has obvious semantic benefit and fixes the interface to `store` properly.

timbeiko:
cc @jordan-brough @Senjai

jordan-brough:
@timbeiko @bizla - I absolutely agree that the code in this area could be better.  But wouldn't it be best to fix the regression right away (e.g. @Senjai's commit) and then take a separate pass to move things forward?  Right now the code is just broken -- you can't pass a string anymore and you also can't pass a `StripePaymentToken`.

Since [the initial commit of Stripe support](https://github.com/activemerchant/active_merchant/commit/ca39e94#diff-8299be227de5ba32c34c30c42a828c9cR110) `StripeGateway` has supported passing in a credit card token as a string.  And it has continued to do so until [this commit](https://github.com/activemerchant/active_merchant/commit/f2a7ea3) which broke it unintentionally.  Even in some [remaining parts of the current code](https://github.com/activemerchant/active_merchant/blob/v1.59.0/lib/active_merchant/billing/gateways/stripe.rb#L372) passing in a String is supported.  @bizla [added `StripePaymentToken`](https://github.com/activemerchant/active_merchant/commit/a56712f53ac685ebdcfa6886b513dafb6298cb9e#diff-8299be227de5ba32c34c30c42a828c9cL314) but only added it halfway -- supporting it in some places but not others (e.g. not supporting it in `StripeGateway#add_credit_card`).

If someone has an alternative backward-compatible fix that's great!  Otherwise could we fix the breakage and then work on improving things after that?

jordan-brough:
It seems like a decent backward-compatible fix could be:
- Update `StripeGateway#store` to convert `payment` into a `StripePaymentToken` if it's a string (possibly with a deprecation warning).
- Update `StripeGateway#add_creditcard` to handle `StripePaymentToken`s.
But again, if it were me I'd say fix the broken code first.

bizla:
> @bizla added StripePaymentToken but only added it halfway -- supporting it in some places but not others (e.g. not supporting it in StripeGateway#add_credit_card).

That's correct. `StripeGateway#add_credit_card` is not a public interface. It's a private helper method meant for specific consumers. It doesn't declare support for anything nor is it reasonable to think that it does.

[Look at the unit tests](https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/stripe_test.rb#L323-L331). While passing a string into `store` may have worked, there are no tests that suggest that this was an intended input with a guaranteed behaviour. 
That doesn't mean it has to stay that way though. :wink: The interface to `purchase` does set a string-as-a-payment precedent. So, your code (and tests!) should coerce that expectation of `store`. Instead of... 
```ruby 
if !payment.is_a?(String) && card_brand(payment) == "check"
```
...which is a logically unreasonable, since it's not about a `String` input specifically, it's about dodging the exception-throwing `card_brand` call, [you could address the fear and fix `card_brand`](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L146-L149). Have it return `nil` if it doesn't additionally respond to `type`, then you're no longer worried about passing garbage into it. It's more reasonable to pass a bad input into a pure function and get nothing rather than defensively avoiding input without a logical basis for the reader of the code. 

Then, as you desire, the original fall-through to `add_creditcard` happens, the code reads logically (you don't have to reason about why you have to worry explicitly about strings ever) and finally, your _new_ tests which explicitly expect a string as an input pass, and everyone goes home happy.

/cc @timbeiko

jordan-brough:
> StripeGateway#add_credit_card is not a public interface. It's a private helper method meant for specific consumers

I mentioned `#add_credit_card` because that is what `#store` uses, and `#store` is a public interface.  `StripePaymentToken` was added half-baked.  There simply is no other way to use `#store` with a stripe token except as a string, and now even that is broken.

> Look at the unit tests

There are no unit tests for passing in a string.  This is part of the problem.  There is plenty of code supporting it and it has always been the only way to handle stripe tokens in `#store`.  Whoever added the code originally did not add enough specs.  @Senjai is helping to fix that.

> you could address the fear and fix card_brand. Have it return nil if it doesn't additionally respond to type

I would think it would be better to have `card_brand`, which is shared by many different gateways, complain about garbage input than return nil.

Senjai:
@bizla @timbeiko I understand. I get that we want a better interface here but that's outside the scope of this PR.

>The proper fix for this problem would be to construct a StripePaymentToken

I agree, but we can't without a major version bump 

>Look at the unit tests. While passing a string into store may have worked, there are no tests that suggest that this was an intended input with a guaranteed behaviour. 

By this logic the `StripePaymentToken` isn't the right choice either, so what is? The commit that broke this interface is f2a7ea3ed298038529acce7552b733ade8f3bc90 when we _restricted_ the type of the argument passed in.

The lack of tests is unfortunate but that doesn't mean something does not _support_ a specific use case, it just means that tests weren't written for it and that is an important distinction. The person who originally added the stripe gateway and the store method probably intended it to accept stripe.js string tokens (otherwise nobody could use active merchant with stripe.js which is unlikely) and I'll make my case for this shortly.

The history of the store method goes _wayyyyyyy_ back to the initial commit for the gateway here: ca39e94032097c53c2848119431da32ff6da3f31. This commit was absent a commit message and didn't have any tests at all just yet.

In this original interface, up to where the funcitonality was removed in f2a7ea3ed298038529acce7552b733ade8f3bc90 there are _explicit_ code paths for string tokens, that do not make sense outside of this context.

Back then (ca39e94032097c53c2848119431da32ff6da3f31) the store method looked like this:

```ruby
      def store(creditcard, options={})
        post = {}
        add_creditcard(post, creditcard, options)

        if options[:customer]
i thought your PR was minimally invasive, included specs, had a great commit message, fixes what seems like a significant regression.  i'm somewhat perplexed by their unwillingness to fix what's broken, given that you handed all of that to them on a platt
          commit("customers/#{CGI.escape(options[:customer])}", :post, post)
        else
          commit('customers', :post, post)
        end
      end
```

creditcard is passed to add_creditcard which is implemented thusly:
```ruby
      def add_creditcard(post, creditcard, options)
        if creditcard.kind_of?(CreditCard)
          card = {}
          card[:number] = creditcard.number
          card[:exp_month] = creditcard.month
          card[:exp_year] = creditcard.year
          card[:cvc] = creditcard.verification_value if creditcard.verification_value?
          card[:name] = creditcard.name if creditcard.name
          post[:card] = card

          add_address(post, options)
        elsif creditcard.kind_of?(String)
          post[:card] = creditcard
        end
      end
```

since then `add_creditcard` has expanded in functionality but this code path _still exists and is still supported_ in master

```ruby
def add_creditcard(post, creditcard, options)
# ...
        elsif creditcard.kind_of?(String)
          if options[:track_data]
            card[:swipe_data] = options[:track_data]
          elsif creditcard.include?("|")
            customer_id, card_id = creditcard.split("|")
            card = card_id
            post[:customer] = customer_id
          else
            card = creditcard
          end
          post[:card] = card
        end
# ...
```

We can also infer that the only time stripe accepts a plain string in the card param is when that string represents a token. Additionally this logic for "track_data" would also not work because this code path would also be broken which also means a stripe token here may be incorrect as this is probably magstripe data too depending on the `options[:track_data]`

I expect the tests were written for store after the fact and a code path was missed, but it's not fair to say that this probably wasn't intended just by looking at tests alone. To make that assertion we would have to say our tests have 100% code coverage, and test every possible code path. 

What I have in this PR is the _least invasive and smallest  possible fix_ as well as solid regression tests to prevent it from happening in the future. Changing the interface to require a StripePaymentToken is a breaking change and I don't think we should do it in a minor version release.

Changing card brand (which as @jordan-brough mentioned) that is used by _all_ gateways and having it accept bad input in a system we control I think is a step back. If I don't have a "card" I should not be asking for the brand of it.

This string check is a code smell for sure, but all the following code paths are currently focused around this being a string. I put in time diagnosing this issue and I would like to unrestrict active merchant versions for people who use the solidus gateway extension. The solution is simple, and is the least invasive fix we have. 

If we want to do a larger refactoring of the private interface I'm on-board with that, but I'd like to open it up as an issue and have someone interested in contributing to the project further take it up, or I can when I have more time. Given the limited amount of free time I have for this, I'm doing what I can to help this project.

I really hope you'll accept this PR as a small iterative approach to fixing the regression so we could have a larger discussion later on how encapsulate this type of data and handle it better in the private interface and it will let our extension and others using Stripe.js (there are people I know who also hit this without our extension who use active merchant directly) stay up to date with recent versions.

@bizla If you wanted to help out with the implementation I would also accept a pull request to the branch I'm using for this PR.

bizla:
> i thought your PR was minimally invasive, included specs, had a great commit message, fixes what seems like a significant regression.  i'm somewhat perplexed by their unwillingness to fix what's broken...

I've opened an issue since this needs a proper fix given the discussion here. I agree that this is a functional regression, but it requires a properly robust fix for merging upstream. 
Keep an eye on #2130 -- let me know if you're keen on testing in your staging environment once the fix has merged.

Senjai:
@bizla @timbeiko You're kidding me right?

I maintain a few gems, solidus extensions and cancancan. I try to model what I would hope from one of the contributors to any of them.

I gave you a _very_ robust fix. It passes all tests. It adds two new tests to ensure regressions are covered. It was a _one line fix_. One line. This feels like a total slap in the face to contributors who decide to fix problems rather than just open issues. I've contributed to this project in the past and it feels like there was an entirely different atmosphere at the time and I'm likely not to continue.

The place to have this discussion was here https://github.com/activemerchant/active_merchant/commit/f2a7ea3ed298038529acce7552b733ade8f3bc90 two months ago when card_brand was being applied to the argument. It was missed in code review which is totally fine and happens sometimes.

This would have made it feature equivalent to the code that existed before that commit which is only two months old or so. 

If you want to spend time and refactor things, go for it. But when someone has a one liner to solve the problem now, and doesn't want to spend hours fixing a private interface that has existed for 3 years that's okay. The fix has no downsides whatsoever, it breaks no tests and it is blatantly obvious and self contained behavior. 

It is _extremely_ simple to get the fix in now, and fix it however you like later.

bizla:
You're of course not restricted from forking the project as you require for your needs in the meantime. You've illustrated something that needs a little more attention that was initially assumed and for deployment onto Shopify and its scale, I have to be strict about a proper forward fix. That said, the next release of ActiveMerchant will have this issue resolved, and you should be able to submit the `store` calls as you expect.

Senjai:
@bizla How is this not _scale_? It works fine. The only argument against this is that you dont like all the other code around it that created the issue :(

You even wrote https://github.com/activemerchant/active_merchant/pull/1790/files how did that get through? It seemed like you fixed a bug that needed to be fixed but I would argue that it's not the best solution.

Fix first, clean later. There are no downsides, and I'm not going to fork a open source project to a more specific active merchant and keep it up to date with this one that would be silly. This can suffice for a patch version with a proper implementation in the minor.

PR #2099, opened 2016-05-10
Update Elavon Integration: Remove customer code, default sales taxes to zero

chobson:
The Elavon Customer Code is being set as the first ten characters of the customer email. This is incorrect; we do not have a field in Shopify for Customer Code, so generating an incorrect one based on email address is interfering with merchant transactions (specifically, refunds fail for invalid customer code).

As well, refunds on tax-exempt items fail if the tax value is nil; we need to default it to zero instead.

Based on a ZD issue; ping me for context.

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -254,7 +254,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_salestax(form, options)
-        form[:salestax] = options[:tax] if options[:tax].present?
+        form[:salestax] =(options[:tax].present? options[:tax] : 0)
       end
 
       def add_address(form, options)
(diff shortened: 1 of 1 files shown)


chobson:
@girasquid I believe this is online now, would you be able to take a peek?

chobson:
Sorry, Luke's on the Capital team. @aprofeit ?

aprofeit:
Just because we don't use `customer_code`, I don't think that means we should remove it for everyone. Other people other than Shopify use this gem.

If the tax thing is unrelated, it should be in its own PR. Do you have documentation explaining how this field works? I'm worried it may do other things we're not aware of.

chobson:
👍 

Hearing from Elavon on this subject, the issues that led to this PR were caused by a defect on their end which they are now looking into.

PR #2100, opened 2016-05-10
Authorize.net: Pass value of Recurring option

curiousepic:
@duff to review and merge


curiousepic:
After consideration this is likely not needed, closing.

PR #2103, opened 2016-05-12
Include method to check whether or not a payment method is a credit card

ThereExistsX:
TransFirst's `is_a` check here was pretty brittle, and was forcing the use of Active Merchant's credit card. This PR decouples Active Merchant from `ActiveMerchant::Billing::CreditCard`

**Attn:**
@aprofeit @lyverovski


lib/active_merchant/billing/network_tokenization_credit_card.rb
```diff
@@ -27,6 +27,10 @@ def source
         end
       end
 
+      def credit_card?
+        true
```
aprofeit: I _think_ this should be `false` as the old call of `is_a?(CreditCard)` would have failed if the class was a `NetworkTokenizationCreditCard`.

aprofeit:
@ntalbott do you have opinions on this? From what I can tell, the original intent of `CreditCardMethods` was to allow consumer projects to define their own `CreditCard` model and pass it in place of a `ActiveRecord::Billing::CreditCard`.

Conditional checks directly against the class like `is_a?(CreditCard)` make this difficult, if not impossible.

Should we work towards removing the `is_a?` checks from AM, or should we give up on the idea of passing custom `CreditCard` models and force passing an `ActiveRecord::Billing::CreditCard`?

ThereExistsX:
^^ @duff @rwdaigle

ntalbott:
@ThereExistsX @aprofeit Spreedly only uses ActiveMerchant's credit card, but I'm fine avoiding is_a tests and preferring respond_to or other tests. In general we've done that in the past.

lib/active_merchant/billing/gateways/trans_first.rb
```diff
@@ -40,10 +40,10 @@ def purchase(money, payment, options = {})
         add_amount(post, money)
         add_payment(post, payment)
         add_address(post, options)
-        add_invoice(post, options) if payment.is_a?(CreditCard)
+        add_invoice(post, options) if payment.credit_card?
         add_pair(post, :RefID, options[:order_id], required: true)
 
-        commit((payment.is_a?(Check) ? :purchase_echeck : :purchase), post)
```
aprofeit: Let's just leave this how it was. If we define `check?` methods down the road this can be changed, but currently it's unrelated to the other changes.

lib/active_merchant/billing/check.rb
```diff
@@ -50,6 +50,9 @@ def type
         'check'
       end
 
+      def credit_card?
+	false
```
aprofeit: Indentation.

lib/active_merchant/billing/gateways/trans_first.rb
ThereExistsX: I'm fine with that -- @lyverovski , thoughts?

lyverovski: Yes - I am OK with it for now. At some point however, I would like the method of detecting Check vs. Credit Card to follow the same logic.

Issue #2014, opened 2016-02-05
activemerchant contains also activeshipping?


megetron:
Hi there,
I am using application which use active_shipping gem. I would like to add the support of the active_merchant for a gateway system. 

should I remove from my application the active_shipping gem and use it through the active_merchant gem?

timbeiko:
Accidentally closed instead of https://github.com/activemerchant/active_merchant/commit/b0eb93a166273f6c0e8ebdb5a473f514c05cd65c . Re-opened.

Amritpal363:
Good morning  🙏🌹india punjabi city code 143604

![Image](https://github.com/user-attachments/assets/ffe19dc9-5598-4f1e-a857-bcdf03093bc1)

PR #2104, opened 2016-05-13
Set error code for AuthorizeNetCimGateway response

ka8725:


timbeiko:
Just reviewed this (and got @ThereExistsX to have a quick look). LGTM. 

ping @aprofeit @lyverovski

test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -632,6 +632,42 @@ def test_should_create_duplicate_customer_profile_transactions_with_duplicate_wi
...
+        },
+        :amount => @amount
+      }
+    }
+
+    assert response = @gateway.create_customer_profile_transaction(customer_profile_transaction)
+    assert_success response
+    assert_equal "Successful.", response.message
+
+    sleep(5)
```
aprofeit: Why is this necessary? Timing things like this is bound to run into errors eventually, can this not be done a different way?

lib/active_merchant/billing/gateways/authorize_net_cim.rb
```diff
@@ -864,10 +864,12 @@ def commit(action, request)
...
         transaction_id = response_params['direct_response']['transaction_id'] if response_params['direct_response']
 
-        Response.new(success, message, response_params,
-          :test => test_mode,
-          :authorization => transaction_id || response_params['customer_profile_id'] || (response_params['profile'] ? response_params['profile']['customer_profile_id'] : nil)
-        )
+        response_options = {}
+        response_options[:test] = test_mode
+        response_options[:authorization] = transaction_id || response_params['customer_profile_id'] || (response_params['profile'] ? response_params['profile']['customer_profile_id'] : nil)
+        response_options[:error_code] = response_params['messages']['message']['code'] unless success
```
aprofeit: Does this data structure _always_ exist? I'm worried if `messages` or `message` isn't present, the code will error out. Is there documentation to support this structure always existing?

test/remote/gateways/remote_authorize_net_cim_test.rb
ka8725: I don't know, I just followed this [example](https://github.com/mezuka/active_merchant/blob/722ff73564385d71cf8326abf41ce0d298d6e7b4/test/remote/gateways/remote_authorize_net_cim_test.rb#L630).

aprofeit: Does it work without it? Although remote tests are generally not the fastest things, I'm just concerned about _forcing_ it to take longer.

timbeiko: @aprofeit just tested and it doesn't, but it does work with `sleep(2)`, which might be worth changing.

aprofeit: Ideally we could have a mechanism that retried until the remote server had done whatever we're waiting for, but that's probably overcomplicating things. Let's just leave the `sleep(5)` I guess. Worried if we do actually need to wait for something, `sleep(2)` may work now, but during a small network blip it would be flakey.

timbeiko: :+1:

lib/active_merchant/billing/gateways/authorize_net_cim.rb
timbeiko: @aprofeit best I found in the docs is [here](http://developer.authorize.net/api/reference/index.html#payment-transactions-charge-a-credit-card).

It mentions them as part of the response (see "Response" tab of "Charge a Credit Card"), yet doesn't state that it will always be there. 

<b>But</b>  for another param, `cavvResultCode` it says "Blank or not present = CAVV not validated.". 

So you'd think they would mention it if this was sometimes blank or not present.

ka8725: @timbeiko I made changes for this [request](http://developer.authorize.net/api/reference/index.html#payment-transactions-charge-a-customer-profile).

Look there the description for `messages` and `message` param: https://db.tt/qBsHeLWu
It says "one or more elements".

Further there is a Ruby code example there from official [library](https://github.com/AuthorizeNet/sample-code-ruby/blob/master/PaymentTransactions/charge-customer-profile.rb#L25-L32). As you see they rely on the same construction to get the error code.

timbeiko: @ka8725 my bad, thanks for the link. It seems to be how they demo its usage, so LGTM.

@aprofeit what do you think?

aprofeit: 👍

aprofeit:
👍

timbeiko:
@lyverovski are you ok with this?

lyverovski:
👍

timbeiko:
Accidentally closed https://github.com/activemerchant/active_merchant/issues/2014 instead. Manually closing and re-opened https://github.com/activemerchant/active_merchant/issues/2014 .

PR #2105, opened 2016-05-13
Adds basic tokenisation support for banwire

bruno:
This commit adds basic support for tokenisation via the `recurrentes` API http://docs.banwire.com/recurrentes to Banwire.

The one off payments for tokens seems to require an account, I have not added tests for that yet. I still need to work out with the Banwire technical team how the one-off payments with a token work. When testing, I get

``` JSON
{"id_transaction":"1631604","error":{"code":"83","message":"no_accounts","detail":""}}
```

The docs don't really have much information for that particular API call.

### DIFF
--- a/lib/active_merchant/billing/gateways/banwire.rb
+++ b/lib/active_merchant/billing/gateways/banwire.rb
@@ -119,18 +119,18 @@ module ActiveMerchant #:nodoc:
           response = json_error(raw_response)
         end
 
-        Response.new(success?(action, response),
+        Response.new(success?(response, action),
                      response["message"],
                      response,
                      :test => test?,
-                     :authorization => authorization_from_response(action, response))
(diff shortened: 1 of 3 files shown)


bruno:
Also, an important note ... it seems that when testing tokenisation (`store`) you can only use a PAN (cc number) once ... and these numbers don't seem to be able to be removed ... so it's likely that, since we're using a dev account, lots of the test PANs have already been used (for when running the remote tests)

bruno:
After extensive conversations with the Banwire team, it appears that we can avoid the exceptions when attempting to tokenise the same card twice. I will be adding support for that shortly.

davidsantoso:
Hi - I'm grooming old pull requests and if this is still something you're interested in, please:

1. Reopen the PR
2. Rebase on `master`
3. Squash all commits
4. Tag me for review

Thanks!

PR #2113, opened 2016-05-24
PayU Latam: Add new gateway

shasum:
@duff only purchase and authorize works in sandbox

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -1,0 +1,400 @@
+require 'digest/md5'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PayuLatamGateway < Gateway
+      self.display_name = "PayU Latam"
+      self.homepage_url = "http://www.payulatam.com"
+
+      self.test_url = "https://sandbox.api.payulatam.com/payments-api/4.0/service.cgi"
+      self.live_url = "https://api.payulatam.com/payments-api/4.0/service.cgi"
+
+      self.supported_countries = ["AR", "BR", "CO", "MX", "PA", "PE"]
+      self.default_currency = "USD"
+      self.money_format = :dollars
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club]
+
+      BRAND_MAP = {
+        "visa" => "VISA",
+        "master" => "MASTERCARD",
+        "american_express" => "AMEX",
+        "diners_club" => "DINERS"
+      }
+
+      MINIMUMS = {
+        "ARS" => 1700,
+        "BRL" => 600,
+        "MXN" => 3900,
+        "PEN" => 500
+      }
+
+      def initialize(options={})
+        requires!(options, :merchant_id, :account_id, :api_login, :api_key)
+        super
+      end
+
+      def purchase(amount, payment_method, options={})
+        post = {}
+        auth_or_sale(post, 'AUTHORIZATION_AND_CAPTURE', amount, payment_method, options)
+        commit('purchase', post)
+      end
+
+      def authorize(amount, payment_method, options={})
+        post = {}
+        auth_or_sale(post, 'AUTHORIZATION', amount, payment_method, options)
+        commit('auth', post)
+      end
+
+      def capture(authorization, options={})
+        post = {}
+
+        add_credentials(post, 'SUBMIT_TRANSACTION')
+        add_transaction_type(post, 'CAPTURE')
+        add_reference(post, authorization)
+
+        commit('capture', post)
+      end
+
+      def void(authorization, options={})
+        post = {}
+
+        add_credentials(post, 'SUBMIT_TRANSACTION')
+        add_transaction_type(post, 'VOID')
+        add_reference(post, authorization)
+
+        commit('void', post)
+      end
+
+      def refund(authorization, options={})
+        post = {}
+
+        add_credentials(post, 'SUBMIT_TRANSACTION')
+        add_transaction_type(post, 'REFUND')
+        add_reference(post, authorization)
+
+        commit('refund', post)
+      end
+
+      def verify(credit_card, options={})
+        minimum = MINIMUMS[options[:currency].upcase] if options[:currency]
+        amount = minimum || 100
+
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(amount, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      def store(payment_method, options = {})
+        post = {}
+
+        add_credentials(post, 'CREATE_TOKEN')
+        add_payment_method_to_be_tokenized(post, payment_method)
+
+        commit('store', post)
+      end
+
+      def verify_credentials
+        post = {}
+        add_credentials(post, 'GET_PAYMENT_METHODS')
+        response = commit('verify_credentials', post)
+        response.success?
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((\"creditCard\\\":{\\\"number\\\":\\\")\d+), '\1[FILTERED]').
+          gsub(%r((\"securityCode\\\":\\\")\d+), '\1[FILTERED]').
+          gsub(%r((\"apiKey\\\":\\\")\w+), '\1[FILTERED]')
+      end
+
+      private
+
+      def auth_or_sale(post, transaction_type, amount, payment_method, options)
+        add_credentials(post, 'SUBMIT_TRANSACTION')
+        add_transaction_type(post, transaction_type)
+        add_order(post, options)
+        add_buyer(post, options)
+        add_invoice(post, amount, options)
+        add_signature(post)
+        add_payment_method(post, payment_method, options)
+        add_payer(post, options)
+        add_extra_parameters(post, options)
+      end
+
+      def add_credentials(post, command)
+        post[:test] = test? unless command == 'CREATE_TOKEN'
+        post[:language] = 'en'
+        post[:command] = command
+        merchant = {}
+        merchant[:apiLogin] = @options[:api_login]
+        merchant[:apiKey] = @options[:api_key]
+        post[:merchant] = merchant
+      end
+
+      def add_transaction_type(post, type)
+        transaction = {}
+        transaction[:type] = type
+        post[:transaction] = transaction
+      end
+
+      def add_order(post, options)
+        order = {}
+        order[:accountId] = @options[:account_id]
+        order[:referenceCode] = options[:order_id] || generate_unique_id
+        order[:description] = options[:description] || 'unspecified'
+        order[:language] = 'en'
+        post[:transaction][:order] = order
+      end
+
+      def add_buyer(post, options)
+        if address = options[:shipping_address]
+          buyer = {}
+          buyer[:fullName] = address[:name]
+          shipping_address = {}
+          shipping_address[:street1] = address[:address1]
+          shipping_address[:street2] = address[:address2]
+          shipping_address[:city] = address[:city]
+          shipping_address[:state] = address[:state]
+          shipping_address[:country] = address[:country]
+          shipping_address[:postalCode] = address[:zip]
+          shipping_address[:phone] = address[:phone]
+          buyer[:shippingAddress] = shipping_address
+          post[:transaction][:order][:buyer] = buyer
+        end
+      end
+
+      def add_invoice(post, money, options)
+        tx_value = {}
+        tx_value[:value] = amount(money)
+        tx_value[:currency] = options[:currency] || currency(money)
+
+        additional_values = {}
+        additional_values[:TX_VALUE] = tx_value
+
+        post[:transaction][:order][:additionalValues] = additional_values
+      end
+
+      def add_signature(post)
+        post[:transaction][:order][:signature] = signature_from(post)
+      end
+
+      def signature_from(post)
+        signature_string = [
+          @options[:api_key],
+          @options[:merchant_id],
+          post[:transaction][:order][:referenceCode],
+          post[:transaction][:order][:additionalValues][:TX_VALUE][:value],
+          post[:transaction][:order][:additionalValues][:TX_VALUE][:currency]
+        ].compact.join("~")
+
+        Digest::MD5.hexdigest(signature_string)
+      end
+
+      def add_payment_method(post, payment_method, options)
+        if payment_method.is_a?(String)
+          brand, token = split_authorization(payment_method)
+          credit_card = {}
+          credit_card[:securityCode] = options[:cvv] if options[:cvv]
+          post[:transaction][:creditCard] = credit_card
+          post[:transaction][:creditCardTokenId] = token
+          post[:transaction][:paymentMethod] = brand.upcase
+        else
+          credit_card = {}
+          credit_card[:number] = payment_method.number
+          credit_card[:securityCode] = payment_method.verification_value
+          credit_card[:expirationDate] = format(payment_method.year, :four_digits).to_s + '/' + format(payment_method.month, :two_digits).to_s
+          credit_card[:name] = payment_method.name.strip
+          post[:transaction][:creditCard] = credit_card
+          post[:transaction][:paymentMethod] = BRAND_MAP[payment_method.brand.to_s]
+        end
+      end
+
+      def add_payer(post, options)
+        if address = options[:billing_address]
+          payer = {}
+          post[:transaction][:paymentCountry] = address[:country]
+          payer[:fullName] = address[:name]
+          payer[:contactPhone] = address[:phone]
+          billing_address = {}
+          billing_address[:street1] = address[:address1]
+          billing_address[:street2] = address[:address2]
+          billing_address[:city] = address[:city]
+          billing_address[:state] = address[:state]
+          billing_address[:country] = address[:country]
+          billing_address[:postalCode] = address[:zip]
+          billing_address[:phone] = address[:phone]
+          payer[:billingAddress] = billing_address
+          post[:transaction][:payer] = payer
+        end
+      end
+
+      def add_extra_parameters(post, options)
+        extra_parameters = {}
+        extra_parameters[:INSTALLMENTS_NUMBER] = options[:installments_number] || 1
+        post[:transaction][:extraParameters] = extra_parameters
+      end
+
+      def add_reference(post, authorization)
+        order_id, transaction_id = split_authorization(authorization)
+        order = {}
+        order[:id] = order_id
+        post[:transaction][:order] = order
+        post[:transaction][:parentTransactionId] = transaction_id
+        post[:transaction][:reason] = 'n/a'
+      end
+
+      def add_payment_method_to_be_tokenized(post, payment_method)
+        credit_card_token = {}
+        credit_card_token[:payerId] = generate_unique_id
+        credit_card_token[:name] = payment_method.name.strip
+        credit_card_token[:identificationNumber] = generate_unique_id
+        credit_card_token[:paymentMethod] = BRAND_MAP[payment_method.brand.to_s]
+        credit_card_token[:number] = payment_method.number
+        credit_card_token[:expirationDate] = format(payment_method.year, :four_digits).to_s + '/' + format(payment_method.month, :two_digits).to_s
+        credit_card_token[:securityCode] = payment_method.verification_value
+        post[:creditCardToken] = credit_card_token
+      end
+
+      def commit(action, params)
+        begin
+          raw_response = ssl_post(url, post_data(params), headers)
+          response = parse(raw_response)
+        rescue ResponseError => e
+          raw_response = e.response.body
+          response_error(raw_response)
+        rescue JSON::ParserError
+          unparsable_response(raw_response)
+        else
+          success = success_from(action, response)
+          Response.new(
+            success,
+            message_from(action, success, response),
+            response,
+            authorization: success ? authorization_from(action, response) : nil,
+            error_code: success ? nil : error_from(action, response),
+            test: test?
+          )
+        end
+      end
+
+      def headers
+        {
+          "Content-Type"  => "application/json",
+          "Accept"  => "application/json"
+        }
+      end
+
+      def post_data(params)
+        params.to_json
+      end
+
+      def url
+        test? ? test_url : live_url
+      end
+
+      def parse(body)
+        JSON.parse(body)
+      end
+
+      def success_from(action, response)
+        case action
+        when 'store'
+          response["code"] == "SUCCESS" && response["creditCardToken"] && response["creditCardToken"]["creditCardTokenId"].present?
+        when 'verify_credentials'
+          response["code"] == "SUCCESS"
+        else
+          response["code"] == "SUCCESS" && response["transactionResponse"] && (response["transactionResponse"]["state"] == "APPROVED")
+        end
+      end
+
+      def message_from(action, success, response)
+        case action
+        when 'store'
+          if success
+            response["code"]
+          elsif response["error"]
+            response["error"]
+          elsif response["creditCardToken"]
+            response["creditCardToken"]["errorDescription"]
+          else
+            "FAILED"
+          end
+        when 'verify_credentials'
+          if success
+            "VERIFIED"
+          else
+            "FAILED"
+          end
+        else
+          if success
+            response["transactionResponse"]["responseCode"]
+          elsif response["error"]
+            response["error"]
+          elsif response["transactionResponse"]
+            response["transactionResponse"]["responseMessage"] || response["transactionResponse"]["responseCode"]
+          else
+            "FAILED"
+          end
+        end
+      end
+
+      def authorization_from(action, response)
+        case action
+        when 'store'
+          [
+            response["creditCardToken"]["paymentMethod"],
+            response["creditCardToken"]["creditCardTokenId"]
+          ].compact.join("|")
+        when 'verify_credentials'
+          nil
+        else
+          [
+            response["transactionResponse"]["orderId"],
+            response["transactionResponse"]["transactionId"]
+          ].compact.join("|")
+        end
+      end
+
+      def split_authorization(authorization)
+        authorization.split("|")
+      end
+
+      def error_from(action, response)
+        case action
+        when 'store'
+          response["creditCardToken"]["errorDescription"] if response["creditCardToken"]
+        when 'verify_credentials'
+          response["error"] || "FAILED"
+        else
+          response["transactionResponse"]["errorCode"] || response["transactionResponse"]["responseCode"] if response["transactionResponse"]
+        end
+      end
+
+      def response_error(raw_response)
+        begin
+          response = parse(raw_response)
+        rescue JSON::ParserError
+          unparsable_response(raw_response)
+        else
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,414 @@
...
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club]
+
+      BRAND_MAP = {
+        "visa" => "VISA",
+        "master" => "MASTERCARD",
+        "american_express" => "AMEX",
+        "diners_club" => "DINERS"
+      }
+
+      # Override defaults using options
```
duff: I'm thinking we can remove these comments.

duff:
@shasum I'd love to look at this PR with all of the commented out tests removed.  It'll simplify things a bunch.  If support for those operations are ever added to the gateway sandbox, we can add tests then.  For now, a small comment in the remote test file explaining that only some operations are supported in the sandbox should suffice.

shasum:
@duff feedback updates and more tests that cover all supported Latam countries

duff:
I haven't forgotten about this one!

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,393 @@
...
+      end
+
+      # Gateway supports authorize for certain countries only,
+      # hence a combination of purchase and void for verify
+      #
+      # Each country has different minimum amounts,
+      # hence 100.00 instead of typical 1.00
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { purchase(10000, credit_card, options) }
```
duff: I'm thinking we should go with `authorize` here even if some countries don't support it.  We wouldn't want a `verify` to actually charge someone's card since  the `void` could fail.  If a country doesn't support `authorize`, then it doesn't support `verify`.

duff: I also wonder if $100 is the lowest in all countries.  That's a pretty high minimum!

test/remote/gateways/remote_payu_latam_test.rb
```diff
@@ -0,0 +1,213 @@
...
+      billing_address: address(address1: "Viamonte",
+                       address2: "1366",
+                       city: "Plata",
+                       state: "Buenos Aires",
+                       country: "AR",
+                       zip: "64000",
+                       phone: "7563126")
+    }
+
+    @options_brazil = {
```
duff: I'm mulling on whether we need tests for each country.  Is there any coding difference in the gateway for each country, or is it mostly a different billing address and `account_id`?

test/remote/gateways/remote_payu_latam_test.rb
```diff
@@ -0,0 +1,213 @@
...
+
+  def test_successful_purchase_peru
+    gateway = PayuLatamGateway.new(fixtures(:payu_latam).update(:account_id => "512323"))
+    response = gateway.purchase(@amount, @credit_card, @options_peru)
+    assert_success response
+    assert_equal "APPROVED", response.message
+    assert response.test?
+  end
+
+  def test_successful_purchase_sans_options
```
duff: This is super useful methinks!  👍

duff:
It would be great if we could also support `verify_credentials` in new gateways.  Examples:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L240

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L164

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/blue_snap.rb#L117

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/braintree_blue.rb#L178

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,393 @@
...
+      def add_transaction_type(post, type)
+        transaction = {}
+        transaction[:type] = type
+        post[:transaction] = transaction
+      end
+
+      def add_order(post, options)
+        order = {}
+        order[:accountId] = @options[:account_id]
+        order[:referenceCode] = options[:order_id] || SecureRandom.hex(16)
```
duff: Could use this : 
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L107

duff: And in the other places where we're referring to `SecureRandom.hex(16)`.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,393 @@
...
+          credit_card[:securityCode] = payment_method.verification_value
+          credit_card[:expirationDate] = format(payment_method.year, :four_digits).to_s + '/' + format(payment_method.month, :two_digits).to_s
+          credit_card[:name] = payment_method.name.strip
+          post[:transaction][:creditCard] = credit_card
+          post[:transaction][:paymentMethod] = BRAND_MAP[payment_method.brand.to_s]
+        end
+      end
+
+      def add_payer(post, options)
+        if address = options[:billing_address] || options[:address]
```
duff: Thinking we can just use `options[:billing_address]` and not worry about `options[:address]`.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,393 @@
...
+      end
+
+      def add_buyer(post, options)
+        buyer = {}
+        if dni_number = options[:buyer_dni_number] || options[:dni_number]
+          buyer[:dniNumber] = dni_number
+        end
+        buyer[:cnpj] = options[:cnpj] if options[:cnpj]
+        buyer[:emailAddress] = options[:shipping_email] || options[:email] || 'unspecified@example.com'
+        if address = options[:shipping_address] || options[:billing_address] || options[:address]
```
duff: Thinking we can just rely on `options[:shipping_address]` here and disregard the other 2.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,393 @@
...
+        post = {}
+
+        action, _, _ = split_authorization(authorization)
+
+        case action
+        when 'auth'
+          add_credentials(post, 'SUBMIT_TRANSACTION')
+          add_transaction_type(post, 'VOID')
+          add_reference(post, authorization)
+        when 'purchase', 'capture'
```
duff: Usually `void` happens on authorizations and `refund` happens on captures and purchases.  I'm wondering if we can have a `def refund` and split it out into 2 operations.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,393 @@
...
+                            post[:transaction][:order][:additionalValues][:TX_VALUE][:value] + '~' +
+                            post[:transaction][:order][:additionalValues][:TX_VALUE][:currency]
+        Digest::MD5.hexdigest(signature_string)
+      end
+
+      def add_payment_method(post, payment_method)
+        if payment_method.is_a?(String)
+          _, brand, cvv, token = split_authorization(payment_method)
+          credit_card = {}
+          credit_card[:securityCode] = cvv
```
duff: Can this work without the `cvv`?  Since the `cvv` can't be stored for PCI reasons, we can't keep it in the authorization.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,393 @@
...
+
+      def initialize(options={})
+        requires!(options, :merchant_id, :account_id, :api_login, :api_key)
+        super
+      end
+
+      def purchase(amount, payment_method, options={})
+        post = {}
+
+        add_credentials(post, 'SUBMIT_TRANSACTION')
```
duff: Looks like there's some logic that shows up in `purchase` and `authorize`.  Perhaps it can be refactored a bit to DRY things up some.

lib/active_merchant/billing/gateways/payu_latam.rb
duff: Guess like a bunch of operations, it's difficult to tell since the sandbox doesn't support store.  😭

duff:
@shasum Getting closer!  👍 

Made a few comments / questions of things I noticed.

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: Agreed. There is a getting charged risk if/when `void` fails. Will switch to `authorize`. I wasn't sure if it was okay to implement an op that worked selectively. Now, I figure we just let users know about upstream gateway's limits in terms how things work.

test/remote/gateways/remote_payu_latam_test.rb
shasum: As there was lot of back n forth while testing this gateway, I had at various times tried out each different `account_id` that maps to a different country as you rightly guessed. In the final commit, I just left them all in.

Pattern wise, there is Brazil and the rest, so this can be trimmed down to just those two.

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: 👍  had noticed it in [test_helper.rb](https://github.com/activemerchant/active_merchant/blob/master/test/test_helper.rb#L235), didn't realize there's a similar one under [gateway.rb](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L107)

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: `shipping_address` else `billing_address` else `address`

think I saw this address fallback sequence on another gateway and borrowed it.

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: 👍 

Cool. The association between `void-authorization` and `refund-purchase` makes sense. Noticing the gateway does not support partial refunds, I somehow bunched those two together.

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: Right. Can't think of a way to find out for sure without running it in prod. Gateway [docs](http://developers.payulatam.com/en/api/tokenization.html) has this to say about stored tokens:
> Note: It is impossible to process transactions without validation of the cvv2 security code for the following countries. You need to send the SecurityCode parameter in the request. Argentina, Mexico, Panama, Peru.

Brazil and Colombia seem to be excluded from this `cvv` requirement as per [docs](http://developers.payulatam.com/en/api/tokenization.html).

Do you see a clean alternative to NOT keeping `cvv` in the returned authorization and still supporting `store` for all countries (or) we choose to support `store` for only Brazil and Colombia for now?

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: Yup, agreed, definitely needs some DRYing 👍

shasum:
@duff updated commit pushed

Previous feedback items have been addressed except for the following,
- Brazil fails when `shipping_address` is not provided. Hence, `billing_address` has been retained as a shipping default with override.
- Removing `cvv` from authorization. Refer previous comment thread to finalize this part.

lib/active_merchant/billing/gateways/payu_latam.rb
duff: I'm thinking our only option is to not store the cvv in the authorization.  This means it may work in some areas and not in others.

duff:
So if a customer specifies no billing_address and no shipping address, then it'll work, right?

The Brazil shipping address thing is an odd one!  Does it happen if the country is specified for the billing address and Brazil is specified?  I don't think I've ever heard of a shipping_address being required.  What if nothing is being shipped?

I'm wondering if we can bag on using the billing_address as a default shipping address.  Then if a customer gets an error message about shipping address being required in some circumstances, then they might be able to adjust their gateway settings in their gateway UI to turn off such a requirement.  Or they can start passing shipping address in those circumstances.  I'm a bit cautious about an implicit default shipping address.  Whatcha think?

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: 😿  will go ahead and change

shasum:
>So if a customer specifies no billing_address and no shipping address, then it'll work, right?

For countries other than Brazil, it works, no address no problem. But, in case of Brazil, it fails Antifraud check if either address is omitted ( both addresses needed for successful Brazil txn )

>The Brazil shipping address thing is an odd one! Does it happen if the country is specified for the billing address and Brazil is specified? I don't think I've ever heard of a shipping_address being required. What if nothing is being shipped?

It happens when the Brazilian `account_id` is used along with Brazilian currency and billing address inputs. Under these circumstances it expects a non-null shipping address, else returns `ANTIFRAUD_REJECTED` error. Good question about nothing being shipped, not sure what the gateway expects in such cases.

>I'm wondering if we can bag on using the billing_address as a default shipping address. Then if a customer gets an error message about shipping address being required in some circumstances, then they might be able to adjust their gateway settings in their gateway UI to turn off such a requirement. Or they can start passing shipping address in those circumstances. I'm a bit cautious about an implicit default shipping address. Whatcha think?

Initially, I was hesitant as well in having such implicit defaults. Later, we had use cases like [this](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L252-L262), which had me think it might be okay to go with such address related defaults.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,400 @@
...
+        post[:transaction][:order] = order
+      end
+
+      def add_buyer(post, options)
+        buyer = {}
+        if dni_number = options[:buyer_dni_number] || options[:dni_number]
+          buyer[:dniNumber] = dni_number
+        end
+        buyer[:cnpj] = options[:cnpj] if options[:cnpj]
+        buyer[:emailAddress] = options[:shipping_email] || options[:email] || 'unspecified@example.com'
```
duff: I wonder if we can simplify this to just:

```
buyer[:emailAddress] = options[:email] || 'unspecified@example.com'
```

duff:
You're right about the implicit defaults for billing address!  

I wonder if there's a way to turn off the `ANTIFRAUD_REJECTED` error for a Brazilian `account_id` in their gateway UI.

Perhaps an option is to have a default shipping_address like you mentioned in [this](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L252-L262).  So it would use the specified `shipping_address` from the customer, or it would use the hard coded default.  But it would be separate from the `billing_address`.

Hoping I'm not being too nitzy.

shasum:
Additional [feedback](https://spreedly.freshdesk.com/helpdesk/tickets/22726) from gateway support. Need to think through changes needed, if any, from Vladimir's observations.

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: 👍

shasum:
>I wonder if there's a way to turn off the ANTIFRAUD_REJECTED error for a Brazilian account_id in their gateway UI.

Could be, but no way to check that immediately. Using open test credentials for their sandbox from the API [docs](http://developers.payulatam.com/en/api/sandbox.html) page. There doesn't seem to be a self service portal for the sandbox.
>Perhaps an option is to have a default shipping_address like you mentioned in this. So it would use the specified shipping_address from the customer, or it would use the hard coded default. But it would be separate from the billing_address.

Thinking it might be better to let the error pass through in this case. Allow customers to send in the shipping_address when required. Hard coding an unrelated address feels odd.

duff:
@shasum Sounds great!

shasum:
@duff end in sight!

Quite a few changes:
- cvv removed from authorization
- billing_address, shipping_address, and email cleaned up
- incorporated all feedback from gateway support
- added query method that helps check status of pending transactions

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+        post = {}
+
+        action, _, _ = split_authorization(authorization)
+
+        case action
+        when 'purchase', 'capture'
+          add_credentials(post, 'SUBMIT_TRANSACTION')
+          add_transaction_type(post, 'REFUND')
+          add_reference(post, authorization)
+        else
```
duff: Same here - I don't know that we need this, do we?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+          message = "Unsupported operation: successful purchase and capture transactions can only be refunded."
+          return Response.new(false, message)
+        end
+
+        commit('refund', post)
+      end
+
+      # Supported countries have different minimum amounts and
```
duff: I wonder if the currency code factors in.  If they specify the currency in the options, can we lower the amount?

Most gateways have 100 for the `verify` amount.  4000 seems so high!   

I also wonder if for most currencies perhaps 100 might work but for a few currencies we could bump it up to 4000.  Or maybe 4000 all the time is simpler.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+        post = {}
+
+        add_credentials(post, 'CREATE_TOKEN')
+        add_payment_method_to_be_tokenized(post, payment_method)
+
+        commit('store', post)
+      end
+
+      def verify_credentials
+        response = authorize(4000, "_|VISA|00000000-0000-0000-0000-000000000000", {cvv: '000'})
```
duff: That's a pretty cool idea.  

I wonder if they have an API call that could be used that doesn't take a card.  The main thought I had was would we break if they changed the message text.  Seems like historically, messages are more likely to change than response codes.  But perhaps messages are one of the only ways to distinguish between error states.  Or maybe this message won't change and we'll be fine.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+        response.message.include? "The ID [00000000-0000-0000-0000-000000000000] of credit card token is invalid"
+      end
+
+      def query(authorization, options={})
```
duff: Not sure if we need `query`.  We typically let folks do things like this outside of active merchant since active merchant is primarily for the main X operations like `purchase`, `authorize`, etc.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+        order[:referenceCode] = options[:order_id] || generate_unique_id
+        order[:description] = options[:description] || 'unspecified'
+        order[:language] = 'en'
+        post[:transaction][:order] = order
+      end
+
+      def add_buyer(post, options)
+        buyer = {}
+        buyer[:dniNumber] = options[:buyer_dni_number] if options[:buyer_dni_number]
+        buyer[:cnpj] = options[:buyer_cnpj] if options[:buyer_cnpj]
```
duff: Not sure if we need this, do we?  Same goes for `dniNumber`.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+          shipping_address[:state] = address[:state]
+          shipping_address[:country] = address[:country]
+          shipping_address[:postalCode] = address[:zip]
+          shipping_address[:phone] = address[:phone]
+          buyer[:shippingAddress] = shipping_address
+        end
+        post[:transaction][:order][:buyer] = buyer
+      end
+
+      def add_shipping_address(post, options)
```
duff: I know they sent an email about specifying the shipping address more than once.  I do wonder why that's needed!  Is there one that's more important than the other?  Seems weird to be doing it twice, once for the buyer and once for the transaction order.  Would it be bad to only specify it for the transaction order?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+      def add_invoice(post, money, options)
+        tx_value = {}
+        tx_value[:value] = amount(money)
+        tx_value[:currency] = options[:currency] || currency(money)
+
+        tx_tax = {}
+        tx_tax[:value] = '0'
+        tx_tax[:currency] = options[:currency] || currency(money)
+
+        tx_tax_return_base = {}
```
duff: I saw the email too where they were talking about these TAX fields.  Do transactions fail without all of these `additional_values`?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+      def add_signature(post)
+        post[:transaction][:order][:signature] = signature_from(post)
+      end
+
+      def signature_from(post)
+        signature_string = [
+          @options[:api_key],
+          @options[:merchant_id],
+          post[:transaction][:order][:referenceCode],
+          post[:transaction][:order][:additionalValues][:TX_VALUE][:value],
```
duff: The fact that these are part of the signature leads me to think that the TAX values are required.  :(

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+          credit_card[:expirationDate] = format(payment_method.year, :four_digits).to_s + '/' + format(payment_method.month, :two_digits).to_s
+          credit_card[:name] = payment_method.name.strip
+          post[:transaction][:creditCard] = credit_card
+          post[:transaction][:paymentMethod] = BRAND_MAP[payment_method.brand.to_s]
+        end
+      end
+
+      def add_payer(post, options)
+        payer = {}
+        payer[:dniNumber] = options[:payer_dni_number] if options[:payer_dni_number]
```
duff: Necessary?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+        payer[:cnpj] = options[:payer_cnpj] if options[:payer_cnpj]
```
duff: I'm not sure what `dniNumber` and `payer_cnpj` represent or how likely customers are to use them.  Do you think it's likely someone might want them?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+        payer[:emailAddress] = options[:email] || 'unspecified@example.com'
```
duff: So `email` gets specified for transaction->payer and for transaction->order->buyer.    I wonder if both are needed.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,484 @@
...
+          billing_address[:country] = address[:country]
+          billing_address[:postalCode] = address[:zip]
+          billing_address[:phone] = address[:phone]
+          payer[:billingAddress] = billing_address
+        end
+        post[:transaction][:payer] = payer
+      end
+
+      def add_browser_data(post, options)
+        post[:transaction][:deviceSessionId] = 'n/a'
```
duff: Do the transactions fail if `deviceSessionId`, `cookie`, and `userAgent` aren't specified.  If so, the 👍 on filling them the with `n/a`.

duff:
Nice!  I looked through and made a few comments, asked a few questions for background, etc.

All in all, it's looking pretty great; I'm just pushing a little bit to see if there's anything we can simplify.

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: It plays tricks on the eyes, but it's the TX_VALUE ( transaction value? ) that's part of signature. Got rid of all other TAX related values.

shasum:
@duff the no frills version!

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,400 @@
...
+
+      def url
+        test? ? test_url : live_url
+      end
+
+      def parse(body)
+        JSON.parse(body)
+      end
+
+      def success_from(action, response)
```
duff: I wonder if `response["code"] == "SUCCESS"` would work for all of 'em.  Are there cases where the response code is SUCCESS and there isn't a `creditCardToken` or a `transactionResponse`?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -0,0 +1,400 @@
...
+        when 'store'
+          response["code"] == "SUCCESS" && response["creditCardToken"] && response["creditCardToken"]["creditCardTokenId"].present?
+        when 'verify_credentials'
+          response["code"] == "SUCCESS"
+        else
+          response["code"] == "SUCCESS" && response["transactionResponse"] && (response["transactionResponse"]["state"] == "APPROVED")
+        end
+      end
+
+      def message_from(action, success, response)
```
duff: Any good ideas how we might refactor this a bit?  Perhaps extracting parts into other methods?

duff:
@shasum I made a comment or two and then merged it since those comments were nice to haves (and those could be new PR's if you have ideas as to how to address them).

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: For `auth` and `purchase`, it checks `transactionResponse.state` is Approved. 

iirc from a past error, validating that response has the item that we need (eg: `tokenId`) is an approach I adopted to prevent runtime `nil` errors when we subsequently read that value.

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: Yep, could use a better refactor. Will do a separate PR.

lib/active_merchant/billing/gateways/payu_latam.rb
duff: I wonder if... there is no `transactionResponse` info and response["code"] == SUCCESS, did money actually actually change hands?  If so, I'm wondering if we want to capture that as a successful transaction and then handle the nil `transactionResponse` subsequently?

lib/active_merchant/billing/gateways/payu_latam.rb
duff: Sounds great! 👍

lib/active_merchant/billing/gateways/payu_latam.rb
shasum: with `code == SUCCESS`, `transactionResponse.state` can be one of `APPROVED`, `DECLINED` or `PENDING`. Only the first one is a successful transaction, which cannot be determined using `code` alone.

duff: Sounds good @shasum.  Thanks for the clarification!  Different gateways seem to have a varying number of fields that need to be checked for success. 😢

PR #2117, opened 2016-05-30
Quickpay v10: Update test fixtures, tests

curiousepic:
@duff to confirm and merge

### DIFF
--- a/test/fixtures.yml
+++ b/test/fixtures.yml
@@ -773,11 +773,11 @@ quickbooks:
 # Quickpay offers a test account.
 # To log in to the manager, use demo@quickpay.net and test234
 quickpay:
-  login: 89898978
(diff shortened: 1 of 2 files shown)


duff:
When I run the `test/remote/gateways/remote_quickpay_v10_test.rb` tests, all but one of them work.

When I run `test/remote/gateways/remote_quickpay_test.rb` tests, they all seem to break with:

```
Error: test_unsuccessful_purchase_with_missing_cvv2(RemoteQuickpayTest): ArgumentError: Missing required parameter: api_key
/Users/duff/code/active_merchant/lib/active_merchant/billing/gateway.rb:303:in `block in requires!'
/Users/duff/code/active_merchant/lib/active_merchant/billing/gateway.rb:296:in `each'
/Users/duff/code/active_merchant/lib/active_merchant/billing/gateway.rb:296:in `requires!'
```

The Quickpay stuff can be confusing with all of the versions.  Is there something special I need to do to run them?  Were the tests busticated with the old creds?

PR #2121, opened 2016-06-01
Add SG to stripe support list

rhlrjv:
Stripe is now in Singapore


rhlrjv:
https://stripe.com/sg

PR #2125, opened 2016-06-02
Add support for scrubbing to Orbital

bruno:
Adding support for scrubbing to Orbital gateway. It would seem that
every test suite sent by Orbital might require different test cards, I
am only committing the small change to point to the correct fixture
hash.


timbeiko:
LGTM :+1:

Issue #2130, opened 2016-06-07
Stripe: Store method does not cleanly accept string inputs interpreted as tokens


bizla:
Currently the Stripe gateway integration doesn't have a tested code path for accepting [Stripe card tokens](https://stripe.com/docs/stripe.js?#collecting-card-details).
Specifically, an explicit call to `card_brand` throws an exception because of a conditional check that assumes that the input is a credit-card-like object, whereas for other methods like `purchase`, a string input is handled correctly; it's interpreted as a Stripe token string. 

See the discussion in https://github.com/activemerchant/active_merchant/pull/2098 for an example of an affected consumer.

jordan-brough:
Can we move `StripePaymentToken` out of the `private` area ([here](https://github.com/activemerchant/active_merchant/blob/4881e20572604eec6cf787d38e82708e58bd71c5/lib/active_merchant/billing/gateways/stripe.rb#L269-L275)) so that it's a bit more prominent?  Or else move it to its own file?  Also can we add a bit of documentation to it?  I'm happy to submit a PR for any of that if you'd like.

I ask because `StripePaymentToken` is now something that can be used across the board in `StripeGateway` and we have public methods that expect to receive it (e.g. `StripeGateway#store`).

PR #2131, opened 2016-06-07
[Bugfix] Stripe: Interpret string input to store method as token identifier

bizla:
Addresses #2130 

### Problem
We had an implicit and untested code path that allowed `StripeGateway` consumers to pass strings into `store` without headache. After https://github.com/activemerchant/active_merchant/commit/f2a7ea3ed298038529acce7552b733ade8f3bc90, the call to `card_brand` effectively disallowed string inputs, but no tests failed.

### Fix
Strings are interpreted (via if-else fallthrough) to be token identifiers in other methods like purchase, so this fix restores that behaviour by replacing the `card_brand` check, adds the ability to also pass in a `StripePaymentToken` for consistency, and the appropriate tests with some minor code cleanup along the way. 

Please review @timbeiko @aprofeit

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,7 +14,12 @@
 * AuthorizeNet: Support verify_credentials [duff]
 * BraintreeBlue: Support verify_credentials [duff]
 * Redsys: Added SAR currency [agseco]
+<<<<<<< HEAD
 * Stripe: Interpret string input to store method as token identifier [bizla]
+=======
+* MaxiPago: add void and refund [shasum]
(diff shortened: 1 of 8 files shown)


timbeiko:
LGTM 👍

aprofeit:
Add _all_ the tests! 👍

bizla:
Closed via 48b58efe660a9fc9522bc4d3fd8fb09c601c357c

jordan-brough:
@bizla thanks for getting this fixed!

PR #2133, opened 2016-06-07
Fixing a copy-paste error in the Stripe handling of track-only contactless EMV data

ryanbalsdon:
I think the title says it all! This was a really dumb mistake on my part... Fortunately, using raw track data for Contactless EMV transactions is not legal so this fix is not a high-priority.


bizla:
Can you add a test?

jasonwebster:
Test added.

jasonwebster:
This was fixed back in https://github.com/activemerchant/active_merchant/releases/tag/v1.63.0, https://github.com/activemerchant/active_merchant/commit/6798d1539ee95c93708fb803ab56d91181defd3d specifically.

PR #2138, opened 2016-06-16
do not overwrite response attribute if it's set before

alovak:
Recently I got the response from CyberSource that contains multiple 'DECISION' elements in xml. Due to existing logic in AM module the last decision will be used as a value to see if transaction was 'successful'.

This is a quick fix. I don't think it should be merged as is but more 'smart' solution should be found and discussed.

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -733,7 +733,7 @@ module ActiveMerchant #:nodoc:
             parent = node.parent.name + (node.parent.attributes["id"] ? "_" + node.parent.attributes["id"] : '')
             reply[(parent + '_' + node.name).to_sym] = node.text
           else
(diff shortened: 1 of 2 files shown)


alovak:
As alternative n. 1

we can change 'parse' method to this (tests pass):

```
      def parse(xml)
        reply = {}
              parse_element(reply, node)
          end
          reply[:decision] = REXML::XPath.first(root, "c:decision").text
          reply[:message] = REXML::XPath.first(root, "c:reasonCode").text
        elsif root = REXML::XPath.first(xml, "//soap:Fault")
          parse_element(reply, root)
          reply[:message] = "#{reply[:faultcode]}: #{reply[:faultstring]}"
        end
        return reply
      end
```

but if c:decision or c:reasonCode do not exist then this method will fail (undefined method `text' for nil:NilClass).

alovak:
I have created a different PR...

PR #2139, opened 2016-06-16
Cyber source should store all response elements correctly

alovak:
I got the response from CyberSource that contains multiple 'DECISION' elements in xml (you can find them in updated tests). Due to existing logic in AM module the last parsed decision will be used as a value to see if transaction was 'successful'.

In this PR we store all values according to its hierarchy e.g.:

``` xml
<c:authReply>
  <c:something>bla</c:something>
</c:authReply>
```

will be stored in response as [:authReply][:something] thus we do not overwrite values randomly... and can access values we want.

**Warning** It breaks existing integrations if you use more than just items from 'replyMessage'. Probably alternative fixes are more appropriate due to this incompatibility.... I would like to get some feedback on it.

P.S. here are two alternative ways to fix this:
1. `reply[node.name.to_sym] = node.text unless reply[node.name.to_sym]` durty but tests pass
2. 

```
      def parse(xml)
        reply = {}
              parse_element(reply, node)
          end
          reply[:decision] = REXML::XPath.first(root, "c:decision").text
          reply[:message] = REXML::XPath.first(root, "c:reasonCode").text
        elsif root = REXML::XPath.first(xml, "//soap:Fault")
          parse_element(reply, root)
          reply[:message] = "#{reply[:faultcode]}: #{reply[:faultstring]}"
        end
        return reply
      end
```

means we parse and store values as it was before (overwriting values ...) but decision and message are taken from the right place. But if c:decision or c:reasonCode do not exist then this method will fail (undefined method `text' for nil:NilClass).

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -699,8 +699,8 @@ module ActiveMerchant #:nodoc:
         Response.new(success, message, response,
           :test => test?,
           :authorization => authorization,
-          :avs_result => { :code => response[:avsCode] },
-          :cvv_result => response[:cvCode]
+          :avs_result => { :code => (response[:ccAuthReply] && response[:ccAuthReply][:avsCode]) },
+          :cvv_result => (response[:ccAuthReply] && response[:ccAuthReply][:cvCode])
         )
       end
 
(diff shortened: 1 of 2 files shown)


alovak:
@shasum @curiousepic @marquisong guys can you give me a feedback on this PR? I see that you work (worked) with CyberSource too.

alovak:
@duff maybe you  can provide some feedback on this?

davidsantoso:
Hi - I'm grooming old pull requests and if this is still something you're interested in, please:

1. Reopen the PR
2. Rebase on `master`
3. Squash all commits
4. Tag me for review

Thanks!

alovak:
@davidsantoso do you have any thoughts on it?

If I review and update, will you merge it?

PR #2142, opened 2016-06-23
Undefined variable in gateway generator remote test template

Stratus3D:
In `test_successful_authorize_and_capture` test assigns the response to `capture` but `response` is used in the response message assertion. Replace `response` with `capture`.


bpollack:
An equivalent of this appears to have already been merged! I apologize. Closing.

PR #1777, opened 2015-07-08
Stripe: Map incorrect_pin error code to new standard error code

abecevello:
Adds a new standard error code called `incorrect_pin` and maps it from Stripe.

This is used for Online PIN transactions to provide a better UX when the gateway lets us know the PIN was incorrect.

PR #2145, opened 2016-06-29
Stripe: Map test_mode_live_card error code to new standard error code

berkcaputcu:
Adds a new standard error code called `test_mode_live_card` and maps it from Stripe.

This is used to inform the user that they are using a real credit card number but the gateway is in test mode. We would like to translate this error to provide better UX, and we need to add the error to the list for it to be picked up as `error_code` [here](https://github.com/activemerchant/active_merchant/blob/16d8557dea8957263d72793dbb63394f3560b9e7/lib/active_merchant/billing/gateways/stripe.rb#L562-L569)

A similar PR is here #1777
Please review @bizla @girasquid 
/cc @emilienoel @Amos47

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -76,6 +76,7 @@ module ActiveMerchant #:nodoc:
       # :processing_error - Processor error
       # :call_issuer - Transaction requires voice authentication, call issuer
       # :pickup_card - Issuer requests that you pickup the card from merchant
+      # :test_mode_live_card - Card was declined. Request was in test mode, but used a non test card.
 
       STANDARD_ERROR_CODE = {
         :incorrect_number => 'incorrect_number',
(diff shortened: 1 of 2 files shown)


girasquid:
@duff can you take a look too? This might affect you folks

girasquid:
Code changes make sense. 👍

emilienoel:
👍

bizla:
LGTM!

duff:
👍 

Long term, I'd really like to rename `:error_code`:

https://billing.spreedly.com/core/subscribers/6KvMC0IyXPjs7vzAuovaYablrZc

to `:standard_error_code` or something like it.  And we'd still have an `:error_code`.  

The goal would be to be able to discern from a `Response` what the *actual* error code was from the gateway and to also be able to get the `standard_error_code` which is Active Merchant's generic mapping.

bizla:
> The goal would be to be able to discern from a Response what the actual error code was from the gateway and to also be able to get the standard_error_code which is Active Merchant's generic mapping.

@duff Maybe we can add this to our AM2.0 wishlist in the repo's issues?

duff:
> @duff Maybe we can add this to our AM2.0 wishlist in the repo's issues?

@bizla That's a good idea.  I wonder if there'd be a way to do it sooner than that such that it doesn't break anyone.  Perhaps `error_code` could remain as it is and we could have another field on the `Response` that is `gateway_error_code` (not sure of the best name for it).

PR #2147, opened 2016-06-30
Citruspay integrated plugin modified

viatechs:
Citruspay payment gateway plugin upgraded to use SSLV2 hosted payment page. Also a new payment form parameter 'notifyUrl' introduced which will take similar value of returnUrl. notifyUrl will allow direct server response in case of delayed payment clearance.


bpollack:
There's some good in this, but we really can't merge this as-is; the PR deletes comments from the README, modifies supported countries for Stripe, and tons of other things that have nothing to do with the stated objectives. Please feel free to open smaller, more targeted PRs/

PR #2160, opened 2016-07-12
Iugu Gateway: authorize, capture and purchase methods

alexandremcosta:
reference: iugu.com/referencias/api

### DIFF
--- a/lib/active_merchant/billing/gateways/iugu.rb
+++ b/lib/active_merchant/billing/gateways/iugu.rb
@@ -1,0 +1,243 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class IuguGateway < Gateway
+      self.live_url = 'https://api.iugu.com/v1/'
+
+      self.supported_countries = %w(BR)
+      self.default_currency = 'BRL'
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club]
+
+      self.homepage_url = 'https://iugu.com/'
+      self.display_name = 'Iugu'
+
+      def initialize(options = {})
+        requires!(options, :login)
+        @api_key = options[:login]
+
+        super
+      end
+
+      def authorize(money, payment, options = {})
+        token = payment_token(payment, options)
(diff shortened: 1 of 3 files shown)


henriquesacramento:
hi, nice job.
How can i use that implementation in my project?

Thanks.

alexandremcosta:
Until activemerchant/active_merchant accept this pull request, the only options is to include my repository in your project. According to this doc (http://bundler.io/git.html), this is the way to go:
`gem 'active_merchant', github: 'alexandremcosta/active_merchant', branch: 'iugu'`

henriquesacramento:
Ok. I do that.
But in application backend i can't see any option to configure iugu.

Can u help me?

iagopiimenta:
Any chance of being merged?

gcmartins93:
It's been a while since this pull was requested, no updates since then? It would be great to have any response from author/contributors.

alexandremcosta:
Looks like it's the fifth next in the line. I have no permission to do the merge by myself, so unfortunatelly I cannot do anything besides waiting, @gcmartins93. Additionally, the project that required me to create this PR doesn't exist anymore :(  
Good luck!

gcmartins93:
Thanks for reply, @alexandremcosta, it's being a while since this PR was open. Do you think that it would take too long to be merged? In the meanwhile do you suggest to use your branch?

alexandremcosta:
As I can see, there is a queue of approval, from the last to the first. Its taking more than a year, from the submission to the approval. Since there is nothing we can do, I recommend you to use my branch, or to drop activemerchant and create your own integration using Iugu API.

gcmartins93:
Thanks for the fast reply, it helped to move forward with my decision.

Vipkry:
It might be a good idea to have :order_id as a param inside create_post_for_auth so we can link both our order to Iugu charge and Iugu charge to our order.

Also, why does the comments indicate us the use of the purchase method to create bank_slips if those are only captured after some days? It doesn't make much sense to try and capture it right after creating the bank_slip because the auto capture will never work...

PR #2162, opened 2016-07-13
Add supported countries by MONEI

alexandresaiz:
Adding all the supported countries. AD, AT, BE, BG, CH, CY, CZ, DE, DK, EE, ES, FI, FO, FR, GB, GI, GR, HU, IE, IL, IS, IT, LI, LT, LU, LV, MT, NL, NO, PL, PT, RO, SE, SI, SK, TR, VA


alexandresaiz:
@duff any chance to have this merged before the end of July?

duff:
@alexandresaiz We track the supported countries in the code:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/monei.rb#L17

And then generate that part of the readme with a rake task.  Also important to note that supported_countries means countries where the gateway onboards merchants.

alexandresaiz:
@duff thanks! Closing this then, as @davidgf as created a new PR following your instructions

PR #2164, opened 2016-07-14
Barclays EPDQ extra plus TLS bumped to 1.2

tahb:
- We’ve received warning that EPDQ will be turning off support for TLSv1 on the 31st August 2016 so to keep ActiveMerchant functional we needed to increase this value
- This information was given to us via email, the only link provided was pcisecuritystandards.org/documents/PCI_DSS_v3-1_Summary_of_Changes.pdf which is what they are conforming to
- 1.1 could have been set but they’ve also advise that that will be disabled in 2018

### DIFF
--- a/lib/active_merchant/billing/gateways/barclays_epdq_extra_plus.rb
+++ b/lib/active_merchant/billing/gateways/barclays_epdq_extra_plus.rb
@@ -10,6 +10,7 @@ module ActiveMerchant #:nodoc:
       self.supported_countries = ["GB"]
       self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :discover, :jcb, :maestro]
       self.default_currency = "GBP"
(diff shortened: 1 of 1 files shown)


bpollack:
@tahb I don't think this patch is actually necessary; generally, Active Merchant will use the highest version of TLS available, so if they only support TLS v1.2, then Active Merchant should automatically default to that. Is that not the behavior you're seeing?

tahb:
@bpollack This was a very long time ago, so long that we no longer use ActiveMerchant :( 

At the time this was a preemptive bump that we started using it before they made v1.2 mandatory to ensure everything would remain functional. Our confidence with EPDQ was very low.

I would assume you are correct since this change was meant to happen 1.5 years ago and there hasn't been any fix made after the fact that I can see.

Closing 👋

PR #2165, opened 2016-07-14
Jetpay: Support endpoint for Canada

shasum:


lib/active_merchant/billing/gateways/jetpay.rb
```diff
@@ -150,6 +153,7 @@ class JetpayGateway < Gateway
 
       def initialize(options = {})
         requires!(options, :login)
+        options[:region] = 'US' unless options[:region]
```
duff: Thinking this line may not be needed.

duff:
@shasum Looks good!  Just one little question/comment.

PR #2178, opened 2016-07-26
Rename CHANGELOG to CHANGELOG.rdoc

swrobel:
CHANGELOG appears to be in rdoc format, so it should have that extension. Not having an extension breaks changelog parsers like [vandamme](https://github.com/tech-angels/vandamme)


bpollack:
While this may be the right thing to do, we're unfortunately at a place where changing the `CHANGELOG` extension would break downstream users in a way that's probably not worth it. Thank you for the proposal, though.

Issue #2172, opened 2016-07-20
QuickPay: enable recurring payments in v10


wsmoak:
As I commented on https://github.com/activemerchant/active_merchant/pull/1873, the current implementation of 'store' for QuickPay v10 returns a single-use token.

The existing test demonstrates that you can store and then use it to make a purchase, but if you attempt a second purchase, as for a subscription renewal, it will fail because the token has already been used.

I tested this with the API directly with curl and checked with QuickPay support to confirm that...

Storing a card is:
1. POST https://api.quickpay.net/cards (no body)
2. POST https://api.quickpay.net/cards/$CARD_ID/authorize?synchronized (card details)

Now the card is stored and can be reused, so

Creating a payment is:
1. POST https://api.quickpay.net/cards/$CARD_ID/tokens?synchronized (no body)
2. POST https://api.quickpay.net/payments (currency, order_id)
3. POST https://api.quickpay.net/payments/$PAYMENT_ID/authorize?synchronized (amount, token from step 1)
4. POST https://api.quickpay.net/payments/$PAYMENT_ID/capture?synchronized (amount)

I'll branch and see if I can re-work this. Let me know if anyone else is already working on it or if you have any comments/suggestions!

wsmoak:
Based on discussion in https://github.com/activemerchant/active_merchant/pull/1873, 'purchase' needs to support both

1. a single purchase with full card details, and
2. a recurring purchase using a token

It appears that `/payments/:id/authorize` takes either a token or the full card info, so this should be possible... http://tech.quickpay.net/api/services/?scope=merchant#POST-payments--id-authorize---format-

wsmoak:
I've got a fix working for this, but I'm having to dig through the response to find the payment id I need, when based on the existing code I think it _should_ be in 'r.authorization'.

https://github.com/activemerchant/active_merchant/compare/master...chargify:quickpay_recurring

Could someone take a look?  Or should I go ahead and open a PR since there is now code to discuss?  Thanks!  /cc @duff

duff:
Looks like they don't make it super cinchy to run recurring purchases against a stored card.  Tokens, cards, references, auths, captures.  Whew!  Pretty great that you've got a remote test demonstrating the issue.

👍 for a PR

davidsantoso:
Closed with ed4ed2d03837a3a7db6a22fc37861c22f3779c90. Thanks for opening up this issue @wsmoak!

PR #2180, opened 2016-07-26
Quickpay: enable recurring payments

wsmoak:
As discussed in
- https://github.com/activemerchant/active_merchant/pull/1873 and
- https://github.com/activemerchant/active_merchant/issues/2172

The current implementation of store for QuickPay v10 is returning a single-use token, which prevents using the stored card details for recurring payments.  We should be storing the card id instead.

This moves the single-use token into 'purchase' and 'authorize' where the stored card id is swapped for a token just before the authorization/capture is done.

### DIFF
--- a/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
+++ b/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
@@ -16,25 +16,33 @@ module ActiveMerchant
 
       def purchase(money, credit_card_or_reference, options = {})
         MultiResponse.run(true) do |r|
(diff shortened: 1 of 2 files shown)


mroach:
I wish I had seen this before implementing nearly the exact same fix haha. Oh my. Facepalm on me. WebGents/active_merchant@5ee841d

wsmoak:
@mroach I'm glad to hear that you independently arrived at a similar solution!

curiousepic:
@wsmoak @mroach Thanks for spotting this issue and contributing, and apologies for the lag in merging a good solution!

rickywiens:
Thanks for resolving this.

PR #2182, opened 2016-07-28
CyberSource: Combine auth_reversal with Void

curiousepic:
Void was failing for authorizations because voiding captures and
reversing auths are distinct actions on the gateway's side. We now
remember the type of action, amount, and currency of each transaction
in the multi-auth, and condition on it to determine which action to
take when void is called on either type of transaction.

@duff to review and merge.


duff:
Looking good!

PR #2189, opened 2016-08-08
Pass metadata when voiding an authorization on Stripe

berkcaputcu:
Currently we're not passing the metadata to Stripe when it's passed as an option. This is something we do for `sale`, `auth` and `refund` transactions. However it's missing for void.

We'd like to make use of this metadata on `void` transactions, so I'd like to add this parameter to `void` requests as well.

@girasquid @emilienoel @Amos47 @timbeiko @lyverovski

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -117,6 +117,7 @@ module ActiveMerchant #:nodoc:
 
       def void(identification, options = {})
         post = {}
+        post[:metadata] = options[:metadata] if options[:metadata]
         post[:expand] = [:charge]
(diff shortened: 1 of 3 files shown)


girasquid:
Can you add a remote test that confirms this works? Right now you're confirming that your option gets passed through, but nothing here shows that Stripe supports this

berkcaputcu:
> Can you add a remote test that confirms this works? Right now you're confirming that your option gets passed through, but nothing here shows that Stripe supports this

👍 done

Amos47:
This LGTM 👍

girasquid:
😍 lgtm

emilienoel:
👍

PR #2195, opened 2016-08-19
Remove weak hard-coded TLS versions

mutemule:
All modified gateways support up to TLSv1.2, so let's not keep some capped at TLSv1.

As far as I can tell, all these gateways were forced to TLSv1 as a way to avoid SSLv3. Since then, modern versions of Ruby [only support TLSv1+](https://github.com/ruby/ruby/blob/61a3fff66141ffc0b6e384729456c6ca87f61776/ext/openssl/lib/openssl/ssl.rb#L61), so let's not punish more modern Rubys by capping the TLS version allowed by a specific gateway.

I'm also considering added specific TLS version negotiation into `Connection::configure_ssl`, so we have a consistent experience across all of AM, regardless of the underlying Ruby.

@duff @ntalbott

### DIFF
--- a/lib/active_merchant/billing/gateways/citrus_pay.rb
+++ b/lib/active_merchant/billing/gateways/citrus_pay.rb
@@ -16,7 +16,6 @@ module ActiveMerchant
       self.supported_countries = %w(AR AU BR FR DE HK MX NZ SG GB US)
       self.default_currency = 'USD'
       self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb, :maestro, :laser]
-      self.ssl_version = :TLSv1
 
     end
   end
--- a/lib/active_merchant/billing/gateways/dibs.rb
+++ b/lib/active_merchant/billing/gateways/dibs.rb
@@ -9,7 +9,6 @@ module ActiveMerchant #:nodoc:
       self.supported_countries = ["US", "FI", "NO", "SE", "GB"]
       self.default_currency = "USD"
       self.money_format = :cents
-      self.ssl_version = :TLSv1
       self.supported_cardtypes = [:visa, :master, :american_express, :discover]
 
       def initialize(options={})
--- a/lib/active_merchant/billing/gateways/global_transport.rb
+++ b/lib/active_merchant/billing/gateways/global_transport.rb
@@ -9,7 +9,6 @@ module ActiveMerchant #:nodoc:
       self.supported_countries = %w(CA PR US)
       self.default_currency = 'USD'
       self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb]
-      self.ssl_version = :TLSv1
 
       self.homepage_url = 'https://www.globalpaymentsinc.com'
       self.display_name = 'Global Transport'
--- a/lib/active_merchant/billing/gateways/netbilling.rb
+++ b/lib/active_merchant/billing/gateways/netbilling.rb
@@ -31,7 +31,6 @@ module ActiveMerchant #:nodoc:
       self.display_name = 'NETbilling'
       self.homepage_url = 'http://www.netbilling.com'
       self.supported_countries = ['US']
-      self.ssl_version = :TLSv1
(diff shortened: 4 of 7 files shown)


duff:
We've had trouble connecting without the hard coded values.  I don't think the version of Ruby we're using is very old.  Any idea what version of Ruby one needs to be at?

Would moving the hardcoded ones to `TLSv1_2` be a possibility?

mutemule:
> We've had trouble connecting without the hard coded values. I don't think the version of Ruby we're using is very old. Any idea what version of Ruby one needs to be at?

Do you remember any specifics? I know we had troubles with a very small number of gateways at one point, but I am under the impression these were all solved.

> Would moving the hardcoded ones to TLSv1_2 be a possibility?

For sure. I've actually been working through whether we can force a global TLSv1.2 minimum (so we support TLSv1.3 by default when it lands), and then we can force-downgrade the specific gateways that don't support it (I only know of one thus far). But this may still cause troubles, depending on the specifics of the troubles experienced.

duff:
If I remember correctly, it seemed to be attempting SSLv3 even though those gateways don't support it.

mutemule:
Yeah, I know that's why this went in originally, was to force an upgrade from SSLv3.

If Ruby <2.1.4 is still considered supported, what about porting in the same negotiation stuff from 2.1.4 into `Connection::configure_ssl`? That would get us to a known-good standard across the board, and then we can discuss bumping things a bit once that's settled?

Or would you still prefer to hard-code these all to TLSv12?

duff:
I'm definitely not an expert on this stuff (evidenced by my less than specific answers!)

We're using Ruby 2.1.7 and I don't know that the negotiation was working as anticipated.  Once we forced it from SSLv3, we were able to connect.

I'm unsure if I'm alone in this, but an across the board change scares me a bit since each gateway is so different and deploying a change where we have difficulty connecting to many gateway types would be sub-optimal.  That being said, my fear is likely caused by a lack of knowledge.

mutemule:
The only across-the-board change I'd suggest at this point is emulating the SSL negotiation introduced in Ruby 2.1.4, which means we would only support TLSv1, TLSv1.1, and TLSv1.2. Which seems like a reasonable set of protocols to support, given AM. Anything beyond that would likely be behind some configuration; so, allow a site to restrict that further, but keep operational defaults.

All this being said, I recognize that this is fine and dandy theoretically, but it's an extremely difficult thing to test in practice: we would need to confirm the `test` and `live` URLs across all gateways across a multitude of Ruby versions, or mark this as a potentially breaking change that should be introduced with care.

So long as we stick as close as possible to the default Ruby behaviour, I would be surprised to find any protocol negotiation bumps, so I'm pretty confident moving down that road. Anything beyond that is definitely considerably more risky, though.

One thing to note against hard-coding these gateways to TLSv1.2: when TLSv1.3 lands, and becomes the preferred protocol, we'll have to remember to come back in and fix these up again. Hard-coding the versions creates a manual maintenance burden that will always be surrounded by some confusion, a la "why did we do this in the first place, and can we safely change it?" -- much like what we're doing in this PR.

mutemule:
@duff: I've rebased, and added another commit that directly copies the pertinent options from Ruby's [ssl options](https://github.com/ruby/ruby/blob/v2_1_4/ext/openssl/lib/openssl/ssl.rb#L64-L66) as of 2.1.4. All this does is remove SSLv2 and SSLv3 support, and optionally remove compression if supported by the local OpenSSL.

I expect, long-term, this is going to be considerably less painful than hard-coding gateways to specific TLS versions, and considerably safer.

duff:
Sounds good @mutemule.  I think it's worth a shot.  If we see any issues we'll report back.  :shipit:

mutemule:
Looks like I'll need to revamp the approach a bit. I'll tackle it in the coming days and re-bump for review when ready.

bpollack:
It's been about 19 months since the last discussion here; should I close this out?

bpollack:
Closing for the moment, since the build has failed and it's been nearly two years since the last update, but please feel free to reopen in the future.

PR #2196, opened 2016-08-20
[In Progress] Stripe: Add L2 and L3 Data Support

timbeiko:
In Progress

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -485,6 +485,7 @@ module ActiveMerchant #:nodoc:
 
       def commit(method, url, parameters = nil, options = {})
         add_expand_parameters(parameters, options) if parameters
+        add_level_iii_data(parameters, options) if (method == :post) && (url == "charges")
         response = api_request(method, url, parameters, options)
 
         success = !response.key?("error")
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -505,6 +506,26 @@ def commit(method, url, parameters = nil, options = {})
...
+        post[:level3][:line_items] = [{
+            :product_code => 0
+        }]
+
+        # This one produces the following error:
+        # "Invalid keys: all first indices must be numeric, you provided product_code"
+        # post[:level3][:line_items] = {
+        #     :product_code => 0
+        # }
+      end
```
timbeiko: It seems that I'm unable to add the line items here. 

The format in the Stripe docs has an empty array in the middle and I can't get the format right
screen shot 2016-08-20 at 3 55 28 pm

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -505,6 +506,26 @@ def commit(method, url, parameters = nil, options = {})
         )
       end
 
+      def add_level_iii_data(post, options)
+        # Hard coded, no validations, just to see that it gets added
+        post[:level3] = {}
+        post[:level3][:merchant_reference] = 55823
```
timbeiko: These hardcoded values are placeholders

lib/active_merchant/billing/gateways/stripe.rb
timbeiko: From Stripe:
screen shot 2016-08-20 at 5 19 19 pm

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -621,6 +625,60 @@ def ach?(payment_method)
           card_brand(payment_method) == "check"
         end
       end
+
+    def encode_parameters(params)
```
timbeiko: From here: https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/util.rb#L117

timbeiko:
test

PR #2198, opened 2016-08-22
PayJunction(v2): Add new gateway

shasum:
Integrates with PayJunction's newer REST API.

### DIFF
--- a/lib/active_merchant/billing/gateways/pay_junction_v2.rb
+++ b/lib/active_merchant/billing/gateways/pay_junction_v2.rb
@@ -1,0 +1,190 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PayJunctionV2Gateway < Gateway
+      self.display_name = "PayJunction"
+      self.homepage_url = "https://www.payjunction.com/"
+
+      self.test_url = "https://api.payjunctionlabs.com/transactions"
+      self.live_url = "https://api.payjunction.com/transactions"
+
+      self.supported_countries = ["US"]
+      self.default_currency = "USD"
+      self.money_format = :dollars
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      def initialize(options={})
+        requires!(options, :api_login, :api_password, :api_key)
+        super
+      end
+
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/pay_junction_v2.rb
```diff
@@ -0,0 +1,224 @@
...
+
+      def void(authorization, options={})
+        post = {}
+        post[:status] = "VOID"
+        post[:transactionId] = authorization
+
+        commit("void", post)
+      end
+
+      def refund(money, authorization, options={})
```
davidsantoso: @shasum wondering about this `refund` method since it seems a little more complicated than a typical implementation. Of course there could be something I'm unaware of considering I didn't work on the remote tests like you did. 😄 

Is it not possible for us to skip doing a GET on a transaction for the `amountBase` and instead just pass the `money` parameter to the request? So, a more pared down implementation like-

```ruby
def refund(money, authorization, options={})
  post = {}
  post[:transactionId] = authorization
  post[:amountBase] = sprintf("%.2f", amount(money))
  commit("refund", post)
end
```

I'm thinking that we should leave it up to the method invoker to pass in the `money` param to set the refund amount?

davidsantoso:
@shasum let me know about the `refund` method above- If we can nix doing a GET on the transaction first, we may be able to simplify the `commit` and response `parse` further down the line as well!

lib/active_merchant/billing/gateways/pay_junction_v2.rb
shasum: @davidsantoso think I made a mistake in understanding PayJunction's `REFUND` transaction. thought it was allowing only AM `credit` style operation. re-reading the docs, looks like it supports both AM `credit` and AM `refund`

due to the misunderstanding I tried to accomplish a read-modify-write to implement AM `refund`. as in, read previously charged amount, reduce the amount to be refunded, and write back the new balance to be charged. but that seems a wrong approach now. thanks for spotting, let me rework this.

shasum:
Just a note to record that `refund` has been modified, GET removed and minor changes around generating AM response from gateway response.

PR #2209, opened 2016-09-01
Add US and CA as new supported countries

davidgf:


jasonwebster:
Hi! Thanks for the PR. Do you think you could also update the README? You can generate the list of gateways and their supported countries by using a rake task: `rake gateways:print:markdown`

alexandresaiz:
@jasonwebster hey Jason, we'll definitely do so !

alexandresaiz:
@jasonwebster hey Jason, we did it. When do you think this will be live?

PR #2213, opened 2016-09-09
Openpay new features

darkaz:
-Add merchant_account_id param to change dynamically between account
-Add metadata object to receive additional information for fraud. Allow up to 15 fields
-Add customer object to receive name, last name, external id and phone number 
-Add shipment address to customer.
-Refactor customer method.

### DIFF
--- a/lib/active_merchant/billing/gateways/openpay.rb
+++ b/lib/active_merchant/billing/gateways/openpay.rb
@@ -63,7 +63,7 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-     def store(creditcard, options = {})
+      def store(creditcard, options = {})
         card_params = {}
         add_creditcard(card_params, creditcard, options)
         card = card_params[:card]
--- a/lib/active_merchant/billing/gateways/openpay.rb
+++ b/lib/active_merchant/billing/gateways/openpay.rb
@@ -21,35 +21,39 @@ module ActiveMerchant #:nodoc:
         requires!(options, :key, :merchant_id)
         @api_key = options[:key]
         @merchant_id = options[:merchant_id]
+        @merchant_account_id = options[:merchant_account_id]
         super
       end
 
       def purchase(money, creditcard, options = {})
         post = create_post_for_auth_or_purchase(money, creditcard, options)
-        commit(:post, 'charges', post, options)
+        commit(:post, get_path('charges', options), post, options)
       end
 
       def authorize(money, creditcard, options = {})
         post = create_post_for_auth_or_purchase(money, creditcard, options)
         post[:capture] = false
-        commit(:post, 'charges', post, options)
+        commit(:post, get_path('charges', options), post, options)
       end
 
       def capture(money, authorization, options = {})
-        post = {}
+         post = {}
         post[:amount] = amount(money) if money
-        commit(:post, "charges/#{CGI.escape(authorization)}/capture", post, options)
+        resource = get_path("charges/#{CGI.escape(authorization)}/capture", options)
+        commit(:post, resource, post, options)
       end
 
       def void(identification, options = {})
-        commit(:post, "charges/#{CGI.escape(identification)}/refund", nil, options)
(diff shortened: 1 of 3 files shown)


davidsantoso:
@darkaz thanks for rebasing on `master`. However, the commits still need to be squashed into one single commit. Thanks!

davidsantoso:
@darkaz I'm reviewing now, however before I push I'll probably squash all your commits into one and rebase again since there have been commits since you reopened this PR. I'll let you know if I have questions. Thanks!

lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -109,14 +114,49 @@ def create_post_for_auth_or_purchase(money, creditcard, options)
...
+              if (customer_opt = options[:customer])
+                customer[:external_id] = customer_opt[:external_id] if customer_opt[:external_id]
+                customer[:name] = customer_opt[:name] if customer_opt[:name]
+                customer[:last_name] = customer_opt[:last_name] if customer_opt[:last_name]
+                customer[:phone_number] = customer_opt[:phone_number] if customer_opt[:phone_number]
+              else 
+                #Customer name is required
+                customer[:name] = options[:order_id] || email
+              end
+              add_shipment_address(customer, options)
```
duff: It would be great to see a remote test flexing this.

lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -109,14 +114,49 @@ def create_post_for_auth_or_purchase(money, creditcard, options)
...
 
+      def add_customer(post, options)
+          if !is_customer(options)
+            if(email = options[:email])
+              customer = {}
+              customer[:email] = email
+              if (customer_opt = options[:customer])
+                customer[:external_id] = customer_opt[:external_id] if customer_opt[:external_id]
+                customer[:name] = customer_opt[:name] if customer_opt[:name]
+                customer[:last_name] = customer_opt[:last_name] if customer_opt[:last_name]
```
duff: For the `name` and `last_name`, we typically use the name from the credit card rather than using part of the options hash.

lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -109,14 +114,49 @@ def create_post_for_auth_or_purchase(money, creditcard, options)
...
+                customer[:phone_number] = customer_opt[:phone_number] if customer_opt[:phone_number]
```
duff: We usually use `phone` of the billing address for this.  Examples:

https://github.com/activemerchant/active_merchant/blob/b2f62a950faab76d1e0417fd090602ae5ac3363b/lib/active_merchant/billing/gateways/authorize_net.rb#L490

https://github.com/activemerchant/active_merchant/blob/b2f62a950faab76d1e0417fd090602ae5ac3363b/lib/active_merchant/billing/gateways/mastercard.rb#L141

duff:
Thanks for the PR!  

The `metadata` feature looks pretty useful, as does the shipping address.

I'm thinking it would be great if we could simplify the `customer` handling some by lessening the nesting.  So we could support things like `customer_ external_id` on the `options` hash.  If the names and the phone number field get handled by my comments above, then I think `customer_external_id` is the only remaining customer field.

I'm not sure I understand the motivation behind being able to change the merchant_account_id on a per transaction basis rather than relying on the parameters passed into the initialize.  Can you help us understand the goal more?  I think it's somewhat of a convention in active merchant for the instance of the gateway to represent the merchant account.  If folks need to refer to another merchant_account, they typically create another instance.

darkaz:
@duff I did some changes based on your comments.

- Added a shipping address test.
- Name and LastName for customer are taking from card object
- Customer simplified. Added customer_external_id and customer_id params

merchant_account_id it will be used by common our customer to change dynamically between his different accounts using that same gateway instance.

Regards

davidsantoso:
Closing due to inactivity.

PR #2220, opened 2016-09-22
Payflow: Update supported countries

shasum:
@davidsantoso 

**ref:**
https://developer.paypal.com/docs/classic/products/payflow-gateway/
https://developer.paypal.com/docs/classic/howto_product_matrix/


davidsantoso:
@shasum thanks!

PR #2222, opened 2016-09-23
SecurionPay: Update supported countries

shasum:
@davidsantoso 

https://securionpay.com/faq/general-admission-boarding/

>We accept companies that are registered in Europe. Additionally, we accept entities from Israel, but only for Visa cards processing.


shasum:
@davidsantoso rebased n ready

PR #2229, opened 2016-09-30
Adds quick_update_customer method to USA ePay Advanced SOAP API

joshreeves:
Allows the updating of customer records on a field-by-field basis without needing to provide a complete customer object.

### DIFF
--- a/lib/active_merchant/billing/gateways/usa_epay_advanced.rb
+++ b/lib/active_merchant/billing/gateways/usa_epay_advanced.rb
@@ -77,12 +77,14 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'http://www.usaepay.com/'
       self.display_name = 'USA ePay Advanced SOAP Interface'
 
-      CUSTOMER_OPTIONS = {
+      CUSTOMER_PROFILE_OPTIONS = {
         :id => [:string, 'CustomerID'], # merchant assigned number
         :notes => [:string, 'Notes'],
         :data => [:string, 'CustomData'],
-        :url => [:string, 'URL'],
-        # Recurring Billing
+        :url => [:string, 'URL']
+      } #:nodoc:
+
+      CUSTOMER_RECURRING_BILLING_OPTIONS = {
         :enabled => [:boolean, 'Enabled'],
         :schedule => [:string, 'Schedule'],
         :number_left => [:integer, 'NumLeft'],
@@ -92,18 +94,24 @@ module ActiveMerchant #:nodoc:
         :user => [:string, 'User'],
         :source => [:string, 'Source'],
         :send_receipt => [:boolean, 'SendReceipt'],
-        :receipt_note => [:string, 'ReceiptNote'],
-        # Point of Sale
+        :receipt_note => [:string, 'ReceiptNote']
+      } #:nodoc:
+
+      CUSTOMER_POINT_OF_SALE_OPTIONS = {
         :price_tier => [:string, 'PriceTier'],
         :tax_class => [:string, 'TaxClass'],
         :lookup_code => [:string, 'LookupCode']
       } #:nodoc:
 
-      ADDRESS_OPTIONS = {
+      CUSTOMER_OPTIONS = [
+        CUSTOMER_PROFILE_OPTIONS,
+        CUSTOMER_RECURRING_BILLING_OPTIONS,
+        CUSTOMER_POINT_OF_SALE_OPTIONS
+      ].inject(:merge) #:nodoc:
+
+      COMMON_ADDRESS_OPTIONS = {
         :first_name => [:string, 'FirstName'],
         :last_name => [:string, 'LastName'],
-        :address1 => [:string, 'Street'],
-        :address2 => [:string, 'Street2'],
         :city => [:string, 'City'],
         :state => [:string, 'State'],
         :zip => [:string, 'Zip'],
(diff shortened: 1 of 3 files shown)


joshreeves:
@davidsantoso @jasonwebster - Hey guys. Would love some feedback on this. Not sure if there's some additional protocol I need to follow, but it would be great to get this merged so that I can switch back to using an un-forked version of AM.

davidsantoso:
@joshreeves thanks for the PR! 

> Not sure if there's some additional protocol I need to follow

We're working on getting a little more formal process in place, but for now a quick ping works! Do you happen to have a link to documentation on this? Just want to cross reference to feel confident.

joshreeves:
@davidsantoso - Thanks for the response. Here is the documentation for the method (https://wiki.usaepay.com/developer/soap/methods/quickupdatecustomer).

The reason this feature is critical for us is because the `updateCustomer` method, which AR currently only supports, requires a complete payment method at the time of updating. If a complete payment method is not present then the existing payment method is overwritten. Since it is impossible for us to retrieve a complete payment method from the USA ePay customer store, we would have to resort to an "all-or-nothing" approach to updating customer data. In other words, to update a customer's zip code, we need their credit card number. Using the `quick_update_customer` solves this problem for us.

davidsantoso:
@joshreeves thanks for the docs! A couple of thoughts for you-

1. I'd like to avoid creating a custom `CUSTOMER_UPDATE_DATA_FIELDS` option at the top if at all possible. Generally options mappings like these in Active Merchant tend to encapsulate common groupings of fields so similar to how you split out the profile, recurring, and point of sale fields. The usual pattern is to encapsulate those then add them as necessary to the request.
2. Likewise, I think that will remove the need of a custom `build_field_value_array` method since we can piece together to SOAP request body similar to how the [`add_customer_payment_method`](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/usa_epay_advanced.rb#L1022-L1030) is doing it.

Overall, this is certainly headed in the right direction! I think it could use a little more work re-using the existing builders/pattern the other request body builders are following.

joshreeves:
@davidsantoso - Sorry in advance for the long-winded response....

Thanks for the feedback. I think there might be some confusion (possibly on my side). If you look at the docs (https://wiki.usaepay.com/developer/soap/methods/quickupdatecustomer), you'll see under the "Arguments" heading a table of arguments that the `quickUpdateCustomer` accepts. One of them is an array of things called a `FieldValue`. There are a number of places in the USA ePay API where it accepts a `FieldValue`, however AM currently does not support any of them. My PR is adding in support for creating both a `FieldValue` and a `FieldValue` array (https://wiki.usaepay.com/developer/soap/objects/fieldvalue). My attempt was to add them in such a way that they could be used for other end-points if needed (for example, `build_customer_data` contains a `build_custom_fields` call which is currently marked as `TODO`. `CustomFields` on the `Customer` object is similarly a `FieldValue` array).

If you'd prefer to not have `build_field_value` and `build_field_value_array` as reusable methods for building those types of params, I can move that logic into the `build_quick_update_customer` method directly. However, as far as I can tell, `build_field_value_array` is really no different than `build_transaction_field_array` in terms of its purpose of being a reusable method for building an array of common params.

To demonstrate the reusability of what I've added, I could add to this PR support for the `CustomFields` property of the `build_customer_data` (currently marked `TODO`); however, my intention was to keep this PR focused for just the purpose of adding support for `quickUpdateCustomer`, while doing it in a way that would be reusable so that `CustomFields` could eventually benefit from it.

As far as the `CUSTOMER_UPDATE_DATA_FIELDS`, at some point I need to construct a single hash of all the options that the `FieldValue` array can accept. Are you saying you'd rather see something like this directly inside `build_quick_update_customer`:
```
def build_quick_update_customer(soap, options)
  soap.tag! "ns1:quickUpdateCustomer" do
    build_token soap, options
    build_tag soap, :integer, 'CustNum', options[:customer_number]
    fields = [
        CUSTOMER_PROFILE_OPTIONS,
        CUSTOMER_RECURRING_BILLING_OPTIONS,
        COMMON_ADDRESS_OPTIONS,
        EXTRA_ADDRESS_OPTIONS,
        GENERAL_PAYMENT_METHOD_OPTIONS
      ].inject(:merge)
    soap.tag! "UpdateData", 'SOAP-ENC:arryType' => "xsd:FieldValue[#{options.length}]", 'xsi:type' => "ns1:FieldValueArray" do
      options[:update_data].each do |k, v|
        build_field_value soap, fields[k][1], v, fields[k][0] if fields.keys.include? k
      end
    end
  end
end
```

davidsantoso:
@joshreeves I had missed the "Arguments" heading on the docs page so this whole PR makes a lot more sense now. My apologies and thanks for pointing it out/taking the time to really explain.

> If you'd prefer to not have build_field_value and build_field_value_array as reusable methods 

Lets keep these in. Reusable is good 👍 

> Are you saying you'd rather see something like this directly inside build_quick_update_customer

Originally yes, but looking back again at the adapter it seems like the transaction request options are building out their requests as well so I guess it's fine to add it at the top in a similar fashion. I think this adapter is on the older side and might not follow some of the more recent conventions.

I did have one last question just to be sure-

Looks like the `address1` and `address2` are separate from the `COMMON_ADDRESS_OPTIONS` and I was just curious as to why that might be?

joshreeves:
@davidsantoso - Awesome. Glad to hear it. We've been using this code in production since September and I'm eager to get back to using the AM repo.

So the reason for breaking out the `COMMON_ADDRESS_OPTIONS` is because in some cases ePay refers to those params as `address1, address2` and some times `street1, street2`. I assume this was a mistake on their part, but I actually attempted to combine them and it didn't work, so I *reluctantly* broke them out.

joshreeves:
You can see here they are `street1, street2` (https://wiki.usaepay.com/developer/soap/objects/address)

And here they are `address1, address2` (https://wiki.usaepay.com/developer/soap/methods/quickupdatecustomer)

davidsantoso:
@joshreeves sorry for the delay- that makes sense. Thanks for clarifying. If you could just rebase the commit on `master` I'd be happy to merge in.

joshreeves:
@davidsantoso - Rebased and tested in our app. Thanks for working with me on this!

davidsantoso:
@joshreeves thanks for the rebase! and for answering all of my questions 😸 Merging.

davidsantoso:
@joshreeves merged. Thanks for contributing to Active Merchant!

PR #2231, opened 2016-10-01
update payment_express payment urls

oklas:
The payment url for testing is not equal production url.
I do not found proof link for pxpost interface, but
that instruction we received in welcome email.
I found such instruction for pxpay interface, it is not proof but it is:
https://www.paymentexpress.com/developer-e-commerce-paymentexpress-hosted-pxpay
Also I have test this, and it is worked

### DIFF
--- a/lib/active_merchant/billing/gateways/payment_express.rb
+++ b/lib/active_merchant/billing/gateways/payment_express.rb
@@ -21,7 +21,8 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'http://www.paymentexpress.com/'
       self.display_name = 'PaymentExpress'
 
-      self.live_url = self.test_url = 'https://sec.paymentexpress.com/pxpost.aspx'
+      self.live_url = 'https://sec.paymentexpress.com/pxpost.aspx'
+      self.test_url = 'https://uat.paymentexpress.com/pxpost.aspx'
 
       APPROVED = '1'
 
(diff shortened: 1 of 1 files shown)


bpollack:
Thanks for the PR! While this is technically correct after a fashion, the URLs are identical according both to the docs and your PR, so I don't think this actually fixes anything. For that reason, I'm going to close out this PR. Please let me know if I've misunderstood anything.

oklas:
Hi @bpollack, thanks for attention to this pr. Actually the original code has a bug wich appears in test environment.

The first bug is in the [line 24](https://github.com/activemerchant/active_merchant/pull/2231/files#diff-e5cbc93d3b47452501244f371363d541L24) where both test and production variables (`test_url` and `live_url`) is initilaizing by merchant production url - this is not rigth, *for test variable must be used test url*. And the second bug is that test variable does not used anywhere when it must be used as gateway url in test environment at [line 296](https://github.com/activemerchant/active_merchant/pull/2231/files#diff-e5cbc93d3b47452501244f371363d541L296).

The documentation which declare both test and production url on provider site is [here](https://www.paymentexpress.com/developer-e-commerce-paymentexpress-hosted-pxpay). As you noticed the production url it same as in origin code. But `test` environment implementation is wrong. This pr is a fix.

This will be useful for future users wich wants to integrate this merchant and for systems which wanted to be actual and tested each time before deploy. Thanks.

bpollack:
Ah, I see; I misread the URLs @oklas . I'll reopen and try to land this today.

PR #2232, opened 2016-10-03
Stripe: Set minimum authorize amount depending on currency

davidsantoso:
Previously the Stripe `verify` action was updated to default to authorizing an amount of $1 USD. Increasing the authorize amount and forcing the USD currency was done primarily to overcome currency exchange rate issues for the required minimum charge amount. See documentation below for more details:

https://support.stripe.com/questions/what-is-the-minimum-amount-i-can-charge-with-stripe

However, some Stripe accounts do not support the USD currency so forcing that during `verify` caused those accounts to break. This change allows for the authorize amount during verify to be dynamically set depending on the currency passed in, defaulting to `50` if none is given. It also doubles the minimum amounts shown in the Stripe documentation in an effort to keep currency exchange rates from becoming an issue when running the authorize in non USD.

@duff any other thoughts?

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -51,6 +51,22 @@ module ActiveMerchant #:nodoc:
         "business" => "company",
       }
 
+      MINIMUM_AUTHORIZE_AMOUNTS = {
+        "USD" => 100,
+        "CAD" => 100,
+        "GBP" => 60,
+        "EUR" => 100,
+        "DKK" => 500,
+        "NOK" => 600,
+        "SEK" => 600,
+        "CHF" => 100,
+        "AUD" => 100,
+        "JPY" => 100,
+        "MXN" => 2000,
+        "SGD" => 100,
+        "HKD" => 800
+      }
+
       def initialize(options = {})
         requires!(options, :login)
         @api_key = options[:login]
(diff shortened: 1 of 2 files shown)


duff:
@davidsantoso Looks great!  Only question is what will happen if they specify an actual `currency_code` but it's not in the hash.  Perhaps it should default to 100 in that case?

davidsantoso:
@duff ah! Very good catch. Let me add that real quick.

davidsantoso:
@duff updated. Now returns `100` if the currency passed in is not in the hash.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -600,6 +616,12 @@ def ach?(payment_method)
           card_brand(payment_method) == "check"
         end
       end
+
+      def auth_minimum_amount(options)
+        return 100 unless options[:currency]
+        return 100 unless MINIMUM_AUTHORIZE_AMOUNTS.has_key?(options[:currency].upcase)
+        return MINIMUM_AUTHORIZE_AMOUNTS[options[:currency].upcase]
+      end
```
duff: Little tweak to simplify?

```
def auth_minimum_amount(options)
  return 100 unless options[:currency]
  MINIMUM_AUTHORIZE_AMOUNTS[options[:currency].upcase] || 100
end
```

Not a biggie though.  😄 

ship_little_-_google_search

davidsantoso: Ha no, that's definitely nicer 😄 Thanks!

PR #2233, opened 2016-10-04
Element: Pass order_id and shipping address

curiousepic:

### DIFF
--- a/test/remote/gateways/remote_element_test.rb
+++ b/test/remote/gateways/remote_element_test.rb
@@ -8,6 +8,7 @@ class RemoteElementTest < Test::Unit::TestCase
     @credit_card = credit_card('4000100011112224')
     @check = check
     @options = {
+      order_id: '1',
       billing_address: address,
       description: 'Store Purchase'
     }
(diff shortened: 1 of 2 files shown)


test/remote/gateways/remote_element_test.rb
```diff
@@ -41,6 +42,12 @@ def test_successful_purchase_with_payment_account_token
     assert_equal 'Approved', response.message
   end
 
+  def test_successful_purchase_with_shipping_address
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(shipping_address: address(address1: "Shipping")))
+    assert_success response
+    assert_equal 'Approved', response.message
+  end
```
curiousepic: @davidsantoso In looking into that, I realized the shipping fields needed to go in the same element as the billing address fields. But apparently even after that fix, no shipping fields are returned in a response. Now I'm not sure if that's intended or if it's still not being passed correctly. Investigating...

curiousepic: @davidsantoso I think that's the expected behavior; in fact it only returns the first line and zipcode of the billing address in the response as well. PR is updated with the address element fix.

curiousepic:
Travis failure is `Error: test_avs_enabled_and_provided(MonerisUsTest): ActiveMerchant::ResponseError: Failed with 503 Service Temporarily Unavailable` Is there a way to re-run the Travis tests or should I resubmit the PR?

davidsantoso:
@curiousepic I restarted the Travis build so hopefully we don't get a timeout this time!

davidsantoso:
@curiousepic also real quick- mind rebasing on `master`? I think it might be a commit or two behind.

PR #2238, opened 2016-10-14
Only get alpha2 country code when it's a known country

bruno:
Fixes the case when the country code lookup in Cybersource would fail
(for example when using `Faker::Address.country`). Only attempt to call
`.code(:alpha2)` when `country_code` is not `nil`.


jasonwebster:
Closed #2238 via a0b3ac284ff06285ec3bf36681b8a237ab340ab4

PR #2241, opened 2016-10-21
Braintree supports network tokenization

elfassy:
Verify that braintree gateways supports network tokenization, similar to what was done for auth.net (https://github.com/activemerchant/active_merchant/pull/1766)

ping @Krystosterone @duff @lyverovski @jnormore

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -185,7 +185,7 @@ module ActiveMerchant
 
       def supports_network_tokenization?
         request = post_data(:authorize) do |xml|
-          add_auth_purchase(xml, "authOnlyTransaction", 1, Billing::NetworkTokenizationCreditCard.test_credit_card, {})
+          add_auth_purchase(xml, "authOnlyTransaction", 1, NetworkTokenizationCreditCard.test_credit_card, {})
         end
         raw_response = ssl_post(url, request, headers)
         response = parse(:authorize, raw_response)
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -184,18 +184,8 @@ module ActiveMerchant
       end
 
       def supports_network_tokenization?
-        card = Billing::NetworkTokenizationCreditCard.new({
-          :number => "4111111111111111",
-          :month => 12,
-          :year => 20,
-          :first_name => 'John',
-          :last_name => 'Smith',
-          :brand => 'visa',
-          :payment_cryptogram => 'EHuWW9PiBkWvqE5juRwDzAUFBAk='
-        })
-
         request = post_data(:authorize) do |xml|
-          add_auth_purchase(xml, "authOnlyTransaction", 1, card, {})
+          add_auth_purchase(xml, "authOnlyTransaction", 1, Billing::NetworkTokenizationCreditCard.test_credit_card, {})
         end
         raw_response = ssl_post(url, request, headers)
         response = parse(:authorize, raw_response)
--- a/lib/active_merchant/billing/gateways/braintree/braintree_common.rb
+++ b/lib/active_merchant/billing/gateways/braintree/braintree_common.rb
@@ -1,4 +1,9 @@
 module BraintreeCommon
+
+  ERROR_CODES = {
+    payment_instrument_not_supported: '91577'
+  }.freeze
+
   def self.included(base)
     base.supported_countries = %w(US CA AD AT BE BG HR CY CZ DK EE FI FR GI DE GR GG HU IS IM IE IT JE LV LI LT LU MT MC NL NO PL PT RO SM SK SI ES SE CH TR GB SG HK MY AU NZ)
     base.supported_cardtypes = [:visa, :master, :american_express, :discover, :jcb, :diners_club]
@@ -17,5 +22,10 @@ module BraintreeCommon
       gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
       gsub(%r((&?ccnumber=)\d*(&?)), '\1[FILTERED]\2').
       gsub(%r((&?cvv=)\d*(&?)), '\1[FILTERED]\2')
+  end
+
+  def supports_network_tokenization?
+    response = authorize(1, ActiveMerchant::Billing::NetworkTokenizationCreditCard.test_credit_card)
+    response.params['braintree_transaction']['processor_response_code'] != ERROR_CODES.fetch(:payment_instrument_not_supported)
   end
 end
--- a/lib/active_merchant/billing/gateways/braintree/braintree_common.rb
+++ b/lib/active_merchant/billing/gateways/braintree/braintree_common.rb
@@ -26,9 +26,9 @@ module BraintreeCommon
 
   def supports_network_tokenization?
     response = authorize(1, ActiveMerchant::Billing::NetworkTokenizationCreditCard.test_credit_card)
-    response_code = response.params['braintree_transaction'].try!(:[], 'processor_response_code')
+    response_code = response.params['braintree_transaction'].try(:[], 'processor_response_code')
     raise(ActiveMerchant::InvalidResponseError, response.message) unless response_code
 
-    response.params['braintree_transaction']['processor_response_code'] != ERROR_CODES.fetch(:payment_instrument_not_supported)
+    response_code != ERROR_CODES.fetch(:payment_instrument_not_supported)
   end
 end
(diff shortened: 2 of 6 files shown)


lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -184,18 +184,8 @@ def scrub(transcript)
...
-          :year => 20,
-          :first_name => 'John',
-          :last_name => 'Smith',
-          :brand => 'visa',
-          :payment_cryptogram => 'EHuWW9PiBkWvqE5juRwDzAUFBAk='
-        })
-
         request = post_data(:authorize) do |xml|
-          add_auth_purchase(xml, "authOnlyTransaction", 1, card, {})
+          add_auth_purchase(xml, "authOnlyTransaction", 1, Billing::NetworkTokenizationCreditCard.test_credit_card, {})
```
Krystosterone: Why the change?

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -171,10 +171,6 @@ def unstore(customer_vault_id, options = {})
       end
       alias_method :delete, :unstore
 
-      def supports_network_tokenization?
```
Krystosterone: I'm curious to know why this was _always_ true. Does `Braintree Blue` always support network tokenization?

elfassy: good question, @lyverovski should we leave this to true on `Braintree Blue` but make it conditional on other `Braintree`?

elfassy: confirmed with Andre, `Braintree Blue` might not always support network tokenization so we'll need to verify it.

lyverovski: It was always defaulted to true because there was no way to verify it before. There still isn't an "official" way, but the guidance from Braintree was to check for the specific error response coming back when attempting a transaction with a network tokenized card & cryptogram.

elfassy:
@duff how does this PR look to you? Thanks

lib/active_merchant/billing/network_tokenization_credit_card.rb
```diff
@@ -34,6 +34,18 @@ def credit_card?
       def type
         "network_tokenization"
       end
+
+      def self.test_credit_card
```
jasonwebster: nit: class methods at the top of the definition (above instance methods) please.

lib/active_merchant/billing/gateways/authorize_net.rb
jasonwebster: nit: `NetworkTokenizationCreditCard` should be enough. `Billing` module is already implied and will be checked first due to resolution rules and nesting.

lib/active_merchant/billing/gateways/braintree/braintree_common.rb
```diff
@@ -18,4 +23,9 @@ def scrub(transcript)
       gsub(%r((&?ccnumber=)\d*(&?)), '\1[FILTERED]\2').
       gsub(%r((&?cvv=)\d*(&?)), '\1[FILTERED]\2')
   end
+
+  def supports_network_tokenization?
+    response = authorize(1, ActiveMerchant::Billing::NetworkTokenizationCreditCard.test_credit_card)
+    response.params['braintree_transaction']['processor_response_code'] != ERROR_CODES.fetch(:payment_instrument_not_supported)
```
jasonwebster: What happens in case of a connection error or 5xy response and this body isn't there? It is appropriate that we let that bubble up (probably, yes)? I'm assuming an error will be raised before we get to the hash traversal, yeah?

Krystosterone:
(( 📟  ))

elfassy:
@jasonwebster I've updated the code to address your comments, let me know how this looks to you now

lib/active_merchant/billing/gateways/braintree/braintree_common.rb
```diff
@@ -26,6 +26,9 @@ def scrub(transcript)
 
   def supports_network_tokenization?
     response = authorize(1, ActiveMerchant::Billing::NetworkTokenizationCreditCard.test_credit_card)
+    response_code = response.params['braintree_transaction'].try!(:[], 'processor_response_code')
+    raise(ActiveMerchant::InvalidResponseError, response.message) unless response_code
```
jasonwebster: So, was this actually necessary?

My original questions were: 

> What happens in case of a connection error or 5xy response and this body isn't there? It is appropriate that we let that bubble up (probably, yes)? I'm assuming a request error will be raised before we get to the hash traversal, yeah?

None of which were really answered.

This is also the incorrect error to raise here. It's only ever used elsewhere in Active Merchant in case of accept-encoding errors, or actually malformed HTTP.

elfassy: > What happens in case of a connection error or 5xy response and this body isn't there?

The response object is created wether or not it is successful (https://github.com/activemerchant/active_merchant/blob/98bdad6d1fe3140df43228814abcbf2429791937/lib/active_merchant/billing/gateways/braintree_blue.rb#L420). It will not raise which will cause errors where we are expecting specific params.

> It is appropriate that we let that bubble up (probably, yes)?

Yes, as far as I can tell

> I'm assuming a request error will be raised before we get to the hash traversal, yeah?
> So, was this actually necessary?

Yes, as far as I can tell, there's nothing raising when the auth fails. It just returns the response with the error code. 

> This is also the incorrect error to raise here

The params we're getting back don't match our expectation. I agree it's probably not the same expectation as other exception using this error. How about a new error type `UnexpectedResponseError`

lib/active_merchant/billing/gateways/braintree/braintree_common.rb
```diff
@@ -18,4 +23,12 @@ def scrub(transcript)
       gsub(%r((&?ccnumber=)\d*(&?)), '\1[FILTERED]\2').
       gsub(%r((&?cvv=)\d*(&?)), '\1[FILTERED]\2')
   end
+
+  def supports_network_tokenization?
+    response = authorize(1, ActiveMerchant::Billing::NetworkTokenizationCreditCard.test_credit_card)
+    response_code = response.params['braintree_transaction'].try(:[], 'processor_response_code')
+    raise(ActiveMerchant::InvalidResponseError, response.message) unless response_code
+
+    response.params['braintree_transaction']['processor_response_code'] != ERROR_CODES.fetch(:payment_instrument_not_supported)
```
Krystosterone: Why not re-use `response_code` here? Also, isn't it inconsistent to raise an error here, but in the Authorize.net one (https://github.com/activemerchant/active_merchant/pull/1766/files#diff-32b8ad7eb3755b05bb1a03b6e1f9a3cdR194), we're not?

elfassy: >  isn't it inconsistent to raise an error here, but in the Authorize.net one 

Authorize.net already raises here (https://github.com/activemerchant/active_merchant/blob/9b40c0b1d7feb20d3330b78e7c284d1e8083e0d7/lib/active_merchant/posts_data.rb#L79)

elfassy: > it's in a module that is used in this gateway as well 

Braintree does not use `ssl_post`, it uses the braintree gem here https://github.com/braintree/braintree_ruby/blob/02db6f8d75830aa3338d61985183aab4eebfab9b/lib/braintree/http.rb#L32 instead. 

I think it's ok to let this raise via hash fetch if the Authorize.net implementation isn't consistent anymore.

chrisboulton:
With Braintree, they also support a `provision_raw_apple_pay` call, which enables  Pay on the Braintree account and also returns a list of supported card brands. (https://github.com/braintree/braintree_ruby/blob/02db6f8d75830aa3338d61985183aab4eebfab9b/lib/braintree/merchant_gateway.rb#L8). Is it worth calling this instead of attempting a test transaction? Or is it worth doing both?

I feel like this is probably a good start, but longer term, is it worthwhile evolving the network tokenization checks, into something where a list of card brands/payment instruments supported is returned? If that list is empty, tokenization isn't supported.

Things are somewhat problematic in this space right now. A merchant might be using a PSP with their gateway that only supports tokenization on specific card brands - for example, if they're using the AMEX processing network, AMEX only supports tokenization for AMEX cards. Right now there's an assumption in Active Merchant that tokenization is supported if a test against a VISA bin returns successfully.

With Braintree this kind of check is simple, because `provision_raw_apple_pay` gives you that list. It's accurate from what I can tell. For other gateways such as Auth.net, I'm not aware of any API call that returns such information, so some kind of test transaction might be the only option for those.

jasonwebster:
Thanks @chrisboulton, that's extraordinarily helpful. I think generally, where possible, we should move towards a pattern of asking the gateway what they do  indeed support for the current merchant's account. Looks like that's possible here, and what we should be doing.

elfassy:
closing for now in favour of https://github.com/activemerchant/active_merchant/pull/2275

PR #2242, opened 2016-10-23
Adding support for Square's ecommerce payments API and customer card on file

jawspeak:
This implements the following features for Squareup.com:

```
      def purchase(money, card_nonce, options={})
      def authorize(money, card_nonce, options={})
      def capture(ignored_money, txn_id, ignored_options={})
      def refund(money, txn_id, options={})
      def void(txn_id, options={})
      def verify(card_nonce, options={})
      def store(card_nonce, options = {})
      def update(customer_id, card_id, options = {})
      def update_customer(customer_id, options = {})
      def unstore(card_id, options = {}, deprecated_options = {})
```

Note it does NOT accept raw PANs because square's payment integration requires an iframe based tokenization approach. This is documented here: read more in https://docs.connect.squareup.com/articles/ecommerce-overview/

Remote tests pass, but the credentials are not shareable. Results here:

```
Loaded suite test/remote/gateways/remote_square_test
Started
...........................

Finished in 167.627982 seconds.
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
27 tests, 175 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
0.16 tests/s, 1.04 assertions/s
```

Please let me know if you need anything to improve this so it can be merged in. Thanks!

### DIFF
--- a/lib/active_merchant/billing/gateways/square.rb
+++ b/lib/active_merchant/billing/gateways/square.rb
@@ -1,0 +1,396 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class SquareGateway < Gateway
+
+      # test uses a 'sandbox' access token, same URL
+      self.live_url = 'https://connect.squareup.com/v2/'
+
+      self.money_format = :cents
+      self.supported_countries = ['US', 'CA']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb, :maestro]
+
+      self.homepage_url = 'https://squareup.com/developers'
+      self.display_name = 'Square'
+
+      # Map to Square's error codes:
+      # https://docs.connect.squareup.com/api/connect/v2/#handlingerrors
+      STANDARD_ERROR_CODE_MAPPING = {
+        'INVALID_CARD' => STANDARD_ERROR_CODE[:invalid_number],
+        'INVALID_EXPIRATION' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        'INVALID_EXPIRATION_DATE' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        'INVALID_EXPIRATION_YEAR' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+
+        # Something invalid in the card, e.g. verify declined when linking card to customer.
+        'INVALID_CARD_DATA' => STANDARD_ERROR_CODE[:processing_error],
+        'CARD_EXPIRED' => STANDARD_ERROR_CODE[:expired_card],
+        'VERIFY_CVV_FAILURE' => STANDARD_ERROR_CODE[:incorrect_cvc],
+        'VERIFY_AVS_FAILURE' => STANDARD_ERROR_CODE[:incorrect_zip],
+        'CARD_DECLINED' => STANDARD_ERROR_CODE[:card_declined],
+        'UNAUTHORIZED' => STANDARD_ERROR_CODE[:config_error]
+      }
+
+      # The `login` key is the client_id (also known as application id) 
+      #   in the dev portal. Get it after you create a new app:
+      #   https://connect.squareup.com/apps/
+      # The `password` is the access token (personal or OAuth)
+      # The `location_id` must be fetched initially 
+      #   https://docs.connect.squareup.com/articles/processing-payment-rest/
+      # The `test` indicates if these credentials are for sandbox or 
+      #   production (money moving) access
+      def initialize(options={})
+        requires!(options, :login, :password, :location_id, :test)
+        @client_id = options[:login].strip
+        @bearer_token = options[:password].strip
+        @location_id = options[:location_id].strip
+
+        super
+      end
+
+      # To create a charge on a card using a card nonce:
+      #     purchase(money, card_nonce, { ...create transaction options... })
+      #
+      # To create a customer and save a card (via card_nonce) to the customer:
+      #     purchase(money, card_nonce, {customer: {...params hash same as in store() method...}, ...})
+      #   Note for US and CA, you must have {customer: {billing_address: {zip: 12345}}} which passes AVS to store a card.
+      #   Note this always creates a new customer, so it may make a duplicate 
+      #   customer if this card was associated to another customer previously.
+      #
+      # To use a customer's card on file:
+      #     purchase(money, nil, {customer: {id: 'customer-id', card_id: 'card-id'}})
+      # Note this does not update any fields on the customer.
+      #
+      # To use a customer, and link a new card to the customer:
+      #     purchase(money, card_nonce, {customer: {id: 'customer-id', billing_address: {zip: 12345}})
+      # Note the zip is required to store the new nonce, and it must pass AVS.
+      # Note this does not update any other fields on the customer.
+      #
+      # As this may make multiple requests, it returns a MultiResponse.
+      def purchase(money, card_nonce, options={})
+        raise ArgumentError('money required') if money.nil?
+        if card_nonce.nil?
+          requires!(options, :customer)
+          requires!(options[:customer], :card_id, :id)
+        end
+        if card_nonce && options[:customer] && options[:customer][:card_id]
+          raise ArgumentError('Cannot call with both card_nonce and' +
+            ' options[:customer][:card_id], choose one.')
+        end
+
+        post = options.slice(:buyer_email_address, :delay_capture, :note,
+            :reference_id)
+        add_idempotency_key(post, options)
+        add_amount(post, money, options)
+        add_address(post, options)
+        post[:reference_id] = options[:order_id] if options[:order_id]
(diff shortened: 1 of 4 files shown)


jawspeak:
FYI @sn0storm this is the branch I mentioned, I'd love your feedback on this!

jawspeak:
@rwdaigle or @shasum is there anything I can do to help?

rwdaigle:
Hi @jawspeak - thanks for this! I'm not sure if there's precedent for merging in a gateway that doesn't support a direct PAN API, but it's not something I've personally done before. Do you have anymore relevant context for this being in ActiveMerchant?

jawspeak:
Great question @rwdaigle, this is my first PR but here are some things you may please consider:

- This is the same model as used in stripe's integration, but they also accept a raw pan submission in addition to a token that comes back from their stripe.js tokenizer. (Docs: https://stripe.com/docs/api#create_charge-source, impl: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L408)

- There is a large Square user base that want to use ActiveMerchant for their ecommerce sites. I work at Square and we have some large partner sites that want to offer Square to their merchants but their codebase is all using ActiveMerchant and it would take 6+ months to prioritize the raw square integration vs using it in ActiveMerchant. This would help them adopt faster.

- (The biggest reason) - Square's non-PAN API lets merchants avoid all PCI compliance since they do not handle payments. With EMV growing it is harder to use cloned cards and more card fraud is moving online, also many data breaches have made merchants very concerned w/ card harvesting. (See this recent list of skimming sites https://www.reddit.com/r/netsec/comments/56xou4/another_5900_stores_found_skimming_full_list)

rwdaigle:
@ntalbott @jnormore @duff any thoughts on this?

@jawspeak I can definitely see why a customer would want to use AM w/ Square. I think I'm wrestling with where the line for AM stops. It's not a general purpose gateway library, it's serving a very specific niche and I want to protect it from becoming a catch all for any type of gateway interaction.

Perhaps I'm being too protective, though, and the really old-timers can weigh in.

jnormore:
I'm not against it, we handle non PAN payments already (gateway tokens, network tokens, etc) so it wouldn't be too much of a deviation from the current behavior. I also don't necessarily see AM as a PAN only abstraction, it's interface defines general payment interactions (in whatever form that could be).

cc @lyverovski

duff:
My feeling is that Active Merchant is for gateway API's that allow you to pass them a PAN.  It's not for transparent redirects, iframe things, offsite gateways, etc.  I think a separate gem that depends on active merchant could add support for Square if folks want that.  There's value in keeping Active Merchant as focused as we can.

duff:
Perhaps one litmus test could be, does the gateway support the [simple usage example](https://github.com/activemerchant/active_merchant#usage)?

rwdaigle:
I understand from @ntalbott that we've drawn the line in the past here - with only supporting a direct PAN interface (which does include follow-on token workflows, just not in isolation). 

Of course, you're free to maintain your own fork or provide a Square patch if it's worth the effort. Sorry we couldn't pull this into the main project for you!

jawspeak:
Thanks @jnormore, @rwdaigle, and @duff for the feedback and comments so far! 

While we (I work at Square) can't comment about future products, would you be comfortable if the Square gateway accepted direct PANs? For instance if that was an option sophisticated merchants could enable, would adding direct PAN support to this PR be better?

Or, would this better make sense as an "Integration"? https://github.com/activemerchant/active_merchant/wiki/GatewaysVsIntegrations Note: it seems that has moved to the `offsite_payments` gem: https://github.com/activemerchant/offsite_payments/tree/master/lib/offsite_payments/integrations

The area I'm wondering about is we do support all of these APIs, however we offer the benefit preventing PCI compliance from merchants, since they use the token for the PAN instead of the PAN directly. 

Supported:

```
 def purchase(money, card_nonce, options={})
      def authorize(money, card_nonce, options={})
      def capture(ignored_money, txn_id, ignored_options={})
      def refund(money, txn_id, options={})
      def void(txn_id, options={})
      def verify(card_nonce, options={})
      def store(card_nonce, options = {})
      def update(customer_id, card_id, options = {})
      def update_customer(customer_id, options = {})
      def unstore(card_id, options = {}, deprecated_options = {})
```

However, instead of raw, sensitive PAN, it uses a tokenized PAN. 

So this simple usage example would be included:

```ruby
require 'active_merchant'

# Use the Square sandbox
ActiveMerchant::Billing::Base.mode = :test

gateway = ActiveMerchant::Billing::SquareGateway.new(
  login: 'sandbox-sq0idp-Wq3...'
  password: 'sandbox-sq0atb-DC-...'
  location_id: 'CBxx...')

# ActiveMerchant accepts all amounts as Integer values in cents
amount = 1000  # $10.00
card = 'tokenized-pan'

# This purchase call could also be:
#     authorize, for a later delayed capture
#     capture, within 6.5 days of an auth
#     refund, to refund fully or partially a previous purchase or capture
#     void, for an authorized but not yet captured
#     store, to store a card on file with a customer_id
#     unstore, to de-link the card-vaulted card from the customer
#     update, to update the customer's vaulted card on file
response = gateway.purchase(amount, card)
if response.success?
  puts "Successfully charged $#{sprintf("%.2f", amount / 100)} to the credit card token #{card}"
else
  raise StandardError, response.message
end
```

The standard [universal API](https://github.com/activemerchant/offsite_payments/blob/master/lib/offsite_payments/integrations/universal.rb) for "Integrations" doesn't seem to quite fit the above. 

Thoughts? Thanks (:

rwdaigle:
Hi @jawspeak!

> would you be comfortable if the Square gateway accepted direct PANs

Yes! This is directly inline with the core responsibilities of Active Merchant (even if it requires a corresponding switch on your side to enable PAN submission mode for interested customers).

lyverovski:
I'm going to agree with @jnormore on this one. This is similar to the Stripe gateway token behaviour, minus the acceptance of raw PANs. The interfaces aren't changing, and there is still the ability to perform auth/capture/refund/void transactions, which is really what defines a "direct" gateway vs. an "integration".

If Square is comfortable adding raw PAN support to their integration in order to conform more closely to the definition of a direct gateway, then that's fine, but I don't think it should be necessary.

rwdaigle:
@lyverovski we should get on the same page regarding the responsibilities of ActiveMerchant (perhaps offline), so we can provide consistent guidance to contributors.

@jawspeak if you're looking for the most expedient route while we get our story straight, feel free to go the direct PAN (and tokenized card) route :)

jawspeak:
Thanks for the feedback. Since we want to encourage all merchants to not need to worry about PCI compliance by offloading that responsibility to Square, yet still provide all the functionality of `auth`, `capture`, `purchase`, `refund`, `void`, plus recurring billing / stored card on file that a traditional direct PAN gateway has - hopefully the tokenized APIs will be something you all feel comfortable with adding :) 

If not, I'm going to need to continue working with people on our side as accepting raw PANs is something that we're discouraging merchants to have to think about. Hope we can make a bit more headway before the holidays, although I understand this is an important and thoughtful consideration that may take some time. cheers!

rwdaigle:
Hi @jawspeak!

Some of the AM maintainers got together in a cold, dark, Canadien bunker to try and come up with some general-purpose guidance that might help us decide when to accept what functionality, based on the goals of the project. Here's what we came up with:

> ActiveMerchant is focused on providing a set of transaction processing primitives that make it easy to transact against multiple gateways in a consistent fashion.

As it relates to this PR, gateway adapters that **only** provide a token-based transaction flow are necessarily breaking the gateway agnosticism ActiveMerchant provides since the only way that token can be created is through some other ancillary gateway-specific flow. In the interest of ensuring ActiveMerchant remains a focused and purposeful library, we don't think it makes sense to include contributions that don't support the primary transaction primitives of a multiple gateway environment.

I hope this adds clarity to why the functionality in this PR, while technically sound, doesn't quite match with the project's goals. If the adapter's functionality can be expanded to include a direct PAN interface (thus maintaining AM's goal of gateway agnosticism and portability), we'd be happy to include it.

Thanks so much for your patience while we figured out our position here!

jawspeak:
Thanks @rwdaigle I apologize for my delay in following up on this. I do appreciate the feedback and thoughtful response. We'll certainly keep this in mind at the point a direct PAN interface is available from square.

alisyed:
@jawspeak thank you - even though it didnt make it into AM, we're finding this very useful. I wish we had known about AM's mission beforehand so we could have gone about building our own abstraction layer to ensure our logic isnt filled with ifs, buts and elses to cater to different payment gateways!

frojasg:
we published a gem called [active_merchant_square](https://github.com/square/active_merchant_square) to extend active_merchant to include Square API

jawspeak:
Here's the details for people that reach this declined PR and want to use Active Merchant with Square: https://medium.com/square-corner-blog/activemerchantsquare-for-squares-e-commerce-api-111ea63e530a

PR #2245, opened 2016-10-28
Add void method to Conekta Gateway

MauricioMurga:
This PR adds the void endpoint to the Conekta Gateway.

### DIFF
--- a/lib/active_merchant/billing/gateways/conekta.rb
+++ b/lib/active_merchant/billing/gateways/conekta.rb
@@ -12,7 +12,7 @@ module ActiveMerchant #:nodoc:
 
       def initialize(options = {})
         requires!(options, :key)
-        options[:version] ||= '0.3.0'
+        options[:version] ||= '1.0.0'
         super
       end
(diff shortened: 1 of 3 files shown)


davidsantoso:
@MauricioMurga thanks for the PR! Just added one inline question about the version number. Also, we've committed on `master` so could you rebase and update your PR? Thanks!

lib/active_merchant/billing/gateways/conekta.rb
```diff
@@ -12,7 +12,7 @@ class ConektaGateway < Gateway
 
       def initialize(options = {})
         requires!(options, :key)
-        options[:version] ||= '0.3.0'
+        options[:version] ||= '1.0.0'
```
MauricioMurga: It's not necessary, we just updated to the most recent version.

MauricioMurga: I've already updated the PR as requested! Let me know if you need anything else!

PR #2246, opened 2016-10-28
Build request from description field

MauricioMurga:
This PR allows that a complete request to be built parsing the description field as JSON.


jasonwebster:
Hi! Thanks for the PR. Can you explain the use case for this? Why is it necessary for this gateway to have an alternative API for building requests that does not conform to the Active Merchant standard? Doesn't that kind of defeat the point of abstraction?

jasonwebster:
Feel free to re-open once my questions and concerns above have been addressed.

Issue #2247, opened 2016-10-29
OrbitalGateway#capture does not recognize order_id parameter


yaychris:
I'm going through the Orbital certification process and my "capture" test cases were rejected because I did not provide the `OrderID`.

In trying to do so, I discovered that ActiveMerchant [does not allow the `order_id` parameter for capture transactions](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/orbital.rb#L589). By comparison, void transactions [do allow it](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/orbital.rb#L607).

**ActiveMerchant**: 1.60.0
**Ruby**: 2.3.1
**ActiveSupport**: 5.0.0.1
**Nokogiri**: 1.6.8.1

bpollack:
Looks to me as if this was resolved in #2248, so closing, but please reopen if I’ve misunderstood.

PR #2248, opened 2016-10-29
Orbital: accept order_id option for capture [#2247]

yaychris:
This commit resolves the issue identified in #2247.


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -586,7 +586,7 @@ def build_mark_for_capture_xml(money, authorization, parameters = {})
         xml.tag! :Request do
           xml.tag! :MarkForCapture do
             add_xml_credentials(xml)
-            xml.tag! :OrderID, format_order_id(order_id)
+            xml.tag! :OrderID, format_order_id(order_id || parameters[:order_id])
```
jasonwebster: In the linked issue you mention

> "capture" test cases were rejected because I did not provide the OrderID.

Why is it that the `order_id` from the authorization did not work in your case? Line 584 (the top of this method) attempts to extract it from the authorization, which is what would be typically expected in an auth and capture flow. Is that behavior buggy? It seems weird to me that you'd have to pass this in separately when trying to capture an auth.

yaychris: Apologies for not being more detailed in the issue.

To my knowledge, that behavior is not buggy. However, it only works if you performed the auth using Active Merchant. I'm using Chase's Hosted Pay Form, which performs the auth in an iframe and provides the authorization token. I send the token to the server and capture there, using Active Merchant.

This can be resolved by combining the token and order ID into a string, like Active Merchant does. That's non obvious, though. Given:

1. Chase requires both the authorization token and the order ID to be certified for production;
2. There are two common ways to perform an auth outside of Active Merchant (Hosted Pay Form and Hosted Pay Page);
3. Combining into a string breaks encapsulation by requiring my app to know about Active Merchant's internals (`authorization_string`, which does the combining, is a private method); and
3. This change preserves existing behavior

It seems reasonable to make this a first-class parameter, rather than tacking it on to the authorization.

bpollack:
@jasonwebster Is this something you're still interested in landing?

albaer:
Hi! I'm interested in getting this PR merged, if that's still on the table, and I'd be happy to help if I can. I'm working with the iframe auth/server capture setup Chris described above, and it would be nice to be able to pass in the order id.

bpollack:
Awesome, I'll look @albaer . I've been going through and cleaning up/landing a lot of PRs, but it's a slow burn. I'll try to get you feedback within the next week or so.

bpollack:
@albaer What would you need to update this and help land it?

joeyjoejoejr:
@bpollack I was just talking to some coworkers, and they said they had this open.  It's been rebased and should be ready to go.  Let me know if there are any changes you would need to it.

PR #2250, opened 2016-11-03
Stripe: Ensure ECI values for tokenized cards are padded

jasonwebster:
Stripe requires all ECI values to be properly 0 padded strings. They
will silently ignore the value if not (and typically the processor will
decline the transaction as a result).

I had considered changing this at the NetworkTokenizationCreditCard
level (and you can see what that implementation would look like here
https://github.com/activemerchant/active_merchant/compare/master...jasonwebster:ensure_eci_value_format?expand=1),
however, after reading some other gateway's documentation, it became
increasingly unclear if that change would be compatible with them. While
the standard always indicates zero padded values, some gateways such as
[Payeezy](https://support.payeezy.com/hc/en-us/articles/203730589-Ecommerce-Flag-Values)
indicate they don't, so I decided to just leave this as a Stripe
specific change.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -389,7 +389,7 @@ module ActiveMerchant #:nodoc:
 
           if creditcard.is_a?(NetworkTokenizationCreditCard)
             card[:cryptogram] = creditcard.payment_cryptogram
-            card[:eci] = creditcard.eci
+            card[:eci] = creditcard.eci.rjust(2, '0') if creditcard.eci =~ /^[0-9]+$/
             card[:tokenization_method] = creditcard.source.to_s
           end
           post[:card] = card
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -389,7 +389,7 @@ def add_creditcard(post, creditcard, options)
 
           if creditcard.is_a?(NetworkTokenizationCreditCard)
             card[:cryptogram] = creditcard.payment_cryptogram
-            card[:eci] = creditcard.eci
+            card[:eci] = creditcard.eci.rjust(2, '0') if creditcard.eci =~ /^[0-9]+$/
```
jasonwebster: Yes, this contains another change wherein it is not sent if non-numeric. 

I dug though the history looking at _why_ we allow setting `eci` to `"recurring"` to flip the recurring option, and it seems to just be some legacy and should eventually be dropped.

jasonwebster:
@rwdaigle Got a spare moment to take a look at this? ❤️

rwdaigle:
> I decided to just leave this as a Stripe specific change

👍 Can always generalize in the future should it prove to be broadly applicable.

PR #2255, opened 2016-11-08
Implemented Realex 3ds-verifysig request.

marjaimate:
This request is there to check the results of a 3D secure verification from the user. It needs a 'PaRes'
attribute that the redirect callback got back from the verified by visa
flow.

Details about the Realex API:
https://developer.realexpayments.com/#!/api/3d-secure/3ds-verifysig

### DIFF
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -83,6 +83,13 @@ module ActiveMerchant
         commit(request)
       end
 
+      def verify_sig(money, creditcard, options = {})
+        requires!(options, :order_id, :pares)
+
+        request = build_verify_sig_request(money, creditcard, options)
+        commit(request)
+      end
+
       def supports_scrubbing
         true
       end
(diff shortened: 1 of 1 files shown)


jasonwebster:
Hi! Thanks for the PR.

Can you explain why you think this and https://github.com/activemerchant/active_merchant/pull/2254 belong in Active Merchant? You're introducing public methods to a single gateway that are not part of Active Merchant's established API and abstraction. How does this influence or work with the purchase flow? Do you think that 3DS support is something that should be solved by Active Merchant at all, and if yes, how do you think it could be done in a way that implements a common interface across all gateways?

marjaimate:
Hi @jasonwebster,

When doing checkouts / purchases in Europe (or with a European card), it's very common to be faced with a 3D secure flow from the customer point of view. From the customer's point of view, it's a small inconvenience to complete, but at least they know they are covered.

However, from a developer / integration angle, it's not the easiest flow to get right. 3D secure related code [already exists in active_merchant](https://github.com/activemerchant/active_merchant/search?utf8=%E2%9C%93&q=3d) so I say it's ok to add it for Realex as well.

On the common interface, 3DS has strict rules how to handle, which are all common between these solutions. Having a common interface for it would be great, all it is, is a redirect to a 3rd party with set params (TermURL, PaReq, MD), and then handling the callback when that 3rd party site finished, and sending on the appropriate params (PaRes, MD, etc.) to the next call. Active Merchant is there to make checkout code and integration easier, so having a simpler out-of-the-box solution for 3D secure flows would be an asset to the library I think.

jasonwebster:
Thanks for the response!

I realized my initial questions could have be clearer. I'm well aware of what 3D Secure is, and that we support it throughout Active Merchant.

My primary concern is adding new public methods to a single gateway. Active Merchant is supposed to be an abstraction that provides a common interface to all gateways. If we could identify a common abstraction for this capability that we could adopt as an Active Merchant interface, I'd be all for it.

marjaimate:
Sure thing, I wouldn't be that familiar with the internals of Active Merchant to identify that common ground for all 3DS and implement nicely. I get the argument on keeping it checkout focused, although there are already some gateways that have other APIs - see [some](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/eway_rapid.rb#L109) [not so common](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L173) examples - so I thought this can be added too.

If you feel like, this is not appropriate to Active Merchant, that's totally fine by me.

bpollack:
I reluctantly think that it's not, @marjaimate, but I really appreciate you taking the time to try to supply this for Active Merchant.

Issue #2256, opened 2016-11-09
Migs stops support for MD5 algorithm


mpospelov:
The message from Bank Gateway

We have identified that you are one of a small number of merchants currently using MD5 algorithms and would like to notify you in order to continue processing transactions successfully, you need to make the necessary technical changes and start to use the SHA256 algorithm standard by 30th November 2016.

According to it the hashing algorithm for MiGs should be changed to SHA256.

PR #2257, opened 2016-11-11
implement sha256 algorithm encryption for migs gateway

mpospelov:
Implementation of sha256 for migs gateway based on issue #2256


bpollack:
Hey @mpospelov, I saw this after doing the work for #2676, which solves the same issue slightly differently. I'm going to close this at this point, but thank you for doing this work last year. I apologize it wasn't merged at the time.

abarrak:
Hi @bpollack,

I wonder why closure of this fine PR by @mpospelov .
active merchant `master` is not working with migs `#purchase_offsite_url` unless such changes applied.

This is the rendered error from MiGS unless SHA256 is used:

screen shot 2017-12-12 at 8 29 10 pm

PR #2259, opened 2016-11-15
Maxipago: Send currency with request

curiousepic:
Maxipago was not sending currency, presumably since it is most often
used with, and the gateway defaults to, the BRL currency. But some
acquirers support additional currencies.


davidsantoso:
@curiousepic looks great, thanks!

PR #2260, opened 2016-11-15
Credorax: Map order_id to field H9

curiousepic:
H9 (Merchant Reference Number) is a more suitable field for order_id
than A1. A1 is still sent as a unique ID where required.


davidsantoso:
@curiousepic looks good! Mind adding a quick remote test just to verify? We can probably just add it/update the `test_successful_purchase` test.

curiousepic:
@davidsantoso Yep, updated remote test. Units and their mock responses would be another matter 😬

PR #2267, opened 2016-11-24
Worldpay: Format non-fractional currency amounts correctly

jasonwebster:
https://github.com/activemerchant/active_merchant/pull/2084 introduced a regression to the Worldpay gateway wherein it was no longer implementing the correct contract for non-fractional currencies.

Our [base gateway tests](https://github.com/jasonwebster/active_merchant/blob/07e28c4936c2ae46f0470d878c19f36564c3df6e/test/unit/gateways/gateway_test.rb#L83-L91) and [Stripe tests](https://github.com/jasonwebster/active_merchant/blob/07e28c4936c2ae46f0470d878c19f36564c3df6e/test/unit/gateways/stripe_test.rb#L431-L440) for example, which is another gateway with a `:cents` money format, ensure that you want to charge ¥100, you actually have to pass in `10000`, as if the non-fractional currency _was_ fractional. That's the external contract of Active Merchant--all transaction amounts accepted by the public API methods are expected to be in cents, period.

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -187,7 +187,7 @@ module ActiveMerchant #:nodoc:
         currency = options[:currency] || currency(money)
 
         amount_hash = {
-          :value => amount(money),
+          :value => localized_amount(money, currency),
           'currencyCode' => currency,
           'exponent' => non_fractional_currency?(currency) ? 0 : 2
         }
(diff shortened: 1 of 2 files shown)


davidsantoso:
@jasonwebster took a look at this and it's 👍 from me. You're correct in your interpretation of the external contract of Active Merchant:

> all transaction amounts accepted by the public API methods are expected to be in cents, period.

Essentially it should be that you can always take whatever amount, multiply by 100, and send the currency if it's not the default, and the adapter will handle formatting it properly for you.

PR #2272, opened 2016-12-01
added Adyen adapter

adyenpayments:
API Integration to the Adyen platform.


rwdaigle:
@adyenpayments in the past we've had issues b/c Adyen hasn't allowed third party access to its APIs. Can you clarify this policy or point us to where Adyen's position is stated so we can determine if including it in ActiveMerchant is appropriate?

Thanks!

rikterbeek:
Hi @rwdaigle,

Clients only need to setup a Test account to get access to our APIs, but that’s it. You can signup here: https://www.adyen.com/signup. Very happy to give you more background and context around this initiative: just email support@adyen.com asking for Rik, and we can set-up a call? 

Regards,
Rik
Adyen

elfassy:
@adyenpayments looks like you have a failing test in Rails3.2. Please add
```ruby
require 'active_support/core_ext/hash/slice'
```
to `lib/active_merchant.rb` 👍

adyenpayments:
@elfassy,

Thanks for the help! Did this indeed solve the issue :) Could you merge it into the active_merchant repo ?

Regards,
Rik
Adyen

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -0,0 +1,227 @@
...
+    assert_failure response
+    assert_match Gateway::STANDARD_ERROR_CODE[:invalid_cvc], response.error_code
+  end
+
+  # def test_missing_cvc_for_purchase
+  #   @credit_card.verification_value = '004'
+  #   assert response = @gateway.purchase(@amount, @credit_card, @options)
+  #   assert_failure response
+  #   assert_match Gateway::STANDARD_ERROR_CODE[:invalid_cvc], response.error_code
+  # end
```
jasonwebster: Why is this commented out? Looks like it is covered above so can be removed.

test/unit/gateways/adyen_test.rb
```diff
@@ -0,0 +1,330 @@
...
+      @gateway.verify(@credit_card, @options)
+    end.respond_with(successful_verify_response)
+    assert_success response
+    assert_equal '7914776426645103', response.authorization
+    assert_equal 'Authorised', response.message
+    assert response.test?
+  end
+
+  def test_successful_verify_with_failed_void
+  end
```
jasonwebster: Either implement this test or remove the empty method.

adyenpayments:
Hi @jasonwebster and @elfassy 

I removed thise unused tests from the PR. Let me know if you have any other questions.

Regards,
Rik
Adyen

jasonwebster:
@rwdaigle @davidsantoso This looks good to me. Do either of you have any outstanding thoughts or concerns?

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -0,0 +1,228 @@
...
+          response,
+          authorization: authorization_from(response),
+          test: test?,
+          error_code: success ? nil : error_code_from(response)
+        )
+
+      end
+
+      def basic_auth
+        Base64.encode64("#{@username}:#{@password}").strip.delete("\r\n")
```
rwdaigle: I believe if you use `strict_encode64`, you won't need to manually strip off the newline info.

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -0,0 +1,220 @@
...
+      :verification_value => '737',
+      :brand => 'visa'
+    )
+
+    @declined_card = credit_card('4000300011112220')
+
+    @options = {
+      reference: '345123',
+      shopperEmail: "john.smith@test.com",
+      shopperIP: "77.110.174.153",
```
davidsantoso: We should probably underscore-ize these options. Which actually brings up another point- we have a couple gateway options [that are standardized](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L25), `email` and `ip` being two of them.

Looks like the gateway adapter is just mapping option keys to API request fields which works but does break that soft contract we have with the `options` hash. Other gateways generally take something like `email` from the options and map them to a appropriate API request key. For example, [Be2Bill](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/be2bill.rb#L63)

```ruby
post[:CLIENTEMAIL] = options[:email]
```

Formalizing the options hash is probably something we revisit.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -0,0 +1,228 @@
...
+      }
+
+      CUSTOMER_DATA = %i[
+        shopperEmail shopperIP shopperReference fraudOffset selectedBrand
+        deliveryDate riskdata.deliveryMethod merchantOrderReference
+        shopperInteraction selectedRecurringDetailReference
+      ]
+
+      def initialize(options={})
+        requires!(options, :username, :password, :merchantAccount)
```
jasonwebster: Same applies here regarding snake casing the options.

test/remote/gateways/remote_adyen_test.rb
jasonwebster: @adyenpayments Let's fix these up, then this should be good to go.

adyenpayments:
Hi @jasonwebster @davidsantoso @elfassy,

Many thanks for you feedback I made changes according to your request changes are you able to merge the PR ?

Regards,
Rik
Adyen

renatodex:
This is already merged? There is something pending?

PR #2275, opened 2016-12-02
supports_network_tokenization? implementation for Braintree

elfassy:
Verify that braintree gateways supports network tokenization, similar to what was done for auth.net.
Implementation using `provision_raw_apple_pay` following discussion in https://github.com/activemerchant/active_merchant/pull/2241

ping @Krystosterone @duff @jasonwebster @chrisboulton

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -193,6 +193,10 @@ module ActiveMerchant #:nodoc:
         raise RuntimeError.new("This gateway does not support scrubbing.")
       end
 
+      def network_tokenization_supported
+        []
+      end
+
       def supports_network_tokenization?
         false
       end
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -193,10 +193,6 @@ module ActiveMerchant #:nodoc:
         raise RuntimeError.new("This gateway does not support scrubbing.")
       end
 
-      def network_tokenization_supported
-        []
-      end
-
       def supports_network_tokenization?
         false
       end
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -171,17 +171,8 @@ module ActiveMerchant #:nodoc:
       end
       alias_method :delete, :unstore
 
-      def network_tokenization_supported
-        result = Braintree::MerchantGateway.new(@braintree_gateway).provision_raw_apple_pay
-        if result.success?
-          result.supported_networks.map { |brand| format_card_brand(brand) }
-        else
-          []
-        end
-      end
-
       def supports_network_tokenization?
-        network_tokenization_supported.present?
+        supports_network_tokenization_brands.present?
       end
 
       def verify_credentials
@@ -197,6 +188,15 @@ module ActiveMerchant #:nodoc:
       end
 
       private
+
+      def supports_network_tokenization_brands
+        result = Braintree::MerchantGateway.new(@braintree_gateway).provision_raw_apple_pay
+        if result.success?
+          result.supported_networks.map { |brand| format_card_brand(brand) }
+        else
+          []
+        end
+      end
 
       def check_customer_exists(customer_vault_id)
         commit do
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -191,11 +191,7 @@ module ActiveMerchant #:nodoc:
 
       def supports_network_tokenization_brands
         result = Braintree::MerchantGateway.new(@braintree_gateway).provision_raw_apple_pay
-        if result.success?
-          result.supported_networks.map { |brand| format_card_brand(brand) }
-        else
-          []
-        end
+        result.success? ? result.supported_networks : []
       end
 
       def check_customer_exists(customer_vault_id)
@@ -618,19 +614,6 @@ module ActiveMerchant #:nodoc:
         end
 
         parameters
-      end
-
-      def format_card_brand(card_brand)
-        case card_brand
-        when 'visa'
-          :visa
-        when 'mastercard'
-          :master
-        when 'amex'
-          :american_express
-        when 'discover'
-          :discover
-        end
       end
     end
   end
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -171,8 +171,17 @@ module ActiveMerchant #:nodoc:
       end
       alias_method :delete, :unstore
 
+      def network_tokenization_supported
+        result = Braintree::MerchantGateway.new(@braintree_gateway).provision_raw_apple_pay
+        if result.success?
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateway.rb
```diff
@@ -193,6 +193,10 @@ def scrub(transcript)
         raise RuntimeError.new("This gateway does not support scrubbing.")
       end
 
+      def network_tokenization_supported
```
Krystosterone: Why have `network_tokenization_supported` public and have it on `Gateway`? It's only ever used (for now) by Braintree Blue; by making it public, we're acknowledging this is now part of the public API. We're still not sure what this logic should look like, which means it will probably change in the near future. Having it public means dealing with deprecation and breaking changes as soon as you release the gem. I'd keep it as a hidden implementation detail in `BraintreeBlue`.

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -171,8 +171,17 @@ def unstore(customer_vault_id, options = {})
       end
       alias_method :delete, :unstore
 
+      def network_tokenization_supported
```
Krystosterone: Shouldn't this be named `support_network_tokenized_brands`?

Krystosterone: I understand the discussion that was previously held https://github.com/activemerchant/active_merchant/pull/2241#issuecomment-262103147, but I don't like the lack of query / command segregation here. 

Also, and I lack understanding around this, but doesn't Braintree Vault also use network tokenization? What if someone using the gem wants to ensure Braintree Blue supports network tokenization but isn't setup to provision apple pay? As an integrator, I'd call `supports_network_tokenization?` to ensure I can use Braintree Vault, but I'd get back false. Furthermore, I'd probably get an exception specific to Apple Pay provisioning, which is really confusing.

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -171,8 +171,17 @@ def unstore(customer_vault_id, options = {})
...
+        if result.success?
+          result.supported_networks.map { |brand| format_card_brand(brand) }
+        else
+          []
+        end
+      end
+
       def supports_network_tokenization?
-        true
+        network_tokenization_supported.present?
```
Krystosterone: Once you've provisioned apple pay, can you re-provision a second time? i.e. I call `supports_network_tokenization?` twice in a row, will the second call fail?

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -610,6 +619,19 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
 
         parameters
       end
+
+      def format_card_brand(card_brand)
```
Krystosterone: What prompted this?

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -171,8 +171,17 @@ def unstore(customer_vault_id, options = {})
       end
       alias_method :delete, :unstore
 
+      def network_tokenization_supported
+        result = Braintree::MerchantGateway.new(@braintree_gateway).provision_raw_apple_pay
```
elfassy: see [provision_raw_apple_pay](https://github.com/braintree/braintree_ruby/blob/02db6f8d75830aa3338d61985183aab4eebfab9b/lib/braintree/merchant_gateway.rb#L8)

elfassy: the goal of having a normalized set of card brands across gateways

lib/active_merchant/billing/gateways/braintree_blue.rb
elfassy: no since  `network_tokenization_supported` will always return an array of the supported card brands

lib/active_merchant/billing/gateways/braintree_blue.rb
elfassy: what do you think of `support_network_tokenization_brands`

lib/active_merchant/billing/gateways/braintree_blue.rb
elfassy: Speaking with Jason we agreed it would be best to normalize the card brands with what is used elsewhere in the AM code base

Krystosterone: I don't think this is necessary anymore.

test/remote/gateways/remote_braintree_blue_test.rb
```diff
@@ -17,6 +17,14 @@ def setup
     }
   end
 
+  def test_supports_network_tokenization_brands
```
Krystosterone: Shouldn't we be testing the public interface instead? (`supports_network_tokenization?`)

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -189,6 +189,15 @@ def verify_credentials
 
       private
 
+      def supports_network_tokenization_brands
```
Krystosterone: Re-iterating my previous concern (https://github.com/activemerchant/active_merchant/pull/2275#discussion_r91172790):

> I understand the discussion that was previously held https://github.com/activemerchant/active_merchant/pull/2241#issuecomment-262103147, but I don't like the lack of query / command segregation here. 
> 
> Also, and I lack understanding around this, but doesn't Braintree Vault also use network tokenization? What if someone using the gem wants to ensure Braintree Blue supports network tokenization but isn't setup to provision apple pay? As an integrator, I'd call `supports_network_tokenization?` to ensure I can use Braintree Vault, but I'd get back false. Is this something that is common and that can happen? Furthermore, I'd maybe even get an exception specific to Apple Pay provisioning (if it fails), which is really confusing.

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -189,6 +189,15 @@ def verify_credentials
...
+        result = Braintree::MerchantGateway.new(@braintree_gateway).provision_raw_apple_pay
```
Krystosterone: This method seems unnecessary, as we can now have only:
```
def supports_network_tokenization?
  response = Braintree::MerchantGateway.new(@braintree_gateway).provision_raw_apple_pay
  response.success? && response.supported_networks.present?
end
```

This is if we decide to address the concerns on `provision_raw_apple_pay` though.

test/remote/gateways/remote_braintree_blue_test.rb
elfassy: this is a _remote_ test that is meant to verify we get the correct brands back from braintree

test/unit/gateways/braintree_blue_test.rb
```diff
@@ -646,10 +646,16 @@ def test_android_pay_card
     assert_equal "transaction_id", response.authorization
   end
 
-  def test_supports_network_tokenization
+  def test_supports_network_tokenization_true
+    @gateway.stubs(supports_network_tokenization_brands: ['visa'])
```
Krystosterone: Shouldn't you be stubbing `Braintree::MerchantGateway` rather than a private method?

PR #2280, opened 2016-12-08
Authorize.net: Fix line items quirk

shasum:
@davidsantoso

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -159,7 +159,6 @@ module ActiveMerchant
             add_payment_source(xml, payment)
             add_invoice(xml, options)
             add_customer_data(xml, payment, options)
-            add_line_items(xml, options)
             add_settings(xml, payment, options)
             add_user_fields(xml, amount, options)
           end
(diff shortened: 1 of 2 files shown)


davidsantoso:
Wowza, very strange! One thought- lets move the adding of line items in the to `add_invoice` method. Line items are naturally part of an invoice so I think it's ok to group them in a single method. on top of the implicit ordering issue, that should allow the comment to be isolate to a single place. So, I'm thinking to the `add_invoice` method to look something like:

```ruby
def add_invoice(xml, options)
  xml.order do
    xml.invoiceNumber(truncate(options[:order_id], 20))
    xml.description(truncate(options[:description], 255))
  end

  # Authorize.net API requires lineItems to be placed directly after order tag
  if !options[:line_items].blank?
    xml.lineItems do
      options[:line_items].each do |line_item|
        xml.lineItem do
          line_item.each do |key, value|
            xml.send(camel_case_lower(key), value)
          end
        end
      end
    end
  end
end
```

shasum:
Makes sense to locate it within invoice. Yep, really strange that it behaves this way.

shasum:
Updated

Issue #2271, opened 2016-11-30
PagoFacil does not accept the abbreviated country


Ligator:
Hi,

I don't know if this is a issue of the gem or this is the way it should work but I have this problem only with this gateway. I am passing the country as "MX" but I am getting the error: `'MX' tiene menos de '3' caracteres`. So I found out that I should pass "Mexico".

jasonwebster:
Hi! Thanks for the report!

In the mean time, yes, just passing `Mexico` as your `country` attribute will work. If you felt like submitting a PR to update Active Merchant to handle both full names and codes, and normalize to the gateway's expected value, you'd make that change [here] (https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/pago_facil.rb) and use our [country library](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/country.rb#L319) to normalize.

PR #2281, opened 2016-12-13
fix to handle MX or Mexico in pagofacil gateway

joseluistorres:
This will addressed the issue mentioned here https://github.com/activemerchant/active_merchant/issues/2271
About passing an abbreviature or a FullName for PagoFacil gateway.

### DIFF
--- a/lib/active_merchant/billing/gateways/pago_facil.rb
+++ b/lib/active_merchant/billing/gateways/pago_facil.rb
@@ -40,7 +40,7 @@ module ActiveMerchant #:nodoc:
         post[:colonia] = address[:address2]
         post[:municipio] = address[:city]
         post[:estado] = address[:state]
-        post[:pais] = address[:country]
+        post[:pais] = country_code(address[:country])
         post[:telefono] = address[:phone]
         post[:cp] = address[:zip]
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/pago_facil.rb
```diff
@@ -117,6 +117,14 @@ def json_error(response)
           "raw_response" => response
         }
       end
+
+      def country_code(country)
+        if country
+          country = ActiveMerchant::Country.find(country)
+          country.to_s
+        end
+      rescue InvalidCountryCodeError
```
jasonwebster: I don't think we should be swallowing this error. Shouldn't the caller know if they're passing in bad country data?

joseluistorres: @jasonwebster I agree sorry I totally forgot about this PR

bpollack:
@joelcogen Any interest in updating this PR? I can't merge it in its current state, but I'd be happy to look if it passed on `master`.

bpollack:
Closing based on lack of a response, but please feel free to re-open and update if you're still interested in seeing this land.

PR #2282, opened 2016-12-14
Worldpay: Remove test for Money like object

shasum:
@davidsantoso 

>Support for Money objects is deprecated and will be removed from a future release of ActiveMerchant. Please use an Integer value in cents

Removed test that triggers above warning. It's targeted to test the handling of money-like object. Modifying makes it equivalent to subsequent test `test_currency_exponent_handling`, makes it redundant, hence removal.

Open to ideas if there is a better alternative to removal.


jasonwebster:
:+1: From me, I really shouldn't have added it, it was a red herring for the real issue (which was an incorrect assumption in another test)

PR #2284, opened 2016-12-19
Omise: Update supported country and card type

zdk:
Ref: https://www.omise.co/currency-and-amount

Changes:
Add Japan as supported country
Add JPY as supported currency
Add JCB as supported card brand
Add test for currency and change test card to fix failed test
Add doc for api_version and update User-Agent header


zdk:
@girasquid what do you think about this PR?

jasonwebster:
Hi! Thanks for the PR.

Could you split out the UserAgent change into its own commit from the JP/JPY support? That's a change that should stand on its own.

zdk:
Hi! @jasonwebster 
Thanks for reviewing this PR. I have just splitted out on its own.
If you could check it out again?
Thank you very much!

PR #2286, opened 2016-12-19
Update placement within Shopify

andrewpaliga:
Update "Gateway Placement within Shopify" section to reflect our current stance.

@louiskearns @lyverovski @ntalbott


lyverovski:
👍

PR #2289, opened 2016-12-20
Remove leading or trailing whitespace from CreditCard name

davidsantoso:
@rwdaigle @duff @jasonwebster wondering if any of you had thoughts?

---

When a credit card name is set using `name=` method, the first and last
names will be split into their respective fields. For example-

```ruby
card = ActiveMerchant::Billing::CreditCard.new name: "John"
card.first_name # => ""
card.last_name # => "John"
```

The issue is if we call `name` again on the card, we get a leading space-

```ruby
card.name # => " John"
```

This update will remove any leading or trailing whitespaces when calling
the name method regulardless of the value of first and last name or how
it was originally set.

```ruby
card = ActiveMerchant::Billing::CreditCard.new name: "John"
card.first_name # => ""
card.last_name # => "John"
card.name # => "John"
```

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -245,7 +245,7 @@ module ActiveMerchant #:nodoc:
       #
       # @return [String] the full name of the card holder
       def name
-        [first_name, last_name].compact.join(' ')
+        "#{first_name} #{last_name}".strip
       end
 
       def name=(full_name)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,8 @@
 * PaymentExpress: Update supported countries [shasum]
 * CyberSource: Add Lebanon to supported countries [shasum]
 * Vanco: Update test URL [davidsantoso]
+* Remove leading or trailing whitespace from credit card name [davidsantoso]
+
 
 == Version 1.62.0 (December 5, 2016)
 * AuthorizeNet: Map to standard AVSResult codes [shasum]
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/credit_card.rb
```diff
@@ -245,7 +245,7 @@ def last_name?
       #
       # @return [String] the full name of the card holder
       def name
-        [first_name, last_name].compact.join(' ')
+        [first_name, last_name].compact.reject{ |n| n == "" }.join(' ')
```
rwdaigle: Am I missing an edge case that negates the clarity of logic like this (vs. existing array-based approach)?

```ruby
"#{first_name} #{last_name}".strip
```

davidsantoso: > Am I missing an edge case that negates the clarity of logic like this
Well, not that I know of. Honestly, I was wondering if there was some edge case I wasn't aware of so seemed safer to just use the existing array based approach. Using `strip` certainly seems clearer...

davidsantoso:
I've updated the `name` method to use the clearer approach of-

```ruby
"#{first_name} #{last_name}".strip
```

I'll wait a little longer if others want to chime in before merging.

jasonwebster:
:+1: from me. I like the simplification. It's a minor nitpick, but I question the necessity of the double assertions on all the tests--the `assert_equal` would also always fail in the cases where the `assert_not_equal` you added would.

davidsantoso:
@jasonwebster that's fair. I'll drop the double assertion!

PR #2292, opened 2016-12-24
Add STANDARD_ERROR_CODE: 'api_error' to billing/gateway and gateways/stripe

bl:
Currently there doesn't exist any support for handling generic API errors on `Response#error_code`. This PR looks to add an `api_error` code to `billing/gateway`'s standard error codes. This PR also adds support for Stripe's `api_error` response type through the `type` attribute. This implementation is similar to https://github.com/activemerchant/active_merchant/pull/2145.

Currently if Stripe's API response with an error that doesn't include a `code` or `decline_code`, then the error responses result in `Response#error_code == nil`. Stripe's generic API error is one such example of this response.
```ruby
{
  "error": {
    "type": "api_error",
    "message": "This is a Stripe API error resposne"
  }
}
```

These changes will assign the `Response#error_code` to the response `type` if no code is provided. This allows us to have a "default" error code for each Stripe response type when no actual code is provided.

**How?**
- add `api_error` to the `billing/gateway`'s `STANDARD_ERROR_CODE`
- add `api_error` to `gateways/stripe`'s `STANDARD_ERROR_CODE_MAPPING`
- assign `Response#error_code` to response `type` if no `code` or `decline_code` was provided.

Added billing/stripe tests for
- api_error error responseresponse
- test_mode_live_card error response

**For Review**
@berkcaputcu @jasonwebster

### DIFF
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -711,6 +711,28 @@ class StripeTest < Test::Unit::TestCase
     assert_match(/^Invalid response received from the Stripe API/, response.message)
   end
 
+  def test_api_error_response
+    @gateway.expects(:ssl_request).returns(api_error_response)
+
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_failure response
+
+    assert_equal Gateway::STANDARD_ERROR_CODE[:api_error], response.error_code
+    refute response.test? # unsuccessful request defaults to live
+    #assert_equal 'ch_test_charge', response.authorization
+  end
+
+  def test_test_mode_live_card_error_response
+    @gateway.expects(:ssl_request).returns(test_mode_live_card_error_response)
+
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_failure response
+
+    assert_equal Gateway::STANDARD_ERROR_CODE[:test_mode_live_card], response.error_code
+    refute response.test? # unsuccessful request defaults to live
+    #assert_equal 'ch_test_charge', response.authorization
+  end
+
   def test_add_creditcard_with_credit_card
     post = {}
     @gateway.send(:add_creditcard, post, @credit_card, {})
@@ -2017,6 +2039,30 @@ class StripeTest < Test::Unit::TestCase
     <<-RESPONSE
     {
        foo : bar
+    }
+    RESPONSE
+  end
+
+  def api_error_response
+    <<-RESPONSE
+    {
+      "error": {
+        "type": "api_error",
+        "message": "This is a Stripe API error resposne"
+      }
+    }
+    RESPONSE
+  end
+
+  def test_mode_live_card_error_response
+    <<-RESPONSE
+    {
+      "error": {
+        "type": "card_error",
+        "code": "card_declined",
+        "decline_code": "test_mode_live_card",
+        "message": "Your request was in teset mode but used a non test card"
+      }
     }
     RESPONSE
   end
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -719,7 +719,6 @@ class StripeTest < Test::Unit::TestCase
 
     assert_equal Gateway::STANDARD_ERROR_CODE[:api_error], response.error_code
     refute response.test? # unsuccessful request defaults to live
-    #assert_equal 'ch_test_charge', response.authorization
   end
 
   def test_test_mode_live_card_error_response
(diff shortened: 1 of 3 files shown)


berkcaputcu:
> These changes will assign the Response#error_code to the response type if no code is provided. This allows us to have a "default" error code for each Stripe response type when no actual code is provided.

In which cases, we don't get an actual error code from Stripe? What's the use case for this change?

test/unit/gateways/stripe_test.rb
```diff
@@ -2021,6 +2043,30 @@ def invalid_json_response
...
+  end
+
+  def test_mode_live_card_error_response
+    <<-RESPONSE
+    {
+      "error": {
+        "type": "card_error",
+        "code": "card_declined",
+        "decline_code": "test_mode_live_card",
+        "message": "Your request was in teset mode but used a non test card"
```
berkcaputcu: /s/teset/test

jasonwebster:
If this is an abstraction we want, I'd like to revisit the naming rather than copying it directly from Stripe. `api_error` isn't clear at a glance (unless you're familiar with Stripe's error typing) that it's a "shit's on fire" type response. This is especially true considering your comment, "generic error", which implies it could be anything, including a not-otherwise-typed decline.

bl:
@berkcaputcu The rationale for this change is Stripe's `api_error`s return a `Response` where `error_code: nil`, instead of an actual error code. This 

Currently I've only checked for `api_error` type responses. The Stripe API has returned responses structured in the JSON format above in the description. These responses don't include any other error codes other than the broad `error_type`.

I haven't checked other Stripe responses to see if they also **only** include an error `type` field (and not a `code` or `decline_code` field). This shouldn't be an issue though as errors are filtered on `STANDARD_ERROR_CODE`s, and the only Stripe API error `type` included is `api_error`. This means responses other than `api_error` that return **only** a `type` field still result in `Response.error_code: nil`

@jasonwebster I agree that `api_error` is not a great description, as these errors are generic and revolve around the API itself and (probably) not actual payment service data. I couldn't come up with something better at the time though, so I'm all ears for suggestions.

bl:
Additionally it looks like the Redsys gateway is failing on `master`, which is causing this branch to fail. I'm currently 👀 to figure out what exactly is going on.

berkcaputcu:
`api_error` types (and other error types) are explained in more detail here: https://stripe.com/docs/api#errors

This LGTM

bl:
@jasonwebster @berkcaputcu 
I was wondering if we could get some feedback on a better name to categorize this generic API error instead of `api_error`.

jasonwebster:
@bl I'm going to pick this up today/tomorrow and combine it with another change that's adding another standard error code, if that's all right?

bl:
@jasonwebster sounds good.

jasonwebster:
I've been thinking about this more, and I've realized that this error category should be considered exceptional (literally) behavior. This is a 5xy response from Stripe's APIs, and that should result in an error being raised in the Active Merchant adapter. That's consistent with the behavior of other gateways. 

That is to say, I don't think this is the right approach in this case. 5xy responses should raise errors, and then it's up to the implementor to deal with them.

@davidsantoso Can you confirm if there's an existing pattern that is preferred here?

davidsantoso:
> That is to say, I don't think this is the right approach in this case. 5xy responses should raise errors, and then it's up to the implementor to deal with them.

@jasonwebster I agree with you here- existing pattern is Active Merchant [raises when the response is outside of the `200..300` range](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/posts_data.rb#L79) and it's up to the implementor to deal with it.

jasonwebster:
Perfect, thanks for the confirmation. Surprised to see that the default includes raising on 4xy, since more often than not, that's concretely implemented in gateways as declines or validation failures and just returns non-success Response objects. Anyway, that's a different conversation.

Ok! @bl thanks again for the PR. If you're still interested in persuing this, we'll accept one that changes the Stripe gateway to raise a ResponseError on 5xy response codes.

davidsantoso:
> Surprised to see that the default includes raising on 4xy

Yeah, usually if the gateway returns of 4xx response but it's not a case where we want to raise then we rescue and parse. Here's an example from the [Mastercard a gateway](https://github.com/activemerchant/active_merchant/blob/ccb9f6de262bb025a8d22fb502b05aaafce30b78/lib/active_merchant/billing/gateways/mastercard.rb#L190)  I was recently working on.

```ruby
def commit(action, post)
  url = build_url(post.delete(:orderid), post.delete(:transactionid))
  post[:apiOperation] = action.upcase
  begin
    raw = parse(ssl_request(:put, url, build_request(post), headers))
  rescue ResponseError => e
    raw = parse(e.response.body)
  end
  succeeded = success_from(raw)
  Response.new(
    succeeded,
    message_from(succeeded, raw),
    raw,
    :authorization => authorization_from(post, raw),
    :test => test?
  )
end
```

Like you said, different conversation 😸

PR #2293, opened 2016-12-28
Add #unstore support for AuthorizeNetGateway

jimryan:
This adds `#unstore` support for the Authorize.net gateway.

Usage is very simple: Pass the authorization token that is included in the `#store` response to `#unstore` and it will delete the entire customer profile.  This does not support deleting specific payment profiles because I was unsure of how best to expose that.  I'm happy to discuss and work that up as well, but think this PR also stands on its own and adding payment profile support could be tackled in another PR.

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -180,6 +180,12 @@ module ActiveMerchant
         end
       end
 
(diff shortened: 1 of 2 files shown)


jasonwebster:
Hi! Thanks for the PR. Do you mind rebasing this on to latest master to pick up some CI configuration changes?

jimryan:
Done!

PR #2295, opened 2017-01-02
CardStream: Add dynamic descriptor option fields

curiousepic:
The options merchant_name and dynamic_descriptor correspond to
statementNarrative1 and statementNarrative2 respectively.


jasonwebster:
Should we document these options anywhere? I see `merchant_name` used in a couple other gateways.

davidsantoso:
@jasonwebster that's a thought! In some ways those fields have started to become Active Merchant specific/defined fields since it's a pretty common feature among gateways, but the manner in which the gateway accepts them varies pretty widely.

Seems like it could be another section in the wiki under the [Patterns and Standards](https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards) section?

PR #2296, opened 2017-01-03
Fix Redsys test broken due to year rollover

bl:
Currently a redsys gateway test is failing due to its it's stubbed response relying on a generated signature using the current date. I've opted to revert back to using the year `2017` explicitly because the Base64 signature is generated using a combination of test data, including date and OpenSSL cipher encrypted secret key + order.

Reverting back will also prevent this test from breaking on every year rollover. Alternatively we dould still rely on `Time.now.year + 1` on this test, but then the stubbed response would need to be updated ever year.

**For Review**
@berkcaputcu @jasonwebster


test/unit/gateways/redsys_sha256_test.rb
```diff
@@ -260,7 +262,7 @@ def generate_order_id
...
+    "entrada=%3C%3Fxml+version%3D%221.0%22+encoding%3D%22UTF-8%22%3F%3E%3CREQUEST%3E%3CDATOSENTRADA%3E%3CDS_Version%3E0.1%3C%2FDS_Version%3E%3CDS_MERCHANT_CURRENCY%3E978%3C%2FDS_MERCHANT_CURRENCY%3E%3CDS_MERCHANT_AMOUNT%3E100%3C%2FDS_MERCHANT_AMOUNT%3E%3CDS_MERCHANT_ORDER%3E144742736014%3C%2FDS_MERCHANT_ORDER%3E%3CDS_MERCHANT_TRANSACTIONTYPE%3EA%3C%2FDS_MERCHANT_TRANSACTIONTYPE%3E%3CDS_MERCHANT_PRODUCTDESCRIPTION%2F%3E%3CDS_MERCHANT_TERMINAL%3E1%3C%2FDS_MERCHANT_TERMINAL%3E%3CDS_MERCHANT_MERCHANTCODE%3E091952713%3C%2FDS_MERCHANT_MERCHANTCODE%3E%3CDS_MERCHANT_TITULAR%3ELongbob+Longsen%3C%2FDS_MERCHANT_TITULAR%3E%3CDS_MERCHANT_PAN%3E4548812049400004%3C%2FDS_MERCHANT_PAN%3E%3CDS_MERCHANT_EXPIRYDATE%3E1709%3C%2FDS_MERCHANT_EXPIRYDATE%3E%3CDS_MERCHANT_CVV2%3E123%3C%2FDS_MERCHANT_CVV2%3E%3C%2FDATOSENTRADA%3E%3CDS_SIGNATUREVERSION%3EHMAC_SHA256_V1%3C%2FDS_SIGNATUREVERSION%3E%3CDS_SIGNATURE%3Eq9QH2P%2B4qm8w%2FS85KRPVaepWOrOT2RXlEmyPUce5XRM%3D%3C%2FDS_SIGNATURE%3E%3C%2FREQUEST%3E"
```
bl: I've removed the `#{(Time.now.year + 1).to_s.slice(2,2)}` bit here as this test is now always operating on the year `2017`.

For a more detailed breakdown of the cause of this error, the `DS_SIGNATURE` was being generated from the [sign_request](https://github.com/bl/active_merchant/blob/8c7d505c74615ed4e160c4a92283295ccf2a2b82/lib/active_merchant/billing/gateways/redsys.rb#L486) method. This generated a base 64 encoded number using a bunch of test data, which does not change, but also used the `credit_card.year` and `credit_card.month`

The credit card data is added to the `Gateway.options` on [add_payment](https://github.com/bl/active_merchant/blob/8c7d505c74615ed4e160c4a92283295ccf2a2b82/lib/active_merchant/billing/gateways/redsys.rb#L291) called within `RedsysGateway.purchase`. It's later used in `sign_request` when generating the base 64 value [here](https://github.com/bl/active_merchant/blob/8c7d505c74615ed4e160c4a92283295ccf2a2b82/lib/active_merchant/billing/gateways/redsys.rb#L387) via `merchant_data_xml` method.

bl:
Hmm 🤔, It looks like nokogiri `1.7.0` was released over the break. This version no longer supports ruby `2.0`. This is failing CI. I'm not too sure what the approach to solving this would be.
https://rubygems.org/gems/nokogiri/versions/1.7.0

jasonwebster:
Ugh. I'll deal with that separately, hold on to your horses.

jasonwebster:
For now, just ignore 2.0 CI. I opened https://github.com/activemerchant/active_merchant/issues/2297 to collect opinions on dropping support for it, which has been officially unsupported by the Ruby organization itself for almost a year now.

Issue #2297, opened 2017-01-03
Drop Ruby 2.0 support


jasonwebster:
2.0 was officially EOL'd February 24th, 2016. I'm exceptionally wary of supporting Active Merchant on versions of the language that are not officially supported anymore. 

The release of Nokogiri 1.7.0, which dropped support for Ruby =< 2.0 last week has kind of forced this issue for us, since our gemspec only specifies a minimum required version. Working around it on the gem dependency level is possible, but something that I'd like to avoid.

/cc @davidsantoso @rwdaigle

rwdaigle:
I think it makes sense not to support old versions of Ruby, especially considering the sensitive data a library like AM deals with 👍

davidsantoso:
Sounds reasonable to me!

davidsantoso:
Closed with #2298

PR #2298, opened 2017-01-04
Drop Ruby 2.0 support

jasonwebster:
2.0 was officially EOL'd February 24th, 2016. Given the sensitive nature
of the information Active Merchant deals with, there's no excuse to be
using unmaintained versions of Ruby. Additionally, the release of
Nokogiri 1.7.0, which dropped support for Ruby =< 2.0 last week forced
this issue for us, since our gemspec only specifies a minimum required
version.

Fixes https://github.com/activemerchant/active_merchant/issues/2297

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ sudo: false
 cache: bundler
 
 rvm:
-- 2.0
 - 2.1
 - 2.2.3
+- 2.3.3
 
 gemfile:
 - Gemfile.rails32
@@ -17,8 +17,6 @@ gemfile:
 
 matrix:
   exclude:
-    - rvm: 2.0
-      gemfile: Gemfile.rails5
     - rvm: 2.1
       gemfile: Gemfile.rails5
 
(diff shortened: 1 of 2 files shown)


jasonwebster:
How would that play with our versioning practices? It's not a functional change, just some metadata, so I could argue that a minor release would be ok. Nokogiri, for example, dropped old ruby support in a minor version.

davidsantoso:
> How would that play with our versioning practices?

That's a good question! As far as I know we're following standard [RubyGem semantic versioning](http://guides.rubygems.org/patterns/#semantic-versioning). In any case, I'm on the same with you for a minor release since there aren't any Active Merchant API level changes being made.

Thanks for your work on this!

davidsantoso:
@jasonwebster wondering if there was anything else before merging this in? Got a couple small gateway fixes I'd like to get in but I thought I'd wait for this update first so Travis is green again!

jasonwebster:
Sorry, been a bit distracted. I can pick this back up later this afternoon. 

Looks like we've got a bit of work to do before we can support 2.4. I'm going to have to drop testing on that version for now, and I'll create an issue to look into dealing with the outstanding issues shortly.

davidsantoso:
Sounds good. Thanks much!

jasonwebster:
@davidsantoso All good here. Builds are green. Want to do the honours?

davidsantoso:
@jasonwebster will do, thanks!

PR #2299, opened 2017-01-06
Worldpay: Add session id attribute

shasum:
@davidsantoso


shasum:
@davidsantoso rebased to drop ruby 2.0, passes all Travis checks

Issue #2237, opened 2016-10-11
Authorize.net - allow settings for CIM purchases


fwilkens:
Currently you can specify settings for non-cim purchases: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L216

I'm proposing to add the same for CIM purchases: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L221

This would allow setting things like the duplicate window, recurring billing, etc.

mward-nerdery:
:+1: for this. Our team would certainly take advantage of being able to set the duration window

bpollack:
This was implemented in b4de7a7b6892db2a605890fbeebea186e76d0657

PR #2300, opened 2017-01-08
[FEATURE]: Allow settings to be passed for authorize.net cim purchases

fwilkens:
This adds support for passing settings to CIM purchases, as is allowed with standard purchases. Matching issue: https://github.com/activemerchant/active_merchant/issues/2237.

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -243,6 +243,7 @@ module ActiveMerchant
           xml.send(transaction_type) do
             xml.amount(amount(amount))
             add_payment_source(xml, payment)
+            add_settings(xml, payment, options)
             add_invoice(xml, options)
           end
         end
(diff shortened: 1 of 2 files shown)


jasonwebster:
Looks good to me, but, to be honest, I don't understand the difference between the two Authorize.net APIs/systems.

fwilkens:
Hi Jason,

Thanks for considering this. Some background that might be helpful:

The entry point for non-cim (ad-hoc transactions with CC/ACH details):
https://developer.authorize.net/api/reference/index.html#payment-transactions-charge-a-credit-card

For CIM transactions (charging a payment profile): https://developer.authorize.net/api/reference/index.html#payment-transactions-charge-a-customer-profile

In each case, if you look under the available request parameters for `TransactionSetting` > , you'll notice there are a number of optional settings. For our use case, we want to be able to specify a `duplicateWindow` for CIM transactions. It's possible right now with ad-hoc transactions, but not for CIM transactions, because the options specified in AM are not applied to the request that's built.

Let me know if I can clarify anything on that. Thanks!

jasonwebster:
Oh. So CIM is just their branding for their vaulting service. Hah. Thanks for taking the time to explain that.

PR #2305, opened 2017-01-13
Travis update

bdewater:
- Add Ruby 2.4 to the test matrix
- Use latest Rubies
- Use release version of Active Support 5
- Add a canary build of Active Support 5.1 that is in the RC stage before release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,9 @@ cache: bundler
 
 rvm:
 - 2.1
-- 2.2.3
-- 2.3.3
+- 2.2.7
+- 2.3.4
+- 2.4.1
 
 gemfile:
 - Gemfile.rails32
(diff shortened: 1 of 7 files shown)


davidsantoso:
Hey @bdewater I [recently updated](https://github.com/activemerchant/active_merchant/commit/54cf5c538e16c18d869f2011730fea4ecd1a8b16) the travis.yml file to take into account Rails 5. But I didn't go further than that. Wondering if you still want to update for your other changes?

bdewater:
Forgot about this one :) will update for latest Rubies and Rails 5.1 Monday at the latest.

davidsantoso:
No major rush! I had done that and was combing through the issues last night and noticed/thought I should probably mention something 😸

davidsantoso:
@bdewater I'm curious about replacing `rails` with `activesupport` in the travis file.

```ruby
-gem 'rails', '~> 5.0.0'
 +gem 'activesupport', '~> 5.0.0'
```

Although we're really only using Active Support from Rails as a whole (hopefully we can remove Active Support usage completely), I think this is meant to show that it would work in a Rails X.X app?  I could be misunderstanding how this might work exactly in a Travis build.

bdewater:
It aligns better with our [gemspec](https://github.com/activemerchant/active_merchant/blob/master/activemerchant.gemspec#L24). By requiring a specific `rails` gem version (which is just [a readme file and gem requirements](https://github.com/rails/rails/blob/master/rails.gemspec#L21)) we indirectly require a newer version of `activesupport`.  Now that version requirement is made directly. Also we don't need to download and Rails and it's dependencies, which means we cut down from [44](https://travis-ci.org/activemerchant/active_merchant/jobs/219414626#L220) to [18](https://travis-ci.org/activemerchant/active_merchant/jobs/220562619#L201) gems on the CI for the ActiveSupport 5.0.x test.

If anything, I should rename those files to show we're not tightly coupled to Rails :)

PR #2308, opened 2017-01-17
Raise ResponseError on 5xx Stripe API response codes

bl:
Currently 5xx responses from Stripe executes with no exceptions, but instead returns a `Billing::Response` initialized in the same way as a traditional 4xx error. This behaviour isn't ideal as 5xx responses should be handled differently than valid API responses.

The following gateways have the behaviour of raising `ResponseError` on 5xx (or 4xx) responses:
- [netpay](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/netpay.rb#L204)
- [blue_snap](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/blue_snap.rb#L339)
- [blue_pay](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/blue_pay.rb#L516)
- [cashnet](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cashnet.rb#L136)
- [forte](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/forte.rb#L197)

This PR looks to perform similar 5xx exception raising on the Stripe gateway as the gateways previously described.

**How?**
- keep same behaviour for 200...400 requests
- raise ResponseError for 500...600 internal server errors

Added billing/stripe tests for
  - raising `ResponseError` on 5xx Stripe API responses
  - `test_mode_live_card` error response, previously untested

**For Review**
@jasonwebster @davidsantoso

### DIFF
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -711,6 +711,25 @@ class StripeTest < Test::Unit::TestCase
     assert_match(/^Invalid response received from the Stripe API/, response.message)
   end
 
+  def test_api_error_response
+    http_response = stub(code: 500, message: "", body: "")
+    @gateway.expects(:ssl_request).raises(ActiveMerchant::ResponseError.new(http_response))
+
+    assert_raises ActiveMerchant::ResponseError do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+  end
+
+  def test_test_mode_live_card_error_response
+    @gateway.expects(:ssl_request).returns(test_mode_live_card_error_response)
+
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_failure response
+
+    assert_equal Gateway::STANDARD_ERROR_CODE[:test_mode_live_card], response.error_code
+    refute response.test? # unsuccessful request defaults to live
+  end
+
   def test_add_creditcard_with_credit_card
     post = {}
     @gateway.send(:add_creditcard, post, @credit_card, {})
@@ -2017,6 +2036,19 @@ class StripeTest < Test::Unit::TestCase
     <<-RESPONSE
     {
        foo : bar
+    }
+    RESPONSE
+  end
+
+  def test_mode_live_card_error_response
+    <<-RESPONSE
+    {
+      "error": {
+        "type": "card_error",
+        "code": "card_declined",
+        "decline_code": "test_mode_live_card",
+        "message": "Your request was in test mode but used a non test card"
+      }
     }
     RESPONSE
   end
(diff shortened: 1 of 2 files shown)


test/unit/gateways/stripe_test.rb
```diff
@@ -711,6 +711,25 @@ def test_invalid_raw_response
...
+  def test_api_error_response
+    http_response = stub(code: 500, message: "This is an API error response", body: "")
+    @gateway.expects(:ssl_request).raises(ActiveMerchant::ResponseError.new(http_response))
+
+    assert_raises ActiveMerchant::ResponseError do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+  end
+
+  def test_test_mode_live_card_error_response
```
bl: This test simply covers `test_mode_live_card` standard error codes are being assigned correctly on `decline_code: test_mode_live_card` Stripe API responses. This is unrelated to this PR and can be moved to a separate PR if it makes more sense.

davidsantoso:
@bl thanks for submitting this PR! Been thinking about this one a bit this morning and have a couple thoughts/questions:

1. I don't feel great about a `raise` directly following a `rescue`. It would probably be better to override the `handle_response` method like some of the other gateways have done.
2. I'm a bit curious as to the forcing factor behind this change. I can see how the `error_code` will be set to `nil` on those particular responses from Stripe, but is there a reason there must be an `error_code` vs just leaving it as `nil` in the same way that we didn't receive a code from Stripe?
3. We need to be a bit cautious when making changes that raise errors where they weren't before. Although the general pattern is for Active Merchant to `raise` on `400..500` HTTP responses, the reality is that gateway APIs are just different and therefore at times we deviate from that pattern. By raising an exception now where it wasn't being raise previously, we risk breaking people that are using the adapter but are not handling the exception at the moment.

bl:
@davidsantoso 
1. I agree this approach might not be most ideal. I'm currently looking into a cleaner alternative that doesn't rely on raising a previously rescued exception.
2. The goal of this change is to have a way of correctly categorizing an API/service error from the Stripe gateway. Currently `error_code: nil` doesn't tell us much more than the error received couldn't be mapped to an AM supported error code. This could have been a valid 4xx response with an error code not currently supported, or a 5xx API error response. Either approach of adding `api_error` to AM's supported error codes, or raising the ResponseError on 5xx responses satisfies this and gives more insight into the type of error received.
3. I agree that raising the error is new behaviour that forces the hand of people integrating AM without explicitly handling 5xx responses in this way. I think this is one of the major benefits of adding an `api_error` code to the Stripe gateway's supported error codes .`nil` error codes become `api_error` without raising unexpected exceptions.

jasonwebster:
No matter what we do, we're breaking a contract or promise at some level. Currently, standardized error codes are all only for decline type errors. They indicate that he API request happened just fine, but the request was rejected by the processor for some reason or another. They've never been used to communicate communication failures.

bl:
@jasonwebster I agree. Additionally implementing this into `handle_response` as @davidsantoso suggested would require minor refactoring as the Stripe gateways call to `ssl_request` already expects `handle_response` to raise a ResponseError exception.

davidsantoso:
Closing due to inactivity.

PR #2311, opened 2017-01-22
Add support for scrubbing to Payway

bruno:
Adds support for scrubbing transcripts to Payway Gateway.

### DIFF
--- a/lib/active_merchant/billing/gateways/payway.rb
+++ b/lib/active_merchant/billing/gateways/payway.rb
@@ -143,6 +143,17 @@ module ActiveMerchant
         commit(:status, 'customer.orderNumber' => options[:order_id])
       end
 
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((&?customer.password=)[^&]*), '\1[FILTERED]').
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/payway.rb
```diff
@@ -143,6 +143,17 @@ def status(options={})
         commit(:status, 'customer.orderNumber' => options[:order_id])
       end
 
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
```
elskwid: Hi @bruno. I was looking through some scrubbing in another gateway and it seems the `.gsub` chain is a common pattern. It has the downside that all the regexes need to be created on each invocation of `.scrub` . Since you're PR hasn't been merged I thought I'd suggest an alternative (and you can freeze the constant to be even more nice to Ruby 😄 ):

```ruby
SCRUB_PATTERNS = [
  %r((&?customer.password=)[^&]*),
  %r((&?card.PAN=)[^&]*),
  %r((&?card.CVN=)[^&]*)
].freeze

SCRUB_REPLACEMENT = '\1[FILTERED]'

# ...

def scrub(transcript)
  SCRUB_PATTERNS.inject(transcript) do |text, pattern|
    text.gsub(pattern, SCRUB_REPLACEMENT)
  end
end
```

PR #2312, opened 2017-01-23
Use the correct param to extract authorization from Payway

bruno:
Payway does not return the authorization from a transaction in its
response, the implementation used the `:order_number` however the right
key is `customer.orderNumber`.


bpollack:
Thanks for the PR @bruno . Would you mind adding unit and remote tests so we can avoid a regression here?

bpollack:
Closing based on the lack of feedback, but I'm happy to reopen if you'd like.

PR #2314, opened 2017-01-24
fix handling of nonfractional currencies and remove non-existent subunit concept for those

Azdaroth:
Thanks for this awesome gem. I found one issue which looks like a serious problem.

I think the current way of handling non-fractional currencies is quite troublesome and the behaviour is really surprising. I believe there shouldn't be a concept of subunits for currencies that, well, don't have it. When passing e.g. 1000 JPY it should be 1000 regardless if the expected amount is dollars or cents and it's the application's responsibility to pass it in a right format. 

This is really a problem when using a user is supposed to pay 1000 JPY, but then the payment is executed for 10 JPY. I For these currencies it's highly unlikely that the amount would need any conversion, the amount passed as the input most likely won't be in a wrong format. 

Such change probably introduces a big change and is not backwards compatible, but I think it's worth going that way as conversions for such currencies is probably unexpected.

Thoughts?

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -260,14 +260,24 @@ module ActiveMerchant #:nodoc:
       end
 
       def localized_amount(money, currency)
-        amount = amount(money)
-
-        return amount unless non_fractional_currency?(currency)
-
-        if self.money_format == :cents
-          sprintf("%.0f", amount.to_f / 100)
+        return nil if money.nil?
+        amount = if money.respond_to?(:cents)
+          ActiveMerchant.deprecated "Support for Money objects is deprecated and will be removed from a future release of ActiveMerchant. Please use an Integer value in cents"
+          money.cents.to_s
         else
+          money.to_s
+        end
+
+        if money.is_a?(String)
+          raise ArgumentError, 'money amount must be a positive Integer in cents.'
+        end
+
+        if non_fractional_currency?(currency)
           amount.split('.').first
+        elsif money_format == :cents
+          amount
+        else
+          sprintf("%.2f", amount.to_f / 100)
         end
       end
 
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -395,7 +395,7 @@ class CyberSourceTest < Test::Unit::TestCase
 
   def test_nonfractional_currency_handling
     @gateway.expects(:ssl_post).with do |host, request_body|
-      assert_match %r(<grandTotalAmount>1</grandTotalAmount>), request_body
+      assert_match %r(<grandTotalAmount>100</grandTotalAmount>), request_body
       assert_match %r(<currency>JPY</currency>), request_body
       true
     end.returns(successful_nonfractional_authorization_response)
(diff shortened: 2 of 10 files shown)


Azdaroth:
Hmm, I think this issue is more complex... So HUF is considered as non-fractional currency, but apparently it's not somehow, according to Wikipedia (https://en.wikipedia.org/wiki/Cash_rounding):
```
The centesimal subdivision of the Hungarian Forint (fillér) was discontinued in 1999, and is not used for transactions, but may appear in contexts where precision is needed, such as petrol prices. Cash rounding is done to the level of 5 Ft, which is also the smallest denominated coin available after a new set of coins was introduced in 2011. Full precision to the Forint level remains possible through bank transactions.
```
So given it's about payment gateways, I think this currency is fractional in fact?

jasonwebster:
Thanks for the PR.

You're right. The current way of handling non-fractional currencies is confusing, and worse, [the documentation is misleading (or just wrong)](https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards#amount).

This has come up before in a couple issues, such as https://github.com/activemerchant/active_merchant/pull/2267. The current expectation of behaviour is:

> 
Our [base gateway tests](https://github.com/jasonwebster/active_merchant/blob/07e28c4936c2ae46f0470d878c19f36564c3df6e/test/unit/gateways/gateway_test.rb#L83-L91) and [Stripe tests](https://github.com/jasonwebster/active_merchant/blob/07e28c4936c2ae46f0470d878c19f36564c3df6e/test/unit/gateways/stripe_test.rb#L431-L440) for example, which is another gateway with a `:cents` money format, ensure that you want to charge ¥100, you actually have to pass in `10000`, as if the non-fractional currency _was_ fractional. That's the external contract of Active Merchant--all transaction amounts accepted by the public API methods are expected to be in cents, period.

The problem is, changing this contract requires a major release, communication, and a willingness to support 1.x for the foreseeable future. Couple that with the fact that there's a lot of things that are not perfect in Active Merchant, and lots of breaking changes we _do_ want to make. I'd rather be making a significant amount of them in a 2.0 release, rather than just a couple. It'll hopefully ease burden of adoption (by, and ease maintenance load, as we won't have to support a number of previous major releases.

Azdaroth:
Thanks for the answer.

Yeah, I definintely understand the problem with this kind of change, I'm currently handling that with checking if given currency is nonfractional and "normalizing" it if it is, but would be great to see in 2.0 release.

What do you think about updating the docs and mention it explicitly that you need some extra "normalization" process for nonfractional currencies?

jasonwebster:
I definitely want to update the docs, yes, and I personally would love to see this change happen in the next major release.

akihitoalextsuboi:
I will patch this code until next major update will be released, thank you @Azdaroth !!!
Though active_merchant is really awesome!

bpollack:
I'm going to close this out at this point; we don't have any plans for a 2.0 release soon, and @jasonwebster 's points about backcompat are spot-on. If there's a way to do this that doesn't break existing clients, I'm all ears, but I think this would cause too much instability on our payment platforms to merge otherwise. Thank you for taking the time to tackle this refactoring, though. I really do wish we could implement this functionality safely.

PR #2318, opened 2017-01-30
Firstdata E4 (Payeezy): Set correct ECI value for card present swipes

jasonwebster:
The `Ecommerce_Flag` field was missing for swipe transactions, which was
causing them to be declined. Payeezy requires this is set explicitly to
`R` for any card present retail transaction.

See
https://support.payeezy.com/hc/en-us/articles/203730589-Ecommerce-Flag-Values

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4.rb
@@ -226,9 +226,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_credit_card(xml, credit_card, options)
-
         if credit_card.respond_to?(:track_data) && credit_card.track_data.present?
           xml.tag! "Track1", credit_card.track_data
+          xml.tag! "Ecommerce_Flag", "R"
         else
           xml.tag! "Card_Number", credit_card.number
           xml.tag! "Expiry_Date", expdate(credit_card)
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -164,8 +164,10 @@ module ActiveMerchant
     end
 
     def credit_card_with_track_data(number = '4242424242424242', options = {})
+      exp_date = default_expiration_date.strftime("%y%m")
+
       defaults = {
-        :track_data => '%B' + number + '^LONGSEN/L. ^15121200000000000000**123******?',
+        :track_data => "%B#{number}^LONGSEN/L. ^#{exp_date}1200000000000000**123******?",
       }.update(options)
 
       Billing::CreditCard.new(defaults)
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/firstdata_e4.rb
```diff
@@ -275,7 +275,7 @@ def add_network_tokenization_credit_card(xml, credit_card)
       def add_card_authentication_data(xml, options)
         xml.tag! "CAVV", options[:cavv]
         xml.tag! "XID", options[:xid]
-        xml.tag! "Ecommerce_Flag", options[:eci]
+        xml.tag! "Ecommerce_Flag", options[:eci] if options[:eci].present?
```
jasonwebster: Not sure if this is totally necessary, but I wanted to guard against cases where the node would be set twice. It's really not ideal how there's 3 different places in the  flow that this value can be set.

test/test_helper.rb
```diff
@@ -164,8 +164,10 @@ def credit_card(number = '4242424242424242', options = {})
     end
 
     def credit_card_with_track_data(number = '4242424242424242', options = {})
+      exp_date = default_expiration_date.strftime("%y%m")
+
       defaults = {
-        :track_data => '%B' + number + '^LONGSEN/L. ^15121200000000000000**123******?',
+        :track_data => "%B#{number}^LONGSEN/L. ^#{exp_date}1200000000000000**123******?",
```
jasonwebster: https://en.wikipedia.org/wiki/ISO/IEC_7813#Track_1 if you're curious

jasonwebster:
In the course of doing this, I noticed that its quite possible to set the `Ecommerce_Flag` multiple times depending on the options passed in. I want to take a look at how the methods are split up that all add card data to the request, because it's pretty weird and there's some awkward duplication for sure.

jasonwebster:
@davidsantoso I'd appreciate your review when you've got a spare moment.

davidsantoso:
@jasonwebster overall LGTM. I do see what you mean about possible duplicate `Ecommerce_Flag` fields being set. I think it should be ok though since:

1. One case it'll only duplicate if it's a `NetworkTokenizationCreditCard` which if it's a card present swipe then it shouldn't be.
2. In `add_card_authentication_data` method where it's being set again, that looks like 3DSecure stuff which also wouldn't happen during a card present swipe.

jasonwebster:
Yeah, I came to the same conclusion. It's fine for now, and some tests we have actually show that we trigger the duplicate scenario already without any impact. I'll still look at cleaning it up down the road. Thanks for your time!

PR #2320, opened 2017-01-31
Add better support for UnionPay

joshnuss:
This PR adds better support for Union Pay cards

- Union Pay cards fall under the Discover brand (start with "62")
- No checksum/luhn validation is required for Union Pay cards (but are required for discover cards)
- Union Pay cards can be 16-19 digits

https://en.wikipedia.org/wiki/Payment_card_number


joshnuss:
@lyverovski Should we contact someone at China Union Pay to verify this makes sense?

Also, do you think we should surface a new brand called "unionpay" or should we continue to piggyback off "discover"?

joshnuss:
@jasonwebster this PR is ready for review

jasonwebster:
Some context for those following along:

Union Pay is a distinct card brand and processing network. However, it has a partnership with Discover for cross-acceptance in North and Central American markets. So, people can use their Union Pay cards in the U.S. and the transaction will run over Discover's network. This is _exactly_ how JCB works too.

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -234,6 +239,12 @@ def valid_checksum?(numbers) #:nodoc:
 
           sum % 10 == 0
         end
+
+        def checksum_exempt?(number)
+          brand = brand?(number)
```
jasonwebster: Ugh. Return value from a predicate. (I know this isn't your doing, just generally complaining)

joshnuss: 😄  felt the same way

bpollack:
@davidsantoso This still make sense to merge?

joshnuss:
Cleaning up stale PRs from my GitHub list, hope it's OK if I close this one.

Issue #2269, opened 2016-11-24
Create new standard error code for declines due to no network tokenization support


jasonwebster:
The world isn't going to be running PANs over the internet forever. Apple Pay, Android Pay, etc. all rely on gateways and processors supporting network tokenization (and each major network has their own way of doing this). While Active Merchant supports processing these payment tokens, it does not provide any standardized abstraction for when processing them fails due to lack of gateway/processor support.

bpollack:
Looks to me as if you handled this in 30028b69602891b0cece80ca6ea18a7dab78639d, so closing.

PR #2322, opened 2017-02-03
Add new `unsupported_feature` standard error code

jasonwebster:
This should be used when the a transaction fails due to the gateway,
processor or merchant configuration not supporting a feature used by the
transaction, such as network tokenization.

I've included a concrete implementation using Authorize.net. I had also
wanted to add Braintree support, but it looks like that gateway doesn't use
the standard error code mapping system at all currently (ugh), so that'll
require a whole other PR.

Addresses https://github.com/activemerchant/active_merchant/issues/2269

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -36,24 +36,24 @@ module ActiveMerchant
       }
 
       STANDARD_ERROR_CODE_MAPPING = {
-        '36' => STANDARD_ERROR_CODE[:incorrect_number],
-        '237' => STANDARD_ERROR_CODE[:invalid_number],
-        '2315' => STANDARD_ERROR_CODE[:invalid_number],
-        '37' => STANDARD_ERROR_CODE[:invalid_expiry_date],
-        '2316' => STANDARD_ERROR_CODE[:invalid_expiry_date],
-        '378' => STANDARD_ERROR_CODE[:invalid_cvc],
-        '38' => STANDARD_ERROR_CODE[:expired_card],
-        '2317' => STANDARD_ERROR_CODE[:expired_card],
-        '244' => STANDARD_ERROR_CODE[:incorrect_cvc],
-        '227' => STANDARD_ERROR_CODE[:incorrect_address],
         '2127' => STANDARD_ERROR_CODE[:incorrect_address],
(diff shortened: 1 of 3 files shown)


jasonwebster:
@rwdaigle @davidsantoso Would appreciate your thoughts here, especially on the naming. I struggled with it. I found it hard to strike a balance between too specific and too generic.

davidsantoso:
@jasonwebster this looks 👍 for me naming wise.

> but it looks like that gateway doesn't use the standard error code mapping system

So, the `STANDARD_ERROR_CODE` map is quite hit or miss in terms of gateway adapter implementation. I think it started out as a convention but then as more gateways were added, it became a bit unwieldy to try and map all of the different errors a gateway might return since it actually varied quite a bit. We've found it tends to be clearer to return the exact error message from a gateway for the sake of looking through gateway docs or contacting their support.

That being said, feel free to add it to the Braintree gateway! Just thought I'd let you know where that stands in terms of usage.

jasonwebster:
Thanks for the context! Always helpful to know the history of things.

I'm finding that standardized response codes like this are becoming more and more important for my use cases, so I wouldn't be surprised to see a few more PRs adding the code normalization (and preserving current messages) to other gateways.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -36,24 +36,25 @@ class AuthorizeNetGateway < Gateway
...
+        '2316' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        '2317' => STANDARD_ERROR_CODE[:expired_card],
         '235' => STANDARD_ERROR_CODE[:processing_error],
+        '237' => STANDARD_ERROR_CODE[:invalid_number],
         '24' => STANDARD_ERROR_CODE[:pickup_card],
+        '244' => STANDARD_ERROR_CODE[:incorrect_cvc],
         '300' => STANDARD_ERROR_CODE[:config_error],
-        '384' => STANDARD_ERROR_CODE[:config_error]
+        '3153' => STANDARD_ERROR_CODE[:processing_error],
+        '3155' => STANDARD_ERROR_CODE[:unsupported_feature],
```
AnotherJoSmith: For future reviewers, this is the only new line. The others were just sorted.

jasonwebster: :D And that's why they're separate commits

PR #2324, opened 2017-02-03
Braintree Blue: Pass cardholder_name with card

curiousepic:
Previously, cardholder_name was only passed with a credit card on store
actions. Other actions lacked it, which may affect decline rates. It is
also required for AVS.

@dsantoso to confirm and merge

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -590,7 +590,8 @@ module ActiveMerchant #:nodoc:
               :number => credit_card_or_vault_id.number,
               :cvv => credit_card_or_vault_id.verification_value,
               :expiration_month => credit_card_or_vault_id.month.to_s.rjust(2, "0"),
-              :expiration_year => credit_card_or_vault_id.year.to_s
+              :expiration_year => credit_card_or_vault_id.year.to_s,
+              :cardholder_name => "#{credit_card_or_vault_id.first_name} #{credit_card_or_vault_id.last_name}"
             }
           end
         end
--- a/test/unit/gateways/braintree_blue_test.rb
+++ b/test/unit/gateways/braintree_blue_test.rb
@@ -497,6 +497,13 @@ class BraintreeBlueTest < Test::Unit::TestCase
     @gateway.purchase(100, credit_card("41111111111111111111"), :billing_address => {:zip => "1234567890"})
   end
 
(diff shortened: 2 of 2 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -590,7 +590,8 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
               :number => credit_card_or_vault_id.number,
               :cvv => credit_card_or_vault_id.verification_value,
               :expiration_month => credit_card_or_vault_id.month.to_s.rjust(2, "0"),
-              :expiration_year => credit_card_or_vault_id.year.to_s
+              :expiration_year => credit_card_or_vault_id.year.to_s,
+              :cardholder_name => "#{credit_card_or_vault_id.first_name} #{credit_card_or_vault_id.last_name}"
```
davidsantoso: @curiousepic just wanna double check- any particular reason for not using the [`name` method on the credit card](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card.rb#L247)?

curiousepic: Really only because this was the way it was already being constructed for Apple Pay and I was in context. You're right that either would work.

davidsantoso: @curiousepic 👍 sounds good. I'll just update/amend the commit it to use `name` instead if that's ok.

davidsantoso:
> Should we be adding it to the Android Pay cryprogram params too (just above)?

@jasonwebster good thought- I did a quick check of the available fields for an `AndroidPayCard` in the Braintree gem docs and it looks like `cardholder_name` isn't actually an accepted field. 

https://developers.braintreepayments.com/reference/response/android-pay-card/ruby

Interestingly enough, it's not an accepted field for an `ApplePayCard` either so that makes me think that the field is ignored. I'll leave it in the `ApplePayCard` for now, but I think maybe we leave off adding it to Android Pay? Any objection?

jasonwebster:
👍 Hah, sorry. I should have just checked the docs.

davidsantoso:
> I should have just checked the docs.
@jasonwebster no worries at all. It was a good thing to ask about since it could've very well been available and a good field to add!

davidsantoso:
@curiousepic updated and merged. Thanks!

PR #2326, opened 2017-02-07
Kushki: Add new gateway

shasum:
@davidsantoso

### DIFF
--- a/lib/active_merchant/billing/gateways/kushki.rb
+++ b/lib/active_merchant/billing/gateways/kushki.rb
@@ -1,0 +1,217 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class KushkiGateway < Gateway
+      self.display_name = "Kushki"
+      self.homepage_url = "https://www.kushkipagos.com"
+
+      self.test_url = "https://api-uat.kushkipagos.com/v1/"
+      self.live_url = "https://api.kushkipagos.com/v1/"
+
+      self.supported_countries = ["CO", "EC"]
+      self.default_currency = "USD"
+      self.money_format = :dollars
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club]
+
+      def initialize(options={})
+        requires!(options, :public_merchant_id, :private_merchant_id)
+        super
+      end
+
+      def purchase(amount, payment_method, options={})
+        MultiResponse.run() do |r|
+          r.process { tokenize(amount, payment_method, options) }
+          r.process { charge(amount, r.authorization, options) }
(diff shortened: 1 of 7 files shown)


davidsantoso:
@duff @rwdaigle @jasonwebster would appreciate if we could get your eyes on this one. The Kushki gateway requires a `DELETE` request with a request body to run their `void` so we're adding an additional request type. Wondering if you have any thoughts on naming, general implementation, other thoughts around this kind of addition?

lib/active_merchant/delete_with_body.rb
```diff
@@ -0,0 +1,10 @@
+require 'net/http'
+require 'net/https'
+
+module ActiveMerchant
```
rwdaigle: Not gonna act like I'd build a library that uses class constants to define functionality, but at least it's convenient 👍

davidsantoso:
@duff @jasonwebster another quick ping to see if either of you have any thoughts. No worries if not, just wanna make sure things gets a couple more eyes to be sure we're 👍 considering it's got a new HTTP `DELETE` request that allows a body.

duff:
👍 👍

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -0,0 +1,217 @@
...
+        post = {}
+        add_reference(post, authorization, options)
+        add_invoice(action, post, amount, options)
+
+        commit(action, post)
+      end
+
+      def add_invoice(action, post, money, options)
+        if action == "tokenize"
+          post[:totalAmount] = amount(money).to_f
```
jasonwebster: Should this include a precision?

lib/active_merchant/connection.rb
```diff
@@ -77,6 +77,9 @@ def request(method, body, headers = {})
               # very unambiguously does not.
               raise ArgumentError, "DELETE requests do not support a request body" if body
               http.delete(endpoint.request_uri, headers)
+            when :delete_with_body
+              debug body
```
jasonwebster: Oh, nevermind. They all do this. Sorry

lib/active_merchant/billing/gateways/kushki.rb
shasum: @jasonwebster The `amount()` method does that internally.

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L254

davidsantoso:
Alrighty, looks like no major objections. I'm going to merge this in.

davidsantoso:
@shasum thanks!

PR #2329, opened 2017-02-08
Check `:pem` exists before trying to mutate to strip

jasonwebster:
Introduced in https://github.com/activemerchant/active_merchant/commit/df2fc16cd7b1e7791d6073e4bbf712c1795f5b6e this regressed the ArgumentError that was raised if no `pem` option was provided.

/cc @curiousepic


curiousepic:
@jasonwebster Ah, LGTM. Thanks!

PR #2330, opened 2017-02-08
GlobalCollect: Pass options to Refund

curiousepic:
Refund had been calling add_amount with the argument setting options to
an empty hash. Now options (namely currency) will be passed correctly
when Refunding, and the default happens in the add_amount method.


davidsantoso:
@curiousepic thanks!

PR #2331, opened 2017-02-08
Remove Barclays ePDQ MPI

bdewater:
At Shopify we're seeing 'connection refused' errors to the endpoint and after reaching out to their support it turns out that this product is not available to new accounts. It has been replaced by [ePDQ DirectLink](https://support.epdq.co.uk/en/guides/integration%20guides/directlink).


elfassy:
should we close this PR?

bdewater:
No, still valid.

davidsantoso:
Thanks for the ping- 👍 on the Spreedly end.

elfassy:
looks like we forgot to remove the remote tests for this gateway https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_barclays_epdq_test.rb. I'll open a PR

PR #2334, opened 2017-02-09
PayU LATAM: Let Refund take amount value

curiousepic:
The Refund action previously did not take an amount argument, and so was
failing when present. PayU LATAM's sandbox appears to not fully test
refunds, so a remote test to check that it fails in the expected way has
been added. Remote tests for Void and unit tests for both are also
added.


davidsantoso:
@curiousepic thanks!

PR #2335, opened 2017-02-09
Stripe: Remove idempotency re-submission

shasum:
@davidsantoso

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -71,7 +71,6 @@ module ActiveMerchant #:nodoc:
         requires!(options, :login)
         @api_key = options[:login]
         @fee_refund_api_key = options[:fee_refund_login]
-
         super
       end
 
@@ -159,6 +158,7 @@ module ActiveMerchant #:nodoc:
       def verify(payment, options = {})
         MultiResponse.run(:use_first_response) do |r|
           r.process { authorize(auth_minimum_amount(options), payment, options) }
+          options[:idempotency_key] = nil
           r.process(:ignore_result) { void(r.authorization, options) }
         end
       end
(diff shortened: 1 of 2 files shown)


davidsantoso:
@shasum sorry for the delay in reviewing this. I'll take a look sometime on Monday morning.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -478,7 +480,7 @@ def headers(options = {})
           "X-Stripe-Client-User-Agent" => user_agent,
           "X-Stripe-Client-User-Metadata" => {:ip => options[:ip]}.to_json
         }
-        headers.merge!("Idempotency-Key" => idempotency_key) if idempotency_key
+        headers.merge!("Idempotency-Key" => idempotency_key) if idempotency_key && @enable_idempotency
```
jasonwebster: Can't we just delete the key from the options for void requests instead of introducing this additional piece of state?

davidsantoso: 👍 for deleting it from the `options` hash on line 163.

davidsantoso:
@shasum if you could make that small update noted in the inline comments we should be 👍 .

lib/active_merchant/billing/gateways/stripe.rb
shasum: Cool, will update. I wasn't sure if it was okay to alter the `options` hash.

jasonwebster: Yeah, I get that. I'd usually shy away from mutating input as well, but in this case, I think it's better than having potentially leaky state.

shasum: Agreed.

shasum:
Updated!

davidsantoso:
@shasum @jasonwebster thanks!

PR #2340, opened 2017-02-17
Sage: Default billing state

shasum:
@davidsantoso

### DIFF
--- a/lib/active_merchant/billing/gateways/sage.rb
+++ b/lib/active_merchant/billing/gateways/sage.rb
@@ -1,6 +1,8 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class SageGateway < Gateway
+      include Empty
+
       self.display_name = 'http://www.sagepayments.com'
       self.homepage_url = 'Sage Payment Solutions'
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/sage.rb
```diff
@@ -226,7 +226,7 @@ def add_addresses(post, options)
 
         post[:C_address]    = billing_address[:address1]
         post[:C_city]       = billing_address[:city]
-        post[:C_state]      = billing_address[:state]
+        post[:C_state]      = billing_address[:state].blank? ? "Outside of US" : billing_address[:state]
```
davidsantoso: @shasum lets swap the `blank?` method with Active Merchants [Empty module](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/empty.rb). You can also search `include Empty` to see how it's used in other gateways.

Main reason is that `blank?` is an Active Support helper method so anyone using this gateway will now need the Active Support gem installed and it'd be good to not have that as a dependency. Also, don't want to assume we're within the context of a Rails application! Also also, if you see it used other places feel free to open an issue/PR to remove it. 😸

shasum:
@davidsantoso swapped!

PR #2343, opened 2017-02-22
Added functionality to Elavon client to enable to pass custom fields…

dmitrypol:
… configured for specific accounts

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -10,7 +10,17 @@ module ActiveMerchant #:nodoc:
     #               :login     => "my_virtual_merchant_id",
     #               :password  => "my_virtual_merchant_pin",
     #               :user      => "my_virtual_merchant_user_id" # optional
+    #               :custom_fields_array => [:my_field1, :my_field2] # optional.
     #            )
+    #
+    #  pass values for custom_fields_array in the options hash
+    # options = {
+    #   email: user.email,
+    #   address: {address1: '123', ...},
+    #   my_field1: 'my custom value1',
+    #   my_field2: 'my custom value2',
+    #   ...
+    # }
     #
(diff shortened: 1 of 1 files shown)


dmitrypol:
I was wondering if you had a chance to look at this PR?  Elavon supports custom fields and I think this could be a valuable feature for other devs.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -66,8 +76,10 @@ class ElavonGateway < Gateway
       # * <tt>:password</tt> -- PIN
       # * <tt>:user</tt> -- Specify a subuser of the account (optional)
       # * <tt>:test => +true+ or +false+</tt> -- Force test transactions
+      # * <tt>:custom_fields_array => custom fields configured on specific Elavon account</tt> (optional)
       def initialize(options = {})
         requires!(options, :login, :password)
+        @custom_fields_array = options[:custom_fields_array] || [:customer_number]
```
jasonwebster: `custom_fields_array` feels a bit off to me, how about just `custom_fields`? And how do you feel about moving the default to a constant?

dmitrypol: no problem changing custom_fields_array to custom_fields.  Can you clarify what you mean by changing default to constant?  Do you mean `customer_number`?

jasonwebster: Yep! Like, adding a:

```ruby
DEFAULT_CUSTOM_FIELDS = [:customer_name]
```

to the class and using it here.

dmitrypol: no problem, just pushed.  Thank you for considering my PR

dmitrypol:
Hi @jasonwebster - did the PR changes I make reflect your ideas?

jasonwebster:
Yeah! Looks perfect. Thanks for the PR!

Issue #2276, opened 2016-12-04
Checkout.com: invalid money_format (defined as decimals instead of dollars or cents).


RobertChanphakeo:
Refer to https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/checkout.rb

`      self.money_format = :decimals`

money_format can either be "dollars" or "cents" however "checkout.rb" moneyformat is decimals. According to checkout.com documentation the money format should be defined as "cents" (http://developers.checkout.com/docs/server/api-reference/charges)

jasonwebster:
Hey! Thanks for the issue. Feel like submitting a quick PR to fix the issue?

aemonm:
I dropped in a PR for this, #2344.

bdewater:
PR merged 😄

PR #2344, opened 2017-02-23
change money_format from :decimal to :cents #2276

aemonm:


bdewater:
Thanks for your contribution!

PR #2347, opened 2017-02-27
Braintree: Add Android Pay meta data fields

jknipp:
Braintree requested two additional fields be added for Android Pay
requests - `source_card_type` and `source_card_last_four`.

From Braintree:

For Android Pay, we noticed that when we are passed decrypted
tokens, there are two fields missing from your payload. Both fields are
related to the details of the source card that is in the user's Android
Pay wallet.

* `source_card_type`
* `source_card_last_four`

We currently use these fields within our search functionality, as such
it'd be super helpful if you could make this update.

@davidsantoso see spreedly/core#2379


jknipp:
At least in the sandbox, the `android_pay_card` details are not exposed on the Braintree Transaction response object.  I was able to confirm the changes are saved by using the sandbox ui.  Attached are screenshots showing before and after adding the attributes
before adding braintree android pay fields
after adding braintree android pay fields

.

jknipp:
Fwiw, the attribute on the transaction request is `source_card_last_four`, but in the response it is `source_card_last_4`, verified in the Braintree gem source.

davidsantoso:
@jknipp thanks!

PR #2348, opened 2017-02-28
Kushki: Remove body from `void` call

shasum:
@davidsantoso


davidsantoso:
@shasum thanks! A bit unfortunate Kushki decided after the fact to make a body option on the `void` call. But, glad we can get back to the usual Active Merchant connection conventions.

davidsantoso:
@shasum merged, thanks!

PR #2351, opened 2017-03-06
Adding Trexle gateway

hossamhossny:
adding Trexle API; https://trexle.com

### DIFF
--- a/lib/active_merchant/billing/gateways/trexle.rb
+++ b/lib/active_merchant/billing/gateways/trexle.rb
@@ -1,0 +1,217 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class TrexleGateway < Gateway
+      self.test_url = 'https://core.trexle.com/api/v1'
+      self.live_url = 'https://core.trexle.com/api/v1'
+
+      self.default_currency = 'USD'
+      self.money_format = :cents
+      self.supported_countries = %w(AD AE AT AU BD BE BG BN CA CH CY CZ DE DK EE EG ES FI FR GB
+                                    GI GR HK HU ID IE IL IM IN IS IT JO KW LB LI LK LT LU LV MC
+                                    MT MU MV MX MY NL NO NZ OM PH PL PT QA RO SA SE SG SI SK SM
+                                    TR TT UM US VA VN ZA)
+      self.supported_cardtypes = [:visa, :master, :american_express]
+      self.homepage_url = 'https://trexle.com'
+      self.display_name = 'Trexle'
+
+      def initialize(options = {})
+        requires!(options, :api_key)
+        super
+      end
+
+      # Create a charge using a credit card, card token or customer token
+      #
+      # To charge a credit card: purchase([money], [creditcard hash], ...)
+      # To charge a customer: purchase([money], [token], ...)
+      def purchase(money, creditcard, options = {})
+        post = {}
+
+        add_amount(post, money, options)
+        add_customer_data(post, options)
+        add_invoice(post, options)
+        add_creditcard(post, creditcard)
+        add_address(post, creditcard, options)
+        commit(:post, 'charges', post, options)
+      end
+
+      # Create a customer and associated credit card. The token that is returned
+      # can be used instead of a credit card parameter in the purchase method
+      def store(creditcard, options = {})
+        post = {}
+
+        add_creditcard(post, creditcard)
+        add_customer_data(post, options)
+        add_address(post, creditcard, options)
+        commit(:post, 'customers', post, options)
+      end
+
+      # Refund a transaction
+      def refund(money, token, options = {})
+        commit(:post, "charges/#{CGI.escape(token)}/refunds", { amount: amount(money) }, options)
+      end
+
+      # Authorize an amount on a credit card. Once authorized, you can later
+      # capture this charge using the charge token that is returned.
+      def authorize(money, creditcard, options = {})
+        post = {}
+
+        add_amount(post, money, options)
+        add_customer_data(post, options)
+        add_invoice(post, options)
+        add_creditcard(post, creditcard)
+        add_address(post, creditcard, options)
+        post[:capture] = false
+        commit(:post, 'charges', post, options)
+      end
+
+      # Captures a previously authorized charge. Capturing only part of the original
+      # authorization is currently not supported.
+      def capture(money, token, options = {})
+        commit(:put, "charges/#{CGI.escape(token)}/capture", { amount: amount(money) }, options)
+      end
+
+      # Updates the credit card for the customer.
+      def update(token, creditcard, options = {})
+        post = {}
+
+        add_creditcard(post, creditcard)
+        add_customer_data(post, options)
+        add_address(post, creditcard, options)
+        commit(:put, "customers/#{CGI.escape(token)}", post, options)
+      end
+
+      def supports_scrubbing
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
+          gsub(/(number\\?":\\?")(\d*)/, '\1[FILTERED]').
+          gsub(/(cvc\\?":\\?")(\d*)/, '\1[FILTERED]')
+      end
+      private
+
+      def add_amount(post, money, options)
+        post[:amount] = amount(money)
+        post[:currency] = (options[:currency] || currency(money))
+        post[:currency] = post[:currency].upcase if post[:currency]
+      end
+
+      def add_customer_data(post, options)
+        post[:email] = options[:email] if options[:email]
+        post[:ip_address] = options[:ip] if options[:ip]
+      end
+
+      def add_address(post, creditcard, options)
+        return if creditcard.kind_of?(String)
+        address = (options[:billing_address] || options[:address])
+        return unless address
+
+        post[:card] ||= {}
+        post[:card].merge!(
+          address_line1: address[:address1],
+          address_line2: address[:address_line2],
+          address_city: address[:city],
+          address_postcode: address[:zip],
+          address_state: address[:state],
+          address_country: address[:country]
+        )
+      end
+
+      def add_invoice(post, options)
+        post[:description] = options[:description] || "Active Merchant Purchase"
+      end
+
+      def add_creditcard(post, creditcard)
+        if creditcard.respond_to?(:number)
+          post[:card] ||= {}
+
+          post[:card].merge!(
+            number: creditcard.number,
+            expiry_month: creditcard.month,
+            expiry_year: creditcard.year,
+            cvc: creditcard.verification_value,
+            name: creditcard.name
+          )
+        elsif creditcard.kind_of?(String)
+          if creditcard =~ /^token_/
+            post[:card_token] = creditcard
+          else
+            post[:customer_token] = creditcard
+          end
+        end
+      end
+
+      def headers(params = {})
+        result = {
+          "Content-Type" => "application/json",
+          "Authorization" => "Basic #{Base64.strict_encode64(options[:api_key] + ':').strip}"
+        }
+
+        result['X-Partner-Key'] = params[:partner_key] if params[:partner_key]
+        result['X-Safe-Card'] = params[:safe_card] if params[:safe_card]
+        result
+      end
+      
+      def commit(method, action, params, options)
+        url = "#{test? ? test_url : live_url}/#{action}"
+        raw_response = ssl_request(method, url, post_data(params), headers(options))
+        parsed_response = parse(raw_response)
+        success_response(parsed_response) 
+      rescue ResponseError => e
+        error_response(parse(e.response.body))
+      rescue JSON::ParserError
+        unparsable_response(raw_response)
+      end
+  
+      def success_response(body)
+        return invalid_response unless body['response']
+      
+        response = body['response']
+        Response.new(
+         true,
+         response['status_message'],
+         body,
+         authorization: token(response),
+         test: test?
+        )
+      end
+
+      def error_response(body)
+        return invalid_response unless body['error']
+        Response.new(
+          false,
+          body['error'],
+          body,
+          authorization: nil,
+          test: test?
+        )
+      end
+
+      def unparsable_response(raw_response)
+        message = "Invalid JSON response received from Trexle. Please contact support@trexle.com if you continue to receive this message."
+        message += " (The raw response returned by the API was #{raw_response.inspect})"
+        return Response.new(false, message)
+      end
+      
+      def invalid_response
+        message = "Invalid response."
+        return Response.new(false, message)
+      end
+
+      def token(response)
+        response['token']
+      end
+
+      def parse(body)
+        return {} if body.blank?
+        JSON.parse(body) 
+      end
+
+      def post_data(parameters = {})
+        parameters.to_json
+      end
+    end
+  end
+end
--- a/test/remote/gateways/remote_trexle_test.rb
+++ b/test/remote/gateways/remote_trexle_test.rb
@@ -1,0 +1,170 @@
+require 'test_helper'
+
+class RemoteTrexleTest < Test::Unit::TestCase
+  def setup
+    @gateway = TrexleGateway.new(fixtures(:trexle))
+
+    @amount = 100
+    @credit_card = credit_card('5555555555554444', year: Time.now.year + 2)
+    @visa_credit_card = credit_card('4242424242424242', year: Time.now.year + 3)
+    @declined_card = credit_card('4000000000000119')
+
+    @options = {
+      email: 'john@trexle.com',
+      ip: '66.249.79.118',
+      order_id: '1',
+      billing_address: address,
+      description: "ActiveMerchant Demo Purchase #{DateTime.now.to_i}"
+    }
+  end
+
+  def test_successful_purchase
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal true, response.params['response']['captured']
+  end
+
+  def test_successful_authorize_and_capture
+    authorization = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success authorization
+    assert_equal false, authorization.params['response']['captured']
+
+    response = @gateway.capture(@amount, authorization.authorization, @options)
+    assert_success response
+    assert_equal true, response.params['response']['captured']
+  end
+
+  def test_failed_authorize
+    response = @gateway.authorize(@amount, @declined_card, @options)
+    assert_failure response
+  end
+
+  def test_failed_capture_due_to_invalid_token
+    response = @gateway.capture(@amount, "bogus", @options)
+    assert_failure response
+  end
+
+  def test_failed_capture_due_to_invalid_amount
+    authorization = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success authorization
+    assert_equal authorization.params['response']['captured'], false
+
+    response = @gateway.capture(@amount + 1, authorization.authorization, @options)
+    assert_failure response
+    assert_equal 'Payment failed', response.params['error']
+  end
+
+  def test_successful_purchase_without_description
+    @options.delete(:description)
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+  end
+
+  def test_unsuccessful_purchase
+    response = @gateway.purchase(@amount, @declined_card, @options)
+    assert_failure response
+  end
+
+  # This is a bit manual as we have to create a working card token as
+  # would be returned from trexle.js / the card tokens API which
+  # falls outside of active merchant
+  def test_store_and_charge_with_trexle_js_card_token
+    headers = {
+      "Content-Type" => "application/json",
+      "Authorization" => "Basic #{Base64.strict_encode64(@gateway.options[:api_key] + ':').strip}"
+    }
+    # Get a token equivalent to what is returned by trexle.js
+    card_attrs = {
+      number: @credit_card.number,
+      expiry_month: @credit_card.month,
+      expiry_year: @credit_card.year,
+      cvc: @credit_card.verification_value,
+      name: "#{@credit_card.first_name} #{@credit_card.last_name}",
+      address_line1: "321 Shoreline Park",
+      address_line2: "suite #7",
+      address_city: "Mountain View",
+      address_postcode: "94043",
+      address_state: "CA",
+      address_country: "United States"
+    }
+    url = @gateway.test_url + "/tokens"
+
+    body = JSON.parse(@gateway.ssl_post(url, card_attrs.to_json, headers))
+
+    card_token = body["response"]["token"]
+
+    store = @gateway.store(card_token, @options)
+    assert_success store
+    assert_not_nil store.authorization
+
+    purchase = @gateway.purchase(@amount, card_token, @options)
+    assert_success purchase
+    assert_not_nil purchase.authorization
+  end
+
+  def test_store_and_customer_token_charge
+    response = @gateway.store(@credit_card, @options)
+    assert_success response
+    assert_not_nil response.authorization
+
+    token = response.params['response']['card']['token']
+
+    assert response1 = @gateway.purchase(@amount, token, @options)
+    assert_success response1
+
+    assert response2 = @gateway.purchase(@amount, token, @options)
+    assert_success response2
+    assert_not_equal response1.authorization, response2.authorization
+  end
+
+  def test_store_and_update
+    response = @gateway.store(@credit_card, @options)
+    assert_success response
+    assert_not_nil response.authorization
+    assert_equal @credit_card.year, response.params['response']['card']['expiry_year']
+
+    response = @gateway.update(response.authorization, @credit_card, address: address)
+    assert_success response
+    assert_not_nil response.authorization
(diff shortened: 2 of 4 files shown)


hossamhossny:
Can someone take care of this please?

hossamhossny:
@jasonwebster Can someone take care of this please?

jasonwebster:
Hi!

We can't accept any gateways without unit and remote tests. Take a look over our [Contributing guide in the wiki](https://github.com/activemerchant/active_merchant/wiki/Contributing) to see what's expected of any new gateway submission

hossamhossny:
@jasonwebster Sorry for missing those and for the delay!
I have just committed the unit and remote test files.

hossamhossny:
@davidsantoso I noticed that you are very active for approving new gateways commits, so mentioning you here =) I hope you don't mind.

jasonwebster:
That's looking much better! Thanks! I'll take a thorough look through it as soon as I can.

hossamhossny:
Hi @jasonwebster 
Haven't you had the chance to review this yet? Can I get any estimates on reviewing this?
Also, is there any idea on the release date of v1.65?
I see several gateways implemented and waiting for review since February, but I have no idea how to contribute/apply as a commiter to speed things up a little bit since everyone seems too busy here.

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,218 @@
...
+      end
+
+      def add_address(post, creditcard, options)
+        return if creditcard.kind_of?(String)
+        address = (options[:billing_address] || options[:address])
+        return unless address
+
+        post[:card] ||= {}
+        post[:card].merge!(
+          :address_line1 => address[:address1],
```
jasonwebster: According to the docs, it looks like `:address_line2` is supported, so we should have it here too.

hossamhossny:
@jasonwebster I have added the missing address_line2 to post options. With regard to the :capture optional option it is a flag if set to false, the charge just authorized. If left blank - default is yes - then the charge would be authorized and captured. It is tested successfully with the remote test unit.

Anything wrong regarding this I should take care of?

hossamhossny:
@jasonwebster any updates please?

hossamhossny:
@jasonwebster It's been a couple of weeks since you lastly reviewed this, but then I am not sure if this requires further improvement or is it ready to be merged.

hossamhossny:
@bizla @shasum @curiousepic 
Can anyone please review this and feedback if there is anything needs improvement, or merge it if it's good to go?


```
Units test: 37 tests, 112 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

```

```
Remote test: 14 tests, 41 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

lib/active_merchant/billing/gateways/trexle.rb
hossamhossny: added `:address_line2`

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,219 @@
...
+      end
+
+      def add_invoice(post, options)
+        post[:description] = options[:description] || "Active Merchant Purchase"
+      end
+
+      def add_capture(post, options)
+        capture = options[:capture]
+
+        post[:capture] = capture == false ? false : true
```
bizla: This implementation can be simplified:
```ruby
def add_capture_option(post, options)
  post[:capture] = options[:capture] || true
end
```

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,219 @@
...
+
+        if body["response"]
+          success_response(body)
+        elsif body["error"]
+          error_response(body)
+        end
+
+      rescue JSON::ParserError
+        return unparsable_response(raw_response)
+      end
```
bizla: This looks like it can be cleaned up too:
```ruby
def commit(...)
  url = ...
  raw_response = ssl_request(...) # throws ResponseError
  parsed_response = parse(raw_response) # throws JSON::ParserError
  success_response(parsed_response) 
rescue ResponseError => e
  error_response(parse(e.response.body))
rescue JSON::ParserError
  unparsable_response(raw_response)
end
```

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,219 @@
...
+
+      def success_response(body)
+        response = body["response"]
```
bizla: You'll want to do handle `body["response"]` being `nil`

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,219 @@
...
+          body,
+          :authorization => token(response),
+          :test => test?
+        )
+      end
+
+      def error_response(body)
+        Response.new(
+          false,
+          body['error_description'],
```
bizla: What if this is `nil`?

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,219 @@
...
+      end
+
+      def add_address(post, creditcard, options)
+        return if creditcard.kind_of?(String)
+        address = (options[:billing_address] || options[:address])
+        return unless address
+
+        post[:card] ||= {}
+        post[:card].merge!(
+          :address_line1 => address[:address1],
```
bizla: Use Ruby 1.9+ Hash syntax instead:
```ruby
post[:card].merge!(
  address_line1: address[:address1] 
  address_line2: address[:address2]
  ...
```

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,219 @@
...
+
+        post[:capture] = capture == false ? false : true
+      end
+
+      def add_creditcard(post, creditcard)
+        if creditcard.respond_to?(:number)
+          post[:card] ||= {}
+
+          post[:card].merge!(
+            :number => creditcard.number,
```
bizla: Same here re: Ruby 1.9+ Hash syntax, i.e.
```ruby
number: credit_card.number,
expiry_month: creditcard.month,
...
```

test/remote/gateways/remote_trexle_test.rb
```diff
@@ -0,0 +1,170 @@
+require 'test_helper'
+
+class RemoteTrexleTest < Test::Unit::TestCase
+  def setup
+    @gateway = TrexleGateway.new(fixtures(:trexle))
+
+    @amount = 100
+    @credit_card = credit_card('5555555555554444', :year => Time.now.year + 2)
```
bizla: Use new Hash syntax here throughout this file too. (Sorry for nagging so much! 😬)

test/unit/gateways/trexle_test.rb
```diff
@@ -0,0 +1,528 @@
...
+    assert_equal 'https://core.trexle.com/api/v1', TrexleGateway.test_url
+  end
+
+  def test_live_url
+    assert_equal 'https://core.trexle.com/api/v1', TrexleGateway.live_url
+  end
+
+  def test_supported_countries
+    assert_equal ["AD",
+ "AE",
```
bizla: You might as well use the same construct that reads beautifully in the implementation! :)
```ruby
expected_supported_countries = %w(AD AE AT AU BD BE BG BN CA CH CY CZ DE DK EE EG ES FI FR GB
                                  GI GR HK HU ID IE IL IM IN IS IT JO KW LB LI LK LT LU LV MC
                                  MT MU MV MX MY NL NO NZ OM PH PL PT QA RO SA SE SG SI SK SM
                                  TR TT UM US VA VN ZA)
assert_equal expected_supported_countries, TrexleGateway.supported_countries

File: lib/active_merchant/billing/gateways/trexle.rb (hunk 2 above)
Comment (hossamhossny) [inline lib/active_merchant/billing/gateways/trexle.rb]: This resulted in a couple of failed tests where the `:capture` flag being `true` when it needed to be `false`. I think the right way is:

```
def add_capture(post, options)
  post[:capture] = options[:capture] == false ? false : true
end
```

File: lib/active_merchant/billing/gateways/trexle.rb (hunk 4 above)
Comment (hossamhossny) [inline lib/active_merchant/billing/gateways/trexle.rb]:
```
        if body["response"]
          success_response(body)
        elsif body["error"]
          error_response(body)
        end
```
Shouldn't this forbids `success_response(body)` from being called if `body["response"]` is `nil`?

File: lib/active_merchant/billing/gateways/trexle.rb (hunk 5 above)
Comment (hossamhossny) [inline lib/active_merchant/billing/gateways/trexle.rb]: Are we talking here about `body['error_description']` or `body['error']` being `nil`?

File: lib/active_merchant/billing/gateways/trexle.rb (hunk 3 above)
Comment (hossamhossny) [inline lib/active_merchant/billing/gateways/trexle.rb]: But we omitted here the if/elseif block that checks for `body["response"]` and `body["error"]` for not being `nil`. Am I missing something?

Comment (bizla) [inline lib/active_merchant/billing/gateways/trexle.rb]: As you have it implemented right now, every caller to `#success_response` would have to know to check for `body["response"]` first, as you have it now.

The smell is that I could theoretically write:
```ruby
body = { 'invalid_key': 'invalid_value' }
success_response(body) 
```
with possibly unexpected results, because `#success_response` just assumes that `body["response"]` is there. What if it isn't? 😉 

So that's why I'd prefer to see: 
```ruby
def success_response(body) 
  return uh_oh_something_bad_happened unless body['response']

  # do things now that I've ensured I have a body that I can work with
end
```
and also why I posed the questions of what to do if these functions are called with `nil` or a Hash that's not formed as you expect. (This can happen if suddenly the response schema from a gateway changes even slightly!)

lib/active_merchant/billing/gateways/trexle.rb
bizla: If I write...
```ruby 
body = { something_unrelated: '123' }
error_response(body) 
```
...should `error_response` not complain?

lib/active_merchant/billing/gateways/trexle.rb
bizla: My mistake; you want capture by default if not specified. 
Still, this ternary is a bit ugly. 

Try:
```ruby 
def add_capture(post, options)
  post[:capture] = options[:capture]
  post[:capture] ||= true
end
```

lib/active_merchant/billing/gateways/trexle.rb
bizla: Yes, but the logic can be expressed more cleanly. See my [other response](https://github.com/activemerchant/active_merchant/pull/2351#discussion_r114861620). 😄

lib/active_merchant/billing/gateways/trexle.rb
hossamhossny: Thanks for the amazing explanation! 🥇 😄 
What should I return though as an error message? I couldn't find something similar in other gateways implementations. Same thing for `error_response(body)`, what should I return?
I know I cannot come with error messages from my head here 😆

lib/active_merchant/billing/gateways/trexle.rb
hossamhossny: Totally clear for me now.

lib/active_merchant/billing/gateways/trexle.rb
hossamhossny: I will fix once I know what should I return when validation checking fails 👍

hossamhossny: Oops! I should just return `unparsable_response(raw_response)`

hossamhossny:
@bizla Sorry to bother but where did I go wrong here? 1 test has failed.

bizla:
@hossamhossny For the `add_capture` implementation, I misspoke about my suggestion. 
The `||=` operator will overwrite a value if it's currently `nil` _or_ `false`. 

Try this instead:
```ruby
def add_capture(post, options)
  post[:capture] = if options.key?(:capture)
    options[:capture]
  else
    true
  end
end
```
 
I think this will fix your `add_capture` test.

hossamhossny:
It worked but something fails here, I think it's coming from the `invalid_response` but not sure how is that relevant to the failed tests.

hossamhossny:
I think it's failing because the gateway returns nil with 401 unauthorized, and JSON tries to parse that in the response.

hossamhossny:
@bizla The last commit fixed the failed remote test. Indeed the failed test was [as I thought](https://github.com/activemerchant/active_merchant/pull/2351#issuecomment-299359550). There are 3 failed unit tests all due to test **expecting `nil` as a result** but instead getting the `invalid_response` message! :)
What do you suggest to fix this?

hossamhossny:
Should be working now, thanks! Please re-review 😄

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,223 @@
...
+
+      # Refund a transaction
+      def refund(money, token, options = {})
+        commit(:post, "charges/#{CGI.escape(token)}/refunds", { :amount => amount(money) }, options)
+      end
+
+      # Authorize an amount on a credit card. Once authorized, you can later
+      # capture this charge using the charge token that is returned.
+      def authorize(money, creditcard, options = {})
+        options[:capture] = false
```
bizla: In general, you shouldn't mutate parameters. You should reuse the `add_amount`, `add_customer_data` functions here and instead protect against a `capture: true` option being passed in.

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,223 @@
...
+
+        purchase(money, creditcard, options)
```
bizla: Remove and replace with your `add_*` functions with a `commit` at the end.

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,223 @@
...
+      end
+
+      # Captures a previously authorized charge. Capturing only part of the original
+      # authorization is currently not supported.
+      def capture(money, token, options = {})
+        commit(:put, "charges/#{CGI.escape(token)}/capture", { :amount => amount(money) }, options)
```
bizla: Use Ruby 1.9+ Hash syntax here.

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,223 @@
...
+      end
+
+      def add_invoice(post, options)
+        post[:description] = options[:description] || "Active Merchant Purchase"
+      end
+
+      def add_capture(post, options)
+        post[:capture] = if options.key?(:capture)
+          options[:capture]
+          else
```
bizla: Indentation should be 
```ruby
variable = if condition
  statement
else
  statement
end
```

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,223 @@
...
+
+        result['X-Partner-Key'] = params[:partner_key] if params[:partner_key]
+        result['X-Safe-Card'] = params[:safe_card] if params[:safe_card]
+        result
+      end
+      
+      def commit(method, action, params, options)
+        url = "#{test? ? test_url : live_url}/#{action}"
+        raw_response = ssl_request(method, url, post_data(params), headers(options)) # throws ResponseError
+        parsed_response = parse(raw_response) # throws JSON::ParserError
```
bizla: Those comments were to aid your understanding, but probably don't need to be in the actual code. :)

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,223 @@
...
+        success_response(parsed_response)
+
```
bizla: Trim whitespace

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,223 @@
...
+      
+      def commit(method, action, params, options)
+        url = "#{test? ? test_url : live_url}/#{action}"
+        raw_response = ssl_request(method, url, post_data(params), headers(options)) # throws ResponseError
+        parsed_response = parse(raw_response) # throws JSON::ParserError
+        success_response(parsed_response)
+        
+      rescue ResponseError => e
+        error_response(e.response.body)
+
```
bizla: Trim

test/unit/gateways/trexle_test.rb
```diff
@@ -0,0 +1,463 @@
...
+      }
+    }'
+  end
+
+  def scrubbed_transcript
+    '{
+      "amount":"100",
+      "currency":"USD",
+      "email":"john@trexle.com",
+      "ip_address":"66.249.79.118",
```
bizla: Whose IP is this? :)

hossamhossny: an orphaned GoogleBot living in Mountain View 😃

hossamhossny:
@bizla should be good to go now, please review :)

lib/active_merchant/billing/gateways/trexle.rb
bizla: This line still needs removal.

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,229 @@
...
+      def authorize(money, creditcard, options = {})
+        options[:capture] = false
+        post = {}
+
+        add_amount(post, money, options)
+        add_customer_data(post, options)
+        add_invoice(post, options)
+        add_creditcard(post, creditcard)
+        add_address(post, creditcard, options)
+        add_capture(post, options)
```
bizla: You still want the `capture` option added to the post for authorizations?

hossamhossny: We have to send a capture flag set to false in order to just authorize the charge and not actually capture it. Unless you want me to pass capture = false in the `add_capture`?

hossamhossny: https://docs.trexle.com/#api-Charges-Charges

bizla: If I call `authorize(100, credit_card, capture: true)`, what should happen?

hossamhossny: It will capture the charge.

bizla: Which is wrong. :) `authorize` should never capture... but you should also never mutate the parameters from the caller, so you shouldn't mutate `options`. 
What you want is to ensure `post[:capture] = false`, i.e. you just want to ignore `options[:capture]`... can you think of how to change the implementation to accomplish this?

hossamhossny: I definitely can remove the `add_capture(post, options)` and substitute it with:
commit(:put, "charges/#{CGI.escape(token)}", { amount: amount(money) }, :capture = false)

Will you want this?

bizla: How about just a `post[:capture] = false` before the `commit`?

hossamhossny: That did it! :)

hossamhossny: See @jasonwebster was saying [this](https://github.com/activemerchant/active_merchant/pull/2351#pullrequestreview-31555083) earlier but I never got it correctly.

bizla: So does `add_capture` still need to exist?

hossamhossny: Nope I ditched it.

hossamhossny: @bizla So, are we there yet?

hossamhossny:
@bizla So is this mergeable now or is it pending @jasonwebster approval? Stripe has just listed Trexle gateway in its [integrations page](https://stripe.com/works-with/trexle) 🙌

hossamhossny:
Can you guys update me on this? What is it waiting for?

bizla:
@davidsantoso Wouldn't mind a Spreedly 👀  on this!

lib/active_merchant/billing/gateways/trexle.rb
davidsantoso: @hossamhossny I still see it? Seems like we don't need this line and the associated `add_capture` method since  `purchase` will default to `true` and `authorize` forces `post[:capture] = false`.

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,227 @@
...
+        result
+      end
+      
+      def commit(method, action, params, options)
+        url = "#{test? ? test_url : live_url}/#{action}"
+        raw_response = ssl_request(method, url, post_data(params), headers(options))
+        parsed_response = parse(raw_response)
+        success_response(parsed_response) 
+      rescue ResponseError => e
+        error_response(parse(e.response.body))
```
davidsantoso: On errors, does Trexle return HTTP response codes outside of the 200..300 range?

davidsantoso:
👍 have some questions of my own!~

lib/active_merchant/billing/gateways/trexle.rb
hossamhossny: Yes, 4xx errors too

lib/active_merchant/billing/gateways/trexle.rb
hossamhossny: Do I need to get rid of the `post[:capture] = true` line?

hossamhossny:
@davidsantoso Take a look now please.

hossamhossny:
@davidsantoso Can you let me know your questions regarding this, if any? I understand you are busy, but this is being reviewed for weeks already and mostly is ready to go.

lib/active_merchant/billing/gateways/trexle.rb
```diff
@@ -0,0 +1,217 @@
...
```
davidsantoso: So, the pattern we've been following this that instead of having two separate methods to build a Response object, the rescue instead will take care of pulling together the appropriate fields, namely the success, authorization, message, error_code etc. You can see how the gateway generator script automatically adds these methods to be used.

https://github.com/activemerchant/active_merchant/blob/master/generators/gateway/templates/gateway.rb#L88-L114

The Digitz gateway is another example, although that gateway overrides the handle_response method. You don't need to do that to handle HTTP response codes outside of 200..300 but it's a pattern we often use.

All that goes to say, if we can consolidate the building of the Response object into a single place regardless of it being an error or not I think that would be good.

hossamhossny: I implemented this the same way other gateways do it. For an example, look at wepay..

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/wepay.rb#L177-L186

Or pin payments..

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/pin.rb#L160-L174

I see the Digitz gateway is not doing it differently handling response codes outside 300..499
which is the same concept for Trexle.

Can you recommend a better way of doing this based on existent gateways?

hossamhossny: @davidsantoso Can I have an answer please?

davidsantoso: @hossamhossny apologies on the tardy replies- lots going on these days. 

WePay and Pin have been in Active Merchant a while so they're not following more recent patterns. But, given that we're not as explicit about what we're looking for or conventions I'll let this one in as is. @hossamhossny apologies on the tardy replies- lots going on these days. 

WePay and Pin have been in Active Merchant a while so they're not following more recent patterns. But, given that we're not as explicit about what we're looking for or conventions I'll let this one in as is. 

If you wanna squash your commits and rebase on `master`@hossamhossny apologies on the tardy replies- lots going on these days. 

WePay and Pin have been in Active Merchant a while so they're not following more recent patterns. But, given that we're not as explicit about what we're looking for or conventions I'll let this one in as is. 

If you wanna squash all your commits and rebase on master I can merge in.

hossamhossny: @davidsantoso Thanks for understanding! I have done the squash and rebase; all good now.

davidsantoso:
@hossamhossny merged. Thanks!

hossamhossny:
@davidsantoso Thanks, I am very confused though about this https://github.com/activemerchant/active_merchant/compare/master@%7B1496960317%7D...master and this https://github.com/Trexle/active_merchant/commit/3555c7cc93b9afed4ed54ebb1f39bd8504d2beb3 but I think it is a github issue, so I contacted them. However, did you have to close this PR as oppose to "merge" it? and how can I make future updates for the gateway?

Sorry for bugging you about this.

davidsantoso:
@hossamhossny 

> but I think it is a github issue, so I contacted them. 

Yeah, I'm not 100% sure what happened either. Wondering if something happened when you squashed and rebased but somehow the commit didn't get authored by the email address your GitHub account has?

> However, did you have to close this PR as oppose to "merge" it?
We follow a linear flow with the git history for Active Merchant so when we merge a commit, we update the change log, amend the contributors commit, then merge to `master` closing the commit by adding the PR number in the commit description.

> and how can I make future updates for the gateway?
Same as usual! Just open a PR and tag someone!

PR #2355, opened 2017-03-08
Implementation of new type of Braintree Gateway

ecdemis123:
This new gateway is to offer support for the "v.zero SDK" that Paypal/Braintree offers.  The main difference between how this works with regular braintree is the authentication, where one authenticates with an `access_token` instead of username/password.  Most of this is copypasta from the Braintree Blue gateway, with a few minor changes to options building and things of that nature.  Cheers! :shipit:


bpollack:
The CI tests on this have been broken for over a year with no movement; closing.

PR #2357, opened 2017-03-08
JetPay: Adjust refund and void transaction types

### LINKED ISSUE #2356: JetPay: Adjust refund and void transaction types
At the moment Active Merchant has mapped the `CREDIT` transaction type for `refund` and the `VOID` transaction type for `void`. Although those sound correct, there's a slight nuance when it comes to JetPay. See docs below (page 17 in the XML Authorization Specification PDF docs)-

screen shot 2017-03-07 at 2 32 23 pm

What this means is if you run an `authorize` then subsequent `capture` you can't run a refund on that previous capture transaction as `CREDIT` can only be used on previous transactions that were of transaction type `SALE` (which is what `purchase` is mapped to).

Moreover, `REVERSEAUTH` will work to `void` any previous authorize, capture, or purchase that has not yet settled.

davidsantoso:
Closes #2356

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* JetPay: Adjust refund and void transaction types [davidsantoso] #2356
 
 == Version 1.64.0 (March 6, 2017)
 * Authorize.net: Allow settings to be passed for CIM purchases [fwilkins] #2300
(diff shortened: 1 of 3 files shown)


davidsantoso:
Successful remote test run proof:

```
➜ ruby -I test test/remote/gateways/remote_jetpay_test.rb
Loaded suite test/remote/gateways/remote_jetpay_test
Started
................

Finished in 7.986752 seconds.
16 tests, 87 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
2.00 tests/s, 10.89 assertions/s
```

PR #2358, opened 2017-03-08
add attr_accessor for echeck_type and modify add_bank_account to work…

eddkerr:
… with an object instead of a hash

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net_cim.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net_cim.rb
@@ -815,24 +815,24 @@ module ActiveMerchant #:nodoc:
       # Note: This element should only be included
       # when the payment method is bank account.
       def add_bank_account(xml, bank_account)
-        raise StandardError, "Invalid Bank Account Type: #{bank_account[:account_type]}" unless BANK_ACCOUNT_TYPES.include?(bank_account[:account_type])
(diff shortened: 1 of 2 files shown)


bacrossland:
I would edit the title of this PR to better reflect that this change is for Authorize.net. Place "Authorize.net:" in front of your title.

lib/active_merchant/billing/gateways/authorize_net_cim.rb
```diff
@@ -815,24 +815,24 @@ def add_credit_card(xml, credit_card)
       # Note: This element should only be included
       # when the payment method is bank account.
       def add_bank_account(xml, bank_account)
```
bacrossland: This method currently accepts a Hash. Switching it to an ActiveMerchant::Billing::Check object will cause code that relies on this to break on an upgrade of the ActiveMerchant gem. My suggestion is that you modify the method to accept both the existing Hash and the new ActiveMerchant::Billing::Check object. If it's a Hash a Deprecation Warning to be displayed that in a future release Hash will no longer be supported.

bpollack:
Any interest in implementing the note mentioned by @bacrossland, @eddkerr, or should I close this PR?

bpollack:
Closing due to lack of response, but really happy to reopen if there's movement.

PR #2361, opened 2017-03-11
Add Genesis Gateways

dimitrovv:
* Add support for **Genesis Gateways**
   * Integrate [`eMerchantPay Direct`](https://www.emerchantpay.com/) Gateway
   * Integrate [`E-ComProcessing Direct`](https://www.e-comprocessing.com/) Gateway

### DIFF
--- a/lib/active_merchant/billing/gateways/ecomprocessing_direct.rb
+++ b/lib/active_merchant/billing/gateways/ecomprocessing_direct.rb
@@ -1,0 +1,15 @@
+require_relative 'genesis_gateway'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class EcomprocessingDirectGateway < GenesisGateway
+
+      self.test_url     = 'https://staging-shopify.e-comprocessing.com/payments/api/v1/process'
+      self.live_url     = 'https://shopify.e-comprocessing.com/payments/api/v1/process'
+
+      self.homepage_url = 'https://www.e-comprocessing.com/'
+      self.display_name = 'E-ComProcessing Direct'
+
+    end
+  end
+end
--- a/lib/active_merchant/billing/gateways/emerchantpay_direct.rb
+++ b/lib/active_merchant/billing/gateways/emerchantpay_direct.rb
@@ -1,0 +1,15 @@
+require_relative 'genesis_gateway'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class EmerchantpayDirectGateway < GenesisGateway
+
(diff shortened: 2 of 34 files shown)


dimitrovv:
@jasonwebster @davidsantoso this PR is ready for review

davidsantoso:
@dimitrovv thanks for the PR! A couple things- gateways added to Active Merchant generally follow a structure of a gateway adapter, remote tests, and unit tests. This PR is a bit out of the expected file structure with including helper files and breaking out the tests. See our [wiki page on contributing a gateway](https://github.com/activemerchant/active_merchant/wiki/Contributing) for more details. Unfortunately we won't be adding these adapters until they fall in line with our conventions. 

We do appreciate new gateway adapters so if you're still interested in adding them please open a new pull request utilizing the gateway patterns as outlined in contribution wiki. If anything in the wiki is unclear, feel free to open an issue to discuss.

dimitrovv:
@davidsantoso thanks for the update.

> A couple things- gateways added to Active Merchant generally follow a structure of a gateway adapter, remote tests, and unit tests. This PR is a bit out of the expected file structure with including helper files and breaking out the tests.

We decided to separate our integration into modules and helper classes, because the gateway class should have become large and would have been difficult for review and etc. 

I have read the [wiki page for contributing](https://github.com/activemerchant/active_merchant/wiki/Contributing), before starting with the development. I have also paid attention to the `breantree` and `paypal` integrations, they also use helper classes for their gateway integrations.

Do you mean that we should organize our integration into 3 files:
* **Gateway Class** - inherits from [Gateway Class](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb)  
* **Remote Tests** - include all test cases without using shared examples in modules
* **Unit Tests** - include all test cases without using shared examples in modules

davidsantoso:
> I have also paid attention to the breantree and paypal integrations, they also use helper classes for their gateway integrations.

Yes, I apologize- those are certainly special cases to some extent. The preference is that gateways not follow that pattern unless really necessary so we don't have too many extra helper files everywhere. I can be a bit difficult to parse through all of it! I will say, it's ok if two gateways share the exact same behavior or one is a while label of another, etc. We do have gateways that only change meta data so they include another gateway to keep things DRY. Mastercard/TNS/Citrus Pay are a good example.

> Do you mean that we should organize our integration into 3 files:
> 
> Gateway Class - inherits from Gateway Class
> Remote Tests - include all test cases without using shared examples in modules
> Unit Tests - include all test cases without using shared examples in modules

Yes. Also if it's two gateways then there would be 6 files total.

Thanks!

dimitrovv:
@davidsantoso thanks for the details, the integrations will be restructured and then submitted

PR #2362, opened 2017-03-13
Qvalent: Add soft descriptor fields. Add authorize, capture, and void

davidsantoso:
Overall fairly standard gateway update. However, while working on it I noticed that the authentication setup was not quite fully baked. @duff since there are some authentication changes that include using a certificate, wondering if I could get a second set of eyes on this one?

Successful full remote test run output for verification of changes. 
```
/Users/david/dev/active_merchant qvalent-updates ✔
➜ ruby -Itest test/remote/gateways/remote_qvalent_test.rb

Loaded suite test/remote/gateways/remote_qvalent_test
Started
.............

Finished in 30.773634 seconds.
13 tests, 43 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
0.42 tests/s, 1.40 assertions/s
```


lib/active_merchant/billing/gateways/qvalent.rb
```diff
@@ -12,9 +12,28 @@ class QvalentGateway < Gateway
...
         requires!(options, :username, :password, :merchant)
-        super
+
+        @options = {
+          :pem => QvalentGateway.pem_file,
+          :pem_password => QvalentGateway.pem_password
+        }.update(options)
+
+        raise ArgumentError, "You need to pass in your pem file using the :pem parameter or set it globally using " \
+        "ActiveMerchant::Billing::LinkpointGateway.pem_file = File.read( File.dirname(__FILE__) + '/../mycert.pem' ) " \
```
duff: Thinking this message could say Qvalent.

duff:
@davidsantoso Looks good to me.  Only little question I have is if we need the `cattr_accessor` stuff or if the `pem` and `pem_password` could be specified like the other options passed to `initialize`.

lib/active_merchant/billing/gateways/qvalent.rb
davidsantoso: Yep! I was following the pattern Linkpoint was using.

davidsantoso:
> Only little question I have is if we need the cattr_accessor stuff 

@duff we could just have them be passed in as the other options, bypassing the ability to do a global set. I was using the same pattern as Linkpoint and I was thinking that provided a little flexibility for someone using the adapter in terms of how they wanted to pass in the PEM.

But, I'll defer to your opinion since Linkpoint may not have set a good example in the first place 😸

duff:
Gotcha. Looks like Linkpoint was done awhile back! Started a decade ago! That's like 4 billion in internet years. I lean toward adding only the thing we need now, knowing that if someone wants more flexibility we could add it later with a PR.

davidsantoso:
Right on. Alrighty thanks for your input! I'll push the changes and merge.

PR #2368, opened 2017-03-21
Orbital: Don't send CVV indicator if CVV is not present

curiousepic:
Previously, the CardSecValInd field was being passed with a value of
9 ("Cardholder states data not available") when no CVV was present
and the card brand was Visa or Discover. Chase has indicated that the
field should not be passed at all when no CVV is provided with the
request. This may have been causing declines.

Unfortunately remote tests cannot be run as we do not currently have
credentials for an active test account.

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -417,10 +417,12 @@ module ActiveMerchant #:nodoc:
         #   Do not submit the attribute at all.
         # - http://download.chasepaymentech.com/docs/orbital/orbital_gateway_xml_specification.pdf
         unless creditcard.nil?
-          if %w( visa discover ).include?(creditcard.brand)
-            xml.tag! :CardSecValInd, (creditcard.verification_value? ? '1' : '9')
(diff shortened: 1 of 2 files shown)


jasonwebster:
I have a sandbox Paymentech account, so I can run the remotes for you.

davidsantoso:
@jasonwebster that would be awesome. Thanks!

jasonwebster:
Logs from the remote run: https://gist.github.com/jasonwebster/717a03128b953226689ae6eda3cab381 

One failure that seems unrelated, but we should look into regardless.

davidsantoso:
@jasonwebster thanks for the remote run log- So that failure is interesting cause we have this note in the remote tests-

> Amounts of x.01 will fail

But running a `verify` will do an `authorize` of 100 so I'm not sure that would ever pass in the test environment?

jasonwebster:
Well, it's supposed to be a failed verify, with a declined card, so it might just be stale test data. However, other test cases that rely on that card work, so /shrug

davidsantoso:
👍 alrighty. Ok with you to merge then?

PR #2370, opened 2017-03-22
GlobalCollect: Make message and error reporting more robust

curiousepic:
For some transactions, particularly refunds when rejected,
the transaction was unsusscessful but no error code or
useful message was being mapped in the response. This change
maps statusCode to error_code when errors are absent, and
status to message when an error message is absent.

It also updates some remote test message expectations, so
they are all passing now.

15 tests, 35 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@davidsantoso to review and merge.

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -271,8 +271,10 @@ EOS
         else
           if errors = response["errors"]
             errors.first.try(:[], "message")
+          elsif status = response["status"]
+            "Status: " + status
           else
-            "Unable to read error message"
+            "No message available"
           end
         end
       end
@@ -289,8 +291,10 @@ EOS
         unless succeeded
           if errors = response["errors"]
             errors.first.try(:[], "code")
+          elsif status = response.try(:[], "statusOutput").try(:[], "statusCode")
+            status.to_s
           else
-            "Unable to read error code"
+            "No error code available"
           end
         end
       end
--- a/test/remote/gateways/remote_global_collect_test.rb
+++ b/test/remote/gateways/remote_global_collect_test.rb
@@ -73,7 +73,7 @@ class RemoteGlobalCollectTest < Test::Unit::TestCase
   def test_failed_capture
     response = @gateway.capture(@amount, '123', @options)
     assert_failure response
-    assert_match %r{The given paymentId is not correct}, response.message
+    assert_match %r{UNKNOWN_PAYMENT_ID}, response.message
   end
 
   # Because payments are not fully authorized immediately, refunds can only be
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -289,8 +291,10 @@ def error_code_from(succeeded, response)
         unless succeeded
           if errors = response["errors"]
             errors.first.try(:[], "code")
+          elsif status = response["statusOutput"]["statusCode"]
```
davidsantoso: Wondering if we should use `response.try(:[], "statusOutput").try(:[] "statusCode")`? The Global Collect API responses seem to have a number of different structures so in the chance we get a response that doesn't have either types we won't run into an exception trying to reference `statusCode` of `nil`.

curiousepic: I'm not sure I see how it would hit an exception in that case. Won't it just kick to `else "No error code available"` if there's no `statusCode`? Testing seems to confirm that.

curiousepic: The first `if` is already in that position. I'd be fine adding it across the board though; kind of wondering why I don't see it in more places.

davidsantoso: > I'm not sure I see how it would hit an exception in that case.

So, `response["statusOutput"]` would be `nil` if that key doesn't exist at the top level of the response hash. Which means it'd be the same as `nil["statusCode"]` which would raise the exception.

> The first if is in the same position.

Fortunately this is slightly different since if you reference a nonexistent key in a hash, `nil` is returned-

```ruby
➜ irb
irb(main):001:0> hash = {a: 1}
=> {:a=>1}
irb(main):002:0> hash[:b]
=> nil
```

> kind of wondering why I don't see it in more places.

I've been using it anytime I'm going more than one level down in a hash. Purely for the sake of avoiding a no method error on `nil` since getting nil value is better than an exception being raised in the middle of parsing out a transaction response.

curiousepic: Gotcha 👍

davidsantoso:
@curiousepic I'm not sure what's up the Travis 🤔 Looking at the log it _may_ be a Travis issue. But I can kick it off again to check.

davidsantoso:
@curiousepic Travis issues should now be fixed (54cf5c538e16c18d869f2011730fea4ecd1a8b16). Mind rebasing on `master` and updating this branch? That'll kick off Travis again.

curiousepic:
@davidsantoso Good to go.

davidsantoso:
@curiousepic thanks!

PR #2371, opened 2017-03-23
Respect Net::Http expectations for proxies

techbelly:
Net::Http has a slightly odd constructor semantics that aren't
immediately obvious. If no proxy parameters are sent then, by default, it
looks up the proxy from the standard HTTP_PROXY/http+proxy environment 
variables. To set no proxy at all, ignoring the environment variable, you can pass nil 
as the p_address parameter.

ActiveMerchant::Connection was always passing proxy parameters 
to Net::Http.new even when a proxy had not been set, meaning that 
the default environment variable configuration behaviour was never respected.

This commit ensures that if you do not explicitly set proxy parameters
on your Gateway the default behaviour of Net::Http to retrieve proxy 
settings from the environment applies. Proxy parameters can still explicitly
set to nil to avoid environment variable configuration.

### DIFF
--- a/lib/active_merchant/connection.rb
+++ b/lib/active_merchant/connection.rb
@@ -44,7 +44,7 @@ module ActiveMerchant
       @max_retries  = MAX_RETRIES
       @ignore_http_status = false
       @ssl_version = nil
-      @proxy_address = nil
+      @proxy_address = :DEFAULT
       @proxy_port = nil
     end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/connection.rb
```diff
@@ -94,7 +94,12 @@ def request(method, body, headers = {})
 
     private
     def http
-      http = Net::HTTP.new(endpoint.host, endpoint.port, proxy_address, proxy_port)
+
+      http = if proxy_address == :DEFAULT
```
bdewater: If I'm reading [the docs and the source](https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html#method-c-new) right, can't we just have `@proxy_address = :ENV` by default and have Net::HTTP take care of it?

bbergstrom: That is my understanding as well.

bbergstrom: @bdewater Created https://github.com/activemerchant/active_merchant/pull/2800 with the revised approach.  Please review.

PR #2380, opened 2017-03-29
Cybersource: Add xml parser exception handling

shasum:
@davidsantoso 

Rescue `REXML` parser exceptions due to malformed XML.

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -686,6 +686,9 @@ module ActiveMerchant #:nodoc:
           response = parse(ssl_post(test? ? self.test_url : self.live_url, build_request(request, options)))
         rescue ResponseError => e
           response = parse(e.response.body)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -686,6 +686,9 @@ def commit(request, action, amount, options)
           response = parse(ssl_post(test? ? self.test_url : self.live_url, build_request(request, options)))
         rescue ResponseError => e
           response = parse(e.response.body)
+        rescue REXML::ParseException => e
+          response = {}
+          response[:message] = e.to_s
```
davidsantoso: @shasum looks good- just FYI when I merge this in I'll probably merge these into a single line like-

```ruby
response = { message: e.to_s }
```

shasum: Awesome!

_muscle memory, making it 2 when 1 should suffice_ :)

PR #2381, opened 2017-03-30
Forte: syntax fix

whitby3001:
Fix for a small syntax error


jasonwebster:
This feels like it's worth a quick regression unit test, eh?

bpollack:
Thanks for this, @whitby3001 ! I'll rebase and merge.

PR #2389, opened 2017-04-04
TransFirst Transaction Express: Support ACH

curiousepic:
This adds support for purchase, refund, and void of eChecks.
eCheck general credit transactions ("blind credit") are
strictly controlled by TransFirst and are not included in
this implementation. It also cleans up some remote tests and
removed tests for "partial purchases", which aren't actually
a thing.

29 tests, 99 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@davidsantoso to review and merge


davidsantoso:
@curiousepic thanks!

PR #2390, opened 2017-04-05
rebrand gateway IPP to Bambora

InfraRuby:
This commit rebrands gateway IPP to Bambora. Note there is now an optional :account option that will need to be supplied through Shopify's interface.


bruno:
Wouldn't it be better to add Bambora as a new gateway and make a note in the `Ipp.rb` gateway implementation that it's now deprecated and people using it should move to Bambora? 

Just because renaming it will cause issues for users, like ourselves, that depend on the `ipp.rb` implementation if suddenly it's gone ... ?

InfraRuby:
@bruno: how about `IppGateway = BamboraGateway` in `ipp.rb`?

davidsantoso:
@InfraRuby I think @bruno makes a great suggestion. If you're still interested in a Bambora gateway adapter, feel free to submit a PR that includes that as a new gateway and a comment that Ipp has been deprecated/renamed.

InfraRuby:
@davidsantoso: ok, new PR for Bambora gateway: https://github.com/activemerchant/active_merchant/pull/2394

PR #2394, opened 2017-04-07
Add Bambora gateway

InfraRuby:
This commit adds a new gateway, Bambora, previously known as IPP, and adds a deprecation message to the IPP gateway.

### DIFF
--- a/lib/active_merchant/billing/gateways/bambora.rb
+++ b/lib/active_merchant/billing/gateways/bambora.rb
@@ -2,15 +2,15 @@ require "nokogiri"
 
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
-    class IppGateway < Gateway
-      self.live_url = 'https://www.ippayments.com.au/interface/api/dts.asmx'
-      self.test_url = 'https://demo.ippayments.com.au/interface/api/dts.asmx'
+    class BamboraGateway < Gateway
+      self.live_url = 'https://www.bambora.co.nz/interface/api/dts.asmx'
+      self.test_url = 'https://demo.bambora.co.nz/interface/api/dts.asmx'
 
-      self.supported_countries = ['AU']
+      self.supported_countries = ['AU', 'NZ']
       self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :jcb]
 
-      self.homepage_url = 'http://www.ippayments.com.au/'
-      self.display_name = 'IPP'
+      self.homepage_url = 'http://www.bambora.com/'
+      self.display_name = 'Bambora'
 
       self.money_format = :cents
 
@@ -22,7 +22,6 @@ module ActiveMerchant #:nodoc:
       }
 
       def initialize(options={})
-        ActiveMerchant.deprecated("IPP gateway is now named Bambora")
         requires!(options, :username, :password)
         super
       end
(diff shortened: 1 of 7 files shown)


lib/active_merchant/billing/gateways/bambora.rb
```diff
@@ -0,0 +1,179 @@
+require "nokogiri"
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class BamboraGateway < Gateway
+      self.live_url = 'https://www.bambora.co.nz/interface/api/dts.asmx'
+      self.test_url = 'https://demo.bambora.co.nz/interface/api/dts.asmx'
```
jknipp: Are the url's country specific? https://dev-apac.bambora.com/checkout/guides/get-up-and-running/migration-to-production

jknipp:
@molbrown Should this be renamed to reflect that this gateway is in support of Bambora Asia Pacific, rather than implying generic support of Bambora gateways?

lib/active_merchant/billing/gateways/bambora.rb
molbrown: This was not spotted in the internal reviews, but I'm confirming with a technical contact that it will support both countries. Looking closer, the correct URL should be  https://www.bambora.co.nz/access/index.aspx

molbrown:
@jknipp I can only find evidence of Bambora Ready being an older version/ something not supported anymore. I agree that this should be named descriptively, but I'm not sure how (maybe `Bambora Asia Pacific`?). The Beanstream gateway represents Bambora's North America region.

I am working in a branch off of this to add `void` and `store` and will open a separate PR, to preserve authorship.

jknipp:
That or `Bambora APAC`? Not sure if it is best to use the acronym or not here.

lib/active_merchant/billing/gateways/bambora.rb
```diff
@@ -0,0 +1,179 @@
...
+        transcript.
+          gsub(%r((<CardNumber>)[^<]+(<))i, '\1[FILTERED]\2').
+          gsub(%r((<CVN>)[^<]+(<))i, '\1[FILTERED]\2').
+          gsub(%r((<Password>)[^<]+(<))i, '\1[FILTERED]\2')
+      end
+
+      private
+
+      def add_credentials(xml)
+        username, account_number = options[:username].split(":")
```
curiousepic: I'd prefer these being separate credentials, unless they are always presented in this concatenated form.

curiousepic:
@molbrown @jknipp I think we would prefer request contents to be assembled before being passed to the commit method, rather than assembling them within a block. It would not take much refactoring.

jknipp:
@curiousepic For my education, what lines are you referring to?

molbrown:
My understanding of that is to move the role of `new_submit_xml`, called on line 129 inside of `commit`, to take place before the `commit` method. Or at least make the request build before `commit` so it is ready to be passed in. (@curiousepic please correct me if this is off...)

curiousepic:
I actually take back my original request after realizing this is how at least BlueSnap is built. I'm actually only referring to how each action starts by calling commit with a block, rather than building up a param hash and passing that to commit.

molbrown:
Cybersource and AuthorizeNet also build requests in blocks, and it looks like Payflow first builds the request then passes it in. So, there's not really a standard pattern it seems.

molbrown:
Continued in #3145 to include needed updates.

PR #2395, opened 2017-04-11
SagePay: Support Repeat transactions

curiousepic:
Also deactivates the remote test for a Maestro card since it
is not supported for accounts set to use GBP currency, and
prevents an order_id conflict for another test.

35 tests, 95 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@davidsantoso to review and merge.

### DIFF
--- a/test/remote/gateways/remote_sage_pay_test.rb
+++ b/test/remote/gateways/remote_sage_pay_test.rb
@@ -159,12 +159,13 @@ class RemoteSagePayTest < Test::Unit::TestCase
     assert !response.authorization.blank?
   end
 
-  def test_successful_maestro_purchase
-    assert response = @gateway.purchase(@amount, @maestro, @options)
-    assert_success response
-    assert response.test?
-    assert !response.authorization.blank?
-  end
+  # Maestro is not available for GBP
+  # def test_successful_maestro_purchase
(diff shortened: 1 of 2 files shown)


test/remote/gateways/remote_sage_pay_test.rb
```diff
@@ -159,12 +159,13 @@ def test_successful_visa_purchase
...
-    assert response.test?
-    assert !response.authorization.blank?
-  end
+  # Maestro is not available for GBP
+  # def test_successful_maestro_purchase
+  #   assert response = @gateway.purchase(@amount, @maestro, @options)
+  #   assert_success response
+  #   assert response.test?
+  #   assert !response.authorization.blank?
+  # end
```
davidsantoso: @curiousepic wondering if we should uncomment these? Or if they're not needed just delete the test.

curiousepic: @davidsantoso Theoretically it works and is needed for other test accounts not using GBP, so I thought I'd leave it in. I'm fine with uncommenting it and having one failing test, though.

davidsantoso: @curiousepic got it. Thanks for the clarification. We can leave it in.

PR #2396, opened 2017-04-18
Beanstream: Map iso province codes for US and CA

shasum:
Beanstream expects char(2) iso codes for province

image

ref: https://developer.beanstream.com/docs/references/merchant_API/v1-0-2/
ref: https://en.wikipedia.org/wiki/ISO_3166-2:CA
ref: https://en.wikipedia.org/wiki/ISO_3166-2:US

---

A few remote tests fail with available credentials. Hence, pasting a `purchase` run which is most relevant in this case.
```
$ bundle exec rake test:remote TEST=test/remote/gateways/remote_beanstream_test.rb TESTOPTS="--name=test_successful_visa_purchase"
Started
.

Finished in 1.927778 seconds.
1 tests, 4 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
0.52 tests/s, 2.07 assertions/s
```

### DIFF
--- a/lib/active_merchant/billing/gateways/beanstream/beanstream_core.rb
+++ b/lib/active_merchant/billing/gateways/beanstream/beanstream_core.rb
@@ -59,6 +59,72 @@ module ActiveMerchant #:nodoc:
         :cancel => 'C'
       }
 
+      STATES = {
+        "ALBERTA" => "AB",
+        "BRITISH COLUMBIA" => "BC",
+        "MANITOBA" => "MB",
+        "NEW BRUNSWICK" => "NB",
+        "NEWFOUNDLAND AND LABRADOR" => "NL",
+        "NOVA SCOTIA" => "NS",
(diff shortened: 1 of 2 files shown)


shasum:
@davidsantoso

davidsantoso:
@shasum looks good. One small tweak- wondering if we can add back in a remote tests that uses a state/province that is already the correct 2 character code? Reason being that we want to ensure that if someone already passes in the correct 2 character code then it doesn't return `nil` if it can't find it in the `STATES` hash.

I do see that you have the `||` in there which is good and covers this case, but want to make sure we have a test that ensures that `||` stays in there and if that gets removed it'll fail.

shasum:
@davidsantoso sure. I did test that case in-place but was a bit lazy to add it additionally 😀  . Will be back with an update.

shasum:
@davidsantoso updated!

PR #2398, opened 2017-04-18
Braintree Blue: Refund unsettled payments via void

bizla:
**What:** Allows refunds of unsettled payments by voiding them instead. 

**How:** Issues a void call if (and only if) the refund fails specifically with [error code 91506](https://developers.braintreepayments.com/reference/general/validation-errors/all/ruby).

**Why:** Consider the scenario: 
1. A purchase payment is submitted.
1. ActiveMerchant responds with a successful response. 
1. A refund for the purchase is submitted.
1. ActiveMerchant responds with a failure response, because the purchase payment has not settled. 

This violates the expectation that successful ActiveMerchant purchases can be fully refunded. 
The gateway-specific details of what's required to make that happen are properly abstracted away with this patch. 

Note that since there's no such thing a partial void, this only applies to full refunds via a `full_refund` option passed into`#refund`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@
 * Checkout V2: Fix success_from not properly checking two possible success codes [davidsantoso]
 * SagePay: Support Repeat transactions [curiousepic] #2395
 * PayU LATAM: Fix incorrect capture method definition [davidsantoso]
+* Braintree Blue: Force refund of unsettled payments via void [bizla]
 
 == Version 1.64.0 (March 6, 2017)
 * Authorize.net: Allow settings to be passed for CIM purchases [fwilkins] #2300
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -42,6 +42,10 @@ module ActiveMerchant #:nodoc:
 
       self.display_name = 'Braintree (Blue Platform)'
 
+      ERROR_CODES = {
+        cannot_refund_if_unsettled: 91506
+      }
+
       def initialize(options = {})
         requires!(options, :merchant_id, :public_key, :private_key)
         @merchant_account_id = options[:merchant_account_id]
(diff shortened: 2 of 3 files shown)


pi3r:
My only concern is maybe it shouldn't be the responsibility of `#refund` to `void` in case of an error? Is there a precedent in AM where a `#refund` method voids a transaction?

bizla:
> My only concern is maybe it shouldn't be the responsibility of #refund to void in case of an error?

Ultimately, AM modules should be written to guarantee that successful purchases can be fully refunded and successful authorizations can be voided/refunded. 
In this case, I'm proposed we handle Braintree-specific error that exposes a detail of their payment state machine (settled vs. unsettled) which can't be resolved with the AM protocol; there is no `#settle_payment`, so the fact that we void in the case of a full refund is really just an implementation detail. 
Now, a consumer of this module can write, with:
```ruby
gateway = ActiveMerchant::BraintreeGateway.new(...)
```
Simply writing, with confidence: 
```ruby
purchase = gateway.purchase(...) 
gateway.refund(..., full_refund: true)
```
...knowing AM will do whatever wrangling is necessary to ensure the money ends up where it's supposed to. This is in contrast to handling the potential leakage of Braintree-specific situations like this: 
```ruby 
refund = gateway.refund(...)
unless refund.success? 
  if refund.params[:errors].any? { |_, v| v =~ ERROR_CODES[:braintree_cannot_refund_unsettled_payment] } 
    gateway.void(...) 
  end
end
```

> Is there a precedent in AM where a #refund method voids a transaction?

There hasn't been yet AFAIK. In this case, the fact that we're issuing a "void" is mostly an implementation detail. It just happens to be what's necessary to ensure that a successful-but-unsettled Braintree purchase is refunded/cancelled/undone/settled-as-cancelled/whatever :)

pi3r:
Thanks for the clear explanation, I see your point better now.

bizla:
ping @duff @ntalbott - this shouldn't interfere with anything on the Spreedly end but I'd like your stamp of approval before merging this :)

bizla:
Closed via https://github.com/activemerchant/active_merchant/commit/1ba258f4bc4029983e378f6be83e776b0ded9d2b.

PR #2399, opened 2017-04-20
Authorize.Net: Force refund of unsettled payments via void

bizla:
**What: This is the same situation as https://github.com/activemerchant/active_merchant/pull/2398.** 
Allows refunds of unsettled payments by voiding them instead. 

**How:** Issues a void call if (and only if) the refund fails specifically with [error code 54](http://developer.authorize.net/api/reference/dist/json/responseCodes.json) and the `full_refund` option is included.

**Why:** Consider the scenario: 
1. A purchase payment is submitted.
1. ActiveMerchant responds with a successful response. 
1. A refund for the purchase is submitted.
1. ActiveMerchant responds with a failure response, because the purchase payment has not settled. 

This violates the expectation that successful ActiveMerchant purchases can be fully refunded.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,7 +21,9 @@
 * SagePay: Support Repeat transactions [curiousepic] #2395
 * PayU LATAM: Fix incorrect capture method definition [davidsantoso]
 * Beanstream: Map ISO province codes for US and CA [shasum] #2396
+* Braintree Blue: Force refund of unsettled payments via void [bizla] #2398
 * Openpay: Support card points [shasum] #2401
+* Authorize.Net: Force refund of unsettled payments via void [bizla] #2399
 
 == Version 1.64.0 (March 6, 2017)
 * Authorize.net: Allow settings to be passed for CIM purchases [fwilkins] #2300
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -92,6 +92,7 @@ module ActiveMerchant
       APPLE_PAY_DATA_DESCRIPTOR = "COMMON.APPLE.INAPP.PAYMENT"
 
       PAYMENT_METHOD_NOT_SUPPORTED_ERROR = "155"
+      INELIGIBLE_FOR_ISSUING_CREDIT_ERROR = "54"
 
       def initialize(options={})
         requires!(options, :login, :password)
@@ -131,10 +132,17 @@ module ActiveMerchant
       end
 
       def refund(amount, authorization, options={})
-        if auth_was_for_cim?(authorization)
+        response = if auth_was_for_cim?(authorization)
           cim_refund(amount, authorization, options)
         else
           normal_refund(amount, authorization, options)
+        end
+
+        return response if response.success?
+        return response unless options[:force_full_refund_if_unsettled]
+
+        if response.params["response_reason_code"] == INELIGIBLE_FOR_ISSUING_CREDIT_ERROR
+          void(authorization, options)
         end
       end
 
--- a/test/unit/gateways/authorize_net_test.rb
+++ b/test/unit/gateways/authorize_net_test.rb
@@ -549,6 +549,20 @@ class AuthorizeNetTest < Test::Unit::TestCase
     assert_equal "The record cannot be found", refund.message
   end
 
+  def test_failed_refund_due_to_unsettled_payment
+    @gateway.expects(:ssl_post).returns(failed_refund_for_unsettled_payment_response)
+    @gateway.expects(:void).never
+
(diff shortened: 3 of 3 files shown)


pi3r:
small typo: I believe you meant with error code `54` in the description in the pr not `error code 91506`.

CHANGELOG
```diff
@@ -21,6 +21,8 @@
 * SagePay: Support Repeat transactions [curiousepic] #2395
 * PayU LATAM: Fix incorrect capture method definition [davidsantoso]
 * Beanstream: Map ISO province codes for US and CA [shasum] #2396
+* Braintree Blue: Force refund of unsettled payments via void [bizla] #2398
```
bizla: Forgot to add this on my last merge 😬

bizla:
ping @davidsantoso (or someone else from Spreedly 😄)

davidsantoso:
@bizla thinking about this a little more- wondering if the option could be more explicit? (I apologize considering the Braintree change was already merged in) Something like `auto_void_unsettled` instead of `full_refund`. Thinking mainly of the method signature being a little strange:

```ruby
gateway.refund(100, authorization, full_refund: true)
```

Where `full_refund` seems to be a little at odds to the first `100` amount parameter. Questions like why would I pass in that option if I'm already sending in the full amount? Or the opposite- why would I need that option if I'm wanting to refund a partial amount? Considering in the past that this has been considered out of scope for Active Merchant, it seems like being as explicit as possible might be good?

bizla:
@davidsantoso I was at odds with the naming of the option too. To answer your questions: 

> why would I pass in that option if I'm already sending in the full amount? ... Considering in the past that this has been considered out of scope for Active Merchant.

ActiveMerchant still doesn't know that the amount is the full amount unless you explicitly say so. That's why I tried to make it as optional as possible; you'll still get "but I'm not settled!" errors by default. The AM-consuming system is still ultimately responsible for setting that flag since it has the actual context. So, in those cases, we have to somehow communicate that what we're after is a full refund by any means necessary. 

> why would I need that option if I'm wanting to refund a partial amount?

Good question... maybe `force_full_refund_if_unsettled` would be better in general? That way the interface is more clear that you're aware of possible unsettlement and that you want AM to do whatever's necessary to make a full refund happen.
How about this?
```ruby
gateway.refund(100, authorization, force_full_refund_if_unsettled: true)
```
(I didn't want to mention `void` specifically because forcing a full refund may not necessarily be done by voiding. What's involved to force a full refund should remain an AM implementation detail.)

davidsantoso:
@bizla sounds good to me. Thanks for the explanations!

bizla:
Thanks @davidsantoso - I'll merge this on the CI ✅

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -131,11 +132,18 @@ def capture(amount, authorization, options={})
...
+        response = if auth_was_for_cim?(authorization)
           cim_refund(amount, authorization, options)
         else
           normal_refund(amount, authorization, options)
         end
+
+        return response if response.success?
+        return response unless options[:force_full_refund_if_unsettled]
+
+        if response.params["response_reason_code"] == INELIGIBLE_FOR_ISSUING_CREDIT_ERROR
```
emilienoel: Can you add a test for when the response_reason_code is not that?

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -131,11 +132,18 @@ def capture(amount, authorization, options={})
...
       def refund(amount, authorization, options={})
-        if auth_was_for_cim?(authorization)
+        response = if auth_was_for_cim?(authorization)
           cim_refund(amount, authorization, options)
         else
           normal_refund(amount, authorization, options)
         end
+
+        return response if response.success?
+        return response unless options[:force_full_refund_if_unsettled]
```
emilienoel: Are you going to change this param for braintree as well?

lib/active_merchant/billing/gateways/authorize_net.rb
bizla: `test_failed_refund` covers that.

lib/active_merchant/billing/gateways/authorize_net.rb
bizla: Already up @ https://github.com/activemerchant/active_merchant/pull/2403

lib/active_merchant/billing/gateways/authorize_net.rb
emilienoel: Does the test assert that void is never called?

bizla: Yep, because there's an `expects(:ssl_post)` 
If void (or anything) was called after the refund response, the test would 💥

PR #2400, opened 2017-04-20
iVeri: Add gateway support

curiousepic:
18 tests, 44 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

@shasum and/or @davidsantoso to review - particularly request building logic (requires specific ordering of elements) and parsing.

### DIFF
--- a/lib/active_merchant/billing/gateways/iveri.rb
+++ b/lib/active_merchant/billing/gateways/iveri.rb
@@ -1,0 +1,240 @@
+require 'nokogiri'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class IveriGateway < Gateway
+      self.live_url = self.test_url = 'https://portal.nedsecure.co.za/iVeriWebService/Service.asmx'
+
+      self.supported_countries = ['US', 'ZA', 'GB']
+      self.default_currency = 'ZAR'
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express]
+
+      self.homepage_url = 'http://www.iveri.com'
+      self.display_name = 'iVeri'
+
+      def initialize(options={})
+        requires!(options, :app_id, :cert_id)
+        super
+      end
+
+      def purchase(money, payment_method, options={})
+        post = build_vxml_request('Debit', options) do |xml|
+          add_auth_purchase_params(xml, money, payment_method, options)
+        end
+
+        commit(post)
+      end
+
+      def authorize(money, payment_method, options={})
+        post = build_vxml_request('Authorisation', options) do |xml|
+          add_auth_purchase_params(xml, money, payment_method, options)
+        end
+
+        commit(post)
+      end
+
+      def capture(money, authorization, options={})
+        post = build_vxml_request('Debit', options) do |xml|
+          add_authorization(xml, authorization, options)
+        end
+
+        commit(post)
+      end
+
+      def refund(money, authorization, options={})
+        post = build_vxml_request('Credit', options) do |xml|
+          add_amount(xml, money, options)
+          add_authorization(xml, authorization, options)
+        end
+
+        commit(post)
+      end
+
+      def void(authorization, options={})
+        post = build_vxml_request('Void', options) do |xml|
+          add_authorization(xml, authorization, options)
+        end
+
+        commit(post)
+      end
+
+      def verify(credit_card, options={})
+        authorize(0, credit_card, options)
+      end
+
+      def verify_credentials
+        void = void('', options)
+        return true if void.message ==  'Missing OriginalMerchantTrace'
+        false
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((CertificateID=\\\")[^\\]*), '\1[FILTERED]').
+          gsub(%r((&lt;PAN&gt;)[^&]*), '\1[FILTERED]').
+          gsub(%r((&lt;CardSecurityCode&gt;)[^&]*), '\1[FILTERED]')
+      end
+
+      private
+
+      def build_xml_envelope(vxml)
+        builder = Nokogiri::XML::Builder.new(:encoding => 'UTF-8') do |xml|
+          xml[:soap].Envelope 'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance', 'xmlns:xsd' => 'http://www.w3.org/2001/XMLSchema', 'xmlns:soap' => 'http://schemas.xmlsoap.org/soap/envelope/' do
+            xml[:soap].Body do
+              xml.Execute 'xmlns' => 'http://iveri.com/' do
+                xml.validateRequest 'true'
+                xml.protocol 'V_XML'
+                xml.protocolVersion '2.0'
+                xml.request vxml
+              end
+            end
+          end
+        end
+
+        builder.to_xml
+      end
+
+      def build_vxml_request(action, options)
+        builder = Nokogiri::XML::Builder.new do |xml|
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/iveri.rb
```diff
@@ -0,0 +1,231 @@
...
+            end
+          end
+        end
+
+        builder.doc.root.to_xml
+      end
+
+      def add_params_in_order(post, action, money, pm, options)
+        post.Amount(amount(money))                    if money
+        post.Currency(options[:currency])             if [:purchase, :capture, :refund].include? action
```
curiousepic: Good catch, this was an oversight; best to pass it along since it's accepted.

lib/active_merchant/billing/gateways/iveri.rb
```diff
@@ -0,0 +1,231 @@
...
+      def commit(post)
+        parsed = parse(ssl_post(live_url, build_xml_envelope(post), headers(post)))
+        succeeded = success_from(parsed)
+
+        Response.new(
+          succeeded,
+          message_from(parsed, succeeded),
+          parsed,
+          authorization: authorization_from(parsed),
+          avs_result: AVSResult.new(code: parsed["some_avs_response_key"]),
```
curiousepic: iVeri doesn't support AVS or seem to return CVV "results"(?), so these were untouched defaults. I'm not sure what's best to replace these with in that case... (some other gateways still have these defaults)

shasum: @curiousepic In the past I've totally skipped them when undefined.

lib/active_merchant/billing/gateways/iveri.rb
```diff
@@ -0,0 +1,231 @@
...
+            xml.Transaction('ApplicationID' => @options[:app_id], 'Command' => action, 'Mode' => mode) do
+              yield(xml)
+            end
+          end
+        end
+
+        builder.doc.root.to_xml
+      end
+
+      def add_params_in_order(post, action, money, pm, options)
```
curiousepic: Hah, I originally shied away from this because there were a lot more potentially accepted params in between each of these that ended up not being required, but which obscured the grouping that is now obvious after removing them! I guess ExpiryDate being part of invoice rather than payment method is still an exception...

curiousepic: Hmm... This still ends up being pretty messy in practice. Yet another alternative is to have separate `add_action_params` methods for each action (and possibly sub-types within) them. It's hard to know what to optimize for here. Anyhow I'd like @davidsantoso's perspective as well...

shasum: Yep. The jumble of fields we start out with influence our grouping and structure in totally different ways. I guess this is where gateways with good docs and clear examples stand out by making things as clear as possible right at the beginning.

>I guess ExpiryDate being part of invoice rather than payment method is still an exception

⬛️  🐑

curiousepic: Hmm, this isn't terrible, with some guard clauses...

```ruby
def add_params_in_order(post, action, money, payment_method, options)
  add_expiration(post, payment_method, options)
  add_references(post, action, payment_method, options)
  add_credit_card(post, payment_method, options)
end

def add_amount(post, money, options)
  return unless money
  post.Amount(amount(money))
  post.Currency(options[:currency])
end

def add_expiration(post, payment_method, options)
  return unless payment_method.respond_to?(:month)
  post.ExpiryDate("#{format(payment_method.month, :two_digits)}#{payment_method.year}")
end

def add_references(post, action, payment_method, options)
    post.MerchantReference(split_auth(payment_method)[2])
  else
    post.MerchantReference(options[:order_id] || generate_unique_id)
  end
  post.TransactionIndex(split_auth(payment_method)[1])  if [:capture, :refund].include? action
  post.OriginalRequestID(split_auth(payment_method)[0]) if [:void].include? action
end

def add_credit_card(post, payment_method, options)
  return unless payment_method.respond_to?(:verification_value)
  post.CardSecurityCode(payment_method.verification_value)
  post.PAN(payment_method.number)
end
```

shasum: Separating out `expiration` to it's own method is definitely better 👍 

`if payment_method.is_a?(String)` -- pattern is what I'm bit unsure about as it's commonly used to recognize stored tokens and may/may not cause trouble down the line when implementing `store`.

davidsantoso: Personally I like the `return unless ...` pattern mainly because it's more explicit. Basically, I know that none of the fields below the guard clause are getting added unless some piece of data comes through.

curiousepic: After some sifting I was able to break things down into a more idiomatic structure with fewer conditionals. The potential downside is it may be less easy to extend in the future if we need to fit in additional request fields or modes. Updating the PR...

curiousepic:
@davidsantoso @shasum Updated to more idiomatic request building structure.

lib/active_merchant/billing/gateways/iveri.rb
```diff
@@ -0,0 +1,241 @@
...
+
+      def add_authorization(post, authorization, options)
+        post.MerchantReference(split_auth(authorization)[2])
+        post.TransactionIndex(split_auth(authorization)[1])
+        post.OriginalRequestID(split_auth(authorization)[0])
+      end
+
+      def add_credit_card(post, credit_card, options)
+        post.ExpiryDate("#{format(credit_card.month, :two_digits)}#{credit_card.year}")
+        post.MerchantReference(options[:order_id] || generate_unique_id)
```
shasum: Looks clean overall. Could this one item `MerchantReference` be separated out to keep clear semantics. E.g.: `def add_new_merchant_reference` ( the new part is to market it different from the one extracted from authorization).

shasum: 🤔  I forget order may be important and that's a reason it's living within `add_credit_card`.

lib/active_merchant/billing/gateways/iveri.rb
```diff
@@ -0,0 +1,241 @@
...
+        builder.doc.root.to_xml
+      end
+
+      def add_amount(post, money, options)
+        post.Amount(amount(money))
+        post.Currency(options[:currency] || default_currency)
+      end
+
+      def add_payment_method(post, payment_method, options)
+        if payment_method.is_a?(String)
```
shasum: Just thinking what would be the impact here when someone revisits to add support for stored cards. At that point, they may have to add `action` as a param to `add_authorization` to do the right thing inside. Should be okay I guess.

curiousepic:
@davidsantoso Yet another update; thanks for the feedback @shasum, I think I addressed everything!

shasum:
@curiousepic Awesome, the refactor looks very clean 👍 

The changes might have accidentally broken unit tests at couple spots. I'll mark them in-line.

curiousepic:
@shasum Good catch, I've corrected them. Thanks!

lib/active_merchant/billing/gateways/iveri.rb
```diff
@@ -0,0 +1,240 @@
+require 'nokogiri'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class IveriGateway < Gateway
+      self.live_url = 'https://portal.nedsecure.co.za/iVeriWebService/Service.asmx'
```
davidsantoso: Guessing their test URL is the same as their live URL? I've seen this pattern in other adapters-

`self.live_url = self.test_url = 'https://example.com'`

Wondering if we should follow suit for the sake of consistency.

PR #2405, opened 2017-04-26
ProPay: Add gateway

davidsantoso:
Adding the ProPay gateway. Just to note- still waiting on the production URL. I'll update the PR once it's given before merging in.

**Remote Test Run**
```
➜ ruby -Itest test/remote/gateways/remote_pro_pay_test.rb
Loaded suite test/remote/gateways/remote_pro_pay_test
Started
...................

Finished in 58.230136 seconds.
19 tests, 49 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
0.33 tests/s, 0.84 assertions/s
```

**Unit Test Run**
```
➜ ruby -Itest test/unit/gateways/pro_pay_test.rb
Loaded suite test/unit/gateways/pro_pay_test
Started
................

Finished in 0.069866 seconds.
16 tests, 68 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
229.01 tests/s, 973.29 assertions/s
```

### DIFF
--- a/lib/active_merchant/billing/gateways/pro_pay.rb
+++ b/lib/active_merchant/billing/gateways/pro_pay.rb
@@ -1,0 +1,326 @@
+require 'nokogiri'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class ProPayGateway < Gateway
+      self.test_url = 'https://xmltest.propay.com/API/PropayAPI.aspx'
+      self.live_url = 'https://epay.propay.com/api/propayapi.aspx'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'https://www.propay.com/'
+      self.display_name = 'ProPay'
+
+      STATUS_RESPONSE_CODES = {
+        "00" => "Success",
+        "20" => "Invalid username",
+        "21" => "Invalid transType",
+        "22" => "Invalid Currency Code",
+        "23" => "Invalid accountType",
+        "24" => "Invalid sourceEmail",
+        "25" => "Invalid firstName",
+        "26" => "Invalid mInitial",
+        "27" => "Invalid lastName",
+        "28" => "Invalid billAddr",
+        "29" => "Invalid aptNum",
+        "30" => "Invalid city",
+        "31" => "Invalid state",
+        "32" => "Invalid billZip",
+        "33" => "Invalid mailAddr",
+        "34" => "Invalid mailApt",
+        "35" => "Invalid mailCity",
+        "36" => "Invalid mailState",
+        "37" => "Invalid mailZip",
+        "38" => "Invalid dayPhone",
+        "39" => "Invalid evenPhone",
+        "40" => "Invalid ssn",
+        "41" => "Invalid dob",
+        "42" => "Invalid recEmail",
+        "43" => "Invalid knownAccount",
+        "44" => "Invalid amount",
+        "45" => "Invalid invNum",
+        "46" => "Invalid rtNum",
+        "47" => "Invalid accntNum",
+        "48" => "Invalid ccNum",
+        "49" => "Invalid expDate",
+        "50" => "Invalid cvv2",
+        "51" => "Invalid transNum and/or Unable to act perform actions on transNum due to funding",
+        "52" => "Invalid splitNum",
+        "53" => "A ProPay account with this email address already exists AND/OR User has no account number",
+        "54" => "A ProPay account with this social security number already exists",
+        "55" => "The email address provided does not correspond to a ProPay account.",
+        "56" => "Recipient’s email address shouldn’t have a ProPay account and does",
+        "57" => "Cannot settle transaction because it already expired",
+        "58" => "Credit card declined",
+        "59" => "Invalid Credential or IP address not allowed",
+        "60" => "Credit card authorization timed out; retry at a later time",
+        "61" => "Amount exceeds single transaction limit",
+        "62" => "Amount exceeds monthly volume limit",
+        "63" => "Insufficient funds in account",
+        "64" => "Over credit card use limit",
+        "65" => "Miscellaneous error",
+        "66" => "Denied a ProPay account",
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/pro_pay.rb
```diff
@@ -0,0 +1,327 @@
...
+    class ProPayGateway < Gateway
+      self.test_url = 'https://xmltest.propay.com/API/PropayAPI.aspx'
+      self.live_url = 'https://example.com/live'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.money_format = :cents
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'http://www.example.net/'
```
curiousepic: https://www.propay.com ?

lib/active_merchant/billing/gateways/pro_pay.rb
```diff
@@ -0,0 +1,327 @@
...
+          xml.transType "06"
+        end
+
+        commit(request)
+      end
+
+      def refund(money, authorization, options={})
+        request = build_xml_request do |xml|
+          xml.transNum authorization
+          xml.amount money if money
```
curiousepic: Should this be 'amount(money)'?

lib/active_merchant/billing/gateways/pro_pay.rb
```diff
@@ -0,0 +1,327 @@
...
+          xml.transType "05"
+        end
+
+        commit(request)
+      end
+
+      def capture(money, authorization, options={})
+        request = build_xml_request do |xml|
+          xml.amount money if money
+          xml.transNum authorization
```
curiousepic: Not sure how I feel about having both these direct additions to the xml in the action methods, _and_ the one-line `add_account` and `add_recurring` methods extracted. Not a big deal though.

lib/active_merchant/billing/gateways/pro_pay.rb
```diff
@@ -0,0 +1,327 @@
...
+          add_account(xml, options)
+          xml.transType "35"
+        end
+
+        commit(request)
+      end
+
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
```
curiousepic: Doesn't accept 0-value `authorize`s?

lib/active_merchant/billing/gateways/pro_pay.rb
```diff
@@ -0,0 +1,327 @@
...
+      end
+
+      def add_invoice(xml, money, options)
+        xml.amount amount(money)
+        xml.currencyCode options[:currency] || currency(money)
+        xml.invNum options[:order_id] || SecureRandom.hex(25)
+      end
+
+      def add_recurring(xml, options)
+        xml.recurringPayment options[:recurring_payment]
```
curiousepic: Any best practice for options passed to a gateway that are boolean being `true/false` vs the value of what's actually passed to the gateway? Or are there more options than `Y/N`?

lib/active_merchant/billing/gateways/pro_pay.rb
davidsantoso: yep 😞

lib/active_merchant/billing/gateways/pro_pay.rb
davidsantoso: It should! Good catch.

lib/active_merchant/billing/gateways/pro_pay.rb
davidsantoso: Hmmm I didn't try it actually. I've actually always gone off of the `100` that comes with the generator. I'll give it a shot and see what happens.

lib/active_merchant/billing/gateways/pro_pay.rb
davidsantoso: I've personally always thought it's better to pass the same value as what's in the gateway docs. Seems more explicit to me instead of relying on the underly adapter to do the right thing depending on `true`/`false`. That being said, I don't think there's actually a convention around this yet...

curiousepic: I think that does make sense; I'll try to remember to follow that in the future.

lib/active_merchant/billing/gateways/pro_pay.rb
davidsantoso: Hmmm. Sometimes I think it's ok to have the method just add the single value instead of always using any extracted methods. But, the more I think about it, it's probably better for me to keep using to extracted methods.

lib/active_merchant/billing/gateways/pro_pay.rb
davidsantoso: Just tried `0` auth during `verify` and got an `Invalid amount` error. 😿

davidsantoso:
Alrighty, updates made. Thanks for the second set of eyes @curiousepic. Just waiting on the production URL from ProPay now...

davidsantoso:
Live URL added. Just waiting on a 👍!

curiousepic:
@shasum Want to give a once-over and 👍 ?

lib/active_merchant/billing/gateways/pro_pay.rb
```diff
@@ -0,0 +1,327 @@
...
+          xml.addr address[:address1]
+          xml.aptNum address[:address2]
+          xml.city address[:city]
+          xml.state address[:state]
+          xml.zip address[:zip]
+        end
+      end
+
+      def add_account(xml, options)
+        xml.accountNum options[:account_num]
```
davidsantoso: @shasum great question! Yes it is. ProPay is a payment gateway for platforms so for each transaction you need to specify the merchant (on your platform) that you want to actually receive the funds. Similar to WePay.

davidsantoso:
Thanks @shasum and @curiousepic!

lib/active_merchant/billing/gateways/pro_pay.rb
shasum: Cool 👍

PR #2407, opened 2017-04-27
Add Rails 5.1 support

jhawthorn:
[Rails 5.1 final is out](http://weblog.rubyonrails.org/2017/4/27/Rails-5-1-final/)!

This fixes the CI setup for the Rails 5.1 final release and loosens the gemspec dependency to support it.

### DIFF
--- a/activemerchant.gemspec
+++ b/activemerchant.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
 
   s.has_rdoc = true if Gem::VERSION < '1.7.0'
 
-  s.add_dependency('activesupport', '>= 3.2.14', '< 5.1')
+  s.add_dependency('activesupport', '>= 3.2.14', '< 5.2')
   s.add_dependency('i18n', '>= 0.6.9')
   s.add_dependency('builder', '>= 2.1.2', '< 4.0.0')
   s.add_dependency('nokogiri', "~> 1.4")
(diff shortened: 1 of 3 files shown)


activemerchant.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
 
   s.has_rdoc = true if Gem::VERSION < '1.7.0'
 
-  s.add_dependency('activesupport', '>= 3.2.14', '< 5.1')
+  s.add_dependency('activesupport', '>= 3.2.14', '< 5.2')
```
jasonwebster: 🤔 I wonder why we bother with the upper bound limit.

jhawthorn: :wave: :smile:  Hi Jason

I always feel there should be _some_ upper bound, just because there's no way to retroactively take it away in the future.

How about `'< 6'`? Or even better `'< 6.x'`, which will also forbid any `'6.0.0.alphas'`

jhawthorn:
I've pushed a commit loosening the dependency to `'< 6.x'`

activemerchant.gemspec
jasonwebster: Well, what do you think is best? I was mostly just asking for opinions. 

As someone integrating ActiveMerchant, I'd personally find it frustrating that I couldn't test edge Rails versions without some dependency monkeying.

I like what you've proposed as a solution to that.

richpeck:
Would be good to get this merged

PR #2414, opened 2017-05-04
Opp: Clean up options

shasum:
Lotsa automatically trimmed trailing spaces showing up in the diff. Sorry about that. Minor changes to keep options in-line with other AM adapters.

Some portions of the implementation is different from typical adapters. The way stored tokens are handled needs some change. I guess that can happen when there's a need for `store` to be added. Right now focus is on minimal changes.

In addition to the diffs, please review transaction success criteria as defined in
`def success_from(raw_response)` 
It's deciding based on http response code. Should it actually be looking at `result.code` within the response. I can make that change once reviewer confirms.

ref: https://docs.acaptureservices.com/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@
 * WePay: Add scrub method [shasum] #2406
 * iVeri: Add gateway support [curiousepic] #2400
 * iVeri: Support 3DSecure data fields [davidsantoso] #2412
+* Opp: Fix transaction success criteria and clean up options [shasum] #2414
 
 == Version 1.65.0 (April 26, 2017)
 * Adyen: Add Adyen v18 gateway [adyenpayments] #2272
--- a/lib/active_merchant/billing/gateways/opp.rb
+++ b/lib/active_merchant/billing/gateways/opp.rb
@@ -3,19 +3,19 @@ module ActiveMerchant #:nodoc:
     class OppGateway < Gateway
 # = Open Payment Platform
     #
-    #  The Open Payment Platform includes a powerful omni-channel transaction processing API, 
-    #   enabling you to quickly and flexibly build new applications and services on the platform. 
-    #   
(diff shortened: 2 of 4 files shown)


davidsantoso:
>  Should it actually be looking at result.code

@shasum we should definitely be relying on a field in the response body instead of an HTTP code in general. Unless the Opp documentation is explicit on using the HTTP status code. HTTP status codes can be a bit misleading in that we get a 200 response, but the transaction was a failure. The 200 only denotes that a request was received, processed, and a response given.

shasum:
@davidsantoso Absolutely! I was thinking along the same lines skimming through their documentation. Taking a more closer look it explicitly says to look beyond http status codes for transaction status. Result codes in the response body seem the right place to look for pass/fail info.

>For each request you send to our API the HTTP status code of the response will already tell you the basic result.

>200 - successful request

>307 - temporary redirect

>400 - bad request. This might either point to e.g. invalid parameters or values sent. It's also returned if the payment failed e.g. because the acquirer declined.

>403 - incorrect authentication information, e.g. one of the authentication.* parameters is wrong - please check them or contact us for correct parameters

>404 - requested resource or endpoint is not found. I.e. endpoint/url doesn't exist. This can also be caused by typos like POST /v1/paymnets instead of payments or wrong IDs like GET /v1/payments/{id} where no payment with {id} exists.

>For payments you'll want more fine grained information to find out why a payment failed. You're getting this information in the result codes.

shasum:
@davidsantoso I'll go ahead change the `success_from` method. Appreciate if you could give a quick 👍  on rest of the changes.

shasum:
```
$ RUBYOPT=W0 be rake test:remote TEST=test/remote/gateways/remote_opp_test.rb 

Finished in 29.771993 seconds.
100% passed
0.50 tests/s, 1.61 assertions/s
```

PR #2416, opened 2017-05-05
Elavon: Support custom fields

curiousepic:
Also updates test card number to work with changes on Elavon's end, and
cleans up reference comments.

Remote:
23 tests, 102 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
28 tests, 138 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@shasum and/or @davidsantoso to review - this isn't the prettiest way to implement these fields, but they require not having `ssl_` prepended to them like all other request fields; any suggestions that don't require rewriting how requests are built (unless that's the best way forward...)?

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -2,33 +2,6 @@ require 'active_merchant/billing/gateways/viaklix'
 
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
-    # = Elavon Virtual Merchant Gateway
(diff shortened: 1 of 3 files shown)


curiousepic:
@shasum Yeah I considered that initially. I thought this method would be slightly more resilient since it doesn't mutate, it just keeps the original values along for the ride.

shasum:
@curiousepic Good point. I'm a 👍  on this PR.

PR #2423, opened 2017-05-10
cybersource handle nested AuthReply with different reasonCode

elfassy:
# What
Use the top level `ResponseCode` from the XML response

# Why
In cases where a `c:ccAuthReply` was provided, we would recursively append its data to the response. The problem is that the `c:ccAuthReply` data also contained a different `ResponseCode`. This could lead to situations where we had a response code for success (100) but a `decision` which was `REJECT`. 
Cybersource confirmed with me over the phone that the authorization (AuthReply) can be a success but the merchant rules (Decision Manager) can still reject the transaction. https://support.cybersource.com/cybskb/index?page=content&id=C156&impressions=false

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -691,7 +691,8 @@ module ActiveMerchant #:nodoc:
         end
 
         success = response[:decision] == "ACCEPT"
-        message = @@response_codes[('r' + response[:reasonCode]).to_sym] rescue response[:message]
+        message = response[:message]
+
         authorization = success ? [ options[:order_id], response[:requestID], response[:requestToken], action, amount, options[:currency]].compact.join(";") : nil
 
         Response.new(success, message, response,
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -691,7 +691,8 @@ def commit(request, action, amount, options)
         end
 
         success = response[:decision] == "ACCEPT"
-        message = @@response_codes[('r' + response[:reasonCode]).to_sym] rescue response[:message]
+        message = response[:message]
```
elfassy: should we add a default error message when success is false but message is nil?

jasonwebster: Not using the normalized error messages is a big change (even if it wasn't always working as intended). This seems unrelated to actually extracting the correct `reasonCode`. Why'd you decide to do this too?

elfassy: the normalized error message logic was moved here: https://github.com/activemerchant/active_merchant/pull/2423/files#diff-9f21097553e010a708d707ede0637ca8R744

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -709,9 +710,10 @@ def parse(xml)
         xml = REXML::Document.new(xml)
         if root = REXML::XPath.first(xml, "//c:replyMessage")
           root.elements.to_a.each do |node|
-            case node.name
+            case node.expanded_name
```
pi3r: why the change to `expanded_name`?

elfassy: `node.expanded_name == 'c:reasonCode'` vs `node.name == 'reasonCode'

PR #2425, opened 2017-05-11
Remove old RemoteBarclaysEpdqTest

elfassy:
# What
Remove test file that is no longer used

# Why
Barclays Epdq was removed https://github.com/activemerchant/active_merchant/pull/2331
Running this test blows up because the `BarclaysEpdqGateway` class no longer exists


bdewater:
🤦‍♂️

PR #2428, opened 2017-05-12
Rename invalid province in remote beanstream test

AnotherJoSmith:
No need to use offensive terms in our codebase, let's try to be inclusive and welcoming instead.

@elfassy

### DIFF
--- a/test/remote/gateways/remote_beanstream_test.rb
+++ b/test/remote/gateways/remote_beanstream_test.rb
@@ -110,7 +110,7 @@ class RemoteBeanstreamTest < Test::Unit::TestCase
   end
 
   def test_failed_purchase_due_to_invalid_billing_state
-    @options[:billing_address][:state] = "Quebecistan"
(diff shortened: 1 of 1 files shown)


shasum:
I'm cool with the change :)

The `stan` ending is pretty normal for countries in our region (and not just muslim majority ones). India, a hindu majority state is called Hindustan locally. I don't see it being derogatory/offensive. I imagine other folks are likely to misinterpret.

AnotherJoSmith:
Hey @shasum, no worries!

Quebecistan has been used in the past as a derogatory expression towards muslims in Quebec, but it's not the most known term for sure.

see https://en.wikipedia.org/wiki/Barbara_Kay_controversy for more information.

shasum:
hmm...that's a strange coincidence I happened to use that exact term. Well, at least I can now claim I've made some controversial commits to a public repo ;)

thanks for cleaning it up 👍

PR #2431, opened 2017-05-16
NMI: Add Network Tokenization support

shasum:
**UPDATED**
```
$ RUBYOPT=W0 be rake test:remote TEST=test/remote/gateways/remote_nmi_test.rb 
33 tests, 106 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```
---
```
$ RUBYOPT=W0 be rake test:units TEST=test/unit/gateways/nmi_test.rb 
27 tests, 192 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* NMI: Add Network Tokenization support [shasum] #2431
 
 == Version 1.66.0 (May 4, 2017)
 * Support Rails 5.1 [jhawthorn] #2407
--- a/lib/active_merchant/billing/gateways/nmi.rb
+++ b/lib/active_merchant/billing/gateways/nmi.rb
@@ -116,7 +116,12 @@ module ActiveMerchant #:nodoc:
           gsub(%r((ccnumber=)\d+), '\1[FILTERED]').
           gsub(%r((cvv=)\d+), '\1[FILTERED]').
           gsub(%r((checkaba=)\d+), '\1[FILTERED]').
-          gsub(%r((checkaccount=)\d+), '\1[FILTERED]')
+          gsub(%r((checkaccount=)\d+), '\1[FILTERED]').
+          gsub(%r((cryptogram=)[^&]+(&?)), '\1[FILTERED]\2')
+      end
+
+      def supports_network_tokenization?
+        true
       end
 
       private
@@ -135,6 +140,10 @@ module ActiveMerchant #:nodoc:
       def add_payment_method(post, payment_method, options)
         if(payment_method.is_a?(String))
           post[:customer_vault_id] = payment_method
+        elsif (payment_method.is_a?(NetworkTokenizationCreditCard))
+          post[:ccnumber] = payment_method.number
+          post[:ccexp] = exp_date(payment_method)
+          post[:token_cryptogram] = payment_method.payment_cryptogram
         elsif(card_brand(payment_method) == 'check')
           post[:payment] = 'check'
           post[:checkname] = payment_method.name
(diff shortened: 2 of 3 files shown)


jasonwebster:
Typically the ECI indicator is also required for processing tokenized requests (especially like those decrypted from digital wallets, like Apple Pay). It looks like they support passing it in according to their docs: https://secure.networkmerchants.com/gw/merchants/resources/integration/integration_portal.php#transaction_variables

shasum:
@jasonwebster Good point.

NMI docs is not clear/up-to-date about network tokenization requests (**example**: there is no mention of the `token_cryptogram` field which has been added in this PR). After several back and forth with their support folks `ccnumber` and `token_cryptogram` are what they confirm as needed.

I don't think it hurts to add ECI. Adding it made `cavv` mandatory, which I mapped to the cryptogram. Also, added cryptogram to the list of things that get scrubbed.

Cc: @davidsantoso @curiousepic

davidsantoso:
> Typically the ECI indicator is also required for processing tokenized requests

I wonder if we should double check with NMI on this. Just last week I was working on Apple Pay for Moneris and their docs say it's optional and they don't use it in their examples. Makes me wonder if some gateways require it while others don't. 

> Adding it made cavv mandatory, which I mapped to the cryptogram.

@shasum the docs that Jason linked show `cavv` as the verification value/CVV. Did you happen to get confirmation from NMI that it should be mapped to the cryptogram? Something seems amiss to me if we send the cryptogram it two different fields. Beyond that, I don't think we know if a `NetworkTokenizationCreditCard` coming from Apple Pay will even include the CVV.

Seems like getting a better sense of `eci` usage and why it made the `cavv` value required would help.

shasum:
I've sent out an email to NMI to get some clarity around their usage of `eci`.

[NMI doc](https://secure.networkmerchants.com/gw/merchants/resources/integration/integration_portal.php#transaction_variables) indicates that `eci, cavv, xid` fields are passed in for 3DS transactions. Earlier, reading up a bit on Apple Pay I came across this [article](https://pomcor.com/2014/11/09/how-apple-pay-uses-3-d-secure-for-internet-payments/) that suggests under the hood Apple Pay using 3DS like protocol. Also, couple other gateways ( [firstdata_e4](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/firstdata_e4.rb#L264) and [cybersource](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L517) ) map `cryptogram` to `cavv` which made me think maybe it's a typical mapping.

I realize it's best to sort this out with the gateway folks. I'll update here once I hear back from them.

davidsantoso:
> Also, couple other gateways ( firstdata_e4 and cybersource ) map cryptogram to cavv which made me think maybe it's a typical mapping.

Interesting! I hadn't seen that before. Curious to see what you hear back from NMI.

jasonwebster:
Apple Pay does indeed rely on an (ab)use of the 3DS protocol for their transactions. I can say with 100% confidence that the requirement of the ECI indicator is dependent on something deep within the interaction between the issuer and acquirer, and have seen plenty of cases (funnily, mostly w/ Canadian issued Visas) where they will never, ever be accepted unless the value is set.

shasum:
NMI came back with a No.

>No, the ECI indicator is not needed with the token_cryptogram

Could be their focus market (US) allows them that leeway. Time to rollback and update the PR.

jasonwebster:
Cool! Thanks for taking the time to confirm that, sorry about the hassle.

shasum:
@jasonwebster Not at all. It was a valid observation. Thanks to you 🙇  we got it properly clarified.

davidsantoso:
🤗 all around!

davidsantoso:
@shasum let me know when this is fully updated and I'll take another look!

shasum:
@davidsantoso updated and ready for a 2nd look!

PR #2432, opened 2017-05-17
Opp: Send disable3DSecure field by default

davidsantoso:
Quick update to the Opp gateway- sends a `disable3DSecure` custom parameter by default but allows the option to flip the switch via an option.

Remote test run:

```
➜ ruby -Itest test/remote/gateways/remote_opp_test.rb
Loaded suite test/remote/gateways/remote_opp_test
Started
...............

Finished in 16.832362 seconds.
100% passed
0.89 tests/s, 2.85 assertions/s
```


curiousepic:
@davidsantoso What's driving this change? Other 3Dsecure functionality doesn't seem to be supported yet, unless they go in with customParameters...?

davidsantoso:
@curiousepic great question- since Opp is a white label gateway some of the "front end"/white labeled gateways that use them require 3DS unless specifically told to disable. Spreedly has a gateway implemented in which this is the case.

However, you bring up a good point in that we probably shouldn't force that custom field on Opp or any of the white labels that implement. I've updated the PR to only include it if it's sent as an option.

PR #2436, opened 2017-05-18
Orbital gateway now requires new test and production urls …

jcowhigjr:
… for connections as part of a certificate upgrade process, deadline May 31, 2017


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -422,9 +422,9 @@ def add_creditcard(xml, creditcard, currency=nil)
               xml.tag! :CardSecValInd, '1'
             end
             xml.tag! :CardSecVal,  creditcard.verification_value
+            end
           end
         end
-      end
```
sunblaze: These two changes are messing up the indentation 👎

WilsonChiang:
@jasonwebster mind reviewing as well, please?

PR #2440, opened 2017-05-24
Fix success response code validation for checkout_v2 gateway (Checkout.com)

nicolas-maalouf-cko:
This is to cater to an edge case that can happen with Checkout.com's Gateway:

When using [AVS risk rules](https://docs.checkout.com/the-hub/risk), a transaction can be handled as "Void Authorize". 

In such cases the responseCode would be 10000, but the responseMessage shows the error code and description instead of "Approved".

Sample response fields:
```
"responseMessage": "40111 - Street Match Only",
"responseCode": "10000",
"status": "Authorised",
```

Shopify needs to treat such transactions as declined.

Approved transactions show the below response fields:
```
"responseMessage": "Approved",
"responseCode": "10000",
"status": "Authorised",
```
or, when flagged (but still to be considered valid by Shopify)
```
"responseMessage": "40141 - Threshold Risk",
"responseCode": "10100",
"status": "Flagged",
```


nicolas-maalouf-cko:
@davidsantoso would you please be able to assist here?
I'm not sure why the checks failed.

Please let me know if anything else is required from our end to have this pull request merged.
I'll also inform Shopify when done.

Thanks a lot,
Cheers,
Nicolas

bdewater:
To get the tests to pass you probably need to change [this line](https://github.com/activemerchant/active_merchant/blob/88da65269810fe9c82e16ab1f8c50303a4d9ac36/test/unit/gateways/checkout_v2_test.rb#L233) to match the additional check you added whether responseMessage is "Approved".

nicolas-maalouf-cko:
Thanks @bdewater I have created a new PR using a different logic

PR #2442, opened 2017-05-25
JetPay V2: Add new gateway

shasum:
New adapter that talks to [JetPay Host XML Specification V2.03.01.pdf](https://github.com/activemerchant/active_merchant/files/1030427/JetPay.Host.XML.Specification.V2.03.01.pdf)

---

```
$ RUBYOPT=W0 be rake test:remote TEST=test/remote/gateways/remote_jetpay_v2_test.rb 
19 tests, 86 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

```
$ RUBYOPT=W0 be rake test:units TEST=test/unit/gateways/jetpay_v2_test.rb 
17 tests, 66 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/lib/active_merchant/billing/gateways/jetpay_v2.rb
+++ b/lib/active_merchant/billing/gateways/jetpay_v2.rb
@@ -1,0 +1,406 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class JetpayV2Gateway < Gateway
+      self.test_url = 'https://test1.jetpay.com/jetpay'
+      self.live_url = 'https://gateway20.jetpay.com/jetpay'
+
+      self.money_format = :cents
+      self.default_currency = 'USD'
+      self.supported_countries = ['US', 'CA']
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'http://www.jetpay.com'
+      self.display_name = 'JetPay'
+
+      API_VERSION = '2.2'
+
+      ACTION_CODE_MESSAGES = {
+        "000" =>  "Approved.",
+        "001" =>  "Refer to card issuer.",
+        "002" =>  "Refer to card issuer, special condition.",
+        "003" =>  "Invalid merchant or service provider.",
+        "004" =>  "Pick up card.",
+        "005" =>  "Do not honor.",
+        "006" =>  "Error.",
+        "007" =>  "Pick up card, special condition.",
+        "008" =>  "Honor with ID (Show ID).",
+        "010" =>  "Partial approval.",
+        "011" =>  "VIP approval.",
+        "012" =>  "Invalid transaction.",
+        "013" =>  "Invalid amount or exceeds maximum for card program.",
+        "014" =>  "Invalid account number (no such number).",
+        "015" =>  "No such issuer.",
+        "019" =>  "Re-enter Transaction.",
+        "021" =>  "No action taken (unable to back out prior transaction).",
+        "025" =>  "Transaction Not Found.",
+        "027" =>  "File update field edit error.",
+        "028" =>  "File is temporarily unavailable.",
+        "030" =>  "Format error.",
+        "039" =>  "No credit account.",
+        "041" =>  "Pick up card (lost card).",
+        "043" =>  "Pick up card (stolen card).",
+        "051" =>  "Insufficient funds.",
+        "052" =>  "No checking account.",
+        "053" =>  "Mp savomgs accpimt.",
+        "054" =>  "Expired Card.",
+        "055" =>  "Incorrect PIN.",
+        "057" =>  "Transaction not permitted to cardholder.",
+        "058" =>  "Transaction not allowed at terminal.",
+        "061" =>  "Exceeds withdrawal limit.",
+        "062" =>  "Restricted card (eg, Country Exclusion).",
+        "063" =>  "Security violation.",
+        "065" =>  "Activity count limit exceeded.",
+        "068" =>  "Response late.",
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/jetpay_v2.rb
```diff
@@ -0,0 +1,406 @@
...
+      def build_xml_request(transaction_type, options = {}, transaction_id = nil, &block)
+        xml = Builder::XmlMarkup.new
+        xml.tag! 'JetPay', 'Version' => API_VERSION do
+          # Basic values needed for any request
+          xml.tag! 'TerminalID', @options[:login]
+          xml.tag! 'TransactionType', transaction_type
+          xml.tag! 'TransactionID', transaction_id.nil? ? generate_unique_id.slice(0, 18) : transaction_id
+          xml.tag! 'Origin', options[:origin] || 'INTERNET'
+          xml.tag! 'IndustryInfo', 'Type' => options[:industry_info] || 'ECOMMERCE'
+          xml.tag! 'Application', options[:application] || 'n/a', 'Version' => options[:application_version] || '1.0'
```
davidsantoso: I think this syntax will error out? ` 'n/a', 'Version' => options[:application_version]` both the comma after `n/a` and the `'Version' => options[...]`. 

I guess that makes me wonder if we can just simplify it with:

`options[:application] || '1.0`

Unless I'm missing something about the JetPay v2 API where we might need to send something related to that middle value.

davidsantoso: I should note this goes for the line directly beneath this one as well.

lib/active_merchant/billing/gateways/jetpay_v2.rb
```diff
@@ -0,0 +1,406 @@
...
+      end
+
+      def supports_scrubbing
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
+          gsub(%r((>)\d+(</CardNum>)), '\1[FILTERED]\2').
```
davidsantoso: Any particular reason to not have this be: %r((<CardNum>)\d+(</CardNum>)?

lib/active_merchant/billing/gateways/jetpay_v2.rb
shasum: @davidsantoso 

Based off of the XML builder incantation `xml.tag! <name>, <value>, <attributes hash>`

the following

```
xml.tag! 'Application', options[:application] || 'n/a', 'Version' => options[:application_version] || '1.0'
```
helps output a markup such as

`<Application Version=1.0>spreedly</Application>`


The commas and || could be bit misleading but essentially it's 3 comma separated arguments with defaults for argument 2 and 3.

Couldn't find a better ref: https://apidock.com/rails/Builder/XmlMarkup but it talks about `tag! will also take text and attribute arguments (after the tag name) like normal markup methods`.

lib/active_merchant/billing/gateways/jetpay_v2.rb
shasum: @davidsantoso Reason is to target numbers that fall within `CardNum` and `CVV2` tags only. Not including tag names as part of the pattern match may scrub off things like amount and order number.

davidsantoso: @shasum so sorry- I just realized that I forgot to add tick marks around my code and GitHub escaped the contents 😞 See my updated question!

lib/active_merchant/billing/gateways/jetpay_v2.rb
davidsantoso: @shasum ah I see. So really it's this-

```
xml.tag! 'Application', (options[:application] || 'n/a'), ('Version' => options[:application_version] || '1.0')
```

shasum: Yep, the parens definitely helps :)

lib/active_merchant/billing/gateways/jetpay_v2.rb
shasum: @davidsantoso The start tag encompasses couple attributes.

`<CardNum CardPresent="false" Tokenize="true">4xxx3xxx2xxx1xxx</CardNum>`
 
Hence the focus on `>4xxx3xxx2xxx1xxx</CardNum>`

lib/active_merchant/billing/gateways/jetpay_v2.rb
```diff
@@ -0,0 +1,406 @@
...
+        commit(money, build_capture_request(transaction_id, money, options), token)
+      end
+
+      def void(reference, options = {})
+        transaction_id, approval, amount, token = reference.split(";")
+        commit(amount.to_i, build_void_request(amount.to_i, transaction_id, approval, token, options), token)
+      end
+
+      def credit(money, credit_card, options = {})
+        commit(money, build_credit_request('CREDIT', money, nil, credit_card, nil, options))
```
curiousepic: Slight inconsistency, unless there's some logic nuance: 'CREDIT' is sent from here to the `build_redit_request` for both `refund` and general `credit` actions, but the transaction type for other actions is sent from their `build_` methods.

curiousepic: Otherwise, LGTM!

shasum: Good catch. I'll edit that plus the parens before merging.

PR #2444, opened 2017-05-29
Authorize.net: Pass Level 2 Data Fields

curiousepic:
This required some shifting of the existing request building flow due to
strange element ordering/structuring requirements from ANET.

    Unit:
    83 tests, 473 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed
    
    Remote:
    59 tests, 204 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -166,7 +166,7 @@ module ActiveMerchant
             xml.amount(amount(amount))
 
             add_payment_source(xml, payment)
-            add_invoice(xml, options)
+            add_invoice(xml, 'refundTransaction', options)
             add_customer_data(xml, payment, options)
             add_settings(xml, payment, options)
             add_user_fields(xml, amount, options)
@@ -244,7 +244,12 @@ module ActiveMerchant
           xml.transactionType(transaction_type)
           xml.amount(amount(amount))
           add_payment_source(xml, payment)
-          add_invoice(xml, options)
+          add_invoice(xml, transaction_type, options)
+          add_tax_fields(xml, options)
+          add_duty_fields(xml, options)
+          add_shipping_fields(xml, options)
+          add_tax_exempt_status(xml, options)
+          add_po_number(xml, options)
           add_customer_data(xml, payment, options)
           add_market_type_device_type(xml, payment, options)
           add_settings(xml, payment, options)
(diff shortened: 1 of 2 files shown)


curiousepic:
@shasum Thanks for the look over - There is a separate CIM-specific gateway which is unused by Spreedly, that those remote tests are for; this base ANET gateway had CIM txn flow integrated with it some time ago. Re: PO Number method(s), it's difficult to unify/simplify them further, given the multiple conditional dependencies 😕

shasum:
@curiousepic Not sure if this fits well with the PO logic but I was imagining something like this.

```ruby
xml.order do
  xml.invoiceNumber(truncate(options[:order_id], 20))
  xml.description(truncate(options[:description], 255))
  add_po_number(xml, options, transaction_type)
end

def add_po_number(xml, options, transaction_type='')
  return unless options[:po_number]
  if transaction_type.start_with?("profileTrans")
    xml.purchaseOrderNumber(options[:po_number])
  else
    xml.poNumber(options[:po_number])
  end
end
```

curiousepic:
@shasum I tried that, but because it is also added in a different location for normal transactions, it gets added a second time because the transaction_type isn't actually enough for it to go on to determine which place it should go 🤦‍♂️

shasum:
@curiousepic Gotchya. I suspected some intricacy around it that I was definitely missing. What if we throw in part of the original guard clause inside the order element. Would that eliminate duplicates?
```ruby
xml.order do
  xml.invoiceNumber(truncate(options[:order_id], 20))
  xml.description(truncate(options[:description], 255))
  add_po_number(xml, options, transaction_type) if transaction_type.start_with?("profileTrans")
end

def add_po_number(xml, options, transaction_type='')
  return unless options[:po_number]
  if transaction_type.start_with?("profileTrans")
    xml.purchaseOrderNumber(options[:po_number])
  else
    xml.poNumber(options[:po_number])
  end
end
```

curiousepic:
Well, that will then require passing `transaction_type` through basically every action, even though it's not needed in most of them. It seems as un-reason-able, if not more, than the split, isolated logic of the original... but not sure which is the lesser of evils. Anyhow this feels like an awful lot of time invested in a simple feature so I think I'm going to call it here, but totally open to revisiting it if I see other opinions - thanks for the effort!

curiousepic:
Closed by https://github.com/activemerchant/active_merchant/commit/a7bf55caea9df1789d97d7ecd6509fe3a3915e08

PR #2447, opened 2017-06-05
Ebanx: Add gateway support

davidsantoso:


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -0,0 +1,234 @@
...
+        post[:payment][:creditcard][:auto_capture] = false
+
+        commit(:authorize, post)
+      end
+
+      def capture(money, authorization, options={})
+        post = {}
+        add_integration_key(post)
+        post[:hash] = authorization
+        post[:amount] = amount(money)
```
curiousepic: Maybe extract to a method since it's used twice, at least?

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -0,0 +1,234 @@
...
+        commit(:capture, post)
+      end
+
+      def refund(money, authorization, options={})
+        post = {}
+        add_integration_key(post)
+        add_operation(post)
+        add_authorization(post, authorization)
+        post[:amount] = amount(money)
+        post[:description] = options[:description]
```
curiousepic: This is weird. Is it a refund-only thing?

davidsantoso: yeah... so for `purchase` and `authorize`, it's `amount_total` but for refund it's just `amount`.

lib/active_merchant/billing/gateways/ebanx.rb
davidsantoso: So, I went back and forth on this one. On the one hand, I prefer to have them in methods if they're used more than once. But, on the other hand, I wondered if that was more confusing cause for `purchase` and `authorize` the field name is `amount_total` and it's added in the `add_invoice` method.

This felt a little more explicit to me to say that we're adding this exact `amount` field rather than having amounts being added in two different methods.

lib/active_merchant/billing/gateways/ebanx.rb
davidsantoso: Yeah. Refunds require a description field. ☹️

lib/active_merchant/billing/gateways/ebanx.rb
curiousepic: Ugh, gotcha, I'm cool with that.

davidsantoso:
Thanks for the review!

PR #2448, opened 2017-06-05
FirstData Payeezy: Set default ECI value for auth/purchase transactions

jasonwebster:
I've heard from a few users that this gateway (and/or their processors
behind it) are now soft-requiring the ECI indicator to always be set.
Typical of this gateway, [documentation is unclear at
best](https://support.payeezy.com/hc/en-us/articles/206601408-First-Data-Payeezy-Gateway-Web-Service-API-Reference-Guide).

I've done this in such a way that all previous cases were accounted for,
and a 3DS/network transaction's embedded value will always take
precedence.

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4.rb
@@ -236,7 +236,9 @@ module ActiveMerchant #:nodoc:
           xml.tag! "Expiry_Date", expdate(credit_card)
           xml.tag! "CardHoldersName", credit_card.name
           xml.tag! "CardType", card_type(credit_card.brand)
-          xml.tag! "Ecommerce_Flag", (credit_card.try(:eci) || options[:eci] || DEFAULT_ECI)
+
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/firstdata_e4.rb
```diff
@@ -275,7 +278,6 @@ def add_network_tokenization_credit_card(xml, credit_card)
       def add_card_authentication_data(xml, options)
         xml.tag! "CAVV", options[:cavv]
         xml.tag! "XID", options[:xid]
-        xml.tag! "Ecommerce_Flag", options[:eci]
```
pi3r: `eci` isn't needed for capture/void/refund right (only for authorize/purchase)?

jasonwebster: Correct.

davidsantoso:
👋 Taking a look now.

davidsantoso:
@jasonwebster LGTM. Only remote tests still failing (with the default ECI of 07) are currency related so I think it should be ok. Most likely some stale test data.

jasonwebster:
Yeah, I had the same failures. I'm guessing our test account that we use isn't set up to process in the currencies those tests use.

PR #2449, opened 2017-06-06
CyberSource: Gateway tweaks

troyswanson:
Hello. Here are some pretty minor tweaks that I think might help this gateway...

### Allow payments from subscriptions to include addresses

- [x] Fixes bug in `add_address` when using a subscription

One problem that I've come across is the inability to add address data to a transaction that uses a "subscription" (tokenized credit card) since the `add_address` method fails when trying to add the first/last name.

### Send additional data on transaction methods

- [x] Includes billing and shipping address data on transaction methods
- [x] Includes line item data on transaction methods

It doesn't seem like there would be any reason *not* send as much metadata about a transaction as possible, so I feel like we should at least try. Decision Manager can use data for both billing and shipping addresses, as well as line item data for the transaction. (Continuing the work from #1877.)

### Tweak data retention logic

- [x] Includes new logic for `fraud_review`
- [x] Includes authorization data regardless of payment failure

Along the lines of fraud review, adding a new `fraud_review` property in the response hash, in keeping with the pattern of other gateways, would be helpful for upstream consumers of AM. (Continuing the work from #1687.)

Also, the `authorization` property seems like it would be useful to retain regardless of failure. As far as I know, a "REJECT" or a "REVIEW" in CyberSource might have actually resulted in a successful auth hold, so it makes sense to keep this data so that you can do an auth revert once you review the transaction.

### Test coverage

- [ ] Includes new tests

I would love to write some tests for this new functionality, but I'm not very strong in Ruby so some guidance would be helpful on that front. Thanks!

---

/ref #1681

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -177,27 +177,27 @@ module ActiveMerchant #:nodoc:
       #
       # The line_item hash goes in the options hash and should look like
       #
-      #         :line_items => [
-      #           {
-      #             :declared_value => '1',
-      #             :quantity => '2',
-      #             :code => 'default',
-      #             :description => 'Giant Walrus',
-      #             :sku => 'WA323232323232323'
-      #           },
-      #           {
-      #             :declared_value => '6',
-      #             :quantity => '1',
-      #             :code => 'default',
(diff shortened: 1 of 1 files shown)


troyswanson:
Does anyone have any time to help review this and make any comments?

PR #2017, opened 2016-02-10
Element express terminal values

asanger:
If no options are passed, everything works exactly as normal. If any options _are_ passed, they will be populated in the defined locations.

This change was made to get us past "certification" with ElementExpress. Instead of passing the default values of the modified lines, we were required to pass in values based on the specific use case.

PR #2450, opened 2017-06-06
Fix Element Express endpoint problems

asanger:
If no options are passed, everything works exactly as normal. If any options are passed, they will be populated in the defined locations.

This change was made to get us past "certification" with ElementExpress. Instead of passing the default values of the modified lines, we were required to pass in values based on the specific use case.

Also needed to rename the production endpoint, as the original endpoint that was added was incorrect.

Original PR: #2017 

Tagging @davidsantoso as requested.

/cc @yjukaku

### DIFF
--- a/test/unit/gateways/element_test.rb
+++ b/test/unit/gateways/element_test.rb
@@ -10,7 +10,8 @@ class ElementTest < Test::Unit::TestCase
     @options = {
       order_id: '1',
       billing_address: address,
-      description: 'Store Purchase'
+      description: 'Store Purchase',
+      card_present_code: 'ManualKeyed'
     }
   end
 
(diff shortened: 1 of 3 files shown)


yjukaku:
Anything holding this up?

test/unit/gateways/element_test.rb
```diff
@@ -10,7 +10,8 @@ def setup
     @options = {
       order_id: '1',
       billing_address: address,
-      description: 'Store Purchase'
+      description: 'Store Purchase',
+      card_present_code: 'ManualKeyed'
```
stevenjonescgm: it seems Vantiv/Mercury updated their API recently to either separate `card_present_code` from `card_input_code`, or now properly validates these values. 

fyi
CardPresentCode 0/1/2/3 Default/Unknown/Present/NotPresent
CardInputCode 0/1/2/3/4/5 Default/Unknown/MagstripeRead/ContactlessMagstripeRead/ManualKeyed/ManualKeyedMagstripeFailure

I'm working on my own changes to fix Check/ACH, `store` and more. I'll try to remember to clean up my changes once they are stable, make a PR and link this

PR #2452, opened 2017-06-07
Fix success response code validation for checkout_v2 gateway (Checkout.com)

nicolas-maalouf-cko:
This is to cater to an edge case that can happen with Checkout.com's Gateway:

When using [AVS risk rules](https://docs.checkout.com/the-hub/risk), a transaction can be handled as "Void Authorize". 

In such cases the responseCode would be 10000, but the responseMessage shows the error code and description instead of "Approved".

Sample response fields:
```
"responseMessage": "40111 - Street Match Only",
"responseCode": "10000",
"status": "Authorised",
```

Shopify needs to treat such transactions as declined.

Approved transactions show the below response fields:
```
"responseMessage": "Approved",
"responseCode": "10000",
"status": "Authorised",
```
or, when flagged (but still to be considered valid by Shopify)
```
"responseMessage": "40141 - Threshold Risk",
"responseCode": "10100",
"status": "Flagged",
```

++ **Add the list of currencies without fractions**


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -166,7 +166,7 @@ def parse(body)
       end
 
       def success_from(response)
-        response["responseCode"] == "10000" || response["responseCode"] == "10100"
+        (response["responseCode"] == "10000" && !(response["responseMessage"].include? "40")) || response["responseCode"] == "10100"
```
bdewater: How about `!response["responseMessage"].start_with?("40")` ? This puts the opening parenthesis for the second conditional method in a more clear place, and is more precise in matching the 40xxx codes at the start of the string.

Also it would be very helpful to codify the cases you mentioned in the pull request in the tests to prevent regressions. Could you add them to https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/checkout_v2_test.rb ? Thanks!

nicolas-maalouf-cko:
@bdewater thanks for the review, I improved the logic as suggested.
I also added a list of all our supported currencies without fractions.
There's a merchant using the extension who got ISK transactions charged 100 times the amount. 

I'll try to add tests to cater for the cases described in the PR later, but can we please merge this especially for the currencies without fractions?

Thanks

bdewater:
ISK is already [defined](https://github.com/nicolas-maalouf-cko/active_merchant/blob/243f5ff1430f7f74d65400348cbd52b1e756d0cb/lib/active_merchant/billing/gateway.rb#L129) as a non-fractional currency, but you must use [localized_amount](https://github.com/nicolas-maalouf-cko/active_merchant/blob/243f5ff1430f7f74d65400348cbd52b1e756d0cb/lib/active_merchant/billing/gateway.rb#L266-L276) in this gateway's code to properly format amounts.

nicolas-maalouf-cko:
Thanks for the help @bdewater 
Will using:
```
post[:value] = localized_amount(money, options[:currency])
```
on line 79 do it?

And should I remove 
```
self.currencies_without_fractions = %w(BIF BYR CLF DJF GNF ISK JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
```
from the pull request in that case?

nicolas-maalouf-cko:
@bdewater I added localized_amount in the latest commit

jasonwebster:
Hey @nicolas-maalouf-cko! Thanks for the contribution here. I'd really like to see a test for the response code parsing, if you've got the time. Otherwise, looks good.

jasonwebster:
We'll merge once there's test coverage for the change, as I requested above. Thanks!

nicolas-maalouf-cko:
@jasonwebster just added the test, please let me know if it looks good.

Thanks again

nicolas-maalouf-cko:
@jasonwebster @bdewater would you please be able to let me know if anything else should be done to be able to merge?

We're having customers using the gateway getting charged wrong amounts in some currencies because of localized_amount.

Thanks

dtykocki:
@nicolas-maalouf-cko would it be possible to squash all of the commits down?

nicolas-maalouf-cko:
@dtykocki do you mean doing a squash and merge on your end?
Or you prefer if I do the squash on my end?

Thanks

dtykocki:
@nicolas-maalouf-cko Squash on your end

nicolas-maalouf-cko:
@dtykocki squashing my commits (first one dating back to April) along with many commits coming from other contributors seems to be less clean to do than I expected.

Should I just close this PR and create a new clean one?

Please let me know,
Thanks

jasonwebster:
I can do it when I merge, don't worry about it! Thanks again for your contribution.

nicolas-maalouf-cko:
Thanks @jasonwebster @dtykocki !

PR #2453, opened 2017-06-07
Element: Add 'credit' method

yjukaku:
Adds the ability to credit a payment method a certain amount using the Element (Vantiv / Element Express) gateway.

Note that this working in production depends on #2450 which corrects the live url

/cc @asanger


asanger:
Y U CLOSE!?

yjukaku:
@asanger I'm worried about the potential for [salami slicing](https://www.youtube.com/watch?v=N7JBXGkBoFc) when using Element Express endpoints. 😉

PR #2461, opened 2017-06-13
QuickPay: Return last response for purchase and authorize

curiousepic:
Previously, the MultiResponse used for Authorize and Purchase were
"using first response", which meant that for stored cards, this was
returning the single-use token for its authorization, causing refunds
and voids of auths and purchases by stored cards to fail. Authorize and
Purchase now return the last response as their result, allowing the proper
id to be returned as the authorization.

Unit:
14 tests, 72 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
26 tests, 119 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

I'd appreciate some eyes on this from @davidsantoso - I want to make sure the method I'm using here is kosher because while simple, it seems a bit unorthodox, and I'm not sure if it would confuse any error reporting that we don't currently account for in any tests.

An alternative might be to use a split-auth system, but given that auth and purchase can consist of 3 or 4 individual requests, and only under certain circumstances would we need to record the token, I'm not sure it would be preferable.

### DIFF
--- a/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
+++ b/lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
@@ -15,7 +15,7 @@ module ActiveMerchant
       end
 
       def purchase(money, credit_card_or_reference, options = {})
-        MultiResponse.run(true) do |r|
+        MultiResponse.run do |r|
           if credit_card_or_reference.is_a?(String)
             r.process { create_token(credit_card_or_reference, options) }
             credit_card_or_reference = r.authorization
@@ -34,7 +34,7 @@ module ActiveMerchant
       end
 
       def authorize(money, credit_card_or_reference, options = {})
-        MultiResponse.run(true) do |r|
+        MultiResponse.run do |r|
           if credit_card_or_reference.is_a?(String)
             r.process { create_token(credit_card_or_reference, options) }
             credit_card_or_reference = r.authorization
@@ -132,7 +132,6 @@ module ActiveMerchant
 
         def create_token(identification, options)
           post = {}
-          # post[:id] = options[:id]
           commit(synchronized_path("/cards/#{identification}/tokens"), post)
         end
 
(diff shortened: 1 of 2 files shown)


curiousepic:
Just noticed this method is used in one other gateway, Stripe.

lib/active_merchant/billing/gateways/quickpay/quickpay_v10.rb
```diff
@@ -15,7 +15,7 @@ def initialize(options = {})
       end
 
       def purchase(money, credit_card_or_reference, options = {})
-        MultiResponse.run(true) do |r|
+        process = MultiResponse.run(true) do |r|
```
davidsantoso: @curiousepic I believe if you don't pass `true` to the `MultiResponse.run` it'll default to using the last response as the primary response?

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/response.rb#L44

I think that would avoid setting it to a variable then pulling the last response.

davidsantoso: I should also note, if you look in the Stripe adapter you'll see that it's just `MultiResponse.run do |r|` and since the initializer defaults to false, it'll use the last response.

curiousepic: Indeed that works. I was thinking that it was set to use the first response for error/message reporting purposes, but removing it doesn't cause any negative remote test failures, at least.

curiousepic: FYI Stripe does also call `.responses.last`on the MultiResponse.

davidsantoso: Ahhh good eye. I suppose following that pattern wouldn't hurt.

curiousepic: Well, there are plenty more that just don't specify use_first and just return the multiresponse, so I'll go with that if you think my imagined error reporting concerns are safe.

davidsantoso: I think it's ok. If the `use_first_response` isn't set to true, I think the default is that the last response becomes the primary response.

PR #2462, opened 2017-06-13
moved to name value pair (NVP) with payflow

jusleg:
A bug was happening with payflow pro when a lot of items were present in the cart (40+ items). Paypal would respond with `10413 error: The totals of the cart item amounts do not match order amounts.` After discussing with paypal, it was determined that we should use PAYPAL-NVP as it will fix the problem.

PAYPAL-NVP adds new exdata in the response (which we ignore at this moment) Here are the new fields:
  - TIMESTAMP
  - ACK
  - VERSION
  - BUILD 

Both paypal express and direct flow have been tested and this had no negative impact. Carts with 40+ items will not generate the error anymore. Since the line item fields are deprecated in paypal, we seem to lose the items description on the paypal checkout page. This is a small compromise to make in order to get payflow to work with any carts.

### DIFF
--- a/lib/active_merchant/billing/gateways/payflow/payflow_common_api.rb
+++ b/lib/active_merchant/billing/gateways/payflow/payflow_common_api.rb
@@ -187,7 +187,8 @@ module ActiveMerchant #:nodoc:
           "X-VPS-Client-Timeout" => timeout.to_s,
           "X-VPS-VIT-Integration-Product" => "ActiveMerchant",
           "X-VPS-VIT-Runtime-Version" => RUBY_VERSION,
-          "X-VPS-Request-ID" => SecureRandom.hex(16)
+          "X-VPS-Request-ID" => SecureRandom.hex(16),
+          "PAYPAL-NVP" => "Y"
         }
       end
 


jusleg:
@jasonwebster @ntalbott for review

jasonwebster:
Closed via 66ca4da3fe38ff3daf1e42c936f714a5c451c07f

davidsantoso:
@jusleg @jasonwebster wondering if we could actually make this header optional? Looking at the Payflow docs, it looks like it changes a couple things about the request and response format. Not sure if it's needed (or maybe even wanted) across the board for all Payflow transactions.

https://developer.paypal.com/docs/classic/payflow/integration-guide/#sending-requests-directly-to-paypal-bypassing-payflow

I was thinking something along the lines of:

```ruby
def build_headers(content_length, options)
  {
    "Content-Type" => "text/xml",
    "Content-Length" => content_length.to_s,
    "X-VPS-Client-Timeout" => timeout.to_s,
    "X-VPS-VIT-Integration-Product" => "ActiveMerchant",
    "X-VPS-VIT-Runtime-Version" => RUBY_VERSION,
    "X-VPS-Request-ID" => SecureRandom.hex(16),
    "PAYPAL-NVP" => options[:paypal_nvp] if options[:paypal_nvp]
  }
end

```

davidsantoso:
@jusleg @jasonwebster just friendly ping on the question above- any concerns with this change?

jasonwebster:
:+1: Sure, I don't see why not. 

Why don't you let me/us take another look at it and see if we can detect when it needs to be set, as well as normalize the responses.

For ref: https://developer.paypal.com/docs/classic/payflow/integration-guide/#sending-requests-directly-to-paypal-bypassing-payflow

jasonwebster:
Actually, that's the wrong documentation, whoops. https://www.paypalobjects.com/webstatic/en_US/developer/docs/pdf/pp_payflowpro_xmlpay_guide.pdf is what we're using, I believe. In which case, this header may not even have an effect. @jusleg Could you provide some more details on the conversations you had with PayPal?

davidsantoso:
👍 so many different PayPal docs! 

For what it's worth- the docs I linked earlier are what we go off of. Request fields seem to match up with what we generate in the Payflow adapter as well. Although, that adapter was before my time contributing to Active Merchant so could be stale by now.

jusleg:
PayPal's response regarding NVP:
> After looking over the logs and playing with the code, it looks like passing “PAYPAL-NVP:Y” in your header clears up the issue (noted here). This essentially tells the system to bypass Payflow and lets all the line items be processed correctly.
> 
> The testing shows that up to a small number of line items can be processed without the NVP flag before the Payflow system truncates the request. Because it gets truncated, the totals don’t match up. The NVP Y header is a best practice if you’re going to pass multiple line items. If you think there will be line items regularly, I would suggest implementing the header.

From our testing, this does not affect the way we handle the response. The additional fields in the response are being ignored for now.

davidsantoso:
> This essentially tells the system to bypass Payflow and lets all the line items be processed correctly.

So, I suppose this is what worries me just a tad. I'm not sure if bypassing is what we want the default behaviour to be. Could be that some people require some functionality in Payflow that bypassing would alter how they are processing. Making it an optional header would allow for both setups to coexist.

davidsantoso:
@jusleg @jasonwebster unless y'all have any major concerns with making this an optional vs default header, I'm gonna go ahead and make that change in the next day or so!

davidsantoso:
Closed with 8ccfbcc1f251ffa79ca6830c5f84a63b68d2fe73.

PR #2466, opened 2017-06-14
Support three-decimal currencies

curiousepic:
Similar to non-fractional currency support, gateways can override which
currencies should be treated as three-decimal.

Unit tests:
3615 tests, 66645 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

I'm unsure about adding some nominally three-decimal currencies (BHD KWD OMR RSD TND) to `gateway` itself, so I left it blank.

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -125,8 +125,9 @@ module ActiveMerchant #:nodoc:
       class_attribute :supported_cardtypes
       self.supported_cardtypes = []
 
-      class_attribute :currencies_without_fractions
+      class_attribute :currencies_without_fractions, :currencies_with_three_decimal_places
       self.currencies_without_fractions = %w(BIF BYR CLP CVE DJF GNF HUF ISK JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
+      self.currencies_with_three_decimal_places = %w()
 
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateway.rb
```diff
@@ -125,8 +125,9 @@ def generate_unique_id
       class_attribute :supported_cardtypes
       self.supported_cardtypes = []
 
-      class_attribute :currencies_without_fractions
+      class_attribute :currencies_without_fractions, :currencies_with_three_decimal_places
       self.currencies_without_fractions = %w(BIF BYR CLP CVE DJF GNF HUF ISK JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
+      self.currencies_with_three_decimal_places = %w()
```
davidsantoso: I wonder if we should add them here as defaults. Seems we know which ones are and if it needs to change at the gateway level it can be overridden?

curiousepic: My only hesitance would be if anyone was already transacting in 3-dec currencies with a gateway that don't consider them to be 3-dec or don't need special formatting.

davidsantoso: Hmmm. 🤔 That's a fair point I suppose.

PR #2467, opened 2017-06-14
Fixed Ability for IATS Gateway to make purchased using 'CustomerCode' ('stored' card token)

justinh24:
Store and Unstore methods were already implemented
Passing the 'stored token' to purchase would return an error.

Implimented 'ProcessCreditCardWithCustomerCodeV1'
Added Unit Test
Added Remote Test

### DIFF
--- a/lib/active_merchant/billing/gateways/iats_payments.rb
+++ b/lib/active_merchant/billing/gateways/iats_payments.rb
@@ -15,6 +15,7 @@ module ActiveMerchant #:nodoc:
 
       ACTIONS = {
         purchase: "ProcessCreditCardV1",
+        purchase_token: "ProcessCreditCardWithCustomerCodeV1",
         purchase_check: "ProcessACHEFTV1",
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/iats_payments.rb
```diff
@@ -42,7 +43,7 @@ def purchase(money, payment, options={})
         add_ip(post, options)
         add_description(post, options)
 
-        commit((payment.is_a?(Check) ? :purchase_check : :purchase), post)
+        commit((payment.is_a?(Check) ? :purchase_check : payment.is_a?(String) ? :purchase_token : :purchase), post)
```
bdewater: Nested ternary operators 🙅‍♂️  could you please change this to something similar to the structure used in `add_payment`?

bpollack:
@bdewater Is this something you're still interested in landing?

Issue #2465, opened 2017-06-14
[Openpay] Purchase transactions fail when using stored card


dtykocki:
ActiveMerchant: master
Ruby: 2.1.7p400

The Openpay adapter remote test for `test_successful_purchase_with_card_stored` fails with:
```
Response expected to succeed: <#<ActiveMerchant::Billing::Response:0x007fcfa1c22c20
   @authorization=nil,
   @avs_result=
    {"code"=>nil, "message"=>nil, "street_match"=>nil, "postal_match"=>nil},
   @cvv_result={"code"=>nil, "message"=>nil},
   @emv_authorization=nil,
   @error_code=nil,
   @fraud_review=nil,
   @message="Please specify a valid source_id",
   @params=
    {"category"=>"request",
     "description"=>"Please specify a valid source_id",
     "http_code"=>422,
     "error_code"=>1003,
     "request_id"=>"8eb61789-7311-4cc0-8fcf-2c338941ea6f"},
   @success=false,
   @test=true>>
test_successful_purchase_with_card_stored(RemoteOpenpayTest)
```

The issue here is that AM is storing cards at the customer level and then transacting at the merchant level:
1. `store` performs a POST to `/customers/:customer_id/cards` with the card data.
2. `purchase` and `authorize` performs a POST to `/charges` passing the `card_id` from the store call.

Example curl requests that reproduce this:
#### Store
```
curl https://sandbox-api.openpay.mx/v1/mzdtln0bmtms6o3kck8f/customers/ag4nktpdzebjiye1tlze/cards \
   -u sk_e568c42a6c384b7ab02cd47d2e407cab: \
   -H "Content-type: application/json" \
   -X POST -d '{
   "card_number":"4111111111111111",
   "holder_name":"Juan Perez Ramirez",
   "expiration_year":"20",
   "expiration_month":"12",
   "cvv2":"110"
 }' 

{"id":"kddkis234glfutacmdrw","type":"debit","brand":"visa","address":null,"card_number":"411111XXXXXX1111","holder_name":"Juan Perez Ramirez","expiration_year":"20","expiration_month":"12","allows_charges":true,"allows_payouts":true,"creation_date":"2017-06-14T15:54:25-05:00","bank_name":"Banamex","customer_id":"ag4nktpdzebjiye1tlze","bank_code":"002"}
```

#### Purchase
Using the `id` attribute from the above.
```
[...]

PR #2468, opened 2017-06-15
Send customer email and name in authorize and purchase

dtykocki:
Previously only name and email were sent to Openpay when storing a card.
This adds support for sending the customer's name and email when performing
authorize or purchase transactions.

Unit:
```
ruby -Itest test/unit/gateways/openpay_test.rb
19 tests, 98 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100%
```

Remote:
```
ruby -Itest test/remote/gateways/remote_openpay_test.rb
23 tests, 78 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 95.6522% passed
```

Closes #2661

The remote test failure above currently fails on master. The details about the failure have been captured in https://github.com/activemerchant/active_merchant/issues/2465

@davidsantoso / @curiousepic

### DIFF
--- a/lib/active_merchant/billing/gateways/openpay.rb
+++ b/lib/active_merchant/billing/gateways/openpay.rb
@@ -130,8 +130,20 @@ module ActiveMerchant #:nodoc:
             holder_name: creditcard.name
           }
           add_address(card, options)
+          add_customer_data(post, creditcard, options)
           post[:card] = card
         end
+      end
+
+      def add_customer_data(post, creditcard, options)
+        if options[:email]
+          customer = {
+            name: creditcard.name || options[:name],
+            email: options[:email]
+          }
+          post[:customer] = customer
+        end
+        post
       end
 
       def add_address(card, options)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -130,10 +130,22 @@ def add_creditcard(post, creditcard, options)
...
           add_address(card, options)
+          add_customer_data(post, creditcard, options)
           post[:card] = card
         end
       end
 
+      def add_customer_data(post, creditcard, options)
+        if options[:email]
+          customer = {
+            name: options[:name] || creditcard.name,
```
curiousepic: I know it's what was being used before, but pulling name straight from options is a pretty rare pattern. Since this method is called only when adding a credit card, I think the preference order here should be switched.  Otherwise, LGTM 👍

dtykocki: @curiousepic Update made.

PR #2470, opened 2017-06-16
SafeCharge: Track currency from original transaction

davidsantoso:
Sending in a currency is required for all SafeCharge transaction types.
Moreover, SafeCharge does not keep track of the original currency used
for an authorize or purchase so if an authorize is done using EUR and
 then a subsequent capture is run on that authorization, the capture
 will be done in USD instead of EUR since the adapter defaults to USD.

Of course this could be mitigated by sending in a currency code into
the options hash for a void/refund/capture, but it's a bit silly to need
to remember that and pass that in when you'll most likely want to run
subsequent transaction requests (e.g. refund, void, and capture) on the
same currency.

This add the currency code used in the original transaction to the
authorization response field so running a void/refund/capture will
always do so in the original currency used vs defaulting to USD.

Remote test run: again, there's a single failure for stale test card data 
on a `credit` request which requires a specific card number to properly 
exercise the transaction.

```
➜ ruby -Itest test/remote/gateways/remote_safe_charge_test.rb
Loaded suite test/remote/gateways/remote_safe_charge_test
Started
...........F
==============================================================================================================================================================================================================================================
Failure:
  Response expected to succeed: <#<ActiveMerchant::Billing::Response:0x007fe19d0295f0
   @authorization="|101508748226|bwBpAFAAcABGAE0ANAAzAEwAZQA0AE0ANAA6AD4ASgBbADsAUwB6AHQAewAsAG4AJgBCACUAUgBDAFoAKgA+AD0ANQBUACoAQwBiAG4AdgA+AFgAMwA=|09|18|1.00|EUR",
   ...
   @message="This card is not supported in the CFT Program",
   @params={...},
   @success=false,
   @test=true>>
test_successful_credit(RemoteSafeChargeTest)
test/remote/gateways/remote_safe_charge_test.rb:96:in `test_successful_credit'
     93:
     94:   def test_successful_credit
     95:     response = @gateway.credit(@amount, @credit_card, @options)
  => 96:     assert_success response
     97:     assert_equal 'Success', response.message
     98:   end
     99:
==============================================================================================================================================================================================================================================
......

Finished in 24.59002 seconds.
18 tests, 44 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.4444% passed
0.73 tests/s, 1.79 assertions/s
```

### DIFF
--- a/lib/active_merchant/billing/gateways/safe_charge.rb
+++ b/lib/active_merchant/billing/gateways/safe_charge.rb
@@ -40,8 +40,8 @@ module ActiveMerchant #:nodoc:
 
       def capture(money, authorization, options={})
         post = {}
-        add_transaction_data("Settle", post, money, options)
-        auth, transaction_id, token, exp_month, exp_year, _ = authorization.split("|")
+        auth, transaction_id, token, exp_month, exp_year, _, original_currency = authorization.split("|")
+        add_transaction_data("Settle", post, money, (options.merge!({currency: original_currency})))
         post[:sg_AuthCode] = auth
         post[:sg_TransactionID] = transaction_id
         post[:sg_CCToken] = token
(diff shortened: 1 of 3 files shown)


davidsantoso:
I should probably note that all authorize/purchase transactions that happened prior to this change will still have the bug since the fix is in the authorization concatenation. But, going forward it should be fixed.

curiousepic:
Actually `test_successful_void` seems to be failing.

davidsantoso:
@curiousepic ah, I made that last minute change and forgot to rerun unit tests before updating the PR! Fixing...

PR #2471, opened 2017-06-17
Add razorpay gateway

djsmentya:
HI! This is not final version of gateway. Currently stuck with remote tests.
Gateway do not allow to create/authorize payment via api, only via JavaScript library. That means I have to create manually payment and copy payment id to test each example. Because if I run test with one payment_id i can't run it again or other one with the same payment_id.

The only way I see - add a payment_id for each test to 'test/fixtures.yml' and change it each time i want to run a test. Or may be use capybara gem to create payments via js and use payment_ids in tests. But the contribution guide said that ' - No new gem dependencies will be accepted.' Is this a good reason to propose capybara with Poltergeist?

### DIFF
--- a/lib/active_merchant/billing/gateways/razorpay.rb
+++ b/lib/active_merchant/billing/gateways/razorpay.rb
@@ -1,13 +1,12 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class RazorpayGateway < Gateway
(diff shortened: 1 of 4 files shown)


bpollack:
@djsmentya Is this something you're still interested in? The CI failed and it's been over a year.

bpollack:
I'm closing this based on the lack of movement, but please feel free to reopen if you're interested in making tests pass.

PR #2475, opened 2017-06-20
JetPay V2: Support `store` and token based payments

shasum:
- Support `store`
- Modify other transactions to accept stored tokens
- Commit certification tests to the repo

---

```ruby
$ be rake test:units TEST=test/unit/gateways/jetpay_v2_test.rb 
17 tests, 66 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

```ruby
$ be rake test:remote TEST=test/remote/gateways/remote_jetpay_v2_test.rb 
20 tests, 80 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

```ruby
$ be rake test:remote TEST=test/remote/gateways/remote_jetpay_v2_certification_test.rb
25 tests, 100 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@
 * Openpay: Send customer name and email in authorize and purchase [dtykocki] #2468
 * SafeCharge: Track currency from original transaction [davidsantoso] #2470
 * Braintree Blue: Braintree Blue: Add ECI indicator to Android Pay transactions [davidsantoso] #2474
+* JetPay V2: Support store transactions and token based payments [shasum] #2475
 
 == Version 1.67.0 (June 8, 2017)
 * Acapture: Pass 3D Secure fields [davidsantoso] #2451
--- a/lib/active_merchant/billing/gateways/jetpay_v2.rb
+++ b/lib/active_merchant/billing/gateways/jetpay_v2.rb
@@ -153,39 +153,39 @@ module ActiveMerchant #:nodoc:
         super
       end
 
-      def purchase(money, credit_card, options = {})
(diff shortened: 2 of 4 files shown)


davidsantoso:
@shasum thanks for adding `store`! Two thoughts-

1. Wondering if consolidating the adding of a payment token vs credit card can be put in a single `add_payment()` method? So the `build_sale_request` and `build_auth_only_request` both just call `add_payment(...)` which has the `if` block and takes care of adding the payment to the hash correctly. Down the line, maybe we want to add additional payment method types which would be a single place to add them.
2. I think we should add the certification tests into the single Jetpay v2 remote tests. See Credorax for an example. Mainly to just keep things in a 1-1 match and tidy up the number of files.

shasum:
@davidsantoso 

1. Makes sense 👍 
2. Took a leaf out of [Litle](https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_litle_certification_test.rb) and [Mercury](https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_mercury_certification_test.rb). Since certification tests make us run through specific scenarios, separating them out seemed cleaner.

shasum:
@davidsantoso Updated and all tests pass (including the certification ones)!

davidsantoso:
> Took a leaf out of Litle and Mercury. Since certification tests make us run through specific scenarios, separating them out seemed cleaner.

Sounds good. Thanks for the pointer!

PR #2481, opened 2017-06-28
Fix CyberSource network tokenization XML parse errors

bdewater:
Per email exchange with CyberSource support, `PaymentNetworkToken` should appear after `ccCaptureService`, e.g.:

```xml
<ccCaptureService run="true"/>
<paymentNetworkToken>
 <transactionType>1</transactionType>
</paymentNetworkToken>
</requestMessage>
```

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -471,6 +471,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_decision_manager_fields(xml, options)
+        return unless options[:decision_manager_enabled]
+
         xml.tag! 'decisionManager' do
           xml.tag! 'enabled', options[:decision_manager_enabled] if options[:decision_manager_enabled]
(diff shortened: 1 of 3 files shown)


jasonwebster:
Is there a reproducible remote test for the before/after? Could we have a regression unit test? Seems like a good idea for some coverage, because the reason for the order is non-obvious and super not intuitive.

bdewater:
I guess we could add the XSD to the test suite and validate a generated Apple Pay request against that. It should fail without this change.

AvaNguyen:
Hi team! I just wanted to provide some additional context behind this. The merchant is looking to switch payment providers asap and so they need to get the Apple pay part working correctly before they do. There's a bit of urgency in this request and I would appreciate any prompt assistance that can be provided. Thanks!

bdewater:
Took longer then expect but adding XSD validation to all network tokenization tests uncovered that our code for Mastercard was also not putting out valid XML. CI should be all green now.

jasonwebster:
This has been replaced by https://github.com/activemerchant/active_merchant/pull/2497

PR #2483, opened 2017-06-28
Adyen: Use Active Merchant standard order_id option for reference

jasonwebster:
From https://docs.adyen.com/developers/api-reference/payments-api#paymentrequest

> A reference to uniquely identify the payment. This reference is used
> in all communication with you about the payment status. We recommend
> using a unique value per payment; however, it is not a requirement.

The `order_id` option is what is used for this throughout Active
Merchant. Let's do that for the Adyen gateway too instead of requiring a
gateway specific option.


jasonwebster:
Merged via https://github.com/activemerchant/active_merchant/commit/9321be0495dd6da2e27764841a4bdd27b3783b14

PR #2486, opened 2017-06-29
Correct calculation for three-exponent currencies

curiousepic:
The first implementation of support for three-exponent currencies used
both incorrect calculations and incorrect tests. These have been
corrected.

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -280,9 +280,9 @@ module ActiveMerchant #:nodoc:
           end
         elsif three_decimal_currency?(currency)
           if self.money_format == :cents
-            (amount.to_i * 10).to_s
+            amount.to_s
           else
-            sprintf("%.3f", amount.to_f)
+            sprintf("%.3f", (amount.to_f / 10))
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateway.rb
```diff
@@ -275,14 +275,14 @@ def localized_amount(money, currency)
         if non_fractional_currency?(currency)
           if self.money_format == :cents
             sprintf("%.0f", amount.to_f / 100)
-          else
+          elsif self.money_format == :dollars
```
davidsantoso: Curious about the explicit check for `:dollars`- since we only have `cents` and `dollars` in Active Merchant, then single `else` seems ok?

curiousepic: I did this for clarity in readability. Would it have any "logical" impact?  I'm fine reverting it on the off chance.

davidsantoso: I don't think it has a logical impact. I was just curious that maybe I was missing something. 

I think style wise I tend to say just use an `else` if there are only two options. Interestingly enough I would probably feel a little less clear since we call the money format "dollars" (which really just means "with a decimal") when the currency in question isn't actually called a "dollar".

curiousepic: 👌

PR #2489, opened 2017-06-30
SagePay: Use VPSTxId from authorization for refunds

dtykocki:
SagePay does not return the `VPSTxId` field in the response of
successful capture transactions. For refunds, the `VPSTxId` field from
authorization transactions must be used instead.

### DIFF
--- a/lib/active_merchant/billing/gateways/sage_pay.rb
+++ b/lib/active_merchant/billing/gateways/sage_pay.rb
@@ -359,9 +359,9 @@ module ActiveMerchant #:nodoc:
           response['Token']
         else
          [ params[:VendorTxCode],
-           response["VPSTxId"],
+           response["VPSTxId"] || params[:VPSTxId],
            response["TxAuthNo"],
-           response["SecurityKey"],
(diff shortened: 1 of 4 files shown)


dtykocki:
@davidsantoso / @curiousepic

davidsantoso:
@dtykocki could you drop the remote test run suite results in the original PR comment for posterity sake? Unfortunately I can't run them and add since it looks like there's IP whitelisting involved 😿

dtykocki:
@davidsantoso - I intended to, but the remote tests will not fully complete. I was forced to drop in a `sleep 200` (sometimes 300) between each HTTP request to SagePage. Even with the `sleep`, some remote tests fail with `2002 : The Authorisation timed out.`. If there's some secret handshake I need to perform with SagePay, I'd love to know what that is at this point.

davidsantoso:
@dtykocki ah gotcha. The typical gateway sandbox quirk! Alrighty, overall this looks good since using the `||` should keep previous transactions with the `VPSTxId` in the response the same. 👍

curiousepic:
LG2M2 👍

PR #2491, opened 2017-06-30
Optimal Payments: Pass CVD indicator accurately

curiousepic:
Previously, CVD Indicator was sent as '1' along with a blank CVD field,
if a CVV was not present, causing a schema error failure. Now, CVD
Indicator is properly passed as '0' with no CVD field when CVV is not
present, which appears to allow CVV-less transactions to be accepted, at
least in the test environment.

Unit:
100% passed

Remote:
14 tests, 68 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@dtykocki


curiousepic:
@davidsantoso `test_minimal_request` now tests this implicitly, but an explicit one is probably warranted.

PR #2480, opened 2017-06-26
Payflow: Set PAYPAL_NVP header as optional

davidsantoso:
This header was originally added for requests across the board, but then we discussed making it optional instead since not all users of the adapter may want the functionality it provides. See #2462 for more details.

Unit test run:

```
Finished in 14.758576 seconds.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
3615 tests, 66653 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
244.94 tests/s, 4516.22 assertions/s
```

PR #2492, opened 2017-07-04
Payflow: Move PAYPAL-NVP header option to a class attribute on the payment gateway

deuxpi:
Change the way the PAYPAL-NVP header is made optional. Adding it as an option on the `commit` method creates a special case that makes the Payflow gateway handling not to be agnostic anymore. Moving that option to the class allows integrators to set that option once during the initialization of the payment gateway.

See https://github.com/activemerchant/active_merchant/pull/2462 and https://github.com/activemerchant/active_merchant/pull/2480.

### DIFF
--- a/lib/active_merchant/billing/gateways/payflow/payflow_common_api.rb
+++ b/lib/active_merchant/billing/gateways/payflow/payflow_common_api.rb
@@ -25,6 +25,13 @@ module ActiveMerchant #:nodoc:
         # subsequent Responses will have a :duplicate parameter set in the params
         # hash.
         base.retry_safe = true
+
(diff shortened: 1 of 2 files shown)


deuxpi:
**cc** @davidsantoso @jasonwebster @jusleg

deuxpi:
Positive and negative tests (from default value) added for option flag. I amended/squashed in the same commit (not sure if that's what is usually done in active_merchant.)

jasonwebster:
It is, perfect. 👌 Thanks for putting that together.

davidsantoso:
👍 thanks! This is a good update since I imagine integrators will want to default their requests to one or the other vs on a per request basis.

Issue #2493, opened 2017-07-05
Upgrade WePay API to 2017-05-31


OldSneerJaw:
The library currently targets an outdated version of the WePay API (2017-02-01) by default. Upgrade to the most recent version: 2017-05-31.

See the [WePay API Versions](https://developer.wepay.com/docs/versions) page for more info.

bpollack:
As noted in #2494, this is part of how WePay works, and we can’t change the baseline without breaking downstream clients. You can override the version with the options hash or the WePay dashboard.

PR #2494, opened 2017-07-05
WePay: Upgrade to API version 2017-05-31

OldSneerJaw:
Updates the _default_ version targeted by the WePay gateway type to match the latest version of the WePay API: 2017-05-31.

Addresses issue #2493.


bpollack:
Thanks so much for taking the time to put this together. I'm regretfully going to reject this for similar reasons as noted in #2879: WePay's API works by assuming you will set a baseline in Dashboard, so the default only matters if you don't have one set. In that context, this PR could break existing users, while new users can simply set a new baseline appropriately.

PR #2495, opened 2017-07-05
SagePay: Make Repeat purchase if payment is a past authorization

curiousepic:
The previous implementation of Repeat purchases required a flag to be
passed in as an option. Now, the payment method is checked and if it is
a past purchase authorization string, the transaction will automatically
be sent as a Repeat purchase.

Unit:
35 tests, 117 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
36 tests, 101 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/sage_pay.rb
+++ b/lib/active_merchant/billing/gateways/sage_pay.rb
@@ -88,7 +88,7 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, options)
         add_optional_data(post, options)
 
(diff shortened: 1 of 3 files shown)


dtykocki:
@curiousepic - Just curious, how were you able to get the full set of remote tests to pass? When I was working with SagePay, I had to drop in `sleep` statements before each HTTP request.

PR #2496, opened 2017-07-06
Auth.Net: Use two character default for billing state

dtykocki:
For addresses in North America, billing state is a required field, should
not contain any symbols, and must be a valid two-character state code.
Instead of defaulting to "N/A", now default to "NC".

Remote tests:
```
ruby -Itest test/remote/gateways/remote_authorize_net_test.rb
Started
.............................................................

Finished in 44.356651 seconds.
---------------------------------------------------------------------------------------------------------------------------------------------------------------
61 tests, 210 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
---------------------------------------------------------------------------------------------------------------------------------------------------------------
1.38 tests/s, 4.73 assertions/s
```

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -555,6 +555,7 @@ module ActiveMerchant
 
         xml.billTo do
           first_name, last_name = names_from(payment_source, address, options)
+          state = state_from(address, options)
           full_address = "#{address[:address1]} #{address[:address2]}".strip
 
           xml.firstName(truncate(first_name, 50)) unless empty?(first_name)
@@ -562,7 +563,7 @@ module ActiveMerchant
           xml.company(truncate(address[:company], 50)) unless empty?(address[:company])
           xml.address(truncate(full_address, 60))
           xml.city(truncate(address[:city], 40))
-          xml.state(empty?(address[:state]) ? 'n/a' : truncate(address[:state], 40))
+          xml.state(truncate(state, 40))
           xml.zip(truncate((address[:zip] || options[:zip]), 20))
           xml.country(truncate(address[:country], 60))
           xml.phoneNumber(truncate(address[:phone], 25)) unless empty?(address[:phone])
@@ -711,6 +712,14 @@ module ActiveMerchant
           [(payment_source.first_name || first_name), (payment_source.last_name || last_name)]
         else
           [options[:first_name], options[:last_name]]
+        end
+      end
+
+      def state_from(address, options)
+        if ["US", "CA"].include?(address[:country])
+          address[:state] || 'NC'
+        else
+          address[:state]
         end
       end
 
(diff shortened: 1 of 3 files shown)


jasonwebster:
May I ask why we defaulted it to a real value rather than leaving it blank and letter the caller know it is required? Seems like it would lead to some hard to debug situations when building an integration, accidentally leaving off the state, and having it set to a valid, yet incorrect value.

dtykocki:
@jasonwebster - Unfortunately, state is not _always_ required. See https://support.authorize.net/authkb/index?page=content&id=A1454&actp=RSS for some background. Whether or not state is required seems to depend on the country of the address - a field that is also not required. 

More specifically, we've seen transaction failures due to the "N/A" default value. I was originally going to default state "XX" like I've seen in other adapters, but I suspect that value will still trigger validation errors from ANET since it's not a valid state code.

davidsantoso:
Looking in that Authorize.Net knowledge base article:

> If the address is in the U.S. or Canada, the two-digit State/Province code must be provided, along with the Zip/Postal Code.

So could be that we have a method that checks the country and if it's US or Canada, uses a default. Otherwise it allows `nil`. Although I can see the default possibly causing confusion I'm also guessing that if someone isn't collecting state then they don't really care would rather it "just work". 

In the past we've erred on the side of make successful transactions as easy as possible which meant general (but still sane) defaults where needed. It has been helpful in the past and if a default seems strange, it should be easy to override.

I'm ok with this default, but making it a little more concise by checking country would also be a worthwhile improvement.

davidsantoso:
@jasonwebster any concerns with the above?

jasonwebster:
It's fine. I still disagree (I'd just drop the `n/a` and pass through the `nil`, and let integrators fix their issues), but I don't want to be in the way. Feel free to ship on! 🌮

davidsantoso:
👍 Makes sense. @dtykocki I'll leave it up to you to decide if we wanna make it a little more concise in terms of not defaulting a state if the country is outside the US and Canada.

dtykocki:
Cleaned this up a bit to only send a default when the country is US or CA.

davidsantoso:
@dtykocki Thanks 🌮 !

jasonwebster:
Heya, just a friendly reminder to update the changelog when you merge. 🙏

jasonwebster:
Funny story! 

I'm starting to see some `Bill To State/Province is required` declines from Authorize.net now, at least for UK customers. 

I took a look at some other Authorize.net  documentation here http://developer.authorize.net/api/reference/#payment-transactions-charge-a-credit-card and it says something totally different than the one above for `billTo.state`:

> State of customer’s billing address. Required only when using a European Payment Processor.

So, I suspect that the `n/a` value was initially meant to cover those cases-- European countries that don't use a state/province.

davidsantoso:
> it says something totally different than the one above for billTo.state

☹️ 

I suppose we can't really know when a European Payment Processor is being used. Makes me wonder if we default to back to `n/a` when it's outside the US and Canada?

```ruby
def state_from(address, options)
  if ["US", "CA"].include?(address[:country])
    address[:state] || 'NC'
  else
    address[:state] || 'n/a'
  end
end
```

jasonwebster:
Yeah, I think that's best?

davidsantoso:
👍 I'll put up a PR!

jasonwebster:
I already did! https://github.com/activemerchant/active_merchant/pull/2563

davidsantoso:
🐎 !

PR #2497, opened 2017-07-06
Add XSD schema validation to CyberSource tests

jasonwebster:
The CyberSource Simple (heh) Order API has some strict element ordering
requirements around auths, captures and network tokenization fields.
This adds a schema validator to ensure that we're always compliant.

As per an email exchange we had with CyberSource support, the
`paymentNetworkToken` element must always come after the
`ccCaptureService` element. This was not the case for us for purchase
transactions, as we set it in the `add_auth_service` method, which means
it was sandwiched between `ccAuthService` and `ccCaptureService`.

This also cleans up a few other XSD validation violations, none of which
should be significant--just some empty tags here and there.

Also, I renamed `add_network_tokenization` to
`add_auth_network_tokenization`, since that's actually what it was
doing--it was only used for authorization transactions (as it was
responsible to for adding the `ccAuthService` commands to the request).

Originally by @bdewater in https://github.com/activemerchant/active_merchant/pull/2481

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -258,6 +258,7 @@ module ActiveMerchant #:nodoc:
         add_decision_manager_fields(xml, options)
         add_mdd_fields(xml, options)
         add_auth_service(xml, creditcard_or_reference, options)
+        add_payment_network_token(xml) if network_tokenization?(creditcard_or_reference)
         add_business_rules_data(xml, creditcard_or_reference, options)
         xml.target!
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -471,13 +474,17 @@ def add_creditcard(xml, creditcard)
       end
 
       def add_decision_manager_fields(xml, options)
+        return unless options[:decision_manager_enabled]
+
         xml.tag! 'decisionManager' do
           xml.tag! 'enabled', options[:decision_manager_enabled] if options[:decision_manager_enabled]
```
bdewater: Come to realize the `if options[:decision_manager_enabled]` is redundant now

PR #2499, opened 2017-07-06
NMI: Add first and lastname to echeck transactions

dtykocki:
Unit:
```
27 tests, 194 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

Remote:
```
33 tests, 106 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/lib/active_merchant/billing/gateways/nmi.rb
+++ b/lib/active_merchant/billing/gateways/nmi.rb
@@ -146,6 +146,8 @@ module ActiveMerchant #:nodoc:
           post[:token_cryptogram] = payment_method.payment_cryptogram
         elsif(card_brand(payment_method) == 'check')
           post[:payment] = 'check'
+          post[:firstname] = payment_method.first_name
+          post[:lastname] = payment_method.last_name
           post[:checkname] = payment_method.name
           post[:checkaba] = payment_method.routing_number
           post[:checkaccount] = payment_method.account_number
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/nmi.rb
```diff
@@ -146,6 +146,8 @@ def add_payment_method(post, payment_method, options)
           post[:token_cryptogram] = payment_method.payment_cryptogram
         elsif(card_brand(payment_method) == 'check')
           post[:payment] = 'check'
+          post[:firstname] = payment_method.first_name
+          post[:lastname] = payment_method.last_name
```
curiousepic: I'm not positive of this, but might a card/check not have a last name?

dtykocki: I wasn't sure at first either, but both `CreditCard` and `Check` have first and lastname:
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card.rb#L117-L122
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/check.rb#L9

curiousepic: @dtykocki Ah, I did not see the validation at first, but I found it 👍

PR #2503, opened 2017-07-11
Worldpay: Support Credit on CFT-enabled merchant IDs

curiousepic:
Worldpay only supports Credits (aka Payouts) by sending a transaction
with the same structure as an Authorization, but with its action flagged
as REFUND and a Merchant ID (login credential) that has been flagged by
Worldpay to treat these transactions as a Credit (they call it a Credit
Fund Transfer or CFT). Purchases, normal Refunds, and other transactions
should only be performed with the separate Merchant ID flagged as eCom.

Unit:
37 tests, 205 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
22 tests, 73 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -72,6 +72,14 @@ module ActiveMerchant #:nodoc:
         void(authorization, options ) if response.params["last_event"] == "AUTHORISED"
       end
 
+      # Credits only function on a Merchant ID/login/profile flagged for Payouts
+      #   aka Credit Fund Transfers (CFT), whereas normal purchases, refunds,
+      #   and other transactions should be performed on a normal eCom-flagged
+      #   merchant ID.
+      def credit(money, payment_method, options = {})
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -217,7 +229,7 @@ def add_payment_method(xml, amount, payment_method, options)
             end
           end
         else
-          xml.tag! 'paymentDetails' do
+          xml.tag! 'paymentDetails', add_attribute_for_cft_if_credit(options) do
```
davidsantoso: Sorry, this is very much a nitpick... how would you feel about a method name of:

`credit_fund_transfer_attribute(options)`?

curiousepic: Sounds good - updated and added a guard clause in the method.

PR #2509, opened 2017-07-14
Realex: update avs/address code creation

slogsdon:


bpollack:
Thanks for the PR. If this is still relevant to you (sorry it's been awhile!), I'd appreciate at least some information about what this is correcting/fixing and why, and ideally also an equivalent note in the remote tests so we can verify that it works properly.

slogsdon:
@bpollack Thanks for following up.

This change corrects AVS behavior for non-GB countries that support AVS through the Realex gateway. Here's the relevant Realex documentation that covers this functionality: https://developer.realexpayments.com/#!/api/fraud-management/avs. Essentially, only GB address should strip non-numeric characters from the postal code and first address line when constructing the `code` field that is sent to the gateway. Previously, it would incorrectly perform this for all cases.

Are you able to confirm if you're looking to add tests/assertions for this change to match the unit test assertions?

PR #2521, opened 2017-07-31
Allow `device_data` to be provided when storing card

eric1234:
This parameter is documented in [Braintree's API](https://developers.braintreepayments.com/reference/request/customer/create/ruby#device_data) but not available to be used in ActiveMerchant. This commit makes the parameters available.

You can consider this a follow-up to 8dba9d3c which added the `device_data` option when creating a transaction. This is the same only it is for when vaulting a card.

NOTE: When creating a card on an existing customer, [Braintree's API](https://developers.braintreepayments.com/reference/request/credit-card/create/ruby) doesn't seem to have the `device_data` param so this only affects cases where we are creating a customer and card together via `store`.

NOTE: When updating an existing customer there is [the option](https://developers.braintreepayments.com/reference/request/customer/update/ruby) to
provide `device_data`. I left that out of this commit as my app is not using `update` and I'm hesitant to add code I am not also testing within the context of an app. Leaving the scope of this commit to just what I am using.

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -230,6 +230,7 @@ module ActiveMerchant #:nodoc:
             :email => scrub_email(options[:email]),
             :id => options[:customer],
           }.merge credit_card_params
+          parameters[:device_data] = options[:device_data] if options[:device_data]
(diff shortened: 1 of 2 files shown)


curiousepic:
This looks good, but needs a correction in the test by including a `:phone` in the customer stubs.

eric1234:
I no longer work on this code. Will close it. If anybody want to take it up they can of course.

PR #2522, opened 2017-08-01
Litle: Update schema and certification tests to v9.12

curiousepic:
Also updates Android Pay support to pass the ordersource correctly.

Unit:
32 tests, 137 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
28 tests, 117 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

I'm happy to consider an alternate method of passing the prelive url for the cert tests.

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -3,7 +3,7 @@ require 'nokogiri'
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class LitleGateway < Gateway
-      SCHEMA_VERSION = '9.4'
+      SCHEMA_VERSION = '9.12'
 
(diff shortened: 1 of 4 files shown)


davidsantoso:
> I'm happy to consider an alternate method of passing the prelive url for the cert tests.

@curiousepic was there an issue with the pervious setup of merging in a `:url` to the initializer? Seems like we could continue with that method and just update the URL in the certification test vs needing to passing in a new `prelive_url: true` option.

curiousepic:
@davidsantoso Good catch... not sure what happened there but it does seem to work fine. I'll switch back to that method.

curiousepic:
Actually it wasn't quite that simple, but found a good similar method.

kjbkjb:
@curiousepic Thank you for maintaining the Vantiv/Litle AM module. It is appreciated.

PR #2524, opened 2017-08-01
stripe emv transactions send statement_addess parameters

malcolm-mergulhao:
related to: https://gist.github.com/bct/53ffdf4dd4dab2540f7b#including-a-custom-address-on-an-emv-transaction
This will allow us to pass `statement_address` in through options and have them added to emv authorize and purchase requests. This is accomplished by adding these parameters to the post in `create_post_for_auth_or_purchase` if the transaction is an EMV payment.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -303,7 +303,9 @@ module ActiveMerchant #:nodoc:
           add_creditcard(post, payment, options)
         end
 
-        unless emv_payment?(payment)
+        if emv_payment?(payment)
+          add_statement_address(post, options)
+        else
           add_amount(post, money, options, true)
           add_customer_data(post, options)
           post[:description] = options[:description]
@@ -363,6 +365,18 @@ module ActiveMerchant #:nodoc:
           post[:card][:address_state] = address[:state] if address[:state]
           post[:card][:address_city] = address[:city] if address[:city]
         end
+      end
+
+      def add_statement_address(post, options)
+        return unless statement_address = options[:statement_address]
+        return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? } 
+
+        post[:statement_address] = {}
+        post[:statement_address][:line1] = statement_address[:address1]
+        post[:statement_address][:line2] = statement_address[:address2] if statement_address[:address2].present?
+        post[:statement_address][:city] = statement_address[:city]
+        post[:statement_address][:postal_code] = statement_address[:zip]
+        post[:statement_address][:state] = statement_address[:state]
       end
 
       def add_creditcard(post, creditcard, options)
--- a/test/unit/gateways/stripe_test.rb
+++ b/test/unit/gateways/stripe_test.rb
@@ -12,6 +12,7 @@ class StripeTest < Test::Unit::TestCase
 
     @options = {
       :billing_address => address(),
+      :statement_address => statement_address(),
       :description => 'Test Purchase'
     }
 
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -365,6 +367,21 @@ def add_address(post, options)
         end
       end
 
+      def add_statement_address(post, options)
+        statement_address = options[:statement_address]
```
bizla: You can `return unless statement_address = options[:statement_address]`

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -365,6 +367,21 @@ def add_address(post, options)
...
+
+        return unless statement_address && statement_address[:address1] &&
```
bizla: Would you maybe prefer
```ruby
return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? } 
```

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -365,6 +367,21 @@ def add_address(post, options)
...
+          statement_address[:city] &&
+          statement_address[:zip] &&
+          statement_address[:state]
+        post[:statement_address] = {}
+        post[:statement_address][:line1] = statement_address[:address1]
+        post[:statement_address][:line2] = statement_address[:address2]
+        post[:statement_address][:city] = statement_address[:city]
+        post[:statement_address][:postal_code] = statement_address[:zip]
```
bizla: What if I pass in a `statement_address[:postal_code]` value instead of `:zip`? Should this accept either one?

malcolm-mergulhao: it could. I chose zip instead of postal code to stay consistent with the billing address parameters that get passed in but if we want the values to match with the ones stripe accepts thats fine too. I guess it depends on what we were thinking of passing in through the options.

lib/active_merchant/billing/gateways/stripe.rb
malcolm-mergulhao: yeah, that would be better

lib/active_merchant/billing/gateways/stripe.rb
jasonwebster: Nope, you've got it right. ActiveMerchant's contact for `billing_address` and `shipping_address` is as follows: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L39-L49

jasonwebster: Which, I suppose, raises a question... is this a  gateway specific option, or an Active Merchant normalization?

davidseal:
Thank you! 🎉

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -311,6 +311,8 @@ def create_post_for_auth_or_purchase(money, payment, options)
           post[:receipt_email] = options[:receipt_email] if options[:receipt_email]
           add_customer(post, payment, options)
           add_flags(post, options)
+        else
```
ryanbalsdon: Optional: I think this would read a bit easier now as:
```
if emv_payment?(payment)
  add_statement_address(post, options)
else
  add_amount(post, money, options, true)
  post[:description] = options[:description]
  post[:statement_descriptor] = options[:statement_description]
  post[:receipt_email] = options[:receipt_email] if options[:receipt_email]
  add_customer(post, payment, options)
  add_flags(post, options)
end
```

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -365,6 +367,18 @@ def add_address(post, options)
         end
       end
 
+      def add_statement_address(post, options)
+        return unless statement_address = options[:statement_address]
+        return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? } 
+
+        post[:statement_address] = {}
+        post[:statement_address][:line1] = statement_address[:address1]
+        post[:statement_address][:line2] = statement_address[:address2]
```
ryanbalsdon: You aren't checking for presence a address2 up above. Would setting this to nil affect the serialization at all?

malcolm-mergulhao: oh yeah i meant to have a check on that line to make sure address2 was present. The reason i didnt check above is that its not required.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -365,6 +367,18 @@ def add_address(post, options)
         end
       end
 
+      def add_statement_address(post, options)
+        return unless statement_address = options[:statement_address]
+        return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? }
```
bizla: We should add a test for this, something like:
```ruby 
def test_add_statement_address_returns_nil_if_fields_missing
  post = {}
  [:address1, :city, :zip, :state].each do |required_key|
    assert_equal nil, @gateway.send(:add_statement_address, post, @options.delete_if { |k| k == required_key })
  end
end
```

malcolm-mergulhao:
@bizla I added the test for `statement_address` not being populated when required fields are missing.

test/unit/gateways/stripe_test.rb
```diff
@@ -898,6 +899,32 @@ def test_add_address
...
+
+    assert_equal @options[:statement_address][:zip], post[:statement_address][:postal_code]
+    assert_equal @options[:statement_address][:state], post[:statement_address][:state]
+    assert_equal @options[:statement_address][:address1], post[:statement_address][:line1]
+    assert_equal @options[:statement_address][:address2], post[:statement_address][:line2]
+    assert_equal @options[:statement_address][:country], post[:statement_address][:country]
+    assert_equal @options[:statement_address][:city], post[:statement_address][:city]
+  end
+
+  def test_add_statement_address_returns_nil_if_required_fields_missing
```
bizla: 😍

malcolm-mergulhao:
I squashed my commits, its looks like I cant merge though so whenever you guys want this should be good to go.

AnotherJoSmith:
@malcolm-mergulhao Mind adding a changelog entry?

PR #2525, opened 2017-08-01
FirstData E4: Override ECI value to `04` for tokenized Discover cards

jasonwebster:
We've been in contact with Bank of America and FirstData, and they've
informed us that Discover is **requiring** all Apple Pay transactions to
use the "in-app" indicator of `04`. I chose to do so for all tokenized cards,
because I have to imagine the same requirements would be true for
all digital wallets--not just Apple Pay.

In addition, I've added a loose XSD validation for the request payloads.
I've extracted this from the official WSDL for the version we use, which
you can find here:
https://api.globalgatewaye4.firstdata.com/transaction/v11/wsdl

I've made some changes to the `Transaction` object to _not_ define a
sequence. We do not send the elements in the order that the schema defines,
but that doesn't appear to have ever been a problem. Now, This isn't great,
but it avoids a significant refactor of the code to get elements in the
correct sequence.

The above led me to fix duplicate XID and CAVV values when processing tokenized cards.
These were being added twice, once with the correct values and again empty.

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4.rb
@@ -236,11 +236,19 @@ module ActiveMerchant #:nodoc:
           xml.tag! "CardHoldersName", credit_card.name
           xml.tag! "CardType", card_type(credit_card.brand)
 
-          eci = (credit_card.respond_to?(:eci) ? credit_card.eci : nil) || options[:eci] || DEFAULT_ECI
-          xml.tag! "Ecommerce_Flag", eci
-
+          add_eci(xml, credit_card, options)
           add_credit_card_verification_strings(xml, credit_card, options)
         end
+      end
+
+      def add_eci(xml, credit_card, options)
+        eci = if credit_card.is_a?(NetworkTokenizationCreditCard) && credit_card.brand == "discover"
+          "04"
+        else
+          (credit_card.respond_to?(:eci) ? credit_card.eci : nil) || options[:eci] || DEFAULT_ECI
+        end
+
+        xml.tag! "Ecommerce_Flag", eci
       end
 
       def add_credit_card_verification_strings(xml, credit_card, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/firstdata_e4.rb
```diff
@@ -237,13 +236,21 @@ def add_credit_card(xml, credit_card, options)
...
-          eci = (credit_card.respond_to?(:eci) ? credit_card.eci : nil) || options[:eci] || DEFAULT_ECI
-          xml.tag! "Ecommerce_Flag", eci
-
+          add_eci(xml, credit_card, options)
           add_credit_card_verification_strings(xml, credit_card, options)
         end
       end
 
+      def add_eci(xml, credit_card, options)
+        eci = if credit_card.is_a?(NetworkTokenizationCreditCard) && credit_card.brand == "discover"
```
jasonwebster: I'd like the record to reflect how dumb I think this is.

test/unit/gateways/firstdata_e4_test.rb
```diff
@@ -286,6 +309,13 @@ def test_supports_network_tokenization
 
   private
 
+  def assert_xml_valid_to_wsdl(data)
```
bdewater: 🏆

lib/active_merchant/billing/gateways/firstdata_e4.rb
```diff
@@ -237,13 +236,21 @@ def add_credit_card(xml, credit_card, options)
...
+          "04"
```
elfassy: Is it `Discover is requiring 04` or `FirstData is requiring that Discover cards have 04`? I'm assuming we haven't seen similar errors with discover elsewhere.

jasonwebster: IMO this is a failure of abstraction on FirstData's part. Who knows what other gateways do. I'd assume most normalize data across card networks, but here we're being bit by a leaky abstraction.

jnormore: It's almost certainly a network level requirement, so all discover cards, and not just FD.

jasonwebster:
Going to close this and re-open with just the schema validation and CAVV dupe fix. Going to solve the ECI issue generally elsewhere.

jasonwebster:
Relevant bits extracted into https://github.com/activemerchant/active_merchant/pull/2529

PR #2528, opened 2017-08-01
PayU Latam: Add additional mandatory fields.

deedeelavinder:
Remote:
15 tests, 39 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
86.6667% passed
(These two failures both have a responseCode of 'INTERNAL_PAYMENT_PROVIDER_ERROR' and pre-date these changes.)

Unit:
17 tests, 71 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -49,7 +49,7 @@ module ActiveMerchant #:nodoc:
         post = {}
 
         add_credentials(post, 'SUBMIT_TRANSACTION')
-        add_transaction_type(post, 'CAPTURE')
+        add_transaction_elements(post, 'CAPTURE', options)
         add_reference(post, authorization)
 
         commit('capture', post)
@@ -59,7 +59,7 @@ module ActiveMerchant #:nodoc:
         post = {}
 
         add_credentials(post, 'SUBMIT_TRANSACTION')
-        add_transaction_type(post, 'VOID')
+        add_transaction_elements(post, 'VOID', options)
         add_reference(post, authorization)
 
         commit('void', post)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -236,6 +253,9 @@ def add_payer(post, options)
           payer[:fullName] = address[:name]
           payer[:contactPhone] = address[:phone]
           payer[:dniNumber] = options[:dni_number]
+          payer[:dniType] = options[:dni_type]
```
dtykocki: Should this have a guard like we do in https://github.com/deedeelavinder/active_merchant/blob/dbd6f141c6ae0e3db15d3b3f37d0bf50f5c79663/lib/active_merchant/billing/gateways/payu_latam.rb#L163?

deedeelavinder: 👍

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -174,8 +181,18 @@ def add_invoice(post, money, options)
         tx_value[:value] = amount(money)
         tx_value[:currency] = options[:currency] || currency(money)
 
+        tx_tax = {}
+        tx_tax[:value] = options[:tx_tax] || '0'
```
bdewater: Other gateways seem to use `options[:tax]`

curiousepic: @deedeelavinder True, the `tx` for `tx_tax` and `tx_tax_base` is extraneous from the perspective of the transactor, probably worth changing.

deedeelavinder: 👍

PR #2529, opened 2017-08-02
FirstData E4: Fix duplicate XID and CAVV values in tokenized transactions

jasonwebster:
I've added a loose XSD validation for the request payloads.
I've extracted this from the official WSDL for the version we use, which
you can find here:
https://api.globalgatewaye4.firstdata.com/transaction/v11/wsdl

I've made some changes to the Transaction object to not define a
sequence. We do not send the elements in the order that the schema defines,
but that doesn't appear to have ever been a problem. Now, This isn't great,
but it avoids a significant refactor of the code to get elements in the
correct sequence. Ticked this debt out here: https://github.com/activemerchant/active_merchant/issues/2533

The above led me to fix duplicate XID and CAVV values when processing tokenized cards.
These were being added twice, once with the correct values and again empty.

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4.rb
@@ -156,7 +156,7 @@ module ActiveMerchant #:nodoc:
         xml = Builder::XmlMarkup.new
 
         xml.instruct!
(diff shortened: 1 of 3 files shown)


jnormore:
:+1:

PR #2530, opened 2017-08-02
Allow Response Code 4 to be returned as successful

nfarve:
Currently only response code 1 is being recorded as a successful
transaction, however response_code 4 is also successful but held for
fraud review.

Started
.............................................................

Finished in 40.926721 seconds.
61 tests, 210 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
1.49 tests/s, 5.13 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -898,7 +898,7 @@ module ActiveMerchant
         if cim?(action) || (action == :verify_credentials)
           response[:result_code] == "Ok"
         else
-          response[:response_code] == APPROVED && TRANSACTION_ALREADY_ACTIONED.exclude?(response[:response_reason_code])
+          [APPROVED, FRAUD_REVIEW].include?(response[:response_code]) && TRANSACTION_ALREADY_ACTIONED.exclude?(response[:response_reason_code])
         end
       end
 
--- a/test/unit/gateways/authorize_net_test.rb
+++ b/test/unit/gateways/authorize_net_test.rb
@@ -826,7 +826,7 @@ class AuthorizeNetTest < Test::Unit::TestCase
     @gateway.expects(:ssl_post).returns(fraud_review_response)
 
     response = @gateway.purchase(@amount, @credit_card)
-    assert_failure response
+    assert_success response
     assert response.fraud_review?
     assert_equal "Thank you! For security reasons your order is currently being reviewed", response.message
   end


nfarve:
@davidsantoso @dtykocki @curiousepic @deedeelavinder 
Another activemerchant change. 

I didn't see a way to create a fraud transaction, so was unable to write a test. If you have suggestions definitely open to them. 

https://developer.authorize.net/hello_world/testing_guide/

davidsantoso:
> I didn't see a way to create a fraud transaction

@nfarve it can be hit or miss with remote tests/gateway sandboxes providing a mechanism to test all possible responses. My only suggestion would be to write a unit test instead with a response that would be returned in the case of a fraud review. So, a stubbed out response that returns `<responseCode>4</responseCode>` instead and checking that it's still considered to be successful.

nfarve:
@davidsantoso looks there was already a unit test. I updated it to test for success instead of failure.

jasonwebster:
Heya friends! 👋 Could you provide a bit of clarification on this change?

If I use Authorize.net's docs, it seems to indicate that a Reason Code of 4 should be treated like a decline, at least for sale/purchase transactions.

https://www.dropbox.com/s/02gn89bq3s2ftds/Screenshot%202018-02-22%2010.16.06.png?dl=0
https://developer.authorize.net/api/reference/responseCodes.html

dtykocki:
Howdy @jasonwebster - hopefully I can help shed some light on this one. There were instances where funds were debited from an account even after receiving this responseCode 4 and under the previous implementation, the transaction was treated as a failure. This left a number of folks quite confused. 

The documentation we used for this was generally ambiguous when describing responseCode 4:
response_codes Also, based on a [community article](https://community.developer.authorize.net/t5/Integration-and-Testing/The-transaction-is-currently-under-review/td-p/23077), we were left with the impression that there's really know way to know whether or not the transaction will complete.

Does that make sense?

dtykocki:
That said, after looking through the links you posted @jasonwebster I am thinking that `responseCode` of 4 should be a decline or the success criteria needs to account for a combination of codes.

jasonwebster:
Huh, yeah, thanks for the quick follow up. I wonder if we're looking at different versions of documentation.

jasonwebster:
Do you recall exactly which documentation you found that in? I found another link (https://developer.authorize.net/api/reference/features/errorandresponsecodes.html) that seems to support `4` as being a decline as well.

dtykocki:
That came from a PDF document named "Merchant Web Services API Advanced Integration Method (AIM) XML Guide" we have saved off. I just tried to find it on ANET's site, but it looks like it's been removed https://www.authorize.net/content/dam/authorize/documents/AIM_guide_XML.pdf - which is not too promising 😄.

PR #2532, opened 2017-08-03
Remove param[:order_number] from captures in Forte Gateway

nfarve:
When processing captures in the forte gateway only admin messages are
allowed. Adding the param[:order_number] produces an F03 invalid field name
error.

Loaded suite test/remote/gateways/remote_forte_test
Started

Finished in 40.85056 seconds.
18 tests, 45 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,

0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_forte_test.rb
+++ b/test/remote/gateways/remote_forte_test.rb
@@ -21,8 +21,10 @@ class RemoteForteTest < Test::Unit::TestCase
 
     @options = {
       billing_address: address,
-      description: 'Store Purchase'
+      description: 'Store Purchase',
+      order_id: '1'
     }
+
   end
 
   def test_invalid_login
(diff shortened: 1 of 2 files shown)


nfarve:
@davidsantoso @curiousepic @dtykocki @deedeelavinder

test/remote/gateways/remote_forte_test.rb
```diff
@@ -23,6 +23,9 @@ def setup
       billing_address: address,
       description: 'Store Purchase'
     }
+
+    @options_with_order_id = @options
+    @options_with_order_id.update(order_id:'2')
```
curiousepic: Is there any reason to not send an `order_id` with all transactions? Most test suites include the `order_id` in the default `options` hash.

nfarve: I wasn't sure if we try to only have the bare minimums in the options hash. Definitely can add to all though.

curiousepic: Cool. I think `order_id` is typically considered part of the minimum.

nfarve:
@curiousepic added order_id to the order instance and went back to using that for all of the tests!

PR #2536, opened 2017-08-04
Vantiv (Litle): Pass 3DS fields

curiousepic:
Remote:
29 tests, 120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
32 tests, 137 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -175,7 +175,7 @@ module ActiveMerchant #:nodoc:
         add_order_source(doc, payment_method, options)
         add_billing_address(doc, payment_method, options)
         add_shipping_address(doc, payment_method, options)
-        add_payment_method(doc, payment_method)
+        add_payment_method(doc, payment_method, options)
         add_pos(doc, payment_method)
         add_descriptor(doc, options)
         add_debt_repayment(doc, options)
@@ -194,7 +194,7 @@ module ActiveMerchant #:nodoc:
         doc.debtRepayment(true) if options[:debt_repayment] == true
       end
 
-      def add_payment_method(doc, payment_method)
+      def add_payment_method(doc, payment_method, options)
         if payment_method.is_a?(String)
           doc.token do
             doc.litleToken(payment_method)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -214,6 +214,11 @@ def add_payment_method(doc, payment_method)
             doc.cardholderAuthentication do
               doc.authenticationValue(payment_method.payment_cryptogram)
             end
+          elsif options[:order_source] && options[:order_source].start_with?('3ds')
```
davidsantoso: Are there other values for `order_source` that do not include 3DS related info? Just wondering if we need the additional `options[:order_source].start_with?('3ds')` conditional.

test/remote/gateways/remote_litle_certification_test.rb
```diff
@@ -803,6 +822,25 @@ def sale_assertions(amount, card, options, assertions={})
...
+      order_id: test_id,
+      order_source: source,
+      cavv: 'BwABBJQ1AgAAAAAgJDUCAAAAAAA=',
+      xid: 'BwABBJQ1AgAAAAAgJDUCAAAAAAA='
+    }
+
+    assert response = @gateway.authorize(10100, credit_card, options)
+    assert_success response
+    assert_equal result, response.params['fraudResult_authenticationResult']
+    puts "Test #{test_id}: #{txn_id(response)}"
```
davidsantoso: 🔪

lib/active_merchant/billing/gateways/litle.rb
curiousepic: @davidsantoso There are; it's a pre-existing gateway specific field.

test/remote/gateways/remote_litle_certification_test.rb
curiousepic: This isn't just for debugging, I added these `puts`es to aid in certification - running the cert test will output all the txn_ids needed to provide to Vantiv and which test they're associated with. It's conceivable this would need to be done again in the future.

curiousepic: (note the certification test suite is separate from normal remote tests)

lib/active_merchant/billing/gateways/litle.rb
davidsantoso: 👍

test/remote/gateways/remote_litle_certification_test.rb
davidsantoso: I gotcha. Thanks for the clarification!

PR #2541, opened 2017-08-09
Fixing issue where partial refund actually voids the whole transaction

nfarve:
The incorrect endpoint was being used for refunds. This updates the
refund method so that a partial refund can be done.

Remote tests are not possible since refunds can only be done on settled
transctions. A unit test was updated to check the amount returned.

Closes ENRG-6905

Remote tests:
Finished in 0.016094 seconds.
11 tests, 25 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
683.48 tests/s, 1553.37 assertions/s

Remote tests:
Finished in 8.146803 seconds.
12 tests, 41 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
1.47 tests/s, 5.03 assertions/s

### DIFF
--- a/test/remote/gateways/remote_trans_first_test.rb
+++ b/test/remote/gateways/remote_trans_first_test.rb
@@ -70,13 +70,32 @@ class RemoteTransFirstTest < Test::Unit::TestCase
     assert_equal 'Insufficient funds', response.message
   end
 
-  def test_successful_refund
+  def test_successful_void
     assert purchase = @gateway.purchase(@amount, @credit_card, @options)
     assert_success purchase
 
-    assert refund = @gateway.refund(@amount, purchase.authorization)
-    assert_success refund
+    assert void = @gateway.void(purchase.authorization)
+    assert_success void
   end
+
+  # Refunds can only be successfully run on settled transactions which take 24 hours 
+  # def test_successful_refund
+  #   assert purchase = @gateway.purchase(@amount, @credit_card, @options)
+  #   assert_success purchase
+
+  #   assert refund = @gateway.refund(@amount, purchase.authorization)
+  #   assert_equal @amount, refund.params["amount"].to_i*100
+  #   assert_success refund
+  # end
+
+  # def test_successful_partial_refund
+  #   assert purchase = @gateway.purchase(@amount, @credit_card, @options)
+  #   assert_success purchase
+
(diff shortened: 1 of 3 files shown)


nfarve:
@deedeelavinder @dtykocki @davidsantoso @curiousepic

test/remote/gateways/remote_trans_first_test.rb
```diff
@@ -70,12 +70,12 @@ def test_failed_purchase
     assert_equal 'Insufficient funds', response.message
   end
 
-  def test_successful_refund
```
davidsantoso: @nfarve we'll actually want to keep this test even though TransFirst has a long delay between a purchase and the transaction fully settling. In the future someone could want to do a remote test for refund and just manually run a purchase and wait to run the refund test till the next day. Although we should have a remote `void` test so we should add that one while we're at it! I'm wondering why this never had a `void` test to begin with 🤔  

Also, it could be good to add a partial refund remote test- [for example EBANX](https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_ebanx_test.rb#L90-L97)- and this can follow the same test pattern as a full refund test where the "tester" just waits until the next day to exercise the test.

nfarve: Okay so keep the refund method but maybe test for failure with the right message?

davidsantoso: Well, this is where remote tests get tricky since we sort of have to work around how a gateway's sandbox is functioning. We should still test that it returns the right message, is successful, etc., but just know (or write a short comment to explain) that we need to wait for the transaction to settle which could take ~24 hours. It's unfortunate that the full remote test will never fully pass in one run but the test will still need to assert the response we want assuming the transaction was settled.

curiousepic: @nfarve Here's a past example of what I've done in this kind of situation: https://github.com/activemerchant/active_merchant/blob/9c44975612deec9d6795d1ff385071a9ae5106ae/test/remote/gateways/remote_global_collect_test.rb#L79

nfarve:
@davidsantoso @curiousepic 
Okay so I added the refund and partial_refund and just commented them out per DP's suggestion. 

I also added the void tests to the unit tests since they were missing.

PR #2543, opened 2017-08-11
Parsing additional params from responses

nfarve:
Currently we are not parsing the CardDetails from the response message,
which is passed on auth and refund transactions.

Loaded suite test/remote/gateways/remote_creditcall_test
Started
Finished in 0.603947 seconds.
1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0
notifications
100% passed
1.66 tests/s, 3.31 assertions/s

Loaded suite test/unit/gateways/creditcall_test
Started
Finished in 0.099526 seconds.
16 tests, 50 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
160.76 tests/s, 502.38 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/creditcall.rb
+++ b/lib/active_merchant/billing/gateways/creditcall.rb
@@ -26,10 +26,16 @@ module ActiveMerchant #:nodoc:
           r.process { capture(money, r.authorization, options) }
         end
 
+        if multi_response.responses[1].nil?
+          merged_params = multi_response.primary_response.params
+        else
+          merged_params =  multi_response.responses[0].params.merge(multi_response.responses[1].params)
+        end
+
         Response.new(
           multi_response.primary_response.success?,
           multi_response.primary_response.message,
-          multi_response.primary_response.params,
+          merged_params,
           authorization: multi_response.responses.first.authorization,
           test: test?
         )
@@ -156,6 +162,16 @@ module ActiveMerchant #:nodoc:
             childnode_to_response(response, childnode)
           end
         end
+
+        node = xml.xpath("//Response/CardDetails")
+        node.children.each do |childnode|
+          if childnode.elements.empty?
+            response[childnode.name] = childnode.text
+          else
+            childnode_to_response(response, childnode)
+          end
+        end
+
 
         response
       end
--- a/test/unit/gateways/creditcall_test.rb
+++ b/test/unit/gateways/creditcall_test.rb
@@ -147,7 +147,7 @@ class CreditcallTest < Test::Unit::TestCase
 
   def successful_purchase_response
     %(
-    <?xml version=\"1.0\" encoding=\"utf-8\"?><Response type=\"CardEaseXML\" version=\"1.0.0\"><TransactionDetails><CardEaseReference>0999da90-b342-e511-b302-00505692354f</CardEaseReference><LocalDateTime format=\"yyyyMMddHHmmss\">20150814143753</LocalDateTime><UTC format=\"yyyyMMddHHmmss\">20150814183753</UTC></TransactionDetails><Result><LocalResult>0</LocalResult><AuthorisationEntity>Unknown</AuthorisationEntity></Result></Response>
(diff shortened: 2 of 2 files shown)


nfarve:
@davidsantoso @dtykocki @curiousepic @deedeelavinder

lib/active_merchant/billing/gateways/creditcall.rb
```diff
@@ -26,10 +26,16 @@ def purchase(money, payment_method, options={})
           r.process { capture(money, r.authorization, options) }
         end
 
+        if multi_response.responses[1].nil?
```
curiousepic: Do any of the remote tests exercise this condition?

lib/active_merchant/billing/gateways/creditcall.rb
```diff
@@ -26,10 +26,16 @@ def purchase(money, payment_method, options={})
...
+          merged_params = multi_response.primary_response.params
+        else
+          merged_params =  multi_response.responses[0].params.merge(multi_response.responses[1].params)
```
curiousepic: This would overwrite any `auth` params of the same name from the `capture`. Just want to make sure that's safe in this context?

davidsantoso:
@nfarve was just looking at this and can see the params from the previous `Authorize` transaction as being pretty helpful/easy to access from the Active Merchant response. Interestingly enough, all of the params returned from a `capture` are present in the params of the `authorize`:

Authorize
```
{
  "CardEaseReference"=>"589a356a-f980-e711-ba27-0050569277e2",
  "LocalDateTime"=>"20170814100404",
  "UTC"=>"20170814140404",
  "LocalResult"=>"0",
  "AcquirerResponseCode"=>"00",
  "AuthCode"=>"D2F278",
  "AuthorisationEntity"=>"Unknown",
  "AmountOnlineApproved"=>"1.00",
  "CardReference"=>"1031a7a3-2414-e711-94f9-002219649f24",
  "CardHash"=>"YO5CqO1QQbHbMaFRuSV2vVq7uqQ=",
  "PAN"=>"xxxxxxxxxxxx2224",
  "ExpiryDate"=>"1809",
  "Description"=>"VISA",
  "CSC"=>"matched",
  "Address"=>"matched",
  "Zip"=>"matched",
  "ICCTag"=>"000000000100"
}
```

Capture
```
{
  "CardEaseReference"=>"5f9a356a-f980-e711-ba27-0050569277e2",
  "LocalDateTime"=>"20170814100406",
  "UTC"=>"20170814140406",
  "LocalResult"=>"0",
  "AuthorisationEntity"=>"Unknown"
}
```

Makes me wonder if we should even use the params from the `capture`! Or, if we want to merge them together) maybe we can reduce it to a one liner to avoid a check of the number of response elements in the multi response array:

```ruby
multi_response.responses.map { |r| r.params }.reduce({}, :merge)
```

lib/active_merchant/billing/gateways/creditcall.rb
nfarve: Yes the failed purchase.

nfarve:
@davidsantoso I think your right we technically don't need to response from the capture. We can just return the first response at all times since that has all of the information they actually want.

lib/active_merchant/billing/gateways/creditcall.rb
curiousepic: 👍

davidsantoso:
@nfarve thinking about it a little more, I wonder if it'll be tough to know where the authorize response params will be- the multi response could have 1 element or two elements depending on whether or not the initialize `Authorize` request was successful.

Makes me start to lean towards merging the params together since the only values that are really changing seem to be the timestamps and those will be right next to each other since we don't delay between those two during a `purchase` call.

nfarve:
@davidsantoso so it looks like localresult is just the result code (0 for success 1 for failure) so merging would never overwrite a failure code. I can't find authorizationentity in their docs to check what exactly that is and if there is a problem if we overwrite. 

But it sounds like if we are fine updating the timestamps there isn't any real harm in overwriting params, per @curiousepic concern.

curiousepic:
Is there any threat of getting the success state, if the auth is successful but the capture fails for some reason, and we're merging params?

davidsantoso:
>  if the auth is successful but the capture fails for some reason, and we're merging params?

I was about to say- I think we still want to use the second (`capture`) transaction to determine the success since it's for a `purchase`. I think if we merge and the `LocalResult` is `1` for failed capture, then we'd get that value from the merge. Seems like overwriting the timestamps is ok too since we're more concerned about when the capture happened over the auth.

Overall, seems like merge is ok to me...

PR #2545, opened 2017-08-14
AVS on TransFirst Express are failing when no address2

nfarve:
Removed the address2 field unless a value is given.

Loaded suite
test/remote/gateways/remote_trans_first_transaction_express_test
Started

Finished in 23.362846 seconds.
30 tests, 105 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
1.28 tests/s, 4.49 assertions/s

Loaded suite test/unit/gateways/trans_first_transaction_express_test
Started

Finished in 0.068235 seconds.
21 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
307.76 tests/s, 1509.49 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
+++ b/lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
@@ -542,7 +542,7 @@ module ActiveMerchant #:nodoc:
               doc["v1"].nr billing_address[:phone].gsub(/\D/, '') if billing_address[:phone]
             end
             doc["v1"].addrLn1 billing_address[:address1]
-            doc["v1"].addrLn2 billing_address[:address2]
+            doc["v1"].addrLn2 billing_address[:address2] if billing_address[:address2]
             doc["v1"].city billing_address[:city]
             doc["v1"].state billing_address[:state]
             doc["v1"].zipCode billing_address[:zip]
--- a/test/remote/gateways/remote_trans_first_transaction_express_test.rb
+++ b/test/remote/gateways/remote_trans_first_transaction_express_test.rb
@@ -45,6 +45,20 @@ class RemoteTransFirstTransactionExpressTest < Test::Unit::TestCase
     assert_equal "Street address does not match, but 5-digit postal code matches.", response.avs_result["message"]
     assert_equal "CVV matches", response.cvv_result["message"]
   end
+ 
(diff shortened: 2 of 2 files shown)


nfarve:
@curiousepic @davidsantoso @dtykocki @deedeelavinder

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -542,7 +542,7 @@ def add_contact(doc, fullname, options)
               doc["v1"].nr billing_address[:phone].gsub(/\D/, '') if billing_address[:phone]
             end
             doc["v1"].addrLn1 billing_address[:address1]
-            doc["v1"].addrLn2 billing_address[:address2]
+            doc["v1"].addrLn2 billing_address[:address2] unless billing_address[:address2].blank?
```
davidsantoso: @nfarve I know there are a couple places in Active Merchant where we currently use the `.blank?` method, but we're trying to slowly remove those/not continue using it since it's actually an [Active Support](https://github.com/rails/rails/tree/master/activesupport) extension and we'd like to remove our dependency on Active Support. It's current usage is a holdover from when Active Merchant was more tied to being used in a Rails application.

That being said, it's not well documented that we're no longer using that convention and instead going with something like:

`doc["v1"].addrLn2 billing_address[:address2] if billing_address[:address2]`

Apologies on the poor documentation! We're not terribly clear about this in general and it's (unfortunately) become more tribal knowledge. Likewise, looking at the remote test we should probably change the `options` hash to instead have:

```ruby
options[:shipping_address][:address2] = nil
options[:billing_address][:address2] = nil
```

Since in general someone will probably not add those keys to a hash to begin with so accessing that key in the hash will just return a `nil`.

nfarve: Gotcha!

nfarve:
@davidsantoso Think I captured your suggestions, not sure why this last check wont start though.

davidsantoso:
@nfarve hrrrm. I just restarted the build so hopefully it's just a transient issue. Based on the change I imagine it'll be fine, but the restart will hopefully be fairly quick!

PR #2546, opened 2017-08-15
Orbital: Upgrade schema version to 7.1

curiousepic:
Also dynamically converts API_VERSION constant to the Content-Type
header which actually specifies the version.

This is in support of sending 3DS data which 5.6 couldn't seem to validate properly.

Unit:
68 tests, 412 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
18 tests, 121 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -30,11 +30,11 @@ module Billing #:nodoc:
     class OrbitalGateway < Gateway
       include Empty
 
-      API_VERSION = "5.6"
+      API_VERSION = "7.1"
 
       POST_HEADERS = {
         "MIME-Version" => "1.1",
-        "Content-Type" => "application/PTI56",
+        "Content-Type" => "application/PTI#{API_VERSION.gsub(/\./, '')}",
```
davidsantoso: Just out of curiosity- do they list their API version as `7.1` but are expecting it sent as `PTI71`? Just wondering if we should set the `API_VERSION = "PTI71"` instead.

curiousepic: Hmm, interesting thought. But they do not refer to the schema version as `PTI71`, that's just the format it's expected in for the schema validator to pick up. This way seems marginally readable, though I suppose it's possible the gsub could break somehow...

davidsantoso: > But they do not refer to the schema version as PTI71... This way seems marginally more readable

I figured as much, just thought I'd double check! I do think using the `7.1` as the `API_VERSION` is clearer for someone grok-ing the adapter so I'm ok with this. 

> though I suppose it's possible the gsub could break somehow...

Considering it's gsub-ing on a constant (well, as constant as constants get it Ruby...) I think the changes of it breaking are pretty close to zero!

PR #2549, opened 2017-08-16
CreditCall: Only allow AVS when specified

curiousepic:
CreditCall: Only allow AVS when specified
    
    This disables AVS checking by default, and only sends the
    AdditionalVerification element if either verify_zip or verify_address
    flags are sent in options. It also exposes `manual_type` to an option
    field. These options can help with decline rates.
    
    Remote:
    100% passed
    
    Unit:
    100% passed

Tagging @nfarve too.

### DIFF
--- a/lib/active_merchant/billing/gateways/creditcall.rb
+++ b/lib/active_merchant/billing/gateways/creditcall.rb
@@ -126,17 +126,22 @@ module ActiveMerchant #:nodoc:
 
       def add_card_details(xml, payment_method, options={})
         xml.CardDetails do
-          xml.Manual(type: "ecommerce") do
+          xml.Manual(type: manual_type(options)) do
             xml.PAN payment_method.number
             xml.ExpiryDate exp_date(payment_method)
             xml.CSC payment_method.verification_value unless empty?(payment_method.verification_value)
           end
 
-          if address = options[:billing_address]
-            xml.AdditionalVerification do
-              xml.Address address[:address1]
-              xml.Zip address[:zip]
-            end
+          add_additional_verification(xml, options)
+        end
+      end
+
+      def add_additional_verification(xml, options)
+        return unless (options[:verify_zip].to_s == 'true') || (options[:verify_address].to_s == 'true')
+        if address = options[:billing_address]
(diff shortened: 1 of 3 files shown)


curiousepic:
I'm rethinking this implementation after looking at their docs more closely, update coming.

curiousepic:
@dtykocki @davidsantoso Updated with revised version. This disables AVS checking by default. This shouldn't disrupt anyone already using the gateway, since CreditCall does not factor AVS results into whether to fail a transaction, and AVS results have not been available in the response until today (https://github.com/activemerchant/active_merchant/commit/97c4a125705ca119a94e014fac558add98dba377). This simplifies the options and makes them more explicit.

lib/active_merchant/billing/gateways/creditcall.rb
```diff
@@ -126,17 +126,22 @@ def add_terminal_details(xml, options={})
...
-            xml.AdditionalVerification do
-              xml.Address address[:address1]
-              xml.Zip address[:zip]
-            end
+          add_additional_verification(xml, options)
+        end
+      end
+
+      def add_additional_verification(xml, options)
+        return unless (options[:verify_zip] == 'true') || (options[:verify_address] == 'true')
```
davidsantoso: any particular reason to check for `'true'` string vs boolean `true`?

curiousepic: There seemed to be an equal number of both in use for other gateways. I actually wasn't sure if Spreedly sends any actual boolean fields to send to AM. Happy to change if we do, it would be cleaner.

davidsantoso: I guess the more I think about it, with Ruby we could just do:

```ruby
return unless (options[:verify_zip]) || (options[:verify_address])
```

and it would work since the conditional is only checking absence of `nil` or `false`, in which case if it was `true` or `"true"` would both work.

curiousepic: unless it was sent as `"false"`, which would still add the element and activate AVS.

PR #2550, opened 2017-08-16
TransFirst Express: Fix Optional Fields Being Passed Blank

nfarve:
Continuation of issue where address2 field being passed as an empty
string caused an error. Currently, only address1 and zip are required
for AVS. Previous fix to only pass address2 if is present doesn't fix
the larger issue that none of the optional fields should be passed
blank.

Updated previous remote test to not only remove address2 but everything in
shipping and billing that is not required.

Loaded suite
test/remote/gateways/remote_trans_first_transaction_express_test
Started

Finished in 25.935593 seconds.
30 tests, 105 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
1.16 tests/s, 4.05 assertions/s

Loaded suite test/unit/gateways/trans_first_transaction_express_test
Started

Finished in 0.104014 seconds.
21 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
201.90 tests/s, 990.25 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
+++ b/lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
@@ -537,15 +537,17 @@ module ActiveMerchant #:nodoc:
           doc["v1"].title options[:title] if options[:title]
 
           if (billing_address = options[:billing_address])
-            doc["v1"].phone do
-              doc["v1"].type (options[:phone_number_type] || "4")
-              doc["v1"].nr billing_address[:phone].gsub(/\D/, '') if billing_address[:phone]
+            if billing_address[:phone]
+              doc["v1"].phone do
+                doc["v1"].type (options[:phone_number_type] || "4")
+                doc["v1"].nr billing_address[:phone].gsub(/\D/, '')
+              end
             end
-            doc["v1"].addrLn1 billing_address[:address1]
+            doc["v1"].addrLn1 billing_address[:address1] if billing_address[:address1]
             doc["v1"].addrLn2 billing_address[:address2] if billing_address[:address2]
-            doc["v1"].city billing_address[:city]
-            doc["v1"].state billing_address[:state]
-            doc["v1"].zipCode billing_address[:zip]
+            doc["v1"].city billing_address[:city] if billing_address[:city]
+            doc["v1"].state billing_address[:state] if billing_address[:state]
+            doc["v1"].zipCode billing_address[:zip] if billing_address[:zip]
             doc["v1"].ctry "US"
           end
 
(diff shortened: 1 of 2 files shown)


nfarve:
@deedeelavinder @dtykocki @davidsantoso @curiousepic

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -537,14 +537,16 @@ def add_contact(doc, fullname, options)
...
+                doc["v1"].nr billing_address[:phone].gsub(/\D/, '')
+              end
             end
             doc["v1"].addrLn1 billing_address[:address1]
             doc["v1"].addrLn2 billing_address[:address2] if billing_address[:address2]
-            doc["v1"].city billing_address[:city]
-            doc["v1"].state billing_address[:state]
+            doc["v1"].city billing_address[:city] if billing_address[:city]
+            doc["v1"].state billing_address[:state] if billing_address[:state]
             doc["v1"].zipCode billing_address[:zip]
```
davidsantoso: Is zip expected to always be sent? Wondering if we should follow this same pattern?

lib/active_merchant/billing/gateways/trans_first_transaction_express.rb
```diff
@@ -537,14 +537,16 @@ def add_contact(doc, fullname, options)
...
-            doc["v1"].phone do
-              doc["v1"].type (options[:phone_number_type] || "4")
-              doc["v1"].nr billing_address[:phone].gsub(/\D/, '') if billing_address[:phone]
+            if billing_address[:phone]
+              doc["v1"].phone do
+                doc["v1"].type (options[:phone_number_type] || "4")
+                doc["v1"].nr billing_address[:phone].gsub(/\D/, '')
+              end
             end
             doc["v1"].addrLn1 billing_address[:address1]
```
davidsantoso: Is address1 expected to always be sent? Wondering if we should follow this same pattern?

nfarve: Note: Address and Zip Code are required to participate in the address verification service (AVS) and highly recommended for appropriate card not present qualification.

From reading their docs its conditional for just a regular credit, with the stipulation above. Not sure if if best to also pass it only when given so that non-AVS calls can be made.

davidsantoso: Gotcha, so seems like you should really send the `address1` and `zip` fields (even if they're empty) whereas the other address fields shouldn't be added to the XML if they are empty.

nfarve: I just checked passing the address1 and zip as an empty string and get the same error as I would get with the other params.

davidsantoso: Hmmm. Well that makes me think we should have guard clauses on them as well 😸

nfarve:
@davidsantoso cool! Went ahead and made sure nothing in the shipping or billing address is passed empty and added a test for a purchase not using AVS.

davidsantoso:
@nfarve LGTM! :shipit:

PR #2552, opened 2017-08-17
CreditCall: Parsing additional params from responses

nfarve:
Currently we are not parsing the CardDetails from the response message,
which is passed on auth and refund transactions.

This information is now being passed in the response object as well.

Loaded suite test/remote/gateways/remote_creditcall_test
Started
Finished in 45.137631 seconds.
18 tests, 46 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
0.40 tests/s, 1.02 assertions/s

Loaded suite test/unit/gateways/creditcall_test
Started
Finished in 0.099526 seconds.
16 tests, 50 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
160.76 tests/s, 502.38 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/creditcall.rb
+++ b/lib/active_merchant/billing/gateways/creditcall.rb
@@ -14,6 +14,32 @@ module ActiveMerchant #:nodoc:
 
       self.homepage_url = 'https://www.creditcall.com'
       self.display_name = 'Creditcall'
+      
+      CVV_CODE = {
+        "matched" => "M",
+        "notmatched" => "N",
+        "notchecked" => "P",
+        "partialmatch" => "N"
+      }
+      
+      AVS_CODE = {
+        "matched;matched" => "D",
+        "matched;notchecked" =>"B",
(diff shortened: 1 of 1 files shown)


nfarve:
@curiousepic @davidsantoso @deedeelavinder @dtykocki

davidsantoso:
@nfarve this is great, thanks! Only minor tweak I would do is probably great separate methods for getting the AVS, CVV, and error code into their own methods. For example with AVS:

```ruby
Response.new(
  ...
  avs_result: AVSResult.new()
  ...
)

def avs_result_code_from(merged_params)
  AVS_CODE["#{merged_params['Address']};#{merged_params['Zip']}"]
end
```

Since this is happening in two different places, it's nice to have an method to handle doing the work should that ever need to change. Then we can change it in just one spot.

nfarve:
@davidsantoso good point! Updated.

PR #2553, opened 2017-08-18
Orbital: Support Network Tokenization Credit Cards

curiousepic:
Remote:
22 tests, 137 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
69 tests, 417 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@nfarve

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -191,7 +191,7 @@ module ActiveMerchant #:nodoc:
 
       # A – Authorization request
       def authorize(money, creditcard, options = {})
-        order = build_new_order_xml(AUTH_ONLY, money, options) do |xml|
+        order = build_new_order_xml(AUTH_ONLY, money, creditcard, options) do |xml|
           add_creditcard(xml, creditcard, options[:currency])
           add_address(xml, creditcard, options)
           if @options[:customer_profiles]
@@ -211,7 +211,7 @@ module ActiveMerchant #:nodoc:
 
       # AC – Authorization and Capture
       def purchase(money, creditcard, options = {})
-        order = build_new_order_xml(AUTH_AND_CAPTURE, money, options) do |xml|
+        order = build_new_order_xml(AUTH_AND_CAPTURE, money, creditcard, options) do |xml|
           add_creditcard(xml, creditcard, options[:currency])
           add_address(xml, creditcard, options)
           if @options[:customer_profiles]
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -463,6 +463,15 @@ def add_creditcard(xml, creditcard, currency=nil)
         end
       end
 
+      def add_cdpt_authentication_value(xml, creditcard)
+        xml.tag! :AuthenticationECIInd, nil
```
curiousepic: The presence of this field is required, but a value is not.

davidsantoso: @curiousepic if a value is sent, will it cause transactions to fail? Seems like we should send something along if we do have one.

jasonwebster: The in the case of Apple Pay, the Payment Token's payment data contains an ECI value that must be passed along for certain network/issuer combinations: https://developer.apple.com/library/content/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.html

It's not always required, but in some cases it is in order for the transaction to process properly. I think it would be best to pass along our standard `:eci` option from the options hash here.

davidsantoso: > the docs say "the ECI Indicator field can be left empty and the appropriate value is derived by the gateway."

Interesting. Gateways seem all over the place when it comes to accepting `eci`. I'd still vote for including it since it's actually a standard field on a `NetworkTokenizationCreditCard` which still leaves the option for the value to be `nil` depending on how the `NetworkTokenizationCreditCard` was initialized.

jasonwebster: Ahh yeah, sorry, the `NetworkTokenizationCreditCard` attribute is what I meant, not the option.

curiousepic:
@davidsantoso @jasonwebster Updated - also now passing XID, which similar to ECI doesn't seem necessary for most transactions but could be needed for specific ones.

davidsantoso:
@curiousepic just to double check- is this a gateway where the ordering of the fields matter?

curiousepic:
@davidsantoso It is.

davidsantoso:
@curiousepic yep figured as much since you had split them out!

PR #1965, opened 2015-12-11
Mercury e2e

theinventor:
Hey everyone!

We added Mercury end to end encryption support a while back and wanted to see if you wanted this upstream.

This means you can support people with an encrypted card reader using e2e encryption with Mercury using their SOAP API.

We have this in production with them and it's working well. 

Happy to try to answer questions about this..

(tell me if I'm doing this wrong.)

PR #2557, opened 2017-08-23
Mercury (MercuryPay and E2E, aka Vantiv)

stevenjonescgm:
We needed a recently added Gateway as well as Mercury end-to-end encryption from https://github.com/webstream-io/active_merchant/tree/Mercury-e2e and #1965 so I cut this new fork to test the merge/updates. While checking the tests I found issues with the MercuryGateway suite, so I updated those as well.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
 * Checkout V2: Add UAE to country list [shasum] #2548
 * Orbital: Updgrade schema version to 7.1 [curiousepic] #2546
 * CreditCall: Parse more response params [nfavre] #2543
+* CreditCall: Only allow AVS when specified [curiousepic] #2549
 
 == Version 1.70.0 (August 4, 2017)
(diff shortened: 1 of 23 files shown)


stevenjonescgm:
I added on top of this pr (because it is less effort than cutting independent prs) some fixes for tests for other Gateways that concern me.

Stripe only needed modern (dynamic) expiration on Track1
AuthrizeNetCim has a "A duplicate record with ID 1810384386 already exists." error. this was being obfuscated by assuming the response was valid (bad assert pattern?) and so surfaced as an Error instead of a Failure. I have not determined the root or solution to this failure, but this at least makes it easier to identify for the next person.

stevenjonescgm:
CI failure seems to be due to instability in moneris_us_test, which should be independent from the above changes, and passes locally

bpollack:
I'm closing this for now, but please feel free to reopen if you've got some cycles and can fix the failing tests.

PR #2559, opened 2017-08-23
Ebank: Adds Brazil Specific Parameters

nfarve:
Birth_date and payment.responsible parameters are required for customers in
Brazil. Adding these fields to the gateway.

Loaded suite test/remote/gateways/remote_ebanx_test

16 tests, 44 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Loaded suite test/unit/gateways/ebanx_test

14 tests, 46 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -48,7 +48,7 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, payment, options)
         add_payment(post, payment)
         add_address(post, options)
-
+        add_customer_responsible_person(post, payment, options) if post[:payment][:country] == 'BR'
         commit(:purchase, post)
       end
 
@@ -60,6 +60,7 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, payment, options)
         add_payment(post, payment)
         add_address(post, options)
+        add_customer_responsible_person(post, payment, options) if post[:payment][:country] == 'BR'
         post[:payment][:creditcard][:auto_capture] = false
 
         commit(:authorize, post)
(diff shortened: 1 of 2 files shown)


nfarve:
@curiousepic @davidsantoso @dtykocki @deedeelavinder

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -41,25 +41,27 @@ def initialize(options={})
       end
 
       def purchase(money, payment, options={})
-        post = { payment: {} }
+        post = { payment: { responsible: {}} }
```
curiousepic: Let's move the addition of this sub-element to the `add_customer_responsible_person` method.

curiousepic:
@nfarve Is the `responsible` sub-element required for all transactions or only for Brazil? This  implementation would always send this element since the name comes from the payment method. If there is some way to condition on the country, it might be best to restrict it to Brazil.

nfarve:
@curiousepic its only required for Brazil and businesses. Since we aren't supporting the business fields, I added a conditional to only add it for Brazil.

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -48,7 +48,9 @@ def purchase(money, payment, options={})
         add_customer_data(post, payment, options)
         add_payment(post, payment)
         add_address(post, options)
-
+        if post[:payment][:country] == 'BR'
+          add_customer_responsible_person(post, payment, options)
```
curiousepic: I think `add_customer_responsible_person(post, payment, options) if post[:payment][:country] == 'BR'` would be slightly preferred despite being a long line.

curiousepic:
@nfarve Looks good other than the formatting mention.

nfarve:
@curiousepic cool updated!

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -129,6 +130,14 @@ def add_customer_data(post, payment, options)
         post[:payment][:name] = payment.name
         post[:payment][:email] = options[:email] || "unspecified@example.com"
         post[:payment][:document] = options[:document]
+        post[:payment][:birth_date] = options[:birth_date] if options[:birth_date]
+      end
+
+      def add_customer_responsible_person(post, payment,  options)
+        post[:payment][:responsible] = {}
```
curiousepic: Sorry one more thing, I think the creation of this hash is implicit in the line below, so this isn't necessary. This pattern of first instantiating an empty hash is more often seen when you return the hash itself at the end. But here we're adding further structure directly to the base `post` hash that's already available.

nfarve: When I remove it I get errors for an undefined method '[]=' for nil:NilClass

curiousepic: Hah ...or I could just be completely wrong! I guess it doesn't like adding new elements two-deep. Merging!

PR #2562, opened 2017-08-25
PayU Latam: Pass payment_country option

curiousepic:
payment_country must be the country of the merchant, rather than the
buyer/payer. This opens an option field for the merchant to pass this
value.

Two tests unrelated to this change still fail as "Internal payment
provider error. "

Remote:
15 tests, 39 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
86.6667% passed

Unit:
17 tests, 71 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@nfarve

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -249,7 +249,7 @@ module ActiveMerchant #:nodoc:
       def add_payer(post, options)
         if address = options[:billing_address]
           payer = {}
-          post[:transaction][:paymentCountry] = address[:country]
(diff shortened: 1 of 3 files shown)


curiousepic:
I'm actually reconsidering whether this should should in fact be a Gateway attribute. Originally I thought the value of this field should be specify-able per transaction, but since each PayU Latam merchant account is tied to a country, it will be the same for every transaction through that gateway.

nfarve:
You can pull the country from the merchant's account?

nfarve:
Gotcha. So when they set up the gateway they would add their country and we would use that every time they process a transaction with that gateway?

PR #2564, opened 2017-08-28
Braintree: Send phone in options hash

deedeelavinder:
Adds phone parameter to customer details, adds one specific test, and
updates other relevant unit and remote tests.

Also updates failing remote test
`test_successful_authorize_with_merchant_account_id`. Instead of
comparing sent versus received `merchant_account_id`, it now asserts
the correct braintree status.

Unit Tests:
48 tests, 121 assertions, 0 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications
100% passed

Remote Tests:
59 tests, 346 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_braintree_blue_test.rb
+++ b/test/remote/gateways/remote_braintree_blue_test.rb
@@ -265,6 +265,20 @@ class RemoteBraintreeBlueTest < Test::Unit::TestCase
     assert_equal 'customer@example.com', transaction["customer_details"]["email"]
   end
 
+  def test_successful_purchase_with_phone
+    assert response = @gateway.purchase(@amount, @credit_card, :phone => "123-345-5678")
+    assert_success response
+    transaction = response.params["braintree_transaction"]
+    assert_equal '123-345-5678', transaction["customer_details"]["phone"]
+  end
+
+  def test_successful_purchase_with_phone_from_address
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    transaction = response.params["braintree_transaction"]
+    assert_equal '(555)555-5555', transaction["customer_details"]["phone"]
+  end
+
   def test_purchase_with_store_using_random_customer_id
     assert response = @gateway.purchase(
       @amount, credit_card('5105105105105100'), @options.merge(:store => true)
@@ -294,7 +308,8 @@ class RemoteBraintreeBlueTest < Test::Unit::TestCase
         :first_name => 'Old First', :last_name => 'Old Last',
         :month => 9, :year => 2012
       ),
-      :email => "old@example.com"
+      :email => "old@example.com",
+      :phone => "321-654-0987"
     )
     payment_method_token = response.params["braintree_customer"]["credit_cards"][0]["token"]
     assert response = @gateway.purchase(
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_braintree_blue_test.rb
```diff
@@ -294,7 +303,9 @@ def test_purchase_using_specified_payment_method_token
         :first_name => 'Old First', :last_name => 'Old Last',
         :month => 9, :year => 2012
       ),
-      :email => "old@example.com"
+      :email => "old@example.com",
+      :phone => "321-654-0987"
+
```
curiousepic: Blank line here

curiousepic:
So this looks good, though the commit message is incorrect, since this allows the field to be passed in as an option, not pulling it from the payment method (PMs don't have `phone` attributes).

curiousepic:
@jasonwebster That's true, and I'm actually looking to do something similar for another gateway. More gateways seem to be requiring a phone number usually associated with a PM, and there are some situations where having the option available is more convenient than relying on it being passed in the billing address. So, we could use `options[:phone] || options[:billing_address][:phone]` as a fallback.

jasonwebster:
Nah, I think the best thing would be to document an aspirational standard set of options for `#store` and `#update`. Carrying over ones from the purchase API may not make sense.

curiousepic:
For the record, Payeezy does currently look for and pass `options[:phone]` in its `details` element, which is required, for all actions, not just store. We're about to change that to my aforementioned `options[:phone] || options[:billing_address][:phone]` so as not to break anyone already using it that way.

I also see it available for CardStream, HPS, PayHub, and Psigate.

jasonwebster:
Great! Then let's add it to what little documentation there is around standardized options: https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards#options

curiousepic:
@jasonwebster Funny thing, I've never seen that page until now 😅  I'm all about it though; I'll update it with a note about this sort of fledgling standard. Let me know if you have any further suggestions about standardizing this field, since it's definitely a minority among the others. 

As a note bolstering the case for this change, a handful of gateways also look inside `billing_address` for `email`, so `phone` and `email` both seem like fields that have use cases where it's desirable to specify independently of billing address.

test/remote/gateways/remote_braintree_blue_test.rb
```diff
@@ -527,10 +544,12 @@ def test_successful_update
         :first_name => 'New First', :last_name => 'New Last',
         :month => 10, :year => 2014
       ),
-      :email => "new@example.com"
+      :email => "new@example.com",
+      :phone => "321-654-0987"
```
curiousepic: Minor point but this is the same phone as the old one, so it's not technically testing Update.

deedeelavinder: @curiousepic Are you thinking of the new test with phone from billing? That is using the other phone number. Line 279.

curiousepic: Nope - this test is updating the info stored with a card. It is first `store`d with this value for `phone` https://github.com/activemerchant/active_merchant/pull/2564/files/3d128699edc6afd49f7961b32baf638a6e6e2a5d#diff-d3e31ce1d7cbbc102d604bd1b6642dc2R526 And then in the `update`, the same value is sent for `phone` here. So it's not testing that it is truly updated because there's no change in the value.

deedeelavinder: 👍 Of course!  Thanks, @curiousepic -- updating now.

PR #2567, opened 2017-09-01
WePay: Don't default API version header

curiousepic:
WePay has restricted compatibility for the version specified in the
request header, to equal to or one ahead of the version set in WePay's
Dashboard for that merchant account.

This means that letting the adapter default the version header to a
certain version (likely one or more version behind current) has greater
potential to break integration than allowing WePay's API to assume the
version set in the dashboard for that account, which occurs if the
header is absent.

api_version can still be sent in options and will override the dashboard
setting, but be advised that this will only work for one version ahead
of what is set in the dashboard.

I have updated the version in the dashboard for the test account in
Fixtures to the current version, so these tests are run against that
version.

Remote:
22 tests, 37 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
100% passed

@nfarve

### DIFF
--- a/lib/active_merchant/billing/gateways/wepay.rb
+++ b/lib/active_merchant/billing/gateways/wepay.rb
@@ -10,8 +10,6 @@ module ActiveMerchant #:nodoc:
       self.default_currency = 'USD'
       self.display_name = 'WePay'
 
-      API_VERSION = "2017-02-01"
-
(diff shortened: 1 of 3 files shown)


curiousepic:
@dtykocki 👍  Definitely a good idea, updated.

PR #2568, opened 2017-09-01
Conekta: Update Required Fields

nfarve:
The key previously used was the default key for testing and not actually
tied to an account. Conekta requires that all fields in the details
parameter be passed for a successful purchase.

Loaded suite test/unit/gateways/conekta_test

11 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Loaded suite test/remote/gateways/remote_conekta_test

13 tests, 51 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/conekta.rb
+++ b/lib/active_merchant/billing/gateways/conekta.rb
@@ -82,15 +82,14 @@ module ActiveMerchant #:nodoc:
 
       def add_details_data(post, options)
         details = {}
-        details[:name] = options[:customer] if options[:customer]
+        details[:name] = options[:customer] || options[:billing_address][:name]
         details[:email] = options[:email] if options[:email]
-        details[:phone] = options[:phone] if options[:phone]
+        details[:phone] = options[:phone] || options[:billing_address][:phone]
         post[:device_fingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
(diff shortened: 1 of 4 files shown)


nfarve:
@curiousepic @deedeelavinder @davidsantoso @dtykocki

lib/active_merchant/billing/gateways/conekta.rb
```diff
@@ -82,9 +82,9 @@ def add_order(post, money, options)
 
       def add_details_data(post, options)
         details = {}
-        details[:name] = options[:customer] if options[:customer]
+        details[:name] = options[:customer] || options[:billing_address][:name]
         details[:email] = options[:email] if options[:email]
-        details[:phone] = options[:phone] if options[:phone]
+        details[:phone] = options[:phone] || options[:billing_address][:phone] || options[:shipping_address][:phone]
```
curiousepic: Let's actually not look in the shipping address for this, since it could be different from what it should be, even if it isn't provided, in which case it does need to be added to the options/billing address.

nfarve: @curiousepic just pushed updated that only checks billing_address.

PR #2570, opened 2017-09-04
PayU Latam: Pass unique buyer fields and country requirements

curiousepic:
Buyer data can be separate from the Payer's (tied to the payment
method). This change allows options to be passed for a unique buyer, and
adding the buyer element no longer conditions only on the presence of a
shipping address. Also adds fields that were missing but are required
for Brazil, Colombia, and Mexico, and allows passing paymentCountry as
an option.

Unrelated Capture and Void remote tests still failing

Remote:
17 tests, 45 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
88.2353% passed

Unit:
21 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

@nfarve

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -122,7 +122,7 @@ module ActiveMerchant #:nodoc:
         add_invoice(post, amount, options)
         add_signature(post)
         add_payment_method(post, payment_method, options)
-        add_payer(post, options)
+        add_payer(post, payment_method, options)
         add_extra_parameters(post, options)
       end
 
@@ -138,6 +138,7 @@ module ActiveMerchant #:nodoc:
 
       def add_transaction_elements(post, type, options)
         transaction = {}
+        transaction[:paymentCountry] = options[:payment_country] || (options[:billing_address][:country] if options[:billing_address])
         transaction[:type] = type
         transaction[:ipAddress] = options[:ip] if options[:ip]
         transaction[:userAgent] = options[:user_agent] if options[:user_agent]
@@ -152,28 +153,59 @@ module ActiveMerchant #:nodoc:
         order[:referenceCode] = options[:order_id] || generate_unique_id
         order[:description] = options[:description] || 'unspecified'
         order[:language] = 'en'
+        order[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
         post[:transaction][:order] = order
       end
 
+      def add_payer(post, payment_method, options)
+        address = options[:billing_address]
+        payer = {}
+        payer[:fullName] = payment_method.name.strip
+        payer[:contactPhone] = address[:phone] if (address && address[:phone])
+        payer[:dniNumber] = options[:dni_number] if options[:dni_number]
+        payer[:dniType] = options[:dni_type] if options[:dni_type]
+        payer[:emailAddress] = options[:email] if options[:email]
+        payer[:birthdate] = options[:birth_date] if options[:birth_date] && options[:payment_country] == 'MX'
+        payer[:billingAddress] = billing_address_fields(options)
+        post[:transaction][:payer] = payer
+      end
+
+      def billing_address_fields(options)
+        return unless address = options[:billing_address]
+        billing_address = {}
+        billing_address[:street1] = address[:address1]
+        billing_address[:street2] = address[:address2]
+        billing_address[:city] = address[:city]
+        billing_address[:state] = address[:state]
+        billing_address[:country] = address[:country]
+        billing_address[:postalCode] = address[:zip] if options[:payment_country] == 'MX'
+        billing_address[:phone] = address[:phone]
+        billing_address
+      end
+
       def add_buyer(post, options)
-        if address = options[:shipping_address]
-          buyer = {}
-          buyer[:fullName] = address[:name]
-          buyer[:dniNumber] = options[:dni_number] if options[:dni_number]
-          buyer[:dniType] = options[:dni_type] if options[:dni_type]
-          buyer[:emailAddress] = options[:email] if options[:email]
-          buyer[:contactPhone] = address[:phone]
-          shipping_address = {}
-          shipping_address[:street1] = address[:address1]
-          shipping_address[:street2] = address[:address2]
-          shipping_address[:city] = address[:city]
-          shipping_address[:state] = address[:state]
-          shipping_address[:country] = address[:country]
-          shipping_address[:postalCode] = address[:zip]
-          shipping_address[:phone] = address[:phone]
-          buyer[:shippingAddress] = shipping_address
-          post[:transaction][:order][:buyer] = buyer
-        end
+        buyer = {}
+        buyer[:fullName] = options[:buyer_name] if options[:buyer_name]
+        buyer[:dniNumber] = options[:buyer_dni_number] if options[:buyer_dni_number]
+        buyer[:dniType] = options[:buyer_dni_type] if options[:buyer_dni_type]
+        buyer[:cnpj] = options[:buyer_cnpj] if options[:buyer_cnpj] && options[:payment_country] == 'BR'
+        buyer[:emailAddress] = options[:buyer_email] if options[:buyer_email]
+        buyer[:contactPhone] = options[:shipping_address][:phone] if options[:shipping_address]
+        buyer[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
+        post[:transaction][:order][:buyer] = buyer
+      end
+
+      def shipping_address_fields(options)
+        return unless address = options[:shipping_address]
+        shipping_address = {}
+        shipping_address[:street1] = address[:address1]
+        shipping_address[:street2] = address[:address2]
+        shipping_address[:city] = address[:city]
+        shipping_address[:state] = address[:state]
+        shipping_address[:country] = address[:country]
+        shipping_address[:postalCode] = address[:zip]
+        shipping_address[:phone] = address[:phone]
+        shipping_address
       end
 
       def add_invoice(post, money, options)
@@ -191,8 +223,8 @@ module ActiveMerchant #:nodoc:
 
         additional_values = {}
         additional_values[:TX_VALUE] = tx_value
-        additional_values[:TX_TAX] = tx_tax
-        additional_values[:TX_TAX_RETURN_BASE] = tx_tax_return_base
+        additional_values[:TX_TAX] = tx_tax if options[:payment_country] == 'CO'
+        additional_values[:TX_TAX_RETURN_BASE] = tx_tax_return_base if options[:payment_country] == 'CO'
 
         post[:transaction][:order][:additionalValues] = additional_values
       end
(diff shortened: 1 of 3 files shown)


curiousepic:
I'm open to other ways of handling this.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -152,28 +153,60 @@ def add_order(post, options)
...
+
+      def add_shipping_address(post, options)
+        address = options[:shipping_address]
+        shipping_address = {}
+        shipping_address[:street1] = address[:address1]
+        shipping_address[:street2] = address[:address2]
+        shipping_address[:city] = address[:city]
+        shipping_address[:state] = address[:state]
+        shipping_address[:country] = address[:country]
+        shipping_address[:postalCode] = address[:zip]
```
deedeelavinder: Is the `if options[:payment_country] == 'MX'` qualifier for `:postalCode` only needed for the billing address and not the shipping address?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -138,6 +138,7 @@ def add_credentials(post, command)
 
       def add_transaction_elements(post, type, options)
         transaction = {}
+        transaction[:paymentCountry] = options[:country] || options[:billing_address][:country] if (options[:billing_address] && options[:billing_address][:country])
```
deedeelavinder: This is the only place using `options[:country]`. Elsewhere `options[:payment_country]` is used (lines 169, 182, 192, 227, and 228).

curiousepic: @deedeelavinder Good catch, went back and forth on this a few times and overlooked the final field name. Thanks!

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -152,28 +153,60 @@ def add_order(post, options)
...
 
+      def add_payer(post, payment_method, options)
+        address = options[:billing_address]
+        payer = {}
+        payer[:fullName] = payment_method.name.strip
+        payer[:contactPhone] = address[:phone]
+        payer[:dniNumber] = options[:dni_number] if options[:dni_number]
+        payer[:dniType] = options[:dni_type] if options[:dni_type]
+        payer[:emailAddress] = options[:email] if options[:email]
+        payer[:contactPhone] = address[:phone]
```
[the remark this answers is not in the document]
curiousepic: 👍

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -152,28 +153,60 @@ def add_order(post, options)
...
+        payer[:birthdate] = options[:birth_date] if options[:birth_date] && options[:payment_country] == 'MX'
+        payer[:billingAddress] = add_billing_address(post, options)
+        post[:transaction][:payer] = payer
+      end
+
+      def add_billing_address(post, options)
```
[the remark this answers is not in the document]
curiousepic: Nope 👍 The method probably be named something other than "add_billing_address" too, since it's only returning it.

lib/active_merchant/billing/gateways/payu_latam.rb
curiousepic: Yep. The shipping address could be somewhere else, so I think it's not validated.

lib/active_merchant/billing/gateways/payu_latam.rb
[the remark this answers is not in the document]
deedeelavinder: `format_billing_address`?

curiousepic:
Updated, addressed the points @deedeelavinder brought up.

deedeelavinder:
👍 :shipit:

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -138,6 +138,7 @@ def add_credentials(post, command)
 
       def add_transaction_elements(post, type, options)
         transaction = {}
+        transaction[:paymentCountry] = options[:payment_country] || options[:billing_address][:country] if (options[:billing_address] && options[:billing_address][:country])
```
davidsantoso: I don't _believe_ the additional ` && options[:billing_address][:country]` is necessary on this guard clause since `options[:billing_address][:country]` will evaluate to `nil` as long as the `:billing_address` key exists.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -152,28 +153,59 @@ def add_order(post, options)
...
 
+      def add_payer(post, payment_method, options)
+        address = options[:billing_address]
+        payer = {}
+        payer[:fullName] = payment_method.name.strip
+        payer[:contactPhone] = address[:phone] if (address && address[:phone])
+        payer[:dniNumber] = options[:dni_number] if options[:dni_number]
+        payer[:dniType] = options[:dni_type] if options[:dni_type]
+        payer[:emailAddress] = options[:email] if options[:email]
+        payer[:birthdate] = options[:birth_date] if options[:birth_date] && options[:payment_country] == 'MX'
```
davidsantoso: Wondering if we should also check the billing address country for MX here?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -152,28 +153,59 @@ def add_order(post, options)
...
+
+      def billing_address_fields(options)
+        return unless address = options[:billing_address]
+        billing_address = {}
+        billing_address[:street1] = address[:address1]
+        billing_address[:street2] = address[:address2]
+        billing_address[:city] = address[:city]
+        billing_address[:state] = address[:state]
+        billing_address[:country] = address[:country]
+        billing_address[:postalCode] = address[:zip] if options[:payment_country] == 'MX'
```
davidsantoso: Ditto to the above question when building the `payer` hash.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -152,28 +153,59 @@ def add_order(post, options)
...
-          buyer[:shippingAddress] = shipping_address
-          post[:transaction][:order][:buyer] = buyer
-        end
+        buyer = {}
+        buyer[:fullName] = options[:buyer_name] if options[:buyer_name]
+        buyer[:dniNumber] = options[:buyer_dni_number] if options[:buyer_dni_number]
+        buyer[:dniType] = options[:buyer_dni_type] if options[:buyer_dni_type]
+        buyer[:cnpj] = options[:buyer_cnpj] if options[:buyer_cnpj] && options[:payment_country] == 'BR'
+        buyer[:emailAddress] = options[:buyer_email] if options[:buyer_email]
+        buyer[:contactPhone] = options[:shipping_address][:phone] if (options[:shipping_address] && options[:shipping_address][:phone])
```
davidsantoso: Also not sure if we need the additional `&& options[:shipping_address][:phone]` in the guard clause here.

test/remote/gateways/remote_payu_latam_test.rb
```diff
@@ -82,8 +118,68 @@ def test_successful_purchase_brazil
     assert response.test?
   end
 
-  def test_successful_purchase_sans_options
-    response = @gateway.purchase(@amount, @credit_card)
+  def test_successful_purchase_colombia
```
davidsantoso: I'm curious about these country specific ones where we're adding the `payment_country` param to `options`- in the standard `test_successful_purchase` test above we don't add it to the options hash and it seems to be ok (assuming cause it's defaulting to "AR" give the changes here) so wondering what we're checking for each specific country and adding the `payment_country` field? Seems like if we didn't add it, then it would pass as well?

lib/active_merchant/billing/gateways/payu_latam.rb
curiousepic: After inquiring with PayU, they indicated that the per-country required fields refers to the country of the merchant account, not the billing address. I agree it is strange.

test/remote/gateways/remote_payu_latam_test.rb
curiousepic: Unfortunately either the sandbox isn't accurate, or the PaymentCountry field is not truly required for a successful transaction and is more of a soft requirement from PayU, because even without adding the field remote purchase tests succeed.

curiousepic: So this test is more checking that the (supposedly) required country-specific fields are added, rather than that the transaction succeeds.

curiousepic:
Updated with David's suggested changes so far.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -138,6 +138,7 @@ def add_credentials(post, command)
 
       def add_transaction_elements(post, type, options)
         transaction = {}
+        transaction[:paymentCountry] = options[:payment_country] # || (options[:billing_address][:country] if options[:billing_address])
```
dtykocki: Excess comment?

test/remote/gateways/remote_payu_latam_test.rb
davidsantoso: > Unfortunately either the sandbox isn't accurate, or the PaymentCountry field is not truly required

☹️ 

> So this test is more checking that the (supposedly) required country-specific fields are added

I gotcha. So, if that's the case wondering if it would be better for us to do the [`check_request`](https://github.com/activemerchant/active_merchant/blob/d27c34e9e0e714afbf03d5eacf2a8cb20dc5b3da/test/unit/gateways/stripe_test.rb#L283) method of ensuring we are adding that field to the request if it exists in the options has AND that it overrides what is in the billing address. Given the value doesn't seem to matter against the sandbox, we can probably 🔪 each specific country and just ensure that we do send that field/it doesn't get accidentally removed. Basically, I feel dubious that changing out the country even gives us anymore confidence since it doesn't seem to be validated in sandbox?

lib/active_merchant/billing/gateways/payu_latam.rb
curiousepic: :facepalm: thanks.

test/remote/gateways/remote_payu_latam_test.rb
curiousepic: @davidsantoso I `check_request` for them in unit tests when they are options but I can add the check for overriding the billing address. It's possible they could add actual checking to their sandbox, so I'm not sure the remote tests should be removed though.

dtykocki: Regardless of how the sandbox behaves, we should have cases that at a minimum exercise each country option.

davidsantoso: > It's possible they could add actual checking to their sandbox, so I'm not sure the remote tests should be removed though.

That's fine. However I'm still not convinced that it gives us any more confidence.

lib/active_merchant/billing/gateways/payu_latam.rb
davidsantoso: weird.

PR #2571, opened 2017-09-04
Add Maestro as a supported card type

### LINKED ISSUE #2217: Braintree supported_card_types does not include Maestro
As per here https://www.braintreepayments.com/payment-methods/accept-credit-cards

Maestro is supported.

matthewheath:
This resolves #2217.


bpollack:
Thanks for the patch, @matthewheath ! Squashed and merged.

PR #2575, opened 2017-09-08
Add refund to Kushki

dtykocki:
Adds support for performing refund transactions on Kushki. The refund
endpoint is a bit quirky in that the response body only returns
a message and status code, not a unique transaction identifier like
other endpoints. Additionally, the amount param is ignored since the
call to perform a refund only accepts a reference identifier.

```
ruby -Itest test/remote/gateways/remote_kushki_test.rb
Loaded suite test/remote/gateways/remote_kushki_test
Started
........

Finished in 11.867091 seconds.
8 tests, 24 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
0.67 tests/s, 2.02 assertions/s

ruby -Itest test/unit/gateways/kushki_test.rb
Loaded suite test/unit/gateways/kushki_test
Started
........

Finished in 0.009668 seconds.
8 tests, 45 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
827.47 tests/s, 4654.53 assertions/s
```

### DIFF
--- a/test/remote/gateways/remote_kushki_test.rb
+++ b/test/remote/gateways/remote_kushki_test.rb
@@ -51,6 +51,24 @@ class RemoteKushkiTest < Test::Unit::TestCase
     assert_equal 'Monto de la transacción es diferente al monto de la venta inicial', response.message
   end
 
+  def test_successful_refund
+    purchase = @gateway.purchase(@amount, @credit_card)
+    assert_success purchase
+
+    assert refund = @gateway.refund(@amount, purchase.authorization)
+    assert_success refund
+    assert_equal 'Succeeded', refund.message
+  end
+
+  def test_failed_refund
+    purchase = @gateway.purchase(@amount, @credit_card)
+    assert_success purchase
+
+    assert refund = @gateway.refund(@amount, nil)
+    assert_failure refund
+    assert_equal 'Missing Authentication Token', refund.message
+  end
+
   def test_successful_void
     purchase = @gateway.purchase(@amount, @credit_card)
     assert_success purchase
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_kushki_test.rb
```diff
@@ -51,6 +51,15 @@ def test_failed_purchase
...
 
+  def test_successful_refund
+    purchase = @gateway.purchase(@amount, @credit_card)
+    assert_success purchase
+
+    assert refund = @gateway.refund(@amount, purchase.authorization)
+    assert_success refund
+    assert_equal 'Succeeded', refund.message
+  end
+
```
curiousepic: Do we want a failed refund remote test? Or does it not actually trigger in sandbox?

davidsantoso: I imagine if no `ticketNumber` is sent the sandbox would fail?

dtykocki: I tried many ways to generate a failure response here. You would think that a failure would occur with an empty `ticketNumber`, but I only get a 500 error with message: ```{"message":"Missing Authentication Token"}%```

dtykocki: Added a remote test for failed refunds.

curiousepic: @dtykocki I wonder if that means we should count 500s as failure. I've seen some adapters do that.

dtykocki: Yeah, treating a 500 as a failure is something I've considered, but question if that change should happen along with switching to a "real" refund. Best I can tell, all of Kushki's endpoints may return either a 400 or a 500 with a response code and message.

PR #2577, opened 2017-09-09
PayU Latam: use address name for payer when using token

ccarruitero:


curiousepic:
@ccarruitero Is this in support for `store`? FYI `store` isn't currently tested and I think not fully implemented yet.

ccarruitero:
@curiousepic is for support `purchase` with token. In my case I was getting the token from the client using payu latam javascript. But I think should be useful for `store` too.

juan267:
This PR as with #2590 also needs to be merged in order to be able to create purchases using a credit card tokens.

PR #2578, opened 2017-09-11
PayU Latam: Correctly condition buyer element fields

curiousepic:
According to PayU, the Buyer element is required, and if no buyer info
is provided, the Payer's info should be sent in its place, and if only
partial Buyer info is provided, the missing fields should instead be
sent empty.

This change ensures these conditions are fulfilled. It also gathers the
buyer fields under one buyer hash option instead of multiple top-level
buyer_* fields, and exposes a top-level cnpj field for the payer.

Remote tests continue to fail unrelated capture and void.

Remote:
18 tests, 47 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
88.8889% passed

23 tests, 89 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -118,7 +118,7 @@ module ActiveMerchant #:nodoc:
         add_credentials(post, 'SUBMIT_TRANSACTION')
         add_transaction_elements(post, transaction_type, options)
         add_order(post, options)
-        add_buyer(post, options)
+        add_buyer(post, payment_method, options)
         add_invoice(post, amount, options)
         add_signature(post)
         add_payment_method(post, payment_method, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -183,15 +183,25 @@ def billing_address_fields(options)
...
+          buyer[:emailAddress] = buyer_hash[:email]
+          buyer[:contactPhone] = options[:shipping_address][:phone] if options[:shipping_address]
+          buyer[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
+        else
+          buyer[:fullName] = payment_method.name.strip
+          buyer[:dniNumber] = options[:dni_number]
+          buyer[:dniType] = options[:dni_type]
+          buyer[:cnpj] = options[:cnpj] if options[:payment_country] == 'BR'
+          buyer[:emailAddress] = options[:email]
+          buyer[:contactPhone] = (options[:shipping_address][:phone] if options[:shipping_address]) || (options[:billing_address][:phone] if options[:billing_address])
```
davidsantoso: Seems like we check 2 places here for a phone number but on line 194 only under the `shipping_address`. Wondering if there's a reason for the discrepancy?

curiousepic: So, a shipping address may be available independent of specified buyer data, so even if the payer's info is substituted in for the buyer fields, they may have specified a different contact phone number for shipping. If there is buyer data provided, the contact phone number should only be picked from the buyer's shipping address. Does that make sense?

davidsantoso: I think so. I also think I may not have full context given the purpose of these fields and how to properly formulate them.

PR #2580, opened 2017-09-12
Cardknox: fix a bug in some dependant gems because some methods were missing

achasveachas:
Cardknox: fix a bug where sending a credit card without track_data would error


bpollack:
I'm going to close this due to the lack of context and tests, but I'd be really happy to take another look if you add either of those.

PR #2588, opened 2017-09-19
WePay: Don't require email for Store

curiousepic:
No longer requires email for Store actions and instead allows the method
to default the value if it's absent.

Remote:
24 tests, 40 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
20 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/wepay.rb
+++ b/lib/active_merchant/billing/gateways/wepay.rb
@@ -76,8 +76,6 @@ module ActiveMerchant #:nodoc:
       end
 
       def store(creditcard, options = {})
-        requires!(options, :email)
-
         post = {}
         post[:client_id] = @options[:client_id]
         post[:user_name] = "#{creditcard.first_name} #{creditcard.last_name}"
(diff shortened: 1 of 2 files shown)


curiousepic:
Updated to not default the field either.

lib/active_merchant/billing/gateways/wepay.rb
```diff
@@ -76,12 +76,10 @@ def refund(money, identifier, options = {})
       end
 
       def store(creditcard, options = {})
-        requires!(options, :email)
-
         post = {}
         post[:client_id] = @options[:client_id]
         post[:user_name] = "#{creditcard.first_name} #{creditcard.last_name}"
-        post[:email] = options[:email] || "unspecified@example.com"
```
dtykocki: Any particular reason why we don't default the email?

curiousepic:
Alright, back to defaulting since it seems a little more idiomatic for ActiveMerchant.

dtykocki:
👍

PR #2589, opened 2017-09-21
Adyen: Use original authorization pspReference on Refunds

lyverovski:
Adyen requires the authorization `pspReference` to be passed in on refund requests, and not the  capture `pspReference` which is currently being sent.

A common pattern used in other Active Merchant gateways is to overload the `authorization` parameter and store multiple downstream gateway specific response parameters. This is the pattern that was chose here as well. For capture requests, the `authorization` parameter will now store the original authorization response `pspReference`,  along with the capture response `pspReference` (joined with '#')

The change is "backwards compatible", in case some implementers already only have the single value as a pspReference, however the Adyen gateway will behave as it does today, and decline those requests.


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -59,7 +59,7 @@ def capture(money, authorization, options={})
       def refund(money, authorization, options={})
         post = init_post(options)
         add_invoice_for_modification(post, money, authorization, options)
-        add_references(post, authorization, options)
+        add_references(post, psp_reference_from(authorization), options)
```
lyverovski: This is a mistake.
I'll have to revisit. It should just be `authorization` and not `psp_reference_from(authorization)`, since `add_references` already does that.

lyverovski:
Actually, the test that was missing was the `gateway.refund("one#two")`, since the compound pspReference occurs as a result of a capture.

The equivalent `gateway.capture("one#two")` test is not really applicable (but I added it anyway). A capture would be receiving the pspReference of the auth, which will not be a compound pspReference (i.e it will not have a # join).

PR #2591, opened 2017-09-25
Payeezy: Adds :store functionality.

deedeelavinder:
Adds :store to Payeezy gateway using Transarmor token according to
documentation here:
https://developer.payeezy.com/payeezy-api/apis/get/securitytokens

Refactors gateway to allow endpoint for tokenization and to handle
returned JSONP object. Concatenates token data for storage.

Unit Tests:
27 tests, 130 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote Tests:
25 tests, 90 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payeezy.rb
+++ b/lib/active_merchant/billing/gateways/payeezy.rb
@@ -3,9 +3,9 @@ module ActiveMerchant
     class PayeezyGateway < Gateway
       class_attribute :integration_url
 
-      self.test_url = 'https://api-cert.payeezy.com/v1/transactions'
-      self.integration_url = 'https://api-cat.payeezy.com/v1/transactions'
-      self.live_url = 'https://api.payeezy.com/v1/transactions'
+      self.test_url = 'https://api-cert.payeezy.com/v1'
+      self.integration_url = 'https://api-cat.payeezy.com/v1'
+      self.live_url = 'https://api.payeezy.com/v1'
 
       self.default_currency = 'USD'
       self.money_format = :cents
@@ -31,7 +31,7 @@ module ActiveMerchant
       end
 
       def purchase(amount, payment_method, options = {})
-        params = {transaction_type: 'purchase'}
+        params = payment_method.is_a?(String) ? { transaction_type: 'recurring' } : { transaction_type: 'purchase' }
 
         add_invoice(params, options)
         add_payment_method(params, payment_method, options)
@@ -69,6 +69,14 @@ module ActiveMerchant
 
         add_authorization_info(params, authorization)
         add_amount(params, (amount || amount_from_authorization(authorization)), options)
+
+        commit(params, options)
+      end
+
+      def store(payment_method, options = {})
+        params = {}
+
+        add_creditcard_for_tokenization(params, payment_method, options)
 
         commit(params, options)
       end
@@ -119,25 +127,32 @@ module ActiveMerchant
         params[:method] = method
       end
 
+      def add_creditcard_for_tokenization(params, payment_method, options)
+        params[:apikey] = @options[:apikey]
+        params[:js_security_key] = options[:js_security_key]
+        params[:ta_token] = options[:ta_token]
+        params[:callback] = 'Payeezy.callback'
+        params[:type] = 'FDToken'
+        card = add_card_data(payment_method)
+        params['credit_card.type'] = card[:type]
+        params['credit_card.cardholder_name'] = card[:cardholder_name]
+        params['credit_card.card_number'] = card[:card_number]
+        params['credit_card.exp_date'] = card[:exp_date]
+        params['credit_card.cvv'] = card[:cvv]
+      end
+
+      def is_store_action?(params)
+        params[:ta_token].present?
+      end
+
       def add_payment_method(params, payment_method, options)
         if payment_method.is_a? Check
           add_echeck(params, payment_method, options)
+        elsif payment_method.is_a? String
+          add_token(params, payment_method, options)
         else
           add_creditcard(params, payment_method)
         end
-      end
-
-      def add_creditcard(params, creditcard)
-        credit_card = {}
-
-        credit_card[:type] = CREDIT_CARD_BRAND[creditcard.brand]
-        credit_card[:cardholder_name] = creditcard.name
-        credit_card[:card_number] = creditcard.number
-        credit_card[:exp_date] = "#{format(creditcard.month, :two_digits)}#{format(creditcard.year, :two_digits)}"
-        credit_card[:cvv] = creditcard.verification_value if creditcard.verification_value?
-
-        params[:method] = 'credit_card'
-        params[:credit_card] = credit_card
       end
 
       def add_echeck(params, echeck, options)
@@ -154,6 +169,44 @@ module ActiveMerchant
 
         params[:method] = 'tele_check'
         params[:tele_check] = tele_check
+      end
+
+      def add_token(params, payment_method, options)
+        token = {}
+        token[:token_type] = 'FDToken'
+
+        type, cardholder_name, exp_date, card_number = payment_method.split('|')
+
+        token[:token_data] = {}
+        token[:token_data][:type] = type
+        token[:token_data][:cardholder_name] = cardholder_name
+        token[:token_data][:value] = card_number
+        token[:token_data][:exp_date] = exp_date
+        token[:token_data][:cvv] = options[:cvv] if options[:cvv]
+
+        params[:method] = 'token'
+        params[:token] = token
+      end
+
+      def add_creditcard(params, creditcard)
+        credit_card = add_card_data(creditcard)
+
+        params[:method] = 'credit_card'
+        params[:credit_card] = credit_card
+      end
+
+      def add_card_data(payment_method)
+        card = {}
+        card[:type] = CREDIT_CARD_BRAND[payment_method.brand]
+        card[:cardholder_name] = payment_method.name
+        card[:card_number] = payment_method.number
+        card[:exp_date] = format_exp_date(payment_method.month, payment_method.year)
+        card[:cvv] = payment_method.verification_value if payment_method.verification_value?
+        card
+      end
+
+      def format_exp_date(month, year)
+        "#{format(month, :two_digits)}#{format(year, :two_digits)}"
       end
 
       def add_address(params, options)
@@ -180,25 +233,18 @@ module ActiveMerchant
       end
 
       def commit(params, options)
-        url = if options[:integration]
-          integration_url
-        elsif test?
-          test_url
-        else
-          live_url
-        end
+        url = base_url(options) + endpoint(params)
 
         if transaction_id = params.delete(:transaction_id)
           url = "#{url}/#{transaction_id}"
         end
 
         begin
-          body = params.to_json
-          response = parse(ssl_post(url, body, headers(body)))
+          response = api_request(url, params)
         rescue ResponseError => e
           response = response_error(e.response.body)
         rescue JSON::ParserError
-          response = json_error(raw_response)
+          response = json_error(e.response.body)
         end
 
         Response.new(
@@ -213,17 +259,34 @@ module ActiveMerchant
         )
       end
 
-      def success_from(response)
-        response['transaction_status'] == 'approved'
+      def base_url(options)
+        if options[:integration]
+          integration_url
+        elsif test?
+          test_url
+        else
+          live_url
+        end
       end
 
-      def authorization_from(params, response)
-        [
-          response['transaction_id'],
-          response['transaction_tag'],
-          params[:method],
-          (response['amount'] && response['amount'].to_i)
-        ].join('|')
+      def endpoint(params)
+        is_store_action?(params) ? '/securitytokens' : '/transactions'
+      end
+
+      def api_request(url, params)
+        if is_store_action?(params)
+          callback = ssl_request(:get, "#{url}?#{post_data(params)}", nil, {})
+          payload = callback[/{(?:\n|.)*}/]
+          parse(payload)
+        else
+          body = params.to_json
+          parse(ssl_post(url, body, headers(body)))
+        end
+      end
+
+      def post_data(params)
+        return nil unless params
+        params.reject { |k, v| v.blank? }.collect { |k, v| "#{k}=#{CGI.escape(v.to_s)}" }.join("&")
       end
 
       def generate_hmac(nonce, current_timestamp, payload)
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -119,28 +135,30 @@ def add_authorization_info(params, authorization)
         params[:method] = method
       end
 
+      def create_token(params, payment_method)
```
curiousepic: Slight misnomer. `add_card_for_tokenization`?

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -119,28 +135,30 @@ def add_authorization_info(params, authorization)
...
+        params[:apikey] = PARAMS_FOR_TOKENIZATION[:api_key]
+	      params[:js_security_key] = PARAMS_FOR_TOKENIZATION[:js_security_key]
+	      params[:ta_token] = PARAMS_FOR_TOKENIZATION[:ta_token]
+	      params[:callback] = PARAMS_FOR_TOKENIZATION[:callback]
+	      params[:type] = PARAMS_FOR_TOKENIZATION[:type]
+        params['credit_card.type'] = CREDIT_CARD_BRAND[payment_method.brand]
+        params['credit_card.cardholder_name'] = payment_method.name
+        params['credit_card.card_number'] = payment_method.number
+        params['credit_card.exp_date'] = format_exp_date(payment_method.month, payment_method.year)
+        params['credit_card.cvv'] = payment_method.verification_value if payment_method.verification_value?
```
curiousepic: and/or could this be DRYed up with `use_creditcard`/`add_creditcard`?

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -156,6 +174,40 @@ def add_echeck(params, echeck, options)
         params[:tele_check] = tele_check
       end
 
+      def use_token(params, payment_method, options)
```
curiousepic: I think `add_` is more idiomatic in AM than `use_` which is less informative.

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -156,6 +174,40 @@ def add_echeck(params, echeck, options)
...
+	      credit_card = {}
+
+	      credit_card[:type] = CREDIT_CARD_BRAND[creditcard.brand]
+	      credit_card[:cardholder_name] = creditcard.name
+	      credit_card[:card_number] = creditcard.number
+	      credit_card[:exp_date] = format_exp_date(creditcard.month, creditcard.year)
+	      credit_card[:cvv] = creditcard.verification_value if creditcard.verification_value?
+
+	      params[:method] = 'credit_card'
+	      params[:credit_card] = credit_card
```
curiousepic: Indents!

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -180,50 +232,63 @@ def add_soft_descriptors(params, options)
...
+          response = json_error(e.response.body)
         end
 
         Response.new(
           success_from(response),
           handle_message(response, success_from(response)),
           response,
           test: test?,
-          authorization: authorization_from(params, response),
+          authorization: success_from(response) ? authorization_from(params, response) : 'Authorization Failed',
```
curiousepic: Most adapters just let the `Authorization` be `nil` for a failed txn and I think we can let that be the case here too.

Also just FYI "Authorization Failed" wouldn't necessarily be accurate, since the word Authorization could mean the type of action or a response reference which is what it means here.

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: 👍

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: I looked at that but it would require a potentially more confusing refactor of `add_creditcard` due to other params being used & set in that method.

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: My thought was the `add` in `add_payment_method` would suffice and that the `add` here was confusing in the context of adding a tokenized card. Happy to switch it back if `use` isn't working.

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: GAH!

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: 👍

lib/active_merchant/billing/gateways/payeezy.rb
curiousepic: 👍 IME, any method that directly adds values to the hash that will become the request uses `add_`.

lib/active_merchant/billing/gateways/payeezy.rb
curiousepic: I think it's reasonable for them to both call a separate method for the card data itself (something like `card_data`), adding a hash it returns to the appropriate element. I think some gateways do this, though I can't seem to find an example. It's not a big deal though.

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -119,27 +135,29 @@ def add_authorization_info(params, authorization)
...
+        params['credit_card.cardholder_name'] = payment_method.name
+        params['credit_card.card_number'] = payment_method.number
+        params['credit_card.exp_date'] = format_exp_date(payment_method.month, payment_method.year)
+        params['credit_card.cvv'] = payment_method.verification_value if payment_method.verification_value?
+      end
+
       def add_payment_method(params, payment_method, options)
         if payment_method.is_a? Check
           add_echeck(params, payment_method, options)
+        elsif payment_method.is_a? String
```
curiousepic: Hmm. If reference purchases were refer implemented, this could collide with those, since they are not tokens. Maybe premature optimization.

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -180,50 +232,63 @@ def add_soft_descriptors(params, options)
...
-          response['transaction_tag'],
-          params[:method],
-          (response['amount'] && response['amount'].to_i)
-        ].join('|')
+      def endpoint(params)
+        params[:ta_token].present? ? '/securitytokens' : '/transactions'
+      end
+
+      def api_request(url, params)
+        if params[:ta_token].present?
```
curiousepic: Hmm. Checking if `transaction_type` is `store` might be more readable/reasonable?

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -180,50 +232,63 @@ def add_soft_descriptors(params, options)
...
+          parse(payload)
+        else
+          body = params.to_json
+          parse(ssl_post(url, body, headers(body)))
+        end
+      end
+
+      def post_data(params)
+        return nil unless params
+        params.to_query
```
curiousepic: Whoa `.to_query` :mind_blown:
That said, I think we're trying to reduce dependency on ActiveSupport, so you might check out how some other adapters construct queries.

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -25,13 +25,21 @@ class PayeezyGateway < Gateway
         'diners_club' => 'Diners Club'
       }
 
+      PARAMS_FOR_TOKENIZATION = {
+        api_key: 'y6pWAJNyJyjGv66IsVuWnklkKUPFbb0a',
+        js_security_key: 'js-6125e57ce5c46e10087a545b9e9d7354c23e1a1670d9e9c7',
+        ta_token: 'NOIW',
+        callback: 'Payeezy.callback',
+        type: 'FDToken'
+      }
```
dtykocki: Where does the `api_key` and `js_security_key` come from? Are these sensitive values that should not be hardcoded in?

curiousepic: Good catch - at first I thought these were static, public values from https://developer.payeezy.com/payeezy-api/apis/post/transactions-14 but those must be example values - api_key is a credential, and it looks like the gateway needs to activate the ability to store on a per-merchant account basis. That might be how the `js_security_key` is provided?

davidsantoso: Looking at the Payeezy docs, looks like the `api_key` is the same as the one we use in the `initialize` so we can reference it in the `@options` hash. Also, wondering if we should move the `js_security_key` to the initialize option since it's a credential. Not a required one, just set it's value there if it was passed in.

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -119,27 +135,29 @@ def add_authorization_info(params, authorization)
         params[:method] = method
       end
 
+      def add_creditcard_for_tokenization(params, payment_method)
+        params[:apikey] = PARAMS_FOR_TOKENIZATION[:api_key]
+        params[:js_security_key] = PARAMS_FOR_TOKENIZATION[:js_security_key]
+        params[:ta_token] = PARAMS_FOR_TOKENIZATION[:ta_token]
+        params[:callback] = PARAMS_FOR_TOKENIZATION[:callback]
+        params[:type] = PARAMS_FOR_TOKENIZATION[:type]
```
davidsantoso: Payeezy docs make it seem like the `ta_token`, `callback`, and `type` fields are static? Correct me if I'm wrong 😬 If they are, then we can just hardcode those inline vs reference them in a hash.

lib/active_merchant/billing/gateways/payeezy.rb
dtykocki: 👍 Their documentation is certainly lacking in this area.

@deedeelavinder - It might be worth digging through their admin / management UI to verify how sensitive these two values are.

lib/active_merchant/billing/gateways/payeezy.rb
davidsantoso: +1 to using something other than `.to_query` purely from an Active Support dependency position.

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: 👍

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: No `transaction_type` in calls to the `/securitytokens` endpoint.

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: 👍

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: 👍

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: As @davidsantoso pointed out, we can use the `:apikey` used to initialize the gateway. The `js_security_key` is merchant-specific but is not treated as ultra-sensitive in their documentation and support forums. This now allows that parameter to be passed in with options. That will allow backward compatibility for current users who wish to use `store` in the future.

lib/active_merchant/billing/gateways/payeezy.rb
dtykocki: As I'm reading through this and run into `params[:ta_token]` used in a few conditionals I start to wonder what the presence of a `ta_token` represents. It seems that one must know that a `ta_token` is used when storing a card. I came to this conclusion by scrolling up and reading `add_creditcard_for_tokenization`. This is fine, but I wonder if we should encapsulate this concept in a single place, i.e. (another) private method.

lib/active_merchant/billing/gateways/payeezy.rb
dtykocki: I'm okay with worrying about this when reference purchases are requested.

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: 👍 Great suggestion -- added and open to naming suggestions.

lib/active_merchant/billing/gateways/payeezy.rb
deedeelavinder: 👍

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -119,27 +127,34 @@ def add_authorization_info(params, authorization)
         params[:method] = method
       end
 
+      def add_creditcard_for_tokenization(params, payment_method, options)
+        params[:apikey] = options[:apikey]
+        params[:js_security_key] = options[:js_security_key]
+        params[:ta_token] = 'NOIW'
```
sebastianGitDL: Hello! How are you?

Thanks for helping improving this particular Gateway! I am currently working on the implementation of Payeezy and I have to use the same method as well. I just wanted to share with you that the ta_token should be configurable, since that token NOIW is only accepted in a SANDBOX environment. If we want to use this tokenization feature in a LIVE environment, the ta_token should be set according to the given by FirstData in the Terminal (take a look at the documentation for that particular field https://developer.payeezy.com/payeezy-api/apis/get/securitytokens)

Hope this helps!

deedeelavinder: Thank you, @sebastianGitDL -- will modify now.

test/remote/gateways/remote_payeezy_test.rb
```diff
@@ -64,7 +92,7 @@ def test_failed_authorize
     @amount = 501300
     assert auth = @gateway.authorize(@amount, @credit_card, @options)
     assert_failure auth
-    assert auth.authorization
+    assert_equal nil, auth.authorization
```
dtykocki: This reads like `purchase` used to return an authorization structure, but will now be `nil`. Seems like a compatibility issue?

deedeelavinder: @dtykocki Restructured the way the authorization is built to accommodate the failed store response and returned these tests to the way they were.

lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -119,27 +127,34 @@ def add_authorization_info(params, authorization)
         params[:method] = method
       end
 
+      def add_creditcard_for_tokenization(params, payment_method, options)
+        params[:apikey] = options[:apikey]
```
dtykocki: I think we can use `@options[:apikey]` here since it's passed in via `initialize`. It looks like the call to `super` will create this instance variable: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L183.

deedeelavinder: 👍

PR #2594, opened 2017-09-26
[Wirecard] use localized_amount to fix non-fractional currencies

bdewater:
We got reports amounts were being overcharged by two orders of magnitude, e.g. for authorizing 149000.00 KRW this was sent to the gateway:

```xml
<amount currency="KRW">14900000</amount> 
<operation>auth</operation> 
```

localized_amount takes the currency and gateway format into account to do the right thing.


davidsantoso:
👍

bdewater:
Merged with https://github.com/activemerchant/active_merchant/commit/2b89e5295c1fdc7769e44f51b9fbd36b0c84fcf9

PR #2603, opened 2017-10-03
Barclaycard Smartpay: Refactors address parsing.

deedeelavinder:
Although using a regex has worked for *most* of the needed address
parsing, there are still a few exceptions. Due to the unusual address
structure, this allows some require fields to
come through in the options hash: `:street`, `:shipping_street`,
`:house_number`, and `:shipping_house_number`.

Unit Tests:
17 tests, 59 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote Tests:
24 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
+++ b/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
@@ -33,15 +33,8 @@ module ActiveMerchant #:nodoc:
         post = payment_request(money, options)
         post[:amount] = amount_hash(money, options[:currency])
         post[:card] = credit_card_hash(creditcard)
-
-        if address = (options[:billing_address] || options[:address])
-          post[:billingAddress] = address_hash(address)
-        end
-
-        if options[:shipping_address]
-          post[:deliveryAddress] = address_hash(options[:shipping_address])
-        end
-
+        post[:billingAddress] = billing_address_hash(options)
+        post[:deliveryAddress] = shipping_address_hash(options) if options[:shipping_address]
         commit('authorise', post)
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -230,17 +223,41 @@ def build_url(action)
...
+
+      def parse_street(address)
+        address_to_parse = "#{address[:address1]} #{address[:address2]}"
+        street = address[:street] || address_to_parse.split(/\s+/).keep_if { |x| x !~ /\d/ }.join(' ')
+        street.empty? ? "Not Provided" : street
+      end
+
+      def parse_house_number(address)
+        address_to_parse = "#{address[:address1]} #{address[:address2]}"
+        house = address[:houseNumberOrName] || address_to_parse.split(/\s+/).keep_if { |x| x =~ /\d/ }.join(' ')
```
curiousepic: I know it was assumed in (my) current code, but I don't think there'd ever be an `address[:houseNumberOrName]` or `address[:street]` field sent to the adapter, so I think they can be removed. I think I was anticipating needing this split (which this commit is taking care of), but it wouldn't be sent in the options' `address` element.

curiousepic: Though that begs the question if we need top-level street and house_number fields for both billing and shipping address...

deedeelavinder: Agreed about `address[:houseNumberOrName]` or `address[:street]` being unlikely, but will leave in case they are currently being used somewhere since they were pre-existing. 

This does include top-level street and house_number fields for both billing and shipping addresses.

deedeelavinder:
@curiousepic -- Unit tests added using `.check_request` to verify request structure with new option fields, both for billing and shipping.

test/unit/gateways/barclaycard_smartpay_test.rb
```diff
@@ -41,6 +86,48 @@ def test_successful_purchase
     assert response.test?
   end
 
+  def test_successful_purchase_with_alternate_address
+    response = stub_comms do
+      @gateway.purchase(@amount, @credit_card, @options_with_alternate_address)
+    end.respond_with(successful_authorize_response, successful_capture_response)
```
curiousepic: It might be good to either `.check_request` the fallback parsing of the `address1: '新北市店溪路3579號139樓'` here, or just remove the test, since there's not much point in testing the success of the mock transaction.

deedeelavinder: 👍 Good point -- `.check_request` now added.The fallback parsing will return 'Not Provided' for the street address which is a new addition.

PR #2604, opened 2017-10-03
Ebanx: Support Store and person_type option

curiousepic:
Adds Store support. The "responsible" element is now only added when it
is required, for transactions in Brazil with a person_type of
"business". Store also passes "country" from options, as it is required.

Four negative test cases are failing due to changes to test card numbers
we have yet to receive an update for.

Remote:
18 tests, 46 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
77.7778% passed

Unit:
15 tests, 51 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -24,7 +24,8 @@ module ActiveMerchant #:nodoc:
         authorize: "direct",
         capture: "capture",
         refund: "refund",
-        void: "cancel"
+        void: "cancel",
+        store: "token"
       }
 
       HTTP_METHOD = {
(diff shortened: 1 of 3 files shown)


deedeelavinder:
👍

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -239,6 +274,12 @@ def error_code_from(response, success)
           response.try(:[], "payment").try(:[], "transaction_status").try(:[], "code")
         end
       end
+
+      def customer_country(options)
+        if country = options[:country] || (options[:billing_address][:country] if options[:billing_address])
+          country.downcase
```
davidsantoso: Curious why the downcase? Usually country abbreviations are uppercase.

curiousepic: Their documentation uses downcase so I figured better safe than sorry.

davidsantoso: 👍 thanks for the explanation!

PR #2605, opened 2017-10-04
Borgun: Include currency code from split authorization

dtykocki:
When performing a void transaction on Borgun, the currency code must
match the original currency code used on the authorization. Because of
this, the adapter will now store the original currency code in the
authorization and include it when void is called.

### DIFF
--- a/lib/active_merchant/billing/gateways/borgun.rb
+++ b/lib/active_merchant/billing/gateways/borgun.rb
@@ -55,9 +55,10 @@ module ActiveMerchant #:nodoc:
 
       def void(authorization, options={})
         post = {}
-        # TransType and TrAmount must match original values from auth or purchase.
-        _, _, _, _, _, transtype, tramount = split_authorization(authorization)
+        # TransType, TrAmount, and currency must match original values from auth or purchase.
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/borgun.rb
```diff
@@ -55,10 +55,11 @@ def refund(money, authorization, options={})
...
       def void(authorization, options={})
         post = {}
-        # TransType and TrAmount must match original values from auth or purchase.
-        _, _, _, _, _, transtype, tramount = split_authorization(authorization)
+        # TransType, TrAmount, and currency must match original values from auth or purchase.
+        _, _, _, _, _, transtype, tramount, currency = split_authorization(authorization)
         post[:TransType] = transtype
-        add_invoice(post, tramount.to_i, options)
+        post[:TrAmount] = tramount.to_i
+        post[:TrCurrency] = currency
```
dtykocki: I'm going with a different approach here; I realized that this change won't be backwards compatible.

dtykocki:
@curiousepic @deedeelavinder @davidsantoso - Updated this slightly to ensure backwards compatability.

PR #2609, opened 2017-10-06
Improve payeezy gateway to allow billing a tokenized credit card

sebastianGitDL:
Hello!

I have made a little change in Payeezy gateway to allow purchases over tokenized credit cards. As you can see I added the add_token method which allows the usage of a Transarmor token instead of the credit card number. 

Let me know if you have any doubt.


bdewater:
Hello! And thanks for contributing. It looks like some folks are already working on this in https://github.com/activemerchant/active_merchant/pull/2591 - could you take a look and see if there's anything missing over there?

sebastianGitDL:
Hi! Nice! I'll take a look at that link and run some tests then. Thanks for for the answer!

PR #2624, opened 2017-10-17
CyberSource: Support 3DSecure requests

curiousepic:
3 remote failures include pinless debit card tests which this commit is
not concerned with and a false alarm from overeager scrubbing assertion.

Remote:
39 tests, 180 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92.3077% passed

Unit:
44 tests, 212 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

I'd appreciate a glance from @davidsantoso

### DIFF
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -9,6 +9,12 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
     @credit_card = credit_card('4111111111111111', verification_value: '321')
     @declined_card = credit_card('801111111111111')
     @pinless_debit_card = credit_card('4002269999999999')
+    @three_ds_enrolled_card = credit_card('4000000000000002',
+      verification_value: '321',
+      month: "12",
+      year: "#{Time.now.year + 2}",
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -363,6 +369,32 @@ def test_successful_retrieve_subscription
     assert response.test?
   end
 
+  def test_3ds_purchase_request
+    assert response = @gateway.purchase(1202, @three_ds_enrolled_card, @options.merge(payer_auth_enroll_service: true))
```
davidsantoso: I wonder if we have other gateways that send a similar, "check if the card is 3DS enabled" field and we've defaulted in Active Merchant? Not sure, but could be worth a check to see if there's a more or less default field we use for this.

curiousepic: Yeah it's kinda difficult to determine, which probably means there's no standard. I'm fine with keeping the option field verbatim to the actual request field sent until there's a more concerted effort to standardize.

PR #2625, opened 2017-10-19
Stripe quickchip support

ryanbalsdon:
This patch does two thing. The first was to add support for the quickchip mode for EMV payments. This brought the number of mutually-exclusive EMV-related states in credit_card up to 5 so this patch combines them into a single param. That second change will be breaking if anyone else is using AM to process chip cards so I'm open to killing it.


test/remote/gateways/remote_stripe_emv_test.rb
```diff
@@ -1,28 +1,31 @@
...
+      contactless: ActiveMerchant::Billing::CreditCard.new(icc_data: '5A08476173900101011957114761739001010119D22122011758928889500D5649534120454C454354524F4E5F20175649534120434445542032312F434152443035202020205F2A0208405F340111820200008407A00000000320109A031505119C01009F02060000000006009F0607A00000000320109F090200019F100706011103A000009F1A0200569F1C0831323334353637389F1E0831303030333236389F260852A5A96394EDA96D9F2701809F3303E0B8C89F3501229F360200069F3704A4428D7A9F410400000289')
     }
 
     @options = {
       :currency => "USD",
       :description => 'ActiveMerchant Test Purchase',
       :email => 'wow@example.com'
     }
+
+    @capture_options = { icc_data: "9F270140" }
```
bizla: A small comment as to what that hex string corresponds to would be helpful.

lib/active_merchant/billing/credit_card.rb
```diff
@@ -176,21 +176,20 @@ def requires_verification_value?
...
-
-      # Returns or sets whether card-present EMV data has been read contactlessly.
-      #
-      # @return [true, false]
-      attr_accessor :contactless_emv
-
-      # Returns or sets whether card-present magstripe data has been read contactlessly.
-      #
-      # @return [true, false]
-      attr_accessor :contactless_magstripe
```
bizla: 👍  Nice cleanup.

lib/active_merchant/billing/credit_card.rb
```diff
@@ -176,21 +176,20 @@ def requires_verification_value?
...
+      attr_accessor :read_method
+
+      READ_METHOD_DESCRIPTIONS = {
+        nil => 'A card reader was not used.',
+        'fallback_no_chip' => 'Magstripe was read because the card has no chip.',
+        'fallback_chip_error' => 'Magstripe was read because the card\'s chip failed.',
+        'contactless' => 'Data was read by a Contactless EMV kernel. Issuer script results are not available.',
+        'contactless_magstripe' => 'Contacless data was read with a non-EMV protocol.',
```
bizla: Typo: `Contacless`

lib/active_merchant/billing/credit_card.rb
```diff
@@ -176,21 +176,20 @@ def requires_verification_value?
...
-      # Returns or sets whether card-present magstripe data has been read contactlessly.
-      #
-      # @return [true, false]
-      attr_accessor :contactless_magstripe
+      attr_accessor :read_method
+
+      READ_METHOD_DESCRIPTIONS = {
+        nil => 'A card reader was not used.',
+        'fallback_no_chip' => 'Magstripe was read because the card has no chip.',
+        'fallback_chip_error' => 'Magstripe was read because the card\'s chip failed.',
```
bizla: Can do double quotes instead of escaping `'`: 
`"Magstripe was read because the card's chip failed."`

lib/active_merchant/billing/credit_card.rb
```diff
@@ -176,21 +176,20 @@ def requires_verification_value?
...
+      attr_accessor :read_method
+
+      READ_METHOD_DESCRIPTIONS = {
+        nil => 'A card reader was not used.',
+        'fallback_no_chip' => 'Magstripe was read because the card has no chip.',
+        'fallback_chip_error' => "Magstripe was read because the card's chip failed.",
+        'contactless' => 'Data was read by a Contactless EMV kernel. Issuer script results are not available.',
+        'contactless_magstripe' => 'Contactless data was read with a non-EMV protocol.',
+        'contact' => 'Data was read using the EMV protocol. Issuer script results may follow.',
+        'contact_quickchip' => 'Data was read by the Quickchip EMV kernel. Issuer script results are not available.',
```
nima-namjouy: 👍 .. we could potentially support QR code based emv processing in the future

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -383,18 +384,19 @@ def add_creditcard(post, creditcard, options)
         card = {}
         if emv_payment?(creditcard)
           add_emv_creditcard(post, creditcard.icc_data)
-          post[:card][:read_method] = "contactless" if creditcard.contactless_emv
-          post[:card][:read_method] = "contactless_magstripe_mode" if creditcard.contactless_magstripe
+          post[:card][:read_method] = "contactless" if creditcard.read_method == 'contactless'
```
nima-namjouy: For normal contact EMV transactions, does `read_method` get set in the post response?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -383,18 +384,19 @@ def add_creditcard(post, creditcard, options)
...
         elsif creditcard.respond_to?(:number)
           if creditcard.respond_to?(:track_data) && creditcard.track_data.present?
             card[:swipe_data] = creditcard.track_data
-            card[:fallback_reason] = creditcard.fallback_reason if creditcard.fallback_reason
-            card[:read_method] = "contactless" if creditcard.contactless_emv
-            card[:read_method] = "contactless_magstripe_mode" if creditcard.contactless_magstripe
+            if creditcard.respond_to?(:read_method)
+              card[:fallback_reason] = 'no_chip' if creditcard.read_method == 'fallback_no_chip'
+              card[:fallback_reason] = 'chip_error' if creditcard.read_method == 'fallback_chip_error'
+              card[:read_method] = "contactless_magstripe_mode" if creditcard.read_method == 'contactless_magstripe'
```
nima-namjouy: does `read_method` get set for standard magstripe and fallback transactions?

test/remote/gateways/remote_stripe_emv_test.rb
```diff
@@ -131,20 +127,19 @@ def test_authorization_and_void_with_emv_credit_card_in_uk
   end
 
   def test_authorization_and_void_with_emv_contactless_credit_card
```
nima-namjouy: is `authorization` a transaction type that's supported for emv contactless?  does it get captured automatically?

jasonwebster:
Can you summarize the contact changes here the break backwards compatibility?

lib/active_merchant/billing/gateways/stripe.rb
ryanbalsdon: The `read_method` param is only used for contactless payments.

test/remote/gateways/remote_stripe_emv_test.rb
ryanbalsdon: It will get captured automatically but you're right, this test doesn't match how we use the gateways. I'll fix it.

ryanbalsdon:
The breaking change here is in credit_card.rb. Before, we had fallback_reason with two valid states and two contactless bools contactless_emv and contactless_magstripe. This patch needed to add _another_ bool to that list called quickchip. The thing that bothered me about that was all these things are mutually exclusive. Setting both contactless flags for example is an error. You can't fallback to a quickchip transaction, etc.

lib/active_merchant/billing/gateways/stripe.rb
ryanbalsdon: `read_method` only gets set for contactless!

test/remote/gateways/remote_stripe_emv_test.rb
```diff
@@ -55,31 +70,28 @@ def test_successful_purchase_with_emv_contactless_credit_card
   end
 
   def test_authorization_and_capture_with_emv_credit_card_in_uk
-    @gateway = StripeGateway.new(fixtures(:stripe_emv_uk))
     assert authorization = @gateway.authorize(@amount, @emv_credit_cards[:uk], @options)
```
nima-namjouy: Is there any need to set the `read_method` for contact transaction tests?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -112,6 +112,7 @@ def purchase(money, payment, options = {})
           end
           r.process do
             post = create_post_for_auth_or_purchase(money, payment, options)
+            post[:card][:processing_method] = 'quick_chip' if quickchip_payment?(payment)
```
nima-namjouy: @ryanbalsdon could you clarify why `quick_chip` processing method has to be set here (instead of here [https://github.com/activemerchant/active_merchant/pull/2625/files#diff-8299be227de5ba32c34c30c42a828c9cR388](url)).  Perhaps a comment might be necessary to prevent someone from accidentally moving this.

bpollack:
@ryanbalsdon Seems like you merged this already in e99f157f; can this be closed?

ryanbalsdon:
Looks like I messed up the commit message and this didn't auto-close, sorry!

PR #2627, opened 2017-10-21
Make library files individually `require`-able

mamhoff:
We would like to be able to use parts of ActiveMerchant without having to load hundreds of Gateways and their dependencies. 

This is a first step towards individually require-able library files. 

I have not checked whether `ActiveSupport` dependencies are in those files. They probably are, but for all the Rails users, `ActiveSupport` will be already loaded.

### DIFF
--- a/lib/active_merchant.rb
+++ b/lib/active_merchant.rb
@@ -51,16 +51,6 @@ require 'active_merchant/posts_data'
 require 'active_merchant/billing'
 require 'active_merchant/version'
 require 'active_merchant/country'
-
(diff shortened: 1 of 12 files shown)


bdewater:
I think this is a great idea, and a fair tradeoff that users opting into this bring their own active_support require if needed. @jasonwebster @davidsantoso WDYT?

bpollack:
I think my previous comment is way off-base. Having looked at this more carefully, I think it'd be great to land, and would help move towards supporting #2923

PR #2631, opened 2017-10-26
Support extra data for Barclay Smartcard credits

bpollack:
Barclays Smartcard will be requiring credits that lack a PSP to include four additional pieces of information: the date-of-birth, entity type, nationality, and full name of the person requesting the credit. This change enables including them.

### DIFF
--- a/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
+++ b/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
@@ -60,6 +60,10 @@ module ActiveMerchant #:nodoc:
         post = payment_request(money, options)
         post[:amount] = amount_hash(money, options[:currency])
         post[:card] = credit_card_hash(creditcard)
+        post[:dateOfBirth] = options[:date_of_birth] if options[:date_of_birth]
+        post[:entityType]  = options[:entity_type] if options[:entity_type]
+        post[:nationality] = options[:nationality] if options[:nationality]
+        post[:shopperName] = options[:shopper_name] if options[:shopper_name]
 
         commit('refundWithData', post)
       end
--- a/test/remote/gateways/remote_barclaycard_smartpay_test.rb
+++ b/test/remote/gateways/remote_barclaycard_smartpay_test.rb
@@ -70,6 +70,31 @@ class RemoteBarclaycardSmartpayTest < Test::Unit::TestCase
         description: 'Store Purchase'
     }
 
+    @options_with_credit_fields = {
+      order_id: '1',
+      billing_address:       {
+              name:     'Jim Smith',
+              address1: '100 Street',
+              company:  'Widgets Inc',
+              city:     'Ottawa',
+              state:    'ON',
+              zip:      'K1C2N6',
+              country:  'CA',
+              phone:    '(555)555-5555',
+              fax:      '(555)555-6666'},
+      email: 'long@bob.com',
+      customer: 'Longbob Longsen',
+      description: 'Store Purchase',
+      date_of_birth: '1990-10-11',
+      entity_type: 'NaturalPerson',
+      nationality: 'US',
+      shopper_name: {
+        firstName: 'Longbob',
+        lastName: 'Longsen',
+        gender: 'MALE'
+      }
+    }
+
     @avs_credit_card = credit_card('4400000000000008',
                                     :month => 8,
                                     :year => 2018,
@@ -175,13 +200,19 @@ class RemoteBarclaycardSmartpayTest < Test::Unit::TestCase
   end
 
   def test_successful_credit
-    response = @gateway.credit(@amount, @credit_card, @options)
+    response = @gateway.credit(@amount, @credit_card, @options_with_credit_fields)
     assert_success response
   end
 
   def test_failed_credit
     response = @gateway.credit(nil, @declined_card, @options)
     assert_failure response
+  end
+
+  def test_failed_credit_insufficient_validation
+    # This test will fail currently (the credit will succeed), but it should succeed after October 29th
+    # response = @gateway.credit(@amount, @credit_card, @options)
+    # assert_failure response
   end
 
   def test_successful_void
--- a/test/unit/gateways/barclaycard_smartpay_test.rb
+++ b/test/unit/gateways/barclaycard_smartpay_test.rb
@@ -64,6 +64,31 @@ class BarclaycardSmartpayTest < Test::Unit::TestCase
         description: 'Store Purchase'
     }
 
+    @options_with_credit_fields = {
+      order_id: '1',
+      billing_address:       {
+              name:     'Jim Smith',
+              address1: '100 Street',
+              company:  'Widgets Inc',
+              city:     'Ottawa',
+              state:    'ON',
+              zip:      'K1C2N6',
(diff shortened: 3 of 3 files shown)


lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -60,7 +60,13 @@ def credit(money, creditcard, options = {})
         post = payment_request(money, options)
         post[:amount] = amount_hash(money, options[:currency])
         post[:card] = credit_card_hash(creditcard)
-
+        if !options.has_key?(:order_id)
```
bpollack: I couldn't find this pattern elsewhere, but it seemed like a really clean way of handling this client-side. Other reasonable options would be to do this, but without the `requires!` line; or just to unconditionally pass through these attributes, if they exist.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -60,7 +60,13 @@ def credit(money, creditcard, options = {})
...
+          requires!(options, [:date_of_birth, :entity_type, :nationality, :shoper_name])
+          post[:dateOfBirth] = options[:date_of_birth]
+          post[:entityType]   = options[:entity_type]
+          post[:nationality]   = options[:nationality]
+          post[:shopperName]   = options[:shopper_name]
```
bpollack: This can be pulled out of the credit card, but I was a bit ambivalent on whether that actually made sense. Nationality, entity type, and date of birth obviously cannot be pulled out of that field, so having this consistently be its own set of data seemed saner to me.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
dtykocki: I've typically seen the pattern of conditionally adding each option individually vs the use of `requires!`. Example:
```
post[:dateOfBirth] = if options[:date_of_birth]
post[:entityType] = if options[:entity_type]
post[:nationality] =  if options[:nationality]
```
I don't have the full scope of changes in my head so I'm assuming that each item can be sent to the gateway individually instead of requiring the entire group of options.

bpollack: > I don't have the full scope of changes in my head so I'm assuming that each item can be sent to the gateway individually instead of requiring the entire group of options.

In this particular case, the absence of an order ID requires _all four options at once_; none are optional, and none are required for other transaction types—hence this style of `requires!`. (More details in the JIRA ticket this branch is named after, if you're bored.) Whether it should be ActiveMerchant's job to do this, and whether shoving a `requires!` halfway through the method, are still points I'm not clear on. It definitely works, but it's a pattern I didn't see in other adapters I glanced at.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -60,7 +60,13 @@ def credit(money, creditcard, options = {})
         post = payment_request(money, options)
         post[:amount] = amount_hash(money, options[:currency])
         post[:card] = credit_card_hash(creditcard)
-
+        if !options.has_key?(:order_id)
+          requires!(options, [:date_of_birth, :entity_type, :nationality, :shoper_name])
```
curiousepic: Typo in `shopper_name`

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
curiousepic: I think it does as well. Typically I add a `|| creditcard.name` as a fallback in these situations.

curiousepic: Though I see with the `require!` it would never trigger...

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
bpollack: Dangit. Fixing.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
curiousepic: What's the purpose of the `order_id` check?

curiousepic: And yeah it's more common to check the fields individually than with a `requires!` block. We usually let users send whatever they have and leave out what they don't, and get a error message from the gateway. Depends on how the gateway responds to missing or blank fields.

bpollack: The `order_id` check is because those fields aren't required if it's present. The alternative to doing it this way would be to say that it's not ActiveMerchant's problem to tell you you're missing fields, at which point this method would instead become:

```
def credit(money, creditcard, options = {})
  post = payment_request(money, options)
  post[:amount] = amount_hash(money, options[:cgurrency])
  post[:dateOfBirth] = options[:date_of_birth] if options[:date_of_birth]
  post[:entityType]  = options[:entity_type] if options[:entity_type]
  post[:nationality] = options[:nationality] if options[:nationality]
  post[:shopperName] = options[:shopper_name] if options[:shopper_name]
  commit('refundWithData', post)
end
```

This seemed harsher to the user and doesn't save us much work, but it permits greater flexibility on the part of the gateway.

bpollack:
Okay, all changes we talked about @curiousepic are in here. Code does much less and doesn't confuse `:order_id` for a pspReference anymore.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -60,6 +60,10 @@ def credit(money, creditcard, options = {})
         post = payment_request(money, options)
         post[:amount] = amount_hash(money, options[:currency])
         post[:card] = credit_card_hash(creditcard)
+        post[:dateOfBirth] = options[:date_of_birth] if options[:date_of_birth]
+        post[:entityType]  = options[:entity_type] if options[:entity_type]
+        post[:nationality] = options[:nationality] if options[:nationality]
+        post[:shopperName] = options[:shopper_name] if options[:shopper_name]
```
curiousepic: Now without the `requires!` we can at least be helpful with a fallback like `|| creditcard.name`

curiousepic:
Otherwise looks good, this is probably the best we can do for now.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
bpollack: I wrote this, but as I started updating the tests to handle it, I realized I disagree with the feature: for us to do this, you'd have to 1) issue a credit, 2) successfully supply the date of birth, entity type, and nationality, but 3) forget shopper name. That's already an edge case of an edge case, but yoinking data from the credit card fields has a fun side-effect that the user may get the credit rejected for _having the wrong shopper name_, and then will be trying to figure out how that's even possible/what the shopper name was in the first place.

I thus think this is probably the best path forward as-is.

curiousepic: I think that's fair 👍

PR #2632, opened 2017-10-26
Shallow copy the options hash for Barclays tests

bpollack:
The intent in the tests was clearly to create a *distinct* options array, but what actually happened is that the original options array was modified. This patch causes the code to do the intended behavior. (Interestingly, this doesn't alter any test passes or failures, because having extra data doesn't upset either the unit tests or remote tests.)


deedeelavinder:
👍  Even though the tests aren't altered, including the test results in the commit maintains consistency and could be helpful in the future.

PR #2633, opened 2017-10-27
Cyber Source: Correctly identifies 'subscriptionID' for store

deedeelavinder:
'subscriptionID' is required to authorize store. Previously 'requestID'
was used to populate the 'subscriptionID'. In the sandbox, they yield
identical values. In production, however, they are two different values.
This correctly captures the 'subscriptionID' separately from the
'requestID'.

Remote Tests:
(Includes 2 failing tests on master regarding pinless_debit_cards. The
code for pinless_debit_cards does not currently run.)

39 tests, 181 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
94.8718% passed

Unit Tests:
44 tests, 212 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -618,7 +618,7 @@ module ActiveMerchant #:nodoc:
 
         xml.tag! 'recurringSubscriptionInfo' do
           if reference
-            _, subscription_id, _ = reference.split(";")
+            subscription_id = reference.split(";")[6]
             xml.tag! 'subscriptionID',  subscription_id
           end
 
@@ -709,7 +709,7 @@ module ActiveMerchant #:nodoc:
         success = response[:decision] == "ACCEPT"
         message = response[:message]
 
-        authorization = success ? [ options[:order_id], response[:requestID], response[:requestToken], action, amount, options[:currency]].compact.join(";") : nil
+        authorization = success ? authorization_from(response, action, amount, options) : nil
 
         Response.new(success, message, response,
           :test => test?,
@@ -758,6 +758,11 @@ module ActiveMerchant #:nodoc:
       def reason_message(reason_code)
         return if reason_code.blank?
         @@response_codes[:"r#{reason_code}"]
+      end
+
+      def authorization_from(response, action, amount, options)
+        [options[:order_id], response[:requestID], response[:requestToken], action, amount,
+         options[:currency], response[:subscriptionID]].join(";")
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -618,7 +618,7 @@ def add_subscription(xml, options, reference = nil)
 
         xml.tag! 'recurringSubscriptionInfo' do
           if reference
-            _, subscription_id, _ = reference.split(";")
+            _, subscription_id = reference.split(";")
```
curiousepic: After playing around with it in irb to confirm, I believe this still assigns the second value in the string (ReferenceID). You'll probably need to specify the number of underscores matching the number of values that precede the SubscriptionID on the end.

bpollack: Or just do `subscription_id = reference.split(';')[6]` (IIRC), which I think reads a bit more clearly anyway. (Although TIL Ruby can do destructuring binds, which I hadn't known previously.)

deedeelavinder: 👍

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -618,7 +618,7 @@ def add_subscription(xml, options, reference = nil)
 
         xml.tag! 'recurringSubscriptionInfo' do
           if reference
-            _, subscription_id, _ = reference.split(";")
+            subscription_id = reference.split(";")[-1]
```
curiousepic: Hrrm. This will break if we have to add yet another thing to the end of the auth. A a positive reference might be better (`[6]`).

deedeelavinder: This doesn't work because not all of the options are passed in every time. The reference can have a variable number of objects.

curiousepic: Are you sure? The semicolons should still be added.

Oh I see why, the auth string is being `compact`ed first. https://github.com/deedeelavinder/active_merchant/blob/8afd6a18a2118b85a871e56c534031c258a6ac72/lib/active_merchant/billing/gateways/cyber_source.rb#L712
That should probably not be done.  It doesn't look like that would impact any of the other `split`s, but it is worth double checking.

It might be nice to pull out it out into its own `authorization_from` method like other gateways, too.

deedeelavinder: @curiousepic 👍  Added an `authorization_from` method with no `compact`.

curiousepic:
@deedeelavinder Well, Travis says lots of unit tests are failing since it they aren't expecting the additional semicolon?

deedeelavinder:
@curiousepic GAH -- added now.

PR #2634, opened 2017-10-30
PayU Latam: Four small updates

deedeelavinder:
1) Adds `partnerId`.
2) Ensures phone_number is passed to 'buyer.contactPhone' and prefers
billing phone over shipping phone.
3) Sends empty field for 'ip_address' when none supplied.
4) Ensures 'creditCard.secutiryCode' node is not included when there is
no 'cvv'.

Remote Tests:
(Includes 5 failing tests on master: 2 'INTERNAL_PAYMENT_PROVIDER_ERROR'
and 3 'DECLINED_TEST_MODE_NOT_ALLOWED'.)

18 tests, 45 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
72.2222% passed

Unit Tests:
24 tests, 94 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -141,7 +141,7 @@ module ActiveMerchant #:nodoc:
         transaction = {}
         transaction[:paymentCountry] = @options[:payment_country] || (options[:billing_address][:country] if options[:billing_address])
         transaction[:type] = type
-        transaction[:ipAddress] = options[:ip] if options[:ip]
+        transaction[:ipAddress] = options[:ip] || ''
         transaction[:userAgent] = options[:user_agent] if options[:user_agent]
         transaction[:cookie] = options[:cookie] if options[:cookie]
         transaction[:deviceSessionId] = options[:device_session_id] if options[:device_session_id]
@@ -151,6 +151,7 @@ module ActiveMerchant #:nodoc:
       def add_order(post, options)
         order = {}
         order[:accountId] = @options[:account_id]
+        order[:partnerId] = options[:partnerId] if options[:partnerId]
         order[:referenceCode] = options[:order_id] || generate_unique_id
         order[:description] = options[:description] || 'unspecified'
         order[:language] = 'en'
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -192,15 +193,15 @@ def add_buyer(post, payment_method, options)
           buyer[:dniType] = buyer_hash[:dni_type]
           buyer[:cnpj] = buyer_hash[:cnpj] if options[:payment_country] == 'BR'
           buyer[:emailAddress] = buyer_hash[:email]
-          buyer[:contactPhone] = options[:shipping_address][:phone] if options[:shipping_address]
+          buyer[:contactPhone] = (options[:billing_address][:phone] if options[:billing_address]) || (options[:shipping_address][:phone] if options[:shipping_address])
```
bpollack: I don't _think_ this is doing what you intend; the result of this line as-is will be `buyer[:contactPhone]` set to `options[:billing_address][:phone]`, `options[:shipping_address][:phone]`, or `nil`, whereas I suspect you meant to leave it unset entirely if neither is available.

I'm not aware of a clever way to short-circuit this in our version of Ruby. The closest I could get would be something like

    val = (options[:billing_address] || options[:shipping_address] || {})[:phone]
    order[:contactPhone] = val if val

but I'm not sure that level of clever is actually wise.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -192,15 +193,15 @@ def add_buyer(post, payment_method, options)
...
           buyer[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
         else
           buyer[:fullName] = payment_method.name.strip
           buyer[:dniNumber] = options[:dni_number]
           buyer[:dniType] = options[:dni_type]
           buyer[:cnpj] = options[:cnpj] if options[:payment_country] == 'BR'
           buyer[:emailAddress] = options[:email]
-          buyer[:contactPhone] = (options[:shipping_address][:phone] if options[:shipping_address]) || (options[:billing_address][:phone] if options[:billing_address])
+          buyer[:contactPhone] = (options[:billing_address][:phone] if options[:billing_address]) || (options[:shipping_address][:phone] if options[:shipping_address])
```
bpollack: Ditto. If you intend to create `:contacTPhone` unconditionally and are okay with it being `nil`, this is fine, but I don't think that's what you intended.

lib/active_merchant/billing/gateways/payu_latam.rb
curiousepic: What she has is actually what is desired, at least according to the ticket. Though looking at the ticket I'm getting a sinking sense that my summary of what should be done is not actually what the customer wants. I'll come back to this.

bpollack: I didn't see the attached ticket, so that may well be the case. I'll yoink my reject.

curiousepic: In fact, after reviewing the ticket again this is in fact what is basically desired, however it is likely best to send an empty string in the case of neither field being available.

curiousepic:
There are many nuances to these items but after reviewing the ticket and attached info I believe this is all correct, aside from the preference of the empty string for `contactPhone`.

PR #2637, opened 2017-10-31
Fix Changelog entry listed in the wrong version

bpollack:
PayU changes are in the not-yet-released 1.75 branch, not 1.74.


bpollack:
This is a *really* tiny bug that crept in with #2634, but it'll keep the changelog clean.

bpollack:
Per discussion, going to remove this and let @curiousepic take care of this in a separate PR.

PR #2640, opened 2017-11-02
Worldpay US: add Worldpay US (Backup) Gateway

bpollack:
Worldpay requires end users to fail over to a different URL if the primary is offline. Rather than try to manage this from within AM, outsource the pain to the user.

Furthers ENRG-6747


bpollack:
Throwing @davidsantoso on this review so he can justify the insanity here, and the immediate insanity that's going to follow elsewhere.

bpollack:
Oh right, one other note: one test in this fails, but the same test fails on Worldpay US proper. Fix on the test isn't obvious, and it didn't seem worth holding up the PR, but I wanted to flag that.

bpollack:
@davidsantoso, @curiousepic and I had a real-life discussion; going to just go a flag route, not an entirely separate adapter route.

PR #2643, opened 2017-11-06
Only send currency and billing_details for auths and sales for Netbanx

AnotherJoSmith:
Paysafe does not allow you to send this information for captures and
purchases. Therefore we shouldn't include them even if they're passed in
the options. The remote tests would previously fail if you included the
options field for captures or refunds, but they now pass.

### DIFF
--- a/lib/active_merchant/billing/gateways/netbanx.rb
+++ b/lib/active_merchant/billing/gateways/netbanx.rb
@@ -31,7 +31,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_invoice(post, money, options)
         add_settle_with_auth(post)
-        add_payment(post, payment)
+        add_payment(post, payment, options)
 
         commit(:post, 'auths', post)
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/netbanx.rb
```diff
@@ -132,13 +132,7 @@ def add_credit_card(post, credit_card, options = {})
 
       def add_invoice(post, money, options)
         post[:amount] = amount(money)
-        post[:currencyCode] = options[:currency] if options[:currency]
```
elfassy: do we also need the currency code in the `add_invoice` in addition to the `add_payment`?

AnotherJoSmith: @elfassy No, it only needs to be in add_payment

AnotherJoSmith:
@elfassy I did run the remote tests and they're all green.

PR #2644, opened 2017-11-06
Revert "Fixes basic auth for netbanx by sending the account_number and api_key"

AnotherJoSmith:
This reverts commit a838219a497144abc2bd72f16729e9d9b07a495c. The
previous change needed to be reverted because I had misunderstood that
the account number and username are different values. By reverting the
last PR, the api_key needs to be in the basic auth format already
(`<username>:<password>`).

### DIFF
--- a/lib/active_merchant/billing/gateways/netbanx.rb
+++ b/lib/active_merchant/billing/gateways/netbanx.rb
@@ -238,13 +238,9 @@ module ActiveMerchant #:nodoc:
         {
           'Accept'        => 'application/json',
           'Content-type'  => 'application/json',
-          'Authorization' => "Basic #{basic_auth}",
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/netbanx.rb
```diff
@@ -241,15 +241,11 @@ def headers
         {
           'Accept'        => 'application/json',
           'Content-type'  => 'application/json',
-          'Authorization' => "Basic #{basic_auth}",
+          'Authorization' => "Basic #{Base64.strict_encode64(@options[:api_key].to_s).strip}",
```
bdewater: We can remove `strip` since base64 will not output spaces.

AnotherJoSmith: 👍

PR #2645, opened 2017-11-07
Worldpay US: Allowing switching to backup gateway per-transaction

bpollack:
Previously, in 42e69f44390d, we allowed creating a Worldpay US gateway that targeted the backup URL. This commit changes that to instead allow specifying the URL per-transaction.

Note that one remote test fails, but it does prior to this commit (and to this entire chain of commits) as well.

Furthers ENRG-6747

Unit tests:
16 tests, 71 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests:
14 tests, 40 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92.8571% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay_us.rb
+++ b/lib/active_merchant/billing/gateways/worldpay_us.rb
@@ -28,7 +28,7 @@ module ActiveMerchant #:nodoc:
         add_payment_method(post, payment_method)
         add_customer_data(post, options)
 
-        commit('purchase', post)
+        commit('purchase', options, post)
       end
 
       def authorize(money, payment, options={})
@@ -37,7 +37,7 @@ module ActiveMerchant #:nodoc:
         add_credit_card(post, payment)
         add_customer_data(post, options)
 
-        commit('authorize', post)
+        commit('authorize', options, post)
       end
 
       def capture(amount, authorization, options={})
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/worldpay_us.rb
```diff
@@ -169,7 +169,7 @@ def parse(xml)
         "void" => "ns_void",
       }
 
-      def commit(action, post)
+      def commit(action, url, post)
```
curiousepic: I feel like it's a slightly more common pattern to pass options straight to commit.  Then you can call the `url` method further downstream.

bpollack: Fair enough, I'll switch this around. Just felt odd to me, since only that one value was needed.

lib/active_merchant/billing/gateways/worldpay_us.rb
```diff
@@ -74,8 +74,8 @@ def verify(credit_card, options={})
 
       private
 
-      def url
-        @options[:use_backup_url] ? self.backup_url : self.live_url
+      def url(options)
+        options[:use_backup_url] ? self.backup_url : self.live_url
```
davidsantoso: In the past we've coerced the type here to be something like-

```
options[:use_backup_url].to_s == "true" ? self.backup_url : self.live_url
```

so we're a little more explicit about what we're expecting vs. `"false"` also causing this to trigger. Just a thought!

bpollack: I'll change that as well to be consistent. Thanks!

PR #2646, opened 2017-11-07
Stop making remote requests for payeezy store unit tests

AnotherJoSmith:
Payeezy uses the :ssl_request method to store CCs.

https://github.com/activemerchant/active_merchant/blob/23c38dd7f5758360fca02b6099254b31211d7573/lib/active_merchant/billing/gateways/payeezy.rb#L276-L285

The problem is, the `stub_comms` method defaults to stubbing the `ssl_post` method.

https://github.com/activemerchant/active_merchant/blob/23c38dd7f5758360fca02b6099254b31211d7573/test/comm_stub.rb#L43-L46

This means that these tests were making remote requests every single time they run, causing flakyness in CI runs.

screen shot 2017-11-07 at 5 12 40 pm

### DIFF
--- a/test/unit/gateways/payeezy_test.rb
+++ b/test/unit/gateways/payeezy_test.rb
@@ -58,7 +58,7 @@ class PayeezyGateway < Test::Unit::TestCase
   end
 
   def test_successful_store
-    response = stub_comms do
+    response = stub_comms(@gateway, :ssl_request) do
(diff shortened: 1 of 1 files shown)


AnotherJoSmith:
As a side comment the whole CommStub class seems to be pretty bad. I think we should eventually get rid of it and use something like WebMock instead.

PR #2650, opened 2017-11-13
Checkout V2: Allow AVS and CVV for Authorizations

deedeelavinder:
AVS and CVV result details will also now come through on authorizations.

Remote Tests:
24 tests, 57 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Unit Tests:
19 tests, 80 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -25,16 +25,7 @@ module ActiveMerchant #:nodoc:
         merged_params = multi.responses.map { |r| r.params }.reduce({}, :merge)
         succeeded = success_from(merged_params)
 
-        Response.new(
-          succeeded,
-          message_from(succeeded, merged_params),
-          merged_params,
-          authorization: authorization_from(merged_params),
-          avs_result: avs_result(:purchase, succeeded, merged_params),
-          cvv_result: cvv_result(:purchase, succeeded, merged_params),
-          error_code: error_code_from(succeeded, merged_params),
-          test: test?
-        )
+        response(:purchase, succeeded, merged_params)
       end
 
       def authorize(amount, payment_method, options={})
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -132,6 +123,11 @@ def commit(action, post, authorization = nil)
         end
 
         succeeded = success_from(response)
+
+        response(action, succeeded, response)
+      end
+
+      def response(action, succeeded, response)
```
bpollack: Nice refactoring! This made the code quite a bit easier to read.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -163,16 +160,16 @@ def base_url
       end
 
       def avs_result(action, succeeded, response)
-        if succeeded
-          action == :purchase ? AVSResult.new(code: response["card"]["avsCheck"]) : nil
+        if succeeded && action == :purchase || :authorize
```
bpollack: I am pretty sure you meant `action == :purchased || action == :authorized`; as-is, `:authorized` is truthy, so this condition is always true.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -163,16 +160,16 @@ def base_url
...
+          AVSResult.new(code: response['card']['avsCheck']) if response['card'] && response['card']['avsCheck']
```
bpollack: This is personal preference, but this condition is getting sufficiently complex I'd generally prefer it not to be a guard clause.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -163,16 +160,16 @@ def base_url
...
         else
           nil
         end
       end
 
       def cvv_result(action, succeeded, response)
-        if succeeded
-          action == :purchase ? CVVResult.new(response["card"]["cvvCheck"]) : nil
+        if succeeded && action == :purchase || :authorize
```
bpollack: Same issue as line [the earlier line](https://github.com/activemerchant/active_merchant/pull/2650/files#diff-390e4ab1e508727fef10daef8ff96d35R163)

lib/active_merchant/billing/gateways/checkout_v2.rb
deedeelavinder: ugh, yes! Thanks for catching.

lib/active_merchant/billing/gateways/checkout_v2.rb
deedeelavinder: Agreed -- removed.

PR #2652, opened 2017-11-15
Credorax: Flag for tokenized card data

curiousepic:
The a9 field should be sent with a value of 8 for visa and mastercard if
the card data is stored (anywhere) as tokenized data. This will send it
if tokenized_card_data is flagged in options.


curiousepic:
Actually this is probably best as an open field that passes any value through. Revising.

PR #2653, opened 2017-11-15
Credorax: Pass Transaction Type field

curiousepic:
The a9 ("Transaction Type") field is required to be sent with a value of
8 for visa and mastercard if the card data is stored (anywhere) as
tokenized data. This commit allows passing the a9 value as an option.

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -129,6 +129,7 @@ module ActiveMerchant #:nodoc:
         add_email(post, options)
         add_3d_secure(post, options)
         add_echo(post, options)
+        add_transaction_type(post, options)
 
         commit(:purchase, post)
       end
@@ -141,6 +142,7 @@ module ActiveMerchant #:nodoc:
         add_email(post, options)
         add_3d_secure(post, options)
         add_echo(post, options)
+        add_transaction_type(post, options)
 
         commit(:authorize, post)
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -129,6 +129,7 @@ def purchase(amount, payment_method, options={})
         add_email(post, options)
         add_3d_secure(post, options)
         add_echo(post, options)
+        post[:a9] = options[:transaction_type] if options[:transaction_type]
```
bpollack: I know it's one line, but you repeat this exact line three times, which is three changes to have a typo, and there's already precedents of `add_*`s in this gateway to ameliorate this exact issue (see `add_echo` or `add_email`). I'd thus really prefer this being `add_a9` or similar.

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -141,6 +142,7 @@ def authorize(amount, payment_method, options={})
...
```
bpollack: Second repetition.

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -182,6 +184,7 @@ def credit(amount, payment_method, options={})
         add_customer_data(post, options)
         add_email(post, options)
         add_echo(post, options)
+        post[:a9] = options[:transaction_type] if options[:transaction_type]
```
bpollack: Third repetition.

curiousepic:
Thanks for catching that 👍  Updated.

PR #2655, opened 2017-11-21
Always override ECI value for Apple Pay transactions with Discover

jasonwebster:
I am so sorry.

Discover has a hard requirement that the ECI value for all Apple Pay
transactions be set to exactly 4, regardless of what is or is not
provided in the Apple PKPaymentToken. This is under subject of merchant
fees and penalties if not, according to people I've talked to at
Discover.

The Discover specification reads:

https://www.dropbox.com/s/fnmofdfnsc37wuj/Screenshot%202017-11-21%2017.43.16.png?dl=0

Yes, that's mostly meaningless, but "field 61 position 9" is what
gateways map the incoming ECI value to. I've confirmed this information
with contacts at FirstData, which said basically "yeah, this is a
Discover requirement, and you must send it in differently, we won't do
it for you."

This was the only way that I could think of that would fix this issue
for all gateways supported by Active Merchant.

### DIFF
--- a/lib/active_merchant/billing/network_tokenization_credit_card.rb
+++ b/lib/active_merchant/billing/network_tokenization_credit_card.rb
@@ -14,8 +14,8 @@ module ActiveMerchant #:nodoc:
       self.require_verification_value = false
       self.require_name = false
 
-      attr_accessor :payment_cryptogram, :eci, :transaction_id
-      attr_writer :source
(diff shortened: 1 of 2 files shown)


jasonwebster:
Closing this in favour of less gross workarounds. In the intermediary, I'm going to encourage this to be fixed on the application side, and in individual gateways. This just feels so wrong--this object should be a logic-less value object--simply a struct of card information.

PR #2658, opened 2017-11-28
Cyber Source: Correctly handle voiding a purchase

eric-pigeon:
treat voiding a purchase as voiding a capture instead of voiding an
auth


bpollack:
Hi @eric-pigeon ! Would you mind linking to the docs that indicate the existing behavior is incorrect/insufficient? If so, I'm happy to merge this.

PR #2659, opened 2017-11-28
Implement CardProcess Gateway

bpollack:
Implement Cardprocess Gateway
    
Furthers ENRG-6967

### DIFF
--- a/lib/active_merchant/billing/gateways/cardprocess.rb
+++ b/lib/active_merchant/billing/gateways/cardprocess.rb
@@ -1,0 +1,254 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CardprocessGateway < Gateway
+      self.test_url = 'https://test.vr-pay-ecommerce.de/v1/payments'
+      self.live_url = 'https://vr-pay-ecommerce.de/v1/payments'
+
+      self.supported_countries = %w[ BE BG CZ DK DE EE IE ES FR HR IT CY LV LT LU
+                                     MT HU NL AT PL PT RO SI SK FI SE GB IS LI NO
+                                     CH ME MK AL RS TR BA ]
+      self.default_currency = 'EUR'
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :jcb]
+
+      self.homepage_url = 'https://vr-pay-ecommerce.docs.oppwa.com/'
+      self.display_name = 'CardProcess VR-Pay'
+      self.money_format = :dollars
+
+      STANDARD_ERROR_CODE_MAPPING = {}
+
+      # Creates a new CardProcess Gateway
+      #
+      # The gateway requires a valid login, password, and entity ID
+      # to be passed in the +options+ hash.
+      #
+      # === Options
+      #
+      # * <tt>:user_id</tt> -- The CardProcess user ID
+      # * <tt>:password</tt> -- The CardProcess password
+      # * <tt>:entity_id</tt> -- The CardProcess channel or entity ID for any transactions
+      def initialize(options={})
+        requires!(options, :user_id, :password, :entity_id)
+        super
+        # This variable exists purely to allow remote tests to force error codes;
+        # the lack of a setter despite its usage is intentional.
+        @test_options = {}
+      end
+
+      def purchase(money, payment, options = {})
+        post = {}
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/cardprocess.rb
```diff
@@ -0,0 +1,252 @@
...
+      #
+      # * <tt>:user_id</tt> -- The CardProcess user ID
+      # * <tt>:password</tt> -- The CardProcess password
+      # * <tt>:entity_id</tt> -- The CardProcess channel or entity ID for any transactions
+      def initialize(options={})
+        requires!(options, :user_id, :password, :entity_id)
+        super
+        # This variable exists purely to allow remote tests to force error codes;
+        # the lack of a setter despite its usage is intentional.
+        @extra_options = {}
```
curiousepic: Maybe name it `@test_options`?

lib/active_merchant/billing/gateways/cardprocess.rb
```diff
@@ -0,0 +1,252 @@
...
+      def void(authorization, _options = {})
+        post = {
+          id: authorization
+        }
+        commit('RV', post)
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run do |r|
+          r.process { authorize(100, credit_card, options) }
```
curiousepic: did we check if they allow auths for 0?

lib/active_merchant/billing/gateways/cardprocess.rb
```diff
@@ -0,0 +1,252 @@
...
+      end
+
+      def purchase(money, payment, options = {})
+        post = {}
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+
+        commit('DB', post)
```
curiousepic: I might consider mapping these action codes to readable aliases.

lib/active_merchant/billing/gateways/cardprocess.rb
bpollack: Ooh, good idea! I'll change this, although I'll still leave the comment.

lib/active_merchant/billing/gateways/cardprocess.rb
bpollack: I actually originally had that, but it felt a bit weird to me, since each alias was used only once, in the appropriate function, at which point it didn't seem to be getting much. It looked like this, roughly:

```ruby
ACTIONS = {
  purchase: 'DB',
  ...
}

def purchase(...)
  ...
  commit(ACTIONS[:purchase], post)
end
```

I don't have strong opinions on this. If you do, or if you have a different spin that feels less redundant, I'm completely happy to go back to this style.

lib/active_merchant/billing/gateways/cardprocess.rb
bpollack: I didn't, but I will.

bpollack: @curiousepic Okay, they do require a non-zero value, and `1.00` seems to be the lowest value for at least some of the processors, so I think this has to stay as-is, sadly.

lib/active_merchant/billing/gateways/cardprocess.rb
curiousepic: Mostly thinking that often when adding options later, we need to condition on actions, and it would be less confusing then, but NBD.

test/unit/gateways/cardprocess_test.rb
```diff
@@ -0,0 +1,280 @@
...
+
+    assert_equal '8a82944a5fe82704016002caa7cd1513', response.authorization
+    assert_equal 'CP', response.params['paymentType']
+    assert response.test?
+  end
+
+  def test_failed_capture
+    @gateway.expects(:ssl_post).returns(failed_capture_response)
+
+    response = @gateway.capture(@money, '123', @options)
```
deedeelavinder: Looks like `@amount` is what is initialized. `@money` is also used on line 88.

bpollack: Thanks for the catch, @deedeelavinder ! This isn't actually why unit tests were failing (that was because XK isn't a country code AM knows about), but it should be fixed anyway. Updated push coming in a second.

deedeelavinder: @bpollack I didn't dig into why the unit tests were throwing errors -- just noticed the discrepancy between `@money` and `@amount`. Glad the tests are easy to fix.

PR #2661, opened 2017-11-30
Safe Charge: Provisions adapter for 3DS purchases

deedeelavinder:
Safe Charge has implemented a system by which they determine whether
purchases should be routed through the 3DS flow or not. If the merchant
is not approved for 3DS, transactions will occur as before, outside of
the 3DS flow. If a merchant is approved for 3DS, Sage Charge will
evaluate the card for enrollment status. If the card is not enrolled, it
will be processed as a 'regular' transaction. If the card is enrolled,
the Safe Charge response will include the 'xid', 'ascurl', and the
'pareq'.

Remote Tests:
21 tests, 62 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Unit Tests:
17 tests, 76 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/safe_charge.rb
+++ b/lib/active_merchant/billing/gateways/safe_charge.rb
@@ -22,7 +22,8 @@ module ActiveMerchant #:nodoc:
 
       def purchase(money, payment, options={})
         post = {}
-        add_transaction_data("Sale", post, money, options)
+        post[:sg_APIType] = 1
+        add_transaction_data("Sale3D", post, money, options)
         add_payment(post, payment)
         add_customer_details(post, payment, options)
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/safe_charge.rb
```diff
@@ -22,7 +22,8 @@ def initialize(options={})
 
       def purchase(money, payment, options={})
         post = {}
-        add_transaction_data("Sale", post, money, options)
+        post[:sg_APIType] = 1
+        add_transaction_data("Sale3D", post, money, options)
```
curiousepic: I think we may want to have the option to flag txns to go through their 3DS system or not at all. It looks like this always defaults to attempting 3DS?

curiousepic: Also, Authorizations should be available for attempting 3DS in addition to Purchases, if that is possible on the gateway end.

deedeelavinder: @curiousepic I think I mentioned their internal assessment: if the merchant is not approved for 3DS, Safe Charge will default to `Sale`. And they do not allow for 3DS authorizations, only purchases.

curiousepic: Chatted offline about this and it looks like this is the way to go, since the determination of "3DS flow or not" really occurs at the merchant account level. 👍

Issue #2657, opened 2017-11-23
cacert.pem outdated; NAB Transact failing


guaguasi:
Using `activemerchant 1.75.0` with `ruby 2.4.2p198` and `active_support 5.1.4` on `CentOS Linux release 7.4.1708`

NAB Transact updated their gateway certificate Nov 23rd, 10:47 AM AEDT (Sydney time) causing every subsequent transaction to fail with the very detailed `The SSL connection to the remote server could not be established` error message. Turns out it was a peer verification error. OpenSSL via the CLI connected just fine.

I ended up replacing `lib/certs/cacert.pem` in the gem source with `/etc/pki/tls/cert.pem`, which appears to be the same file as `/etc/pki/tls/certs/ca-bundle.crt`, and the transactions started to flow once more.

Am I bonkers in doing this? Is this a valid issue?

ralleman:
This is good to know because I'm seeing the same thing. What's the correct resolution for this?

bpollack:
Merging into #2901.

PR #2614, opened 2017-10-10
Entrust Root CA certificates for Barclays EPDQ Extra Plus Gateway

pacso:
As described in #2613 there are missing root certificates when attempting to validate the SSL certificate for `payments.epdq.co.uk`.

This PR adds the missing certificates.

You can verify the added certificates on the following page: https://www.entrust.com/get-support/ssl-certificate-support/root-certificate-downloads/

PR #2662, opened 2017-12-01
Regenerated cacert.pem

guaguasi:
cacert.pem is extremely outdated, regenerated it using libcurl's `mk-ca-bundle.pl`. Should address #2657, #2614, and similar.


rwdaigle:
There's been some discussion amongst @activemerchant/spreedly that we should remove the certs from the project itself all together, instead relying on the certificate management of each environment as is the default for almost all other projects.

Curious if @activemerchant/developers are using AM's certs or are already overriding with system certs?

guaguasi:
Copying the system's ca-bundle was how I initially fixed my issue (#2657) before finding out how to regenerate the mozilla bundle file. I'm currently assigning the system bundle to the `CA_FILE` constant as a stop gap which seems to be working but is throwing a warning. It appears that currently it's still using AM's certs.

Personally, I feel it should use the environment's bundle by default with an up to date `cacert.pem`  file as a fallback, as well as a proper way to specify a custom location for a bundle.

guaguasi:
@bpollack I can understand that. I literally just downloaded `certdata.txt` file the original `cacert.pem` was generated with and used `mk-ca-bundle.pl` to generate the `cacert.pem` in this PR. It should contain all the current certificates, right?

varyonic:
I'm not a member of @activemerchant/developers and I was caught by surprise when the Digicert Global Root G2 cert expired yesterday, used to verify sermepa.es (Redsys). Took me a while first to update our old Ubuntu ca-certificates package and then to overwrite cacert.pem currently provided.  This is a small shop so not much attention paid here to managing certs at the Rails environment level.

bpollack:
I'm going to close this for now; we're trying to figure out what to do internally, and I'm tracking various proposals in #2901.

PR #2663, opened 2017-12-01
PayU Latam: Require payment_country on initialize

curiousepic:
Also cleans up tests, though declined and pending test cards are now
throwing a new unexpected error, unrelated to this change.

Remote:
18 tests, 47 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
83.3333% passed

Unit:
22 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -29,9 +29,8 @@ module ActiveMerchant #:nodoc:
       }
 
       def initialize(options={})
-        requires!(options, :merchant_id, :account_id, :api_login, :api_key)
+        requires!(options, :merchant_id, :account_id, :api_login, :api_key, :payment_country)
         super
-        @options[:payment_country] ||= options[:payment_country] if options[:payment_country]
       end
 
       def purchase(amount, payment_method, options={})
@@ -139,7 +138,7 @@ module ActiveMerchant #:nodoc:
 
       def add_transaction_elements(post, type, options)
         transaction = {}
-        transaction[:paymentCountry] = @options[:payment_country] || (options[:billing_address][:country] if options[:billing_address])
+        transaction[:paymentCountry] = @options[:payment_country]
         transaction[:type] = type
         transaction[:ipAddress] = options[:ip] || ''
         transaction[:userAgent] = options[:user_agent] if options[:user_agent]
@@ -167,7 +166,7 @@ module ActiveMerchant #:nodoc:
         payer[:dniNumber] = options[:dni_number] if options[:dni_number]
         payer[:dniType] = options[:dni_type] if options[:dni_type]
         payer[:emailAddress] = options[:email] if options[:email]
-        payer[:birthdate] = options[:birth_date] if options[:birth_date] && options[:payment_country] == 'MX'
+        payer[:birthdate] = options[:birth_date] if options[:birth_date] && @options[:payment_country] == 'MX'
         payer[:billingAddress] = billing_address_fields(options)
         post[:transaction][:payer] = payer
       end
@@ -180,7 +179,7 @@ module ActiveMerchant #:nodoc:
         billing_address[:city] = address[:city]
         billing_address[:state] = address[:state]
         billing_address[:country] = address[:country]
-        billing_address[:postalCode] = address[:zip] if options[:payment_country] == 'MX'
+        billing_address[:postalCode] = address[:zip] if @options[:payment_country] == 'MX'
         billing_address[:phone] = address[:phone]
         billing_address
       end
@@ -191,7 +190,7 @@ module ActiveMerchant #:nodoc:
           buyer[:fullName] = buyer_hash[:name]
           buyer[:dniNumber] = buyer_hash[:dni_number]
           buyer[:dniType] = buyer_hash[:dni_type]
-          buyer[:cnpj] = buyer_hash[:cnpj] if options[:payment_country] == 'BR'
+          buyer[:cnpj] = buyer_hash[:cnpj] if @options[:payment_country] == 'BR'
           buyer[:emailAddress] = buyer_hash[:email]
           buyer[:contactPhone] = (options[:billing_address][:phone] if options[:billing_address]) || (options[:shipping_address][:phone] if options[:shipping_address]) || ''
           buyer[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
@@ -199,7 +198,7 @@ module ActiveMerchant #:nodoc:
           buyer[:fullName] = payment_method.name.strip
           buyer[:dniNumber] = options[:dni_number]
           buyer[:dniType] = options[:dni_type]
-          buyer[:cnpj] = options[:cnpj] if options[:payment_country] == 'BR'
+          buyer[:cnpj] = options[:cnpj] if @options[:payment_country] == 'BR'
           buyer[:emailAddress] = options[:email]
           buyer[:contactPhone] = (options[:billing_address][:phone] if options[:billing_address]) || (options[:shipping_address][:phone] if options[:shipping_address]) || ''
           buyer[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
@@ -235,8 +234,8 @@ module ActiveMerchant #:nodoc:
 
         additional_values = {}
         additional_values[:TX_VALUE] = tx_value
-        additional_values[:TX_TAX] = tx_tax if options[:payment_country] == 'CO'
-        additional_values[:TX_TAX_RETURN_BASE] = tx_tax_return_base if options[:payment_country] == 'CO'
+        additional_values[:TX_TAX] = tx_tax if @options[:payment_country] == 'CO'
+        additional_values[:TX_TAX_RETURN_BASE] = tx_tax_return_base if @options[:payment_country] == 'CO'
 
         post[:transaction][:order][:additionalValues] = additional_values
       end
--- a/test/unit/gateways/payu_latam_test.rb
+++ b/test/unit/gateways/payu_latam_test.rb
@@ -4,7 +4,7 @@ class PayuLatamTest < Test::Unit::TestCase
   include CommStub
 
   def setup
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -167,7 +166,7 @@ def add_payer(post, payment_method, options)
         payer[:dniNumber] = options[:dni_number] if options[:dni_number]
         payer[:dniType] = options[:dni_type] if options[:dni_type]
         payer[:emailAddress] = options[:email] if options[:email]
-        payer[:birthdate] = options[:birth_date] if options[:birth_date] && options[:payment_country] == 'MX'
+        payer[:birthdate] = options[:birth_date] if options[:birth_date] && @options[:payment_country]  == 'MX'
```
bpollack: Extra space after `]` and before `==`.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -180,7 +179,7 @@ def billing_address_fields(options)
         billing_address[:city] = address[:city]
         billing_address[:state] = address[:state]
         billing_address[:country] = address[:country]
-        billing_address[:postalCode] = address[:zip] if options[:payment_country] == 'MX'
+        billing_address[:postalCode] = address[:zip] if @options[:payment_country]  == 'MX'
```
bpollack: Extra space after `]` and before `==`.

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -191,15 +190,15 @@ def add_buyer(post, payment_method, options)
           buyer[:fullName] = buyer_hash[:name]
           buyer[:dniNumber] = buyer_hash[:dni_number]
           buyer[:dniType] = buyer_hash[:dni_type]
-          buyer[:cnpj] = buyer_hash[:cnpj] if options[:payment_country] == 'BR'
+          buyer[:cnpj] = buyer_hash[:cnpj] if @options[:payment_country]  == 'BR'
```
bpollack: And again, apparently? Not sure what's going on here; buggy formatter?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -191,15 +190,15 @@ def add_buyer(post, payment_method, options)
...
           buyer[:emailAddress] = buyer_hash[:email]
           buyer[:contactPhone] = (options[:billing_address][:phone] if options[:billing_address]) || (options[:shipping_address][:phone] if options[:shipping_address]) || ''
           buyer[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
         else
           buyer[:fullName] = payment_method.name.strip
           buyer[:dniNumber] = options[:dni_number]
           buyer[:dniType] = options[:dni_type]
-          buyer[:cnpj] = options[:cnpj] if options[:payment_country] == 'BR'
+          buyer[:cnpj] = options[:cnpj] if @options[:payment_country]  == 'BR'
```
bpollack: And…again?

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -235,8 +234,8 @@ def add_invoice(post, money, options)
 
         additional_values = {}
         additional_values[:TX_VALUE] = tx_value
-        additional_values[:TX_TAX] = tx_tax if options[:payment_country] == 'CO'
-        additional_values[:TX_TAX_RETURN_BASE] = tx_tax_return_base if options[:payment_country] == 'CO'
+        additional_values[:TX_TAX] = tx_tax if @options[:payment_country]  == 'CO'
```
bpollack: And again.

test/unit/gateways/payu_latam_test.rb
```diff
@@ -187,9 +187,9 @@ def test_partial_buyer_hash_info
         phone: "(11)756312345"
       ),
       buyer: {
-        name: 'Jorge Borges',
-        dni_number: '5415668464456',
-        email: 'axaxaxas@mlo.org'
+        name: "Jorge Borges",
```
bpollack: You didn't actually change these lines, but they're flagged due to quote changes.

test/unit/gateways/payu_latam_test.rb
```diff
@@ -24,9 +24,9 @@ def setup
       tax_return_base: 0,
       email: "username@domain.com",
       ip: "127.0.0.1",
-      device_session_id: 'vghs6tvkcle931686k1900o6e1',
-      cookie: 'pt1t38347bs6jc9ruv2ecpv7o2',
-      user_agent: 'Mozilla/5.0 (Windows NT 5.1; rv:18.0) Gecko/20100101 Firefox/18.0',
+      device_session_id: "vghs6tvkcle931686k1900o6e1",
```
bpollack: These lines aren't actually changed, but they appear to be due to quote changes.

test/unit/gateways/payu_latam_test.rb
```diff
@@ -264,20 +266,21 @@ def test_colombia_required_fields
         zip: "01019-030",
         phone: "(11)756312633"
       ),
-      tx_tax: '3193',
-      tx_tax_return_base: '16806'
+      tx_tax: "3193",
```
bpollack: Quotes again.

test/unit/gateways/payu_latam_test.rb
```diff
@@ -297,58 +300,16 @@ def test_mexico_required_fields
         zip: "01019-030",
         phone: "(11)756312633"
       ),
-      birth_date: '1985-05-25'
+      birth_date: "1985-05-25"
```
bpollack: And again.

test/unit/gateways/payu_latam_test.rb
```diff
@@ -297,58 +300,16 @@ def test_mexico_required_fields
...
     }
 
     stub_comms do
-      @gateway.purchase(@amount, @credit_card, @options.update(options_mexico))
+      gateway.purchase(@amount, @credit_card, @options.update(options_mexico))
```
bpollack: This won't work; `stub_comms`, when not given arguments, alters `@gateway`, but you're running the test against `gateway`, so this will be doing live requests.

test/unit/gateways/payu_latam_test.rb
curiousepic: Yeah I thought I would make the quotes more uniform while I was in here because they were bothering me.

test/unit/gateways/payu_latam_test.rb
curiousepic: It's running against the Mexico-specific gateway local variable created on line 281.

bpollack: But `stub_comms` does literally nothing here, so either this isn't a unit test, or the unit test isn't working properly. Or maybe it somehow is working properly (though I confess to confusion on how), but in that case, `stub_comms` is a no-op and can be deleted.

test/unit/gateways/payu_latam_test.rb
bpollack: That's fair, though two notes on that: first, that should probably be a separate commit, because it'll otherwise confuse `git blame`; and second, I've been told (and tools like RuboCop generally enforce) the quote format to standardize on for strings that do not require interop in Ruby would be `''`, not `""`, specifically so that adding a random `#` doesn't suddenly result in surprising behavior. So I think you'd want to go the other direction here (unless we've got code standards that specify the reverse is the standard within AM).

bpollack:
In addition to what I mention here, since a lot of this is actually about handling Mexico specially, and that's mentioned neither in the code nor in the commit message, I think it might be worth:

  1. Modifying the commit message to give some background would probably be helpful;
  2. Adding to the commit message why _removing some tests_ was okay (several tests that checked for success were altered to instead check for failure) might also be useful; and finally,
  3. At least one unit tests should explain why Mexico needs to be special (see discussion kicked off beginning in https://spreedly.slack.com/archives/C5GBHQTMH/p1512154739000394)

test/unit/gateways/payu_latam_test.rb
curiousepic: AFAIK, it's required for the `.check_request` below.

test/unit/gateways/payu_latam_test.rb
curiousepic: Fair points; Something made me think interops need double instead of single.

lib/active_merchant/billing/gateways/payu_latam.rb
curiousepic: Probably a bad find and replace...ment.

test/unit/gateways/payu_latam_test.rb
bpollack: But `gateway` _is a live gateway_; your call to `.purchase` at https://github.com/activemerchant/active_merchant/pull/2663/files/795afc742131856525521e627984d193c8fdfaac#diff-02139dc0c1d5f38586d52890be485342R307 going to actively call against them. So this either belongs in the remote tests, or you want

```
gateway = PayuLatamGateway.new(merchant_id: "merchant_id", account_id: "account_id", api_login: "api_login", api_key: "api_key", payment_country: "MX")
...
stub_comms(gateway) do ...
```

curiousepic: I see! Thanks.

curiousepic:
I think the test changes had a side effect of looking like it has something to do specifically with Mexico when in fact it does not. This simply requires the `payment_country` field at initialization and removes support for allowing it to be overridden by an `option` field. Some of the countries do have their own specific required fields but I think the tests cover that well enough. I could definitely be more explicit in the commit message; I'll add some background.

curiousepic:
@bpollack Updated commit message and addressed your concerns, have another look.

curiousepic:
Yeah! Thanks for the lookover and suggestions.

PR #2664, opened 2017-12-04
Add verify call for MiGS

bpollack:
This code probably works, but since the remote tests are broken and we can't track down documentation at the moment that would let us figure out *why* they're broken, this commit should not be merged.


bpollack:
Yanking the DNM; this can go out now.

PR #2669, opened 2017-12-07
PayU Latam: Change default text for description

nfarve:
Updates description to include merchant_id if there is no description.
Updates the tests for purchases in Colombia to properly use tax
information and to  match the test example
provided in docs.

3 tests are failing that are not related to these changes

Remote Tests:
Finished in 49.482971 seconds.
19 tests, 50 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
84.2105% passed

Unit Tests:
Finished in 0.015186 seconds.
22 tests, 86 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -152,7 +152,7 @@ module ActiveMerchant #:nodoc:
         order[:accountId] = @options[:account_id]
         order[:partnerId] = options[:partner_id] if options[:partner_id]
         order[:referenceCode] = options[:order_id] || generate_unique_id
-        order[:description] = options[:description] || 'unspecified'
+        order[:description] = options[:description] || 'Compra en ' + @options[:merchant_id]
         order[:language] = 'en'
         order[:shippingAddress] = shipping_address_fields(options) if options[:shipping_address]
         post[:transaction][:order] = order
--- a/test/remote/gateways/remote_payu_latam_test.rb
+++ b/test/remote/gateways/remote_payu_latam_test.rb
@@ -146,11 +146,11 @@ class RemotePayuLatamTest < Test::Unit::TestCase
         zip: "01019-030",
         phone: "(11)756312633"
       ),
-      tx_tax: '3193',
-      tx_tax_return_base: '16806'
+      tax: "3193",
+      tax_return_base: "16806"
     }
 
-    response = gateway.purchase(@amount, @credit_card, @options.update(options_colombia))
+    response = gateway.purchase(2000000, @credit_card, @options.update(options_colombia))
     assert_success response
     assert_equal "APPROVED", response.message
     assert response.test?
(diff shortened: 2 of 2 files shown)


lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -152,7 +152,7 @@ def add_order(post, options)
         order[:accountId] = @options[:account_id]
         order[:partnerId] = options[:partner_id] if options[:partner_id]
         order[:referenceCode] = options[:order_id] || generate_unique_id
-        order[:description] = options[:description] || 'unspecified'
+        order[:description] = options[:description] || 'Compra en ' + @options[:merchant_id]
```
bpollack: Just to sanity check here, it makes to add `Compra en` before the user-supplied string? I trust you, just it's the only instance of this I could find in AM (not that I spent _that_ long looking).

test/remote/gateways/remote_payu_latam_test.rb
```diff
@@ -146,11 +146,11 @@ def test_successful_purchase_colombia
         zip: "01019-030",
         phone: "(11)756312633"
       ),
-      tx_tax: '3193',
-      tx_tax_return_base: '16806'
+      tax: "3193",
```
bpollack: Why the quote change?

test/remote/gateways/remote_payu_latam_test.rb
```diff
@@ -146,11 +146,11 @@ def test_successful_purchase_colombia
...
+      tax_return_base: "16806"
     }
 
-    response = gateway.purchase(@amount, @credit_card, @options.update(options_colombia))
+    response = gateway.purchase(2000000, @credit_card, @options.update(options_colombia))
```
bpollack: What did this alter?

test/remote/gateways/remote_payu_latam_test.rb
nfarve: More so just making it consistent with the rest of test.

test/remote/gateways/remote_payu_latam_test.rb
nfarve: You cant use a value less than the tax values. So I had to increase the purchase amount for the test to pass.

bpollack: Would it make sense to just bump up `@amount` then?

lib/active_merchant/billing/gateways/payu_latam.rb
nfarve: Yea from the ticket: 

"PayU Latam requires their description field. We noted in our documentation that it is a required GSF, however, if it's not sent, then we currently send "unspecified" and PayULatam wants us to default to "Compra en insert merchant name". We can't do this because we don't have the merchant name. But we do have their merchant_id from their gateway credentials which they said are approved."

test/remote/gateways/remote_payu_latam_test.rb
nfarve: Just tried that and it fails another test.

lib/active_merchant/billing/gateways/payu_latam.rb
bpollack: This is a weird gateway.

Issue #2666, opened 2017-12-06
SagePay CVV match issue


singhai0:
I think we have been dealing with the `cvv_result` for SagePay incorrectly. 

We get the CVV result from the gateway in the `CV2Result` parameter, that we try to transform into a `cvv_result` as:
https://github.com/activemerchant/active_merchant/blob/93ec51ad6532aa5464b13233b564b1d06449b59d/lib/active_merchant/billing/gateways/sage_pay.rb#L354

which performs a lookup in the `AVS_CVV_CODE` hash:
https://github.com/activemerchant/active_merchant/blob/93ec51ad6532aa5464b13233b564b1d06449b59d/lib/active_merchant/billing/gateways/sage_pay.rb#L40-L45

As it turns out, we map a CVV result of `MATCHED` to `Y` which is incorrect -- `Y` is not a proper cvv result code (it would even fail ActiveMerchant's own [validations](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/cvv_result.rb)).

#2670 attempts to fix this behavior.

singhai0:
Closing as #2670 got merged.

PR #2670, opened 2017-12-07
Correcting CVV, AVS codes for Sagepay

singhai0:
Fixes https://github.com/activemerchant/active_merchant/issues/2666.

I should have written tests but my environments are all messed up.

### DIFF
--- a/lib/active_merchant/billing/gateways/sage_pay.rb
+++ b/lib/active_merchant/billing/gateways/sage_pay.rb
@@ -37,10 +37,17 @@ module ActiveMerchant #:nodoc:
         :jcb => "JCB"
       }
 
-      AVS_CVV_CODE = {
+      AVS_CODE = {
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/sage_pay.rb
```diff
@@ -37,13 +37,20 @@ class SagePayGateway < Gateway
...
         "NOTPROVIDED" => nil,
         "NOTCHECKED" => 'X',
         "MATCHED" => 'Y',
         "NOTMATCHED" => 'N'
       }
 
+      CVV_CODE = {
+        "NOTPROVIDED" => 'S',
+        "NOTCHECKED" => 'X',
+        "MATCHED" => 'M',
```
damianthe: Looks good. I wonder if we have any other downstream dependencies on the 'Y' value other than risk_analysis. Definitely worth a look

PR #2672, opened 2017-12-11
[wip] Adding a method to generate avs code from postal & street matches

singhai0:
A few gateways provide AVS postal and street checks, but do not provide any value avs `code`. Through this PR, I am attempting to add a rudimentary method that would allow us to fill in this missing `code`.

cc @jasonwebster

### DIFF
--- a/lib/active_merchant/billing/avs_result.rb
+++ b/lib/active_merchant/billing/avs_result.rb
@@ -70,12 +70,17 @@ module ActiveMerchant
       def initialize(attrs)
         attrs ||= {}
         
-        @code = attrs[:code].upcase unless attrs[:code].blank?
+        if attrs[:code].blank?
+          @code = generate_avs_code(street_match: attrs[:street_match], postal_match: attrs[:postal_match])
+        else
+          @code = attrs[:code].upcase
+        end
+
         @message = self.class.messages[code]
-        
+
         if attrs[:street_match].blank?
           @street_match = STREET_MATCH_CODE[code]
-        else  
+        else
           @street_match = attrs[:street_match].upcase
         end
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/avs_result.rb
```diff
@@ -93,6 +98,22 @@ def to_hash
           'postal_match' => postal_match
         }
       end
+
+      def generate_avs_code(street_match:, postal_match:)
+        return if street_match.blank? && postal_match.blank?
```
infiton: should this be `street_match.blank? || postal_match.blank?`

singhai0: No. If either `street_match` or `postal_match` is available, we can generate `A` or `P` codes respectively.

If `if street_match.blank? && postal_match.blank?` is true, these two letter codes offer no valuable information, and thus no `code` can be generated

singhai0: Although it can be said that `U` should get generated in those cases, but in various implementations, I found out that this made the most sense.

bpollack:
Are y'all still interested in moving forward here? It's been over six months, so I'm not sure if you're on a fork, no longer need this, or would still like to merge.

singhai0:
I wrote this while I was an intern at Shopify, where we had a lot of data from various gateways that could've been used to fully build this _reverse lookup_ of sort. Now that I am back in school, I no longer have access to that data and am unable to make progress on this 😩.

Maybe @tothpete or someone else can continue work on this. What do you think @infiton?

rperetti:
At this point I think we should just close it.

Issue #2502, opened 2017-07-11
Vantiv (Litle) Url Update deadline approaching Aug 31 2017


kjbkjb:
My corporation does not allow me to contribute so I'm submitting an issue here:

Vantiv purchase Litle quite a while ago, now they are migrating their endpoints to their own domain with a hard deadline of 8/31
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/litle.rb#L9

The live URL will be changing to https://payments.vantivcnp.com/vap/c
ommunicator/online
The test URL, as far as I can tell will not be changing.

We've monkey-patched for this change but perhaps someone in the contributor list can make this update to AM.

Contact me and I can provide Vantiv's FAQ on the topic.

VantivSDK:
Hi @kjbkjb,
Vantiv can take this issue and submit a pull request with updated URLs. The test URL will change as well, we will update that too when all is ready to go live.

kjbkjb:
THANK YOU!

brianarnold789:
@kjbkjb pull request for prod and cert/prelive URL updates: https://github.com/activemerchant/active_merchant/pull/2504

bruno:
Note this has already been done in #2531

PR #2673, opened 2017-12-11
Vantiv (Litle): Update sandbox url for testing

wsmoak:
Previously the URL for live transactions was updated, however the test url wasn't, and it has now stopped working.

This updates the sandbox url based on the documentation in https://developer.vantiv.com/docs/DOC-1375

Related to #2502


bpollack:
Thanks @wsmoak! Merging.

PR #2675, opened 2017-12-12
Mercado Pago: protect against invalid access tokens

bpollack:
An access token provided that was not validly URL-encoded (e.g., had
spaces) could cause ActiveRecord to crash, since it could result in a
malformed URL. This ensures that even a malformed access token can be
passed through, avoiding a crash and allowing normal error handling
to take place.


bpollack:
(Relates to ENRG-7132.)

PR #2676, opened 2017-12-13
MiGS: improve test suite and SecureHash

bpollack:
Our test cards were out-of-date, and MiGS now requires SHA256 hashes.
Address both of these issues.

Closes #2257


bpollack:
(NB, #2257 did some of this work over a year ago; I didn't see it until after I'd done my own version.)

PR #2685, opened 2017-12-21
Implement Paymentez Gateway

bpollack:
Unit tests:

13 tests, 37 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests:

13 tests, 29 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -1,0 +1,273 @@
+require 'base64'
+require 'digest'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PaymentezGateway < Gateway #:nodoc:
+      self.test_url = 'https://ccapi-stg.paymentez.com/v2/'
+      self.live_url = 'https://ccapi.paymentez.com/v2/'
+
+      self.supported_countries = %w[MX EC VE CO BR CL]
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express diners_club]
+
+      self.homepage_url = 'https://secure.paymentez.com/'
+      self.display_name = 'Paymentez'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        1 => :processing_error,
+        6 => :card_declined,
+        9 => :card_declined,
+        10 => :processing_error,
+        11 => :card_declined,
+        12 => :config_error,
+        13 => :config_error,
+        19 => :invalid_cvc,
+        20 => :config_error,
+        21 => :card_declined,
+        22 => :card_declined,
+        23 => :card_declined,
+        24 => :card_declined,
+        25 => :card_declined,
+        26 => :card_declined,
+        27 => :card_declined,
+        28 => :card_declined
+      }.freeze
+
+      CARD_MAPPING = {
+        'visa' => 'vi',
+        'master' => 'mc',
+        'american_express' => 'ax',
+        'diners_club' => 'di'
+      }.freeze
+
+      def initialize(options = {})
+        requires!(options, :application_code, :app_key)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        post = {}
+
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_customer_data(post, options)
+
+        commit_transaction('debit_cc', post)
+      end
+
+      def authorize(money, payment, options = {})
+        post = {}
+
+        add_invoice(post, money, options)
+        add_customer_data(post, options)
+
+        MultiResponse.run do |r|
+          r.process { store(payment, options) }
+          post[:card] = { token: r.authorization }
+          r.process { commit_transaction('authorize', post) }
+        end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -0,0 +1,274 @@
...
+      def card_message_from(response)
+        if !response.include?('error')
+          response['message'] || response['card']['message']
+        else
+          response['error']['type']
+        end
+      end
+
+      def authorization_from(response)
+        (response['transaction'] || {})['id']
```
curiousepic: @dtykocki For my own edification as well, since I don't think I've seen it before, do you have an opinion on this pattern vs. something like a more conventional verbose
`response['transaction']['id'] if response['transaction'] && response['transaction']['id']`?

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -0,0 +1,274 @@
...
+        response = commit_card('add', post)
+        if !response.success? && !(token = extract_previous_card_token(response)).nil?
+          unstore("#{token}|#{post[:user][:id]}")
+          response = commit_card('add', post)
+        end
+        response
+      end
+
+      def unstore(identification, _options = {})
+        card_token, user_id = identification.split('|', 2)
```
curiousepic: I think we might prefer to pull the user_id from options instead of requiring the user to concatenate it to the token when passing it in here.

lib/active_merchant/billing/gateways/paymentez.rb
dtykocki: It's for sure less verbose, which is pretty nice, but I'd prefer to stick with our current convention for consistency sake.

bpollack: The dying Pythonista in me respects but is saddened by your choice, @dtykocki ;-). I'll switch this pattern though.

lib/active_merchant/billing/gateways/paymentez.rb
bpollack: Oh, point. Sorry, first interaction with TPV. That'll also avoid me having to make `store` munge. Thanks!

lib/active_merchant/billing/gateways/paymentez.rb
dtykocki: Well.. if Ruby had pattern matching like Elixir, this would be a nil point.

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -0,0 +1,274 @@
...
+
+      def store(credit_card, options = {})
+        post = {}
+
+        add_customer_data(post, options)
+        add_payment(post, credit_card)
+
+        response = commit_card('add', post)
+        if !response.success? && !(token = extract_previous_card_token(response)).nil?
+          unstore("#{token}|#{post[:user][:id]}")
```
dtykocki: So remove the previously stored card (if one exist) and store a new one instead. Am I following that correctly?

lib/active_merchant/billing/gateways/paymentez.rb
bpollack: Seriously, @dtykocki, I was just dying for F#/Elixir/Haskell pattern matching throughout this PR. Given the heterogeneous nature of their JSON, that'd have simplified things _so much_ it's not funny.

lib/active_merchant/billing/gateways/paymentez.rb
bpollack: Yep, you've got it. I initially wrote this to just generate a random `user_id` so we could just store the same card multiple times (which is the approach taken by other tokenize-then-`$WHATVER` gateways), but because the following operations require `user_id` to be constant, that wasn't a viable path forward.

dtykocki: Fair enough 👍

bpollack:
All right, all your concerns address (I think). No rush, but let me know if this looks good now.

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -0,0 +1,273 @@
...
+
+      def card_success_from(response)
+        return false if response.include?('error')
+        return true if response['message'] == 'card deleted'
+        response['card']['status'] == 'valid'
+      end
+
+      def message_from(response)
+        if success_from(response)
+          response['transaction'] && response['transaction']['message']
```
curiousepic: Wait, does this not just return a bool?

curiousepic: Ah it returns the value for the last argument. Neat.

bpollack: Yeah, that's why Ruby 2.3 named its safe navigation operator `&.`, rather than e.g. the more common `?.`: naming it `&.` echoes this pattern.

PR #2690, opened 2017-12-29
Authorize.Net: Get customer profile with merchant_customer_id or email

alexcwatt:
I wanted to use the `AuthorizeNetCim#get_customer_profile` method in a project and noticed the API [supports getting a profile with the `merchantCustomerId` or `email`](https://developer.authorize.net/api/reference/index.html#customer-profiles-get-customer-profile) in addition to the `customerProfileId`, so I added support for that.

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net_cim.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net_cim.rb
@@ -216,9 +216,13 @@ module ActiveMerchant #:nodoc:
       #
       # ==== Options
       #
-      # * <tt>:customer_profile_id</tt> -- The Customer Profile ID of the customer to retrieve. (REQUIRED)
+      # * <tt>:customer_profile_id</tt> -- The Customer Profile ID of the customer to retrieve. (CONDITIONAL)
+      # * <tt>:email</tt> -- The Email address associated with the customer to retrieve. (CONDITIONAL)
+      # * <tt>:merchant_customer_id</tt> -- The Merchant assigned ID of the customer to retrieve. (CONDITIONAL)
       def get_customer_profile(options)
-        requires!(options, :customer_profile_id)
+        requires!(options, :customer_profile_id) unless options[:email] || options[:merchant_customer_id]
+        requires!(options, :email) unless options[:customer_profile_id] || options[:merchant_customer_id]
+        requires!(options, :merchant_customer_id) unless options[:customer_profile_id] || options[:email]
 
         request = build_request(:get_customer_profile, options)
         commit(:get_customer_profile, request)
@@ -563,7 +567,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def build_get_customer_profile_request(xml, options)
-        xml.tag!('customerProfileId', options[:customer_profile_id])
+        xml.tag!('customerProfileId', options[:customer_profile_id]) if options[:customer_profile_id]
+        xml.tag!('merchantCustomerId', options[:merchant_customer_id]) if options[:merchant_customer_id]
+        xml.tag!('email', options[:email]) if options[:email]
         xml.target!
       end
 
--- a/lib/active_merchant/billing/gateways/authorize_net_cim.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net_cim.rb
@@ -220,9 +220,7 @@ module ActiveMerchant #:nodoc:
       # * <tt>:email</tt> -- The Email address associated with the customer to retrieve. (CONDITIONAL)
       # * <tt>:merchant_customer_id</tt> -- The Merchant assigned ID of the customer to retrieve. (CONDITIONAL)
       def get_customer_profile(options)
-        requires!(options, :customer_profile_id) unless options[:email] || options[:merchant_customer_id]
-        requires!(options, :email) unless options[:customer_profile_id] || options[:merchant_customer_id]
-        requires!(options, :merchant_customer_id) unless options[:customer_profile_id] || options[:email]
+        raise ArgumentError, "Requires one of customer_profile_id, email, or merchant_customer_id" unless options[:customer_profile_id] || options[:email] || options[:merchant_customer_id]
 
         request = build_request(:get_customer_profile, options)
         commit(:get_customer_profile, request)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/authorize_net_cim.rb
```diff
@@ -216,9 +216,13 @@ def delete_customer_shipping_address(options)
       #
       # ==== Options
       #
-      # * <tt>:customer_profile_id</tt> -- The Customer Profile ID of the customer to retrieve. (REQUIRED)
+      # * <tt>:customer_profile_id</tt> -- The Customer Profile ID of the customer to retrieve. (CONDITIONAL)
+      # * <tt>:email</tt> -- The Email address associated with the customer to retrieve. (CONDITIONAL)
+      # * <tt>:merchant_customer_id</tt> -- The Merchant assigned ID of the customer to retrieve. (CONDITIONAL)
       def get_customer_profile(options)
-        requires!(options, :customer_profile_id)
+        requires!(options, :customer_profile_id) unless options[:email] || options[:merchant_customer_id]
```
bpollack: I think this would be better as a straight conditional, rather than three guarded `requires!` calls. Best as I can figure, what you're trying to do is say you want one of `customer_profile_id`, `email`, or `merchant_customer_id`. In that case, I'd skip `requires!` entirely (which is just a function that throws `ArgumentError` in certain conditions) and throw the exception yourself, or do simpler logic. As an example, something like (didn't think through this, not tested, may be wrong, but points in the direction I'm thinking)

`raise ArgumentError("Requires one of customer_profile_id, email, or merchant_customer_id") unless options[:customer_profile_id] || options[:email] || options[:merchant_customer_id]`

might be a better fit.

test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -74,8 +74,17 @@ def test_successful_profile_create_get_update_and_delete
     assert_nil response.authorization
     assert response = @gateway.get_customer_profile(:customer_profile_id => @customer_profile_id)
     assert_nil response.params['profile']['merchant_customer_id']
-    assert_nil response.params['profile']['description']
+    assert_equal 'Up to 255 Characters', response.params['profile']['description']
```
bpollack: Was this a change in the API?

test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -74,8 +74,17 @@ def test_successful_profile_create_get_update_and_delete
...
     assert_equal 'new email address', response.params['profile']['email']
+
+    assert response = @gateway.update_customer_profile(:profile => {:customer_profile_id => @customer_profile_id, :email => 'johndoe@test.com'})
+    assert response.test?
+    assert_success response
+    assert_nil response.authorization
+    assert response = @gateway.get_customer_profile(:email => 'johndoe@test.com')
```
bpollack: That's not the `response`; that's the `profile`. Use a new variable.

test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -74,8 +74,17 @@ def test_successful_profile_create_get_update_and_delete
...
+    assert_nil response.params['profile']['merchant_customer_id']
+    assert_equal 'Up to 255 Characters', response.params['profile']['description']
+    #assert_equal 'new email address', response.params['profile']['email']
```
bpollack: Commented-out code.

test/unit/gateways/authorize_net_cim_test.rb
```diff
@@ -12,15 +12,17 @@ def setup
     @credit_card = credit_card
     @address = address
     @customer_profile_id = '3187'
+    @email = 'johndoe@test.com'
```
bpollack: Do you truly want these in all tests? If not, I'd modify the tests themselves where you want this. The fact you've got optional modifications at https://github.com/activemerchant/active_merchant/pull/2690/files#diff-068964b7f2d4b4c8f35e220595c2ee62R25 makes me think this shouldn't be a global change.

test/unit/gateways/authorize_net_cim_test.rb
```diff
@@ -342,6 +344,20 @@ def test_should_get_customer_profile_request
...
     assert_equal @customer_profile_id, response.authorization
+
+    @gateway.expects(:ssl_post).returns(successful_get_customer_profile_response)
+
+    assert response = @gateway.get_customer_profile(:email => @email)
+    assert_instance_of Response, response
+    assert_success response
+    assert_equal @email, response.params['email']
+
+    @gateway.expects(:ssl_post).returns(successful_get_customer_profile_response)
```
bpollack: I'd prefer to guard these with `.when()` calls, rather than just adding new `.returns()` calls at specific junctions in tests, but that's a personal preference.

test/unit/gateways/authorize_net_cim_test.rb
```diff
@@ -342,6 +344,20 @@ def test_should_get_customer_profile_request
     assert_instance_of Response, response
     assert_success response
     assert_equal @customer_profile_id, response.authorization
+
+    @gateway.expects(:ssl_post).returns(successful_get_customer_profile_response)
+
+    assert response = @gateway.get_customer_profile(:email => @email)
+    assert_instance_of Response, response
```
bpollack: This line (and its twin on 358) don't seem to give us anything useful.

lib/active_merchant/billing/gateways/authorize_net_cim.rb
alexcwatt: Sounds good, I was just following a pattern I found in `create_customer_profile`. But that would be a lot cleaner.

test/remote/gateways/remote_authorize_net_cim_test.rb
alexcwatt: I don't know how this ever passed before. The description was being set, so it shouldn't have been asserted as nil.

alexcwatt:
@bpollack Thank you for all the feedback; I know my style needs improvement and I love that open source projects allow learning like that :) I pushed another commit a few days ago that I think addresses the change requests; if you could take another look and let me know if you see anything else, that would be great!

bpollack:
@nfarve (Just because you're in authorize.net.) This seems like it might be in line with the general work you're doing right now, and the patch looks clean; any thoughts on merging?

PR #2693, opened 2018-01-02
Payflow: Change Verify Method for Amex Cards

nfarve:
Currently $0 authorization is only supported for AVS. This changes the
verify method for amex cards to be a $1 auth and void.

Tests failures are unrelated to these changes.

Loaded suite test/remote/gateways/remote_payflow_test
27 tests, 99 assertions, 10 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
62.963% passed
----------------------------------------------------------------------------------------
0.60 tests/s, 2.18 assertions/s
Loaded suite test/unit/gateways/payflow_test
44 tests, 192 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/fixtures.yml
+++ b/test/fixtures.yml
@@ -663,7 +663,7 @@ payex:
 # Working credentials, no need to replace
 payflow:
   login: 'spreedlyIntegrations'
-  password: 'y9q)(j7H'
+  password: 'L9DjqEKjXCkU'
   partner: 'PayPal'
 
 payflow_uk:
--- a/lib/active_merchant/billing/gateways/payflow.rb
+++ b/lib/active_merchant/billing/gateways/payflow.rb
@@ -45,7 +45,14 @@ module ActiveMerchant #:nodoc:
       end
 
       def verify(payment, options={})
-        authorize(0, payment, options)
+        if credit_card_type(payment) == 'Amex'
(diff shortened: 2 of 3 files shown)


test/fixtures.yml
```diff
@@ -663,7 +663,7 @@ payex:
 # Working credentials, no need to replace
 payflow:
   login: 'spreedlyIntegrations'
-  password: 'y9q)(j7H'
+  password: 'L9DjqEKjXCkU'
```
bpollack: Was this intentional? Figure it probably was; just was surprised since there wasn't a mention of it in the commit message.

nfarve: Yea I was getting user authenticated errors and Paypal asked that we change our password. It ended up being a IP whitelisting issue though.

AnotherJoSmith:
@nfarve Do you have a link to any documentation to shows why this change is necessary? Also would've been nice to test that the authorization is successfully voided afterwards.

nfarve:
@AnotherJoSmith We had direct conversation with Payflow that stated ```Payflow stated "“..Currently today for AMEX direct or AMEX via processor $0 authorization is only support for AVS not CSC – so if they remove CSC it would work…”```. However our customer wanted to use the csc, which required the change. I can add a test for voiding these transactions.

PR #2696, opened 2018-01-03
Forte and Moneris: ensure all unit tests do not call remote servers

bpollack:
Our Travis runs are failing because Forte and Moneris (at least) have unit tests that are actually partially remote tests. These commits get them local again.


bpollack:
We can squash these; I don't have a strong preference.

bpollack:
Squashed per offline discussion.

PR #2699, opened 2018-01-04
Set the `receipt_email` attribute of charges to `options[:email]`

jasonwebster:
I feel like this is how this should have always worked. The `:email`
option is a core, standardize option across Active Merchant, and is the
"customer's email address", meanwhile, the `receipt_email` attribute on
a Stripe charge is the "address a customer's receipt was sent to."
Definitely sounds like the same thing to me.

I've left the existing, one-off option `receipt_email` be, and it will
take precedence over the `email` option, which should maintain
backwards compatibility.


jasonwebster:
Turns out we can't just do this:

> If receipt_email is specified for a charge in live mode, a receipt will be sent regardless of your email settings.

from: https://stripe.com/docs/api#create_charge-receipt_email

PR #2700, opened 2018-01-08
Add or fix scrubbing in several adapters

bpollack:


bpollack:
@dtykocki Hah, point. That's because I wasn't consistent about what existing adapter's tests I was stealing. I'll look for what's most common as-is and then move them all over to that before merging.

bpollack:
Okay, all unified on `test_transcript_scrubbing`. I'll merge as soon as Travis verifies I didn't break anything on that massive rebase.

PR #2702, opened 2018-01-09
Add more scrubbing support

bpollack:


test/unit/gateways/mercury_test.rb
```diff
@@ -117,91 +117,146 @@ def test_card_present_with_invalid_data
     assert_success response
   end
 
+  def test_transcript_scrubbing
+    assert @gateway.supports_scrubbing?
+    File.open("expected.txt", "w") { |f| f.write post_scrub }
+    File.open("actual.txt", "w") { |f| f.write @gateway.scrub(pre_scrub) }
```
dtykocki: Should both of these `File.open` calls be here?

bpollack: Sigh, nope. Good catch. I'll axe them.

PR #2706, opened 2018-01-11
CardConnect: Add new gateway

nfarve:
Adds a new gateway card_connect to Active Merchant

Loaded suite test/remote/gateways/remote_card_connect_test
...................

19 tests, 45 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/unit/gateways/card_connect_test
................

16 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/card_connect.rb
+++ b/lib/active_merchant/billing/gateways/card_connect.rb
@@ -1,0 +1,286 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CardConnectGateway < Gateway
+      self.test_url = 'https://fts.cardconnect.com:6443/cardconnect/rest/'
+      self.live_url = 'https://fts.cardconnect.com:8443/cardconnect/rest/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'https://cardconnect.com/'
+      self.display_name = 'Card Connect'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        '11' => STANDARD_ERROR_CODE[:card_declined],
+        '12' => STANDARD_ERROR_CODE[:incorrect_number],
+        '13' => STANDARD_ERROR_CODE[:incorrect_cvc],
+        '14' => STANDARD_ERROR_CODE[:incorrect_cvc],
+        '15' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        '16' => STANDARD_ERROR_CODE[:expired_card],
+        '17' => STANDARD_ERROR_CODE[:incorrect_zip],
+        '21' => STANDARD_ERROR_CODE[:config_error],
+        '22' => STANDARD_ERROR_CODE[:config_error],
+        '23' => STANDARD_ERROR_CODE[:config_error],
+        '24' => STANDARD_ERROR_CODE[:processing_error],
+        '25' => STANDARD_ERROR_CODE[:processing_error],
+        '27' => STANDARD_ERROR_CODE[:processing_error],
+        '28' => STANDARD_ERROR_CODE[:processing_error],
+        '29' => STANDARD_ERROR_CODE[:processing_error],
+        '31' => STANDARD_ERROR_CODE[:processing_error],
+        '32' => STANDARD_ERROR_CODE[:processing_error],
+        '33' => STANDARD_ERROR_CODE[:card_declined],
+        '34' => STANDARD_ERROR_CODE[:card_declined],
+        '35' => STANDARD_ERROR_CODE[:incorrect_zip],
+        '36' => STANDARD_ERROR_CODE[:processing_error],
+        '37' => STANDARD_ERROR_CODE[:incorrect_cvc],
+        '41' => STANDARD_ERROR_CODE[:processing_error],
+        '42' => STANDARD_ERROR_CODE[:processing_error],
+        '43' => STANDARD_ERROR_CODE[:processing_error],
+        '44' => STANDARD_ERROR_CODE[:config_error],
+        '61' => STANDARD_ERROR_CODE[:processing_error],
+        '62' => STANDARD_ERROR_CODE[:processing_error],
+        '63' => STANDARD_ERROR_CODE[:processing_error],
+        '64' => STANDARD_ERROR_CODE[:config_error],
+        '65' => STANDARD_ERROR_CODE[:processing_error],
+        '66' => STANDARD_ERROR_CODE[:processing_error],
+        '91' => STANDARD_ERROR_CODE[:processing_error],
+        '92' => STANDARD_ERROR_CODE[:processing_error],
+        '93' => STANDARD_ERROR_CODE[:processing_error],
+        '94' => STANDARD_ERROR_CODE[:processing_error],
+        '95' => STANDARD_ERROR_CODE[:config_error],
+        '96' => STANDARD_ERROR_CODE[:processing_error],
+        'NU' => STANDARD_ERROR_CODE[:card_declined],
+        'N3' => STANDARD_ERROR_CODE[:card_declined],
+        'NJ' => STANDARD_ERROR_CODE[:card_declined],
+        '51' => STANDARD_ERROR_CODE[:card_declined],
+        'C2' => STANDARD_ERROR_CODE[:incorrect_cvc],
+        '54' => STANDARD_ERROR_CODE[:expired_card],
+        '05' => STANDARD_ERROR_CODE[:card_declined],
+        '03' => STANDARD_ERROR_CODE[:config_error],
+        '60' => STANDARD_ERROR_CODE[:pickup_card]
+      }
+
+      def initialize(options = {})
+        requires!(options, :merchant_id, :username, :password)
+        require_valid_domain!(options, :domain)
+        super
+      end
+
+      def require_valid_domain!(options, param)
+        if options.key?(param)
+          raise ArgumentError.new('not a valid cardconnect domain') unless /\Dcardconnect.com:\d{1,}\D/ =~ options[param]
+        end
+      end
+
+      def purchase(money, payment, options = {})
+        if options[:po_number]
+          MultiResponse.run do |r|
+            r.process { authorize(money, payment, options) }
+            r.process { capture(money, r.authorization, options) }
+          end
+        else
+          post = {}
+          add_invoice(post, options)
+          add_money(post, money)
+          add_payment(post, payment)
+          add_currency(post, money, options)
+          add_address(post, options)
+          add_customer_data(post, options)
+          add_3DS(post, options)
+          post[:capture] = 'Y'
+          commit('auth', post)
+        end
+      end
+
+      def authorize(money, payment, options = {})
+        post = {}
+        add_money(post, money)
+        add_currency(post, money, options)
+        add_invoice(post, options)
+        add_payment(post, payment)
+        add_address(post, options)
+        add_customer_data(post, options)
+        add_3DS(post, options)
+        commit('auth', post)
+      end
+
+      def capture(money, authorization, options = {})
+        post = {}
+        add_money(post, money)
+        add_reference(post, authorization)
+        add_additional_data(post, options)
+        commit('capture', post)
+      end
+
+      def refund(money, authorization, options = {})
+        post = {}
+        add_money(post, money)
+        add_reference(post, authorization)
+        commit('refund', post)
+      end
+
+      def void(authorization, options = {})
+        post = {}
+        add_reference(post, authorization)
+        commit('void', post)
+      end
+
+      def verify(credit_card, options = {})
+        authorize(0, credit_card, options)
+      end
+
+      def supports_scrubbing?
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+      private
+
+      def add_customer_data(post, options)
+        post[:email] = options[:email] if options[:email]
+      end
+
+      def add_address(post, options)
+        address = options[:billing_address] || options[:address]
+        if address
+          post[:address] = address[:address1] if address[:address1]
```
bpollack: Does it make sense to concatenate `address[:address2]` in any way, or do we not care here?

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+      end
+
+      def add_additional_data(post, options)
+        post[:ponumber] = options[:po_number]
+        post[:taxamnt] = options[:tax_amount] if options[:tax_amount]
+        post[:frtamnt] = options[:freight_amount] if options[:freight_amount]
+        post[:dutyamnt] = options[:duty_amount] if options[:duty_amount]
+        post[:orderdate] = options[:order_date] if options[:order_date]
+        post[:shipfromzip] = options[:ship_from_zip] if options[:ship_from_zip]
+        if (shipping_address = options[:shipping_address])
```
bpollack: I'm actually totally fine with this pattern, but you didn't use it elsewhere in the file (e.g. lines 147 and 148 could use this pattern). It might be worth being consistent one way or the other. Really up to you, though; I don't have a strong opinion.

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+          post[:shiptozip] = shipping_address[:zip]
+          post[:shiptocountry] = shipping_address[:country]
+        end
+        if options[:items]
+          post[:items] = []
```
bpollack: This works fine, but I had to read it twice to grok what you're actually doing is a mapping operation; might be worth converting it to

    post[:items] = options[:items].map do |item| ...

to make it clearer that's what's happening.

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+          options[:items].each do |item|
+            updated = {}
+            item.each_pair do |k, v|
```
bpollack: You likewise _can_ do this as a mapping operation, but playing with it locally, it was a bit gross. Best I got for the whole thing was

```ruby
post[:items] = options[:items].map do | item|
  item.map { |k, v| [k.to_s.gsub(/_/, ''), v] }.to_h
end
```

which wastes memory and I'm not convinced is cleaner, but your call. (There's a grosser version using `.inject` which doesn't waste RAM, but it gets sufficiently hard to read that I think your existing version is superior to that one.)

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+              updated.merge!(k.to_s.gsub(/_/, '') => v)
+            end
+            post[:items] << updated
+          end
+        end
+      end
+
+      def add_3DS(post, options)
+        post[:secureflag] = options[:secure_flag] if options[:secure_flag]
```
bpollack: This is the same pattern (albeit explicitly) that you did in `add_additional_data` (and looking back, that does it manually too, from lines 193 to 198); maybe worth extracting a method so you could cut this to something

    post.merge!(munge_keys(options, [:secure_value, :secure_xid, :secure_flag]))

May not be worth it though; entirely your call. You don't appear to be doing such direct assignments in any of the other `add_*` methods.

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+        post[:securevalue] = options[:secure_value] if options[:secure_value]
+        post[:securexid] = options[:secure_xid] if options[:secure_xid]
+      end
+
+      def headers
+        {
+          'Authorization' => 'Basic ' + Base64.strict_encode64("#{@options[:username]}:#{@options[:password]}").strip,
```
bpollack: `Base64.strict_encode64` shouldn't be including trailing spaces, so `.strip` ought to be a no-op here; should this be for the `username:password` string?

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+
+      def parse(body)
+        JSON.parse(body)
+      end
+
+      def url(action)
+        if test?
+          test_url + action
+        else
+          (@options[:domain] ? @options[:domain] : live_url) + action
```
bpollack: This needs to be vetted in some way or we're allowing ANY URL to go through here. You could exfiltrate all our cards by just passing in `postbin.com` or something.

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
...
+      def authorization_from(response)
+        response['retref']
+      end
+
+      def post_data(parameters = {})
+        parameters.to_json
+      end
+
+      def error_code_from(response)
+        return false if success_from(response)
```
bpollack: Most of our other gateways return `nil` here, which might allow you to just drop this line in favor of a slight tweak to the next.

test/fixtures.yml
```diff
@@ -122,6 +122,11 @@ cams:
   username: testintegrationc
   password: password9
 
+card_connect:
```
bpollack: It this is safe to share, please add a comment indicating that. If not, this looks like it might be our actual testing creds.

test/remote/gateways/remote_card_connect_test.rb
```diff
@@ -0,0 +1,190 @@
...
+    assert_equal 'Insufficient funds', response.message
+  end
+
+  def test_successful_authorize_and_capture
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount, auth.authorization)
+    assert_success capture
+    assert_equal 'Approval Queued for Capture' || 'Approval Accepted', capture.message
```
bpollack: This will get executed as

    assert_equal 'Approval Queued for Capture', capture.message

since 'Approved Queue for Capture' is truthy. I think you meant something closer to

    assert_includes ['Approval Queued for Capture', 'Approval Accepted'], capture.message

test/remote/gateways/remote_card_connect_test.rb
```diff
@@ -0,0 +1,190 @@
...
+  def test_partial_capture
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount - 1, auth.authorization)
+    assert_success capture
+  end
+
+  def test_failed_capture
+    response = @gateway.capture(@amount, '23221')
```
bpollack: Is this transaction guaranteed to never exist?

lib/active_merchant/billing/gateways/card_connect.rb
bpollack: (BTW, assuming this is from our convo in Slack, I was suggesting doing a regex vet. So if the URLs always have the form `*.foo.com`, we'd just throw an exception if the domain didn't match that.)

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,283 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CardConnectGateway < Gateway
+      self.test_url = 'https://fts.cardconnect.com:6443/cardconnect/rest/'
+      self.live_url = 'https://fts.cardconnect.com:8443/cardconnect/rest/'
```
dtykocki: @nfarve - The 8443 port *might* be an issue for us since it's non-standard. Just to be safe, it's probably worth checking with the ops team on this.

lib/active_merchant/billing/gateways/card_connect.rb
nfarve: Guess it doesn't hurt to concatenate the two together into the one field. 👍

lib/active_merchant/billing/gateways/card_connect.rb
nfarve: So I did have that pattern above and rubocop complained (wanted a guard clause) about it so I modified to the current.

lib/active_merchant/billing/gateways/card_connect.rb
nfarve: Would the test for this be better suited as a unit test?

bpollack: Wouldn't that still allow the customer at runtime to submit any URL?

nfarve: So your saying it should be on the core side? I was going to check in the initialize step before a commit was made.

test/remote/gateways/remote_card_connect_test.rb
nfarve: I always do that. Thanks for the catch

test/remote/gateways/remote_card_connect_test.rb
nfarve: Yea it needs to be 14 digits.

lib/active_merchant/billing/gateways/card_connect.rb
bpollack: No, I was saying you want it somewhere in the actual AM adapter. Putting it in `initialize` (as opposed to right here in `url`) would be fine, and might provide better error reporting. But I don't think we want this check in Core; while that'd help _us_ out, it both puts more logic into Core, and it doesn't help out other users of AM.

test/remote/gateways/remote_card_connect_test.rb
bpollack: Might be worth adding something like `@invalid_txn = '23221'` and then using that here to improve the documentation a bit then, but up to you.

test/remote/gateways/remote_card_connect_test.rb
bpollack: It's doubly tricky for me to catch these, because Python can handle variants of this in certain situations. E.g., `if 0 < x < 5` actually does what it looks like it ought to do. So I always have to think a second when I see these in Ruby.

test/remote/gateways/remote_card_connect_test.rb
nfarve: 👍

nfarve:
@bpollack Round 2 is up

lib/active_merchant/billing/gateways/card_connect.rb
nfarve: Ticket up to whitelist both ports.

dtykocki: We only need 8443 since that's the production endpoint. The test endpoint doesn't need to be whitelisted.

test/remote/gateways/remote_card_connect_test.rb
```diff
@@ -0,0 +1,191 @@
...
+  def test_failed_verify
+    response = @gateway.verify(@declined_card, @options)
+    assert_failure response
+    assert_match %r{Insufficient funds}, response.message
+  end
+
+  def test_invalid_login
+    gateway = CardConnectGateway.new(username: '', password: '', merchant_id: '')
+    assert_raises(ActiveMerchant::ResponseError) do
+      gateway.purchase(@amount, @credit_card, @options)
```
bpollack: Other gateways I've seen _usually_ have returned `STANDARD_ERROR_CODE[:config_error]` for this instead of throwing an exception. (In fact, I'm not used to seeing AM throw exceptions outside an adapter once the gateway's set up, period. I'm only used to seeing that as part of instantiation, such as the `requires!` calls.)

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -0,0 +1,286 @@
...
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript
+          .gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]')
+          .gsub(%r(("cvv2\\":\\")\d*), '\1[FILTERED]')
+          .gsub(%r(("merchid\\":\\")\d*), '\1[FILTERED]')
+          .gsub(%r((&?"account\\":\\")\d*), '\1[FILTERED]')
```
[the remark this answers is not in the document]
nfarve: required?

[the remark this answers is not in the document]
bpollack: Roger, so could you kill the one on line 141 too then?

lib/active_merchant/billing/gateways/card_connect.rb
nfarve: Updated the ticket. Thanks for the catch.

PR #2713, opened 2018-01-16
Stripe: Fix Partial Application Fee Refunds

curiousepic:
Note to reviewers: This was more complicated than it seemed to have any right to be, so I'm not very confident about almost any aspect of it and I'd love to get more/deeper eyes than usual on this change and am open to suggestions on better patterns, etc.

Previously, when a partial application fee refund was requested, the connected stripe_account option was sent in the authentication header for both the fee retrieval call AND the application fee refund itself. Since the fee exists only on the main Platform account, sending the connected stripe_account for the refund meant it was looking in the wrong place, and failed. Now we remove the stripe_account field before the partial refund, so it is not sent in headers.

This also switches to a method of fetching the transaction to obtain the id of the application fee, rather than querying all application fees associated with the transaction, since it simplifies things, and the prior method did not seem to have any benefits over this (having multiple application fees doesn't seem possible currently). Error messaging around this has been updated as well.

This also involved updating the aging remote_stripe_connect_tests to use the stripe_destination in fixtures as the stripe_account option, and to remove assertions that are now invalid due to the changes to response structure in https://stripe.com/docs/upgrades#2013-08-13 Unit tests were likewise updated for the partial fee refund method changes.

Remote (two failing likely due to unrelated stale test data):
59 tests, 253 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.6102% passed

Remote Connect:
3 tests, 10 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
122 tests, 651 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -151,7 +151,7 @@ module ActiveMerchant #:nodoc:
 
           return r unless options[:refund_fee_amount]
 
-          r.process { fetch_application_fees(identification, options) }
+          r.process { fetch_application_fee(identification, options) }
           r.process { refund_application_fee(options[:refund_fee_amount], application_fee_from_response(r.responses.last), options) }
         end
       end
@@ -166,9 +166,7 @@ module ActiveMerchant #:nodoc:
 
       def application_fee_from_response(response)
         return unless response.success?
-
-        application_fees = response.params["data"].select { |fee| fee["object"] == "application_fee" }
-        application_fees.first["id"] unless application_fees.empty?
+        response.params["application_fee"] unless response.params["application_fee"].empty?
       end
 
       def refund_application_fee(money, identification, options = {})
@@ -176,9 +174,11 @@ module ActiveMerchant #:nodoc:
 
         post = {}
         add_amount(post, money, options)
-        options.merge!(:key => @fee_refund_api_key)
+        options.merge!(:key => @fee_refund_api_key) if @fee_refund_api_key
+        options.delete(:stripe_account)
 
-        commit(:post, "application_fees/#{CGI.escape(identification)}/refund", post, options)
+        refund_fee = commit(:post, "application_fees/#{CGI.escape(identification)}/refunds", post, options)
+        application_fee_response!(refund_fee, "Application fee could not be refunded: #{refund_fee.message}")
       end
 
       # Note: creating a new credit card will not change the customer's existing default credit card (use :set_default => true)
@@ -371,7 +371,7 @@ module ActiveMerchant #:nodoc:
 
       def add_statement_address(post, options)
         return unless statement_address = options[:statement_address]
-        return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? } 
+        return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? }
 
         post[:statement_address] = {}
         post[:statement_address][:line1] = statement_address[:address1]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -166,19 +166,19 @@ def verify(payment, options = {})
...
 
         post = {}
         add_amount(post, money, options)
-        options.merge!(:key => @fee_refund_api_key)
+        options.merge!(:key => @fee_refund_api_key) if @fee_refund_api_key
+        options.delete(:stripe_account) if options[:stripe_account]
 
-        commit(:post, "application_fees/#{CGI.escape(identification)}/refund", post, options)
+        refund_fee = commit(:post, "application_fees/#{CGI.escape(identification)}/refunds", post, options)
+        refund_fee.success? ? refund_fee : Response.new(false, "Application fee could not be refunded: #{refund_fee.message}")
```
curiousepic: There may be a better way to hand this and 468. Their current state is based on the preexisting line 173 above.

dtykocki: Agreed. There is a quite a bit going on in just a few lines of code. At a minimum, while I try to load up in my head what's going on, `refund_fee.success? ? refund_fee : Response.new(false, "Application fee could not be refunded: #{refund_fee.message}")` could be moved to a private method as the same pattern is used on line 468.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -461,10 +461,11 @@ def add_emv_metadata(post, creditcard)
         post[:metadata][:card_read_method] = creditcard.read_method if creditcard.respond_to?(:read_method)
       end
 
-      def fetch_application_fees(identification, options = {})
+      def fetch_application_fee(identification, options = {})
         options.merge!(:key => @fee_refund_api_key)
 
-        commit(:get, "application_fees?charge=#{identification}", nil, options)
+        fetch_charge = commit(:get, "charges/#{identification}", nil, options)
```
dtykocki: Do you think we need `#{CGI.escape(identification)}` here? I'm not sure what values identification could hold, but looks like we've escaped this variable in other calls.

curiousepic: 👍 Good call.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -166,19 +166,19 @@ def verify(payment, options = {})
...
+        response.params["application_fee"] unless response.params["application_fee"].empty?
       end
 
       def refund_application_fee(money, identification, options = {})
         return Response.new(false, "Application fee id could not be found") unless identification
 
         post = {}
         add_amount(post, money, options)
-        options.merge!(:key => @fee_refund_api_key)
+        options.merge!(:key => @fee_refund_api_key) if @fee_refund_api_key
```
bpollack: This is petty but why not just `options[:key] = @fee_refund_api` instead of the `merge!` call?

bpollack: Oh. Looking elsewhere, it seems that everywhere in this file does that?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -166,19 +166,19 @@ def verify(payment, options = {})
...
+        options.delete(:stripe_account) if options[:stripe_account]
```
bpollack: Strictly speaking, `a_hash.delete(:no_such_key)` is safe, so you don't need the guard.

lib/active_merchant/billing/gateways/stripe.rb
bpollack: Yeah, I'm with @dtykocki. Maybe something like

```ruby
def require_success!(obj, message)
  obj.success? ? obj : Response.new(false, message)
end
```

which'd make it clearer what's going on here.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -371,7 +371,7 @@ def add_address(post, options)
 
       def add_statement_address(post, options)
         return unless statement_address = options[:statement_address]
-        return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? } 
+        return unless [:address1, :city, :zip, :state].all? { |key| statement_address[key].present? }
```
bpollack: I was sure Ruby had to have an equivalent of Python's `not any({'address1', 'city', 'zip', 'state'} - statement_address.keys())` but nope, this seems like the Ruby way.

curiousepic:
Thanks @bpollack, addressed the suggestions.

PR #2722, opened 2018-01-25
CenPos: Switch to new endpoint URL

bpollack:
In addition to changing the endpoint, something in Active Merchant
or this adapter isn't dealing well with Gzip'd responses, so I've
removed that header for now. (Requests at the old endpoint weren't ever
Gzip'd, so the old path was effectively untested.) I also patched one
remote test that was trivial to fix while I was in here.

The six remote test failures are identical with or without this patch.
(Well, you'll get seven without it due to the above-mentioned trivial
fix.)

Unit:
23 tests, 98 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
23 tests, 51 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
73.913% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cenpos.rb
+++ b/lib/active_merchant/billing/gateways/cenpos.rb
@@ -6,7 +6,7 @@ module ActiveMerchant #:nodoc:
       self.display_name = "CenPOS"
       self.homepage_url = "https://www.cenpos.com/"
 
-      self.live_url = "https://ww3.cenpos.net/6/transact.asmx"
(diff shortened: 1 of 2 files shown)


bpollack:
@dtykocki @deedeelavinder @nfarve @curiousepic Sorry for the explicit ping here, but this has to be live on Core by midnight UTC, so it's a bit of a rush to get this approved (or emended) ASAP.

bpollack:
Thanks @deedeelavinder !

PR #2723, opened 2018-01-25
Payment Express: Scrub merchant password

curiousepic:
Merchant password credential was not being scrubbed. There are many
failing remote tests but they are not related to this change (they fail on 
master as well).

### DIFF
--- a/test/remote/gateways/remote_payment_express_test.rb
+++ b/test/remote/gateways/remote_payment_express_test.rb
@@ -138,6 +138,7 @@ class RemotePaymentExpressTest < Test::Unit::TestCase
 
     assert_scrubbed(@credit_card.number, clean_transcript)
     assert_scrubbed(@credit_card.verification_value.to_s, clean_transcript)
+    assert_scrubbed(@gateway.options[:password], clean_transcript)
   end
 
 end
--- a/lib/active_merchant/billing/gateways/payment_express.rb
+++ b/lib/active_merchant/billing/gateways/payment_express.rb
@@ -128,7 +128,8 @@ module ActiveMerchant #:nodoc:
 
       def scrub(transcript)
         transcript.
-          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
+          gsub(%r((<PostPassword>).+(</PostPassword>)), '\1[FILTERED]\2').
+          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]\2').
           gsub(%r((<CardNumber>)\d+(</CardNumber>)), '\1[FILTERED]\2').
           gsub(%r((<Cvc2>)\d+(</Cvc2>)), '\1[FILTERED]\2')
       end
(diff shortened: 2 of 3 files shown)


test/remote/gateways/remote_payment_express_test.rb
```diff
@@ -135,9 +135,10 @@ def test_transcript_scrubbing
       @gateway.purchase(@amount, @credit_card, @options)
     end
     clean_transcript = @gateway.scrub(transcript)
-
+
```
bpollack: Kill if you've got a second.

test/remote/gateways/remote_payment_express_test.rb
```diff
@@ -135,9 +135,10 @@ def test_transcript_scrubbing
       @gateway.purchase(@amount, @credit_card, @options)
     end
     clean_transcript = @gateway.scrub(transcript)
-
+    
     assert_scrubbed(@credit_card.number, clean_transcript)
     assert_scrubbed(@credit_card.verification_value.to_s, clean_transcript)
+    assert_scrubbed(fixtures(:payment_express)[:password].to_s, clean_transcript)
```
bpollack: This can just be `assert_scrubbed(@gateway.options[:password]...`, which would make it consistent with other remote tests that do this (see e.g. authorizenet)

test/remote/gateways/remote_payment_express_test.rb
```diff
@@ -138,6 +138,7 @@ def test_transcript_scrubbing
 
     assert_scrubbed(@credit_card.number, clean_transcript)
     assert_scrubbed(@credit_card.verification_value.to_s, clean_transcript)
+    assert_scrubbed(fixtures(:payment_express)[:password].to_s, clean_transcript)
```
bpollack: This can just be `assert_scrubbed(@gateway.options[:password]...`, which would make it consistent with other remote tests that do this (see e.g. authorizenet)

dtykocki:
>  There are many failing remote tests but they are not related to this change (they fail on
master as well)

@curiousepic - Remote tests did not fail for me on master when I ran them. Did you use the credentials we have on file?

test/remote/gateways/remote_payment_express_test.rb
```diff
@@ -138,6 +138,7 @@ def test_transcript_scrubbing
 
     assert_scrubbed(@credit_card.number, clean_transcript)
     assert_scrubbed(@credit_card.verification_value.to_s, clean_transcript)
+    assert_scrubbed(@gateway.options[:password].to_s, clean_transcript)
```
bpollack: Okay, I'll drop this at this point, but both the fixture and `@gateway.options[:password]` are already strings, so the `.to_s` isn't needed. I know we're trying to patch this quickly, but I'd rather take a few extra moments to make sure the code doesn't suffer than swing back later.

bpollack:
@dtykocki I also gets tons failing. I'm using the creds in 1Password under "Payment Express"; do those not match yours?

curiousepic:
@dtykocki Strange, yes I am using those. Perhaps it's an IP whitelisting issue?

dtykocki:
Okaaayy weird... remote test are now failing on master even with the creds... Carry on

PR #2725, opened 2018-01-26
Adds support for Google Pay sourced cards

joshnuss:
Adds Google Pay to list of allowed sources.

- Updated `NetworkTokenizedCard::SOURCES` to include `:google_pay`


pi3r:
@Krystosterone not as long as it's still supported by Google.

PR #2728, opened 2018-01-31
Adyen: Return refusal_reason_raw when present

curiousepic:
Remote:
28 tests, 65 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
17 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -205,11 +205,15 @@ module ActiveMerchant #:nodoc:
       end
 
       def message_from(action, response)
-        case action.to_s
-        when 'authorise'
+        return authorize_message_from(response) if action.to_s == 'authorise'
+        response['response'] || response['message']
+      end
+
+      def authorize_message_from(response)
+        if response['refusalReason'] && response['additionalData'] && response['additionalData']['refusalReasonRaw']
+          "#{response['refusalReason']} | #{response['additionalData']['refusalReasonRaw']}"
+        else
           response['refusalReason'] || response['resultCode'] || response['message']
-        when 'capture', 'refund', 'cancel'
-          response['response'] || response['message']
         end
       end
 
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -207,7 +207,11 @@ def success_from(action, response)
       def message_from(action, response)
         case action.to_s
         when 'authorise'
-          response['refusalReason'] || response['resultCode'] || response['message']
+          if response['refusalReason'] && response['additionalData'] && response['additionalData']['refusalReasonRaw']
+            "#{response['refusalReason']} | #{response['additionalData']['refusalReasonRaw']}"
+          else
+            response['refusalReason'] || response['resultCode'] || response['message']
+          end
```
curiousepic: I tried a few iterations of this and gave up on making it more elegant, so suggestions are welcome.

dtykocki: @curiousepic - To start off, you could eliminate the `case` statement use an if/else on the action instead instead - there's the 'authorize` action and then every other action.

dtykocki: Because of the nested hash access, there isn't much else you can do right now. Once we're on Ruby 2.3, `.dig` should help here http://ruby-doc.org/core-2.3.0/Hash.html#method-i-dig

dtykocki: You could take this a step further by isolating moving the nested hash logic into their own methods. Something like:
```
if action.to_s == 'authorize'
  authorization_response_from(response)
else
  other_response_from(response)
end
```

This at least makes things a bit more descriptive.

curiousepic: Sounds good, thanks 👍

curiousepic:
@dtykocki Cleaned up a bit.

PR #2733, opened 2018-02-06
Payeezy: Update Store

nfarve:
This updates the store method based on changes in the Payeezy
documentation taking effect at the end of March.

Loaded suite test/remote/gateways/remote_payeezy_test
...........................

27 tests, 98 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/unit/gateways/payeezy_test
................................

32 tests, 153 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payeezy.rb
+++ b/lib/active_merchant/billing/gateways/payeezy.rb
@@ -136,16 +136,10 @@ module ActiveMerchant
 
       def add_creditcard_for_tokenization(params, payment_method, options)
         params[:apikey] = @options[:apikey]
-        params[:js_security_key] = options[:js_security_key]
         params[:ta_token] = options[:ta_token]
-        params[:callback] = 'Payeezy.callback'
         params[:type] = 'FDToken'
-        card = add_card_data(payment_method)
-        params['credit_card.type'] = card[:type]
-        params['credit_card.cardholder_name'] = card[:cardholder_name]
-        params['credit_card.card_number'] = card[:card_number]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -140,12 +140,8 @@ def add_creditcard_for_tokenization(params, payment_method, options)
         params[:ta_token] = options[:ta_token]
         params[:callback] = 'Payeezy.callback'
```
dtykocki: According to https://developer.payeezy.com/payeezy-api/apis/post/transactions/tokens we don't need to send in the `callback` and `js_security_key`.

dtykocki: Also, there's a strong possibility we'll no longer need to send in the the `apikey` param in `add_creditcard_for_tokenization` anymore https://github.com/activemerchant/active_merchant/blob/ccca1e2e72c3277c1257e27c911acf0d63b95d66/lib/active_merchant/billing/gateways/payeezy.rb#L137

It's worth looking into this since we *should* already be performing the authentication Payeezy expects in https://github.com/activemerchant/active_merchant/blob/ccca1e2e72c3277c1257e27c911acf0d63b95d66/lib/active_merchant/billing/gateways/payeezy.rb#L301-L312

nfarve: @dtykocki thanks for the catch. Will ✂️ that.

PR #2743, opened 2018-02-12
Optimal payments add scrub

curiousepic:
Remote:
15 tests, 71 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
19 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_optimal_payment_test.rb
+++ b/test/remote/gateways/remote_optimal_payment_test.rb
@@ -143,4 +143,16 @@ class RemoteOptimalPaymentTest < Test::Unit::TestCase
     assert_failure response
     assert_equal 'invalid merchant account', response.message
   end
+
+  # Password assertion hard-coded due to the value being the same as the login, which would cause a false-positive
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+    transcript = @gateway.scrub(transcript)
+
+    assert_scrubbed(@credit_card.number, transcript)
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_optimal_payment_test.rb
```diff
@@ -143,4 +143,15 @@ def test_invalid_login
...
+
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+    transcript = @gateway.scrub(transcript)
+
+    assert_scrubbed(@credit_card.number, transcript)
+    assert_scrubbed(@credit_card.verification_value, transcript)
+    assert_scrubbed('%3CstorePwd%3Etest%3C/storePwd%3E', transcript)
```
curiousepic: Hard coding this isn't... optimal, but since the `login` and `password` creds are the same, this will flag a false positive for not scrubbing the login.

bpollack: If you don't want to add a comment here about why you're doing it this way, it might be worth explicitly `CGI.encode`ing it in-place (or equivalent) to make more obvious what you're doing. Just an idea, though.

curiousepic:
@bpollack Updated with the alternation for the potentially double-escaped slash we discussed offline, and added a comment for the hard-coded test assertion.

curiousepic:
And unit test added. One more thumb and I'll ship.

PR #2749, opened 2018-02-13
JetPay: fix garbled action code messages

reynhout:
Trivial fix of message typo. 

Reference: https://www.jetpay.com/doc-library/dev-docs/xml/JetPay-Host-XML-Specification-V2.03.01.pdf


bpollack:
Thanks @reynhout ! Merged to master.

PR #2755, opened 2018-02-20
Barclaycard Smartpay: Georgia is no longer supported

bpollack:


bpollack:
The irony is that "Georgia is no longer supported" is also true politically. 😢

PR #2762, opened 2018-02-26
Orbital: Scrub profile transactions

curiousepic:
Previously, credit card number and merchant ID were not scrubbed from
profile transactions since they use different field names.

Remote:
23 tests, 144 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
69 tests, 418 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


curiousepic:
FWIW, I don't think we actually support profile transactions.

PR #2766, opened 2018-02-28
Spreedly: Add verify and find operations

abarrak:
**Remote:**
31 tests, 146 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

**Unit:**
24 tests, 134 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


.gitignore
```diff
@@ -10,3 +10,4 @@ Gemfile*.lock
 .rvmrc
 /doc/
 .ruby-version
+.ruby-gemset
```
dtykocki: This can be removed. Instead of adding this to the project's .gitignore file, you can add to a global .gitignore that's local to your workstation. See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore for the details.

lib/active_merchant/billing/gateways/spreedly_core.rb
```diff
@@ -44,7 +44,7 @@ def purchase(money, payment_method, options = {})
           purchase_with_token(money, payment_method, options)
         else
           MultiResponse.run do |r|
-            r.process { save_card(false, payment_method, options) }
+            r.process { save_card(options[:store], payment_method, options) }
```
dtykocki: I like the direction you've gone in with `save_card`. However, I do think that there's an additional refactoring opportunity here. Consider removing `options[:store]` as an argument to `save_card` all together and leverage the existing `options` hash instead. The `save_card` method could then be modified to only accept two arguments and determine when the retained attributed should be true:
```
def save_card(credit_card, options)
  ......existing code.....
  doc.retained(true) if options[:retain]
end
```

Additionally, instead of using `store` as key, `retain` is more appropriate when working our API. Using `store` may lead to confusion with the `store` method itself.

Thoughts?

lib/active_merchant/billing/gateways/spreedly_core.rb
```diff
@@ -128,22 +152,32 @@ def authorize_with_token(money, payment_method_token, options)
         commit("gateways/#{@options[:gateway_token]}/authorize.xml", request)
       end
 
+      def verify_with_token(payment_method_token, options)
+        request = build_xml_request('transaction') do |doc|
+          add_invoice(doc, nil, options)
+          doc.payment_method_token(payment_method_token)
+          doc.retain_on_success(true) if options[:store]
```
dtykocki: This should also be `if options[:retain]`

lib/active_merchant/billing/gateways/spreedly_core.rb
```diff
@@ -128,22 +152,32 @@ def authorize_with_token(money, payment_method_token, options)
...
+        commit("gateways/#{@options[:gateway_token]}/verify.xml", request)
+      end
+
       def auth_purchase_request(money, payment_method_token, options)
         build_xml_request('transaction') do |doc|
           add_invoice(doc, money, options)
-          doc.ip(options[:ip])
           add_extra_options(:gateway_specific_fields, doc, options)
           doc.payment_method_token(payment_method_token)
           doc.retain_on_success(true) if options[:store]
```
dtykocki: Annnd here too

lib/active_merchant/billing/gateways/spreedly_core.rb
abarrak: I've updated to maintain `:store` too.

Please advice @dtykocki if it's better to revert back.

dtykocki: @abarrak - Let's just go with using `store` instead of supporting both `store` and `retain`. I think it'll keep things clearer this way. Once you've made the change, we should be ready for a merge.

abarrak: Alright, I restored back old setup in new PR #2798.

PR #2769, opened 2018-03-01
SecureNet: Support scrub

curiousepic:
Remote (sandbox frequently times out, scrub tests pass):
12 tests, 24 assertions, 0 failures, 7 errors, 0 pendings, 0 omissions, 0 notifications
41.6667% passed

Unit: 19 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


transcript.log
```diff
@@ -0,0 +1,18 @@
...
+read 2547 bytes
+Conn close
```
dtykocki: @curiousepic - Should this file be committed? I'm not seeing others like this one.

curiousepic: Whoops, good catch.

curiousepic: Removed.

PR #2772, opened 2018-03-01
Visanet Peru: Refund on unsettled transactions

nfarve:
When transactions are not yet settled a refundSingleTransaction needs to
be called. This tries to use this method if a cancelDeposit fails. If
both calls are made the messages from both transactions are captured. .

Loaded suite test/unit/gateways/visanet_peru_test
.............

13 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Loaded suite test/remote/gateways/remote_visanet_peru_test
.................

17 tests, 56 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/visanet_peru.rb
+++ b/lib/active_merchant/billing/gateways/visanet_peru.rb
@@ -34,25 +34,32 @@ module ActiveMerchant #:nodoc:
         params[:email] = options[:email] || 'unknown@email.com'
         params[:createAlias] = false
 
-        commit("authorize", params)
+        commit("authorize", params, options)
       end
 
       def capture(authorization, options={})
         params = {}
+        options[:id_unico] = split_authorization(authorization)[1]
         add_auth_order_id(params, authorization, options)
-        commit("deposit", params)
+        commit("deposit", params, options)
       end
 
       def void(authorization, options={})
         params = {}
         add_auth_order_id(params, authorization, options)
-        commit("void", params)
+        commit("void", params, options)
       end
 
       def refund(amount, authorization, options={})
         params = {}
+        params[:amount] = amount(amount) if amount
         add_auth_order_id(params, authorization, options)
-        commit("cancelDeposit", params)
+        response = commit("cancelDeposit", params, options)
+        return response if response.success? || split_authorization(authorization).length == 1 || !options[:force_full_refund_if_unsettled]
+
+        # Attempt RefundSingleTransaction if unsettled
+        prepare_refund_data(params, authorization, options)
+        commit('refund', params, options)
       end
 
       def verify(credit_card, options={})
@@ -88,7 +95,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_auth_order_id(params, authorization, options)
-        params[:purchaseNumber] = authorization
+        purchase_number, _ = split_authorization(authorization)
+        params[:purchaseNumber] = purchase_number
         params[:externalTransactionId] = options[:order_id]
       end
 
@@ -118,22 +126,35 @@ module ActiveMerchant #:nodoc:
         params[:antifraud] = antifraud
       end
 
-      def commit(action, params)
+      def prepare_refund_data(params, authorization, options)
+        params.delete(:purchaseNumber)
+        params[:externalReferenceId] = params.delete(:externalTransactionId)
+        _, transaction_id = split_authorization(authorization)
+
+        options.update(transaction_id: transaction_id)
+        params[:ruc] = options[:ruc]
+      end
+
+      def split_authorization(authorization)
+        authorization.split('|')
+      end
+
+      def commit(action, params, options={})
         begin
-          raw_response = ssl_request(method(action), url(action, params), params.to_json, headers)
+          raw_response = ssl_request(method(action), url(action, params, options), params.to_json, headers)
           response = parse(raw_response)
         rescue ResponseError => e
           raw_response = e.response.body
-          response_error(raw_response)
+          response_error(raw_response, options, action)
         rescue JSON::ParserError
           unparsable_response(raw_response)
         else
           Response.new(
             success_from(response),
-            message_from(response),
+            message_from(response, options, action),
             response,
             :test => test?,
-            :authorization => authorization_from(params),
+            :authorization => authorization_from(params, response, options),
             :error_code => response["errorCode"]
           )
         end
@@ -146,20 +167,23 @@ module ActiveMerchant #:nodoc:
         }
       end
 
-      def url(action, params)
+      def url(action, params, options={})
         if (action == "authorize")
           "#{base_url}/#{@options[:merchant_id]}"
+        elsif (action == 'refund')
+          "#{base_url}/#{@options[:merchant_id]}/#{action}/#{options[:transaction_id]}"
         else
           "#{base_url}/#{@options[:merchant_id]}/#{action}/#{params[:purchaseNumber]}"
         end
       end
 
       def method(action)
-        (action == "authorize") ? :post : :put
+        (%w(authorize refund).include? action) ? :post : :put
       end
 
-      def authorization_from(params)
-        params[:purchaseNumber]
+      def authorization_from(params, response, options)
+        id_unico = response['data']['ID_UNICO'] || options[:id_unico]
+        "#{params[:purchaseNumber]}|#{id_unico}"
       end
 
       def base_url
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -34,25 +34,33 @@ def authorize(amount, payment_method, options={})
...
         params = {}
         add_auth_order_id(params, authorization, options)
-        commit("void", params)
+        commit("void", params, options)
       end
 
       def refund(amount, authorization, options={})
+        # Attempt RefundSingleTransaction if unsettled
         params = {}
+        params[:amount] = amount(amount)
```
curiousepic: Should this only be `if amount`, since it may not be provided if it's a full refund, and sending it as nil might cause an issue?

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -118,22 +127,35 @@ def add_antifraud_data(params, options)
...
+      def prepare_refund_data(params, authorization, options)
+        params.delete(:purchaseNumber)
+        params[:externalReferenceId] = params.delete(:externalTransactionId)
+        _ , transaction_id = parse_auth(authorization)
+
+        options.update(transaction_id: transaction_id)
+        params[:ruc] = @options[:ruc]
+      end
+
+      def parse_auth(authorization)
```
curiousepic: Might be nice to use the convention in most other gateways and call this method `split_auth`.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -34,25 +34,33 @@ def authorize(amount, payment_method, options={})
...
       end
 
       def refund(amount, authorization, options={})
+        # Attempt RefundSingleTransaction if unsettled
         params = {}
+        params[:amount] = amount(amount)
         add_auth_order_id(params, authorization, options)
-        commit("cancelDeposit", params)
+        response = commit("cancelDeposit", params, options)
+        return response if response.success? || parse_auth(authorization).length == 1
```
curiousepic: I'm not sure I follow this conditional. So we'd return a failed `cancelDeposit` response if the authorization used for the refund didn't have one of the two ID numbers? Could you clarify why?

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -34,25 +34,33 @@ def authorize(amount, payment_method, options={})
...
+
+
+        prepare_refund_data(params, authorization, options)
```
curiousepic: Hmm. _Incoming sophomoric opinion expression_: Instead of this method that mutates a bunch of stuff, I think it would be a bit cleaner to have an explicit method for building a `cancelDeposit` transaction and an explicit method for building a `refundSingleTransaction` transaction, and the `refund` method would then call `refundSingleTransaction` if `cancelDeposit` fails. It would also make it easier to call one or the other individually, if we need to do that in the future for some reason. 

Maybe premature optimization, and I don't know if you looked into that already and it wasn't as simple as it sounds, but consider it if not?

lib/active_merchant/billing/gateways/visanet_peru.rb
nfarve: We can't process a refund without the id that is the second part of the ID numbers. Trying to do so gives a new error. I felt it wasn't worth making the second call since we don't have the key piece of information we would need.

lib/active_merchant/billing/gateways/visanet_peru.rb
nfarve: Definitely will consider. I generally try to touch others code as little as possible, but definitely agree with your opinion.

lib/active_merchant/billing/gateways/visanet_peru.rb
nfarve: I haven't touched this. Seems like its currently not causing an issue.

lib/active_merchant/billing/gateways/visanet_peru.rb
curiousepic: Gotcha, that make sense.

lib/active_merchant/billing/gateways/visanet_peru.rb
dtykocki: Hmm it looks like this was modified to pass in the amount. The current `refund` method on master ignores the amount: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/visanet_peru.rb#L52-L56. I'm thinking we'll need a guard around the `amount` here.

nfarve: @curiousepic sorry I thinking about the `params[:amount]` in add_invoice. Updated to have the if statement.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -118,22 +127,35 @@ def add_antifraud_data(params, options)
...
       end
 
-      def commit(action, params)
+      def prepare_refund_data(params, authorization, options)
+        params.delete(:purchaseNumber)
+        params[:externalReferenceId] = params.delete(:externalTransactionId)
+        _ , transaction_id = split_authorization(authorization)
+
+        options.update(transaction_id: transaction_id)
+        params[:ruc] = @options[:ruc]
```
dtykocki: What's the rationale for passing `ruc` through `@options` instead of using the `options` hash?

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -174,23 +199,30 @@ def success_from(response)
         response["errorCode"] == 0
       end
 
-      def message_from(response)
+      def message_from(response, options, action)
         if empty?(response["errorMessage"]) || response["errorMessage"] == "[ ]"
-          response["data"]["DSC_COD_ACCION"]
+          "#{response["data"]["DSC_COD_ACCION"]}, #{options[:error_message]}" if action == 'refund'
+          response['data']['DSC_COD_ACCION']
```
dtykocki: This will always return `response['data']['DSC_COD_ACCION']`. Give the modification on line 204, is seems like we don't want to always return `response['data']['DSC_COD_ACCION']`.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -174,23 +199,30 @@ def success_from(response)
...
         else
-          response["errorMessage"]
+          if action == 'refund'
+            message = "#{response["errorMessage"]}, #{options[:error_message]}"
+            options[:error_message] = response["errorMessage"]
```
dtykocki: Is this line needed? `options` is not referenced or returned anywhere after this assignment.

nfarve: In this new configuration, message_from will be called twice (once for cancelDeposit and once for refundSingleTransaction). If it fails in the first call, I want to save the error message so I can show it in the second call. I wanted to capture both messages incase these calls failed for a reason other than the transaction wasn't settled.

curiousepic:
@nfarve Pretty sure it applies in this case? The operative condition is that a refund is attempted but the transaction is found to be unsettled and so the initial refund method failed, so if this flag is present, we attempt whatever is needed to refund the unsettled transaction.

lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -34,25 +34,32 @@ def authorize(amount, payment_method, options={})
...
 
       def void(authorization, options={})
         params = {}
         add_auth_order_id(params, authorization, options)
-        commit("void", params)
+        commit("void", params, options)
       end
 
       def refund(amount, authorization, options={})
+        # Attempt RefundSingleTransaction if unsettled
```
curiousepic: Nitpick, it would be clearer for this to be on line 60.

curiousepic:
So it doesn't look like the new functionality is exercised in any test, is that right? Would it be possible to mock a "transaction is unsettled" gateway response for a unit test?

nfarve:
@curiousepic Yea we worked with visanet, but there is no way to test it out. Cabify has agreed to test it in production for us.

PR #2776, opened 2018-03-06
Litle: Add Support for Echeck

nfarve:
Adds support for Echeck and tests for certification.

Loaded suite test/remote/gateways/remote_litle_test
...................................

35 tests, 136 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Loaded suite test/unit/gateways/litle_test
Started
...................................

35 tests, 151 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -23,23 +23,33 @@ module ActiveMerchant #:nodoc:
       def purchase(money, payment_method, options={})
         request = build_xml_request do |doc|
           add_authentication(doc)
-          doc.sale(transaction_attributes(options)) do
-            add_auth_purchase_params(doc, money, payment_method, options)
+          if check?(payment_method)
+            doc.echeckSale(transaction_attributes(options)) do
+              add_echeck_purchase_params(doc, money, payment_method, options)
+            end
+          else
+            doc.sale(transaction_attributes(options)) do
+              add_auth_purchase_params(doc, money, payment_method, options)
+            end
           end
         end
-
-        commit(:sale, request, money)
+       check?(payment_method) ? commit(:echeckSales, request, money) : commit(:sale, request, money)
       end
 
       def authorize(money, payment_method, options={})
         request = build_xml_request do |doc|
           add_authentication(doc)
-          doc.authorization(transaction_attributes(options)) do
-            add_auth_purchase_params(doc, money, payment_method, options)
+          if check?(payment_method)
+            doc.echeckVerification(transaction_attributes(options)) do
+              add_echeck_purchase_params(doc, money, payment_method, options)
+            end
+          else
+            doc.authorization(transaction_attributes(options)) do
+              add_auth_purchase_params(doc, money, payment_method, options)
+            end
           end
         end
-
-        commit(:authorization, request, money)
+        check?(payment_method) ? commit(:echeckVerification, request, money) : commit(:authorization, request, money)
       end
 
       def capture(money, authorization, options={})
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -207,14 +251,21 @@ def add_debt_repayment(doc, options)
...
             doc.litleToken(payment_method)
           end
         elsif payment_method.respond_to?(:track_data) && payment_method.track_data.present?
           doc.card do
             doc.track(payment_method.track_data)
           end
+        elsif check?(payment_method)
+          doc.echeck do
+            doc.accType(payment_method.account_type)
+            doc.accNum(payment_method.account_number)
```
curiousepic: We'll want to add this field to the scrub method

nfarve: good catch.

test/remote/gateways/remote_litle_test.rb
```diff
@@ -78,6 +102,14 @@ def test_successful_authorization_with_merchant_data
...
+
+  def test_successful_authorization_with_echeck
+    options = @options.merge({
+      order_id: '38',
+      order_source: 'telephone'
+    })
+    assert response = @gateway.authorize(3002, @authorize_check, options)
     assert_success response
     assert_equal 'Approved', response.message
   end
```
curiousepic: I know Github can do weird things to how these diffs are displayed, but wanted to make sure that these lines 113 and 114 weren't missing from the above test, and same for 202/203

lib/active_merchant/billing/gateways/litle.rb
curiousepic: And the scrub tests too, preferably.

test/remote/gateways/remote_litle_test.rb
nfarve: Ooh look like it is. I think it was a merge conflict auto-merging thing. Will fix that.

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -62,19 +72,24 @@ def credit(money, authorization, options = {})
...
-
+      def refund(money, payment, options={})
         request = build_xml_request do |doc|
           add_authentication(doc)
           add_descriptor(doc, options)
-          doc.credit(transaction_attributes(options)) do
-            doc.litleTxnId(transaction_id)
-            doc.amount(money) if money
+          doc.send(refund_type(payment), transaction_attributes(options)) do
+            if payment.is_a?(String) || payment.is_a?(Numeric)
```
dtykocki: Why the need to check for a numeric type now?

lib/active_merchant/billing/gateways/litle.rb
nfarve: 👍 check and check

lib/active_merchant/billing/gateways/litle.rb
nfarve: @dtykocki in the tests it was giving tokens/auth codes as numbers and not strings

curiousepic: Shouldn't `authorization_from` stringify them?

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -62,19 +72,24 @@ def credit(money, authorization, options = {})
...
+              transaction_id, kind, _ = split_authorization(payment)
+              doc.litleTxnId(transaction_id)
+              doc.amount(money) if money
+            elsif check?(payment)
+              add_echeck_purchase_params(doc, money, payment, options)
+            else
+              add_auth_purchase_params(doc, money, payment, options)
+            end
```
dtykocki: Maybe it's the naming of the methods being used in here in `refund` (`add_echeck_purchase_params` and `add_auth_purchase_params`) but do refunds require the exact same params as echeck and auth purchases?

nfarve: @dtykocki Litle allows for "credits" of non-vantiv (litle) transactions. In those cases you give it the same info as a purchase and they "credit" the amount instead.

lib/active_merchant/billing/gateways/litle.rb
dtykocki: Yeah `authorization_from` should turn them into strings so I'm a bit 😕 here.

nfarve: @dtykocki and @curiousepic So I can change remote tests so they pass only strings and not numbers. Cases like this: https://github.com/activemerchant/active_merchant/blob/9885a343a0bd9a004b3a3cff9bb2f34357f53ad3/test/remote/gateways/remote_litle_test.rb#L320

dtykocki: @nfarve - Try that and maybe we'll be able to remove that number check.

nfarve:
@dtykocki and @curiousepic change all auth tokens to strings in remote test and got rid of test for numeric.

test/unit/gateways/litle_test.rb
```diff
@@ -45,6 +69,17 @@ def test_successful_purchase
     assert response.test?
   end
 
+  def test_successful_purchase_with_echeck
+      response = stub_comms do
+      @gateway.purchase(2004, @check)
+    end.respond_with(successful_purchase_with_echeck_response)
```
dtykocki: Nitpick: indentation is slightly off here.

PR #2778, opened 2018-03-08
Adyen: Support store action

curiousepic:
As part of implementation, the previous usage of the authorization_from
method and passing an unnecessary pspReference around was found to be
unnecessary, so was replaced with the split authorization pattern used
for store in other adapters. Also switches Verify to a 0 amount since it
is supported.

Remote:
100% passed

Unit:
19 tests, 91 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -45,21 +45,21 @@ module ActiveMerchant #:nodoc:
         add_invoice(post, money, options)
         add_payment(post, payment)
         add_extra_data(post, options)
-        add_shopper_interaction(post,payment,options)
+        add_shopper_interaction(post, payment, options)
         add_address(post, options)
         commit('authorise', post)
       end
 
       def capture(money, authorization, options={})
         post = init_post(options)
-        add_invoice_for_modification(post, money, authorization, options)
+        add_invoice_for_modification(post, money, options)
         add_references(post, authorization, options)
         commit('capture', post)
       end
 
       def refund(money, authorization, options={})
         post = init_post(options)
-        add_invoice_for_modification(post, money, authorization, options)
+        add_invoice_for_modification(post, money, options)
         add_references(post, authorization, options)
         commit('refund', post)
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -101,7 +112,14 @@ def add_extra_data(post, options)
       end
 
       def add_shopper_interaction(post, payment, options={})
-        shopper_interaction = payment.verification_value ?  "Ecommerce" : "ContAuth"
+        if payment.is_a?(String)
```
nfarve: Was there a reason to break this out?

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -70,9 +70,20 @@ def void(authorization, options={})
...
+        add_extra_data(post, options)
+        add_recurring(post, options)
+        add_address(post, options)
+        commit('authorise', post)
+      end
+
       def verify(credit_card, options={})
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
+          r.process { authorize(0, credit_card, options) }
```
dtykocki: I'm guessing that this change is intentional. $0 auths are supported now?

nfarve: I assumed so from the commit message

dtykocki: 🙈 I didn't read the message. Whoops.

lib/active_merchant/billing/gateways/adyen.rb
curiousepic: `payment.verification_value` errors when `payment` is a string, so I restructured it this way. There is probably a more concise formulation though... hmm.

curiousepic: The alternatives I've come up with are just less readable, for not much benefit. I'll go with this unless you have a specific suggestion.

curiousepic:
I kept having a nagging feeling that I was misinterpreting how the pspReference was being handled previously and I think I finally understand it and I think it needs to be preserved (despite the fact that the tests pass with my changes). I'll rework my changes to incorporate the previous behavior and resubmit.

lyverovski:
@curiousepic You can find some background on `pspReference` in #2589.  

I think the change you made works going forward, but breaks refunds for any existing captures that have `#` stored as their delimiter in the pspReference?

curiousepic:
Thanks for taking a look @lyverovski. I saw that commit but didn't take a look at the PR. I'm working on a revision that is theoretically backward compatible with those, changing the delimiter back to `#`.

However, a confounding factor is that (for better or worse) most adapters using split `authorization`s have a rigid ordering to their elements, and this implementation currently `.compact`s, which means that for any existing `authorization`s that don't have both elements present, there's no delimiter and we don't know if it's an original or normal psp reference. Any thoughts?

curiousepic:
@lyverovski Actually, I could account for the lack of delimiters by making captures and voids assume non-split authorizations are a normal psp reference, and refunds assume its an original psp reference. That seems like it would cover situations that are currently covered as tests...

curiousepic:
Revised PR https://github.com/activemerchant/active_merchant/pull/2784

lyverovski:
Sorry for the late reply @curiousepic.
I had made the same assumption in the original code. You were only dealing with a compound `authorization` value when handling refund transactions, which had to split the `authorization` that was compounded in the capture transaction response.

PR #2784, opened 2018-03-14
Adyen: Support store action

curiousepic:
As part of implementation, I preserved the behavior of retaining
original psp references for refunds, and account for backwards
compatability with pre-existing non-compound authorizations. Also
removes an irrelevant test case (a capture should never have a compound
authorization) and switches Verify to a 0 amount, since it is supported.

Remote:
100% passed

Unit:
18 tests, 86 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -45,34 +45,45 @@ module ActiveMerchant #:nodoc:
         add_invoice(post, money, options)
         add_payment(post, payment)
         add_extra_data(post, options)
-        add_shopper_interaction(post,payment,options)
+        add_shopper_interaction(post, payment, options)
         add_address(post, options)
         commit('authorise', post)
       end
 
       def capture(money, authorization, options={})
         post = init_post(options)
-        add_invoice_for_modification(post, money, authorization, options)
-        add_references(post, authorization, options)
+        add_invoice_for_modification(post, money, options)
+        add_reference(post, authorization, options)
         commit('capture', post)
       end
 
       def refund(money, authorization, options={})
         post = init_post(options)
-        add_invoice_for_modification(post, money, authorization, options)
-        add_references(post, authorization, options)
+        add_invoice_for_modification(post, money, options)
+        add_original_reference(post, authorization, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -45,34 +45,45 @@ def authorize(money, payment, options={})
...
         commit('cancel', post)
       end
 
+      def store(credit_card, options={})
+        requires!(options, :order_id)
+        post = init_post(options)
+        add_invoice(post, 0, options)
+        add_payment(post, credit_card)
+        add_extra_data(post, options)
+        add_recurring_contract(post, options)
```
dtykocki: `add_recurring_contract` is called here an in `add_payment` is that intentional?

curiousepic: Yes, it needs to be added both when a card is sent to be stored, and when a stored card is used for a transaction. No better way to add them in both situations is occurring to me right now but I'm open to suggestions.

dtykocki: Fair enough. I don't have an alternative that would be any better. 👍

test/unit/gateways/adyen_test.rb
```diff
@@ -126,12 +119,26 @@ def test_failed_void
     assert_failure response
   end
 
+  def test_successful_store
+    @gateway.expects(:ssl_post).returns(successful_store_response)
+      response = @gateway.store(@credit_card, @options)
```
dtykocki: Nitpick: indentation is off

test/unit/gateways/adyen_test.rb
```diff
@@ -126,12 +119,26 @@ def test_failed_void
...
+    assert_success response
+    assert_equal "#8835205392522157#8315202663743702", response.authorization
+  end
+
+  def test_failed_store
+    @gateway.expects(:ssl_post).returns(failed_store_response)
+      response = @gateway.store(@credit_card, @options)
```
dtykocki: Also nitpick.

PR #2791, opened 2018-03-22
Mundipagg: New Gateway Implementation

nfarve:
Adds new gateway Mundipagg. Adds two new vouchers as credit cards
(Sodexo and VR). Allows delete requests to pass body parameters.

Loaded suite test/unit/gateways/mundipagg_test
...............

15 tests, 66 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Started
....................

20 tests, 50 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/mundipagg.rb
+++ b/lib/active_merchant/billing/gateways/mundipagg.rb
@@ -1,0 +1,289 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class MundipaggGateway < Gateway
+      self.live_url = 'https://api.mundipagg.com/core/v1'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'https://www.mundipagg.com/'
+      self.display_name = 'Mundipagg'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        '400' => STANDARD_ERROR_CODE[:processing_error],
+        '401' => STANDARD_ERROR_CODE[:config_error],
+        '404' => STANDARD_ERROR_CODE[:processing_error],
+        '412' => STANDARD_ERROR_CODE[:processing_error],
+        '422' => STANDARD_ERROR_CODE[:processing_error],
+        '500' => STANDARD_ERROR_CODE[:processing_error]
+      }
+
+      STANDARD_ERROR_MESSAGE_MAPPING = {
+        '400' => 'Invalid request;',
+        '401' => 'Invalid API key;',
+        '404' => 'The requested resource does not exist;',
+        '412' => 'Valid parameters but request failed;',
+        '422' => 'Invalid parameters;',
+        '500' => 'An internal error occurred;'
+      }
+
+      def initialize(options={})
+        requires!(options, :api_key)
+        super
+      end
+
+      def purchase(money, payment, options={})
+        post = {}
+        add_invoice(post, money, options)
+        add_customer_data(post, options) unless payment.is_a?(String)
+        add_shipping_address(post, options)
+        add_payment(post, payment, options)
+
+        commit('sale', post)
+      end
+
+      def authorize(money, payment, options={})
+        post = {}
+        add_invoice(post, money, options)
+        add_customer_data(post, options) unless payment.is_a?(String)
+        add_shipping_address(post, options)
+        add_payment(post, payment, options)
+        add_capture_flag(post, payment)
+        commit('authonly', post)
+      end
+
+      def capture(money, authorization, options={})
+        post = {}
+        post[:code] = authorization
+        add_invoice(post, money, options)
+        commit('capture', post, authorization)
+      end
+
+      def refund(money, authorization, options={})
+        add_invoice(post={}, money, options)
+        commit('refund', post, authorization)
+      end
+
+      def void(authorization, options={})
+        commit('void', post=nil, authorization)
+      end
+
+      def store(payment, options={})
+        post = {}
+        options.update(name: payment.name)
+        options = add_customer(options) unless options[:customer_id]
+        add_payment(post, payment, options)
+        commit('store', post, options[:customer_id])
+      end
+
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript
+          .gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]')
+          .gsub(%r(("cvv\\":\\")\d*), '\1[FILTERED]')
+          .gsub(%r((card\\":{\\"number\\":\\")\d*), '\1[FILTERED]')
+      end
+
+      private
+
+      def add_customer(options)
+        post = {}
+        post[:name] = options[:name]
+        customer = commit('customer', post)
+        options.update(customer_id: customer.authorization)
+      end
+
+      def add_customer_data(post, options)
+        post[:customer] = {}
+        post[:customer][:email] = options[:email]
+      end
+
+      def add_billing_address(options)
+        billing = {}
+        address = options[:billing_address] || options[:address]
+        billing[:street] = address[:address1].match(/\D+/)[0].strip if address[:address1]
+        billing[:number] = address[:address1].match(/\d+/)[0] if address[:address1]
+        billing[:compliment] = address[:address2] if address[:address2]
+        billing[:city] = address[:city] if address[:city]
+        billing[:state] = address[:state] if address[:state]
+        billing[:country] = address[:country] if address[:country]
+        billing[:zip_code] = address[:zip] if address[:zip]
+        billing[:neighborhood] = address[:neighborhood]
+        billing
+      end
+
+      def add_shipping_address(post, options)
+        if address = options[:shipping_address]
+          post[:address] = {}
+          post[:address][:street] = address[:address1].match(/\D+/)[0].strip if address[:address1]
+          post[:address][:number] = address[:address1].match(/\d+/)[0] if address[:address1]
+          post[:address][:compliment] = address[:address2] if address[:address2]
+          post[:address][:city] = address[:city] if address[:city]
+          post[:address][:state] = address[:state] if address[:state]
+          post[:address][:country] = address[:country] if address[:country]
+          post[:address][:zip_code] = address[:zip] if address[:zip]
+        end
+      end
+
+      def add_invoice(post, money, options)
+        post[:amount] = money
+        post[:currency] = (options[:currency] || currency(money))
+      end
+
+      def add_capture_flag(post, payment)
+        if voucher?(payment)
+          post[:payment][:voucher][:capture] = false
+        else
+          post[:payment][:credit_card][:capture] = false
+        end
+      end
+
+      def add_payment(post, payment, options)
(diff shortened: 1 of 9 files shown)


curiousepic:
I'm remembering some previous discussion on whether deletes should allow a body, with the consensus being that they shouldn't. See https://github.com/activemerchant/active_merchant/pull/2326 and https://github.com/activemerchant/active_merchant/pull/2348
I don't know the details on this though (possibly a Ruby/gem version issue?) which may have changed since, so perhaps it's good to support now?

nfarve:
@curiousepic hmmmm might need to add a new http method

dtykocki:
As of Ruby 2.3.6, `Net::HTTP`, the library used make gateway calls, does not allow for a body when making a delete: http://ruby-doc.org/stdlib-2.3.0/libdoc/net/http/rdoc/Net/HTTP.html#method-i-delete

dtykocki:
Huh.. it actually looks like we're override the call in https://github.com/shasum/active_merchant/blob/714a1153554ce97f7ce4e1d3294b9c212d97af58/lib/active_merchant/delete_with_body.rb

dtykocki:
@nfarve You'll need to test it out, but I think you can try something like https://github.com/activemerchant/active_merchant/pull/2326/files#diff-154e23ec35747841db7f46d7dd8f137aR156.

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,288 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class MundipaggGateway < Gateway
+      self.test_url = 'https://api.mundipagg.com/core/v1/'
+      self.live_url = 'https://api.mundipagg.com/core/v1/'
```
dtykocki: @nfarve - If the gateway doesn't provide a sandbox with a test URL, there's not much point in have both URLs. The `test_url` and conditional logic around the `live_url` can be removed with there's no sandbox. Thoughts?

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,288 @@
...
+      }
+
+      STANDARD_ERROR_MESSAGE_MAPPING = {
+          '400' => 'Invalid request;',
+          '401' => 'Invalid API key;',
+          '404' => 'The requested resource does not exist;',
+          '412' => 'Valid parameters but request failed;',
+          '422' => 'Invalid parameters;',
+          '500' => 'An internal error occurred;'
+      }
```
dtykocki: Nitpick: Indentation is in the hash is different than the rest.

nfarve:
@dtykocki Yea I think `delete_with_body` was removed. I would have to reimplement it. The current code I pushed works though.

nfarve:
@curiousepic and @dtykocki added back `delete_with_body` and a test for it!

dtykocki:
@curiousepic - Did you mean to close this one out?

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,287 @@
...
+        customer = create_customer(options)
+        options.update(customer_id: customer.authorization)
+      end
+
+      def add_customer_data(post, options)
+        post[:customer] = {}
+        post[:customer][:email] = options[:email]
+      end
+
+      def build_billing_address(options)
```
dtykocki: For consistency sake, name this method `add_billing_address`?

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,287 @@
...
+        billing = {}
+        address = options[:billing_address] || options[:address]
+        billing[:street] = address[:address1].match(/\D+/)[0].strip if address[:address1]
+        billing[:number] = address[:address1].match(/\d+/)[0] if address[:address1]
+        billing[:compliment] = address[:address2] if address[:address2]
+        billing[:city] = address[:city] if address[:city]
+        billing[:state] = address[:state] if address[:state]
+        billing[:country] = address[:country] if address[:country]
+        billing[:zip_code] = address[:zip] if address[:zip]
+        billing[:neighborhood] = address[:neighborhood]
```
dtykocki: Should this line have `if address[:neighborhood]`?

lib/active_merchant/billing/gateways/mundipagg.rb
nfarve: So I had it that way and changed it since its "building" the billing address object and not really "adding" it. Maybe I am being too literal.

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,287 @@
...
+        '400' => 'Invalid request;',
+        '401' => 'Invalid API key;',
+        '404' => 'The requested resource does not exist;',
+        '412' => 'Valid parameters but request failed;',
+        '422' => 'Invalid parameters;',
+        '500' => 'An internal error occurred;'
+      }
+
+      def initialize(options={})
+        requires!(options, :api_key)
```
dtykocki: I noticed that the `fixtures.yml` contains an entry for `gateway_id`. Any particular reason we don't require `gateway_id` here?

nfarve: Its a field they use the track with. They only asked us to include it when were trying to debug how to send the voucher requests correctly.

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,287 @@
...
+
+      def url_for(action, auth = nil)
+        url = live_url
+        return "#{url}customers/#{auth}/cards" if action == 'store'
+        return "#{url}customers" if action == 'customer'
+        url = "#{url}charges/"
+        url =  "#{url}#{auth}/" if %w[refund void capture].include? action
+        return "#{url}capture/" if action == 'capture'
+        url
+      end
```
dtykocki: Determining which URL to use would be better suite with a `case` statement here. Additionally, if the trailing slash is removed from https://github.com/activemerchant/active_merchant/pull/2791/files#diff-d6e4d483fbc1ae8bde58049c05c40018R4, it'll be easier to read which endpoint is being used:
```
"#{url}/customers/#{auth}/cards"
```
vs
```
"#{url}customers/#{auth}/cards"
```

lib/active_merchant/connection.rb
```diff
@@ -82,6 +82,11 @@ def request(method, body, headers = {})
               # very unambiguously does not.
               raise ArgumentError, "DELETE requests do not support a request body" if body
               http.delete(endpoint.request_uri, headers)
+            when :delete_with_body
+              debug body
```
dtykocki: This should probably be deleted.

nfarve: This is actually the standard for any request with a `body`

https://github.com/activemerchant/active_merchant/blob/3b428287e5f2d96d41c986abe0a83fb72032e2c9/lib/active_merchant/connection.rb#L77

dtykocki: Neat, I see that now.

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,287 @@
...
+        options = add_customer(options) unless options[:customer_id]
+        add_payment(post, payment, options)
+        commit('store', post, options[:customer_id])
+      end
+
+      def create_customer(options={})
+        post = {}
+        post[:name] = options[:name]
+        commit('customer', post)
+      end
```
dtykocki: We typically don't introduce new methods to AM's public API. I can't say for certain that it hasn't happened before, but I believe it's discouraged. When a gateway requires additional setup, like creating a customer, the process is encapsulated in the public API . So instead of increasing the public footprint, would it be possible to make `create_customer` private or move the logic to `add_customer`?

@curiousepic - Any thoughts here? You've worked with AM longer than I have.

nfarve: I can make it private.

lib/active_merchant/billing/gateways/mundipagg.rb
nfarve: I don't think its just for testing. I think it similar to when we want to monitor traffic from a specific merchant. Or when a gateway wants to track Spreedly traffic.

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -0,0 +1,289 @@
...
+          "#{url}/customers/#{auth}/cards/"
+        when 'customer'
+          "#{url}/customers/"
+        when 'refund', 'void'
+          "#{url}/charges/#{auth}/"
+        when 'capture'
+          "#{url}/charges/#{auth}/capture/"
+        else
+         "#{url}/charges/"
+        end
```
dtykocki: Nice!

test/remote/gateways/remote_mundipagg_test.rb
```diff
@@ -0,0 +1,155 @@
...
+
+  def test_successful_void
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    assert void = @gateway.void(auth.authorization)
+    assert_success void
+  end
+
+  def test_successful_void_with_voucher
```
dtykocki: 👍 Can you add a test for a successful refund with a voucher?

nfarve: Done!

lib/active_merchant/connection.rb
```diff
@@ -82,6 +82,11 @@ def request(method, body, headers = {})
...
+              req = Net::HTTP::Delete.new(endpoint.request_uri, headers)
+              req.body = body
+              http.request(req)
```
dtykocki: @jasonwebster or @bdewater - we'd appreciate another set of eyes on this bit. This gateway requires a `DELETE` with a body when performing a `refund` or `void`. Any thoughts or concerns here? For some additional history, we almost went with the same approach in https://github.com/activemerchant/active_merchant/pull/2326, but the `delete_with_body` addition to  `lib/active_merchant/connection.rb` didn't land in master.

bdewater: Creating a 'pseudo' HTTP verb here doesn't feel right to me me. It also burdens the caller knowing about ambiguous specs and resulting implementation choices. Is there a strong reason against using these lines to replace the current `:delete` implementation?

nfarve: @bdewater our original implementation had it in the `:delete` but we thought this was safer. I don't think we would be against going back to the original

bdewater: As long as the default behaviour remains the same - not sending a body unless one is specified - I think we should be good. Everybody has programmed around that exception up until now.

nfarve:
@curiousepic @dtykocki  new and improved up with old and funky change to `:delete`

lib/active_merchant/connection.rb
```diff
@@ -80,8 +80,15 @@ def request(method, body, headers = {})
               # It's kind of ambiguous whether the RFC allows bodies
               # for DELETE requests. But Net::HTTP's delete method
               # very unambiguously does not.
-              raise ArgumentError, "DELETE requests do not support a request body" if body
-              http.delete(endpoint.request_uri, headers)
+              # raise ArgumentError, "DELETE requests do not support a request body" if body
```
bdewater: Nitpick: we can just completely remove this.

PR #2795, opened 2018-03-27
CardStream: Change `refund` to use `REFUND_SALE`

dtykocki:
Per CardStream, specifying `REFUND` as the action performs a general /
independent credit against the card which may result in chargebacks. Instead,
the `REFUND_SALE` action can be used refund a `SALE` transaction that has settled.
When a `SALE` transaction has not settled, a `CANCEL` request must be made that
will refund the original `SALE`.

Note: I initially found this behavior quite confusing but it turns out we perform this
pattern in the Authorize.net, WorldPay, and BraintreeBlue adapters. These changes follow
the same pattern.

Unit:
24 tests, 126 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
30 tests, 200 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/card_stream.rb
+++ b/lib/active_merchant/billing/gateways/card_stream.rb
@@ -184,7 +184,16 @@ module ActiveMerchant #:nodoc:
         add_pair(post, :xref, authorization)
         add_amount(post, money, options)
         add_remote_address(post, options)
-        commit('REFUND', post)
+        response = commit('REFUND_SALE', post)
+
+        return response if response.success?
+        return response unless options[:force_full_refund_if_unsettled]
+
+        if response.params["responseCode"] == "65541"
+          void(authorization, options)
+        else
+          response
+        end
       end
 
       def void(authorization, options = {})
--- a/test/remote/gateways/remote_card_stream_test.rb
+++ b/test/remote/gateways/remote_card_stream_test.rb
@@ -144,15 +144,29 @@ class RemoteCardStreamTest < Test::Unit::TestCase
     assert responseCapture.test?
   end
 
-  def test_successful_visacreditcard_purchase_and_refund
+  def test_successful_visacreditcard_purchase_and_refund_with_force_refund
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/card_stream.rb
```diff
@@ -184,7 +184,16 @@ def refund(money, authorization, options = {})
         add_pair(post, :xref, authorization)
         add_amount(post, money, options)
         add_remote_address(post, options)
-        commit('REFUND', post)
+        response = commit('REFUND_SALE', post)
+
+        return response if response.success?
+        return response unless options[:force_full_refund_if_unsettled]
```
curiousepic: What's the case for relying on this flag instead of voiding by default if unsettled?

dtykocki: My first pass at this didn't include the flag. I only added it in when I saw that other adapters (those mentioned in the commit message) use `force_full_refund_if_unsettled` to determine if void should be called. I tend to favor existing conventions for consistency sake, but I can see a case for not including the flag.

curiousepic: Interesting, I hadn't seen those in use before... It doesn't look like Spreedly utilizes them? 

I also recall another update recently that fixed a very similar issue where a transaction may be unsettled and need to be voided instead of refunded, but I can't seem to find it... I don't think that used a flag like this.

dtykocki: No, we don't utilize the the force flag.

Hmmm maybe you're thinking of https://github.com/activemerchant/active_merchant/pull/2772?

curiousepic: Ah, yes. So perhaps that should use this flag if we want to begin using it...

dtykocki: @curiousepic - Commented on the `force_full_refund_if_unsettled` flag on #2772. Any other concerns with this one?

PR #2796, opened 2018-03-28
Stripe: Add `exchange_rate` parameter

WilsonChiang:
This adds support for the optional `exchange_rate` parameter during
capture and purchase. This specifies an exchange rate to be used during
a transaction, and if incorrect or stale, Stripe will return a 400
`invalid_request_error` with an `invalid_exchange_rate` code. The
response also contains the updated rate in the event of a failure in the
`new_rate` parameter.

This is currently a preview feature.

Ref: https://stripe.com/docs/api#capture_charge

Unit:
126 tests, 667 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed

Remote:
64 tests, 291 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -128,6 +128,7 @@ module ActiveMerchant #:nodoc:
         else
           add_application_fee(post, options)
           add_amount(post, money, options)
+          add_exchange_rate(post, options)
           commit(:post, "charges/#{CGI.escape(authorization)}/capture", post, options)
         end
       end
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -320,6 +321,7 @@ def create_post_for_auth_or_purchase(money, payment, options)
 
         add_metadata(post, options)
         add_application_fee(post, options)
+        add_exchange_rate(post, options)
```
WilsonChiang: Wrestled with the idea of putting it here, or just where its called for a purchase.

PR #2799, opened 2018-03-30
Adyen: Support merchant-specific subdomains

curiousepic:
Adyen's sandbox doesn't support testing of these subdomains, so the only
way I found to test the functionality was by subverting the check for a
test url in the adapter itself, and checking the request in a unit test,
hence the 1 unit failure.

Happy to take suggestions on more robust methods to test this.

Remote:
100% passed

Unit:
19 tests, 89 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.7368% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -202,8 +202,6 @@ module ActiveMerchant #:nodoc:
       end
 
       def commit(action, parameters)
-        url = (test? ? test_url : live_url)
-
         begin
           raw_response = ssl_post("#{url}/#{action.to_s}", post_data(action, parameters), request_headers)
           response = parse(raw_response)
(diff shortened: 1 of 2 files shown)


curiousepic:
Revising this to make it a gateway-level option rather than options hash-level.

curiousepic:
Updated; would appreciate another look from @nfarve @dtykocki and/or @deedeelavinder

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -47,27 +47,27 @@ def authorize(money, payment, options={})
         add_extra_data(post, options)
         add_shopper_interaction(post, payment, options)
         add_address(post, options)
-        commit('authorise', post)
+        commit('authorise', post, options)
```
nfarve: If its gateway-level do you need these?

curiousepic: Nope, good catch!

curiousepic: Removed unnecessary options params.

Issue #2372, opened 2017-03-23
Respecting http_proxy environment variable


techbelly:
Net::Http respects http_proxy environment variables:

https://ruby-doc.org/stdlib-2.4.0/libdoc/net/http/rdoc/Net/HTTP.html#method-c-new

"If none of the p_ arguments are given, the proxy host and port are taken from the http_proxy environment variable (or its uppercase equivalent) if present. If the proxy requires authentication you must supply it by hand. See URI::Generic#find_proxy for details of proxy detection from the environment. To disable proxy detection set p_addr to nil."

Because ActiveMerchant always gives p_ arguments to the constructor of Net:Http, this behaviour is, I think, unintentionally circumvented.

See pull request: #2371 for a potential fix.

georgiev:
Please reopen - the issue is still a valid one in both - the master and the v1.95.0 tag. See PR #3244 for description.

PR #2800, opened 2018-04-04
Fix ENV based configuration of Net::Http for proxies

bbergstrom:
Net::Http has a slightly odd constructor semantics that aren't
immediately obvious. If no proxy parameters are sent then, by default, it
looks up the proxy from the standard HTTP_PROXY/http+proxy environment
variables. To set no proxy at all, ignoring the environment variable, you can pass nil
as the p_address parameter.

ActiveMerchant::Connection was always passing proxy parameters
to Net::Http.new even when a proxy had not been set, meaning that
the default environment variable configuration behaviour was never respected.

This commit ensures that if you do not explicitly set proxy parameters
on your Gateway the default behaviour of Net::Http to retrieve proxy
settings from the environment applies. Proxy parameters can still explicitly
set to nil to avoid environment variable configuration or to desired proxy values.

Ref: https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html#method-c-new
Fixes: https://github.com/activemerchant/active_merchant/issues/2372
Revision of dormant PR: https://github.com/activemerchant/active_merchant/pull/2371

### DIFF
--- a/test/unit/connection_test.rb
+++ b/test/unit/connection_test.rb
@@ -26,6 +26,25 @@ class ConnectionTest < Test::Unit::TestCase
     end
   end
 
+  def test_connection_passes_env_proxy_by_default
+    @connection.logger.expects(:info).twice
+
+    spy = Net::HTTP.new('example.com', 443)
+    Net::HTTP.expects(:new).with('example.com', 443, :ENV, nil).returns(spy)
+    spy.expects(:get).with('/tx.php', {}).returns(@ok)
+    response = @connection.request(:get, nil, {})
+  end
+
+  def test_connection_does_pass_requested_proxy
+    @connection.logger.expects(:info).twice
+    @connection.proxy_address = "proxy.example.com"
+    @connection.proxy_port = 8080
+    spy = Net::HTTP.new('example.com', 443)
+    Net::HTTP.expects(:new).with('example.com', 443, "proxy.example.com", 8080).returns(spy)
+    spy.expects(:get).with('/tx.php', {}).returns(@ok)
+    response = @connection.request(:get, nil, {})
+  end
+
   def test_successful_get_request
     @connection.logger.expects(:info).twice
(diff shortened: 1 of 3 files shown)


test/unit/connection_test.rb
```diff
@@ -26,6 +26,25 @@ def test_connection_endpoint_raises_uri_error
...
+    @connection.logger.expects(:info).twice
+
+    spy = Net::HTTP.new('example.com', 443)
+    Net::HTTP.expects(:new).with('example.com', 443, :ENV, nil).returns(spy)
+    spy.expects(:get).with('/tx.php', {}).returns(@ok)
+    response = @connection.request(:get, nil, {})
+  end
+
+  def test_connection_does_pass_requested_proxy
+    @connection.logger.expects(:info).twice
```
bdewater: This seems unrelated to what we're testing here, can we remove it? Same question for the other test.

test/unit/connection_test.rb
```diff
@@ -26,6 +26,25 @@ def test_connection_endpoint_raises_uri_error
...
+    @connection.proxy_address = "proxy.example.com"
+    @connection.proxy_port = 8080
+    spy = Net::HTTP.new('example.com', 443)
+    Net::HTTP.expects(:new).with('example.com', 443, "proxy.example.com", 8080).returns(spy)
+    spy.expects(:get).with('/tx.php', {}).returns(@ok)
+    response = @connection.request(:get, nil, {})
```
bdewater: Nitpick: we can remove the unused `response ` variable. Same for the other test.

bbergstrom:
Revised.

bbergstrom:
@bdewater How often does a release get cut of this gem?

bdewater:
More-or-less monthly, there's no firm schedule. You could always point your Gemfile to the merged commit on master if you need it now, the delta with the latest release is pretty small :)

PR #2801, opened 2018-04-04
Include gateway specific fields in Spreedly refund

swebb:
Extends commit: b1436c893

Refunds via Spreedly Core were not including extra options. As a result
the description seen by the customer was incorrect.


swebb:
Superseded by https://github.com/activemerchant/active_merchant/commit/86182b1189ec0818c0892e9429b47a7af7b74ff2

PR #2802, opened 2018-04-05
Remove Inquiry requests in verify transactions

nfarve:
Inquiries need at least 5 minutes after initial transaction in order to
be reliable. This removes inquiries from verify requests.

Errors are also caused by timeouts. Successful tests can be achieved by
adding a sleep, but doesn not produce consistent results.

Loaded suite test/unit/gateways/worldpay_test

-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
37 tests, 210 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Finished in 231.664812 seconds.
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
23 tests, 71 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
78.2609% passed
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

### DIFF
--- a/test/remote/gateways/remote_worldpay_test.rb
+++ b/test/remote/gateways/remote_worldpay_test.rb
@@ -43,6 +43,7 @@ class RemoteWorldpayTest < Test::Unit::TestCase
     assert_success auth
     assert_equal 'SUCCESS', auth.message
     assert auth.authorization
+    sleep(40)
     assert capture = @gateway.capture(@amount, auth.authorization)
     assert_success capture
   end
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -43,6 +43,7 @@ def test_authorize_and_capture
     assert_success auth
     assert_equal 'SUCCESS', auth.message
     assert auth.authorization
+    sleep(40)
```
dtykocki: Doh.. I suppose something changed on their sandbox that requires the sleep statement now.

nfarve: Yea but unfortunately it doesn't work when all the test run, but if I run the test one by one they pass.

PR #2807, opened 2018-04-12
Remove iDeal references

bdewater:
iDeal is a Dutch offsite payment method and belongs in https://github.com/activemerchant/offsite_payments


bpollack:
@bdewater Need an approver? I can take a look if so. (Or if this already merged as a different SHA…)

bdewater:
Kinda forgot about this one; just rebased so a review would be much appreciated!

PR #2810, opened 2018-04-18
Add documented country support for US and CA

a-salty-strudel:
Realex has let us know they've been successfully operating in the US and Canada for over a year and they'd like that reflected in AM, so I'm submitting my first pull request!

I know it's a little change but please let me know if anything's off about this. :)


a-salty-strudel:
I updated the unit tests to expect the new countries as well.

PR #2816, opened 2018-04-23
SafeCharge: Additional gateway options

dtykocki:
Adds merchant descriptor options that may be used to display on a
customer's credit card statement. Additionally, this adds support for
the new `sg_StoredCredentialMode` field. When `card_not_present` is
true, this tells SafeCharge that the transaction was completed
using a previously tokenized and stored card data. When `card_not_present`
is false (the default), this tells SafeCharge the card was entered
for the first time and will be stored for future transactions.

Remote:
21 tests, 62 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
20 tests, 96 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/safe_charge.rb
+++ b/lib/active_merchant/billing/gateways/safe_charge.rb
@@ -25,7 +25,7 @@ module ActiveMerchant #:nodoc:
         post[:sg_APIType] = 1 if options[:three_d_secure]
         trans_type = options[:three_d_secure] ? "Sale3D" : "Sale"
         add_transaction_data(trans_type, post, money, options)
-        add_payment(post, payment)
+        add_payment(post, payment, options)
         add_customer_details(post, payment, options)
 
         commit(post)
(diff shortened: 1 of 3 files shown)


curiousepic:
Do we know if the field should in fact be sent for all transaction types (including ie void?)

dtykocki:
@curiousepic - Their latest documentation, attached to the JIRA issue, says that the field is mandatory for all types.

dtykocki:
I gave the documentation a second glance and the `StoredCredentialMode` option may not apply for all transaction types:

> This parameter is only applicable to merchants that store tokenized card data and is mandatory when sg_cardnumber is sent. Merchants that do not store card data, or that are using SafeCharge’s
tokenization feature (sg_CCToken parameter), should not send this parameter.

The adapter only sends `sg_cardnumber` for `purchase`, `authorize`, and `credit`. The remaining reference based transaction type use `sg_CCToken`.

dtykocki:
I went ahead and renamed the option to `stored_credential_mode`.  I felt that `card_not_present` did not clear describe the purpose of the option. For now, I'd rather keep this option very gateway specific.

PR #2822, opened 2018-04-26
FirstPay: Handle missing billing addresses

dtykocki:
When a billing address or address is not specified, the adapter
will raise an exception. This adds a guard around the address
to prevent these exceptions from occurring.

Remote:
12 tests, 27 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
10 tests, 108 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


dtykocki:
@shasum - I'm unable to add you to the list of reviewers. Tagging you here instead.

PR #2831, opened 2018-05-07
Bug on PLN iso code

chopenhauer:
Bug: wrong iso code for PLN


nfarve:
@chopenhauer is there somewhere we can verify the codes?

curiousepic:
Code is good according to wiki, it was mixed up with PLZ. Merging.

chopenhauer:
@nfarve @curiousepic 
You can check it here
https://en.wikipedia.org/wiki/ISO_4217
You will see that PLZ 616 was deprecated against PLN 985

PR #2834, opened 2018-05-07
Adding 3D Secure pass thru capabilities to Braintree

filipebarcos:
This is simply adding support to `:three_d_secure_pass_thru` params for
Braintree, documented here: https://developers.braintreepayments.com/reference/request/transaction/sale/ruby#three_d_secure_pass_thru


filipebarcos:
@joshnuss I'm down for that, but I could do it in a separated pr, if that's fine for you. We have more PRs coming with similar purpose, so we can just group them all

PR #2849, opened 2018-05-16
QBMS: Support transcript scrubbing

curiousepic:
We do not have test credentials for QBMS and they claim their test
sandbox is not currently working, so the remote tests cannot be run. The
mock transcripts added here in the unit tests should be accurate; they
come from an external scrub test that dates to the same time and API
version for the gateway implementation here in Active Merchant. We also
can confirm properly scrubbed recent transcripts using the same scrub
method externally.

Unit:
15 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/qbms.rb
+++ b/lib/active_merchant/billing/gateways/qbms.rb
@@ -113,6 +113,17 @@ module ActiveMerchant #:nodoc:
         commit(:query, nil, {})
       end
 
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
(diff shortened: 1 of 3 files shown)


dtykocki:
@curiousepic - I presume the gateway can't provide an ETA when their sandbox will be available?

Also I don't follow this statement:
> We also can confirm properly scrubbed recent transcripts using the same scrub
method externally.

curiousepic:
They specifically said they could not.

This scrub method is the same as what Spreedly calls manually (externally) on QBMS transactions made today, and I've confirmed that those transcripts are properly scrubbed.

PR #2850, opened 2018-05-18
Paymentez: Read messages on Failure with no error

nfarve:
Extends parsing messages on failure. Messages can be returned that are not included in the error
field.

Loaded suite test/unit/gateways/paymentez_test
................

100% passed

Loaded suite test/remote/gateways/remote_paymentez_test
................

16 tests, 39 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -219,7 +219,11 @@ module ActiveMerchant #:nodoc:
         if success_from(response)
           response['transaction'] && response['transaction']['message']
         else
-          response['error'] && response['error']['type']
+          if response['error']
+            response['error'] && response['error']['type']
+          else
+            response['transaction']['message']
+          end
         end
       end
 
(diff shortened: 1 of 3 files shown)


nfarve:
I was unable to reproduce the error with a card with an old year, so I think expired here means something different. Thats why only the unit test.

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -219,7 +219,11 @@ def message_from(response)
         if success_from(response)
           response['transaction'] && response['transaction']['message']
         else
-          response['error'] && response['error']['type']
+          if response['error']
+            response['error'] && response['error']['type']
+          else
+            response['transaction']['message']
```
curiousepic: It might be worth considering concatenating both the error type and transaction message in this case, which we do for some other gateways.

nfarve: Might be misunderstanding, but it looks like its one or the other. We get back an error or a transaction.

PR #2854, opened 2018-05-22
Authorize.net: Add ability to pass `customer_payment_profile_id`

nfarve:
Adds the ability to actively set the `customer_payment_profile_id`
field.

..................................................................

66 tests, 224 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/unit/gateways/authorize_net_test
............................................................................................

92 tests, 521 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -167,7 +167,7 @@ module ActiveMerchant
             xml.transactionType('refundTransaction')
             xml.amount(amount(amount))
 
-            add_payment_source(xml, payment, :credit)
+            add_payment_source(xml, payment, :credit, options)
             xml.refTransId(transaction_id_from(options[:transaction_id])) if options[:transaction_id]
             add_invoice(xml, 'refundTransaction', options)
             add_customer_data(xml, payment, options)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -375,10 +375,10 @@ def normal_void(authorization, options)
         end
       end
 
-      def add_payment_source(xml, source, action = nil)
+      def add_payment_source(xml, source, action = nil, options)
```
bpollack: The behavior of a mandatory parameter after an optional parameter can get really weird; I think it's worth reversing these (i.e. `add_payment_source(xml, source, options, action = nil)`).

That will also allow you to ax the `options=options` calls at 271 and 249 to boot.

test/remote/gateways/remote_authorize_net_test.rb
```diff
@@ -392,6 +392,15 @@ def test_failed_authorize_using_stored_card
...
 
+  def test_failed_authorize_using_wrong_token
+    response = @gateway.store(@declined_card)
+    assert_success response
+
+    response = @gateway.authorize(@amount, response.authorization, @options.merge(customer_payment_profile_id: 12345))
+    assert_failure response
+
+    assert_equal 'Customer Profile ID or Customer Payment Profile ID not found', response.message
+  end
```
bpollack: Newline after `end`

nfarve:
@bpollack thanks updated! I always struggle with balancing changing as little as possible and making necessary changes.

PR #2855, opened 2018-05-23
Mundipagg: Remove Billing Address if no Address Sent

nfarve:
Removes the `billing_address` object if no address is sent.

Loaded suite test/unit/gateways/mundipagg_test
...............

15 tests, 65 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

....................

20 tests, 50 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/mundipagg.rb
+++ b/lib/active_merchant/billing/gateways/mundipagg.rb
@@ -109,18 +109,20 @@ module ActiveMerchant #:nodoc:
         post[:customer][:email] = options[:email]
       end
 
-      def add_billing_address(options)
-        billing = {}
-        address = options[:billing_address] || options[:address]
-        billing[:street] = address[:address1].match(/\D+/)[0].strip if address[:address1]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -174,7 +174,9 @@ def add_credit_card(post, payment, options)
           post[:payment][:credit_card][:card][:exp_month] = payment.month
           post[:payment][:credit_card][:card][:exp_year] = payment.year
           post[:payment][:credit_card][:card][:cvv] = payment.verification_value
-          post[:payment][:credit_card][:card][:billing_address] = add_billing_address(options)
+          if options[:billing_address] || options[:address]
```
curiousepic: A bit nitpicky, but it would be a little more compact to just call the `add_billing_address` method here, and put the conditional and [:billing_address] field creation into that method.

nfarve:
@curiousepic updated!

PR #2859, opened 2018-05-25
Stripe: Handle upstream processing errors

bpollack:
Sometimes, Stripe can fail internally, in which case its status will be set to "failed" even though an error key may be absent. This commit handles that case and adds a regression test for it.

Unit: 127 tests, 674 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 64 tests, 288 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


bpollack:
Just updated after a clean run on remote tests (Stripe didn't time out, whee).

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -549,14 +549,15 @@ def commit(method, url, parameters = nil, options = {})
         add_expand_parameters(parameters, options) if parameters
         response = api_request(method, url, parameters, options)
 
-        success = !response.key?("error")
+        success = !response.key?("error") && response["status"] != "failed"
+        error_message = response.fetch("error", {"message" => "No details provided"})["message"]
```
curiousepic: "No error details" might be a little more explicit than "No details provided".

bpollack: Good call; will change.

bpollack:
All right, give that a shot @curiousepic ; it makes the patch larger, but I think I agree it makes it easier to read.

curiousepic:
@bpollack I'll take readable over compact any day. Honestly, this was more about `.fetch` being new/unintuitive for me and giving the logic room to breath past one line, but not a big deal, I do like it and might start using it more!

bpollack:
@curiousepic `Hash#dig` negates 99% of the need I've got for `Hash#fetch`, so I wouldn't get too attached for it if we can indeed drop Ruby 2.1 and 2.2 support for AM shortly. :)

PR #2861, opened 2018-05-29
Remove options parameter from Cybersource add_check_service call

chriscz:
The options parameter is not required by the add_check_service function.


bpollack:
This fix looks good @chriscz . Would you mind adding a regression unit test?

bpollack:
@curiousepic This is missing tests, but also looks patently correct; any concerns with landing it?

curiousepic:
Sure 👍

PR #2863, opened 2018-05-30
Require Ruby 2.3 or higher

bpollack:
Rubies older than 2.3.6 are outside maintenance, so this honestly seems
like the bare minimum for a payment gem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ rvm:
 - 2.5
 - 2.4
 - 2.3
-- 2.2
 
 gemfile:
 - Gemfile.rails52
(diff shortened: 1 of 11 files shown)


bdewater:
I feel we should make our stance on dropping unsupported Ruby/Rails explicit in the readme here as well: https://github.com/activemerchant/active_merchant#api-stability-policy

bpollack:
Good call @bdewater . I took a stab at that just now; thoughts?

bpollack:
BTW, Travis is having a bad day, but I promise these tests do actually pass (which is pretty obviously if you click through).

bpollack:
Would someone from Shopify mind approving or rejecting this so I can merge? Spreedly's good with it landing.

bdewater:
@bpollack can you include Rails (or Active Support) in that compatibility policy?

bpollack:
@bdewater Done. I can flag Active Support instead if you wish, but Rails seemed like it _might_ be clearer for people. No real preference here though.

PR #2865, opened 2018-05-31
Openpay: support payment installments

bpollack:
Remote tests currently fail, I *think* because I'm misunderstanding how
to use the API, but I'm not yet sure.

Unit: 20 tests, 102 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: N/A

### DIFF
--- a/lib/active_merchant/billing/gateways/openpay.rb
+++ b/lib/active_merchant/billing/gateways/openpay.rb
@@ -38,6 +38,7 @@ module ActiveMerchant #:nodoc:
       def capture(money, authorization, options = {})
         post = {}
         post[:amount] = amount(money) if money
+        post[:payments] = options[:payments] if options[:payments]
         commit(:post, "charges/#{CGI.escape(authorization)}/capture", post, options)
       end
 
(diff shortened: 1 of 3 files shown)


dtykocki:
:angry_homer: Come on Travis! 

Locally, on master, I'm seeing four remote test failures that I _think_ could be fixed up. The `test_successful_purchase_with_card_stored` fails with message "Please specify a valid source_id". Maybe we can provide a `source_id` to make this one pass?

The remaining three failures are:
```
"The card is not supported on online transactions" expected but was "The card was declined".
```

It's your call here, but just the message for these failures could changed.

bpollack:
You'll be happy to know I fixed Travis in https://github.com/activemerchant/active_merchant/commit/3efa5c901aa327445056eff0496a398bf93237ba and https://github.com/activemerchant/active_merchant/commit/f75f57e55c4cab9d2673a564243a15f810bdcb9e , but I hadn't rebased yet; just did so and re-pushed, so Travis should go well.

And that's fair on the remotes. I generally avoid touching those if they've been working in prod since they occasionally require adapter tweaks, but you're right that these are pretty straightforward test-local changes. I'll take a look tomorrow.

bpollack:
Okay, @dtykocki I fixed everything but the `source_id` error, which seems like it _might_ be 3DS-related, but at any rate isn't documented in the Openpay documentation. This gets us down to one remote failure.

PR #2868, opened 2018-05-31
Prevent keepalives by default

bpollack:
With b20ad8a287567, the connection logic changed such that keep-alives would suddenly be used, where previously, Connection: close would be sent. This patch reverts to the old behavior. Keep-alives can still be manually requested by passing the relevant header.

I'm not 100% sure this is the right fix, and I'd love a regression test, but this resolves some issues we were seeing over at Spreedly with integrating this version of AM.

### DIFF
--- a/lib/active_merchant/connection.rb
+++ b/lib/active_merchant/connection.rb
@@ -65,6 +65,8 @@ module ActiveMerchant
     end
 
     def request(method, body, headers = {})
+      headers['connection'] ||= 'close'
+
       request_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
 
       retry_exceptions(:max_retries => max_retries, :logger => logger, :tag => tag) do
--- a/test/unit/connection_test.rb
+++ b/test/unit/connection_test.rb
@@ -29,7 +29,7 @@ class ConnectionTest < Test::Unit::TestCase
   def test_connection_passes_env_proxy_by_default
     spy = Net::HTTP.new('example.com', 443)
     Net::HTTP.expects(:new).with('example.com', 443, :ENV, nil).returns(spy)
-    spy.expects(:get).with('/tx.php', {}).returns(@ok)
+    spy.expects(:get).with('/tx.php', {'connection' => 'close'}).returns(@ok)
     @connection.request(:get, nil, {})
   end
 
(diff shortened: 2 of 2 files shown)


bdewater:
For the curious, `req['connection'] ||= 'close'` was previously set automatically here:
https://github.com/ruby/ruby/blob/fef8339f582b125e1fec5f3d47adbb48f7cede92/lib/net/http.rb#L1453-L1455
but now that we start it ourselves, we have to set this ourselves, too. Thanks @bpollack !

grantbdev:
@dtykocki @bdewater This change started failing some connections for me in a not-so-obvious way. If you pass a frozen headers hash, `FrozenError: can't modify frozen Hash` will be raised internally and the method will jump to the `ensure` block. Then `TypeError: nil can't be coerced into Float` will be raised because `request_start` is used for subtraction, but it has not been defined yet.

Assigning `request_start` before `headers['connection'] ||= 'close'` could help people see the actual error.

I am making sure to pass a non-frozen headers hash for now, but if frozen headers are supposed to be a supported use case maybe it should be converted to not do a mutation.

What do you think?

bdewater:
Hi @grantbdev sorry this is causing you trouble, the change had a much larger impact than I anticipated 😞 

> Assigning request_start before headers

Makes sense 👍 should've caught that in the review, anything we do in there should be instrumented, even if it's simply assignment.

> I am making sure to pass a non-frozen headers hash for now, but if frozen headers are supposed to be a supported use case maybe it should be converted to not do a mutation.

I believe this usually gets `dup`'ed in most libraries to prevent mutating passed in arguments.

Would love to see a PR to fix these 😄!

grantbdev:
@bdewater I'd be happy to submit a PR for this.

> I believe this usually gets dup'ed in most libraries to prevent mutating passed in arguments.

Do you have a preference where this happens? [`#raw_ssl_request`](https://github.com/activemerchant/active_merchant/blob/df6690788bc55746f07627a427590cf472b6186a/lib/active_merchant/posts_data.rb#L48) and [`#request`](https://github.com/activemerchant/active_merchant/blob/df6690788bc55746f07627a427590cf472b6186a/lib/active_merchant/connection.rb#L67) seem like the primary candidates to me, but I'm not sure how high/low in the stack this should be done.

Thanks! 😀

PR #2873, opened 2018-06-04
Bridge Pay: don't throw when account_type omitted

bpollack:
One remote test fails (for echecks, no less), but it does so both
with and without this patch, and the error appears internal (it fails
with a .NET XSD validation error for an element we're not sending and
isn't documented).

Unit: 14 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 17 tests, 38 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.1176% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/bridge_pay.rb
+++ b/lib/active_merchant/billing/gateways/bridge_pay.rb
@@ -114,7 +114,7 @@ module ActiveMerchant #:nodoc:
           post[:TransitNum] = payment_method.routing_number
           post[:AccountNum] = payment_method.account_number
           post[:NameOnCheck] = payment_method.name
-          post[:ExtData] = "<AccountType>#{payment_method.account_type.capitalize}</AccountType>"
+          post[:ExtData] = "<AccountType>#{payment_method.account_type[0].capitalize}</AccountType>" if payment_method.account_type
         end
       end
 
--- a/test/unit/gateways/bridge_pay_test.rb
+++ b/test/unit/gateways/bridge_pay_test.rb
@@ -45,6 +45,14 @@ class BridgePayTest < Test::Unit::TestCase
     assert response.test?
   end
 
+  def test_failed_purchase_with_bad_echeck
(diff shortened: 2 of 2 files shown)


lib/active_merchant/billing/gateways/bridge_pay.rb
```diff
@@ -114,7 +114,7 @@ def add_payment_method(post, payment_method)
           post[:TransitNum] = payment_method.routing_number
           post[:AccountNum] = payment_method.account_number
           post[:NameOnCheck] = payment_method.name
-          post[:ExtData] = "<AccountType>#{payment_method.account_type.capitalize}</AccountType>"
+          post[:ExtData] = "<AccountType>#{payment_method.account_type[0].capitalize}</AccountType>" if payment_method.account_type
```
dtykocki: I don't believe the `account_type` property is an array. Is the `[0]` necessary here?

dtykocki: Duh.. this is Ruby - it could also work for strings.

dtykocki: How about adding a unit test for this fix?

bpollack: Gah, I did, but forgot to `git add -p` as I was bouncing between two other branches. Will grab it out of whatever stash it landed in and add.

And re. the array access: AFAICT, the original implementation was wrong. In AM, if I understand properly, `account_type` should be one of `checking` or `savings`, but Bridge Pay actually requires the value `C` or `S` (see https://www.dropbox.com/s/1tg479a9n8e3dvt/BridgeComm%20Communication%20Interface%20API%20Guide.pdf?dl=0 page 96, which I think is what should be happening here).

bpollack:
Found and cherry-picked out the unit test; should be good now.

PR #2875, opened 2018-06-04
Do not make remote connections in unit tests

bpollack:
Blue Snap and Mundipagg unit tests made remote connections. Now they
don't.


bpollack:
Hold; went too quick on this. The idea is right, implementation is wrong.

PR #2879, opened 2018-06-06
Wepay gateway can use global options in headers

mcelicalderon:
Without this you always have to specify `version` in the options hash for every request or you'll get this error from WePay
```
Your app cannot access this version of the API
```
API version can be set int the gateway's initializer like

```
ActiveMerchant::Billing::WepayGateway.new(version: '2018-03-21', .......)
```

### DIFF
--- a/lib/active_merchant/billing/gateways/wepay.rb
+++ b/lib/active_merchant/billing/gateways/wepay.rb
@@ -222,16 +222,14 @@ module ActiveMerchant #:nodoc:
       end
 
       def headers(options)
-        headers = {
-          "Content-Type"      => "application/json",
-          "User-Agent"        => "ActiveMerchantBindings/#{ActiveMerchant::VERSION}",
-          "Authorization"     => "Bearer #{@options[:access_token]}"
-        }
-        headers["Api-Version"] = options[:version] if options[:version]
-        headers["Client-IP"] = options[:ip] if options[:ip]
-        headers["WePay-Risk-Token"] = options[:risk_token] if options[:risk_token]
-
-        headers
+        {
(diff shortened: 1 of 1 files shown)


mcelicalderon:
I'm using v1.70.0. Now I understand why I'm getting the error as that version defaults to api version `2017-02-01` here https://github.com/activemerchant/active_merchant/blob/v1.70.0/lib/active_merchant/billing/gateways/wepay.rb#L13

Still, this change will allow you to configure the gateway object to optionally use the specified API version for all request you make, unless otherwise specified in the options param for each request.

mcelicalderon:
Reading this https://developer.wepay.com/api/guides/how-to-versioning now I'm not completely sure this change is that useful, as to globally change you API version you can do it from the wepay app console and all your requests will use that version unless otherwise specified. The only possible use I see for this change is to upgrade API version and test that it works from your code before actually changing the config in the app console.

bpollack:
Yeah, [we hit a similar issue](https://docs.spreedly.com/payment-gateways/wepay/) when tackling this feature, and it's why we allow the versions to float in the gateway object:

> Due to the fact that WePay’s system restricts transaction compatibility to the version set in your WePay account’s Dashboard, and new WePay merchant accounts cannot be set to older versions, Spreedly cannot assume a single version for all of our customers. Therefore, we do not explicitly specify which version of WePay’s API our transactions conform to. This means that WePay assumes that we are using the version that is set in your WePay account Dashboard.

I appreciate you taking the time to try adding this feature, but I think I'm going to reject this PR for now. If WePay changes how they handle versioning in the future, though, I'm really happy to take another look.

PR #2379, opened 2017-03-27
Stripe: support shipping info in purchases

whitby3001:
Adds support for adding a shipping address and tracking information to purchases and authorizations.

PR #2884, opened 2018-06-12
Pass shipping address to Stripe

jonmast:
Stripe accepts a shipping address as an optional parameter, use the :shipping_address from the options hash to fill this in when available.

Stripe uses this for fraud detection, see https://stripe.com/docs/api#create_charge-shipping


bpollack:
Thanks @jonmast ; I found an existing PR, #2379, that implemented the same functionality, and just merged that. If you see anything unsatisfactory in that implementation, please feel free to submit another PR improving it.

jonmast:
Oops, I somehow missed that. The implementation looks good (better than this one) at first glance, I'll open a PR if I run into any problems integrating it.

PR #2885, opened 2018-06-13
Deferred method for PayU Latam

Shelvak:
- Adding Billing::Deferred object to handle deferred coupon payment method
- Adapting PayuLatam to handle deferred payment method.
- Adding test for deferred case


Shelvak:
:man_shrugging:

PR #2886, opened 2018-06-13
Kuski: do not send extraTaxes if no value provided

bpollack:
The failing remote test is a refund, which fails with a note that
refunds for the test card aren't supported (I think; my Spanish isn't
great).

Unit: 8 tests, 45 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 9 tests, 27 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
88.8889% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/kushki.rb
+++ b/lib/active_merchant/billing/gateways/kushki.rb
@@ -94,14 +94,7 @@ module ActiveMerchant #:nodoc:
         sum[:iva] = 0
         sum[:subtotalIva0] = 0
 
-        if sum[:currency] == "COP"
-          extra_taxes = {}
-          extra_taxes[:propina] = 0
-          extra_taxes[:tasaAeroportuaria] = 0
-          extra_taxes[:agenciaDeViaje] = 0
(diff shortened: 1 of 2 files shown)


bpollack:
@curiousepic Could you take another look, if you don't mind? The first code didn't actually work as intended when I prodded; I added some extra regression tests and fixed the underlying issue.

bpollack:
@curiousepic Specifically https://github.com/activemerchant/active_merchant/pull/2886/files#diff-154e23ec35747841db7f46d7dd8f137aR108 was needed to patch things up, and you can see the regression tests diving in pretty hard there.

curiousepic:
@bpollack I was actually wondering about that, but assumed the tests were sufficient 😕New ones look great 👍

curiousepic:
@bpollack Wait, is there a reason there's no remote tests for extra_taxes?

bpollack:
@curiousepic Yeah: I have absolutely no idea at all what reasonable values for that are.

PR #2888, opened 2018-06-14
Moneris US: Add ACH

nfarve:
Adds ACH capabilities to Moneris US Gateway

One failing test for CVV verification that are unrelated to these
changes.

Loaded suite test/remote/gateways/remote_moneris_us_test
Failure: test_cvv_no_match_when_enabled(MonerisUsRemoteTest)

30 tests, 118 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.6667% passed

Loaded suite test/unit/gateways/moneris_us_test

33 tests, 164 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/moneris_us.rb
+++ b/lib/active_merchant/billing/gateways/moneris_us.rb
@@ -72,9 +72,15 @@ module ActiveMerchant #:nodoc:
         add_payment_source(post, creditcard_or_datakey, options)
         post[:amount]     = amount(money)
         post[:order_id]   = options[:order_id]
-        post[:address]    = options[:billing_address] || options[:address]
+        add_address(post, creditcard_or_datakey, options)
         post[:crypt_type] = options[:crypt_type] || @options[:crypt_type]
-        action = (post[:data_key].blank?) ? 'us_purchase' : 'us_res_purchase_cc'
+        action = if creditcard_or_datakey.is_a?(String)
+          'us_res_purchase_cc'
+        elsif card_brand(creditcard_or_datakey) == 'check'
+          'us_ach_debit'
+        elsif post[:data_key].blank?
+          'us_purchase'
+        end
         commit(action, post)
       end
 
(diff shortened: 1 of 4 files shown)


nfarve:
I went down the path of implementing all of the store features for ACH until I realized that we don't support Store for this gateway. We may want to open a separate ticket to support store, but this gateway is being decommissioned.

curiousepic:
Looks pretty good; this is a kind of convoluted adapter. Not seeing unit tests though, do you think they'd be worth it?

lib/active_merchant/billing/gateways/moneris_us.rb
```diff
@@ -154,14 +157,39 @@ def expdate(creditcard)
         sprintf("%.4i", creditcard.year)[-2..-1] + sprintf("%.2i", creditcard.month)
       end
 
+      def add_address(post, payment_method, options)
+        if !payment_method.is_a?(String) && card_brand(payment_method) == 'check'
+          post[:ach_info][:cust_first_name] = payment_method.first_name if payment_method.first_name
```
bpollack: Does it make sense to `post[:ach_info] ||= {}` here? If someone reordered `add_address` and `add_payment_source` calls in `purchase` or something, it'd suddenly crash, and the lack of unit tests means that might not be caught super quickly. (Alternatively, if there were a purchase unit test, I think I'd be okay with this arrangement.)

lib/active_merchant/billing/gateways/moneris_us.rb
```diff
@@ -154,14 +157,39 @@ def expdate(creditcard)
...
+          ach_info[:routing_num] = source.routing_number
+          ach_info[:account_num] = source.account_number
+          ach_info[:account_type] = source.account_type
+          ach_info[:check_num] = source.number if source.number
+          post[:ach_info] = ach_info
         else
           post[:pan]        = source.number
           post[:expdate]    = expdate(source)
           post[:cvd_value]  = source.verification_value if source.verification_value?
+          post[:crypt_type] = options[:crypt_type] || @options[:crypt_type] if (options[:crypt_type] || @options[:crypt_type])
```
bpollack: This is 100% up to you, but an alternative to the heavy repetition here would be

    if crypt_type = options[:crypt_type] || @options[:crypt_type]
        post[:crypt_type] = crypt_type
    end

which is _slightly_ longer, but might be easier to read. Again, entirely your call.

lib/active_merchant/billing/gateways/moneris_us.rb
nfarve: @bpollack Went ahead and added the unit test.

nfarve:
@curiousepic and @bpollack added unit tests!

bpollack:
Awesome! Thanks for sticking with this. 🚀 Ship it!

PR #2891, opened 2018-06-15
Adyen: Support Network Tokenization Cards via mpiData fields

curiousepic:
Remote:
33 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
20 tests, 94 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -44,7 +44,7 @@ module ActiveMerchant #:nodoc:
         post = init_post(options)
         add_invoice(post, money, options)
         add_payment(post, payment)
-        add_extra_data(post, options)
+        add_extra_data(post, payment, options)
         add_shopper_interaction(post, payment, options)
         add_address(post, options)
         add_installments(post, options) if options[:installments]
@@ -76,7 +76,7 @@ module ActiveMerchant #:nodoc:
         post = init_post(options)
         add_invoice(post, 0, options)
         add_payment(post, credit_card)
-        add_extra_data(post, options)
+        add_extra_data(post, credit_card, options)
         add_recurring_contract(post, options)
         add_address(post, options)
         commit('authorise', post)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -102,18 +102,24 @@ def scrub(transcript)
...
+        "google_pay" => "paywithgoogle"
+      }
+
+      def add_extra_data(post, payment, options)
         post[:shopperEmail] = options[:shopper_email] if options[:shopper_email]
         post[:shopperIP] = options[:shopper_ip] if options[:shopper_ip]
         post[:shopperReference] = options[:shopper_reference] if options[:shopper_reference]
         post[:fraudOffset] = options[:fraud_offset] if options[:fraud_offset]
-        post[:selectedBrand] = options[:selected_brand] if options[:selected_brand]
+        post[:selectedBrand] = options[:selected_brand] || NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
```
nfarve: Is it okay to pass this field empty?

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -17,14 +17,21 @@ def setup
...
+    )
+
     @options = {
       reference: '345123',
       shopper_email: "john.smith@test.com",
       shopper_ip: "77.110.174.153",
       shopper_reference: "John Smith",
       billing_address: address(),
-      order_id: "123",
-      installments: 2
```
nfarve: Did you mean to get rid of the installments?

lib/active_merchant/billing/gateways/adyen.rb
curiousepic: I don't think it's possible to be sent as empty, given the conditions... if there's no `options[:selected_brand]` and it's not a `NetworkTokenizationCreditCard`, the field doesn't get added, otherwise a value will be present.

curiousepic:
Added scrub for cavv

PR #2892, opened 2018-06-15
Prevent headers mutation

grantbdev:
#2868 added modifications to the connection request headers. This PR duplicates the headers argument before modifying the headers to prevent mutation of the original object as a side-effect. This also fixes exceptions being raised if the headers argument was a frozen hash.

This PR also moves assigning `request_start` back to the top of the method so that everything is instrumented and `ensure` won't raise an exception due to an undefined variable.

### DIFF
--- a/lib/active_merchant/connection.rb
+++ b/lib/active_merchant/connection.rb
@@ -67,6 +67,7 @@ module ActiveMerchant
     def request(method, body, headers = {})
       request_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
 
+      headers = headers.dup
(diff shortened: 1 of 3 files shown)


bpollack:
This patch makes sense to me; doing regressions against some of our internal stuff to make sure it won't cause a repeat of some of the patches that led up to this, but if it checks out, I'll rebase and merge.

bpollack:
Okay, this is clean, passes full tests, and doesn't break upstream, either. We're good to go!

bpollack:
Thanks @grantbdev !

PR #2893, opened 2018-06-18
Worldpay: support IAV verify

bpollack:
I'm researching what's required to get remote tests working. It's not that they're failing as such; it's that they don't even complete. (Like, any them.)

Unit: 39 tests, 218 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -81,9 +81,13 @@ module ActiveMerchant #:nodoc:
       end
 
       def verify(credit_card, options={})
-        MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
-          r.process(:ignore_result) { void(r.authorization, options.merge(:authorization_validated => true)) }
+        if options[:iav]&.upcase == "TRUE"
+          verify_request(credit_card, options)
+        else
+          MultiResponse.run(:use_first_response) do |r|
+            r.process { authorize(100, credit_card, options) }
+            r.process(:ignore_result) { void(r.authorization, options.merge(:authorization_validated => true)) }
+          end
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -410,8 +428,11 @@ def authorization_from(raw)
       end
 
       def credit_fund_transfer_attribute(options)
-        return unless options[:credit]
-        {'action' => "REFUND"}
+        if options[:credit]
+          {'action' => 'REFUND'}
+        elsif options[:action]
+          {'action' => options[:action]}
+        end
```
deedeelavinder: The `elsif` in this context caught my eye. Did you mean `else`? This is a great spot for a ternary operator. 

`options[:credit] ? { 'action' => 'REFUND' } : { 'action' => options[:action] }`

Also, what will happen if there is no `options[:action]`?

bpollack: I did mean `elsif`, not `else`. In the case neither branch is taken, `nil` is returned, and XML attributes will be added to the `paymentDetails` tag, which is indeed what we want as the default. This is also why I didn't go with a ternary; you'd have to use two of them to achieve the effect, and doing something like

    return if option[:action] && options[:action] != :credit
    {'action' => options[:action] == :credit ? 'REFUND' : options[:action]}

felt more confusing to me than the version I did. I'm open to alternatives, though; this conditional definitely feels odd, even if it's mechanically correct, and I did feel as if there was probably a cleaner approach.

bpollack: @deedeelavinder Actually, after sleeping on it, I think the second is actually better, because it's more obvious that the code isn't making a mistake and does intend to return `nil` in that use case. I'll swap that out. Thanks!

bpollack: Landed. Slightly different than the above code-snippet, which is wrong, but same idea.

deedeelavinder: @bpollack You could also set a variable for use in the ternary (so that `nil` is returned when appropriate), but the latest also looks good.

deedeelavinder:
👍

bpollack:
Gonna close this PR for the moment; this isn't okay to land as-is, and the branch isn't going anywhere.

PR #2895, opened 2018-06-21
Realex: Pass amount for captures

curiousepic:
Remote (7 failures unrelated to changes verified on master):
21 tests, 103 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
66.6667% passed

Unit:
21 tests, 445 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -64,7 +64,7 @@ module ActiveMerchant
       end
 
       def capture(money, authorization, options = {})
-        request = build_capture_request(authorization, options)
+        request = build_capture_request(money, authorization, options)
         commit(request)
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -147,14 +147,15 @@ def build_purchase_or_authorization_request(action, money, credit_card, options)
...
+      def build_capture_request(money, authorization, options)
         timestamp = new_timestamp
         xml = Builder::XmlMarkup.new :indent => 2
         xml.tag! 'request', 'timestamp' => timestamp, 'type' => 'settle' do
           add_merchant_details(xml, options)
+          add_amount(xml, money, options)
           add_transaction_identifiers(xml, authorization, options)
           add_comments(xml, options)
-          add_signed_digest(xml, timestamp, @options[:login], sanitize_order_id(options[:order_id]), nil, nil, nil)
+          add_signed_digest(xml, timestamp, @options[:login], sanitize_order_id(options[:order_id]), amount(money), (options[:currency] || currency(money)), nil)
```
bpollack: `options[:currency] || currency(money)` is a pattern that shows up several times here (e.g. in `build_refund_request` and `build_purchase_or_authorization_request`). Do you think it might be worth extracting? (And that's a genuine question, because three appearances is enough to make me wonder, but not enough to be an obvious win.)

curiousepic: Honestly I think the `currency` method should accept `options` and build in that conditional. It doesn't otherwise seem to be extracted for any other gateways, so I'm somewhat inclined to leave it as is.

PR #2896, opened 2018-06-21
CardConnect: add store and unstore

bpollack:
Unit: 21 tests, 89 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 23 tests, 54 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/card_connect.rb
+++ b/lib/active_merchant/billing/gateways/card_connect.rb
@@ -130,6 +130,21 @@ module ActiveMerchant #:nodoc:
         authorize(0, credit_card, options)
       end
 
+      def store(payment, options = {})
+        post = {}
+        add_payment(post, payment)
+        add_address(post, options)
+        add_customer_data(post, options)
+        commit('profile', post)
+      end
+
+      def unstore(authorization, options = {})
+        account_id, profile_id = authorization.split('|')
+        commit('profile', {},
+               verb: :delete,
+               path: "/#{profile_id}/#{account_id}/#{@options[:merchant_id]}")
+      end
+
       def supports_scrubbing?
         true
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -2,7 +2,10 @@ module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class CardConnectGateway < Gateway
       self.test_url = 'https://fts.cardconnect.com:6443/cardconnect/rest/'
+      # self.test_cardsecure_url = 'https://fts.cardconnect.com:6443/cardsecure/cs'
```
curiousepic: Do we want to keep this, or the url construction method that seems unused? Or is cardsecure integration still coming?

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -271,7 +295,11 @@ def message_from(response)
       end
 
       def authorization_from(response)
-        response['retref']
+        if response['profileid']
+          "#{response['acctid']}||#{response['profileid']}"
```
curiousepic: Nitpick but this kind of makes it look like a concated split auth, when it isn't. Maybe pull the conditional outside of the interpolation string?

curiousepic: Oh, hah, it is concating. In that case, the opposite of what I said; most separators are single character.

lib/active_merchant/billing/gateways/card_connect.rb
bpollack: Dammit, will axe. Thanks for the catch.

lib/active_merchant/billing/gateways/card_connect.rb
bpollack: Just to be clear, just do `|` instead of `||`?

bpollack:
Killed commented-out lines and switched separator from `||` to `|`. Unless I hear otherwise, I'll merge this when CI finishes.

lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -175,13 +190,19 @@ def add_invoice(post, options)
       end
 
       def add_payment(post, payment)
-        post[:name] = payment.name
-        if card_brand(payment) == 'check'
-          add_echeck(post, payment)
+        if payment.is_a?(String)
+          account_id, profile_id = payment.split('||')
```
curiousepic: This is still splitting on `||`

bpollack: Well that'll teach me not to run remote tests. Fixed (along with the corresponding test).

PR #2899, opened 2018-06-22
Update Psigate endpoint url

Girardvjonathan:
It seem that the Psigate endpoint might have changed causing: `The SSL connection to the remote server could not be established` errors for merchants. These new urls should fix the problems according to their email.


bpollack:
What's 🎩  mean?

bdewater:
Shopify jargon for manual testing. We've rolled this out in production and confirmed it works.

PR #2900, opened 2018-06-22
Paymentez: support captures with different amounts

bpollack:
Previously, Paymentez only supported capturing the original amount. It
now supports capturing any amount the processor is okay with--by
default, 100% or more of the amount specified.

As part of this commit, I've repaired all remote tests, plus altered
the partial cpature test, since it's now explicitly documented that the
test, as written, should fail (which it does).

Unit: 16 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 16 tests, 41 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_paymentez_test.rb
+++ b/test/remote/gateways/remote_paymentez_test.rb
@@ -81,23 +81,31 @@ class RemotePaymentezTest < Test::Unit::TestCase
     assert_equal 'Response by mock', capture.message
   end
 
+  def test_successful_authorize_and_capture_with_different_amount
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+    assert capture = @gateway.capture(@amount + 100, auth.authorization)
+    assert_success capture
+    assert_equal 'Response by mock', capture.message
+  end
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_paymentez_test.rb
```diff
@@ -81,23 +81,31 @@ def test_successful_authorize_and_capture_with_token
...
-    assert_equal nil, response.message
+    assert_equal 'Response by mock', response.message
   end
 
   def test_partial_capture
     auth = @gateway.authorize(@amount, @credit_card, @options)
     assert_success auth
     assert capture = @gateway.capture(@amount - 1, auth.authorization)
-    assert_success capture
+    assert_failure capture # Paymentez explicitly does not support partial capture; only GREATER than auth capture
```
curiousepic: Weird, did this fail before the change?

curiousepic: Ah, right.

PR #2906, opened 2018-06-26
Fix Style/StringLiteral violations

bpollack:
Yes, I cheated and used rubocop to help with this, but I also did
actually audit the result, and it's pretty clean to read through with a
good diff tool since no spacing, etc. was changed. My recommendation
for review would be to look for and notice the absence of `#{}` in any
of the changed strings, meaning that the `"` to `'` conversion is safe.

Unit: 3874 tests, 67956 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed


curiousepic:
FWIW, I've also confirmed that any `#{}`s in the diff still occur only in `"`s.

bdewater:
Devil's advocate: should we configure Rubocop with `double_quotes` style? Rails uses this for example. My gut feeling says it's much more prevalent and will need less fixing.

PR #2908, opened 2018-06-28
Conekta: Allow customer application in headers

curiousepic:
Remote:
15 tests, 57 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
13 tests, 84 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/conekta.rb
+++ b/lib/active_merchant/billing/gateways/conekta.rb
@@ -23,7 +23,7 @@ module ActiveMerchant #:nodoc:
         add_payment_source(post, payment_source, options)
         add_details_data(post, options)
 
-        commit(:post, 'charges', post)
+        commit(:post, 'charges', post, options)
       end
 
       def authorize(money, payment_source, options = {})
@@ -34,7 +34,7 @@ module ActiveMerchant #:nodoc:
         add_details_data(post, options)
 
         post[:capture] = false
-        commit(:post, 'charges', post)
+        commit(:post, 'charges', post, options)
       end
 
       def capture(money, identifier, options = {})
@@ -43,7 +43,7 @@ module ActiveMerchant #:nodoc:
         post[:order_id] = identifier
         add_order(post, money, options)
 
-        commit(:post, "charges/#{identifier}/capture", post)
+        commit(:post, "charges/#{identifier}/capture", post, options)
       end
 
       def refund(money, identifier, options)
@@ -52,12 +52,12 @@ module ActiveMerchant #:nodoc:
         post[:order_id] = identifier
         add_order(post, money, options)
 
-        commit(:post, "charges/#{identifier}/refund", post)
+        commit(:post, "charges/#{identifier}/refund", post, options)
       end
 
       def void(identifier, options = {})
         post = {}
-        commit(:post, "charges/#{identifier}/void", post)
+        commit(:post, "charges/#{identifier}/void", post, options)
       end
 
       def supports_scrubbing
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/conekta.rb
```diff
@@ -173,22 +173,27 @@ def parse(body)
         JSON.parse(body)
       end
 
-      def headers(meta)
+      def headers(options)
         {
-          "Accept" => "application/vnd.conekta-v#{options[:version]}+json",
+          "Accept" => "application/vnd.conekta-v#{@options[:version]}+json",
```
bpollack: This has to become `@options` to distinguish itself from the local `options` variable that is shadowing the `options` method, right? If so, it's correct; just confused by the diff.

lib/active_merchant/billing/gateways/conekta.rb
```diff
@@ -173,22 +173,27 @@ def parse(body)
...
-          "X-Conekta-Client-User-Agent" => user_agent,
-          "X-Conekta-Client-User-Metadata" => meta.to_json
+          "X-Conekta-Client-User-Agent" => conekta_client_user_agent(options),
+          "X-Conekta-Client-User-Metadata" => @options.to_json
         }
       end
 
+      def conekta_client_user_agent(options)
+        return user_agent unless options[:application]
+        JSON.dump(JSON.parse(user_agent).merge!({application: options[:application]}))
```
bpollack: Welp that's the first time I've seen JSON in a header. Probably won't be the last.

lib/active_merchant/billing/gateways/conekta.rb
```diff
@@ -173,22 +173,27 @@ def parse(body)
...
+          "Accept" => "application/vnd.conekta-v#{@options[:version]}+json",
           "Accept-Language" => "es",
-          "Authorization" => "Basic " + Base64.encode64("#{options[:key]}:"),
+          "Authorization" => "Basic " + Base64.encode64("#{@options[:key]}:"),
           "RaiseHtmlError" => "false",
           "Conekta-Client-User-Agent" => {"agent"=>"Conekta ActiveMerchantBindings/#{ActiveMerchant::VERSION}"}.to_json,
-          "X-Conekta-Client-User-Agent" => user_agent,
-          "X-Conekta-Client-User-Metadata" => meta.to_json
+          "X-Conekta-Client-User-Agent" => conekta_client_user_agent(options),
+          "X-Conekta-Client-User-Metadata" => @options.to_json
```
bpollack: Shouldn't this be either `options[:meta].to_json` or `@options[:meta].to_json`? If not, then I really _am_ getting confused what's going on here.

curiousepic: I was also thrown for a loop by this and this was the result of trial and error, so your confusion is a good signal that I should take a step back and really learn what's going on. `@options[:meta]` doesn't find anything when used here. Previously, the `headers` method pulled it from `options[:meta]`. I'll root around a bit more.

curiousepic: I think I know what's going on here, though it's kind of a doozy to explain. Here goes. When the gateway was first added, `commit` called `headers(options[:meta])`, and `headers` had `meta.to_json` for the appropriate header, anticipating a `meta` option field being passed to it. However, none of the calls to `commit` passed `options`. So that has sat there doing nothing since then. But `headers` also called `options[:key]` and `options[:version]`, which, in the absence of a local `options`, apparently resolves to the global values? I'm still not sure why that happens, but I've confirmed in `pry`. 

In any case, my confusion led me to think the Metadata header actually wanted the global options hash, which.... no. So I'll restore it to look for `options[:meta]`.

curiousepic: @bpollack Done, and added test for `options[:meta]`

lib/active_merchant/billing/gateways/conekta.rb
curiousepic: Yep, since I need the local options also, I had to make this distinction.

curiousepic:
Also corrected quote style.

lib/active_merchant/billing/gateways/conekta.rb
bpollack: > But headers also called options[:key] and options[:version], which, in the absence of a local options, apparently resolves to the global values?

Yep, because [`@options` is declared `attr_reader` in `gateway.rb`](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L146), so if there's no local variable called `options`, then you'll grab the gateway's global config from the accessor.

In related news we may want to stop naming variables `options` in gateways…

lib/active_merchant/billing/gateways/conekta.rb
```diff
@@ -173,23 +173,28 @@ def parse(body)
...
+      def headers(options)
         {
-          'Accept' => "application/vnd.conekta-v#{options[:version]}+json",
-          'Accept-Language' => 'es',
-          'Authorization' => 'Basic ' + Base64.encode64("#{options[:key]}:"),
-          'RaiseHtmlError' => 'false',
-          'Conekta-Client-User-Agent' => {'agent'=>"Conekta ActiveMerchantBindings/#{ActiveMerchant::VERSION}"}.to_json,
-          'X-Conekta-Client-User-Agent' => user_agent,
-          'X-Conekta-Client-User-Metadata' => meta.to_json
+          "Accept" => "application/vnd.conekta-v#{@options[:version]}+json",
```
bpollack: I thought `RuboCop` wanted exactly the opposite of these (Travis seems to agree).

curiousepic:
Whoops, didn't actually look at Travis, passes now, merging. Thanks for the 👀 @bpollack

PR #2775, opened 2018-03-05
Allow setting min/max SSL version for a connection on Ruby 2.5

bdewater:
The configuration on the Net::HTTP object is passed through to OpenSSL::SSLContext. The functionality in the OpenSSL gem was introduced in https://github.com/ruby/openssl/pull/142

PR #2909, opened 2018-06-29
Default SSL min_version to TLS 1.1 to comply with PCI DSS deadline

bdewater:
Ruby supports setting this from version 2.5, I backported this for 2.3 and 2.4 with the [min_max_ssl](https://github.com/bdewater/min_max_ssl) gem. Also see https://github.com/activemerchant/active_merchant/pull/2775

For more details about the deadline: https://blog.pcisecuritystandards.org/are-you-ready-for-30-june-2018-sayin-goodbye-to-ssl-early-tls

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* Default SSL min_version to TLS 1.1 to comply with June 30 PCI DSS deadline [bdewater] #2909
 * Paymentez: return a Result object even when the upstream server 500s [bpollack] #2871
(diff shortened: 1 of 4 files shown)


bpollack:
@bdewater The test failures from Travis look legit.

Conceptually, I'm fine with this; I don't think this impacts Spreedly, since we override the connection objects and just went through our own version of this internally. I'll do a double-check if it lands that it doesn't break us, but it shouldn't be a big deal to un-break if it somehow does.

bdewater:
I had forgotten I had written a test for this 😅 should be fixed now.

PR #2910, opened 2018-06-29
CT Payment: Add new gateway

nfarve:
Adds a new gateway (ct_payment) along with a certicication test file

Loaded suite test/unit/gateways/ct_payment_test

14 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/remote/gateways/remote_ct_payment_test

100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ct_payment.rb
+++ b/lib/active_merchant/billing/gateways/ct_payment.rb
@@ -1,0 +1,267 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CtPaymentGateway < Gateway
+      self.test_url = 'https://test.ctpaiement.ca/v1/'
+      self.live_url = 'hhtps://www.ctpaiement.com/v1/'
+
+      self.supported_countries = ['US', 'CA']
+      self.default_currency = 'CAD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club]
+
+      self.homepage_url = 'http://www.ct-payment.com/'
+      self.display_name = 'CT Payment'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        '14' => STANDARD_ERROR_CODE[:invalid_number],
+        '05' => STANDARD_ERROR_CODE[:card_declined],
+        'M6' => STANDARD_ERROR_CODE[:card_declined],
+        '9068' => STANDARD_ERROR_CODE[:incorrect_number],
+        '9067' => STANDARD_ERROR_CODE[:incorrect_number]
+      }
+      CARD_BRAND = {
+        'american_express' => 'A',
+        'master' => 'M',
+        'diners_club' => 'I',
+        'visa' => 'V',
+        'discover' => 'O'
+      }
+
+      def initialize(options={})
+        requires!(options, :api_key, :company_number, :merchant_number)
+        super
+      end
+
+      def purchase(money, payment, options={})
+        requires!(options, :order_id)
+        post = {}
+        add_terminal_number(post, options)
+        add_money(post, money)
+        add_operator_id(post, options)
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+
+        payment.is_a?(String) ? commit('purchaseWithToken', post) : commit('purchase', post)
+      end
+
+      def authorize(money, payment, options={})
+        requires!(options, :order_id)
+        post = {}
+        add_money(post, money)
+        add_terminal_number(post, options)
+        add_operator_id(post, options)
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+
+        payment.is_a?(String) ? commit('preAuthorizationWithToken', post) : commit('preAuthorization', post)
+      end
+
+      def capture(money, authorization, options={})
+        requires!(options, :order_id)
+        post = {}
+        add_invoice(post, money, options)
+        add_money(post, money)
+        add_customer_data(post, options)
+        transaction_number, authorization_number, invoice_number = split_authorization(authorization)
+        post[:OriginalTransactionNumber] = transaction_number
+        post[:OriginalAuthorizationNumber] = authorization_number
+        post[:OriginalInvoiceNumber] = invoice_number
+
+        commit('completion', post)
+      end
+
+      def refund(money, authorization, options={})
+        requires!(options, :order_id)
+        post = {}
+        add_invoice(post, money, options)
+        add_money(post, money)
+        add_customer_data(post, options)
+        transaction_number, _, invoice_number = split_authorization(authorization)
+        post[:OriginalTransactionNumber] = transaction_number
+        post[:OriginalInvoiceNumber] = invoice_number
+
+        commit('refundWithoutCard', post)
+      end
+
+      def credit(money, payment, options={})
+        requires!(options, :order_id)
+        post = {}
+        add_terminal_number(post, options)
+        add_money(post, money)
+        add_operator_id(post, options)
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+
+        payment.is_a?(String) ? commit('refundWithToken', post) : commit('refund', post)
+      end
+
+      def void(authorization, options={})
+        post = {}
+        post[:InputType] = 'I'
+        post[:LanguageCode] = 'E'
+        transaction_number, _, invoice_number = split_authorization(authorization)
+        post[:OriginalTransactionNumber] = transaction_number
+        post[:OriginalInvoiceNumber] = invoice_number
+        add_operator_id(post, options)
+        add_customer_data(post, options)
+
+        commit('void', post)
+      end
+
+      def verify(credit_card, options={})
+        requires!(options, :order_id)
+        post = {}
+        add_terminal_number(post, options)
+        add_operator_id(post, options)
+        add_invoice(post,0, options)
+        add_payment(post, credit_card)
+        add_customer_data(post, options)
+
+        commit('verifyAccount', post)
+      end
+
+      def store(credit_card, options={})
+        requires!(options, :email)
+        post = {
+          LanguageCode: 'E',
+          Name: credit_card.name.rjust(50, ' '),
+          Email: options[:email].rjust(240, ' ')
+        }
+        add_operator_id(post, options)
+        add_payment(post, credit_card)
+        add_customer_data(post, options)
+
+        commit('recur/AddUser', post)
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((&?auth-api-key=)[^&]*)i, '\1[FILTERED]').
+          gsub(%r((&?payload=)[a-zA-Z%0-9=]+)i, '\1[FILTERED]').
+          gsub(%r((&?token:)[^&]*)i, '\1[FILTERED]').
+          gsub(%r((&?cardNumber:)[^&]*)i, '\1[FILTERED]')
+      end
+
+      private
+
+      def add_terminal_number(post, options)
+        post[:MerchantTerminalNumber] = options[:merchant_terminal_number] || ' ' * 5
+      end
+
+      def add_money(post, money)
+        post[:Amount] = money.to_s.rjust(11,'0')
+      end
+
+      def add_operator_id(post, options)
+        post[:OperatorID] = options[:operator_id] || '0' * 8
+      end
+
+      def add_customer_data(post, options)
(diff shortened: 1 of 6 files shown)


lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,297 @@
...
+  module Billing #:nodoc:
+    class CtPaymentGateway < Gateway
+      self.test_url = 'https://test.ctpaiement.ca/v1/'
+      self.live_url = 'https://example.com/live'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = 'http://www.example.net/'
```
bpollack: I think Docs indirectly grabs this, so it's probably worth fixing.

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,297 @@
...
+      def purchase(money, payment, options={})
+        post = {}
+        add_terminal_number(post, options)
+        add_money(post, money)
+        add_operator_id(post, options)
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+        response = if payment.is_a?(String)
```
bpollack: You _might_ be able to fit this on one line via ternary, but your call; I find this very readable as-is.

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,297 @@
...
+        add_customer_data(post, options)
+
+        response = commit('verifyAccount', post)
+        send_ack(response.authorization)
+        response
+      end
+
+      def store(credit_card, options={})
+        requires!(options, :email)
+        post = {}
```
bpollack: This is 100% your call, but given you immediately initialize this, you could just do

```ruby
post = {
  LanguageCode: 'E',
  Name: credit_card.name.rjust(50, ' ')
  Email: options[:email].rjust(240, ' ')
}
```

which decreases some repetition. That said, this is 100% your call.

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,297 @@
...
+      }
+      CARD_BRAND = {
+        'american_express' => 'A',
+        'master' => 'M',
+        'diners_club' => 'I',
+        'visa' => 'V',
+        'discover' => 'O'
+      }
+
+      PARAMETERS_ORDER = [
```
bpollack: I don't actually see this variable used anywhere. Is this a remnant of something?

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,297 @@
...
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((&?auth-api-key=)[^&]*)i, '\1[FILTERED]').
+          gsub(%r((&?token:)[^&]*)i, '\1[FILTERED]').
```
bpollack: I'm confused about how this scrub method works. `token` and `cardNumber` both look as if they'd be in the base64-encoded part of the post, so I'm not clear how this filter would work in general. But beyond that, I'm also unclear why they end with `:` instead of `=`, since they're done as POST params. Could you walk me through what I'm missing here?

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,297 @@
...
+        post[:InputType] = 'I'
+        post[:LanguageCode] = 'E'
+      end
+
+      def add_payment(post, payment)
+        if payment.is_a?(String)
+          post[:Token] = split_authorization(payment)[3].strip
+        else
+          post[:CardType] = CARD_BRAND[payment.brand] || ' '
+          post[:CardNumber] = payment.number #.rjust(40,' ')
```
bpollack: Should this comment be killed?

test/remote/gateways/remote_ct_payment_certification_test.rb
```diff
@@ -0,0 +1,246 @@
...
+    @declined_card = credit_card('4502244713161718')
+    @options = {
+      billing_address: address,
+      description: 'Store Purchase',
+      merchant_terminal_number: '13366',
+      order_id: generate_unique_id[0,11]
+    }
+  end
+
+  def test1
```
bpollack: ~Do these test numbers correlate to ones that are supposed to be in the docs? Why are they all just `test<num>`?~ Okay, I get it; file name says yes.

lib/active_merchant/billing/gateways/ct_payment.rb
bpollack: Okay, I think I get what's going on after reading the tests: we're scrubbing the value they echo back in the response, not our submission. Ditto for the token.

_Should_ we scrub the base64 somehow? Do we do this for any other gateway? I'm asking because I'd generally consider preserving base64 strings a PAN leak, but I also don't want to remove all the debugging info…

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,297 @@
...
+
+      def parse(body)
+        JSON.parse(body)
+      end
+
+      def split_authorization(authorization)
+        authorization.split(';')
+      end
+
+      def commit(action, parameters)
```
bpollack: As far as I can tell, we always, without exception, call `send_ack` after `commit`, and always in the same way:

```ruby
response = commit(some_action, params)
send_ack(response.authorization)
response
```

Given that, I think it'd be worth thinking about just folding it directly into `commit`. It'd be one thing if we sometimes didn't send the `ACK`, but it seems like we literally always do. (Alternatively, if we do call `commit` without `send_ack` sometimes, or if we want some flexibility, I still suspect it'd be a win to just add a `..., send_ack: true)` param to `commit`.)

lib/active_merchant/billing/gateways/ct_payment.rb
nfarve: @bpollack yea this is a remanent of me getting stuck in rubocop looks. It was complaining that the one line ternary was too long so then I did multiline then it complained that it should be a if else statement. Going back and fixing that stuff now.

lib/active_merchant/billing/gateways/ct_payment.rb
nfarve: Good catch, scrubbing the payload.

lib/active_merchant/billing/gateways/ct_payment.rb
nfarve: Was working but putting it back due to your COBOL theory.

lib/active_merchant/billing/gateways/ct_payment.rb
nfarve: It can't be folded into the commit, it would end up in an endless loop since the send_ack is also a commit.

bpollack: It can still be folded; you'd just need to split out `commit` from `commit_raw` (or the other direction, `commit_with_ack` v. `commit`). Or just go the `send_ack:` param route, in which case `send_ack` would pass `false.`

test/fixtures.yml
```diff
@@ -219,6 +219,11 @@ credorax:
   merchant_id: 'merchant_id'
   cipher_key: 'cipher_key'
 
+# Working credentials, no need to replace
```
bpollack: Okay, last thing: I doubt these are working credentials, since they're the default, but if they are, never mind and we're good to go. (I suspect for them we just want to axe out the comment, not put in working creds.)

nfarve: Fixed. Thanks @bpollack for all these catches.

bpollack: You're welcome! You're good to ship now, just to reiterate.

nfarve: Waiting for certification test to be approved so they can give us the actual live url.

lib/active_merchant/billing/gateways/ct_payment.rb
nfarve: Gotcha. Updated!

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,258 @@
...
+        requires!(options, :order_id)
+        post = {}
+        add_terminal_number(post, options)
+        add_money(post, money)
+        add_operator_id(post, options)
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+        payment.is_a?(String) ? commit('purchaseWithToken', post) : commit('purchase', post)
```
curiousepic: Nitpick - add space before commit

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,258 @@
...
+        commit('verifyAccount', post)
+      end
+
+      def store(credit_card, options={})
+        requires!(options, :email)
+        post = {
+          LanguageCode: 'E',
+          Name: credit_card.name.rjust(50, ' '),
+          Email: options[:email].rjust(240, ' ')
+        }
```
curiousepic: NBD Nitpick - this kinda breaks the pattern of creating an empty hash and adding the fields to it piecemeal found in the other action methods and most other gateways.

bpollack: I actually suggested @nfarve do this in the first place. In my opinion, this is easier to read (especially with symbol keys) than the other style, and it's objectively less repetitive. The main reason we do the other style is because you do this pattern:

```ruby
post = {}
add_frob(post, options)
add_quux(post, options)
add_bar(post, options)
...
```

which in this case we're (partially) skipping since we can. Given that, I don't think emulating the old style purely for consistency's sake necessarily makes sense here—but I don't feel strongly about this, either.

lib/active_merchant/billing/gateways/ct_payment.rb
```diff
@@ -0,0 +1,258 @@
...
+          avs_result: AVSResult.new(code: response['avsStatus']),
+          cvv_result: CVVResult.new(response['cvv2Cvc2Status']),
+          test: test?,
+          error_code: error_code_from(response)
+        )
+        if send_ack
+          post = {
+            TransactionNumber: split_authorization(final_response.authorization)[0]
+          }
+          commit('ack', post, false)
```
curiousepic: Hmm... did you consider using a MultiResponse for this instead?

lib/active_merchant/billing/gateways/ct_payment.rb
curiousepic: Ah... well, same here TBH, this is fine.

lib/active_merchant/billing/gateways/ct_payment.rb
nfarve: I did... trying to remember why I went back to this way. 🤔

nfarve: @curiousepic for the MultiResponse wouldn't it need to be on the first commit and not nested inside of it?

curiousepic: Yep. Could be part of the commit method or it could be part of the actions, ala an auth+capture purchase. I'm honestly not sure if it's a good idea yet, but it seems like a pattern that already exists that we're recreating from scratch here. There could be a qualitative difference I can't see, though.

nfarve: Yea I had in each method, but this reduced lines of code.

bpollack: > @curiousepic for the MultiResponse wouldn't it need to be on the first commit and not nested inside of it?

This is moot at this point, but I'm confused. If I get what you're saying, you can control that via whether you pass `use_first_response` as `true` or not to `MultiResponse.run`, so that'd just be a slightly different approach than this; you'd end up with something like

```ruby
def commit_raw(action, parameters)
  ...current contents of commit method...
end

  MultiResponse.run(true) do |r|
    resp = r.commit_raw(action, parameters)
    r.commit_raw('ack', {TransactionNumber: split_authorization(resp.authorization)[0]})
  end
end
```

(To be clear, I'm not asking you change what you're doing here; just not clear on why `MultiResponse` wouldn't do the trick.)

nfarve: @bpollack I as more so asking how could it be done with the current commit structure. Definitely see how add a new commit `commit_raw` could achieve the goal though.

nfarve: @bpollack and @curiousepic new and improved with `MultiResponse`

bpollack:
Test failures look legit.

nfarve:
@bpollack definitely were. Should be fixed.

PR #2911, opened 2018-07-02
Spreedly: Fix some cases where the logged transcript is compressed

abarrak:
When API response gets beyond certain size, it's gizpped before it's sent back to the client.

Since [HTTP debug output](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/connection.rb#L140) is merely a collective string built over time, we finally get back a mixed transcript of plain text and compressed text, _which is undesirable_.

- This is an approach to avoid mixed transcript log.
- Another one is to analyze the string, and uncompress the zipped portion only, which might not always be applicable.


bpollack:
Thanks for the PR @abarrak . I don't really know that we want to support this; gzip compression is generally great for everyone involved. If the logging bothers you, I'd rather see a PR that can better handle the connection switching to `gzip`.

abarrak:
>  If the logging bothers you, I'd rather see a PR that can better handle the connection switching to gzip

@bpollack I'm not following your point here.
Gizp behavior is inconsistent, and occurs in some endpoints only.

The [logging bit](https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L986) is beyond AM, where [internal `::HTTP`](https://github.com/ruby/ruby/blob/trunk/lib/net/protocol.rb#L210) doesn't handle compressed string.

I'll submit another approach of handling inside `scrub`.

bpollack:
> @bpollack I'm not following your point here.

The gzip isn't endpoint-specific _as such_; it's based on the size of the content generated, and some endpoints generate enough content that our servers decide to gzip it if that's available. This line:

https://github.com/activemerchant/active_merchant/pull/2911/files#diff-3d8ec2e8f34093ada56bc96cef17f9ecR290

is working for your purposes by disabling gzip. That's not an acceptable way to resolve this issue. (This issue also applies to other gateways, incidentally, not just `spreedly_core.rb`, which is why a Spreedly Core-specific fix seems like the wrong way to go.) It seems like the issue you're trying to resolve here is that the transcripts contain the raw gzip data. I'm not sure I agree that's an issue, but if you do, I'd rather approach it by making the transcripts show the unzipped data, not disabling gzip for one particular gateway.

Is that enough background/info to help out here? I also got into some of this on the email you sent to me last week.

abarrak:
I agree with you that disabling gzip is not an optimal workaround.

Since you indicated this isn't specific to `SpreedlyCore`, I'll look into another mean to resolve it or keep it outside AM.

PR #2912, opened 2018-07-03
Firstdata E4 (GGE4) v27: Add new gateway

shasum:
This is based off of existing gateway [firstdata_e4](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/firstdata_e4.rb). v27 diverges significantly from existing v11 in terms of additional credentials for HMAC and multiple field changes including address & cvv fields. 

---

```
$ RUBYOPT=W0 be rake test:remote TEST=test/remote/gateways/remote_firstdata_e4_v27_test.rb 
.......................

Finished in 53.177556 seconds.
----------------------------------------------------------------------------------------------------------------------
23 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------
0.43 tests/s, 1.73 assertions/s
```

---

```
$ RUBYOPT=W0 be rake test:units TEST=test/unit/gateways/firstdata_e4_v27_test.rb 
.............................

Finished in 0.302474 seconds.
----------------------------------------------------------------------------------------------------------------------
29 tests, 136 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------
95.88 tests/s, 449.63 assertions/s
```

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
@@ -1,0 +1,444 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class FirstdataE4V27Gateway < Gateway
+      self.test_url = 'https://api.demo.globalgatewaye4.firstdata.com/transaction/v27'
+      self.live_url = 'https://api.globalgatewaye4.firstdata.com/transaction/v27'
+
+      TRANSACTIONS = {
+        sale:          '00',
+        authorization: '01',
+        verify:        '05',
+        capture:       '32',
+        void:          '33',
+        credit:        '34',
+        store:         '05'
+      }
+
+      SUCCESS = 'true'
+
+      SENSITIVE_FIELDS = [:cvdcode, :expiry_date, :card_number]
+
+      BRANDS = {
+        :visa => 'Visa',
+        :master => 'Mastercard',
+        :american_express => 'American Express',
+        :jcb => 'JCB',
+        :discover => 'Discover'
+      }
+
+      DEFAULT_ECI = '07'
+
+      self.supported_cardtypes = BRANDS.keys
+      self.supported_countries = ['CA', 'US']
+      self.default_currency = 'USD'
+      self.homepage_url = 'http://www.firstdata.com'
+      self.display_name = 'FirstData Global Gateway e4 v27'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        # Bank error codes: https://support.payeezy.com/hc/en-us/articles/203730509-First-Data-Global-Gateway-e4-Bank-Response-Codes
+        '201' => STANDARD_ERROR_CODE[:incorrect_number],
+        '531' => STANDARD_ERROR_CODE[:invalid_cvc],
+        '503' => STANDARD_ERROR_CODE[:invalid_cvc],
+        '811' => STANDARD_ERROR_CODE[:invalid_cvc],
+        '605' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        '522' => STANDARD_ERROR_CODE[:expired_card],
+        '303' => STANDARD_ERROR_CODE[:card_declined],
+        '530' => STANDARD_ERROR_CODE[:card_declined],
+        '401' => STANDARD_ERROR_CODE[:call_issuer],
+        '402' => STANDARD_ERROR_CODE[:call_issuer],
+        '501' => STANDARD_ERROR_CODE[:pickup_card],
+        # Ecommerce error codes: https://support.payeezy.com/hc/en-us/articles/203730499-eCommerce-Response-Codes-ETG-e4-Transaction-Gateway-Codes
+        '22' => STANDARD_ERROR_CODE[:invalid_number],
+        '25' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        '31' => STANDARD_ERROR_CODE[:incorrect_cvc],
+        '44' => STANDARD_ERROR_CODE[:incorrect_zip],
+        '42' => STANDARD_ERROR_CODE[:processing_error]
+      }
+
+      def initialize(options = {})
+        requires!(options, :login, :password, :key_id, :hmac_key)
+        @options = options
+
+        super
+      end
+
+      def authorize(money, credit_card_or_store_authorization, options = {})
+        commit(:authorization, build_sale_or_authorization_request(money, credit_card_or_store_authorization, options))
+      end
+
+      def purchase(money, credit_card_or_store_authorization, options = {})
+        commit(:sale, build_sale_or_authorization_request(money, credit_card_or_store_authorization, options))
+      end
+
+      def capture(money, authorization, options = {})
+        commit(:capture, build_capture_or_credit_request(money, authorization, options))
+      end
+
+      def void(authorization, options = {})
+        commit(:void, build_capture_or_credit_request(money_from_authorization(authorization), authorization, options))
+      end
+
+      def refund(money, authorization, options = {})
+        commit(:credit, build_capture_or_credit_request(money, authorization, options))
+      end
+
+      def verify(credit_card, options = {})
+        commit(:verify, build_sale_or_authorization_request(0, credit_card, options))
+      end
+
+      # Tokenize a credit card with TransArmor
+      #
+      # The TransArmor token and other card data necessary for subsequent
+      # transactions is stored in the response's +authorization+ attribute.
+      # The authorization string may be passed to +authorize+ and +purchase+
+      # instead of a +ActiveMerchant::Billing::CreditCard+ instance.
+      #
+      # TransArmor support must be explicitly activated on your gateway
+      # account by FirstData. If your authorization string is empty, contact
+      # FirstData support for account setup assistance.
+      #
+      # https://support.payeezy.com/hc/en-us/articles/203731189-TransArmor-Tokenization
+      def store(credit_card, options = {})
+        commit(:store, build_store_request(credit_card, options), credit_card)
+      end
+
+      def verify_credentials
+        response = void('0')
+        response.message != 'Unauthorized Request. Bad or missing credentials.'
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript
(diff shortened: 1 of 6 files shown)


shasum:
@activemerchant/spreedly-runtime [ENRG-7445](https://spreedly.atlassian.net/browse/ENRG-7445)

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -0,0 +1,445 @@
...
+          :cvv_result => response[:cvv2],
+          :error_code => standard_error_code(response)
+        )
+      end
+
+      def headers(method, url, request)
+        content_type = "application/xml"
+        content_digest = Digest::SHA1.hexdigest(request)
+        sending_time = Time.now.utc.iso8601
+        payload = method + "\n" + content_type + "\n" + content_digest + "\n" + sending_time + "\n" + url.split(".com")[1]
```
shasum: It's important that escape character `\n` is enclosed in double quotes.

shasum: @activemerchant/spreedly-runtime ☝️

bpollack: Assuming this is re. RuboCop, RuboCop ought to be smart enough to allow double quotes for strings that need it (e.g., escapes like what you're doing, or `"#{injections}"`); are you seeing it get irritated about these?

shasum: https://travis-ci.org/activemerchant/active_merchant/jobs/399726009#L596

You're right. It's the `".com"` at the end of the line that it complains about.

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -0,0 +1,445 @@
...
+        end
+      end
+
+      def add_credit_card_eci(xml, credit_card, options)
+        eci = if credit_card.is_a?(NetworkTokenizationCreditCard) && credit_card.source == :apple_pay && card_brand(credit_card) == 'discover'
+          # Discover requires any Apple Pay transaction, regardless of in-app
+          # or web, and regardless of the ECI contained in the PKPaymentToken,
+          # to have an ECI value explicitly of 04.
+          '04'
+        else
```
shasum: Rubocop unhappy with indentation of `else` relative to `if`. Open to ideas on how best to make it happy.

test/unit/gateways/firstdata_e4_v27_test.rb
```diff
@@ -0,0 +1,980 @@
...
+    - IE=Edge,chrome=1
+    content-length:
+    - '49'
+    connection:
+    - Close
+  body_exist: true
+  http_version: '1.1'
+  socket:
+    RESPONSE
+    YAML.load(yamlexcep)
```
shasum: Rubocop suggests `YAML.safe_load()`. But using it throws an exception on objects referenced within the document.

> Psych::DisallowedClass: Tried to load unspecified class: Net::HTTPUnauthorized

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
bpollack: It wants the `else` lined up with the `if` (and will in general want the whole block appropriately indented), so line 224 should start two spaced after the `if`, and the `else` should align with the `if` also.

bpollack: You can run `rubocop --auto-correct` to see what it suggests, BTW.

test/unit/gateways/firstdata_e4_v27_test.rb
bpollack: `YAML.load` will deserialize arbitrary Ruby classes if you specify the object with a `!` (so the `!ruby/exception:ActiveMerchant::ResponseError` here), which is really dangerous (imagine someone deserializing the `File` class with reading from `/etc/passwd` or the like). `safe_load` only allows whitelisted classes. In this case we control the YAML, so the RuboCop recommendation isn't _strictly_ necessary, but it's still a good habit. The only trick is you have to tell it which classes are okay. In this case, if you change `YAML.load(yamlexcep)` to `YAML.safe_load(yamlexcep, ['Net::HTTPBadRequest', 'ActiveMerchant::ResponseError'])` (based on the class serializations on line 857 and 859), both RuboCop and your code will be happy.

shasum:
@bpollack Thanks for the tip on rubocop auto-correct and `yaml` whitelist. The CI checks have passed. Unit and remote tests are good.

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -0,0 +1,445 @@
...
+        void:          '33',
+        credit:        '34',
+        store:         '05'
+      }
+
+      SUCCESS = 'true'
+
+      SENSITIVE_FIELDS = [:cvdcode, :expiry_date, :card_number]
+
+      BRANDS = {
```
bpollack: I'm confused about the split between `BRANDS` and `E4_BRANDS`. [We only use `BRANDS` for `supported_cardtypes`](https://github.com/activemerchant/active_merchant/pull/2912/files#diff-8cf7977aecc61a9b94a9aa5bd7c9a55aR29), which initially made me think that `E4_BRANDS` was totally redundant, but then [we do use `E4_BRANDS` for card types](https://github.com/activemerchant/active_merchant/pull/2912/files#diff-8cf7977aecc61a9b94a9aa5bd7c9a55aR320) later on. I _think_ we're doing this because it was a backwards-compatibility kluge inherited from the original Firstdata gateway; if that's the case, then can we kill it, since we're introducing a specific v27 upgrade here? I think we could just have `BRANDS` and be done.

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -0,0 +1,445 @@
...
+      def scrub(transcript)
+        transcript
+          .gsub(%r((<Card_Number>).+(</Card_Number>)), '\1[FILTERED]\2')
+          .gsub(%r((<CVDCode>).+(</CVDCode>)), '\1[FILTERED]\2')
+          .gsub(%r((<Password>).+(</Password>))i, '\1[FILTERED]\2')
+          .gsub(%r((<CAVV>).+(</CAVV>)), '\1[FILTERED]\2')
+          .gsub(%r((CARD NUMBER\s+: )#+\d+), '\1[FILTERED]')
+      end
+
+      def supports_network_tokenization?
```
bpollack: I can't tell what, if anything, we use this for. Do we need it?

test/fixtures.yml
```diff
@@ -331,6 +331,12 @@ firstdata_e4:
   login: SD8821-67
   password: T6bxSywbcccbJ19eDXNIGaCDOBg1W7T8
 
+firstdata_e4_v27:
```
bpollack: If these creds are legit and it's okay to have them, we should probably stick the `# Working credentials, no need to replace` comment here; otherwise, we should probably nuke 'em.

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -0,0 +1,445 @@
...
+          :cvv_result => response[:cvv2],
+          :error_code => standard_error_code(response)
+        )
+      end
+
+      def headers(method, url, request)
+        content_type = 'application/xml'
+        content_digest = Digest::SHA1.hexdigest(request)
+        sending_time = Time.now.utc.iso8601
+        payload = method + "\n" + content_type + "\n" + content_digest + "\n" + sending_time + "\n" + url.split('.com')[1]
```
bpollack: This is nitpicky, but `payload = [method, content_type, content_digest, sending_time, url.split('.com')[1]].join("\n")` might be easier to read.

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -0,0 +1,445 @@
...
+        end
+      end
+
+      def money_from_authorization(auth)
+        _, _, amount = auth.split(/;/, 3)
+        amount.to_i
+      end
+
+      def message_from(response)
+        if(response[:faultcode] && response[:faultstring])
```
bpollack: Parens here (and the following `elsif`) are strictly unnecessary

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -0,0 +1,445 @@
...
+          xml.tag! 'CardType', card_type(credit_card.brand)
+
+          add_credit_card_eci(xml, credit_card, options)
+          add_credit_card_verification_strings(xml, credit_card, options)
+        end
+      end
+
+      def add_credit_card_eci(xml, credit_card, options)
+        eci = if credit_card.is_a?(NetworkTokenizationCreditCard) && credit_card.source == :apple_pay && card_brand(credit_card) == 'discover'
+                # Discover requires any Apple Pay transaction, regardless of in-app
```
bpollack: wat.

They should just ignore the ECI then, or automatically set it to `04`. This is silly. (Their requirement, not your code.)

jasonwebster: Yes. This is a failure of abstraction at many levels (the gateway, FirstData, and the network) that this leaks all the way here, however, doing this is the only way we could have this "just work" for people without knowing this particular piece of arcana. (I'm the one that patched the original FirstData E4 gateway that this is copied from with this logic.)

bpollack: Yeah, I'm _really_ not questioning that the code needs to be here; I just find the gateway requirements a bit silly. It reminds me unpleasantly of when I used to do Win32 coding and you'd end up with `foo.dwSomething = 6; // required for compatibility mode` and the like.

lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
bpollack: Resolved this by killing `E4_BRANDS`.

bpollack:
Thank you, @shasum ! Merged to master.

PR #2915, opened 2018-07-05
GlobalCollect: Don't overwrite contactDetails

curiousepic:
Previously, if there was an address provided, it would overwrite the
contactDeatails field which may have already had an email field added.
Now we have a conditional for both email and phone number elements
within contactDetails.

Also updates the sandbox url to the current endpoint.

Remote:
100% passed

Unit:
17 tests, 79 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -4,7 +4,7 @@ module ActiveMerchant #:nodoc:
       self.display_name = 'GlobalCollect'
       self.homepage_url = 'http://www.globalcollect.com/'
 
-      self.test_url = 'https://api-sandbox.globalcollect.com/'
+      self.test_url = 'https://eu.sandbox.api-ingenico.com/'
       self.live_url = 'https://api.globalcollect.com/'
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -154,13 +154,9 @@ def add_customer_data(post, options, payment = nil)
         post['order']['companyInformation'] = {
           'name' => options[:company]
         }
-        post['order']['contactDetails'] = {
-          'emailAddress' => options[:email]
-        }
+        post['order']['contactDetails']['emailAddress'] = options[:email] if options[:email]
```
bpollack: Do we know that `post['order']['contactDetails']` will always be defined here? I'm wondering if it might be worth adding `post['order']['contactDetails'] ||= {}` before this line. That might be redundant though.

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -169,10 +165,10 @@ def add_refund_customer_data(post, options)
           post['customer']['address'] = {
             'countryCode' => address[:country]
           }
-          post['customer']['contactDetails'] = {
-            'emailAddress' => options[:email],
-            'phoneNumber' => address[:phone]
-          }
+          post['customer']['contactDetails']['emailAddress'] = options[:email] if options[:email]
```
bpollack: Same here; wondering if we need `post['customer']['contactDetails'] ||= {}` to be safe.

lib/active_merchant/billing/gateways/global_collect.rb
bpollack: Okay, [`nestable_hash`](https://github.com/activemerchant/active_merchant/pull/2915/files/d83f99dc537c52e6283d99194d4c514bf046ab71#diff-476876f224070b3b093c610adc75d562R321) makes this work, so ship it!

Issue #2916, opened 2018-07-06
Stripe: Contactless magstripe mode not set when icc_data is present


abhiin1947:
When sending a magstripe contactless transaction via stripe, the gateway does not set the `contactless_magstripe_mode` on the post even though `creditcard.read_method` is set to `contactless_magstripe`.

PR #2917, opened 2018-07-06
Set contactless_magstripe_mode when transaction is contactless magstripe. Fix silently failing tests.

abhiin1947:
# What was the problem?
- When processing a contactless magstripe transaction if `icc_data` on the credit card is present, stripe gateway does not set `read_method` on the `post` to `contactless_magstripe_mode`

Resolves https://github.com/activemerchant/active_merchant/issues/2916

# Why this change?
- Set `contactless_magstripe_mode` if the read method is contactless magstripe.

# Test failures
- Activemerchant does have a test to make sure that `contactless_magstripe_mode` flag is set when it is part of EMV data. But these tests did not assert that the check is valid.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 * GlobalCollect: Don't overwrite contactDetails [curiousepic] #2915
 * Pin Payments: Pass reference for statement desc [curiousepic] #2919
 * FirstData: introduce v27 gateway [shasum] #2912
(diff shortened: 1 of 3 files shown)


ryanbalsdon:
Can you add a line to the changelog, too? https://github.com/activemerchant/active_merchant/blob/master/CHANGELOG

PR #2110, opened 2016-05-20
PaymentHighway gateway

hasanen:
Gateway for Finnish provider PaymentHighway(https://paymenthighway.fi/).

PR #2920, opened 2018-07-10
Add the Payment Highway gateway

matiaskorhonen:
Adds support for the Payment Highway gateway (https://www.paymenthighway.io/)

Rebased from #2110

### DIFF
--- a/lib/active_merchant/billing/gateways/payment_highway.rb
+++ b/lib/active_merchant/billing/gateways/payment_highway.rb
@@ -1,0 +1,172 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PaymentHighwayGateway < Gateway
+      self.money_format = :cents
+      self.test_url = 'https://v1-hub-staging.sph-test-solinor.com/'
+      self.live_url = 'https://v1.api.paymenthighway.io'
+
+      self.supported_countries = ['FI']
+      self.default_currency = 'EUR'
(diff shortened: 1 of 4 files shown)


test/fixtures.yml
```diff
@@ -679,6 +679,13 @@ payment_express:
   login: LOGIN
   password: PASSWORD
 
+# Working credentials, no need to replace
+payment_highway:
```
bpollack: I doubt these are actually working credentials

test/unit/gateways/payment_highway_test.rb
```diff
@@ -0,0 +1,276 @@
...
+  private
+
+  def pre_scrubbed
+    '
+   Run the remote tests for this gateway, and then put the contents of transcript.log here.
+    '
+  end
+
+  def post_scrubbed
+    '
```
bpollack: Scrubs are missing

lib/active_merchant/billing/gateways/payment_highway.rb
```diff
@@ -0,0 +1,172 @@
...
+          'card' => card_to_json(card)
+        }
+
+        payload = add_ip(payload, options)
+        payload = add_order_id(payload, options)
+
+        commit("/transaction/#{transaction_id}/debit", payload.to_json, transaction_id)
+      end
+
+      def order_status(order_id)
```
bpollack: We don't encourage exposing methods like this in the public part of the class API.

PR #2921, opened 2018-07-10
Upgrade RuboCop to 0.58.1

bpollack:
This version includes a bugfix (https://github.com/rubocop-hq/rubocop/issues/6092)
that should allow us to add PayPal-related code to the RuboCop checks.
For this patch, though, only RuboCop is upgraded (and a tiny tweak to
Mercury implemented to handle a new RuboCop test); the PayPal compliance
changes will be a separate commit.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'jruby-openssl', :platforms => :jruby
-gem 'rubocop', '~> 0.57.2', require: false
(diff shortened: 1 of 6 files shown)


bpollack:
Huh. Dangit. This won't fix things, because RuboCo is incompatible with ruby-parse 2.5.1.1, which is the version that fixes this bug. 🤦‍♂️ 

I still think this is worth landing, but it doesn't get us closer to having PayPal included in tests.

bpollack:
I'm gonna kill this and fold it into #2928 as a demo.

PR #2922, opened 2018-07-10
Update contributing document regarding gateways

bdewater:
I feel it's time we make this more explicit instead if keeping PRs unmerged for months on end.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,16 @@
 # Contributing guidelines
 
-We gladly accept bugfixes and new gateways. Please follow the guidelines here to ensure your work is accepted.
+We gladly accept bugfixes, but are not actively looking to add new gateways. Please follow the guidelines here to ensure your work is accepted.
 
 ## New Gateways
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,10 +1,14 @@
 # Contributing guidelines
 
-We gladly accept bugfixes and new gateways. Please follow the guidelines here to ensure your work is accepted.
+We gladly accept bugfixes, but are not actively looking to add new gateways. Please follow the guidelines here to ensure your work is accepted.
 
 ## New Gateways
 
-Please see the [ActiveMerchant Guide to Contributing a new Gateway](https://github.com/activemerchant/active_merchant/wiki/contributing) for information on adding a new gateway to ActiveMerchant.
+We're not taking on many new gateways at the moment. The team maintining ActiveMerchant is small and with the limited resources available, we generally prefer not to support a gateway than support a gateway poorly.
```
WilsonChiang: Typo here. `maintining`

bpollack: Probably should also be "...prefer not to support a gateway than *to* support a gateway poorly"; it reads a bit awkwardly the way I wrote it originally.

bdewater:
Fixed both issues and added a reference to discuss the plugin/gem idea. Please re-review 😄

PR #2928, opened 2018-07-11
Clean up the build process

bpollack:
Right now, each Gemfile has a *huge* amount of redundancy, so any change to dependencies requires altering a minimum of five files. On top of that, we run RuboCop way too much.

These patches do two things:
 - Clean up the Gemfiles to remove redundancy
 - Clean up Travis to only do RuboCop once, not every build
 - Demonstrate the two working together by upgrading RuboCop only once, not for every Gemfile

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,11 @@ gemfile:
 - Gemfile.rails50
 - Gemfile.rails42
 
-matrix:
+jobs:
   include:
-    - rvm: 2.3
-      gemfile: Gemfile.rails42
+      rvm: 2.5
+      gemfile: Gemfile
+      script: bundle exec rubocop --parallel
 
 notifications:
   email:
--- a/Gemfile.rails42
+++ b/Gemfile.rails42
@@ -1,12 +1,3 @@
-source 'https://rubygems.org'
-gemspec
-
-gem 'jruby-openssl', :platforms => :jruby
-gem 'rubocop', '~> 0.57.2', require: false
-
-group :test, :remote_test do
-  # gateway-specific dependencies, keeping these gems out of the gemspec
-  gem 'braintree', '>= 2.50.0'
-end
+eval File.read('Gemfile')
 
 gem 'activesupport', '~> 4.2.0'
--- a/lib/active_merchant/billing/gateways/mercury.rb
+++ b/lib/active_merchant/billing/gateways/mercury.rb
@@ -103,7 +103,7 @@ module ActiveMerchant #:nodoc:
           xml.tag! 'Transaction' do
             xml.tag! 'TranType', 'Credit'
             xml.tag! 'TranCode', action
-            if options[:allow_partial_auth] && (action == 'PreAuth' || action == 'Sale')
+            if options[:allow_partial_auth] && ['PreAuth', 'Sale'].include?(action)
               xml.tag! 'PartialAuth', 'Allow'
             end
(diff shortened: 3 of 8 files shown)


.travis.yml
```diff
@@ -18,6 +18,13 @@ matrix:
     - rvm: 2.3
       gemfile: Gemfile.rails42
```
bdewater: Can we axe this `matrix` block while we're here? The rvm/gemfile combinations already do it :)

.travis.yml
```diff
@@ -18,6 +18,13 @@ matrix:
...
 
+jobs:
+  include:
+    - stage: RuboCop
+      rvm: 2.5
+      gemfile: Gemfile
+      script: bundle exec rake rubocop
```
bdewater: 🎉

.travis.yml
```diff
@@ -18,6 +18,12 @@ matrix:
     - rvm: 2.3
       gemfile: Gemfile.rails42
 
+jobs:
+  include:
+      rvm: 2.5
+      gemfile: Gemfile
+      script: bundle exec rake rubocop
```
bdewater: We can add the `--parallel` flag 🏎 since we will not autocorrect functionality here.

bpollack:
@bdewater Integrated your changes. The build failed a couple times due to me rebasing a bit too much, but I think we're clean now.

lib/active_merchant/billing/gateways/mercury.rb
```diff
@@ -103,7 +103,7 @@ def build_non_authorized_request(action, money, credit_card, options)
           xml.tag! 'Transaction' do
             xml.tag! 'TranType', 'Credit'
             xml.tag! 'TranCode', action
-            if options[:allow_partial_auth] && (action == 'PreAuth' || action == 'Sale')
+            if options[:allow_partial_auth] && ['PreAuth', 'Sale'].include?(action)
```
curiousepic: I don't know if this was intended to go in with these changes, but it looks fine.

bpollack: It was; RuboCop 0.58.1 has a new check, and that was the only violator, so I just figured I'd roll with it as part of the RuboCop upgrade.

Gemfile.rails42
```diff
@@ -1,12 +1,3 @@
...
-gemspec
-
-gem 'jruby-openssl', :platforms => :jruby
-gem 'rubocop', '~> 0.57.2', require: false
-
-group :test, :remote_test do
-  # gateway-specific dependencies, keeping these gems out of the gemspec
-  gem 'braintree', '>= 2.50.0'
-end
+eval File.read('Gemfile')
```
varyonic: FWIW Bundler has `eval_gemfile`, eg. https://github.com/search?q=eval_gemfile&type=Code

bpollack: Do you know if there's an advantage? I just looked at its implementation, and there's nothing obvious (it handles certain errors a bit better, after a fashion, but it's not as if a stack trace saying the file couldn't be found is gonna be hard to parse). I'd definitely have done this if I'd known about it; just wondering if it actually makes a difference.

varyonic: Not much, mostly cosmetic and reflects this is a common pattern.

bpollack: Also allowed shoving the other `Gemfile`s in their own directory. Thank you for the suggestion, @varyonic ; landed in dde5ab42d5c9

PR #2930, opened 2018-07-12
Adyen: Set Default Name for Apple Pay Transactions

nfarve:
Name is required for Adyen but not always paseed on Apple Pay
transactions. This creates the default name place holder of `Not
Provided` on apply pay transactions. Updates one remote test with new
message.

......................

22 tests, 101 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Started

34 tests, 86 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -179,6 +179,7 @@ module ActiveMerchant #:nodoc:
         }
 
         card.delete_if{|k,v| v.blank? }
+        card[:holderName] ||= 'Not Provided' if credit_card.is_a?(NetworkTokenizationCreditCard)
         requires!(card, :expiryMonth, :expiryYear, :holderName, :number)
         post[:card] = card
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -179,6 +179,7 @@ def add_card(post, credit_card)
         }
 
         card.delete_if{|k,v| v.blank? }
+        card[:holderName] ||= 'Not Provided' if credit_card.is_a?(NetworkTokenizationCreditCard)
```
curiousepic: This won't overwrite if the name is an empty string, not sure if that's possible, but worth looking into.

jasonwebster: It will because of the line above, which removes all `blank?` values (and their keys) from the `card` hash.

curiousepic: 👍

PR #2934, opened 2018-07-17
Beanstream: Update to use api key with login credentials

nfarve:
Beanstream is requiring that users that connect via username and
password also supply a api key. 2 echeck tests are failing unrelated to
this change.

Loaded suite test/remote/gateways/remote_beanstream_test

41 tests, 185 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.122% passed

Loaded suite test/unit/gateways/beanstream_test

23 tests, 108 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_beanstream_test.rb
+++ b/test/remote/gateways/remote_beanstream_test.rb
@@ -384,6 +384,7 @@ class RemoteBeanstreamTest < Test::Unit::TestCase
     assert_scrubbed(@visa.number, clean_transcript)
     assert_scrubbed(@visa.verification_value.to_s, clean_transcript)
     assert_scrubbed(@gateway.options[:password], clean_transcript)
+    assert_scrubbed(@gateway.options[:api_key], clean_transcript)
   end
 
   private
--- a/lib/active_merchant/billing/gateways/beanstream.rb
+++ b/lib/active_merchant/billing/gateways/beanstream.rb
@@ -199,6 +199,7 @@ module ActiveMerchant #:nodoc:
         transcript.
           gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
           gsub(/(&?password=)[^&\s]*(&?)/, '\1[FILTERED]\2').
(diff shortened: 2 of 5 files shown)


curiousepic:
Should this be added to the scrub method in `beanstream.rb`?

nfarve:
@curiousepic updated!

nfarve:
@bpollack I update fixtures which should update all the tests to include the `api_key`. I was using the local and forgot to update the master. 

Also it looks like tests are failing in Travis because the "module has too many lines" any thoughts on how to get around this?

nfarve:
@bpollack deleted some spacing for the time being.

bpollack:
@nfarve #2935 should allow this to pass if that looks okay with you.

test/remote/gateways/remote_beanstream_test.rb
```diff
@@ -384,6 +384,7 @@ def test_transcript_scrubbing
     assert_scrubbed(@visa.number, clean_transcript)
     assert_scrubbed(@visa.verification_value.to_s, clean_transcript)
     assert_scrubbed(@gateway.options[:password], clean_transcript)
+    assert_scrubbed(@gateway.options[:api_key], clean_transcript)
```
curiousepic: Should probably also be added to the example transcript in the scrubbing unit test

bpollack:
> @nfarve #2935 should allow this to pass if that looks okay with you.

@nfarve Just landed, so if you rebase, the Rubocop errors will go away.

test/remote/gateways/remote_beanstream_test.rb
nfarve: It actually is not present in the transcript. I am more so thinking of this as a precaution.

curiousepic: Gotcha 👍

PR #2941, opened 2018-07-24
Barclaycard Smartpay: bump API to v30

bpollack:
This currently shouldn't change any behavior, but preps for handling third-party payouts and also serves as a marker that the API package should run cleanly on v30 of the API.

Unit: 25 tests, 114 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 29 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@
 * FirstPay: Update Fields For Recurring Payments [nfarve] #2940
 * BlueSnap: Update list of supported countries [curiousepic] #2942
 * Remove unused handle_response method [bl] #2309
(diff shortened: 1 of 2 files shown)


bpollack:
Actually hold off a second; going to do this slightly differently.

bpollack:
Okay, reopening. Changed the API version marker to a constant.

bpollack:
Yeah, I honestly was surprised it went in cleanly, but the remote tests pass—and that seems legit based on the docs so far, too.

PR #2945, opened 2018-07-25
Use SHIPTONAME instead of `full_name` for Payflow Express

filipebarcos:
This has a _potential_ backward compatibility issue, where people would expect `full_name` to be used. But this is causing a awkward UX for buyers. If you have Payflow Express, the buyer select a shipping address, the name on that Shipping address is different from the account owner, when selected, we would receive/use the account owner name, not the shipping address name.


filipebarcos:
@pi3r done ✅

PR #2948, opened 2018-08-06
BlueSnap: Handle 403 responses

curiousepic:
403 Forbidden responses due to insufficient account permissions were
causing parsing exception errors. This handles them with a bespoke
response passing the raw text as message.

Remote (1 unrelated failure):
25 tests, 72 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96% passed

Unit:
19 tests, 72 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -116,7 +116,7 @@ module ActiveMerchant
 
       def verify_credentials
         begin
-          ssl_get("#{url}/nonexistent", headers)
+          ssl_get(url.to_s, headers)
         rescue ResponseError => e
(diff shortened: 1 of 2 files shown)


curiousepic:
Updated with a small change and also a fix for a change in gateway behavior that seemed to have been introduced since I started this branch:

    This also fixes the verify_credentials call; a change at the gateways
    caused the response given for a bogus url (url/nonexistent) to be a
    success with a different message. Hitting the root url now seems
    sufficient to provoke a 401.

curiousepic:
Nevermind, there isn't really precedent for what I was going to do and is probably not worth the effort.

PR #2951, opened 2018-08-08
Adyen: Add RecurringProcessingModel

nfarve:
With visa updates, Adyen now requires the type of recurring processing
model for all recurring transactions. This adds the neccessary field.

34 tests, 86 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

......................

22 tests, 102 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 * Paymentez: support partial refunds [bpollack] #2959
 * Payflow: allow support for partial captures [pi3r] #2952
 * CT Payment: Update How Address is Passed [nfarve] #2960
+* Adyen: Add RecurringProcessingModel [nfarve] #2951
 
 == Version 1.81.0 (July 30, 2018)
 * GlobalCollect: Don't overwrite contactDetails [curiousepic] #2915
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -148,6 +148,7 @@ module ActiveMerchant #:nodoc:
           currency: options[:currency] || currency(money)
         }
         post[:amount] = amount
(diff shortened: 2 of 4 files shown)


curiousepic:
Looks like this is added only for Store actions, but I think these visa card on file designations need to be sendable for any type of transaction that involves a card stored elsewhere?

nfarve:
@curiousepic from the post, it looks like this only applies to recurring transactions which are stored cards: https://www.adyen.com/blog/update-on-visas-new-card-on-file-transaction-type

curiousepic:
@nfarve I don't think recurring in this case strictly means cards stored at Adyen... LMK if you want to chat to clarify.

nfarve:
Hmm Let me know think about it some more then. Whats throwing me off is this statement: 
`By default Adyen will flag recurring transactions as 'subscription'. `

nfarve:
@curiousepic It looks like the only place it could be added is for recurring transactions, which we don't seem to currently support. https://docs.adyen.com/developers/api-reference/recurring-api

curiousepic:
Hmmmmm. Well, `store` does currently utilizes https://docs.adyen.com/developers/features/recurring-payments/store-payment-details
It would be nice if they defined "recurring" a little more explicitly in their post.

curiousepic:
Their docs show the field is supported for auth/purches: https://docs.adyen.com/developers/api-reference/payments-api#paymentrequest 
This makes me think it should be passable with any transaction, not just when storing a card with Adyen.

nfarve:
@curiousepic okay moved to `add_invoice`

nfarve:
@curiousepic It did. Fixed

PR #2954, opened 2018-08-08
Add support for Paymentwall

aguayma:
Added support for [Paymentwall](https://www.paymentwall.com/). Test for available response codes from Paymentwall API.


abunsen:
Includes scrubbing as well.

Issue #2955, opened 2018-08-08
[Cardknox Gateway]The response don't include the error_code


Ahmed-Hussein93:
missing the error_code in the response

PR #2956, opened 2018-08-08
 [CardKnox][fix issue #2955] add error_code to the response

Ahmed-Hussein93:


bpollack:
Thanks for the patch, @Ahmed-Hussein93. Could you add at least a unit test to prevent a regression?

PR #2959, opened 2018-08-13
Paymentez: add support for partial refunds

bpollack:
Note, the remote tests that fail do actually include the new noe, but just with "The partial refund is not supported by carrier", which means everything is happening correctly; they just misconfigured stuff on their side.

Unit: 19 tests, 77 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 17 tests, 44 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
88.2353% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,7 @@
 * BlueSnap: Handle 403 responses [curiousepic] #2948
 * BlueSnap: Add StoreCard Field [nfarve] #2953
 * Worldpay: support installments [bpollack] #2957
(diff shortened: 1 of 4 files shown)


deedeelavinder:
CI fails on `Gemset '' does not exist`

bpollack:
@deedeelavinder Yeah, I saw that; I think it's bogus. I kicked off the build against just in case.

bpollack:
w00t, yep. Merging.

PR #2965, opened 2018-08-20
Provide fallback for nil `socket.io.cipher`

bdewater:
We've seen a few odd cases where this array was nil. Instead of blowing up the transaction just log as 'unknown'.

```ruby
irb(main):001:0> nil&.[](0) || 'unknown'
=> "unknown"
irb(main):002:0> %w()&.[](0) || 'unknown'
=> "unknown"
irb(main):003:0> %w(FOO)&.[](0) || 'unknown'
=> "FOO"
```

### DIFF
--- a/lib/active_merchant/net_http_ssl_connection.rb
+++ b/lib/active_merchant/net_http_ssl_connection.rb
@@ -4,7 +4,7 @@ module NetHttpSslConnection
   refine Net::HTTP do
     def ssl_connection
       return {} unless @socket.present?
-      { version: @socket.io.ssl_version, cipher: @socket.io.cipher[0] }
+      { version: @socket.io.ssl_version, cipher: @socket.io.cipher&.[](0) || 'unknown' }
     end
   end
 end


bpollack:
Do you know _how_ it could even be `nil`? Ruby's using some super-secret cipher that's not even worthy of the array that would dare contain it?

bdewater:
It baffled us as well, and for some reason it's consistently happening with only Paypal Payflow. Makes you wonder what `ssl_version` would return if it doesn't blow up...

bdewater:
Synchronicity is real 😅 the nil error spiked for us around the time https://www.paypal-status.com/history/eventdetails/8400 happened (hours after this PR was opened).

bpollack:
@bdewater Do you want to kill this? I have time to merge and test today otherwise.

bdewater:
We monkeypatched this in locally and I believe the errors we saw went away. Let me circle back and see what the logs say about the cipher in these cases.

bpollack:
@bdewater Ping.

pi3r:
I'm gonna close that for now. Will reopen once we revisit this issue on our side. Thanks for keeping track of it :)

PR #2968, opened 2018-08-27
Adyen: add Maestro and Electron support

bpollack:
Unit: 23 tests, 109 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 35 tests, 88 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

PR #2969, opened 2018-08-27
Adyen: add support for customRoutingFlag

bpollack:
Note that this can't really be tested in the sandbox (it's changing how Adyen processes the card, so the sandbox doesn't do anything with it), so remote tests are omitted.

Also note that this depends on/incorporates #2968 (since they both need to muck with `adddionalData`), so please take a look at that first/in addition to this.

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -120,6 +120,7 @@ module ActiveMerchant #:nodoc:
         post[:merchantOrderReference] = options[:merchant_order_reference] if options[:merchant_order_reference]
         post[:additionalData] ||= {}
         post[:additionalData][:overwriteBrand] = normalize(options[:overwrite_brand]) if options[:overwrite_brand]
+        post[:additionalData][:customRoutingFlag] = options[:custom_routing_flag] if options[:custom_routing_flag]
       end
 
       def add_shopper_interaction(post, payment, options={})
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -114,9 +114,13 @@ def add_extra_data(post, payment, options)
...
         post[:shopperReference] = options[:shopper_reference] if options[:shopper_reference]
         post[:fraudOffset] = options[:fraud_offset] if options[:fraud_offset]
-        post[:selectedBrand] = options[:selected_brand] || NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
+        post[:selectedBrand] = options[:selected_brand] if options[:selected_brand]
+        post[:selectedBrand] ||= NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
         post[:deliveryDate] = options[:delivery_date] if options[:delivery_date]
         post[:merchantOrderReference] = options[:merchant_order_reference] if options[:merchant_order_reference]
+        post[:additionalData] ||= {}
+        post[:additionalData][:overwriteBrand] = normalize(options[:overwrite_brand]) if options[:overwrite_brand]
+        post[:additionalData][:customRoutingFlag] = options[:custom_routing_flag] if options[:custom_routing_flag]
```
bpollack: This line, and its test, are the only actual new feature introduced in this PR. The rest is a separate commit getting picked up by this PR that's also identically in #2968.

PR #2971, opened 2018-08-29
Adyen: Add support for GooglePay

dtykocki:
Unit:
24 tests, 113 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
36 tests, 90 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -121,6 +121,7 @@ module ActiveMerchant #:nodoc:
         post[:additionalData] ||= {}
         post[:additionalData][:overwriteBrand] = normalize(options[:overwrite_brand]) if options[:overwrite_brand]
         post[:additionalData][:customRoutingFlag] = options[:custom_routing_flag] if options[:custom_routing_flag]
+        post[:additionalData]['paymentdatasource.type'] = NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
       end
 
       def add_shopper_interaction(post, payment, options={})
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -121,6 +121,7 @@ def add_extra_data(post, payment, options)
         post[:additionalData] ||= {}
         post[:additionalData][:overwriteBrand] = normalize(options[:overwrite_brand]) if options[:overwrite_brand]
         post[:additionalData][:customRoutingFlag] = options[:custom_routing_flag] if options[:custom_routing_flag]
+        post[:additionalData]["paymentdatasource.type"] = NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
```
dtykocki: Note: Setting the `paymentdatasource.type` field is required per an Adyen contact. Unfortunately, I've been unable to find any mention of this field in their public documentation.

dtykocki: I'm pretty sure, but only because it's what Adyen said to do and the GooglePay tests passed right after adding `paymentdatasource.type`. I don't mind trying `[:paymentdatasource][:type]` and see what happens.

dtykocki: Also, Travis isn't happy with this PR, I'll take a look.

dtykocki: We'll for sure need to go with `paymentdatasource.type`. I'm seeing the same "Additional data fields acquirerCode and authorisation MID required for external MPI" error when sending `[:paymentdatasource][:type]`.

dtykocki: Rubocop warning has also been fixed. Waiting on Travis now.

PR #2974, opened 2018-08-30
PayU Latam: add partial capture support

bpollack:
Note that PayU Latam's sandbox does not support capture, so while I
added remote tests, they're more...let's call it hypothetical.

Failing remote tests are unrelated to this change

Unit: 28 tests, 104 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 30 tests, 68 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
86.6667% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payu_latam.rb
+++ b/lib/active_merchant/billing/gateways/payu_latam.rb
@@ -52,6 +52,11 @@ module ActiveMerchant #:nodoc:
         add_transaction_elements(post, 'CAPTURE', options)
         add_reference(post, authorization)
 
+        if !amount.nil? && amount.to_f != 0.0
+          post[:transaction][:additionalValues] ||= {}
+          post[:transaction][:additionalValues][:TX_VALUE] = invoice_for(amount, options)[:TX_VALUE]
+        end
+
         commit('capture', post)
       end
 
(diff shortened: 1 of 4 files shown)


curiousepic:
Looks good, though maybe let's get confirmation from PayU that it won't blow up with the other additional_values in there (since indeed their sandbox won't cut it)

lib/active_merchant/billing/gateways/payu_latam.rb
```diff
@@ -52,6 +52,11 @@ def capture(amount, authorization, options={})
         add_transaction_elements(post, 'CAPTURE', options)
         add_reference(post, authorization)
 
+        if !amount.nil? && amount.to_f != 0.0
+          post[:transaction][:additionalValues] ||= {}
+          post[:transaction][:additionalValues][:TX_VALUE] = invoice_for(amount, options)[:TX_VALUE]
```
bpollack: @curiousepic What extra values? We only include just the `TX_VALUE` key here, which matches the Zendesk ticket.

curiousepic: I see now, kind of ignored that `[:TX_VALUE]` on the end there. They do also include currency in the example (despite saying it's ignored), but we can go with this for now and address any further issues as they come up.

bpollack: Yeah, there's some disagreement there; I'm having to take them at face value on their example, ’cause that's all I've got to go with.

PR #2976, opened 2018-08-31
Check if using SSL before getting SSL conn info.

mastermatt:
Was getting this error in a project when running its acceptance tests
against a mocking service that used a `http` schema.
```
undefined method `ssl_version' for #<TCPSocket:(closed)>
$GEMS/bundler/gems/active_merchant-f3d11caabc66/lib/active_merchant/net_http_ssl_connection.rb:7:in `ssl_connection'
$GEMS/bundler/gems/active_merchant-f3d11caabc66/lib/active_merchant/connection.rb:82:in `block (2 levels) in request'
```

### DIFF
--- a/lib/active_merchant/net_http_ssl_connection.rb
+++ b/lib/active_merchant/net_http_ssl_connection.rb
@@ -3,7 +3,7 @@ require 'net/http'
 module NetHttpSslConnection
   refine Net::HTTP do
     def ssl_connection
-      return {} unless @socket.present?
+      return {} unless use_ssl? && @socket.present?
       { version: @socket.io.ssl_version, cipher: @socket.io.cipher[0] }
     end
   end


bpollack:
Just to clarify, do you mean the mocking service is pure HTTP? If so, that's explicitly not supported by Active Merchant. Even TLS lower than 1.2 isn't sufficient (generally) for PCI compliance, so having support for HTTP endpoints is explicitly something we don't support.

bpollack:
(There are _tons_ of solutions here that cost nothing and are relatively easy, incidentally, ranging from using [Let's Encrypt](https://letsencrypt.org/) to [puma-dev](https://github.com/puma/puma-dev) to manually self-signed certs. The lack of HTTP support is not designed to prevent you from doing development locally.)

bpollack:
@bdewater Pinging you just in case you've got a different take here, since IIRC this is your code.

mastermatt:
yes, you are correct, the mocking service only offers HTTP and I don't control that service unfortunately. 
While I'm not surprised to hear this lib doesn't support HTTP in principal, there is code in place already to protect against it so I took a shot. The first line of `ActiveMerchant::Connection#configure_ssl` for instance.
https://github.com/activemerchant/active_merchant/blob/3169f0c1dd58cc3e195df0e6be7ed392c7cab256/lib/active_merchant/connection.rb#L149

bdewater:
While non-HTTPS production traffic is a big no-no according to the PCI DSS, the case for a HTTP mocking service seems perfectly valid. Certificates are yet another thing to configure and maintain, even with the things @bpollack mentioned. 

Also, I'd argue ActiveMerchant is mostly concerned with OSI layer 7 and not 6 🙂so I'm 👍on merging this.

bpollack:
Fair enough @bdewater ; I'll merge.

bpollack:
Merged as 00287a262d0607b206e0682a789dc368e3fbc2d8

PR #2978, opened 2018-08-31
Reflect correct test mode in Braintree responses

berkcaputcu:
WIP

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -352,7 +352,10 @@ module ActiveMerchant #:nodoc:
           message_from_result(result),
           { braintree_transaction: transaction_hash(result) },
           { authorization: (result.transaction.id if result.transaction) }
-        )
+            authorization: (result.transaction.id if result.transaction),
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -352,7 +352,10 @@ def response_from_result(result)
           message_from_result(result),
           { braintree_transaction: transaction_hash(result) },
           { authorization: (result.transaction.id if result.transaction) }
-        )
+            authorization: (result.transaction.id if result.transaction),
+            test: test?
+          }
+         )
```
elfassy: to fix the failing test:
```ruby
      def response_from_result(result)
        response = { braintree_transaction: transaction_hash(result) }
        Response.new(
          result.success?, 
          message_from_result(result),
          response,
          authorization: (result.transaction.id if result.transaction),
          test: test?
        )
      end
```

elfassy: moved to https://github.com/activemerchant/active_merchant/pull/2980 to be able to edit

Issue #1291, opened 2014-07-07
An error occurred during processing. Please try again


ghost:
I'm using spree 2.0 stable e-commerce platform with gem activemerchant varsion 1.42.9. I'm using Authorize.net CIM method for payment. I can purchase through credit card but facing issue at the time of refund payment. It is giving me respose message text as "An error occurred during processing. Please try again". Please provide me solution on this.
with Error code E00001

PR #2980, opened 2018-08-31
Reflect correct test mode in Braintree responses

elfassy:
# Why

Reflect correct test mode in Braintree responses. 

The request was using the correct environment for test transactions (sandbox) but the response object returned did not include which mode was used. 

# What

Set `test` option, as per template https://github.com/activemerchant/active_merchant/blob/85c57a6c2d15ae86ab3713120bf9d2277260cd26/generators/gateway/templates/gateway.rb#L92-L101

---
## Remote tests

```
Finished in 99.933189 seconds.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
63 tests, 357 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.2381% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
0.63 tests/s, 3.57 assertions/s
```

All three failing remote tests have `You must get credits enabled in your Sandbox account for this to pass` (not sure how to set that up, if anyone has a clue)

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -347,11 +347,14 @@ module ActiveMerchant #:nodoc:
       end
 
       def response_from_result(result)
+        response_hash = { braintree_transaction: transaction_hash(result) }
+
         Response.new(
           result.success?,
           message_from_result(result),
-          { braintree_transaction: transaction_hash(result) },
-          { authorization: (result.transaction.id if result.transaction) }
+          response_hash,
+          authorization: (result.transaction.id if result.transaction),
+          test: test?
         )
       end
 
@@ -369,6 +372,7 @@ module ActiveMerchant #:nodoc:
           options[:avs_result] = { code: avs_code_from(result.transaction) }
           options[:cvv_result] = result.transaction.cvv_response_code
         end
+        options[:test] = test?
         options
       end
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
 == HEAD
 * PayU Latam: support partial captures [bpollack] #2974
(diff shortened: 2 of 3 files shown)


bpollack:
I know y'all are at Shopify anyway, but it seems like this is a full duplicate of #2978

elfassy:
@bpollack correct, i cloned his branch to since he was working off a fork, so i can fix the tests

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -369,6 +372,7 @@ def response_options(result)
           options[:avs_result] = { code: avs_code_from(result.transaction) }
           options[:cvv_result] = result.transaction.cvv_response_code
         end
+        options[:test] = test?
```
elfassy: we may be able to use `response_options` in `response_from_result`, but I'll do the refactoring in another PR since i'm not sure of the implications of adding `avs_result` and `cvv_result`

pi3r:
Any way to add a remote test for it?

elfassy:
@filipebarcos I don't think it's a breaking change, we always expect a response object back, just now it will have the correct test value set. Good call on the changelog, will update.
@pi3r the current remote test cover this case. We're not changing how we communicate with Braintree, only the way we wrap the response.

dwlynch05:
Hi team, 
Double checking that the fix was shipped on the 6th September. I have a Plus Merchant JB Hifi who have incurred transaction fees due to the store processing test orders as false. From the period September to October 11th. For store uat-jbhifi.myshopify.com/. Last affected order was #1291. I cant tell if the test function was disabled as our logs do not go back that far. Reasonable to assume that the fix wasn't pushed to this site for whatever reason ?. Full transaction bill list. https://shopjbhifi.myshopify.com/admin/settings/account/invoice/56046734/fees?category=order_commission&shop_id=3032350789

PR #2983, opened 2018-09-07
Decide CC brand by lambda, not regex

bpollack:
Previously, we were using pure regexes to determine a credit card brand
based on its BIN. The problem is that you can't reliably determine
BINs based on regexes: Maestro overlaps with MasterCard, Electron just
has a pile of buckets, and so on.

As a first step, remove any reliance on raw regexes, and instead switch
to using lambdas. For now, this is functionally equivalent, but will
allow more precise brand determination in a follow-up commit.

Unit: 3927 tests, 68196 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -3,20 +3,20 @@ module ActiveMerchant #:nodoc:
     # Convenience methods that can be included into a custom Credit Card object, such as an ActiveRecord based Credit Card object.
     module CreditCardMethods
       CARD_COMPANIES = {
-        'visa'               => /^4\d{12}(\d{3})?(\d{3})?$/,
-        'master'             => /^(5[1-5]\d{4}|677189|222[1-9]\d{2}|22[3-9]\d{3}|2[3-6]\d{4}|27[01]\d{3}|2720\d{2})\d{10}$/,
-        'discover'           => /^(6011|65\d{2}|64[4-9]\d)\d{12}|(62\d{14})$/,
-        'american_express'   => /^3[47]\d{13}$/,
-        'diners_club'        => /^3(0[0-5]|[68]\d)\d{11}$/,
-        'jcb'                => /^35(28|29|[3-8]\d)\d{12}$/,
-        'switch'             => /^6759\d{12}(\d{2,3})?$/,
-        'solo'               => /^6767\d{12}(\d{2,3})?$/,
-        'dankort'            => /^5019\d{12}$/,
-        'maestro'            => /^(5[06-8]|6\d)\d{10,17}$/,
-        'forbrugsforeningen' => /^600722\d{10}$/,
-        'laser'              => /^(6304|6706|6709|6771(?!89))\d{8}(\d{4}|\d{6,7})?$/,
-        'sodexo'             => /^(606071|603389|606070|606069|606068|600818)\d{8}$/,
-        'vr'                 => /^(627416|637036)\d{8}$/
+        'visa'               => ->(num) { num =~ /^4\d{12}(\d{3})?(\d{3})?$/ },
+        'master'             => ->(num) { num =~ /^(5[1-5]\d{4}|677189|222[1-9]\d{2}|22[3-9]\d{3}|2[3-6]\d{4}|27[01]\d{3}|2720\d{2})\d{10}$/ },
+        'discover'           => ->(num) { num =~ /^(6011|65\d{2}|64[4-9]\d)\d{12}|(62\d{14})$/ },
+        'american_express'   => ->(num) { num =~ /^3[47]\d{13}$/ },
+        'diners_club'        => ->(num) { num =~ /^3(0[0-5]|[68]\d)\d{11}$/ },
+        'jcb'                => ->(num) { num =~ /^35(28|29|[3-8]\d)\d{12}$/ },
+        'switch'             => ->(num) { num =~ /^6759\d{12}(\d{2,3})?$/ },
+        'solo'               => ->(num) { num =~ /^6767\d{12}(\d{2,3})?$/ },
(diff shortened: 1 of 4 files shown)


bpollack:
Recalling this from review for the moment; I may approach this differently.

bpollack:
All right, adding back. I'm actually happy with the direction this is going, and I'd like feedback on how others feel about it before going any further.

bpollack:
Also, apologies for Laser being part of this. I can strip it out before this whole PR series, but Maestro tests do (and should!) fail if Laser's left in, because it's in Maestro's current BIN range.

curiousepic:
While you're in here, might be nice to make the hash name `CARD_COMPANIES` more descriptive, perhaps `CARD_COMPANY_BIN_RANGES`.

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -103,37 +128,18 @@ def valid_number?(number)
...
-        #     card_companies.find([nil]) { |brand, regexp| number =~ regexp }.first.dup
-        #   end
-        #
         def brand?(number)
           return 'bogus' if valid_test_mode_card_number?(number)
 
-          card_companies.reject { |c,p| c == 'maestro' }.each do |company, pattern|
-            return company.dup if number =~ pattern
+          CARD_COMPANIES.each do |company, func|
+            return company.dup if func[number]
```
dtykocki: I'd normally expect to see a lambda invoked via `.call`, so `func.call(number)`. Is there a benefit to invoking the function as `func[number]`?

bpollack: Nope, just me not knowing what's Rubyish.

bpollack: Done.

bpollack:
> While you're in here, might be nice to make the hash name `CARD_COMPANIES` more descriptive, perhaps `CARD_COMPANY_BIN_RANGES`.

Sure. I'll hold off until @bdewater or @joshnuss or someone from Shopify weighs in so I don't have to keep doing odd rebases, though.

bpollack:
@bdewater Any thoughts on this?

bpollack:
I'm going to merge this today; it's been open over a week.

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -39,10 +38,29 @@ module CreditCardMethods
...
+      ]
+
       def self.included(base)
         base.extend(ClassMethods)
       end
 
+      def self.in_bin_range?(number, ranges)
+        bin = number.to_i
+        ranges.any? do |range|
+          range.include?(bin)
```
bdewater: We'll probably want to use `cover` instead of `include`, it's [faster](https://github.com/JuanitoFatas/fast-ruby/#range) since it doesn't traverse the entire range 🙂

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -103,37 +128,18 @@ def valid_number?(number)
...
-        #     card_companies.find([nil]) { |brand, regexp| number =~ regexp }.first.dup
-        #   end
-        #
         def brand?(number)
           return 'bogus' if valid_test_mode_card_number?(number)
 
-          card_companies.reject { |c,p| c == 'maestro' }.each do |company, pattern|
-            return company.dup if number =~ pattern
+          CARD_COMPANIES.each do |company, func|
+            return company.dup if func.call(number)
```
bdewater: We can get rid of the repeated `to_s` calls in the `CARD_COMPANIES` hash by just coercing it here. Although I wonder if that's even needed - are there cases where this is called with anything other than a string?

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -103,37 +128,18 @@ def valid_number?(number)
...
         end
 
-        # Regular expressions for the known card companies.
-        #
-        # References:
-        # - http://en.wikipedia.org/wiki/Credit_card_number
-        # - http://www.barclaycardbusiness.co.uk/information_zone/processing/bin_rules.html
         def card_companies
-          CARD_COMPANIES
+          CARD_COMPANIES.keys
```
pi3r: Isn't that a breaking change? Someone might be using this method.

lib/active_merchant/billing/credit_card_methods.rb
bpollack: Oh, point. Will do!

lib/active_merchant/billing/credit_card_methods.rb
bpollack: It's possible. If we're concerned, I can tweak this by adding a parallel structure that returns the regexes. But the regexes will be (honestly already are) wrong, so I'd personally prefer we just loudly note it's a breaking change in the release notes.

lib/active_merchant/billing/credit_card_methods.rb
bpollack: I'm not sure, honestly; this was just me coding very defensively. If you're feeling good on the Shopify side, then I'm okay switching this to assume strings, since 1) it should be anyway, and 2) we're already going to be breaking backcompat here to some degree.

lib/active_merchant/billing/credit_card_methods.rb
pi3r: That works for me!

bpollack:
@bdewater Implemented your changes, except the `to_s` changes; I agree we should do that, but it gets more involved _really_ quickly (some methods that call `.to_s` appear as if at least we may call them directly, so I'll need to audit that on our side), and I know I've got another patch or two this direction anyway, so I'd rather delay that for a separate PR if that's all right.

bdewater:
Sounds good to me!

PR #2986, opened 2018-09-12
Rubocop changes that effect the indention of comments

nfarve:


bpollack:
@nfarve Test failure was bogus; I kicked it and it's green now.

PR #2990, opened 2018-09-17
Begin removing Solo and Switch

bpollack:
Both [Switch][1] and [Solo][2] were purchased by Maestro, and haven't been issued since 2002 and 2011, respectively. That's good, because their old bin ranges overlap with Maestro's current ones (surprise).

We should fully remove Solo and Switch, but some oddities about them (such as having start dates) has special handling in AM. Some gateways also run remote tests with old Switch/Solo cards. As a result, fully removing all Solo and Switch support *immediately* is counterproductive. This is the minimal amount of change required to properly handle modern Maestro cards.

[Switch]: https://en.wikipedia.org/wiki/Switch_(debit_card)
[Solo]: https://en.wikipedia.org/wiki/Solo_(debit_card)

3922 tests, 68171 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -9,8 +9,6 @@ module ActiveMerchant #:nodoc:
         'american_express'   => ->(num) { num =~ /^3[47]\d{13}$/ },
         'diners_club'        => ->(num) { num =~ /^3(0[0-5]|[68]\d)\d{11}$/ },
         'jcb'                => ->(num) { num =~ /^35(28|29|[3-8]\d)\d{12}$/ },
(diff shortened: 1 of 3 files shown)


bpollack:
Sorry @bdewater ; I'll quit tagging you on this series shortly, but I want to keep at least someone from Shopify in the loop as I clean this up so I don't make things unintentionally exciting.

bpollack:
Also, if you're curious, e2a2a6a9f8a4a5f919b34835272623c9499f66e0 is what the more aggressive version of this would add. I'm happy to fold that in here, but it's definitely riskier.

PR #2991, opened 2018-09-18
Remove the rest of Solo and Sripe support

bpollack:
This is the continuation of #2990, and finishes removing Solo and Switch fully from the code base. It was kept separate because it's riskier (in particular, some of the remote tests were using Switch and Solo cards, and I have a hunch this breaks them--though they haven't been touched in forever and I can't imagine how they were working previously, anyway, given the cards have been sunset for *years*), but I still believe it's safe to land.


bpollack:
I promise this is the last one I'll ping Shopify on for a few days, but this one still felt like one last one you all might want input on. :-)

PR #2992, opened 2018-09-18
Add support for Carnet

bpollack:
Carnet is only used in Mexico, and the only test number I could find I found by googling for pictures of Carnet cards until I found someone with one who didn't blur out the number (although the test number here was mutated to be in the same bin range but have different subsequent digits).

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -13,7 +13,13 @@ module ActiveMerchant #:nodoc:
         'maestro'            => ->(num) { (12..19).include?(num&.size) && in_bin_range?(num.slice(0, 6), MAESTRO_RANGES) },
         'forbrugsforeningen' => ->(num) { num =~ /^600722\d{10}$/ },
         'sodexo'             => ->(num) { num =~ /^(606071|603389|606070|606069|606068|600818)\d{8}$/ },
-        'vr'                 => ->(num) { num =~ /^(627416|637036)\d{8}$/ }
+        'vr'                 => ->(num) { num =~ /^(627416|637036)\d{8}$/ },
+        'carnet'             => lambda { |num|
+          num&.size == 16 && (
+            in_bin_range?(num.slice(0, 6), CARNET_RANGES) ||
+            CARNET_BINS.any? { |bin| num.slice(0, bin.size) == bin }
+          )
+        }
       }
 
       # http://www.barclaycard.co.uk/business/files/bin_rules.pdf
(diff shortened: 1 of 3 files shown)


test/unit/credit_card_methods_test.rb
```diff
@@ -187,6 +187,16 @@ def test_19_digit_maestro_uk
     assert_equal 'maestro', CreditCard.brand?(number)
   end
 
+  def test_carnet_cards
+    numbers = [
+      '5062280000000000'
```
curiousepic: Are there any further dummy numbers you could throw in here for exercising the recognition?

bpollack: Not _yet_; there aren't even test numbers as such. I have two live card numbers, and that's it. I'm open to ideas here, though; I obviously was planning to have more card numbers than I ended up having...

bpollack:
The review this depended on got approved and merged; I updated the description appropriately.

test/unit/credit_card_methods_test.rb
curiousepic: Well I'm mostly thinking that it doesn't really matter if they're test cards or not as long as they qualify for the matching criteria. Right?

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -13,7 +13,13 @@ module CreditCardMethods
         'maestro'            => ->(num) { (12..19).include?(num&.size) && in_bin_range?(num.slice(0, 6), MAESTRO_RANGES) },
         'forbrugsforeningen' => ->(num) { num =~ /^600722\d{10}$/ },
         'sodexo'             => ->(num) { num =~ /^(606071|603389|606070|606069|606068|600818)\d{8}$/ },
-        'vr'                 => ->(num) { num =~ /^(627416|637036)\d{8}$/ }
+        'vr'                 => ->(num) { num =~ /^(627416|637036)\d{8}$/ },
+        'carnet'             => lambda { |num|
```
dtykocki: I take it you're not going with the stabby lambda `->(num)` here for readability sake?

bpollack: It's actually a RuboCop violation to use `->` with a multi-line lambda. Was news to me, too.

test/unit/credit_card_methods_test.rb
bpollack: I'll add a few more test cards to make sure we're flexing at least _one_ of the card buckets and at least a couple splotches of the BIN range.

PR #2993, opened 2018-09-18
Piles of RuboCop fixes

bpollack:
I was trying to stick really hard to ones that improved readability, but I think these are overwhlemingly wins.


bpollack:
> 👍 First time encountering that SymbolProc thing, it seems less readable and un-ruby-like, but if it's a standard RuboCop offense I'm not bothered

I am okay to back that one out, but it was added in Ruby 2.2 and is actually quite common; it's used nearly 300 times in Rails, 70 times in Sequel, etc.

PR #2996, opened 2018-09-20
Stripe: handle much more complicated post structures

bpollack:
Previously, we couldn't handle Stripe endpoints that had arrays ofhashes, among other things. This fixes that. In the process, this alsoswitches from implicit to explicit array indices, which allowssupporting an upcoming feature.

Unit: 129 tests, 682 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 65 tests, 298 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


bpollack:
NB that unit tests had to be changed, but all remote tests worked exactly as-is, which is expected.

bpollack:
I finished up the level 3 stuff (which the first commit was designed to support), so I folded that in here, too.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -491,25 +491,39 @@ def parse(body)
...
+        params.each do |key, value|
           next if value != false && value.blank?
+          flattened_key = prefix.nil? ? key : "#{prefix}[#{key}]"
           if value.is_a?(Hash)
-            h = {}
-            value.each do |k, v|
-              h["#{key}[#{k}]"] = v unless v.blank?
-            end
-            post_data(h)
+            flatten_params(flattened, value, flattened_key)
```
dtykocki: Aw yeaa.. recursion

PR #2999, opened 2018-09-24
Enable Carnet for Conekta and Openpay

bpollack:
Unit: 3927 tests, 68207 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed


bpollack:
Hold; expanding the list of updated gateways.

bpollack:
Okay, adding y'all back; after further convos, we're just doing these two initially.

PR #3001, opened 2018-09-27
Authorize.Net: Support custom delimiter for cim

curiousepic:
By default, cim transaction responses have a field delimiter of ',' but
fields themselves can easily contain unescaped commas, causing bad
parsing. Providing a custom delimiter option can prevent this.

Remote:
68 tests, 234 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
93 tests, 536 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -101,7 +101,7 @@ module ActiveMerchant
 
       def purchase(amount, payment, options = {})
         if payment.is_a?(String)
-          commit(:cim_purchase) do |xml|
+          commit(:cim_purchase, options) do |xml|
             add_cim_auth_purchase(xml, 'profileTransAuthCapture', amount, payment, options)
           end
         else
@@ -113,7 +113,7 @@ module ActiveMerchant
 
       def authorize(amount, payment, options={})
         if payment.is_a?(String)
-          commit(:cim_authorize) do |xml|
+          commit(:cim_authorize, options) do |xml|
             add_cim_auth_purchase(xml, 'profileTransAuthOnly', amount, payment, options)
           end
         else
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -672,8 +676,12 @@ def add_po_number(xml, options)
         xml.poNumber(options[:po_number]) if options[:po_number]
       end
 
+      def add_extra_options_for_cim(xml, options)
+        xml.extraOptions("x_delim_char=#{options[:delimiter]}") if options[:delimiter]
```
bpollack: Is this always legal to add? I'm asking because, if it is, we could fold it directly into `commit`. (I'm aware that it's not currently added for a lot of the customer profile calls, hence the question.)

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -967,11 +975,12 @@ def auth_was_for_cim?(authorization)
...
       end
 
-      def parse_direct_response_elements(response)
+      def parse_direct_response_elements(response, options)
         params = response[:direct_response]
+        delimiter = options[:delimiter] if options[:delimiter]
         return {} unless params
 
-        parts = params.split(',')
+        parts = params.split(delimiter || ',')
```
bpollack: Because you're ending with `|| ','` here, you can just axe line 980 entirely and turn this into `parts = params.split(options[:delimiter] || ',')`

curiousepic: That's what I thought initially but second-guessed myself, restoring.

lib/active_merchant/billing/gateways/authorize_net.rb
curiousepic: I'll check.

curiousepic: Unfortunately no, this is probably safer.

bpollack: No worries! Thanks for checking.

bpollack:
@curiousepic What's going on with this? Should it still be merged?

curiousepic:
@bpollack It was merged somehow, closing.

PR #3004, opened 2018-10-02
RuboCop: AlignHash

nfarve:
Rubocop fix to align all hashes.

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay_online_payments.rb
+++ b/lib/active_merchant/billing/gateways/worldpay_online_payments.rb
@@ -92,21 +92,21 @@ module ActiveMerchant #:nodoc:
       end
 
       def create_post_for_auth_or_purchase(token, money, options)
-      {
-        'token' => token,
-        'orderDescription' => options[:description] || 'Worldpay Order',
-        'amount' => money,
-        'currencyCode' => options[:currency] || default_currency,
-        'name' => options[:billing_address]&&options[:billing_address][:name] ? options[:billing_address][:name] : '',
-        'billingAddress' => {
-          'address1'=>options[:billing_address]&&options[:billing_address][:address1] ? options[:billing_address][:address1] : '',
-          'address2'=>options[:billing_address]&&options[:billing_address][:address2] ? options[:billing_address][:address2] : '',
-          'address3'=>'',
-          'postalCode'=>options[:billing_address]&&options[:billing_address][:zip] ? options[:billing_address][:zip] : '',
-          'city'=>options[:billing_address]&&options[:billing_address][:city] ? options[:billing_address][:city] : '',
-          'state'=>options[:billing_address]&&options[:billing_address][:state] ? options[:billing_address][:state] : '',
-          'countryCode'=>options[:billing_address]&&options[:billing_address][:country] ? options[:billing_address][:country] : ''
-          },
+        {
+          'token' => token,
+          'orderDescription' => options[:description] || 'Worldpay Order',
+          'amount' => money,
+          'currencyCode' => options[:currency] || default_currency,
+          'name' => options[:billing_address]&&options[:billing_address][:name] ? options[:billing_address][:name] : '',
+          'billingAddress' => {
+            'address1'=>options[:billing_address]&&options[:billing_address][:address1] ? options[:billing_address][:address1] : '',
+            'address2'=>options[:billing_address]&&options[:billing_address][:address2] ? options[:billing_address][:address2] : '',
+            'address3'=>'',
+            'postalCode'=>options[:billing_address]&&options[:billing_address][:zip] ? options[:billing_address][:zip] : '',
+            'city'=>options[:billing_address]&&options[:billing_address][:city] ? options[:billing_address][:city] : '',
+            'state'=>options[:billing_address]&&options[:billing_address][:state] ? options[:billing_address][:state] : '',
+            'countryCode'=>options[:billing_address]&&options[:billing_address][:country] ? options[:billing_address][:country] : ''
+           },
           'customerOrderCode' => options[:order_id],
           'orderType' => 'ECOM',
           'authorizeOnly' => options[:authorizeOnly] ? true : false
--- a/test/remote/gateways/remote_opp_test.rb
+++ b/test/remote/gateways/remote_opp_test.rb
@@ -18,37 +18,37 @@ class RemoteOppTest < Test::Unit::TestCase
       merchant_transaction_id: "active_merchant_test_complete #{time}",
       address: address,
       description: 'Store Purchase - Books',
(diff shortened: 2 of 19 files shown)


test/remote/gateways/remote_opp_test.rb
```diff
@@ -18,25 +18,25 @@ def setup
       merchant_transaction_id: "active_merchant_test_complete #{time}",
       address: address,
       description: 'Store Purchase - Books',
-        #      riskWorkflow: true,
-        #      testMode: 'EXTERNAL' # or 'INTERNAL', valid only for test system
+      #      riskWorkflow: true,
+      #      testMode: 'EXTERNAL' # or 'INTERNAL', valid only for test system
 
-        billing_address: {
+      billing_address: {
```
bpollack: Here, too.

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
```diff
@@ -107,9 +107,9 @@ def create_post_for_auth_or_purchase(token, money, options)
           'state'=>options[:billing_address]&&options[:billing_address][:state] ? options[:billing_address][:state] : '',
           'countryCode'=>options[:billing_address]&&options[:billing_address][:country] ? options[:billing_address][:country] : ''
           },
-          'customerOrderCode' => options[:order_id],
-          'orderType' => 'ECOM',
-          'authorizeOnly' => options[:authorizeOnly] ? true : false
+        'customerOrderCode' => options[:order_id],
+        'orderType' => 'ECOM',
+        'authorizeOnly' => options[:authorizeOnly] ? true : false
```
bpollack: The indentation pattern going on here seems weird/wrong; seems like 109 should be indented less, or these lines more, or something. There's also an awkward/insufficient amount of indentation here v. the closing brace on 113. Do you know what's going on here?

test/remote/gateways/remote_opp_test.rb
```diff
@@ -18,25 +18,25 @@ def setup
       merchant_transaction_id: "active_merchant_test_complete #{time}",
       address: address,
       description: 'Store Purchase - Books',
-        #      riskWorkflow: true,
-        #      testMode: 'EXTERNAL' # or 'INTERNAL', valid only for test system
+      #      riskWorkflow: true,
```
bpollack: Might make sense to shorten the gap between `#` and the keyword in this comment, too. I have a hunch it at one point lined up with the previous lines, but that clearly stopped being the case even before your patch.

test/remote/gateways/remote_payflow_express_test.rb
```diff
@@ -51,65 +51,65 @@ def test_set_express_purchase
...
-      :subtotal=>2798,
-      :items => [
-        {:name => 'test4',
-          :description => 'test4',
-          :quantity=>2 ,
-          :amount=> 1399 ,
-          :url=>'http://localhost:3000/products/test4'}],
-      :discount=>280,
-      :no_shipping=>true}
+               :return_url=>'http://localhost:3000/orders/1/8e06ea26f8add7608671d433f13c2193/commit_paypal?buyer_accepts_marketing=true&utm_nooverride=1',
```
bpollack: If we're going with this indentation style, it'd probably make sense to make sure `{`s go on their own line as part of the patch; the following lines (and those for the other hashes in this file) end up being _really_ indented when they wouldn't otherwise need to be.

test/remote/gateways/remote_paypal_express_test.rb
```diff
@@ -10,11 +10,11 @@ def setup
       :order_id => '230000',
       :email => 'buyer@jadedpallet.com',
       :billing_address => { :name => 'Fred Brooks',
-                    :address1 => '1234 Penny Lane',
-                    :city => 'Jonsetown',
-                    :state => 'NC',
-                    :country => 'US',
-                    :zip => '23456'
+                            :address1 => '1234 Penny Lane',
```
bpollack: Here, too.

test/unit/gateways/eway_managed_test.rb
```diff
@@ -25,11 +25,11 @@ def setup
                   :title => 'Mr.',
                   :phone => '(555)555-5555'
                },
-               :email => 'someguy1232@fakeemail.net',
-               :order_id => '1000',
-               :customer => 'mycustomerref',
-               :description => 'My Description',
-               :invoice => 'invoice-4567'
+                 :email => 'someguy1232@fakeemail.net',
```
bpollack: I'm confused again what's going on here. Seems like these should align with the `}` on 27?

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
nfarve: These lines are aligned with line 101

test/unit/gateways/eway_managed_test.rb
nfarve: This rubocop is trying to align the `:`'s with each other. Not he to the `{`

test/remote/gateways/remote_opp_test.rb
nfarve: Here `billing_address` is aligned with `description` on line 20.

nfarve:
@bpollack cleaned it up a bit. Hopefully more readable and less indented.

lib/active_merchant/billing/gateways/worldpay_online_payments.rb
deedeelavinder: @nfarve It looks like the closing brackets on 109 and 113 need to go left two spaces so that they align with the beginning characters of their block or section. That would follow the pattern, e.g. lines 121 & 127 and 202 & 206.

bpollack: Yeah, @deedeelavinder 's got it. In addition, the alignment issues here are probably cause because the hash is starting out underindented (which isn't new with your patch); line 95 and following should really be indented two spaces further, I suspect, and the rest of this is an alignment fight as the editor's trying to make everything work out.

test/remote/gateways/remote_opp_test.rb
bpollack: Yes, but the `}` is aligned neither with the hash key nor with the opening brace, which are the two styles honored by RuboCop. There are a couple ways to address this, but probably the best would be to unindent the entire collection of hashes so that the closing braces align with the hash keys (which looks like a three-space unindent, fi I'm reading this properly, for lines 25–30, 32–38, and 40-whenever that block ends, inclusive)

bpollack:
This looks way better @nfarve! I do think it'd make sense to fix up the entire hash blocks in a couple places, since this still causes some indentation regressions for two sections, but this is almost ready to go.

PR #3005, opened 2018-10-02
Global Collect: handle internal server errors

molbrown:
Ingenico returns HTML when it has a 500 error, which this gateway has
not handled and would crash on a JSON parse error. New rescue provides a
more informative error and prevents crashing in this situation.

Remote:
16 tests, 38 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
19 tests, 85 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -72,8 +72,8 @@ module ActiveMerchant #:nodoc:
       def scrub(transcript)
         transcript.
           gsub(%r((Authorization: )[^\\]*)i, '\1[FILTERED]').
-          gsub(%r(("cardNumber\\":\\")\d+), '\1[FILTERED]').
-          gsub(%r(("cvv\\":\\")\d+), '\1[FILTERED]')
+          gsub(%r(("cardNumber\\+":\\+")\d+), '\1[FILTERED]').
+          gsub(%r(("cvv\\+":\\+")\d+), '\1[FILTERED]')
       end
 
       private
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -248,6 +252,14 @@ def commit(action, post, authorization = nil)
         )
       end
 
+      def json_error(raw_response)
+        {
+          'error_message' => 'Invalid response received from the Global Collect API.  Please contact SOMEONE if you continue to receive this message.' +
```
curiousepic: SOMEONE 😂

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -248,6 +252,14 @@ def commit(action, post, authorization = nil)
...
+            "  (The raw response returned by the API was #{raw_response.inspect})",
```
curiousepic: I'm trying to figure out if it's possible that the contents here would reach the customer un-scrubbed... 🤔

molbrown: Just to clarify, you are saying the contents _should_ be scrubbed and it might not be? Or the other way around?

nfarve: @curiousepic is the complete response gets scrubbed I don't see why having `raw_response.inspect` would cause an issue.

molbrown: @curiousepic @bpollack @nfarve New test to confirm scrubbing in the `message` revealed it may not have been scrubbed, though I had to get creative with the error response I tested with. Updated the `scrub` method to cover bases, in any case.

molbrown: Also, I modeled this solution after other gateways that also use `raw_response.inspect` in error messages, and those do not test errors for scrubbing either. May be worth checking for those other gateways.

curiousepic:
Looks like Rubocop is failing due to line length of the scrubbed response mock, though. You can add the file to the exceptions in `.rubocop_todo.yml`

PR #3010, opened 2018-10-11
Regression: Account for nil CC number

guaguasi:
Fixes exceptions created by 2bc7493 when CC number is nil

```
3933 tests, 68237 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed
```

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -160,7 +160,7 @@ module ActiveMerchant #:nodoc:
         end
 
         def electron?(number)
-          return false unless [16, 19].include?(number.length)
+          return false unless [16, 19].include?(number&.length)
 
           # don't recalculate for each range
           bank_identification_number = first_digits(number).to_i
@@ -176,7 +176,7 @@ module ActiveMerchant #:nodoc:
         end
 
         def first_digits(number)
-          number.slice(0,6)
+          number&.slice(0,6) || ''
         end
 
         def last_digits(number)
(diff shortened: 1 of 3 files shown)


bpollack:
Thanks for the PR. Before I look at merging, it'd be helpful if you could let me know how `nil`s are getting into these functions. Are these calls happening "within" Active Merchant, or are you calling these functions with `nil`s from your code?

guaguasi:
I suppose the `nil`s would be considered as coming from outside ActiveMerchant. If the `ActiveMerchant::Billing::CreditCard` object has a `nil` `number`, the various methods such as `#first_digits` fail with exception in contrast to how it used to behave and return an empty string. 

If you notice in the originally referenced commit, there are various locations that once had `#to_s` and now have the SNO. This PR merely accounts for the `nil` without doing string conversion while also maintaining the original return values.

The other item of note, is that the now changed `#last_digits` matches the return value of `#first_digits` when `number` is `nil`. Prior to 2bc7493, `#last_digits` returned `nil` whereas it now returns an empty string. By the same token, prior to that commit, `#first_digits` returned an empty string and now it just throws an exception of `undefined method `slice' for nil:NilClass`. After this PR, they both return an empty string when `number` is `nil`.

bpollack:
Thanks @guaguasi ; I'll take a look tomorrow and merge if things look good.

bpollack:
Bah, sorry for not circling back around @guaguasi ; this is on my todo list for today.

bpollack:
Merged. Thanks @guaguasi !

PR #3012, opened 2018-10-12
Braintree: Adds `device_data` as top-level parameter. 

deedeelavinder:
According to Braintree docs, `device_data` can be sent as a top-level
parameter when creating a `transaction`, `payment_method`, or
`customer`. And after chatting with their support, there is no way to
test this from the server-side. Deploying this with the hope that a
customer will give us some feedback about its efficacy.

Unit tests:
54 tests, 137 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote tests:
63 tests, 361 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -159,7 +159,7 @@ module ActiveMerchant #:nodoc:
             :last_name => creditcard.last_name,
             :email => scrub_email(options[:email]),
             :phone => options[:phone] || (options[:billing_address][:phone] if options[:billing_address] &&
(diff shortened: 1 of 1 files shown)


deedeelavinder:
@dtykocki I tried lots of variations on remote tests, but the `risk_data` parameter always showed `device_data_captured` as `false` and the value passed in `device_data` was never returned. Since there was no way to verify `device_data` had been received, or even passed correctly, I couldn't picture how to write helpful unit tests. I was thinking remote/unit tests could be added iteratively as we heard back from the customer, perhaps with some examples. Open to suggestions!

PR #3014, opened 2018-10-14
Allow inquiring settlement status for card connect

tywhang:
Allow transactions to be inquired about their settlement status.

Implements this API endpoint: https://developer.cardconnect.com/cardconnect-api#inquire


bpollack:
Thanks for this suggestion, @tywhang. This unfortunately falls outside the scope of what Active Merchant does; the goal of this library is to provide a common payment surface across all the gateways, rather than to provide gateway-specific features like this. It would probably make more sense either to add this functionality entirely separately in your application, or to use a [refinement](https://ruby-doc.org/core-2.5.1/doc/syntax/refinements_rdoc.html) in your own code base to add the functionality to this class just for your use case.

PR #3018, opened 2018-10-22
Rubocop: Layout/CaseIndention and Layout/ElseAlignment

nfarve:
Fixes some issues with indention and alignment with cases and elses.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,3 +29,6 @@ Layout/AlignParameters:
 
 Layout/DotPosition:
   EnforcedStyle: trailing
+
+Layout/CaseIndentation:
+  EnforcedStyle: end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -14,35 +14,10 @@ Gemspec/OrderedDependencies:
   Exclude:
     - 'activemerchant.gemspec'
 
-# Offense count: 113
-# Cop supports --auto-correct.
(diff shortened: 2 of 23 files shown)


bpollack:
Hmm, looks like we [_can_](https://rubocop.readthedocs.io/en/latest/cops_layout/#layoutcaseindentation); that's done by setting `IndentOneStep` to `true`. How do y'all feel about that, though? Having `when` flush with `case` is the default RuboCop style, but IMVHO it makes them harder to follow.

curiousepic:
Honestly the indented `if`s are more annoying to me... not a strong opinion though.

bpollack:
> Honestly the super-indented ifs are more annoying to me... 

This is why I haven't tried to land `Layout/ElseAlignment` in AM; we do this pattern so often there that it actually then overloads the line length limits. But I don't know an equivalent of `EnforceStyle: end` for `else` alignment.

nfarve:
@bpollack there is no equivalent for `Layout/ElseAlignment`. We would just have to agree to ignore it all together.

nfarve:
Also not sure I am on board for the indent on the `when`. I see your point that its a happy medium between the indent with the `=` or with the end.

bpollack:
@nfarve Do you still want to merge this as-is? It's drifted enough from `master` I'm getting worried about it merging cleanly.

PR #3023, opened 2018-10-25
RuboCop and logic fixes exposed by resolving RuboCop issues

bpollack:
This patch series is beginning to fix actual bugs in AM exposed by tightening the RuboCop settings. Please give it more than the cursory scan. I can split this into multiple PRs if it's too long.

### DIFF
--- a/lib/active_merchant/billing/gateways/borgun.rb
+++ b/lib/active_merchant/billing/gateways/borgun.rb
@@ -130,7 +130,6 @@ module ActiveMerchant #:nodoc:
         post[:Processor] = @options[:processor]
         post[:MerchantID] = @options[:merchant_id]
 
-        url = (test? ? test_url : live_url)
         request = build_request(action, post)
         raw = ssl_post(url(action), request, headers)
         pairs = parse(raw)
--- a/lib/active_merchant/billing/gateways/usa_epay_advanced.rb
+++ b/lib/active_merchant/billing/gateways/usa_epay_advanced.rb
@@ -1386,7 +1386,8 @@ module ActiveMerchant #:nodoc:
             build_tag soap, v[0], v[1], options[k]
           end
           case
-          when options[:payment_method] == nil
+          when options[:payment_method].nil?
+            nil
           when options[:payment_method].kind_of?(ActiveMerchant::Billing::CreditCard)
             build_credit_card_data soap, options
           when options[:payment_method].kind_of?(ActiveMerchant::Billing::Check)
--- a/test/remote/gateways/remote_beanstream_test.rb
+++ b/test/remote/gateways/remote_beanstream_test.rb
@@ -267,7 +267,7 @@ class RemoteBeanstreamTest < Test::Unit::TestCase
     assert_success purchase
 
     assert refund = @gateway.refund(@amount, purchase.authorization)
-    assert_success credit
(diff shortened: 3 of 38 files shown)


lib/active_merchant/billing/gateways/borgun.rb
```diff
@@ -130,7 +130,6 @@ def commit(action, post)
         post[:Processor] = @options[:processor]
         post[:MerchantID] = @options[:merchant_id]
 
-        url = (test? ? test_url : live_url)
```
bpollack: This one's fun. You'd think this should fail, given that the `url` _method_ is called two lines later, but somehow this works:

```ruby
irb(main):002:0> class Foo
irb(main):003:1>   def foo
irb(main):004:2>     u = 'hello'
irb(main):005:2>     u('goodbye')
irb(main):006:2>   end
irb(main):007:1>   def u(greeting)
irb(main):008:2>     puts greeting
irb(main):009:2>   end
irb(main):010:1> end
=> :u
irb(main):011:0> Foo.new.foo
goodbye
```

I'd have to look at the compiler to figure out what's going on here, because I have no idea.

test/remote/gateways/remote_beanstream_test.rb
```diff
@@ -267,7 +267,7 @@ def test_successful_check_purchase_and_refund
     assert_success purchase
 
     assert refund = @gateway.refund(@amount, purchase.authorization)
-    assert_success credit
+    assert_success refund
```
bpollack: I'm actually guessing here. `credit` isn't defined, so the author probably meant `refund`, but that implies this test was never run, so…not 100% sure.

test/remote/gateways/remote_visanet_peru_test.rb
```diff
@@ -107,7 +107,7 @@ def test_successful_refund_unsettled
     assert_success response
 
     new_auth = "_|#{response.authorization.split('|')[1]}"
-    refund = @gateway.refund(@amount, new_auth, @options.merge(force_full_refund_if_unsettled: true, ruc: '20341198217'))
+    @gateway.refund(@amount, new_auth, @options.merge(force_full_refund_if_unsettled: true, ruc: '20341198217'))
```
bpollack: This should probably be `_refund`, or just comment out the entire method. I'm favoring the latter.

lib/active_merchant/billing/gateways/usa_epay_advanced.rb
```diff
@@ -1386,8 +1386,7 @@ def build_transaction_request_object(soap, options, name='Params')
             build_tag soap, v[0], v[1], options[k]
           end
           case
-          when options[:payment_method] == nil
-          when options[:payment_method].kind_of?(ActiveMerchant::Billing::CreditCard)
+          when options[:payment_method].nil? || options[:payment_method].kind_of?(ActiveMerchant::Billing::CreditCard)
```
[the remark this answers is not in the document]
bpollack: Brain fart; I was in C mode, where `when`s fall through. I think line 1389 should instead be deleted entirely.

bpollack:
@curiousepic Take a look at the redo and let me know if this looks better.

curiousepic:
@bpollack Seems fine, still approved.

PR #3024, opened 2018-10-25
Add state to the netbanx payload

Girardvjonathan:
This field is needed for certain transaction that could get declined
because of the gateway "Risk Management department"

I manually tested that the field get send, but didn't do a test transaction (I had no test credential).
This gateway tests were passing before with a state defined [code is here](https://github.com/activemerchant/active_merchant/blob/master/test/test_helper.rb#L231)
and are still passing after with it defined.

Please tell me if more tests are needed
Cheers!


pi3r:
Can you update the unit tests here https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/netbanx_test.rb?

PR #3025, opened 2018-10-30
Braintree: Fix passing phone-only billing address

curiousepic:
If a billing address option had only a phone number, the billing address
element was still included in the request with all empty fields, which
causes a failure. Now we remove the phone from the billing address hash
before it's added.

Remote:
100% passed

Unit:
55 tests, 139 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -304,12 +304,13 @@ module ActiveMerchant #:nodoc:
 
         parameters[:credit_card] ||= {}
         parameters[:credit_card].merge!(:options => valid_options)
-        parameters[:credit_card][:billing_address] = map_address(options[:billing_address]) if options[:billing_address]
+        address = options[:billing_address]&.except(:phone)
+        return parameters if address.nil? || address.empty?
+        parameters[:credit_card][:billing_address] = map_address(address)
         parameters
       end
 
       def map_address(address)
-        return {} if address.nil?
         mapped = {
           :street_address => address[:address1],
           :extended_address => address[:address2],
--- a/test/remote/gateways/remote_braintree_blue_test.rb
+++ b/test/remote/gateways/remote_braintree_blue_test.rb
@@ -168,6 +168,20 @@ class RemoteBraintreeBlueTest < Test::Unit::TestCase
     assert_equal purchase_response.params['braintree_transaction']['billing_details'], response_billing_details
   end
 
+  def test_successful_store_with_phone_only_billing_address_option
+    billing_address = {
+      :phone => '123-456-7890'
+    }
+    credit_card = credit_card('5105105105105100')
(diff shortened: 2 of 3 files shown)


curiousepic:
Open to other methods of handling this.

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -304,12 +304,13 @@ def merge_credit_card_options(parameters, options)
 
         parameters[:credit_card] ||= {}
         parameters[:credit_card].merge!(:options => valid_options)
-        parameters[:credit_card][:billing_address] = map_address(options[:billing_address]) if options[:billing_address]
+        address = options[:billing_address]&.except(:phone)
+        return parameters if address.nil? || address.empty?
```
bpollack: This can be `address&.empty?` but I don't really think that's worth altering the patch.

bpollack:
_Generically_, I think we want some sort of "copy this hash if and only if these elements exist in it" as a helper method in `Gateway`. That's the "real" fix. But I think this is absolutely the right approach for now.

lib/active_merchant/billing/gateways/braintree_blue.rb
curiousepic: Actually we need to be checking for either case. `&.empty?` lets `nil` through.

bpollack: `nil` and `false` are bothy falsy, and `nil&.foo` return `nil`, so I think it would indeed be fine to be `... if address&.empty?`. E.g.,

```
irb(main):001:0> puts 'hello' if nil&.empty?
=> nil
irb(main):002:0> 
```

But again, I'm not convinced the half line deletion is terribly useful here.

curiousepic: Interesting, because how I described it is how it seemed to behave in testing. I'll experiment some more for the sake of understanding.

curiousepic:
```
irb(main):006:0> a = nil
=> nil
irb(main):007:0> puts "asdf" if a&.empty?
=> nil
```

curiousepic: So the thing is, in your example, we want 'hello', but it's returning nil instead.

bpollack: Bah, you're totally right; `||`, not `&&`. I got doubly turned around here because `nil.empty?` is defined (and returns `true`) if Active Support is loaded.

curiousepic: I think the confusion lies here:
```

curiousepic: Aaah, ok gotcha. Nevermind the above 👍

PR #3026, opened 2018-10-30
Barclaycard Smartpay: Improves Error Handling

deedeelavinder:
The error handling was not exposing detailed error codes and messages in
the response. This returns those codes and messages where appropriate.
This also adds and updates several tests.

Unit tests:
27 tests, 133 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote tests:
33 tests, 69 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
96.9697% passed

This failure is for "Invalid credentials" for
`test_successful_third_party_payout` and is
unrelated.

### DIFF
--- a/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
+++ b/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
@@ -168,11 +168,11 @@ module ActiveMerchant #:nodoc:
           return Response.new(false, 'Invalid credentials', {}, :test => test?)
         when '403'
           return Response.new(false, 'Not allowed', {}, :test => test?)
-        when '422'
-          return Response.new(false, 'Unprocessable Entity', {}, :test => test?)
-        when '500'
-          if e.response.body.split(' ')[0] == 'validation'
-            return Response.new(false, e.response.body.split(' ', 3)[2], {}, :test => test?)
+        when '422', '500'
+          if e.response.body.split(/\W+/).any? { |word| %w(validation configuration security).include?(word) }
+            error_message = e.response.body[/#{Regexp.escape('message=')}(.*?)#{Regexp.escape('&')}/m, 1].tr('+', ' ')
+            error_code = e.response.body[/#{Regexp.escape('errorCode=')}(.*?)#{Regexp.escape('&')}/m, 1]
+            return Response.new(false, error_code + ': ' + error_message, {}, :test => test?)
           end
         end
         raise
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -168,11 +168,11 @@ def commit(action, post, account = 'ws', password = @options[:password])
           return Response.new(false, 'Invalid credentials', {}, :test => test?)
         when '403'
           return Response.new(false, 'Not allowed', {}, :test => test?)
-        when '422'
-          return Response.new(false, 'Unprocessable Entity', {}, :test => test?)
-        when '500'
-          if e.response.body.split(' ')[0] == 'validation'
-            return Response.new(false, e.response.body.split(' ', 3)[2], {}, :test => test?)
+        when '422', '500'
+          if e.response.body.split(/\W+/).any? { |word| word  == 'validation' || 'configuration' || 'security' }
```
bpollack: `word == 'validation' || 'configuration' || 'security'` is always truthy, since `'configuration'` and `'security'` are truthy. I think you meant `...any? { |word| %q(validation configuration security).include?(word) }` or something similar.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -168,11 +168,11 @@ def commit(action, post, account = 'ws', password = @options[:password])
...
+            error_message = e.response.body[/#{Regexp.escape('message=')}(.*?)#{Regexp.escape('&')}/m, 1].tr('+', ' ')
```
bpollack: Wait, how does this work? I thought `.body` was a string; if it is, then how is that regex getting converted into an integer for offset purposes?

bpollack:
The overall plan here looks right, but the code to check the body content won't work as intended.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
deedeelavinder: 👍

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
deedeelavinder: http://ruby-doc.org/core-2.5.3/String.html#method-i-5B-5D

bpollack: I really hate this language sometimes.

PR #3027, opened 2018-11-01
Various RuboCop fixes

bpollack:
Needed a break from 3DS and wanted to ship something. These are mostly lints that only impacted a file or two.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -241,13 +241,6 @@ Layout/SpaceInsideStringInterpolation:
     - 'lib/active_merchant/billing/gateways/trust_commerce.rb'
     - 'test/unit/gateways/worldpay_test.rb'
 
-# Offense count: 4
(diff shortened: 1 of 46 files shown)


.rubocop_todo.yml
```diff
@@ -231,41 +231,11 @@ Layout/SpaceInsideReferenceBrackets:
...
-Lint/AmbiguousRegexpLiteral:
-  Enabled: false
-
 # Offense count: 148
 # Configuration parameters: AllowSafeAssignment.
 Lint/AssignmentInCondition:
   Enabled: false
 
-# Offense count: 8
-Lint/DuplicateMethods:
```
bpollack: _In general_, for this one, I deleted the first method definition, since the latter would currently be superseding it, but I did audit the deletions to make sure that was a reasonable course of action. Feel free to look and make sure you agree.

PR #3029, opened 2018-11-02
Braintree: Account for nil billing address fields

curiousepic:
The earlier fix for phone-only billing address options did not account
for other elements being present but nil. Now we compact the address.

Remote:
100% passed

Unit:
56 tests, 141 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -305,7 +305,7 @@ module ActiveMerchant #:nodoc:
         parameters[:credit_card] ||= {}
         parameters[:credit_card].merge!(:options => valid_options)
         address = options[:billing_address]&.except(:phone)
-        return parameters if address.nil? || address.empty?
+        return parameters if address.nil? || address.values.compact.empty?
         parameters[:credit_card][:billing_address] = map_address(address)
         parameters
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -305,7 +305,7 @@ def merge_credit_card_options(parameters, options)
         parameters[:credit_card] ||= {}
         parameters[:credit_card].merge!(:options => valid_options)
         address = options[:billing_address]&.except(:phone)
-        return parameters if address.nil? || address.empty?
+        return parameters if address.nil? || address.values.compact.empty?
```
bpollack: `address.all? { |_k, v| v.nil? || v.empty? }` will avoid creating the two extra `Array`s you're making with the `.values.compact` sequence. I'm not sure we care in this particular case, though.

curiousepic: Noted. I wonder if the processing time saved would actually be more than it takes to run remote tests again!

bpollack: Probably not. This kind of thing adds up, especially because Ruby's memory-hungry and doesn't have the best GC ever, but when you've got super-small arrays that are very short-lived (which is the exact case here), the benefit is exceedingly minimal.

PR #3030, opened 2018-11-02
Realex: Add verify

kheang:
Add verify method by making an Open to Buy (OTB) request:
https://developer.realexpayments.com/#!/api/process-payment/otb

Remote (1 unrelated failure from `test_realex_purchase_then_refund`):
23 tests, 122 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.6522% passed

Unit:
22 tests, 526 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -80,6 +80,13 @@ module ActiveMerchant
         commit(request)
       end
 
+      def verify(credit_card, options = {})
+        requires!(options, :order_id)
+
+        request = build_verify_request(credit_card, options)
+        commit(request)
+      end
+
       def supports_scrubbing
         true
       end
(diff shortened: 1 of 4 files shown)


test/unit/gateways/realex_test.rb
```diff
@@ -3,7 +3,8 @@
 class RealexTest < Test::Unit::TestCase
   class ActiveMerchant::Billing::RealexGateway
     # For the purposes of testing, lets redefine some protected methods as public.
-    public :build_purchase_or_authorization_request, :build_refund_request, :build_void_request, :build_capture_request
+    public :build_purchase_or_authorization_request, :build_refund_request, :build_void_request,
+      :build_capture_request, :build_otb_request
```
deedeelavinder: @kheang, What if `build_otb_request` were `build_verify_request`? It has the benefit of being more descriptive and following the established pattern.

kheang: @deedeelavinder I'm happy to change this.  I used "otb" to signal that Realex's request type is "otb", but if that causes confusion, I can swap with verify.  Thoughts?

kheang: I see now how refund/rebate is handled, consistent with your suggestion. I'll revise!

kheang: @deedeelavinder Revised per your comments.  Thanks!

lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -185,6 +192,20 @@ def build_void_request(authorization, options)
         xml.target!
       end
 
+      # Verify initiates an OTB (Open To Buy) request
```
deedeelavinder: 👍

PR #3031, opened 2018-11-02
Various RuboCop fixes

bpollack:


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -147,7 +147,7 @@ def add_customer_data(post, payment, options)
 
       def add_customer_responsible_person(post, payment, options)
         post[:payment][:person_type] = options[:person_type] if options[:person_type]
-        if options[:person_type] && options[:person_type].downcase == 'business'
+        if options[:person_type]&.casecmp('business')&.zero?
```
nfarve: :nice:

lib/active_merchant/billing/gateways/eway_managed.rb
```diff
@@ -165,7 +165,7 @@ def parse(body)
                 reply[:success]=true
               else
                 if root = REXML::XPath.first(xml, '//UpdateCustomerResult') then
-                  if root.text.downcase == 'true' then
+                  if root.text.casecmp('true').zero? then
```
nfarve: 🤔 Do we really want a case insensitive comparison here?

PR #3032, opened 2018-11-05
Braintree: Actually account for nil address fields

curiousepic:
Guess what time it is? Time for another patch to this functionality.

Phone is not the only field a billing address option may contain that
the gateway doesn't want here, so we now instead simply map the fields
first and then don't add the hash if it's empty.

Remote:
100% passed

Unit:
56 tests, 141 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -304,9 +304,10 @@ module ActiveMerchant #:nodoc:
 
         parameters[:credit_card] ||= {}
         parameters[:credit_card].merge!(:options => valid_options)
-        address = options[:billing_address]&.except(:phone)
-        return parameters if address.nil? || address.values.compact.empty?
-        parameters[:credit_card][:billing_address] = map_address(address)
+        if options[:billing_address]
+          address = map_address(options[:billing_address])
+          parameters[:credit_card][:billing_address] = address unless address.all? { |_k, v| v.nil? }
+        end
         parameters
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -304,9 +304,9 @@ def merge_credit_card_options(parameters, options)
 
         parameters[:credit_card] ||= {}
         parameters[:credit_card].merge!(:options => valid_options)
-        address = options[:billing_address]&.except(:phone)
-        return parameters if address.nil? || address.values.compact.empty?
-        parameters[:credit_card][:billing_address] = map_address(address)
+        billing_address = map_address(options[:billing_address]) if options[:billing_address]
+        return parameters if billing_address.nil? || billing_address.values.compact.empty?
```
bpollack: Okay, if we're going back in here anyway, let's fix this to `billing_address.all? { |_k, v| v.nil? || v.empty? }`

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -304,9 +304,9 @@ def merge_credit_card_options(parameters, options)
 
         parameters[:credit_card] ||= {}
         parameters[:credit_card].merge!(:options => valid_options)
-        address = options[:billing_address]&.except(:phone)
-        return parameters if address.nil? || address.values.compact.empty?
-        parameters[:credit_card][:billing_address] = map_address(address)
+        billing_address = map_address(options[:billing_address]) if options[:billing_address]
```
bpollack: Does it make sense to just have `map_address` do something reasonable when passed `nil`? We have this pattern in a few other places, looks like, and I'm wondering if they don't all need the empty check. If they do, it might be simpler to go that route.

curiousepic:
@bpollack Updated with a slightly different tack, LMK what you think.

curiousepic:
Yep!

PR #3033, opened 2018-11-05
Paymentez: Adds support for user.phone field

molbrown:
Adds user.phone field for Paymentez and updates remote test.

Unit tests:
100% passed

Remote tests:
17 tests, 33 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
64.7059% passed
These failures due to `The method authorize is not supported by carrier`.
Country credentials used for testing do not support `authorize`. Unrelated.

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -134,6 +134,7 @@ module ActiveMerchant #:nodoc:
         post[:user][:email] = options[:email]
         post[:user][:ip_address] = options[:ip] if options[:ip]
         post[:user][:fiscal_number] = options[:fiscal_number] if options[:fiscal_number]
+        post[:user][:phone] = options[:phone] || (options[:billing_address][:phone] if options[:billing_address])
       end
 
       def add_invoice(post, money, options)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,7 @@
 * Braintree: Account for nil billing address fields [curiousepic] #3029
 * Realex: Add verify [kheang] #3030
 * Braintree: Actually account for nil address fields [curiousepic] #3032
+* Paymentez: Adds support for user.phone field [molbrown] #3033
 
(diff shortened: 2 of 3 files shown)


curiousepic:
Actually, I just realized, a `phone` field often comes in from `options[:billing_address][:phone]`, so we'd want to default to that, and then check for a possible `options[:phone]`. I think this is a common pattern in other gateways so you can check others for examples.

molbrown:
Interesting, even though [Paymentez](https://paymentez.github.io/api-doc/#debit-with-credit-card) doesn't appear to nest `[:phone]` like that?

curiousepic:
Yep, at least in our case. When Core sends a request to AM, the `billing_address` hash option is filled in with any details we have from the payment method, including phone. So we probably want to allow overriding, by also adding a new top-level `options[:phone]` GSF as you were planning, but also default to the `phone` in the `billing_address` if the top-level one isn't present. This is a pretty common pattern.

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -134,6 +134,7 @@ def add_customer_data(post, options)
         post[:user][:email] = options[:email]
         post[:user][:ip_address] = options[:ip] if options[:ip]
         post[:user][:fiscal_number] = options[:fiscal_number] if options[:fiscal_number]
+        post[:user][:phone] = (options[:billing_address][:phone] if options[:billing_address]) || (options[:phone])
```
curiousepic: Looks good, but if we want the GSF to act as an override, it will need to come first in the `||` check.

molbrown: FYI, the Travis failure is due to a Rubocop problem with a file outside of this commit, being addressed elsewhere.

bpollack: It's been fixed; you should be good if you rebase. (I literally _just_ merged it; it's not like you missed it.)

curiousepic:
LGTM 🚢

PR #3035, opened 2018-11-06
Visanet Peru: fix RuboCop error

bpollack:
Follow-up to #3034


bpollack:
@deedeelavinder I'll hold merging this until you also 👍 so we don't repeat last night. 😬

bpollack:
@deedeelavinder @curiousepic Would you mind checking one last time? I fixed up the unit tests so they'd actually fail/test the right thing.

test/unit/gateways/visanet_peru_test.rb
```diff
@@ -16,20 +16,20 @@ def setup
...
+    @gateway.expects(:ssl_request).with(:post, any_parameters).returns(successful_authorize_response)
+    @gateway.expects(:ssl_request).with(:put, any_parameters).returns(successful_capture_response)
 
     response = @gateway.purchase(@amount, @credit_card, @options)
     assert_success response
     assert_equal 'OK', response.message
 
     assert_match %r([0-9]{9}|$), response.authorization
-    assert_equal @options[:order_id], response.params['externalTransactionId']
+    assert_equal 'de9dc65c094fb4f1defddc562731af81', response.params['externalTransactionId']
```
bpollack: Previously, the requests were returned in the wrong order, so the author of this unit test checked for the _authorization_ response. This is incorrect; the ID checked here is now the one that's actually embedded in `successful_capture_response` (which you can look up in the JSON way below).

test/unit/gateways/visanet_peru_test.rb
```diff
@@ -16,20 +16,20 @@ def setup
...
     assert response.test?
   end
 
   def test_failed_purchase
-    @gateway.expects(:ssl_request).returns(failed_authorize_response_bad_card)
+    @gateway.expects(:ssl_request).with(:post, any_parameters).returns(failed_authorize_response_bad_card)
```
bpollack: I updated some, but not all, tests to make sure we would hit both `:post` and `:put` a bit.

test/unit/gateways/visanet_peru_test.rb
```diff
@@ -90,14 +90,14 @@ def test_successful_refund
     response = @gateway.purchase(@amount, @credit_card, @options)
     assert_success response
 
-    @gateway.expects(:ssl_request).returns(successful_refund_response)
+    @gateway.expects(:ssl_request).with(:put, any_parameters).returns(successful_refund_response)
```
bpollack: You'd think that this would be `post` if you glanced quickly at the `method`…erm, method, but I actually can't find the genuine `refund` route tested, period; just `cancelDespoit`--and that's indeed a `put`.

deedeelavinder: @bpollack Do you think it is beyond the scope of this PR to add tests for `prepare_refund_data`?

test/unit/gateways/visanet_peru_test.rb
deedeelavinder: 👍

test/unit/gateways/visanet_peru_test.rb
deedeelavinder: 👍

test/unit/gateways/visanet_peru_test.rb
bpollack: _Probably_. But more practically, I just can't get to it today; I need to wrap up some other gateway work. So I'd either need someone else to add that test or we'll need to hold off merging this today. Since builds are technically "broken", I'd rather just go ahead and merge this.

deedeelavinder: 👍

deedeelavinder:
@bpollack Re-approved. 👍

PR #3036, opened 2018-11-07
Paymentez: Does not send nil for empty parameter phone

molbrown:
Unit tests:
100% passed

Remote tests:
18 tests, 34 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
61.1111% passed
These failures due to The method authorize is not supported by carrier.
Country credentials used for testing do not support authorize. Unrelated.

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -134,7 +134,8 @@ module ActiveMerchant #:nodoc:
         post[:user][:email] = options[:email]
         post[:user][:ip_address] = options[:ip] if options[:ip]
         post[:user][:fiscal_number] = options[:fiscal_number] if options[:fiscal_number]
-        post[:user][:phone] = options[:phone] || (options[:billing_address][:phone] if options[:billing_address])
+        post[:user][:phone] = options[:phone] || (options[:billing_address][:phone] if options[:billing_address] &&
+          options[:billing_address][:phone])
       end
 
       def add_invoice(post, money, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -134,7 +134,8 @@ def add_customer_data(post, options)
         post[:user][:email] = options[:email]
         post[:user][:ip_address] = options[:ip] if options[:ip]
         post[:user][:fiscal_number] = options[:fiscal_number] if options[:fiscal_number]
-        post[:user][:phone] = options[:phone] || (options[:billing_address][:phone] if options[:billing_address])
+        post[:user][:phone] = options[:phone] || (options[:billing_address][:phone] if options[:billing_address] &&
+          options[:billing_address][:phone])
```
curiousepic: Nice, I thought that might be it as well. A couple small things that probably aren't worth changing for this:
- This is a good use case for the `.dig` method (new to the recently updated ruby version) https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig
- Usually we don't bother sticking to the 80 characters per line in adapters, but NBD.

test/remote/gateways/remote_paymentez_test.rb
```diff
@@ -34,6 +34,18 @@ def test_successful_purchase_with_more_options
     assert_success response
   end
 
+  def test_no_phone
```
curiousepic: Let's make this more descriptive, something like `test_successful_purchase_without_phone_option`

PR #3040, opened 2018-11-08
Credorax: add submerchant_id support

bpollack:
Unit: 20 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


bpollack:
NB, remote tests are largely currently failing because their fake Visa processor is down, so I omitted the results, but they are just failing for that reason, not anything related to the h3 field.

PR #3041, opened 2018-11-09
Worldpay: Pass stored credential option fields

curiousepic:
In preparation for Visa card-on-file requirements.
WorldPay's documentation on this subject can be found at:
http://support.worldpay.com/support/kb/gg/corporate-gateway-guide/content/industryschemeextras/storedcredentials.htm

Remote:
27 tests, 109 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
39 tests, 222 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -248,7 +248,18 @@ module ActiveMerchant #:nodoc:
               xml.tag! 'session', 'shopperIPAddress' => options[:ip] if options[:ip]
               xml.tag! 'session', 'id' => options[:session_id] if options[:session_id]
             end
+            add_stored_credential_options(xml, options) if options[:stored_credential_usage]
           end
+        end
+      end
+
+      def add_stored_credential_options(xml, options={})
+        if options[:stored_credential_initiated_reason]
+          xml.tag! 'storedCredentials', 'usage' => options[:stored_credential_usage], 'merchantInitiatedReason' => options[:stored_credential_initiated_reason] do
+            xml.tag! 'schemeTransactionIdentifier', options[:stored_credential_transaction_id] if options[:stored_credential_transaction_id]
+          end
+        else
+          xml.tag! 'storedCredentials', 'usage' => options[:stored_credential_usage]
         end
       end
 
--- a/test/remote/gateways/remote_worldpay_test.rb
+++ b/test/remote/gateways/remote_worldpay_test.rb
@@ -11,7 +11,10 @@ class RemoteWorldpayTest < Test::Unit::TestCase
     @declined_card = credit_card('4111111111111111', :first_name => nil, :last_name => 'REFUSED')
     @threeDS_card = credit_card('4111111111111111', :first_name => nil, :last_name => '3D')
 
-    @options = {order_id: generate_unique_id, email: 'wow@example.com'}
+    @options = {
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -248,10 +248,29 @@ def add_payment_method(xml, amount, payment_method, options)
...
+          xml.tag! 'storedCredentials', 'usage' => options[:stored_credential_usage]
+        end
+      end
+
+      def merchant_initiated?(xml, options={})
+        options[:stored_credential_usage] && options[:stored_credential_initiated_reason] && options[:stored_credential_transaction_id]
+      end
+
+      def customer_initiated?(xml, options={})
+        options[:stored_credential_usage]
```
bpollack: Okay, this is subtle, but: given that `merchant_initiated?` and `customer_initiated?` both check for `options[:stored_credential_usage]`, it's possible for `customer_initiated?` to bogusly return `true` (or at least a truthy value). You won't hit that behavior right now, because you're testing for `merchant_iniated?` first, but I can _easily_ see how a superficial refactoring could result in unexpected behavior. So what I'd instead suggest is tweaking this so that the order `customer_initiated?` and `merchant_iniated?` are called in doesn't matter. Something like this:

```suggestion
        !merchant_initiated?(xml, options) && options[:stored_credential_usage]
```

test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -246,19 +293,19 @@ def test_transcript_scrubbing
   # def get_authorization
   #   assert_success(response = @gateway.purchase(@amount, @credit_card, @options))
   #   assert response.authorization
-  #   puts "auth: " + response.authorization
+  #   puts 'auth: ' + response.authorization
```
bpollack: Does this test pass now, out of curiosity? If not, based on the blame, maybe we should just ax it; seems like it's been dead for a long time, Jim.

test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -246,19 +293,19 @@ def test_transcript_scrubbing
...
   # end
 
   # def test_refund
   #   refund = @gateway.refund(@amount, 'replace_with_authorization')
   #   assert_success refund
-  #   assert_equal "SUCCESS", refund.message
+  #   assert_equal 'SUCCESS', refund.message
```
bpollack: Ditto for this and the next one.

lib/active_merchant/billing/gateways/worldpay.rb
curiousepic: In fact I'm not entirely sure we even need/want these methods, rather than just passing through what is present. Currently since they aren't validated in the sandbox, I don't know if partial info would fail or how, but since we're basically leaving all this behavior up to the merchant, I don't think there's much harm in it.

test/remote/gateways/remote_worldpay_test.rb
curiousepic: They do in theory, just takes time.

curiousepic:
Updated. Still needs some in-built logic since two of the values go into the same tag (unless there's a smarter way to add a second value to the same tag?)

PR #3042, opened 2018-11-11
Moneris: Adds Credential On File Logic

deedeelavinder:
This allows Credential On File information to be passed in the options
hash. `cof_enabled` must be set to `true`. Values for the three
CoF fields: `issuer_id`, `payment_indicator`, and `payment_information`
must also be included.

Unit Tests:
37 tests, 188 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote Tests:
29 tests, 114 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/moneris.rb
+++ b/lib/active_merchant/billing/gateways/moneris.rb
@@ -33,7 +33,8 @@ module ActiveMerchant #:nodoc:
         requires!(options, :login, :password)
         @cvv_enabled = options[:cvv_enabled]
         @avs_enabled = options[:avs_enabled]
-        options = { :crypt_type => 7 }.merge(options)
+        @cof_enabled = options[:cof_enabled]
+        options[:crypt_type] = 7 unless options.has_key?(:crypt_type)
         super
       end
 
@@ -46,17 +47,18 @@ module ActiveMerchant #:nodoc:
         requires!(options, :order_id)
         post = {}
         add_payment_source(post, creditcard_or_datakey, options)
-        post[:amount]     = amount(money)
-        post[:order_id]   = options[:order_id]
-        post[:address]    = options[:billing_address] || options[:address]
+        post[:amount] = amount(money)
+        post[:order_id] = options[:order_id]
+        post[:address] = options[:billing_address] || options[:address]
         post[:crypt_type] = options[:crypt_type] || @options[:crypt_type]
+        add_cof(post, options) if @cof_enabled
         action = if post[:cavv]
                    'cavv_preauth'
                  elsif post[:data_key].blank?
                    'preauth'
                  else
                    'res_preauth_cc'
-        end
+                 end
         commit(action, post)
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -33,7 +33,8 @@ def initialize(options = {})
         requires!(options, :login, :password)
         @cvv_enabled = options[:cvv_enabled]
         @avs_enabled = options[:avs_enabled]
-        options = { :crypt_type => 7 }.merge(options)
+        @cof_enabled = options[:cof_enabled]
+        options = {:crypt_type => 7}.merge(options)
```
bpollack: I get what's going on here, and I know it's not your code or anything, but this is a _ton_ slower (and wastes way more memory) than just `options[:crypt_type] = 7 if !options.has_key?(:crypt_type)`. _If you're back in here for some other reason_, might be worth switching it.

lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -182,16 +185,16 @@ def expdate(creditcard)
...
         if payment_method.is_a?(String)
-          post[:data_key]   = payment_method
-          post[:cust_id]    = options[:customer]
+          post[:data_key] = payment_method
+          post[:cust_id] = options[:customer]
         else
           if payment_method.respond_to?(:track_data) && payment_method.track_data.present?
-            post[:pos_code]   = '00'
-            post[:track2]     = payment_method.track_data
+            post[:pos_code] = '00'
```
bpollack: It's probably not worth going back and undoing, but the whitespace alignment changes here (and in the rest of this file) make the diff appear much larger than it actually needed to be, and make it a bit harder to find the actual logic changes. (They also shouldn't be necessary; it's fine if you found it annoying, but RuboCop should not have complained, since it allows extra whitespace specifically for this kind of alignment.) It'd be to keep logic and whitespace changes separate when they easily can be in the future.

bpollack:
The RuboCop failure appears legitimate as well (and very quick to fix).

lib/active_merchant/billing/gateways/moneris.rb
deedeelavinder: Sure, no problem.

lib/active_merchant/billing/gateways/moneris.rb
deedeelavinder: 👍 
Rubocop had lots to say about these files and I thought reformatting the file would help. It did some, but there were lots of adjustments like these. I agree that that it is easier to keep logic changes separate and intended to take these out this morning. I hadn't requested any reviewers yet for just that reason. Since you have already approved, and waded through the extra long diff, I will just leave them in.

bpollack: RuboCop shouldn't have complained about any of these, though. I actually do want to figure out why that's happening, since it implies we might've configured RuboCop wrong.

deedeelavinder: Rubocop is fine. Its complaints were about my additions (lots of options massaging) and reformatting addressed most of the issues I introduced. It also deleted all the extra spaces, which as I said, I intended to undo this morning.

bpollack: Oooh, got it! Sorry, I've clearly got a reading comprehension issue this morning. All good! 👍

bpollack: I do find it weird, I indeed don't see you having requested review, but I got pinged by GitHub on this this morning, and thought I saw @dtykocki and me tagged. I wonder if GitHub somehow sent emails to _recommended_ reviewers? Or something else?

PR #3043, opened 2018-11-12
Paymentez: Supports phone field, does not send if empty

molbrown:
Unit:
100% passed

Remote:
19 tests, 35 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
68.4211% passed
These failures due to `The method authorize is not supported by carrier`.
Country credentials used for testing do not support authorize. Unrelated.

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -134,6 +134,9 @@ module ActiveMerchant #:nodoc:
         post[:user][:email] = options[:email]
         post[:user][:ip_address] = options[:ip] if options[:ip]
         post[:user][:fiscal_number] = options[:fiscal_number] if options[:fiscal_number]
+        if phone = options[:phone] || options.dig(:billing_address, :phone)
+          post[:user][:phone] = phone
+        end
       end
 
       def add_invoice(post, money, options)
--- a/test/remote/gateways/remote_paymentez_test.rb
+++ b/test/remote/gateways/remote_paymentez_test.rb
@@ -5,8 +5,8 @@ class RemotePaymentezTest < Test::Unit::TestCase
     @gateway = PaymentezGateway.new(fixtures(:paymentez))
 
     @amount = 100
-    @credit_card = credit_card('4111111111111111', verification_value: '555')
-    @declined_card = credit_card('4242424242424242', verification_value: '555')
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -134,6 +134,9 @@ def add_customer_data(post, options)
         post[:user][:email] = options[:email]
         post[:user][:ip_address] = options[:ip] if options[:ip]
         post[:user][:fiscal_number] = options[:fiscal_number] if options[:fiscal_number]
+        if phone = options[:phone] || (options[:billing_address][:phone] if options[:billing_address])
```
bpollack: I really don't like the `foo || (bar if baz)` pattern, because it's very unclear what actually happens. (You [responded `ew` to this same pattern](https://spreedly.slack.com/archives/C5GBHQTMH/p1537286552000100), BTW.). There's a couple routes here, but since we're on Ruby 2.3.6 now, I'd suggest `if phone = (options[:phone] || options[:billing_address]&.[](:phone))` or `phone = (options[:phone] || (options[:billing_address] || [])[:phone])`, or something similar to be a bit more explicit.

test/remote/gateways/remote_paymentez_test.rb
```diff
@@ -5,8 +5,8 @@ def setup
     @gateway = PaymentezGateway.new(fixtures(:paymentez))
 
     @amount = 100
-    @credit_card = credit_card('4111111111111111', verification_value: '555')
-    @declined_card = credit_card('4242424242424242', verification_value: '555')
+    @credit_card = credit_card('4111111111111111', verification_value: '666')
```
bpollack: 😈

lib/active_merchant/billing/gateways/paymentez.rb
molbrown: Thanks, updated. I was not familiar with `&.`, perfect!

bpollack:
Oh BTW @molbrown don't forget to add to `CHANGELOG` prior to merging.

lib/active_merchant/billing/gateways/paymentez.rb
bpollack: Better than what I just said is to use `.dig`; sorry I spaced that. So that'd be `phone = options[:phone] || options.dig(:billing_address, :phone)`

PR #3044, opened 2018-11-12
Adyen: Return AVS and CVC Result

nfarve:
Returns the result of avs and cvc in the response if it is provided.

........................

24 tests, 115 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------

....................................

36 tests, 90 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


nfarve:
I could not get the remote response to actually return the AVS and CVC so I used a actual response sample from the ticket and update unit tests.

PR #3046, opened 2018-11-13
Credorax: add rich 3DS support

bpollack:


curiousepic:
Let's add the new fields to scrub.

curiousepic:
Well, I've never been clear on whether they really should be, curious to hear your opinion on that.

bpollack:
Well, at a bare minimum, I'm passing the CC in the MPI requests, which definitely needs to be scrubbed, so…let's start there. :)

curiousepic:
I think the unit tests look good, though with the external MPI craziness I can't quite tell if there's a permutation of request/response scenarios that isn't tested. :not_sure_if:

bpollack:
@curiousepic Scrubbing added, plus some cleanup to reflect how `three_d_options` is usually passed.

bpollack:
(And I also added an additional remote tests, but I think this is as far as it goes.)

test/remote/gateways/remote_credorax_test.rb
```diff
@@ -168,6 +186,17 @@ def test_transcript_scrubbing
     assert_scrubbed(@credit_card.verification_value.to_s, clean_transcript)
   end
 
+  def test_3ds_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options.merge({three_d_secure: {}, xid: '12345678901234567890'}))
+    end
+    clean_transcript = @gateway.scrub(transcript)
+    puts "!!!!!!\n#{clean_transcript}\n!!!!!"
```
curiousepic: Do we want this here?

bpollack:
> I think the unit tests look good, though with the extra MPI craziness I can't quite tell if there's a permutation of request/response scenarios that isn't covered (that could be). :not_sure_if:

It might be a good exercise for both of us if you could articulate anything obviously missing, even vaguely; I don't exactly have super-high confidence in this code.

test/remote/gateways/remote_credorax_test.rb
bpollack: 🤦‍♂️ No. Removing.

bpollack:
@curiousepic Nuked the erroneous `puts`, plus fixed the RuboCop error.

bpollack:
@nfarve Yeah. I'm going to retract this; it's been subsumed by the Core PR entirely.

PR #3048, opened 2018-11-14
RuboCop: fix Layout/SpaceAfterComma

bpollack:


bpollack:
@nfarve This is just the first in that RuboCop series.

PR #3051, opened 2018-11-15
Moneris: Allows cof_enabled gateway to process non-cof transactions

deedeelavinder:
Adds a bypass for transactions without credential-on-file details when @cof_enabled is
true.

Unit Tests:
37 tests, 188 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote Tests:
31 tests, 119 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
93.5484% passed

(The two failing tests are not related and appear to be timing-based.)

### DIFF
--- a/lib/active_merchant/billing/gateways/moneris.rb
+++ b/lib/active_merchant/billing/gateways/moneris.rb
@@ -293,7 +293,7 @@ module ActiveMerchant #:nodoc:
           when :cvd_info
             transaction.add_element(cvd_element(parameters[:cvd_value])) if @cvv_enabled
           when :cof_info
-            transaction.add_element(credential_on_file(parameters)) if @cof_enabled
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -324,10 +324,14 @@ def cvd_element(cvd_value)
         element
       end
 
+      def cof_details_present?(parameters)
+        parameters[:issuer_id] && parameters[:payment_indicator] && parameters[:payment_information]
```
curiousepic: Just to verify (since this is not true for Worldpay), are all three fields actually required in both initial and subsequent cof txns?

deedeelavinder: yes, @curiousepic, all three are required for first and subsequent cof transactions.

PR #3059, opened 2018-11-21
Forte: Allow void on capture

nfarve:
Forte uses different transaction id's for captures. When trying to void
a capture this results in a error. This saves the original auth
transaction id and authorization id in the capture authorization so the original auth will be voided
or refunded.

Loaded suite test/unit/gateways/forte_test

20 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Loaded suite test/remote/gateways/remote_forte_test

21 tests, 59 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/forte.rb
+++ b/lib/active_merchant/billing/gateways/forte.rb
@@ -191,7 +191,7 @@ module ActiveMerchant #:nodoc:
           success_from(response),
           message_from(response),
           response,
-          authorization: authorization_from(response),
+          authorization: authorization_from(response, parameters),
           avs_result: AVSResult.new(code: response['response']['avs_result']),
           cvv_result: CVVResult.new(response['response']['cvv_code']),
           test: test?
@@ -219,8 +219,12 @@ module ActiveMerchant #:nodoc:
         response['response']['response_desc']
       end
 
-      def authorization_from(response)
-        [response.try(:[], 'transaction_id'), response.try(:[], 'response').try(:[], 'authorization_code')].join('#')
+      def authorization_from(response, parameters)
+        if parameters[:action] == 'capture'
+          [response['transaction_id'], response.dig('response', 'authorization_code'), parameters[:transaction_id], parameters[:authorization_code]].join('#')
+        else
+          [response['transaction_id'], response.dig('response', 'authorization_code')].join('#')
+        end
       end
 
       def endpoint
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/forte.rb
```diff
@@ -219,8 +219,12 @@ def message_from(response)
         response['response']['response_desc']
       end
 
-      def authorization_from(response)
-        [response.try(:[], 'transaction_id'), response.try(:[], 'response').try(:[], 'authorization_code')].join('#')
+      def authorization_from(response, parameters)
+        if parameters[:action] == 'capture'
+          [response.try(:[], 'transaction_id'), response.try(:[], 'response').try(:[], 'authorization_code'), parameters[:transaction_id], parameters[:authorization_code]].join('#')
```
bpollack:
```suggestion
          [response.dig('transaction_id', 'response', 'authorization_code'), parameters[:transaction_id], parameters[:authorization_code]].join('#')
```

lib/active_merchant/billing/gateways/forte.rb
```diff
@@ -219,8 +219,12 @@ def message_from(response)
...
+        else
+          [response.try(:[], 'transaction_id'), response.try(:[], 'response').try(:[], 'authorization_code'), ''].join('#')
```
bpollack:
```suggestion
          [response['transaction_id'], response.dig('response', 'authorization_code'), ''].join('#')
```

lib/active_merchant/billing/gateways/forte.rb
nfarve: Err I think this is different. Maybe `response.dig('transaction_id'), response.dig('response'), response.dig('authorization_code'), parameters[:transaction_id], parameters[:authorization_code].join('#')`

bpollack: Whoops, point. Fixed the suggestion. `response.dig('foo')`, `response['foo']`, and `response.try(:[], 'foo')` are all the same, though, unless I'm missing something, so we might as well go for just `response['foo']` in that case.

nfarve: Updated!

PR #3064, opened 2018-11-27
Rubocop fixes

bpollack:
These are all leading whitespace fixes. While the number of lines changed looks scary, if you turn on "ignore whitespace," basically the whole thing goes away.

### DIFF
--- a/lib/active_merchant/billing/gateways/eway_managed.rb
+++ b/lib/active_merchant/billing/gateways/eway_managed.rb
@@ -150,7 +150,7 @@ module ActiveMerchant #:nodoc:
         reply = {}
         xml = REXML::Document.new(body)
         if root = REXML::XPath.first(xml, '//soap:Fault') then
-           reply=parse_fault(root)
+          reply=parse_fault(root)
         else
           if root = REXML::XPath.first(xml, '//ProcessPaymentResponse/ewayResponse') then
             # Successful payment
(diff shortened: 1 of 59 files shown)


lib/active_merchant/billing/gateways/eway_managed.rb
```diff
@@ -232,33 +232,33 @@ def soap_request(arguments, action)
...
         xml = Builder::XmlMarkup.new :indent => 2
-          xml.instruct!
-          xml.tag! 'soap12:Envelope', {'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance', 'xmlns:xsd' => 'http://www.w3.org/2001/XMLSchema', 'xmlns:soap12' => 'http://www.w3.org/2003/05/soap-envelope'} do
-            xml.tag! 'soap12:Header' do
-              xml.tag! 'eWAYHeader', {'xmlns' => 'https://www.eway.com.au/gateway/managedpayment'} do
-                xml.tag! 'eWAYCustomerID', @options[:login]
-                xml.tag! 'Username', @options[:username]
-                xml.tag! 'Password', @options[:password]
-              end
+        xml.instruct!
```
nfarve: Shouldn't this be indented?

lib/active_merchant/billing/gateways/paystation.rb
```diff
@@ -100,101 +100,101 @@ def scrub(transcript)
 
       private
 
-        def new_request
```
nfarve: Did most of this file get re-ordered?

lib/active_merchant/billing/gateways/eway_managed.rb
bpollack: No; there's no `do` on the previous line, and this is not a block. It's honestly really misleading for exactly that reason as-is.

lib/active_merchant/billing/gateways/paystation.rb
bpollack: No. I'm not honestly entirely clear what you're responding to; the lines match up one-for-one.

bpollack: Oh, I see, GitHub's being dumb in side-by-side mode. Either turn on "ignore whitespace" or use a different diff viewer. Or you can look at the actual line numbers.

PR #3067, opened 2018-11-29
Experiment: TreeCopy

bpollack:
This is an experiment with having a generic tree copier as a way to simplify gateway implementations. It's not done, and may be overengineered, but it's hopefully an interesting thought-experiment on what a more declarative version of Active Merchant might look like. The new TreeCopy module is largely extracted from what was left of me trying to add this to Stripe on the last time I was in there.

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -1,6 +1,7 @@
 require 'net/http'
 require 'net/https'
 require 'active_merchant/billing/response'
+require 'active_merchant/tree_copy'
 
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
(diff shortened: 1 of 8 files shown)


bpollack:
The build error's just RuboCop, since this patch is so old; just fixed everything up and re-pushed. Again, I doubt this should be merged as-is for a variety of reasons anyway, but this logic _does_ work.

bpollack:
I converted over (part of) Paymentez and pushed that up as a second commit, just to remind myself what this feels like to do fresh. I do think it's overall an improvement; there's *significantly* less redundancy. But it doesn't do anything for the XML gateways, and it's not quite fleshed-out enough to replace _all_ the assignments in the non-XML ones. In particular, we struggle a bit with handling parameters that are _not_ initialized from another hash. That may be fine, or maybe not; I'm not sure.

I do think the name `TreeCopy` is wrong, though, if you want something easy to bikeshed over. Maybe `HashShaper`?

PR #3068, opened 2018-11-29
Worldpay: support MasterCard credits

bpollack:
MasterCard and Visa credits do not work the same way, so we were erroneously counting MasterCard credits as failed. Count them as successful.

Unit: 40 tests, 227 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 28 tests, 111 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* Worldpay: handle Visa and MasterCard payouts differently [bpollack] #3068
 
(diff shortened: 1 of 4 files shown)


bpollack:
@activemerchant/spreedly-endpoints I wouldn't normally ping this quick, but I want to make sure this gets to prod sufficiently far before EOW that I don't leave it hanging.

PR #3072, opened 2018-12-04
Fix Type Error on Orbital Gateway

joeyjoejoejr:
## Issue

On our production system we updated some configs and our chase merchant id changed from a string to an integer.  Since since length is simply called on the merchant_id without coercion or any type of type checking, this was a tough error to track down.

## What Changed
* Added a test demonstrating the error.
* Coerce the merchant_id to a string in the initializer

screen shot 2018-12-04 at 10 46 32 am


bpollack:
Thanks, @joeyjoejoejr . Squashed and merged as e7aad1c72559aaef84634d7b8a43b2a8264442ed

PR #3075, opened 2018-12-05
Fix FirstData E4 Card Type Whitespace Bug

jcbantuelle:
When attempting to process a payment using the Amex card type specified by FirstData/PayEezy, the brand conversion was failing due to the whitespace in "american express" not converting to a symbol that matched the defined brands.

To make this more robust, I've updated the conversion to replace whitespaces with underscores, which allows card types to align more closely with the defined brands and fixes the immediate issue of "american express" not functioning as a card type.


bayprogrammer:
@jcbantuelle, thank you for your patch! After reviewing this, and the history of the FirstData gateway adapter, I do see we have prior art for this approach here:

https://github.com/activemerchant/active_merchant/pull/1504#issuecomment-74782433 

From that PR:

> When using the FirstData token response from their server, that brand value is 'mastercard'.

However, if I understand things correctly, credit card objects instantiated with these non-ActiveMerchant-standard brand strings (including for other special case we've added previously) will be invalid (even if they can still be used), per:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card.rb#L348

I'm thinking it would it make more sense, moving forward, to have client code using ActiveMerchant normalize the card brand value coming from FirstData's token response to one of our supported card brand strings prior to creating an instance of CreditCard. We have a well-defined set of supported brands listed here:

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card.rb#L83

I'm going to close this for now, but *please* feel free to write back if you have further thoughts on this or if I'm missing something crucial. Thank you again!

PR #3082, opened 2018-12-12
Does not fail validation for blank cvv on maestro cards

filipebarcos:
This change is allowing users to pass `cvv: blank` and pass validation
even when `CreditCard.require_verification_value = true` for maestro,
since cvv length is 0.

I believe this was some backward compatibility issue when we changed maestro card validations to include the new/correct numbers.

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -362,7 +362,7 @@ module ActiveMerchant #:nodoc:
           unless valid_card_verification_value?(verification_value, brand)
             errors << [:verification_value, "should be #{card_verification_value_length(brand)} digits"]
           end
-        elsif requires_verification_value?
+        elsif requires_verification_value? && !valid_card_verification_value?(verification_value, brand)
           errors << [:verification_value, 'is required']
         end
         errors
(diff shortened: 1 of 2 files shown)


bpollack:
I removed myself since I can't work as actively on this anymore, but adding @activemerchant/spreedly-endpoints or @activemerchant/spreedly-runtime might be worthwhile.

curiousepic:
I don't have any obvious concerns here, but I feel like it could stand a second opinion so I'll let someone else approve.

lib/active_merchant/billing/credit_card.rb
```diff
@@ -362,7 +362,7 @@ def validate_verification_value #:nodoc:
           unless valid_card_verification_value?(verification_value, brand)
             errors << [:verification_value, "should be #{card_verification_value_length(brand)} digits"]
           end
-        elsif requires_verification_value?
+        elsif requires_verification_value? && !valid_card_verification_value?(verification_value, brand)
```
elfassy: might be a good idea to move the `valid_card_verification_value?` up a level since it's use in both branches. Something like
```ruby
unless valid_card_verification_value?(verification_value, brand)
  if verification_value?
    errors << [:verification_value, "should be #{card_verification_value_length(brand)} digits"]
  elsif requires_verification_value?
     errors << [:verification_value, 'is required']
  end
end
```

PR #3086, opened 2018-12-19
Stripe: Add 3DS

nfarve:
Adds 3DS to the Stripe gateway. Remote tests for 3DS are in their own
file.

Loaded suite test/unit/gateways/stripe_test
.....................................................................................................................................

133 tests, 713 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/remote/gateways/remote_stripe_3ds_test
.....

5 tests, 21 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/remote/gateways/remote_stripe_test
...................................................................

67 tests, 313 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 * Braintree Blue: Support Level 2 and 3 data fields [curiousepic] #3094
 * Braintree Blue: Refactor line_items field [curiousepic] #3100
 * TrustCommerce: Use `application_id` [nfarve] #3103
+* Stripe: Add 3DS Support [nfarve] #3086
 
 == Version 1.89.0 (December 17, 2018)
 * Worldpay: handle Visa and MasterCard payouts differently [bpollack] #3068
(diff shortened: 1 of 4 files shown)


nfarve:
@curiousepic and @molbrown ready for 👀

curiousepic:
Any particular reason remote tests are their own file? I'd think we'd want to have them in the main test run so there's less of a chance something would get missed if a future change impacts them.

nfarve:
@curiousepic Stripe seems to have most of the tests broken out like this. I think because it's become some big. Right now there are 5 remote tests for stripe. The base remote test is pretty huge.

curiousepic:
@nfarve Ah, I didn't notice/forgot about that (which kind of demonstrates my point 😬), but it does make sense.

PR #3093, opened 2019-01-02
Fix usaepay transaction invoice

lancecarlson:
This is potentially a breaking change. The :invoice: option now points to the invoice field in the gateway. This field has limitations, but is required. The :order_id: option now points to the orderid field in the gateway, which is much larger and used for additional reference and is optional. This is the correct way to integrate with the gateway.

### DIFF
--- a/lib/active_merchant/billing/gateways/usa_epay_transaction.rb
+++ b/lib/active_merchant/billing/gateways/usa_epay_transaction.rb
@@ -198,6 +198,7 @@ module ActiveMerchant #:nodoc:
 
       def add_invoice(post, options)
         post[:invoice]      = options[:invoice]
+        post[:orderid]      = options[:order_id]
         post[:description]  = options[:description]
       end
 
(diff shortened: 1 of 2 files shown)


nfarve:
@lancecarlson how is the invoice used differently than the order_id? Am I correct in understanding that `invoice` is only a number but `order_id` can be a string?

lancecarlson:
@nfarve 

order_id: Order identifier. This field can be used to reference the order to which this transaction corresponds. This field can contain up to 64 characters and should be used instead of UMinvoice when orderid is longer that 10 digits. 

invoice: Unique Invoice or order number. 10 digits. While not required, it is strongly recommended that this field be populated for CC:Sale, CC:AuthOnly, CC:PostAuth, CC:Credit, Check:Sale and Check:Credit 

Internally we use both for different things. It makes it pretty confusing when invoice is pointing to order_id. It also makes it difficult to specify both.

PR #3098, opened 2019-01-03
Use contemporary Pin Payments URL

montdidier:
While the old URL(s) are still supported and will be for the foreseeable future Pin Payments is actively preferencing the pinpayments.com domain.


montdidier:
Looks like current check failure is a bundler failure on Travis to me, rather than a problem with the PR.

PR #3100, opened 2019-01-03
Braintree Blue: Refactor line_items field

curiousepic:
Removed hilariously redundant code.

Unit:
57 tests, 150 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
68 tests, 384 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -621,17 +621,7 @@ module ActiveMerchant #:nodoc:
         parameters[:discount_amount] = options[:discount_amount] if options[:discount_amount]
         parameters[:ships_from_postal_code] = options[:ships_from_postal_code] if options[:ships_from_postal_code]
 
-        if options[:line_items]
-          items = []
-          options[:line_items].each do |line_item|
-            item = {}
-            line_item.each do |key, value|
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -621,17 +621,7 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
         parameters[:discount_amount] = options[:discount_amount] if options[:discount_amount]
         parameters[:ships_from_postal_code] = options[:ships_from_postal_code] if options[:ships_from_postal_code]
 
-        if options[:line_items]
-          items = []
```
molbrown: Is it possible that `line_items` ever comes in as something other than an array? That's really the only reason I could see for keeping this. If not, 👍

curiousepic: It's probably unlikely, but if it's not, it will cause a gateway error rather than an AM error, so I'm ok with that.

curiousepic: Most other adapters follow this pattern instead.

PR #3102, opened 2019-01-04
Spreedly additional fields merchant name descriptor, merchant location descriptor

montdidier:


montdidier:
When merging, I would imagine you'll want to squash.

lib/active_merchant/billing/gateways/spreedly_core.rb
```diff
@@ -200,6 +208,7 @@ def add_credit_card(doc, credit_card, options)
           doc.last_name(credit_card.last_name)
           doc.month(credit_card.month)
           doc.year(credit_card.year)
+          doc.verification_value(credit_card.verification_value) if credit_card.verification_value?
```
jknipp: @montdidier This value is already passed [here](https://github.com/activemerchant/active_merchant/pull/3102/files#diff-3d8ec2e8f34093ada56bc96cef17f9ecR206).

Did you intend to send the `verification_value ` only if it is present instead of always sending it?

montdidier: Good catch. It is just wrong.

curiousepic:
Hi @montdidier, FYI, given the storied history of this branch, it was a lot smoother to trim just the relevant changes to a new branch and merge with you as the author https://github.com/activemerchant/active_merchant/commit/86182b1189ec0818c0892e9429b47a7af7b74ff2
Thanks!

montdidier:
@curiousepic Thank you.

PR #3103, opened 2019-01-04
TrustCommerce: Use `application_id`

nfarve:
Allows for `application_id` to be used as `aggregatorID`. 3 failing
remotes test unrelated to this change.

- test_store_failure
- test_unstore_failure
- test_recurring_failure

Loaded suite test/unit/gateways/trust_commerce_test
Started
..........

10 tests, 34 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/remote/gateways/remote_trust_commerce_test

Finished in 32.663168 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
16 tests, 59 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
81.25% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/trust_commerce.rb
+++ b/lib/active_merchant/billing/gateways/trust_commerce.rb
@@ -312,9 +312,9 @@ module ActiveMerchant #:nodoc:
       private
 
       def add_aggregator(params, options)
-        if @options[:aggregator_id]
+        if @options[:aggregator_id] || application_id != Gateway.application_id
           params[:aggregators] = 1
-          params[:aggregator1] = @options[:aggregator_id]
+          params[:aggregator1] = @options[:aggregator_id] || application_id
         end
       end
 
--- a/test/unit/gateways/trust_commerce_test.rb
+++ b/test/unit/gateways/trust_commerce_test.rb
@@ -32,6 +32,7 @@ class TrustCommerceTest < Test::Unit::TestCase
   end
 
   def test_succesful_purchase_with_check
+    ActiveMerchant::Billing::TrustCommerceGateway.application_id = 'abc123'
     stub_comms do
       @gateway.purchase(@amount, @check)
     end.check_request do |endpoint, data, headers|


lib/active_merchant/billing/gateways/trust_commerce.rb
```diff
@@ -312,9 +312,9 @@ def scrub(transcript)
       private
 
       def add_aggregator(params, options)
-        if @options[:aggregator_id]
+        if @options[:aggregator_id] || application_id != Gateway.application_id
```
curiousepic: I'm not following why this is `!=`...? Seems like it should just be `=`. This also means the test might not be sufficient. Unless I'm misunderstanding this...

nfarve: There is a default value for all Gateways. I think the value is `ActiveMerchant` or something like that.

nfarve: Well thats what this made me believe: https://github.com/activemerchant/active_merchant/blob/60efe5c5bd27cafa22edb6cdef2273ca7a8da72f/lib/active_merchant/billing/gateways/eway_rapid.rb#L45

But I can't find where its set.

curiousepic: It's set to nil by default https://github.com/activemerchant/active_merchant/blob/6ea353a450fd48c1df7c539e698327bd5ac40e0d/lib/active_merchant/billing/gateway.rb#L142

nfarve: okay so yea there is a default value. We want to only use it if there's a different value.

curiousepic: I sorted out my confusion, but in any case I don't think we need to check `Gateway`'s value.

curiousepic: If it's not different, `application` would already have inherited the default value, right?

nfarve: Right so in that case we **don't** want to set `aggregatorID`

curiousepic: Gah, got it. I don't know why this was so confusing for me but I'm better now. 👍

Issue #3078, opened 2018-12-07
Refactor Spreedly Core to send one request instead of two


lancecarlson:
I noticed that the current Spreedly Core implementation first creates a payment method and then initiates the transaction. Spreedly charges per API call, so I think it would be great to have an implementation that attempts to reduce the amount of calls needed to make a transaction. With the Spreedly API, you can send the payment information directly without creating a payment method first. It will give you back a payment method token if you need it. I actually started to refactor this, and was going to submit a PR, but wanted to make sure the general idea would get accepted before I started to work on it. Thoughts?

PR #3105, opened 2019-01-06
Spreedly one request + Add support for bank accounts

lancecarlson:
Consolidates purchase and authorization requests down to 1.

See issue: https://github.com/activemerchant/active_merchant/issues/3078

### DIFF
--- a/lib/active_merchant/billing/gateways/spreedly_core.rb
+++ b/lib/active_merchant/billing/gateways/spreedly_core.rb
@@ -35,19 +35,13 @@ module ActiveMerchant #:nodoc:
       # Public: Run a purchase transaction.
       #
       # money          - The monetary amount of the transaction in cents.
-      # payment_method - The CreditCard or the Spreedly payment method token.
+      # payment_method - The CreditCard or Check or the Spreedly payment method token.
       # options        - A hash of options:
       #                  :store - Retain the payment method if the purchase
       #                           succeeds.  Defaults to false.  (optional)
       def purchase(money, payment_method, options = {})
-        if payment_method.is_a?(String)
-          purchase_with_token(money, payment_method, options)
-        else
-          MultiResponse.run do |r|
-            r.process { save_card(options[:store], payment_method, options) }
-            r.process { purchase_with_token(money, r.authorization, options) }
-          end
-        end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/spreedly_core.rb
```diff
@@ -155,12 +143,12 @@ def save_card(retain, credit_card, options)
       end
 
       def purchase_with_token(money, payment_method_token, options)
-        request = auth_purchase_request(money, payment_method_token, options)
+        request = build_purchase_request(money, payment_method_token, options)
```
jknipp: @lancecarlson I'd prefer that we retain `auth` in the method name since it is dual purpose. I'm fine with `build_auth_purchase_request`.

lancecarlson: @jknipp What if we consolidated things a bit and made a method like this:

```
      def build_request(money, payment_method, options)
        build_xml_request('transaction') do |doc|
          add_invoice(doc, money, options)
          add_payment_method(doc, payment_method, options)
          retain_on_success(doc, true) if options[:store]
          add_extra_options(:gateway_specific_fields, doc, options)
        end
      end
```

Then we can chop down ```verify_with_token```. If you're cool with this refactoring, I will start getting the tests to pass. :)

jknipp: 👍 Sound good to me, I like the idea of dry'ing up the code here.

lancecarlson:
@jknipp So, some feedback I wanted from you on this last commit:

1) inside of build_request, I omitted the retain_on_success step because it was in add_payment_method already. Do you think it makes sense to keep in here or promote its functionality to build_request?

2) Should we rename build_request to build_transaction_request now or something considering its envelope's name?

jknipp:
@lancecarlson 
1. I like `retain_on_success` in `add_payment_method` as it relates to the payment method. I think the `retain_on_success` flag is only applicable to `card` and `bank account`, not payment method tokens?
2. 👍 on renaming to `build_transaction_request`.

lancecarlson:
@jknipp I just did the rename. I think that should be sufficient for this commit!

PR #3112, opened 2019-01-15
Braintree: Send all country fields

curiousepic:
According to Braintree docs, sending Level 3 data requires the presence
of the shipping country in alpha3 format. Since we can't easily
determine which format should be sent in various situations, default to
sending all formats that are present (which seems permissible, from
remote testing), and then ensure the alpha3 is sent by using the Country
module.

Unit:
57 tests, 150 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
68 tests, 384 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_braintree_blue_test.rb
+++ b/test/remote/gateways/remote_braintree_blue_test.rb
@@ -12,7 +12,7 @@ class RemoteBraintreeBlueTest < Test::Unit::TestCase
 
     @options = {
       :order_id => '1',
-      :billing_address => address(:country_name => 'United States of America'),
+      :billing_address => address(:country_name => 'Canada'),
       :description => 'Store Purchase'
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_braintree_blue_test.rb
```diff
@@ -12,7 +12,7 @@ def setup
 
     @options = {
       :order_id => '1',
-      :billing_address => address(:country_name => 'United States of America'),
+      :billing_address => address(:country_name => 'Canada'),
```
curiousepic: This was giving a "incompatible country data" error so it needed to be in line with the default test card data.

PR #3117, opened 2019-01-17
Orbital - Support for stored credentials framework

jknipp:
Add support for Stored Credentials Framework (Cardholder vs Merchant Initiated transactions).

Update the XSD schema in remote tests to the latest version since we upgraded the API version for stored credentials.

ECS-91

Remote:
25 tests, 148 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
71 tests, 427 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -30,7 +30,7 @@ module Billing #:nodoc:
     class OrbitalGateway < Gateway
       include Empty
 
-      API_VERSION = '7.1'
+      API_VERSION = '7.7'
```
jknipp: Upgrading from 7.1 to 7.7 seems like a big jump in version to me. Any concerns here? Should the upgrade be separate from the credentials update?

test/unit/gateways/orbital_test.rb
```diff
@@ -567,7 +582,7 @@ def test_american_requests_adhere_to_xml_schema
     response = stub_comms do
       @gateway.purchase(50, credit_card, :order_id => 1, :billing_address => address)
     end.check_request do |endpoint, data, headers|
-      schema_file = File.read("#{File.dirname(__FILE__)}/../../schema/orbital/Request_PTI54.xsd")
+      schema_file = File.read("#{File.dirname(__FILE__)}/../../schema/orbital/Request_PTI77.xsd")
```
jknipp: It felt appropriate to update the XSD used for tests to be consistent with the API version update. I left this as a distinct commit in case we decide against doing so.

lib/active_merchant/billing/gateways/orbital.rb
curiousepic: I don't think so, shouldn't be a big deal if there was no change in behavior with tests.

PR #3119, opened 2019-01-18
Map AVS codes for Adyen

therufs:
Adding more mappings to AVSResult makes me a little ⁉️but it didn't look like there were good explicit corresponding options for the codes Adyen is returning.

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -110,24 +110,28 @@ module ActiveMerchant #:nodoc:
 
       AVS_MAPPING = {
         '0'  => 'R',  # Unknown
-        '1'  => 'A',	# Address matches, postal code doesn't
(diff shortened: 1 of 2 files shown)


curiousepic:
It may be possible and desirable to override `AVSResult`'s `MESSAGES` hash to include the Adyen-specific codes and messages...

therufs:
hey @activemerchant/spreedly-connect, 

The possibility of breaking out AVS messages per card provider has been brought up with product, but we think it's outside the scope of this ticket, so the above is what we're going to go with for now: just populating existing AVS codes/messages where they match.

therufs:
Remote: 39 tests, 100 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 26 tests, 127 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

PR #3120, opened 2019-01-21
Braintree: Adds support for transaction_source

molbrown:
Braintree has deprecated the recurring field in lieu of the transaction_source parameter which allows a 'recurring' value. For now, Braintree still allows both fields to be sent.

ECS-97

Remote:
69 tests, 389 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
58 tests, 151 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -585,7 +585,9 @@ module ActiveMerchant #:nodoc:
           parameters[:merchant_account_id] = merchant_account_id
         end
 
-        if options[:recurring]
+        if options[:transaction_source]
+          parameters[:transaction_source] = options[:transaction_source]
+        elsif options[:recurring]
           parameters[:recurring] = true
         end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -589,6 +589,10 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
           parameters[:recurring] = true
         end
 
+        if options[:transaction_source]
+          parameters[:transaction_source] = options[:transaction_source]
+        end
+
```
jknipp: It looks like we are letting customers drive setting the `recurring` and `transaction_source` fields. Is there any concern that we may set `recurring` to `true` and the `transaction_source` would not be `recurring`? Does one field take priority over the other?

molbrown: I could also see an argument for not limiting this, and leaving it to the merchant to use the fields the right way. I don't _think_ this hurts, though.

PR #3126, opened 2019-01-28
Blue Snap: Supports Level 2/3 data

molbrown:
Adds fields to enable Level 2/ Level 3 data, with existing
Nokogiri-Builder request structure for this gateway.

ECS-61

Unit:
20 tests, 76 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
26 tests, 76 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92.3077% passed
Unrelated failures, caused by updated error messages and store functionality.

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -191,6 +191,7 @@ module ActiveMerchant
         doc.send('merchant-transaction-id', truncate(options[:order_id], 50)) if options[:order_id]
         doc.send('soft-descriptor', options[:soft_descriptor]) if options[:soft_descriptor]
         add_description(doc, options[:description]) if options[:description]
+        add_level_3_data(doc, options)
       end
 
       def add_address(doc, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -204,6 +205,41 @@ def add_address(doc, options)
         doc.zip(address[:zip]) if address[:zip]
       end
 
+      def add_level_3_data(doc, options)
+        return unless options[:customer_reference_number]
```
molbrown: Added return condition here to prevent empty `level-3-data` container tag.

curiousepic:
Are Level 2 fields actually part of this?

molbrown:
The difference between Level 2 or 3 (or 2 with TAA) depends on the fields passed, but all are put in the `level-3-data` container - https://developers.bluesnap.com/docs/level-23-data

lib/active_merchant/billing/gateways/blue_snap.rb
jknipp: Going to ask the obvious, can `customer_reference_number` be excluded but 1 or more of the other fields provided, where we should send L3 data elements?

molbrown: Good q- `customer_reference_number` and `sales_tax_amount` are the only fields required by all 3 processing types: Level 3, Level 2, Level 2 with TAA. There are no processing types that are valid without both of those.

molbrown: (Level 3 requires several more in addition)

jknipp: 👍

PR #3127, opened 2019-01-29
Append error messages to Cecabank response

therufs:
Pursuant to [ECS-96](https://spreedly.atlassian.net/browse/ECS-96) and in lieu of https://github.com/spreedly/core/pull/3527.


therufs:
@jknipp Good catch. The concatenation option is covered in `test_invalid_xml_response_handling`, but I've added a check for the error-only text to `test_unsuccessful_request`.

PR #3130, opened 2019-01-31
NMI: Add customer id to authorization on store

curiousepic:
Previously authorizations for the store action were empty. Now we return
the relevant customer_vault_id as the authorization to use in subsequent
transactions.

Remote:
34 tests, 110 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
27 tests, 195 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/nmi.rb
+++ b/lib/active_merchant/billing/gateways/nmi.rb
@@ -144,7 +144,7 @@ module ActiveMerchant #:nodoc:
 
       def add_payment_method(post, payment_method, options)
         if(payment_method.is_a?(String))
-          post[:customer_vault_id] = payment_method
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/nmi.rb
```diff
@@ -144,7 +144,7 @@ def add_invoice(post, money, options)
 
       def add_payment_method(post, payment_method, options)
         if(payment_method.is_a?(String))
-          post[:customer_vault_id] = payment_method
+          post[:customer_vault_id] = split_authorization(payment_method)[0]
```
curiousepic: Changing this to fit the pattern other calls to `split_authorization` use.

PR #3131, opened 2019-02-01
PaymentExpress: Support client_info field

jknipp:
Adds support for the ClientInfo field used to send the IP address.

ECS-124

Unit:
34 tests, 249 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
15 tests, 70 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


jknipp:
Payment Express does not echo the `ClientInfo` field back in the response so that we can verify it in the response. However, it is saved and displayed in the UI as part of the transaction details.

screen_shot_2019-02-19_at_4_57_36_pm

PR #3133, opened 2019-02-01
Trans First Express: Don't pass blank name field

curiousepic:
Remote (Failures probably due to changes to sandbox, relevant remote
test does pass):
32 tests, 83 assertions, 9 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
71.875% passed

Unit:
21 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_trans_first_transaction_express_test.rb
+++ b/test/remote/gateways/remote_trans_first_transaction_express_test.rb
@@ -109,6 +109,32 @@ class RemoteTransFirstTransactionExpressTest < Test::Unit::TestCase
     assert_equal 'Succeeded', response.message
   end
 
+  def test_successful_purchase_without_name
+    credit_card_opts = {
+      :number => 4485896261017708,
+      :month => Date.new((Time.now.year + 1), 9, 30).month,
+      :year => Date.new((Time.now.year + 1), 9, 30).year,
+      :first_name => '',
+      :last_name => ''
+    }
+
+    credit_card = CreditCard.new(credit_card_opts)
+    response = @gateway.purchase(@amount, credit_card, @options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+
+    credit_card_opts = {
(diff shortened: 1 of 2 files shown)


test/remote/gateways/remote_trans_first_transaction_express_test.rb
```diff
@@ -109,6 +109,21 @@ def test_successful_purchase_with_empty_string_cvv
...
+      :month => Date.new((Time.now.year + 1), 9, 30).month,
+      :year => Date.new((Time.now.year + 1), 9, 30).year,
+      :first_name => '',
+      :last_name => ''
+    }
+
+    credit_card = CreditCard.new(credit_card_opts)
+    response = @gateway.purchase(@amount, credit_card, @options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
```
molbrown: Is it worth looking for the name being absent from the request? Or will the sandbox error if you send it blank

molbrown: It could be a unit test, alternatively.

curiousepic: I did manually test that, but probably a good idea.

curiousepic: Added case to same test.

molbrown: I guess I was thinking about verifying that the blank name field is not added to the request. But I wasn't sure if you needed to write that into the test necessarily, or if the sandbox would error in that case.

curiousepic: Ah, I see. Yes, this was in response to the gateway throwing an error when the field was sent as an empty string, so the remote test should be inherently sufficient.

PR #3139, opened 2019-02-09
Worldpay: Adjust use of normalized stored credentials hash

davidsantoso:
Previously the new/beta normalized stored credential hash was using a
boolean value to indicate if a transaction was part of a recurring
transaction or not. However, this did not allow for the case of gateways
(like Worldpay) to take into account transactions that are part of an
installment. This adjusts the previously "recurring" field to use the
name "transaction_type" which accepts a string of either:

- "installment"
- "recurring"
- "unscheduled"

This should allow more flexibility to for other gateways that also have
the notion of a payment that is part of an installment.

Note that this normalized hash is still in "beta" use. The hash
structure itself may change and until there is solid use around it
across multiple gateways, it will likely be a defined-yet-not-enforced
hash rather than an object which provides guarantees around field
values.

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -285,7 +285,11 @@ module ActiveMerchant #:nodoc:
         if options[:stored_credential][:initial_transaction]
           xml.tag! 'storedCredentials', 'usage' => 'FIRST'
         else
-          reason = options[:stored_credential][:recurring] ? 'RECURRING' : 'UNSCHEDULED'
+          reason = case options[:stored_credential][:reason_type]
+                   when 'installment' then 'INSTALMENT'
+                   when 'recurring' then 'RECURRING'
+                   when 'unscheduled' then 'UNSCHEDULED'
+                   end
 
           xml.tag! 'storedCredentials', 'usage' => 'USED', 'merchantInitiatedReason' => reason do
             xml.tag! 'schemeTransactionIdentifier', options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id]
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,6 +26,7 @@
 * Qvalent: Map CVV Result to responses [curiousepic] #3135
 * Card Connect: Handle 401s as responses [curiousepic] #3137
 * Worldpay: Introduce normalized stored credential options [davidsantoso] #3134
+* Worldpay: Adjust use of normalized stored credentials hash [davidsantoso] #3139
 
 == Version 1.90.0 (January 8, 2019)
 * Mercado Pago: Support "gateway" processing mode [curiousepic] #3087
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -285,7 +285,12 @@ def add_stored_credential_using_normalized_fields(xml, options)
         if options[:stored_credential][:initial_transaction]
           xml.tag! 'storedCredentials', 'usage' => 'FIRST'
         else
-          reason = options[:stored_credential][:recurring] ? 'RECURRING' : 'UNSCHEDULED'
+          reason = case options[:stored_credential][:transaction_type]
```
curiousepic: I don't know if I'd use the term "transaction type" for the normalization, seems a little general for this usage. May I suggest "schedule"?

davidsantoso: I agree, `transaction_type` is pretty generic. I looked up a couple of gateways docs and they all had a different term for this 😸The ones I saw were:

* `reason`
* `type`
* `sequence`
* `indicator`

Of those, I think the word `reason` is pretty good. As in, "the _reason_ for this transaction is a recurring monthly subscription". Or going along those same lines the word `purpose` is another good synonym. 

I do think `schedule` still feels a bit vague since I would then assume it's a recurring transaction and the value would be something along the lines of "monthly".

jknipp: I agree `transaction type` is too general in this context - and could be confused with the general transaction type (auth, credit, etc). 

I like `reason`, though I wonder if it would be interpreted as a free form text field versus a using `*_type` as the field name that seems to imply a limited set of types.

jknipp: What values do we expect this field to hold?

davidsantoso: > I wonder if it would be interpreted as a free form text field versus a using `*_type` as the field name that seems to imply a limited set of types.

🤔 ya that's a good point.

> What values do we expect this field to hold?

`recurring`, `unscheduled`, `installment`. Maybe we can call it `reason_type`? 😸

jknipp: Digging into synonyms, do any of these feel more appropriate and encompassing?

* frequency
* cycle
* period

curiousepic: I'm cool with `reason_type`

davidsantoso: @jknipp my only hesitation with those terms is that they imply either that it is an ongoing thing- which it may not be- or that there is a set time frame between transactions.

I think I'm going to go with `reason_type` which hopefully conveys the "reason" for a transaction (i.e. this transaction is happening cause it's a recurring transaction or this transaction is happening because a customer did a one off purchase on our website) with a `_type` suffix to imply there's a limited number of reasons.

jknipp: 👍 Works for me.

PR #3141, opened 2019-02-13
Merrco partial refunds fix

payfirma1:
Changes made to the add_order_id method to send a uuid in place of order number. This is expected to fix the multiple partial refunds problem for Merrco payments.


lib/active_merchant/billing/gateways/netbanx.rb
```diff
@@ -157,8 +157,8 @@ def expdate(credit_card)
         { :month => month, :year => year }
       end
 
-      def add_order_id(post, options)
-        post[:merchantRefNum] = (options[:order_id] || SecureRandom.uuid)
+      def add_order_id(post)
+        post[:merchantRefNum] = (SecureRandom.uuid)
```
pi3r: You don't need the parenthesis here.

payfirma1: The reason for us to make this change is that the PaySafe gateway expects a unique merchantRefNum each time we make a partial refund for an order. In the code the order id was being sent as the merchantRefNum for each refund and this allowed only 1 refund per order (full or partial).

Given that the earlier code change affected multiple line, we have now moved the only change we wanted into its own method. We just needed this functionality in the refunds method, not anywhere else. 

A test case won't be required now because this line of change will only affect the refunds method and as per the existing test cases it's not possible to test the refunds without going through the full transaction and settlement cycle from the gateway.

With this final commit we hope that we can push this change onto the Shopify platform and it won't have any negative affect on any functionality except refunds.

Additionally all existing test cases are being satisfied after this code change and also the automatic build on our branch is passing.

pi3r:
How will you be able to track to which order the refund belongs to?

payfirma1:
As per the Pay safe API documentation and our discussion with their representative we know that the merchantRefNum has to be unique each time we communicate with the API. Also we specifically asked that if the merchantRefNum has to be the order id, to which the answer was no. It can be any random string, as long as it is unique each time.

We tried the same process via their virtual terminal and it lets us perform multiple partial refunds by giving any random merchantRefNum. We are just hoping that what was told to us is correct.

payfirma1:
You think we can proceed with this pull request or is there more you want from our side ?

payfirma1:
Hi. Is there something else you would like us to add in order to proceed with this pull request ?

payfirma1:
Thank you. What is the next step for us to test this functionality ? When will these changes reflect on the production Shopify site ?

pi3r:
We need to merge this and include the new version in our production environment. We will let you know when this is done, after merging.

payfirma1:
Thank you. Will wait for next piece of communication from your side and will then test this functionality.

payfirma1:
Hi. May I please request to get an estimate on the timeline when we can expect this change to get pushed to production ?

pi3r:
@payfirma1 Changes should be live on our platform.

PR #3142, opened 2019-02-15
Fat Zebra: Support voids

curiousepic:
This requires keeping track of actions in order to properly choose the
url for voiding refunds since it is different from refunding purchases.

Remote (1 unrelated failure):
22 tests, 78 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.4545% passed

Unit:
17 tests, 89 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/fat_zebra.rb
+++ b/lib/active_merchant/billing/gateways/fat_zebra.rb
@@ -46,14 +46,17 @@ module ActiveMerchant #:nodoc:
       end
 
       def capture(money, authorization, options = {})
+        txn_id, _ = authorization.to_s.split('|')
         post = {}
+
         add_amount(post, money, options)
         add_extra_options(post, options)
 
-        commit(:post, "purchases/#{CGI.escape(authorization)}/capture", post)
+        commit(:post, "purchases/#{CGI.escape(txn_id)}/capture", post)
       end
 
-      def refund(money, txn_id, options={})
+      def refund(money, authorization, options={})
+        txn_id, _ = authorization.to_s.split('|')
         post = {}
 
         add_extra_options(post, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/fat_zebra.rb
```diff
@@ -141,23 +152,25 @@ def commit(method, uri, parameters=nil)
...
-      def authorization_from(response, success)
+      def authorization_from(response, success, uri)
+        endpoint = uri.split('/')[0]
         if success
-          (response['response']['id'] || response['response']['token'])
+          id = response['response']['id'] || response['response']['token']
+          "#{id}|#{endpoint}"
         else
-          nil
+          ''
```
jknipp: Not something you have to change. I guess you could avoid the `else` and then call `authorization_from(response, success, uri).to_s`. Probably better to be explicit.

curiousepic: calling `.to_s` would produce authorizations of the form `|purchases` etc. But in any case, now that we have `.split` being called on authorizations, it will choke on `nil`. But I suppose I could call `.to_s.split`...

jknipp: Not sure it's worth the effort, think I'd leave it as is for explicitness.

curiousepic: Actually noticed a lot of split-auth implementations use `.to_s.split`, it is more robust at least. I updated that and also some variable naming conventions, if you wouldn't mind giving another once-over.

curiousepic:
@jknipp A better solution for this metadata would definitely be a nice to have, it's one of the things I was going to bring up in standup. But we're already pretty entrenched in it.

lib/active_merchant/billing/gateways/fat_zebra.rb
curiousepic: Oops, missed amending somehow, should be accurate now.

PR #3143, opened 2019-02-15
BlueSnap: Support ACH/ECP payments

jknipp:
Add support for ACH/ECP/echeck payments. We only support vaulting and using 1 ACH account per customer (<= v2.0). Support for multiple ACH accounts (v3.0) was not included at this time.

ECS-60

Unit:
33 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
25 tests, 97 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -170,7 +170,7 @@ module ActiveMerchant
       def add_auth_purchase(doc, money, payment_method, options)
         doc.send('recurring-transaction', options[:recurring] ? 'RECURRING' : 'ECOMMERCE')
         add_order(doc, options)
-        doc.send('storeCard', options[:store_card] || false)
+        doc.send('store-card', options[:store_card] || false)
         add_amount(doc, money, options)
         add_fraud_info(doc, options)
 
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -59,14 +59,27 @@ module ActiveMerchant
         'line1: N, zip: N, name: N' => 'N',
       }
 
+      BANK_ACCOUNT_TYPE_MAPPING = {
+        'personal_checking' => 'CONSUMER_CHECKING',
+        'personal_savings' => 'CONSUMER_SAVINGS',
+        'business_checking' => 'CORPORATE_CHECKING',
+        'business_savings' => 'CORPORATE_SAVINGS'
+      }
+
       def initialize(options={})
         requires!(options, :api_username, :api_password)
         super
       end
 
       def purchase(money, payment_method, options={})
-        commit(:purchase) do |doc|
-          add_auth_purchase(doc, money, payment_method, options)
+        payment_method_details = PaymentMethodDetails.new(payment_method)
+
+        commit(:purchase, :post, payment_method_details) do |doc|
+          if payment_method_details.alt_transaction?
+            add_alt_transaction_purchase(doc, money, payment_method_details, options)
+          else
+            add_auth_purchase(doc, money, payment_method, options)
+          end
         end
       end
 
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -245,6 +273,53 @@ def add_authorization(doc, authorization)
...
+            add_personal_info(doc, check, options)
+            add_echeck_company(doc, check)
+          end
+        end
+
+        doc.send('ecp-transaction') do
+          add_echeck(doc, check) unless vaulted_shopper
+        end
+
+        doc.send('authorized-by-shopper', options[:authorized_by_shopper])
```
jknipp: This field must always be `true` for ACH/ECP and other alternate payments to be accepted. However, BlueSnap requires that the merchant have a checkbox with specific verbiage on their checkout page and require the customer to agree to terms. 

I left it up to our customers to pass this value in to ensure they were meeting the gateway's requirements.

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -380,5 +457,50 @@ def forbidden_response(body)
         { 'description' => body }
       end
     end
+
+    class PaymentMethodDetails
```
curiousepic: Whoa, this is pretty heavy duty. I don't think any other adapters define internal classes for something like this; a `check` payment method class already exists. How necessary is this?

curiousepic: For reference, here is a typical ACH implementation https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/card_connect.rb#L192-L213

jknipp: @curiousepic There is one other place in AM that defines an internal class. I started off keeping track off the payment method details using a hash, however, this quickly became unruly as I had to perform checks multiple places to see if the payment method is an alt transaction and/or if it is a check. For this particular case, creating a class with accessors allowed for better encapsulation of functionality.

Unlike a [straightforward implementation](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/card_connect.rb#L192-L213), Blue Snap's API changes the resource URL, XML root element and other aspects based on if the transaction is a card or an ACH/SEPA/Paypal/etc (`alt transaction`).  What we end up with is needing to classify 4 payment method types 1) String 2) String, but storing an Alt payment method (ACH) 3) Check, 4) [Default] Credit card.

My first pass required values to be passed repeatedly to different methods - because `url` and `root_element` are deeply nested in `commit` I had to pass data from method to method to end up with `if-elsif-else` statements replacing the ternary form `action == :store ? 'vaulted-shopper' : 'card-transaction'`. 

I'd prefer to not use an internal class but felt it was much cleaner to do so because of the encapsulation. I'm happy to pair and walk through these changes.

curiousepic: I seeeee, got it. Just wanted to make sure it was justified and sounds like it is. Thanks for the details.

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -380,5 +457,50 @@ def forbidden_response(body)
...
+
+      def initialize(payment_method = nil)
+        @payment_method = payment_method
+        @payment_method_type = nil
+        parse(payment_method)
+      end
+
+      def check?
+        @payment_method_type == 'check'
+      end
```
jknipp: @curiousepic I'm to going to make one small update: 

```
def check?
  @payment_method.is_a(Check) || @payment_method_type == 'check'
end
```

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -324,13 +399,15 @@ def message_from(succeeded, parsed_response)
...
+        action == :store ? vaulted_shopper_id(parsed_response, payment_method_details) : parsed_response['transaction-id']
       end
 
-      def vaulted_shopper_id(parsed_response)
+      def vaulted_shopper_id(parsed_response, payment_method_details)
         return nil unless parsed_response['content-location-header']
-        parsed_response['content-location-header'].split('/').last
+        vaulted_shopper_id = parsed_response['content-location-header'].split('/').last
+        vaulted_shopper_id += "|#{payment_method_details.payment_method_type}" if payment_method_details.alt_transaction?
+        vaulted_shopper_id
```
curiousepic: I might just be confused, but shouldn't this mean that for a purchase with a stored card, the token is added to the request as `[id]|card`? I'd think it would need to get split out, but apparently the `test_successful_purchase_using_stored_card` remote test passes?

curiousepic: Oh, no, because it's not an `alt_transaction?` on store. NM!

curiousepic:
But nice work wrangling some complicated dependencies into a semblance of order!

jknipp:
A yet to be realized benefit of introducing the internal class is that adding other alternate payment types should be easy (support the new type in the class, add supporting XML changes).

PR #3144, opened 2019-02-18
Concatenate card and customer token

therufs:
Pinpayments handles stored cards by assigning them to a user. It seems that if
a card is associated with a user the card must be used with the customer ID
rather than the card's, and since a card can only be associated with one user,
it's not clear that the card ID generated when a card is stored will ever
actually be reusable (see the [Customer API docs](https://pinpayments.com/developers/api-reference/customers#put-customer) since cards are irretrievable once they are replaced.

The possible use case for stored card IDs is that customers may have multiple
cards if one is set as primary. Other stored cards may not be able to be used
until/unless they are set as primary.

Remote tests: 16 tests, 46 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit tests: 37 tests, 112 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/pin.rb
+++ b/lib/active_merchant/billing/gateways/pin.rb
@@ -67,6 +67,7 @@ module ActiveMerchant #:nodoc:
       # Updates the credit card for the customer.
       def update(token, creditcard, options = {})
         post = {}
+        token = get_customer_token(token)
 
         add_creditcard(post, creditcard)
         add_customer_data(post, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -206,7 +215,11 @@ def unparsable_response(raw_response)
       end
 
       def token(response)
-        response['token']
+        if response['token'].start_with?('cus')
+          "#{response.dig('card', 'token')}+#{response['token']}"
```
curiousepic: There's no real standardization, but unless you think `+` is particularly suited for this gateway, let's use one of the more common separators, either `;` or `|`.

curiousepic: Also, since prior tokens only consisted of the customer token, that should be the first element so that splitting old tokens will still parse them correctly.

jknipp: +1 for using a common separator.

therufs: 👍

PR #3146, opened 2019-02-19
Detect Discover cards that are longer than 16 digits in length

prashcr:
The current regex does not detect Discover cards that are longer than 16 digits, but their compliance guide states that cards may be between 16-19 digits. [Reference](https://www.discoverglobalnetwork.com/downloads/IPP_VAR_Compliance.pdf)

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -5,7 +5,7 @@ module ActiveMerchant #:nodoc:
       CARD_COMPANY_DETECTORS = {
         'visa'               => ->(num) { num =~ /^4\d{12}(\d{3})?(\d{3})?$/ },
         'master'             => ->(num) { num&.size == 16 && in_bin_range?(num.slice(0, 6), MASTERCARD_RANGES) },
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -104,7 +104,7 @@ def valid_start_year?(year)
       # master:           http://www.mastercard.com/ca/merchant/en/getstarted/Anatomy_MasterCard.html
       # jcb:              http://www.jcbcard.com/security/info.html
       # diners_club:      http://www.dinersclub.com/assets/DinersClub_card_ID_features.pdf
-      # discover:         https://www.discover.com/credit-cards/help-center/glossary.html
+      # discover:         https://www.discoverglobalnetwork.com/downloads/IPP_VAR_Compliance.pdf
```
jknipp: @prashcr The card number length change looks good, will you rollback the change to the URL?

prashcr:
Sure, I'll do that soon.

jknipp:
Thanks @prashcr! Merged in https://github.com/activemerchant/active_merchant/commit/dc9fb68868f82085f42fbe84ee491be465643358

PR #3147, opened 2019-02-21
Bambora Asia-Pacific: Adds Store

molbrown:
ECS-159

Unit:
8 tests, 50 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
17 tests, 32 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/bambora_apac.rb
+++ b/lib/active_merchant/billing/gateways/bambora_apac.rb
@@ -3,8 +3,8 @@ require 'nokogiri'
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class BamboraApacGateway < Gateway
-      self.live_url = 'https://www.bambora.co.nz/interface/api/dts.asmx'
-      self.test_url = 'https://demo.bambora.co.nz/interface/api/dts.asmx'
+      self.live_url = 'https://www.bambora.co.nz/interface/api'
+      self.test_url = 'https://demo.bambora.co.nz/interface/api'
 
       self.supported_countries = ['AU', 'NZ']
       self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :jcb]
@@ -32,7 +32,7 @@ module ActiveMerchant #:nodoc:
             xml.CustRef options[:order_id]
             add_amount(xml, money)
             xml.TrnType '1'
-            add_credit_card(xml, payment)
+            add_payment(xml, payment)
             add_credentials(xml, options)
             xml.TrnSource options[:ip]
           end
@@ -45,7 +45,7 @@ module ActiveMerchant #:nodoc:
             xml.CustRef options[:order_id]
             add_amount(xml, money)
             xml.TrnType '2'
-            add_credit_card(xml, payment)
+            add_payment(xml, payment)
             add_credentials(xml, options)
             xml.TrnSource options[:ip]
           end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/bambora_apac.rb
```diff
@@ -82,6 +82,26 @@ def void(money, authorization, options={})
         end
       end
 
+      def verify(credit_card, options={})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { purchase(200, credit_card, options) }
```
curiousepic: I assume 200 is it the minimum? If not, try 0.

molbrown: I wanted to, but I could not get 100 or 0 to work. Voids only work on purchases, not auths, and it won't accept a $0 purchase. It seems that 100 is an amount that generates an error for tests ("declined_message"=>"Refer Card Issuer"), though that did not seem apparent in the docs. https://dev-apac.bambora.com/checkout/guides/get-up-and-running/test-your-setup

curiousepic: 🤔 I wonder if an un-voided auth is preferable...

molbrown: I was having the same thought, it would be lame if the void fails, then there are random $2 purchases 😬

molbrown: Updated to a 0 un-voided auth.

curiousepic: @molbrown Sorry for leaving this out initially, but there may be other reasons un-voided auths are undesirable. It should be fine if they are automatically canceled after a short time, but I can't tell if this is the case from their Docs. Any insight on that?

curiousepic:
Not a big deal but it might be nice to pull out the endpoint (`sipp` or `dts`) into a method and set them as  a variable in the urls instead of splitting the relevant actions into an if/else that duplicates the actions. Up to you.

test/remote/gateways/remote_bambora_apac_test.rb
```diff
@@ -81,6 +81,46 @@ def test_failed_void
...
+    assert_failure response
+  end
+
+  def test_successful_purchase_using_stored_card
+    assert store_response = @gateway.store(@credit_card, @options)
+    assert_success store_response
+
+    response = @gateway.purchase(500, store_response.authorization, @options)
+    assert_success response
+    assert_equal '', response.message
```
curiousepic: This is a little weird. We usually set the response's `message` to something like "Success".

lib/active_merchant/billing/gateways/bambora_apac.rb
molbrown: I have contacted Bambora support about this, and have decided to remove `verify` from this commit and proceed without it. It will be added in once there is more information about how to best implement.

molbrown:
Updated with some re-factoring around the endpoint determination, removes verify, adds Succeeded to `message_from`.

lib/active_merchant/billing/gateways/bambora_apac.rb
```diff
@@ -150,36 +178,48 @@ def new_submit_xml(action)
...
               end
             end
           end
         end
         xml.target!
       end
 
+      def define_endpoint(action)
+        action == 'TokeniseCreditCard' ? '/sipp' : '/dts'
+      end
```
jknipp: This won't hold up approval, just something to consider. I think I would name the function `endpoint` and avoid including the beginning forward slash in the return values - it was not immediately apparent when reviewing the URL's where it was or how it was included.

```
def endpoint(action)
  action == 'TokeniseCreditCard' ? 'sipp' : 'dts'
end

# Example usage
response = parse(ssl_post("#{commit_url}/#{endpoint(action)}.asmx", new_submit_xml(action, &block), headers))
```

curiousepic: ➕

molbrown: ✅

PR #3148, opened 2019-02-21
Orbital: Pass normalized stored credential fields

curiousepic:
Remote: 27 tests, 152 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 72 tests, 432 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

I'd particularly appreciate @jknipp and @davidsantoso's eyes on this one. Orbital's expectations for message/reason type are not straightforward https://www.dropbox.com/s/532uhzsx5mhydb5/OrbitalGatewayXMLInterfaceSpecification%207.7.pdf?dl=0 (section 3.3)

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -507,9 +507,33 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_credentials(xml, parameters)
-        xml.tag! :MITMsgType, parameters[:mit_msg_type] if parameters[:mit_msg_type]
-        xml.tag! :MITStoredCredentialInd, parameters[:mit_stored_credential_ind] if parameters[:mit_stored_credential_ind]
-        xml.tag! :MITSubmittedTransactionID, parameters[:mit_submitted_transaction_id] if parameters[:mit_submitted_transaction_id]
+        return unless parameters[:mit_stored_credential_ind] == 'Y' || parameters[:stored_credential] && !parameters[:stored_credential].values.all?(&:nil?)
+        if msg_type = get_msg_type(parameters)
+          xml.tag! :MITMsgType, msg_type
+        end
+        xml.tag! :MITStoredCredentialInd, 'Y'
+        if parameters[:mit_submitted_transaction_id]
+          xml.tag! :MITSubmittedTransactionID, parameters[:mit_submitted_transaction_id]
+        elsif parameters.dig(:stored_credential, :network_transaction_id) && parameters.dig(:stored_credential, :initiator) == 'merchant'
+          xml.tag! :MITSubmittedTransactionID, parameters[:stored_credential][:network_transaction_id]
+        end
+      end
+
+      def get_msg_type(parameters)
+        return parameters[:mit_msg_type] if parameters[:mit_msg_type]
+        return 'CSTO' if parameters[:stored_credential][:initial_transaction]
+        return unless parameters[:stored_credential][:initiator] && parameters[:stored_credential][:reason_type]
+        initiator = case parameters[:stored_credential][:initiator]
+        when 'customer' then 'C'
+        when 'merchant' then 'M'
+        end
+        reason = case parameters[:stored_credential][:reason_type]
+        when 'recurring' then 'REC'
+        when 'installment' then 'INS'
+        when 'unscheduled' then 'USE'
+        end
+
+        "#{initiator}#{reason}"
       end
 
       def parse(body)
--- a/test/remote/gateways/remote_orbital_test.rb
+++ b/test/remote/gateways/remote_orbital_test.rb
@@ -157,6 +157,54 @@ class RemoteOrbitalGatewayTest < Test::Unit::TestCase
     assert_equal 'Approved', response.message
   end
 
+  def test_successful_purchase_with_normalized_mit_stored_credentials
+    stored_credential = {
+      stored_credential: {
+        initial_transaction: false,
+        initiator: 'merchant',
+        reason_type: 'unscheduled',
+        network_transaction_id: 'abcdefg12345678'
+      }
+    }
+
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(stored_credential))
+
+    assert_success response
+    assert_equal 'Approved', response.message
+  end
+
+  def test_successful_purchase_with_normalized_cit_stored_credentials
+    stored_credential = {
+      stored_credential: {
+        initial_transaction: true,
+        initiator: 'customer',
+        reason_type: 'unscheduled'
+      }
+    }
+
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(stored_credential))
+
+    assert_success response
+    assert_equal 'Approved', response.message
+  end
+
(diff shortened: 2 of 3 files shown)


jknipp:
@bayprogrammer You will have similar changes for https://spreedly.atlassian.net/browse/ECS-138, taking advantage of the [Stored Credential framework](https://paper.dropbox.com/doc/Spreedlys-Stored-Credential-Framework--AXMhFGUz7ffMjQCioqCfDMHnAg-szo9uWDNHTIFqEvFugOpo) built by @davidsantoso

davidsantoso:
@curiousepic looks good to me in terms of using the "defined" `stored_credential` hash in the options parameter. I didn't look into Orbital's exact field requirements but the general idea of adding them based on the values in the `stored_credential` hash is 💯

curiousepic:
I had the thought that, since Orbital has more `msgType` values than the ones implemented here for the situations Spreedly will using, it may be desirable to allow the GSFs to override the normalized hash when provided, rather than the logic dictating that only either the normalized hash or the GSFs are used... what do we think about that?

jknipp:
@curiousepic I like that idea, but I know that @davidsantoso made updates to the framework expand support for cases such as Orbital. I can't answer if what we have covers all cases. I think for Orbital there are at least 2 message types for customer and merchant that are not covered.

davidsantoso:
> I think for Orbital there are at least 2 message types for customer and merchant that are not covered.

Gotcha. So, I guess this where where the "beta" side of the `stored_credential` hash is clear. I surveyed 3 gateways and the bases were covered there. But, I suspect that there's a fair number of variations still out there. This is also where it seems like a structured object of some sort would come in handy.

curiousepic:
Updated with GSFs overriding the normalized hash, as well as a unique msg_type for the initial transaction that I missed the first go round.

davidsantoso:
@curiousepic for Worldpay I had a top level method which determined if it should use the `stored_credentials` hash or very specific fields in the options. 

https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/worldpay.rb#L276

Wondering if something like that might be a little easier to grok?

curiousepic:
@davidsantoso That wouldn't allow piece-wise overriding, which was my primary reasoning behind overriding at all, in order to utilize the normalized hash for everything except message types that the normalized `reason_type` doesn't cover.

davidsantoso:
ohh hmm 🤔 I hadn't thought about mixing and matching- seems like it could get complex real quick. I'll defer to you on allowing specific field overrides vs. continuing to add to the `stored_credential` hash possible values.

curiousepic:
Hmm, I wonder how common these other types are; and honestly it's difficult to tell whether they are proper reason types, or just a situational combination of all the factors that the gateway wants us to specify. They are:

> **Cardholder-Initiated General Transaction**
> Signifies cardholder-initiated transactions that can support various merchant-initiated transactions.
> Example: A cardholder-initiated authorization that the merchant may later use to request a re-authorization.

There is virtually no other information on this type. I don't see how it's different from an initial Cardholder-Initiated Transaction (which is distinct from an initial cardholder-initiated unscheduled, recurring, or installment), unless it needs to be specified explicitly  in order to allow Merchant-Initiated Reauthorizations. But in that case, how does it fall within the recurring/unscheduled/installment designation? And if so, should this just be the default for initial transactions?

> **Merchant-Initiated Resubmission**
> Signifies when a merchant performs a resubmission in cases where the merchant requested an authorization, but received a decline after the merchant has already delivered the goods or services to the cardholder.
> Example: A merchant receives a decline due to insufficient funds after already shipping goods and performs a resubmission transaction to obtain an approval.

This... seems like a headache to handle either in the hash or in AM logic.

> **Merchant-Initiated Reauthorization**
> Signifies a merchant-initiated re-authorization.
> Example: Merchant receives an order for several items totaling a specific amount. The items are shipped separately and split shipment transactions are created for each shipment.

Headache.

I think these are not true reason types, with the possible exception of "general". So I don't think there's a place for them in the hash, but if a user wants to utilize these, overriding with GSFs allows that, and if they don't care about them, they're free to go all-hash. Bad field combinations should return a useful error message from the gateway, so I don't feel too bad about allowing mixing and matching.

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -507,9 +507,33 @@ def add_managed_billing(xml, options)
       end
 
       def add_stored_credentials(xml, parameters)
-        xml.tag! :MITMsgType, parameters[:mit_msg_type] if parameters[:mit_msg_type]
-        xml.tag! :MITStoredCredentialInd, parameters[:mit_stored_credential_ind] if parameters[:mit_stored_credential_ind]
-        xml.tag! :MITSubmittedTransactionID, parameters[:mit_submitted_transaction_id] if parameters[:mit_submitted_transaction_id]
+        return unless parameters[:mit_stored_credential_ind] && parameters[:mit_stored_credential_ind] == 'Y' || parameters[:stored_credential] && !parameters[:stored_credential].values.all?(&:nil?)
```
jknipp: Can we simplify 
`parameters[:mit_stored_credential_ind] && parameters[:mit_stored_credential_ind] == 'Y'` 
to
`parameters[:mit_stored_credential_ind] == 'Y'`?

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -507,9 +507,33 @@ def add_managed_billing(xml, options)
...
+        if msg_type = get_msg_type(parameters)
+          xml.tag! :MITMsgType, msg_type
+        end
+        xml.tag! :MITStoredCredentialInd, 'Y'
+        if parameters[:mit_submitted_transaction_id]
+          xml.tag! :MITSubmittedTransactionID, parameters[:mit_submitted_transaction_id]
+        elsif parameters[:stored_credential] && parameters[:stored_credential][:network_transaction_id] && parameters[:stored_credential][:initiator] && parameters[:stored_credential][:initiator] == 'merchant'
```
jknipp: Likewise, can we simplify
`parameters[:stored_credential][:initiator] && parameters[:stored_credential][:initiator] == 'merchant'`
to
`parameters.dig(:stored_credential, :initiator) == 'merchant'` ?

curiousepic: Nice, I think so; I didn't properly understand `.dig` when I went back and implemented it, and missed this refactor.

lib/active_merchant/billing/gateways/orbital.rb
curiousepic: I'm less sure about this, I'll test it out with `.dig`.

jknipp: I don't think we need `.dig` because we aren't dealing with a nested hash. The result of `parameters[:mit_stored_credential_ind]` if the key `mit_stored_credential_ind` is missing is `nil`.

curiousepic: You're right; I had tried it that way previously and it failed for some reason, no forgotten, but this is good.

PR #3151, opened 2019-02-25
Adds ELO card type

deedeelavinder:
This adds ELO bin ranges and ELO card brand logic in general.
It also adds `elo` as a supported cardtype for the Adyen gateway.

Unit Tests
27 tests, 132 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications 100% passed

Remote Tests
45 tests, 122 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications 100% passed


dtykocki:
@deedeelavinder - It looks like there are a good number of what appear to be legit failures on CI. We should probably get those fixed up before reviewing in detail.

deedeelavinder:
@dtykocki Yes, they are all the same failure. 😬 A collision with Discover. _back to the drawing board_

PR #3153, opened 2019-02-26
Adds ELO card type

deedeelavinder:
This adds ELO bin ranges and ELO brand logic in general. ELO is in a
partnership with Discover and is part of the Discover Global Network.
Using the ELO BIN ranges before the Discover regex allows those cards
to be identified correctly.

It also adds `elo` as a supported cardtype for the Adyen gateway.

Unit Tests
27 tests, 132 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications 100% passed

Remote Tests
45 tests, 122 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications 100% passed


prashcr:
Related question for maintainers:
Is it worth distinguishing card network vs card brand?

**Current**
`brand? == discover` includes Discover cards, PayPal cards, China UnionPay cards and Elo cards
`diners_club, jcb` are considered as different card brands, even though they're also on the Discover network.

We don't handle partnerships/co-branded cards well

**Ideas**

I think there are users like me who primarily care about which network the card is on. E.g. I want to make sure cards on the Discover network are sent to a suitable processor.

Typically in the case of co-branded cards and partnerships, the cards are designated to be processed on one of the networks in particular (e.g. Diners/Mastercard as Mastercard, China UnionPay/Discover as Discover, Visa/Dankort as Visa etc.)

We could break up the current Discover range into its constituent brands, and provide a separate `network?` method to check the card network.

PR #3154, opened 2019-02-27
[WIP] Fix discover network bin ranges

prashcr:
* Diners Club: add new "39" range
* Diners Club: only "36" range has a minimum of 14 digits
* Diners Club: change min/max for all other ranges to 16/19 digits
* JCB: update regex to support 19 digits

Taken from https://www.discoverglobalnetwork.com/downloads/IPP_VAR_Compliance.pdf


deedeelavinder:
This causes a collision with a `Maestro` number.

PR #3155, opened 2019-02-27
Ecs 138 vantiv litle stored credential

bayprogrammer:
Trying the draft pull request feature. I have a little bit more I want to do before making this a real pull request (expand/refine the remote new remote tests for stored credentials in Litle), but I *think* I'm close. Thought I'd push up what I have for some early feedback.

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -3,7 +3,7 @@ require 'nokogiri'
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class LitleGateway < Gateway
-      SCHEMA_VERSION = '9.12'
+      SCHEMA_VERSION = '9.14'
 
       self.test_url = 'https://www.testvantivcnp.com/sandbox/communicator/online'
       self.live_url = 'https://payments.vantivcnp.com/vap/communicator/online'
@@ -223,6 +223,7 @@ module ActiveMerchant #:nodoc:
         add_descriptor(doc, options)
         add_merchant_data(doc, options)
         add_debt_repayment(doc, options)
+        add_stored_credential_params(doc, options)
       end
 
       def add_merchant_data(doc, options={})
@@ -293,6 +294,38 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def add_stored_credential_params(doc, options={})
+        return unless options[:stored_credential]
+
+        if options[:stored_credential][:initial_transaction]
+          add_stored_credential_params_initial(doc, options)
+        else
+          add_stored_credential_params_used(doc, options)
+        end
+      end
+
+      def add_stored_credential_params_initial(doc, options)
+        case options[:stored_credential][:reason_type]
+        when 'unscheduled'
+          doc.processingType('initialCOF')
+        when 'installment'
+          doc.processingType('initialInstallment')
+        when 'recurring'
+          doc.processingType('initialRecurring')
+        end
+      end
+
+      def add_stored_credential_params_used(doc, options)
+        if options[:stored_credential][:reason_type] == 'unscheduled'
+          if options[:stored_credential][:initiator] == 'merchant'
+            doc.processingType('merchantInitiatedCOF')
+          else
+            doc.processingType('cardholderInitiatedCOF')
+          end
+        end
+        doc.originalNetworkTransactionId(options[:stored_credential][:network_transaction_id])
+      end
+
       def add_billing_address(doc, payment_method, options)
         return if payment_method.is_a?(String)
 
@@ -332,8 +365,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_order_source(doc, payment_method, options)
-        if options[:order_source]
-          doc.orderSource(options[:order_source])
+        order_source = order_source(options)
+        if order_source
+          doc.orderSource(order_source)
         elsif payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source == :apple_pay
           doc.orderSource('applepay')
         elsif payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source == :android_pay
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -223,6 +223,55 @@ def add_auth_purchase_params(doc, money, payment_method, options)
...
+          doc.processingType("initialCOF")
+        when 'installment'
+          doc.processingType("initialInstallment")
+        when 'recurring'
+          doc.processingType("initialRecurring")
+        end
+      end
+
+      def add_stored_credential_params_used(doc, options)
+        network_transaction_id = options[:stored_credential].fetch(:network_transaction_id)
```
curiousepic: I have not often used `fetch`, is it particularly useful here?

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -332,6 +381,10 @@ def add_address(doc, address)
       end
 
       def add_order_source(doc, payment_method, options)
+        # TODO(zmd): mutating options like this feels wrong... ponder the error
+        #   of my ways further before committing to this...
+        order_source_options_for_stored_credential(options)
+
         if options[:order_source]
```
curiousepic: Yeah let's have this be something like `if options[:order_source] || parameters[:stored_credential] && !parameters[:stored_credential].values.all?(&:nil?)` and munge together the condititionals from your `order_source_options_for_stored_credential` with the below.

test/remote/gateways/remote_litle_test.rb
```diff
@@ -231,6 +231,42 @@ def test_authorization_capture_refund_void
     assert_equal 'Approved', void.message
   end
 
+  def test_authorization_capture_with_stored_credential
```
curiousepic: Minor quibble, we usually refer to this as `authorize_and_capture`, partly not to confuse it with a response's `authorization`.

test/remote/gateways/remote_litle_test.rb
```diff
@@ -231,6 +231,42 @@ def test_authorization_capture_refund_void
...
+    initial_stored_creds = {
+      stored_credential: {
+        initial_transaction: true,
+        reason_type: 'unscheduled',
```
curiousepic: We should probably support initial transactions with a nil `reason_type`

curiousepic:
Brought up a few considerations but overall this looks great!

lib/active_merchant/billing/gateways/litle.rb
bayprogrammer: Yah, I didn't like mutating the options hash to achieve the necessary result.

I think I've found a good solution, at least one that has the same desired effect (setting orderSource under certain conditions when stored_credential is being passed, but still ensuring logic falls back to default for CIT used cases, without mutating the options hash directly).

https://github.com/activemerchant/active_merchant/pull/3155/commits/2ccd672242a4743ae39f1c53fbdc6340337bfa51

Please let me know if my proposed solution misses something that you were seeing here in my original hack. Thanks!

test/remote/gateways/remote_litle_test.rb
bayprogrammer: So the stored credential spec I'm implementing from seems to indicate that even for the initial transaction you need to set `processingType` to one of `initialRecurring`, `initialInstallment`, or `initialCOF`.

I'm not sure how we can satisfy that requirement if we don't have the user pass in the reason type even for that initial transaction. (See the spec PDF attached to ECS-138, page 2). Am I missing something there?

lib/active_merchant/billing/gateways/litle.rb
bayprogrammer: Yah, this was one of those soft-blocker questions I alluded to yesterday. What `Hash#fetch` does is get the value at the given key (first argument to `#fetch`), and if that key wasn't found raise an exception. My wanting to explicitly not support calling this method without providing the `:network_transaction_id` is why I used `fetch` here. (This method only gets called if the user is asking to use existing stored credentials, and if they're doing that they should have a network transaction id).

(As a side-note, `#fetch` is also handy if you do want the key to be optional but want to explicitly set the default value to something other than the Hash instance's default when the key isn't provided; I didn't want that here so I didn't pass the second argument to `#fetch`).

BUT, I'm not actually sure if we want to raise an exception here or not. From the perspective of the user of AM, should we alert the user immediately if they're not passing the needed data? Also, if we do still want to raise here, what exception would be appropriate to raise? (Something specific to ActiveMerchant?)

curiousepic: Very good questions that I'm not sure I have an easy answer for. Generally, we have leaned away from raising exceptions and toward allowing the gateway to provide whatever error they throw after receiving the request. But this might be a good candidate for standardization along with normalizing stored credentials. `ArgumentError` might be the closest fit.

test/remote/gateways/remote_litle_test.rb
curiousepic: Gotcha, that's fine. I don't think so, just trying to keep things permissive.

lib/active_merchant/billing/gateways/litle.rb
bayprogrammer: I'll go ahead and not raise in this case, to better match what we're already doing. I don't want to break with convention/consistency needlessly. Thanks for your feedback, it was very helpful!

test/remote/gateways/remote_litle_test.rb
bayprogrammer: Ok, cool. I don't think I'm actually enforcing this in the gateway adapter, so we should still be OK for being permissive (though like with not submitting the network_transaction_id, not submitting the reason_type will likely cause the gateway itself to raise some kind of error). Thanks again!

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -332,7 +365,9 @@ def add_address(doc, address)
       end
 
       def add_order_source(doc, payment_method, options)
-        if options[:order_source]
+        if stored_cred_order_source = order_source_for_stored_credential(options)
```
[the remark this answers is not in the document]
bayprogrammer: @jknipp Good call on assignment in a conditional. I like your simplification and will implement that presently. Thank you!

PR #3157, opened 2019-03-04
Adyen: Idempotency for non-purchase requests

molbrown:
Support of Idempotency-Key for all actions except purchase, due to
multi-key requirement (design still pending).

ECS-166

Unit:
29 tests, 139 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
43 tests, 120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -37,6 +37,7 @@ module ActiveMerchant #:nodoc:
           authorize(money, payment, options)
         else
           MultiResponse.run do |r|
+            options[:idempotency_key] = nil
             r.process { authorize(money, payment, options) }
             r.process { capture(money, r.authorization, options) }
           end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -84,12 +85,13 @@ def store(credit_card, options={})
...
         add_address(post, options)
-        commit('authorise', post)
+        commit('authorise', post, options)
       end
 
       def verify(credit_card, options={})
         MultiResponse.run(:use_first_response) do |r|
           r.process { authorize(0, credit_card, options) }
+          options[:idempotency_key] = nil
           r.process(:ignore_result) { void(r.authorization, options) }
```
jknipp: We set the idempotency_key to nil here because we've already used it in the `authorize`? I'm guessing we have to do this because the void will fail because of a duplicate key.

molbrown: Yes, it can't have the key already used in the auth. Look at remote test `test_verify_with_idempotency_key` to see how I checked repeating voids on auths that were already canceled. I'm also not as concerned about this as purchase, as it's a $0-amount, and there's not a lot of hurt in repeat-voids. This could be a big benefit, as re-trying verify without idempotency might raise a fraud flag at the bank for hitting the card with multiple authorizes.

jknipp: Makes sense 👍

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -326,7 +338,22 @@ def test_successful_verify
...
+    assert_equal 'Authorised', response.message
+    first_auth = response.authorization
+
+    response = @gateway.verify(@credit_card, options)
+    assert_success response
+    assert_equal response.authorization, first_auth
+
+    response = @gateway.void(first_auth, @options)
+    assert_success response
   end
```
jknipp: Nice!

PR #3160, opened 2019-03-05
FirstdataE4: Rename to FirstdataE4V11

curiousepic:
In preparation to rename FirstdataE4V27 to FirstdataE4, making the names
accurately refer to the support gateway versions.

Unit:
32 tests, 157 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
No currently working test creds.

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4_v11.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4_v11.rb
@@ -1,6 +1,6 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
-    class FirstdataE4Gateway < Gateway
+    class FirstdataE4V11Gateway < Gateway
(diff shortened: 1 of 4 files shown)


curiousepic:
Inquiring about current test creds, but thought I'd throw this up for early scrutiny.

curiousepic:
We would also want to cut a minor version before merging this.

curiousepic:
@jknipp Good question; we could certainly have taken that path when implementing v28, but it probably doesn't make sense to move to it retroactively in this case.

jknipp:
@curiousepic Are we ok to close this?

curiousepic:
@jknipp This fell through the cracks, though we might still be waiting on test creds for confirmation? In any case it's something that should still happen.

curiousepic:

PR #3161, opened 2019-03-05
Correctly process Adyen risk_data option

bayprogrammer:


bayprogrammer:
@molbrown I noticed the failing tests wrt 'Refused' vs 'CVC Declined', but that you had fixed those in your recent Adyen commit that you rolled back. Should I go ahead and include those fixes in my branch here, or leave them for when your revisions get merged back in?

molbrown:
Those updates weren't related to my ticket, just an error message update. Ok for you to add the change 👍

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -167,10 +167,11 @@ def add_extra_data(post, payment, options)
...
 
       def add_risk_data(post, options)
-        risk_data = {}
-        risk_data.merge!(options[:risk_data]) if options[:risk_data]
+        risk_data = options.fetch(:risk_data, {})
+        return nil if risk_data.empty?
+        risk_data = Hash[risk_data.map { |k, v| ["riskdata.#{k}", v] }]
 
-        post[:additionalData][:riskData] = risk_data unless risk_data.empty?
+        post[:additionalData].merge!(risk_data)
```
bayprogrammer: It occurs to me that this method can be further simplified since we're not creating a sub-hash under `additionalData`. It makes sense to exit early and avoid creating an empty sub-hash, but if we're merging directly into `additionalData` I'm pretty sure it's neither more efficient than simply merging the hash (even if it's empty). I'm going to tweak this a bit and see if I can't simplify. Won't take much effort to do.

curiousepic:
Looking good. Don't forget a commit message and test run output.

curiousepic:
@bayprogrammer Ah, not a big deal, I'm just used to reviewing theoretically final commits (this was not submitted as a draft AFAICT), and changes due to review being amended.

bayprogrammer:
@curiousepic Yah, didn't do as a draft PR this time. I'll squash and include the proper commit message for future PRs (and will do that shortly for this one). Thanks for the heads up on our process! :)

jknipp:
@bayprogrammer The standard process is to include test run output as reassurance that tests were executed. It also serves as an indicator that the PR is ready to be reviewed and merged once approved.

This is especially true in Core where the absence of test output in a PR is a code smell.

bayprogrammer:
I've properly prepared my commit (squashed, with test run information included). I also tweaked `#add_risk_data` per my comment earlier.

The remote tests that were failing yesterday due to mismatched expectations with the expected failure message string suddenly stopped failing, so I did not change the remote test as I had expected I would.

lib/active_merchant/billing/gateways/adyen.rb
bayprogrammer: Would the following be preferable to avoid needless instantiation of a new `Hash` when not needed?

```
      def add_risk_data(post, options)
        if (risk_data = options[:risk_data])
          risk_data = Hash[risk_data.map { |k, v| ["riskdata.#{k}", v] }]
          post[:additionalData].merge!(risk_data)
        end
      end
```

bayprogrammer: @curiousepic When you have a moment, would you give me your opinion on this? I don't want to try to micro-optimize here, but when I thought about it more it seemed more logical to just skip all of it at least if the user doesn't pass any `risk_data` option at all. Since you already approved, I don't want to sneak in a change after-the-fact without running it by you first. Thanks!

curiousepic: Sure, that's reasonable.

PR #3165, opened 2019-03-07
FirstData e4 v27: Support v28 url and stored creds

curiousepic:
Involves updating response parsing to handle nested elements.

Remote:
26 tests, 110 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
31 tests, 146 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


curiousepic:
I was originally going to wait to push this until we renamed the two adapters, but I think it's safe to go ahead and make this update before that happens. Here's the PR for the first step of renaming: https://github.com/activemerchant/active_merchant/pull/3160

PR #3166, opened 2019-03-11
Base.mode is supports thread local variable assignment

lancecarlson:


lancecarlson:
This is only supported in Rails 5+. Perhaps it makes sense to start supporting 5+ at some point?

lancecarlson:
I can add the active support file into the code and require it if the rails version is older. Is this desired?

PR #3170, opened 2019-03-12
Realex: Add 3DS

AnnaGyergyai:
Add support for ECI, XID and CAVV on the XML request for 3D Secure. Following: https://developer.globalpay.com/#!/api/3d-secure-two#3ds2-authorization

```xml
  <mpi>
    <cavv>AAACBllleHchZTBWIGV4AAAAAAA=</cavv>
    <xid>crqAeMwkEL9r4POdxpByWJ1/wYg=</xid>
    <eci>5</eci>
  </mpi>
```

New remote test passes:
```
Loaded suite test/remote/gateways/remote_realex_test
Started
.
Finished in 0.991542 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 3 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1.01 tests/s, 3.03 assertions/s
```

### DIFF
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -145,7 +145,11 @@ module ActiveMerchant
           add_card(xml, credit_card)
           xml.tag! 'autosettle', 'flag' => auto_settle_flag(action)
           add_signed_digest(xml, timestamp, @options[:login], sanitize_order_id(options[:order_id]), amount(money), (options[:currency] || currency(money)), credit_card.number)
-          add_network_tokenization_card(xml, credit_card) if credit_card.is_a?(NetworkTokenizationCreditCard)
+          if credit_card.is_a?(NetworkTokenizationCreditCard)
+            add_network_tokenization_card(xml, credit_card)
+          else
+            add_three_d_secure(xml, options)
+          end
           add_comments(xml, options)
           add_address_and_customer_info(xml, options)
         end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -145,7 +145,11 @@ def build_purchase_or_authorization_request(action, money, credit_card, options)
           add_card(xml, credit_card)
           xml.tag! 'autosettle', 'flag' => auto_settle_flag(action)
           add_signed_digest(xml, timestamp, @options[:login], sanitize_order_id(options[:order_id]), amount(money), (options[:currency] || currency(money)), credit_card.number)
-          add_network_tokenization_card(xml, credit_card) if credit_card.is_a?(NetworkTokenizationCreditCard)
+          if credit_card.is_a?(NetworkTokenizationCreditCard)
+            add_network_tokenization_card(xml, credit_card)
+          else
```
elfassy: Do we need this `else`? We check `if options[:three_d_secure]` in `add_three_d_secure` and there may be cases were we'll want to pass this option for a network tokenization card.

AnnaGyergyai: I think yes, there's already some logic for cavv and eci, but it's pulled from payment for tokenized cards: https://github.com/activemerchant/active_merchant/blob/8535adc31d6e4285924b8418b61699aee1424fe7/lib/active_merchant/billing/gateways/realex.rb#L279

PR #3173, opened 2019-03-13
QuickPay v10: Add 3DS params

mdeloupy:
Add support for 3DSecure for QuickPay v10. 

From the documentation : https://learn.quickpay.net/tech-talk/api/services/#POST-payments--id-authorize---format-


elfassy:
failing test: `Prefer single-quoted strings when you don't need string interpolation or special symbols`

mdeloupy:
Fixed :-)

PR #3175, opened 2019-03-14
Add 3DS parameters to Worldpay

nebdil:
Add parameters that we receive after doing 3DS with an external MPI in order to pass them to Worldpay.

Related documentation: https://beta.developer.worldpay.com/docs/wpg/reference/3ds2withexternalmpi

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -270,6 +270,16 @@ module ActiveMerchant #:nodoc:
               xml.tag! 'session', 'shopperIPAddress' => options[:ip] if options[:ip]
               xml.tag! 'session', 'id' => options[:session_id] if options[:session_id]
             end
+
+            if three_d_secure = options[:three_d_secure]
+              xml.tag! 'info3DSecure' do
+                xml.tag! 'threeDSVersion', three_d_secure[:version]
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -270,6 +270,16 @@ def add_payment_method(xml, amount, payment_method, options)
               xml.tag! 'session', 'shopperIPAddress' => options[:ip] if options[:ip]
               xml.tag! 'session', 'id' => options[:session_id] if options[:session_id]
             end
+
+            if three_d_secure = options[:three_d_secure]
+              xml.tag! 'info3DSecure' do
+                xml.tag! 'threeDSVersion', three_d_secure[:version]
```
elfassy: doc states:
```
threeDSVersion
Possible Values:
1.0.2
2.1.0
```
how do we plan to handle this? The version passed in the options might be different, say `2.2.0`

filipebarcos: I wonder what really matters, if it's more like 1 vs. 2, or the actual number matters

filipebarcos: I'm confirming this with Worldpay, but rather not block this PR.

nfarve:
Do you have a special parameter to take advantage of these changes? The two remote tests are failing for me.

PR #3176, opened 2019-03-18
Allow to pass empty `landing_page` parameter

PhilibertDugas:
PayPal has logic in their javascript to determine if a visitor should see the sign-in page or the guest checkout page.

This kicks in only when the `landing_page` parameter is not sent and the logic is based off cookies.

### DIFF
--- a/lib/active_merchant/billing/gateways/paypal_express.rb
+++ b/lib/active_merchant/billing/gateways/paypal_express.rb
@@ -146,9 +146,9 @@ module ActiveMerchant #:nodoc:
               xml.tag! 'n2:cpp-payflow-color', options[:background_color] unless options[:background_color].blank?
               if options[:allow_guest_checkout]
                 xml.tag! 'n2:SolutionType', 'Sole'
-                xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing'
-              elsif options[:paypal_chooses_landing_page]
-                xml.tag! 'n2:SolutionType', 'Sole'
+                unless options[:paypal_chooses_landing_page]
+                  xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing'
+                end
               end
               xml.tag! 'n2:BuyerEmail', options[:email] unless options[:email].blank?
 
--- a/lib/active_merchant/billing/gateways/paypal_express.rb
+++ b/lib/active_merchant/billing/gateways/paypal_express.rb
@@ -147,6 +147,8 @@ module ActiveMerchant #:nodoc:
               if options[:allow_guest_checkout]
                 xml.tag! 'n2:SolutionType', 'Sole'
                 xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing'
+              elsif options[:paypal_chooses_landing_page]
+                xml.tag! 'n2:SolutionType', 'Sole'
               end
               xml.tag! 'n2:BuyerEmail', options[:email] unless options[:email].blank?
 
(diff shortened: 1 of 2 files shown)


PhilibertDugas:
@pi3r @filipebarcos

lib/active_merchant/billing/gateways/paypal_express.rb
```diff
@@ -147,6 +147,8 @@ def build_setup_request(action, money, options)
               if options[:allow_guest_checkout]
                 xml.tag! 'n2:SolutionType', 'Sole'
                 xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing'
+              elsif options[:paypal_chooses_landing_page]
```
filipebarcos: Would make sense to leverage the `allow_guest_checkout` for this?

PhilibertDugas: sure, maybe I can pass `paypal_chooses_landing_page` too and that would result in

`xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing' unless options[:paypal_chooses_landing_page` ?

lib/active_merchant/billing/gateways/paypal_express.rb
```diff
@@ -146,7 +146,7 @@ def build_setup_request(action, money, options)
               xml.tag! 'n2:cpp-payflow-color', options[:background_color] unless options[:background_color].blank?
               if options[:allow_guest_checkout]
                 xml.tag! 'n2:SolutionType', 'Sole'
-                xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing'
+                xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing' unless options[:paypal_chooses_landing_page]
```
Krystosterone: Whatever the param, it's hard to understand what the behaviour of the code is because of the way it's written.

PhilibertDugas:
> Do I understand correctly in that passing in "Login" as the `landing_page` wouldn't solve the problem?

That's right. When the `landing_page` parameter is not set at all, PayPal will decide between `Login` and `Billing` based on the browser's information. From what I got, they tweak this algorithm to optimize the buyer's journey.

Snapshot of the conversation I had:

> Philibert Dugas [12:15 PM]
>> remove  landing_page= all together and let the PayPal cookie or Non-Cookied user allow PayPal to decide to show either Login or Billing Page

> How is that done? Removing the param would trigger a fallback mechanism based on the cookies?

> PayPal [12:18 PM]
> Yes. If the user has a PayPal cookie on their device we would display the login page. If the user doesn’t have a cookie we would show the billing screen expanded, but also have an option for a existing PayPal user to login in case they removed their cookies. Make senSe?


> There is a lot of logic/experiments our teams do to show what page and when. But the above is what it would look like with or without a cookie
> And if they were using PayPal OneTouch it just auto logs them in.

Krystosterone:
Thanks for that context! In that case, I'd just slightly tweak the code.

lib/active_merchant/billing/gateways/paypal_express.rb
```diff
@@ -146,7 +146,9 @@ def build_setup_request(action, money, options)
               xml.tag! 'n2:cpp-payflow-color', options[:background_color] unless options[:background_color].blank?
               if options[:allow_guest_checkout]
                 xml.tag! 'n2:SolutionType', 'Sole'
-                xml.tag! 'n2:LandingPage', options[:landing_page] || 'Billing'
+                unless options[:paypal_chooses_landing_page]
```
PhilibertDugas: @Krystosterone is this version more readable?

PhilibertDugas: I opted for a new parameter as opposed to changing the behaviours of existing params (like touching `:landing_page`) because this could affect other integrations outside of our own

Krystosterone: I would also have agreed to adding parenthesis to it:
```
xml.tag!('n2:LandingPage', options[:landing_page] || 'Billing') unless options[:paypal_chooses_landing_page]
```

PR #3177, opened 2019-03-19
Updated Nab Transaction gateway periodic test url

mengqing:
The periodic test url has updated on the gateway. This PR is for updating that


jknipp:
👍 Thanks @mengqing!

Fixes 6 remote test failures, 3 remote test failures are unrelated.

PR #3179, opened 2019-03-22
Adyen: Fix adding phone from billing address

curiousepic:
Now checking for presence of a billing_address in options before
adding the phone.

Remote:
52 tests, 149 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
30 tests, 144 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -152,7 +152,7 @@ module ActiveMerchant #:nodoc:
       }
 
       def add_extra_data(post, payment, options)
-        post[:telephoneNumber] = options[:billing_address][:phone] if options[:billing_address][:phone]
+        post[:telephoneNumber] = options[:billing_address][:phone] if options.dig(:billing_address, :phone)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -152,7 +152,9 @@ def scrub(transcript)
       }
 
       def add_extra_data(post, payment, options)
-        post[:telephoneNumber] = options[:billing_address][:phone] if options[:billing_address][:phone]
```
molbrown: Would this simplify by using:
post[:telephoneNumber] = options[:billing_address][:phone] if options.dig( :billing_address, :phone)

curiousepic: Yep! Updated.

curiousepic:
Closed with 477453d6449e2dc8c98e5944a960dfe9b0e26c42

PR #3180, opened 2019-03-22
Fix partial or missing address exceptions

molbrown:
NilClass errors noted in attempted transactions for DLocal and Mundipagg gateways,
due to absent objects. Made small fixes that will avoid this error.

Mundipagg Unit tests:
17 tests, 76 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

DLocal Unit tests:
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/mundipagg.rb
+++ b/lib/active_merchant/billing/gateways/mundipagg.rb
@@ -128,8 +128,8 @@ module ActiveMerchant #:nodoc:
       def add_shipping_address(post, options)
         if address = options[:shipping_address]
           post[:address] = {}
-          post[:address][:street] = address[:address1].match(/\D+/)[0].strip if address[:address1]
-          post[:address][:number] = address[:address1].match(/\d+/)[0] if address[:address1]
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -128,8 +128,8 @@ def add_billing_address(post, type, options)
       def add_shipping_address(post, options)
         if address = options[:shipping_address]
           post[:address] = {}
-          post[:address][:street] = address[:address1].match(/\D+/)[0].strip if address[:address1]
-          post[:address][:number] = address[:address1].match(/\d+/)[0] if address[:address1]
+          post[:address][:street] = address[:address1].match(/\D+/)[0].strip if address[:address1].match(/\D+/) && address[:address1]
+          post[:address][:number] = address[:address1].match(/\d+/)[0] if address[:address1].match(/\d+/) && address[:address1]
```
jknipp: Should the check that `address[:address1]` is truthy occur before the match?

molbrown: Updated.

PR #3181, opened 2019-03-24
Migrate CheckoutV2Gateway integration to Unified Payments Api and add 3DS support with third party MPI

nicolas-maalouf-cko:
* Migrate the CheckoutV2Gateway to the Unified Payments API
* Update the matching unit tests
* Add 3DS Support with third party MPI along with matching unit tests

Checkout.com Unified Payments API Reference:
https://api-reference.checkout.com/


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -81,43 +82,59 @@ def scrub(transcript)
...
-        post[:transactionIndicator] = options[:transaction_indicator] || 1
-        post[:previousChargeId] = options[:previous_charge_id] if options[:previous_charge_id]
+        post[:card_on_file] = true if options[:card_on_file] == true
+        post[:payment_type] = 'Regular' if options[:transaction_indicator] == 1
+        post[:payment_type] = 'Recurring' if options[:transaction_indicator] == 2
+        post[:previous_payment_id] = options[:previous_charge_id] if options[:previous_charge_id]
+      end
+
+      def add_3ds(post, options)
+        return unless options[:execute_threed] || options[:threed_dynamic]
```
filipebarcos: Since you don't really use `threed_dynamic` value, I'd just drop its dependency. Also, if we can rely on the presence of `three_d_secure` key, would be a bit more aligned with other gateways that already support the param pass thru.
```ruby
{
  three_d_secure: {
    cavv: 'cavv value goes here',
    eci: 'eci value',
    xid: 'xid value',
  }
}
```

then this method would be something like
```suggestion
def add_3ds(post, options)
  if options[:three_d_secure]
    post[:'3ds'] = {}
    post[:'3ds'][:enabled] = true
    post[:'3ds'][:eci] = options[:eci] if options[:eci]
    post[:'3ds'][:cryptogram] = options[:cavv] if options[:cavv]
    post[:'3ds'][:xid] = options[:xid] if options[:xid]
  end
end
```

filipebarcos: a `return unless options[:three_d_secure]` would work too

PR #3182, opened 2019-03-26
Adyen: Update to support normalized stored credential fields

molbrown:
Adapter already supports defining shopperInteraction and
recurringProcessingModel directly. Adding ability to define these fields
using the normalized hash (which can be overidden by passing the fields
directly). Maintains using verification value or Network Tokenization
cards to define shopperInteraction.

ECS-213

Unit:
32 tests, 152 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
52 tests, 149 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -49,7 +49,7 @@ module ActiveMerchant #:nodoc:
         add_invoice(post, money, options)
         add_payment(post, payment)
         add_extra_data(post, payment, options)
-        add_shopper_interaction(post, payment, options)
+        add_stored_credentials(post, payment, options)
         add_address(post, options)
         add_installments(post, options) if options[:installments]
         add_3ds(post, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -177,8 +178,13 @@ def add_risk_data(post, options)
...
       end
 
+      def add_stored_credentials(post, payment, options)
+        add_shopper_interaction(post, payment, options)
+        add_recurring_processing_model(post, options)
+      end
+
       def add_shopper_interaction(post, payment, options={})
-        if (payment.respond_to?(:verification_value) && payment.verification_value) || payment.is_a?(NetworkTokenizationCreditCard)
+        if options.dig(:stored_credential, :initial_transaction) || (payment.respond_to?(:verification_value) && payment.verification_value) || payment.is_a?(NetworkTokenizationCreditCard)
```
curiousepic: Could `.dig` replace the second condition?

molbrown: I wish! `payment` is not a hash, so it cannot.

curiousepic: orite

PR #3183, opened 2019-03-29
DLocal: Send country as string

molbrown:
ECS-246

Unit:
100% passed

Remote:
21 tests, 59 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


molbrown:
This is actually not a great test. You only see the country object if you check `post[:country]` as the request is built, even though the tests appear to send the country string. Not sure why this wasn't exposed in tests, but the country object is also seen in the production transcripts.

PR #3185, opened 2019-04-02
Support stored credentials in Cybersource

therufs:
There are still a few failing remote tests; these seem to fail in the current version as well.

### DIFF
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -425,6 +425,62 @@ class CyberSourceTest < Test::Unit::TestCase
     assert_success response
   end
 
+  def test_successful_auth_first_unscheduled_stored_cred
+    @gateway.stubs(:ssl_post).returns(successful_authorization_response)
+    @options[:stored_credential] = {
+      :initiator => 'cardholder',
+      :reason_type => 'unscheduled',
+      :initial_transaction => true,
+      :network_transaction_id => ''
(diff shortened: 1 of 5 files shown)


therufs:
Reference: 
http://apps.cybersource.com/library/documentation/dev_guides/CC_Svcs_SO_API/Credit_Cards_SO_API.pdf (p. 206)
https://ics2wsa.ic3.com/commerce/1.x/transactionProcessor/CyberSourceTransaction_1.153.xsd

test/unit/gateways/cyber_source_test.rb
```diff
@@ -425,6 +425,62 @@ def test_successful_auth_with_network_tokenization_for_amex
...
+
+  def test_successful_auth_first_recurring_stored_cred
+    @gateway.stubs(:ssl_post).returns(successful_authorization_response)
+    @options[:stored_credential] = {
+      :initiator => 'cardholder',
+      :reason_type => 'recurring',
+      :initial_transaction => true,
+      :network_transaction_id => ''
+    }
+    assert response = @gateway.authorize(@amount, @credit_card, @options)
```
przemyslawhaptas: @therufs as I understand it, one of the points of the first request is to obtain the `network_transaction_id` to be used with subsequent ones, correct? Unfortunately, it seems to not be the case, at least not with my cybersource account. Would you mind checking if the response has the field on your side?

PR #3190, opened 2019-04-04
Adyen: Support adjust action

curiousepic:
This adds support for Adyen's adjustAuthorization action, which can
change the amount of a prior authorization, either increasing or
decreasing the amount and extending the auth period. There is no other
precedent for this type of action in ActiveMerchant, but it seems very
straightforward and simple to support, similar to other followup actions
like refunds and voids.

Remote:
55 tests, 163 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
34 tests, 161 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -305,6 +305,36 @@ class RemoteAdyenTest < Test::Unit::TestCase
     assert_equal 'Original pspReference required for this operation', response.message
   end
 
+  def test_successful_adjust
+    authorize = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success authorize
+
+    assert adjust = @gateway.adjust(200, authorize.authorization)
+    assert_success adjust
+    assert_equal '[adjustAuthorisation-received]', adjust.message
+  end
+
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -305,6 +305,36 @@ def test_failed_void
...
+
+  def test_successful_adjust_and_capture
+    authorize = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success authorize
+
+    assert adjust = @gateway.adjust(200, authorize.authorization)
+    assert_success adjust
+    assert_equal '[adjustAuthorisation-received]', adjust.message
+
+    assert capture = @gateway.adjust(200, authorize.authorization)
```
molbrown: should this be `@gateway.capture`?

curiousepic: Good catch, I think I undo-ed too far after fixing that...

curiousepic: (updated)

PR #3192, opened 2019-04-09
Adyen: Upgrade to v40 API

davidsantoso:
The primary driver behind this is to ensure that we can use this
adapater for 3DS 2.0 transactions given the appropriate fields are in
place. However, considering this is quite a major jump in version
support it deserves it's own commit.

Note that the two remote tests that have been changed are because the
Adyen API no longer seems to return errors for a blank state or country
in the billing address. However I should note that looking at the Adyen
API reference for billingAddress, it appears as though country is
required. Though that may be a typo in their documentation.

https://docs.adyen.com/api-explorer/#/Payment/v40/authorise

Remote Test Run:
➜ ruby -Itest test/remote/gateways/remote_adyen_test.rb
Started
.......................................................

Finished in 62.034313 seconds.
55 tests, 161 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

0.89 tests/s, 2.60 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -4,8 +4,8 @@ module ActiveMerchant #:nodoc:
 
       # we recommend setting up merchant-specific endpoints.
       # https://docs.adyen.com/developers/api-manual#apiendpoints
-      self.test_url = 'https://pal-test.adyen.com/pal/servlet/Payment/v18'
-      self.live_url = 'https://pal-live.adyen.com/pal/servlet/Payment/v18'
+      self.test_url = 'https://pal-test.adyen.com/pal/servlet/Payment/v40'
+      self.live_url = 'https://pal-live.adyen.com/pal/servlet/Payment/v40'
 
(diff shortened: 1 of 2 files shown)


davidsantoso:
> This is quite a version jump, any specifics we should watch out for? Breaking changes, unexpected features?

Unfortunately the best I have is that our remote tests are passing. In their new docs site there's an API explorer where you can toggle the version number, but even that only goes back to v31. 

> Does Adyen publish a changelog?

Also unfortunately no. I've searched everywhere for one but I don't think they do publish one.

The [Adyen docs say this](https://beta-docs.adyen.com/development-resources/live-endpoints/#api-versioning), which is very not comforting:

> While we endeavor to maintain backwards compatibility, we version the API when a change will create compatibility issues.  We maintain older versions of the API which you can continue to use until you are ready to upgrade. 

So I'm not sure what that could mean for this. I was hoping y'all might have more insight into the Adyen adapter from previous experiences using it that could lead to other things we might want to test?

curiousepic:
Heh... Only now seeing that the docs indicate `adjustAuthorisation` wasn't supported on v18, but the sandbox didn't complain when I tested it. Welp.

I'm not sure I have any useful input, but seems alright, just need to keep a close eye on things when we upgrade. Might want to ping Shopify folks on this as well.

molbrown:
^ @curiousepic's observation illustrates how permissive Adyen's sandbox tends to be. Passing tests will not necessarily indicate that the update won't break things as they are implemented currently. 😬 I guess best to watch closely and be prepared for ripple effects requiring more updates. Unfortunately, I can't think of anything obvious to watch for.

davidsantoso:
> Only now seeing that the docs indicate adjustAuthorisation wasn't supported on v18

So double win if this goes smoothly?

> can't think of anything obvious to watch for.

👍 thanks!

PR #3199, opened 2019-04-28
add deviceFingerprintID to cybersource payment gateway

apokalipto:
Some countries require the deviceFingerprintID tag in order to get production keys


rmdydo:
Is there a reason why this was closed? I am looking to add deviceFingerprintID as well.

PR #3200, opened 2019-04-29
Adyen: Correct formatting of Billing Address

nfarve:
With the upgrade of Adyen to version 40, the billing address was removed
from the card object. This corrects the format and returns the two tests
that fail when you pass an incorrect country or state. In order to see
AVS results you may need to make some configuration changes to your
account.

https://docs.adyen.com/developers/api-reference/payments-api#paymentresultadditionaldata

........................................................

56 tests, 166 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Started

34 tests, 161 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -15,6 +15,15 @@ class RemoteAdyenTest < Test::Unit::TestCase
       :brand => 'visa'
     )
 
+    @avs_credit_card = credit_card('4400000000000008',
+      :month => 10,
+      :year => 2020,
+      :first_name => 'John',
+      :last_name => 'Smith',
+      :verification_value => '737',
+      :brand => 'visa'
+    )
+
     @elo_credit_card = credit_card('5066 9911 1111 1118',
       :month => 10,
       :year => 2020,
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -504,13 +531,15 @@ def test_missing_state_for_purchase
   def test_blank_country_for_purchase
     @options[:billing_address][:country] = ''
     response = @gateway.authorize(@amount, @credit_card, @options)
-    assert_success response
+    assert_failure response
+    assert_match Gateway::STANDARD_ERROR_CODE[:incorrect_address], response.error_code
```
curiousepic: Do empty strings now send an empty string, or an `'N/A'`?

curiousepic: (and should that behavior change, rather than the test?)

curiousepic: Well, I guess it doesn't matter if the real behavior change is in what the gateway requires, which seems to be the case.

nfarve: empty strings raise an error. Not sending a values sends a `'N/A'`

nfarve: This returns the test to the state it was before the upgrade.

davidsantoso: > This returns the test to the state it was before the upgrade.

Makes sense that these tests were updated to be incorrect. When I did the upgrade it changed to succeeded because the billing address was no longer being sent where it was expected and thus none of it was present and there was nothing to error on. In hindsight that should've indicated that I needed to double check any changes to where the `billingAddress` hash was added.

curiousepic: 👍s all around.

PR #3201, opened 2019-04-29
Elavon: Send ssl_transaction_currency if not USD

gcatlin:
Elavon returns an error for any API request that includes a currency
UNLESS the merchant's terminal is setup for multi-currency. Previously
we NEVER sent the currency, so only the default currency worked. Now, we
only send the `ssl_transaction_currency` field if the currency is not
the default currency (USD).

Elavon Multi-Currency Conversion (MCC) documentation:
https://developer.elavon.com/#/api/eb6e9106-0172-4305-bc5a-b3ebe832f823.rcosoomi/documents?converge-integration-guide/book/processing_options/../../book/processing_options/mcc.html

ECS-272

Closes #3201 

Unit:
29 tests, 141 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
24 tests, 105 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 * Checkout V2: Checkout V2: Correct success criteria [curiousepic] #3205
 * Adyen: Add normalized hash of 3DS 2.0 data fields from web browsers [davidsantoso] #3207
 * Stripe: Do not attempt application fee refund if refund was not successful [jasonwebster] #3206
+* Elavon: Send transaction_currency if currency is provided [gcatlin] #3201
 
 == Version 1.93.0 (April 18, 2019)
 * Stripe: Do not consider a refund unsuccessful if only refunding the fee failed [jasonwebster] #3188
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -42,6 +42,7 @@ module ActiveMerchant #:nodoc:
         else
           add_creditcard(form, payment_method)
         end
+        add_currency(form, money, options)
         add_address(form, options)
         add_customer_data(form, options)
         add_test_mode(form, options)
@@ -54,6 +55,7 @@ module ActiveMerchant #:nodoc:
         add_salestax(form, options)
         add_invoice(form, options)
         add_creditcard(form, creditcard)
+        add_currency(form, money, options)
         add_address(form, options)
         add_customer_data(form, options)
         add_test_mode(form, options)
@@ -69,6 +71,7 @@ module ActiveMerchant #:nodoc:
           add_approval_code(form, authorization)
           add_invoice(form, options)
           add_creditcard(form, options[:credit_card])
+          add_currency(form, money, options)
           add_customer_data(form, options)
           add_test_mode(form, options)
         else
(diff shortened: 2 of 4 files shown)


gcatlin:
### Concerns / Questions
 - There should probably be tests added to this PR. I'm sorry I didn't get that far, but I ran out of time.
 - The credentials in `fixtures.yml` are bad. I found new creds in the Elavon dev portal, but there are 2 different sets of creds: one that support multi-currency and one that does not. Which of these (or both?) should be added to `fixtures.yml`?
 - Some gateways include a list of supported currencies. Should valid currencies be added to the Elavon gateway? They provide a list:
    > - The alphanumeric code that must be included in the request to indicate the currency in which you wish to process. If omitted, the terminal default currency is assumed (for example: USD, CAD, and JPY).
    > - More than 94 currencies are supported. Refer to the [ISO Currency Codes](https://developer.elavon.com/#/api/eb6e9106-0172-4305-bc5a-b3ebe832f823.rcosoomi/documents?converge-integration-guide/book/transaction_types/credit_card/../../../book/api_reference/iso_currency_codes.html) section for an extensive list of available currencies.
    - I'm also concerned that the default currency is set per-terminal, so the assumption of USD may be a bad idea.
    - Also, the list of ISO Currency Codes contains *more* that 94 entries, so hard-coding the supported currencies may be asking for trouble.
- From the [Elavon Multi-Currency Conversion documentation](https://developer.elavon.com/#/api/eb6e9106-0172-4305-bc5a-b3ebe832f823.rcosoomi/documents?converge-integration-guide/book/processing_options/../../book/processing_options/mcc.html):
    > - Once a terminal is setup for Multi-Currency, the terminal can process by default the following currencies: US Dollar (USD), Canadian Dollar (CAD), Pound Sterling (GBP), Australian Dollar (AUD), Euro (EUR) and Japanese (YEN). You can add or remove currencies in the Converge user interface.
    > - Multi-Currency is supported with MasterCard and Visa transactions only. Other card types cannot be sent with a transaction other than merchant currency: USD or CAD.
    - The complexities underlying those bullet points have not been captured in the pull request, but perhaps should?
 - The following also have not been accounted for, but perhaps AM handles this automatically? (From [Elavon Multi-Currency Conversion documentation](https://developer.elavon.com/#/api/eb6e9106-0172-4305-bc5a-b3ebe832f823.rcosoomi/documents?converge-integration-guide/book/processing_options/../../book/processing_options/mcc.html)):
    > - When specifying amounts, be sure to submit the correct number of decimal places for the transaction currency. All currencies support two exponents (decimals,) other than the following: ...
    > - For Japanese Yen, there are no currency exponents after the decimal place. Any numbers included after the decimal place will be ignored.
    > - For the Bahraini Dinar, there are 3 possible digits after the decimal place. Your website may wish to take advantage of the precision this audience of customers will expect. However, currently the Converge will accommodate only 2 decimals. If you submit three, it will automatically round your transaction to two decimals. (for example: 1.235 will round to 1.23 and 1.236 will round to 1.24).

jknipp:
Revisting this, I think we should send the `ssl_transaction_currency` GSF if `options[:currency_code]` is provided and leave the determination of _**when**_ it should be provided up to the user. The user knows what their Elavon default currency is and if Multi-Currency is enabled, and therefore when to send the non-default currency.

jknipp:
@activemerchant/spreedly-connect Ready for re-review.

PR #3205, opened 2019-05-01
Checkout V2: Correct success criteria

curiousepic:
Also corrects address field value detection.

Unit:
22 tests, 99 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unfortunately we don't currently have working test credentials for remote tests

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -111,12 +111,12 @@ module ActiveMerchant #:nodoc:
         address = options[:billing_address]
         if(address && post[:source])
           post[:source][:billing_address] = {}
-          post[:source][:billing_address][:address_line1] = address[:address1] if address[:address1]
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -215,7 +215,7 @@ def parse(body)
       end
 
       def success_from(response)
-        response['response_summary'] == 'Approved' || response.key?('action_id')
+        response['response_summary'] == 'Approved' || !response.key?('response_summary') && response.key?('action_id')
```
rwdaigle: We should use `()` here to make it clear the precedence of these conditions vs. having to rely solely on our knowledge of Ruby's rules around `||` and `&&`?

(obviously something to change once we've resolved any immediate issues)

nfarve: I don't think it's just ruby. I think the general order of precedence for logic operators is ~, &, ||. But `()` would make it more clear.

PR #3210, opened 2019-05-08
Elavon: Multi-currency support

jknipp:
Send the currency in the ssl_transaction_currency field only if the
user has specified they are using multi-currency. The multi-currency
flag can be set with either a gateway or transaction level field.

Using multi-currency depends on the merchant's multi-currency terminal
setting at Elavon.

ECS-272

Unit:
31 tests, 150 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
26 tests, 115 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_elavon_test.rb
+++ b/test/remote/gateways/remote_elavon_test.rb
@@ -3,6 +3,7 @@ require 'test_helper'
 class RemoteElavonTest < Test::Unit::TestCase
   def setup
     @gateway = ElavonGateway.new(fixtures(:elavon))
+    @multi_currency_gateway = ElavonGateway.new(fixtures(:elavon_multi_currency))
 
     @credit_card = credit_card('4124939999999990')
     @bad_credit_card = credit_card('invalid')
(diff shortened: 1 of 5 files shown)


test/remote/gateways/remote_elavon_test.rb
```diff
@@ -205,8 +206,17 @@ def test_successful_purchase_with_custom_fields
...
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(currency: 'USD', multi_currency: true))
+
+    assert_failure response
+    assert response.test?
+    assert_equal 'Transaction currency is not allowed for this terminal.  Your terminal must be setup with Multi currency', response.message
+    assert response.authorization
+  end
+
+  def test_successful_purchase_with_multi_currency
+    assert response = @multi_currency_gateway.purchase(@amount, @credit_card, @options.merge(currency: 'JPY', multi_currency: true))
```
curiousepic: Do we want to test the gateway-level and option-level flags separately?

jknipp: I think that's a holdover from before I changed to use a gateway-level flag. I'll add a new test.

PR #3212, opened 2019-05-09
Adyen: Support preAuths and Synchronous Adjusts

curiousepic:
Allows passing of authorisationType PreAuth to enable Adjusts to be made
on Auths, and allows passing of adjustAuthorisationData from responses
to support Synchronous Adjusts - this requires Adyen to set your account
to Synchronous Adjust mode.

Remote:
60 tests, 187 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
37 tests, 179 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -80,6 +80,7 @@ module ActiveMerchant #:nodoc:
         post = init_post(options)
         add_invoice_for_modification(post, money, options)
         add_reference(post, authorization, options)
+        add_extra_data(post, nil, options)
         commit('adjustAuthorisation', post, options)
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -396,8 +400,10 @@ def success_from(action, response)
         case action.to_s
         when 'authorise', 'authorise3d'
           ['Authorised', 'Received', 'RedirectShopper'].include?(response['resultCode'])
-        when 'capture', 'refund', 'cancel', 'adjustAuthorisation'
+        when 'capture', 'refund', 'cancel'
           response['response'] == "[#{action}-received]"
+        when 'adjustAuthorisation'
+          response['response'] == 'Authorised' || response['response'] == '[adjustAuthorisation-received]'
```
curiousepic: the first condition is true for sync mode, the second for default/async.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -174,6 +175,9 @@ def add_extra_data(post, payment, options)
         post[:additionalData][:overwriteBrand] = normalize(options[:overwrite_brand]) if options[:overwrite_brand]
         post[:additionalData][:customRoutingFlag] = options[:custom_routing_flag] if options[:custom_routing_flag]
         post[:additionalData]['paymentdatasource.type'] = NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
+        post[:additionalData][:authorisationType] = options[:authorisation_type] if options[:authorisation_type]
+        post[:additionalData][:adjustAuthorisationData] = options[:adjust_authorisation_data] if options[:adjust_authorisation_data]
+        post[:additionalData][:RequestedTestAcquirerResponseCode] = options[:requested_test_acquirer_response_code] if options[:requested_test_acquirer_response_code] && test?
```
curiousepic: This allows us to explicitly invoke arbitrary response codes; actually a pretty nice sandbox feature compared to others.

molbrown:
For async, is there a follow-up message after the received Adjust succeeds? 🤔

curiousepic:
Not built-in. The same is true for refunds that are technically async; we still return the confirmation of receipt as a success.

curiousepic:
To clarify, Adyen does send a notification via their own callback system.

PR #3213, opened 2019-05-09
Worldpay: Support Unknown Card Type

tanyajajodia:


tanyajajodia:
@dtykocki for your initial review and feedback

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -24,6 +24,8 @@ class WorldpayGateway < Gateway
         'elo'              => 'ELO-SSL'
       }
 
+      UNKNOWN_CARD_CODE = 'CARD-SSL'
```
dtykocki: I think it might be more consistent with the existing code to place `CARD-SSL` into the map above. Perhaps with the key "unknown".

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -252,7 +254,11 @@ def add_payment_method(xml, amount, payment_method, options)
           end
         else
           xml.tag! 'paymentDetails', credit_fund_transfer_attribute(options) do
-            xml.tag! CARD_CODES[card_brand(payment_method)] do
+            card_code = CARD_CODES[card_brand(payment_method)]
+            if card_code.nil?
+              card_code = UNKNOWN_CARD_CODE
+            end
+            xml.tag! card_code do
```
dtykocki: 😎 The logic checks out here but I'm wondering if it makes sense to abstract into a private method. Something like: `xml.tag! card_code_for(payment_method)`.

test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -451,6 +452,13 @@ def test_transcript_scrubbing
     assert_scrubbed(@credit_card.verification_value.to_s, clean_transcript)
   end
 
+  def test_failed_xml_validation_unknown_card
+    assert response = @gateway.purchase(@amount, @sodexo_voucher, @options)
+    assert_failure response
+    assert_equal '5', response.error_code
+    assert_match %r{XML failed validation: Invalid payment details : Card number not recognised:}, response.message
+  end
```
dtykocki: 👍 Just to be extra safe, how about adding a test for `authorize` and `verify`?

lib/active_merchant/billing/gateways/worldpay.rb
tanyajajodia: I actually tried that first, but the problem is the key it looks for isn't `unknown`. For example, if we pass in a sodexo card that's not supported / unknown for worldpay `credit_card('6060704495764400', brand: 'sodexo')`, the key here would be `sodexo`. Let me know if I'm missing something...

lib/active_merchant/billing/gateways/worldpay.rb
tanyajajodia: Certainly - that would be a cleaner way for sure.

test/remote/gateways/remote_worldpay_test.rb
tanyajajodia: 👍

PR #3219, opened 2019-05-14
Mundipagg: Make gateway_affiliation_id an option

curiousepic:
gateway_affiliation_id was originally implemented as a gateway-level
credential (named gateway_id). However, this field makes more sense as a
transaction-level field.

Remote:
23 tests, 58 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
17 tests, 76 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/mundipagg.rb
+++ b/lib/active_merchant/billing/gateways/mundipagg.rb
@@ -155,7 +155,8 @@ module ActiveMerchant #:nodoc:
         post[:customer][:name] = payment.name if post[:customer]
         post[:customer_id] = parse_auth(payment)[0] if payment.is_a?(String)
         post[:payment] = {}
-        post[:payment][:gateway_affiliation_id] = @options[:gateway_id] if @options[:gateway_id]
+        affiliation = options[:gateway_affiliation_id] || @options[:gateway_id]
+        post[:payment][:gateway_affiliation_id] = affiliation if affiliation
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -155,7 +155,7 @@ def add_payment(post, payment, options)
         post[:customer][:name] = payment.name if post[:customer]
         post[:customer_id] = parse_auth(payment)[0] if payment.is_a?(String)
         post[:payment] = {}
-        post[:payment][:gateway_affiliation_id] = @options[:gateway_id] if @options[:gateway_id]
+        post[:payment][:gateway_affiliation_id] = options[:gateway_affiliation_id] if options[:gateway_affiliation_id]
```
bayprogrammer: Would there be any value to falling back to the GW option if the txn-level option is omitted (esp. if there's a chance of impacting other users of AM who might benefit from a backwards-compatible fallback)?

curiousepic: That's a good point. Since it was never required on initialize and we aren't messing with that signature anyway, there's no harm in leaving a naive fallback as you suggest.

curiousepic: Updated with fallback

PR #3223, opened 2019-05-16
Checkout.com V2: Fetch payment by ID or session ID support

obahareth:
This is necessary to complete the 3D Secure flow with Checkout, a checkout session ID must be verified using their API. This is necessary if one wants to store 3DS-secured credit cards.

Other changes:

- Some slight refactoring for `commit` to now rely on an `api_request` method which makes requests and parses responses.
- Have all Checkout V2 requests happen via `ssl_request` rather than `ssl_post`, because we now have a GET request in this PR.
- Stub ssl_request instead of ssl_post in checkout_v2 unit test.
- Update signature of `check_request` block params in checkout_v2 unit test.

If this PR is merged, all open PRs for Checkout.com V2 will need the modifications for the unit tests to stub `ssl_request` instead of `ssl_post` and update the signature of `check_request` or their tests will be failing.

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -138,16 +138,25 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def commit(action, post, authorization = nil)
+      def api_request(method, url, post = {})
+        raw_response = response = nil
+
         begin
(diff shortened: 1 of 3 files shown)


obahareth:
If it would help in reviewing/merging, I have all the Checkout V2 PRs I made merged into https://github.com/HungerStation/active_merchant/tree/checkout_v2_updates

PR #3230, opened 2019-05-22
[Gateways::Cybersource] Send tokenization data when card is :master

pi3r:
Currently it will send tokenization data for MasterCard when the card
type is `:mastercard` which is an invalid card type. This commit makes it
so it will send tokenization data when the card type is `:master` and not
`:mastercard`.

Right now, all tokenized mastercard are failing on Cybersource (at least for us).

Tests seem to only fail for the `SoEasyPay` which are unrelated to those changes
```
4069 tests, 69102 assertions, 1 failures, 7 errors, 0 pendings, 0 omissions, 0 notifications
99.8034% passed
```


pi3r:
Looks like it's no kicking a Travis Build anymore?

Girardvjonathan:
@pi3r github is having issue with webhook

PR #3231, opened 2019-05-22
TransFirst Express: Fix blank address2 values

britth:
If a user passes in an empty string for the address2 field, their 
transaction will fail on this gateway. This PR ensures that address2 
only gets passed through if it is present and not blank. Additionally 
it cleans up some remote test failures. test_successful_verify still 
fails in the remote tests, but failure is unrelated to change.

Unit:
21 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
33 tests, 107 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.9697% passed

### DIFF
--- a/test/remote/gateways/remote_trans_first_transaction_express_test.rb
+++ b/test/remote/gateways/remote_trans_first_transaction_express_test.rb
@@ -7,7 +7,7 @@ class RemoteTransFirstTransactionExpressTest < Test::Unit::TestCase
 
     @amount = 100
     @declined_amount = 21
-    @credit_card = credit_card('4485896261017708')
+    @credit_card = credit_card('4485896261017708', verification_value: 999)
     @check = check
 
     billing_address = address({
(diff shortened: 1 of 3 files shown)


britth:
@jknipp

test/remote/gateways/remote_trans_first_transaction_express_test.rb
```diff
@@ -285,16 +305,9 @@ def test_failed_credit
...
-    visa = credit_card('4485896261017708')
-    amex = credit_card('371449635392376', verification_value: 1234)
-    mastercard = credit_card('5499740000000057')
-    discover = credit_card('6011000991001201')
-
-    [visa, amex, mastercard, discover].each do |credit_card|
-      response = @gateway.verify(credit_card, @options)
-      assert_success response
-      assert_match 'Succeeded', response.message
-    end
```
jknipp: Why did we remove the tests for Amercian Express, Mastercard and Discover?

britth: It seems like something has changed since these were initially added such that only the visa card actually succeeds for the verify test. Looks like they've been broken since at least earlier this year (https://github.com/activemerchant/active_merchant/pull/3133). I haven't had luck yet, but I can try to dig into the docs some more to figure out why that's the case - would that be better left to a separate PR (and change this back) or a fix here?

jknipp: I'd do a cursory look to see if TransFirst provides updated test cards, I wouldn't spend more than ~30 mins on that. If you can't find them, revert the change and notate the PR with the test(s) that fail and that they are unrelated to your change. This is an instance where it's ok to ignore the test and move on.

PR #3232, opened 2019-05-24
Implement Worldpay store

bayprogrammer:
We are only supporting shopper-scoped tokens right now, and not
implementing delete/unstore, saving a token along with an authorize or
purchase call, overriding token details on use, or setting the token
lifetime.

The design of the opaque token authorization string we return from
`store` has been intentionally designed to be flexible enough to extend
in a backwards-compatible way should any of the above features need to
be implemented.

Unit:
59 tests, 329 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
50 tests, 219 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96% passed

The two failing remote tests are concerning 3DS parameter pass through
which the test credentials I'm using does not have permission to use
presently.

ECS-347

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -59,10 +59,12 @@ module ActiveMerchant #:nodoc:
 
       def authorize(money, payment_method, options = {})
         requires!(options, :order_id)
-        authorize_request(money, payment_method, options)
+        payment_details = payment_details_from(payment_method)
+        authorize_request(money, payment_method, payment_details.merge(options))
       end
 
       def capture(money, authorization, options = {})
+        authorization = order_id_from_authorization(authorization.to_s)
         MultiResponse.run do |r|
           r.process { inquire_request(authorization, options, 'AUTHORISED') } unless options[:authorization_validated]
           if r.params
@@ -74,6 +76,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def void(authorization, options = {})
+        authorization = order_id_from_authorization(authorization.to_s)
         MultiResponse.run do |r|
           r.process { inquire_request(authorization, options, 'AUTHORISED') } unless options[:authorization_validated]
           r.process { cancel_request(authorization, options) }
@@ -81,6 +84,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def refund(money, authorization, options = {})
+        authorization = order_id_from_authorization(authorization.to_s)
         response = MultiResponse.run do |r|
           r.process { inquire_request(authorization, options, 'CAPTURED', 'SETTLED', 'SETTLED_BY_MERCHANT') }
           r.process { refund_request(money, authorization, options) }
@@ -97,14 +101,20 @@ module ActiveMerchant #:nodoc:
       #   and other transactions should be performed on a normal eCom-flagged
       #   merchant ID.
       def credit(money, payment_method, options = {})
-        credit_request(money, payment_method, options.merge(:credit => true))
+        payment_details = payment_details_from(payment_method)
+        credit_request(money, payment_method, payment_details.merge(:credit => true, **options))
       end
 
-      def verify(credit_card, options={})
+      def verify(payment_method, options={})
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
+          r.process { authorize(100, payment_method, options) }
           r.process(:ignore_result) { void(r.authorization, options.merge(:authorization_validated => true)) }
         end
+      end
+
+      def store(credit_card, options={})
+        requires!(options, :customer)
+        store_request(credit_card, options)
       end
 
       def supports_scrubbing
@@ -142,6 +152,10 @@ module ActiveMerchant #:nodoc:
 
       def credit_request(money, payment_method, options)
         commit('credit', build_authorization_request(money, payment_method, options), :ok, 'SENT_FOR_REFUND', options)
+      end
+
+      def store_request(credit_card, options)
+        commit('store', build_store_request(credit_card, options), options)
       end
 
       def build_request
@@ -184,7 +198,7 @@ module ActiveMerchant #:nodoc:
                 end
               end
               add_payment_method(xml, money, payment_method, options)
-              add_email(xml, options)
+              add_shopper(xml, options)
               if options[:hcg_additional_data]
                 add_hcg_additional_data(xml, options)
               end
@@ -224,6 +238,22 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def build_store_request(credit_card, options)
+        build_request do |xml|
+          xml.tag! 'submit' do
+            xml.tag! 'paymentTokenCreate' do
+              add_authenticated_shopper_id(xml, options)
+              xml.tag! 'createToken'
+              xml.tag! 'paymentInstrument' do
+                xml.tag! 'cardDetails' do
+                  add_card(xml, credit_card, options)
+                end
+              end
+            end
+          end
+        end
+      end
+
       def add_amount(xml, money, options)
         currency = options[:currency] || currency(money)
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -483,9 +548,52 @@ def required_status_message(raw, success_criteria)
...
+            token_id: raw[:payment_token_id],
+            token_scope: 'shopper',
+            customer: options[:customer]
+          )
+        else
+          pair = raw.detect { |k, v| k.to_s =~ /_order_code$/ }
+          (pair ? pair.last : nil)
+        end
+      end
+
```
bayprogrammer: If, in the future, anyone needs to support merchant-scoped tokens, it will be easy to add without breaking compatibility with the current format. Likewise should this form of authorization need to embed the original order id (not something we need to do now, but would be needed to be able to return an authorization from a authorize/purchase call where the caller also wants to store the token in a single call (Worldpay does support this use case)).

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -460,17 +507,35 @@ def handle_response(response)
       #   - a string or an array of strings (if one of many responses)
       #   - An array of strings if one of many responses could be considered a
       #     success.
```
bayprogrammer: We need the ability to return success based on responses materially different than looking for `:ok` in the parsed response. The `:last_event` stuff also wouldn't work for `store` because the response for `store` doesn't include last event information. I broke things down into simpler methods for each type of success-check, and left things more easily extensible for any future types of action-specific success criteria we may need (rather than continue to pile on the existing conditional logic in the original method).

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -144,6 +151,10 @@ def credit_request(money, payment_method, options)
         commit('credit', build_authorization_request(money, payment_method, options), :ok, 'SENT_FOR_REFUND', options)
       end
 
+      def store_request(credit_card, options)
+        commit('store', build_store_request(credit_card, options), options)
```
bayprogrammer: Note that we are intentionally not passing the `*success_requirements` argument. It is optional, and not applicable to `store`. See comments RE: `success_and_message_from`.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -390,14 +432,19 @@ def parse(action, xml)
...
         node.attributes.each do |k, v|
-          raw["#{node.name.underscore}_#{k.underscore}".to_sym] = v
+          raw["#{node_name}_#{k.underscore}".to_sym] = v
         end
         if node.has_elements?
-          raw[node.name.underscore.to_sym] = true unless node.name.blank?
+          raw[node_name.to_sym] = true unless node.name.blank?
           node.elements.each { |e| parse_element(raw, e) }
         else
-          raw[node.name.underscore.to_sym] = node.text unless node.text.nil?
```
bayprogrammer: This allows us to include the error message when there are newlines. I discovered this issue when trying to make assertions about the returned error message from existing canned responses in the unit tests. Now we can return the error message whether the `CDATA` information is preceded and/or followed by newlines or not.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -241,7 +268,7 @@ def add_amount(xml, money, options)
       end
 
       def add_payment_method(xml, amount, payment_method, options)
-        if payment_method.is_a?(String)
+        if options[:payment_type] == :pay_as_order
```
curiousepic: I have yet to fully read your additions, but I'm seeing that for AM users of this feature, they'll need to start sending a new `payment_type` option? Would it hurt to retain the string check for backward compatibility?

curiousepic: Or was this made to distinguish it from stored PMs, since they end up as strings?

curiousepic: I see now that this is a flag created by a method that parses the PM, neat. Disregard for now.

bayprogrammer: Yep, you got it. The problem was that both the TPV tokens and pay-by-reference/pay-as-order string payment methods needed to be distinguished early on. Not a user-facing option, should be fully backwards-compatible.

lib/active_merchant/billing/gateways/worldpay.rb
curiousepic: I'd vote to just axe `success_and_message_from` step too.

lib/active_merchant/billing/gateways/worldpay.rb
curiousepic: So I think I like this. But I do want to bring up a consideration. Obviously the way AM handles `authorization`s is not ideal, but to the extent that there is a standard, that standard seems to be for newly supported elements to be concatenated on the end of whatever is already present, for any action type. So what I would have expected here is that the data for a stored card would start with the second element of a concatenated authorization, regardless of whether that first element actually has a value.

Does that make sense? And if so, does it sound like something we should actually care about?

curiousepic: The advantage of this is that, if we're assuming it needs to remain a string, the ordering allows it to be parsed both by the adapter and by anything else looking at the string (as long as the order is known and maintained), independent of transaction type (some authorizations add the action type for future explicit parsing).

bayprogrammer: Yah, I think it's a good idea (and easy to implement) to ensure that the order_id, if ever it should be present in a compound auth like this one, would take the first ordinal position (and just be empty for now in store where it doesn't apply). That way any future extensions where both the order_id and these token-specific bits are necessary (such as if anyone needs to implement authorize+store in a single call) can rely on the first element (including if there are no "|" characters) always being the order_id.

Not completely sure though if I follow about checking the length of split for pay-as-order though. Are you referring to how we distinguish between pay-as-order payment method strings and stored token id strings?

curiousepic: Cool, I think we should go with that.

>Not completely sure though if I follow about checking the length of split for pay-as-order though. Are you referring to how we distinguish between pay-as-order payment method strings and stored token id strings?

Yes, though I guess it depends on how the current methods change, we might not even need that. I'll leave it up to you and wait to see the changes.

bayprogrammer: @curiousepic Ok, I've pushed up some updates that implement this.

Since we want to ensure there is a slot in our compound authorization strings that can hold the `order_id` value (if and when it's used), I went ahead and ensured that these complex authorization strings could be used in place of a bare order_id authorization.

Was actually a bit uncertain whether to proceed that far, but it would have felt awkward to me to have the ability to store the order_ids in an authorization that would have been invalid to use for capture/etc. Especially welcome feedback on this point.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -483,11 +543,64 @@ def required_status_message(raw, success_criteria)
...
+        token_details[:token_id] = token_id if token_id.present?
+        token_details[:token_scope] = token_scope if token_scope.present?
+        token_details[:customer] = customer if customer.present?
+
+        token_details
+      end
+
+      def payment_details_from(payment_method)
+        payment_details = {}
+        if payment_method.is_a?(CreditCard)
```
bayprogrammer: @curiousepic So I updated this to rely on our parsing the complex auth and checking that the token-specific field is in the resultant hash to determine whether it's a pay-by-order auth string or token auth string. As part of this I flipped the logic around to rely on detecting whether it's a credit card payment method using `is_a?` (I generally try really hard to avoid `is_a?`). One thing I'm uncertain of is if there is the potential for any other kind of payment_method we need to worry about at this stage which would fail this check but also not be a String (and thus cause the later code to break which assume either a `CreditCard` or `String` instance are being used).

Is there a common method that all card-like objects can respond to which could be used to differentiate them from `String` instances without having to resort to `is_a?` here?

curiousepic: Looks good!  We do often use `respond_to?(:number)`.

bayprogrammer: Ah, I like that better. Updated. Thanks!

PR #3235, opened 2019-05-29
Deleted landing_page param to allow Paypal use their logic

qchenme:
Refer to this pull request: https://github.com/activemerchant/active_merchant/pull/3176

Paypal wants us to remove `landing_page` to be able to use their own logic - determine whether directing users to login or guest checkout page. 

The empty `landing_page` was sent previously, now it will be removed entirely.

The deletion includes:
- ` LandingPage` tag with `:paypal_chooses_landing_page` option
- Corresponding test `test_paypal_chooses_landing_page`
- assert `LandingPage` tag in `test_allow_guest_checkout`


qchenme:
Updates have been done in other places, no need to change the code.

PR #3236, opened 2019-05-31
Adyen: Support for additional AVS code mapping

jknipp:
Map additional Adyen AVS result codes to Active Merchant AVS result
codes.

ECS-301

Unit:

37 tests, 179 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:

60 tests, 187 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,7 @@
 * Adyen: Pass updateShopperStatement, industryUsage [curiousepic] #3233
 * TransFirst Transaction Express: Fix blank address2 values [britth] #3231
 * WorldPay: Add support for store method [bayprogrammer] #3232
+* Adyen: Support for additional AVS code mapping [jknipp] #3236
 
 == Version 1.95.0 (May 23, 2019)
 * Adyen: Constantize version to fix subdomains [curiousepic] #3228
--- a/lib/active_merchant/billing/avs_result.rb
+++ b/lib/active_merchant/billing/avs_result.rb
@@ -3,11 +3,10 @@
 module ActiveMerchant
   module Billing
     # Implements the Address Verification System
-    # https://www.wellsfargo.com/downloads/pdf/biz/merchant/visa_avs.pdf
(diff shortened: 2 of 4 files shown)


jknipp:
@curiousepic I started implementing AmEx specific messages, but it's feeling like the wrong solution. The `card brand` isn't easily accessible when extracting the AVS response, which would mean several edits per gateway to retrieve the AmEx specific message.

Further, I realized that the AVS messages are identical to those on `https://www.cybersource.com/developers/other_resources/quick_references/avs_results/`, which likely come from their AVS code documentation in http://apps.cybersource.com/library/documentation/dev_guides/CC_Svcs_SO_API/Credit_Cards_SO_API.pdf. Mysteriously, `J` and `Q` are not covered. This is the clearest source of truth that we have. 

These docs map alpha code `N` to two different messages. I'm now thinking we should leave the message as is or update `N` to be:

```
Street address and postal code do not match.  For the American Express card type: Card member's name, street address and postal code do not match.
```

jknipp:
Maybe we should only update https://docs.spreedly.com/guides/debugging/#avs-response-codes to indicate `N` is different for American Express cards.

curiousepic:
@jknipp I'd like to find a solution for AM in general. I don't know why it took so long to think of this (at least for me) but we could simply add amex provisions to the messages with potentially different meanings, something like: 
"N: Street address and postal code do not match. American Express: Card member's name, street address and postal code do not match."

jknipp:
Updated based on discussion with @curiousepic. I refrained from adding unit tests as I don't feel like they add value to what we currently have. I'm happy to reconsider this position if we think they would be beneficial.

jknipp:
An AVS result message update for https://spreedly.atlassian.net/browse/ECS-56 is tagging along.

PR #3240, opened 2019-06-06
Fix phone number formatting

nicolas-maalouf-cko:


pi3r:
Can you rebase and add a test for it please?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -133,7 +133,8 @@ def add_3ds(post, options)
           post[:'3ds'][:enabled] = true
           post[:'3ds'][:eci] = options[:eci] if options[:eci]
           post[:'3ds'][:cryptogram] = options[:cavv] if options[:cavv]
-          post[:'3ds'][:xid] = options[:xid] if options[:xid]
+          post[:'3ds'][:version] = options[:version] if options[:version]
+          post[:'3ds'][:xid] = options[:ds_transaction_id] || options[:xid]
```
molbrown: @bayprogrammer Do I understand correctly based on [Active Merchant's Standardized 3DS ](https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields) guide that the option fields `version`, `eci`, `cavv`, `ds_transaction_id`, and `xid` should all be nested inside of a `three_d_secure` object, instead of each being top-level `options`?

molbrown: (Or at least that this should be compatible both ways?)

bayprogrammer: @molbrown, Yes, that's right. Good catch! I see that here we are checking for the presence of the `three_d_secure` option before proceeding, but then pulling the options directly out of the `options` hash (rather than e.g. `options[:three_d_secure][:version]`).

@nicolas-maalouf-cko does that sound right to you?

filipebarcos: Right, I totally missed that. @nicolas-maalouf-cko Could you adjust those, please?

PR #3243, opened 2019-06-14
NMI: Add support for stored credentials

bayprogrammer:
Unit:
45 tests, 342 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
41 tests, 151 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-258

Closes #3243

### DIFF
--- a/lib/active_merchant/billing/gateways/nmi.rb
+++ b/lib/active_merchant/billing/gateways/nmi.rb
@@ -31,6 +31,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_invoice(post, amount, options)
         add_payment_method(post, payment_method, options)
+        add_stored_credential(post, options)
         add_customer_data(post, options)
         add_vendor_data(post, options)
         add_merchant_defined_fields(post, options)
@@ -43,6 +44,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_invoice(post, amount, options)
         add_payment_method(post, payment_method, options)
+        add_stored_credential(post, options)
         add_customer_data(post, options)
         add_vendor_data(post, options)
         add_merchant_defined_fields(post, options)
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/nmi.rb
```diff
@@ -175,6 +176,35 @@ def add_payment_method(post, payment_method, options)
...
+        # :reason_type, when provided, overrides anything previously set in
+        # post[:billing_method] (see `add_invoice` and the :recurring) option
+        case stored_credential[:reason_type]
+        when 'recurring'
+          post[:billing_method] = 'recurring'
+        when 'installment'
+          post[:billing_method] = 'installment'
+        when 'unscheduled'
+          post.delete(:billing_method)
+        end
```
bayprogrammer: Verified with NMI integration support over the phone that for unscheduled stored credential transactions the billing_method need not be set (it's only valid values are 'recurring' or 'installment').

lib/active_merchant/billing/gateways/nmi.rb
```diff
@@ -175,6 +176,35 @@ def add_payment_method(post, payment_method, options)
           post[:ccnumber] = payment_method.number
           post[:cvv] = payment_method.verification_value unless empty?(payment_method.verification_value)
           post[:ccexp] = exp_date(payment_method)
+          add_stored_credential(post, options)
```
bayprogrammer: Adding the stored credential any place we would take send a `ccnumber` field (which seems consistent with how things are implemented in Core), but it's not clear to me whether or not the idea of stored credentials applies to tokenized payment methods. Visa does have this to say:

> A stored credential is information (including, but not limited to, an account number or payment token) that is stored by a merchant or its agent, a payment facilitator, or a staged digital wallet operator to process future transactions.

source: https://usa.visa.com/dam/VCOM/global/support-legal/documents/stored-credential-transaction-framework-vbs-10-may-17.pdf

So it sounds like the idea of a token used as a payment method isn't necessarily precluding the use of the stored credential fields. Should we just always add stored credential info, when provided, any time we're adding payment information (and not worry about the type of payment method)? Welcome thoughts on this.

test/test_helper.rb
```diff
@@ -245,6 +245,26 @@ def statement_address(options = {})
       }.update(options)
     end
 
+    def stored_credential(*args, **options)
```
bayprogrammer: Added this helper to make generating the stored credential hash less verbose. Since this is useful for any gateway we write stored credential tests for using the standardized hash format, I added it to the test helper directly.

curiousepic: Nice, this also serves as good documentation of the format in code, since it is only otherwise present in the PR/commit message.

test/remote/gateways/remote_nmi_test.rb
```diff
@@ -298,8 +375,30 @@ def test_network_tokenization_transcript_scrubbing
...
+  def assert_cvv_scrubbed(transcript)
+    assert_match(/cvv=\[FILTERED\]/, transcript)
+  end
 
-    # "password=password is filtered, but can't be tested b/c of key match"
-    # assert_scrubbed(@gateway.options[:password], clean_transcript)
+  def sc_options(*args, id: nil)
```
curiousepic: Let's use the full word for explicitness.

lib/active_merchant/billing/gateways/nmi.rb
curiousepic: I think it's right to add them whenever they are present in options, independent of type of PM. However, I believe they only need to be sent for auths and purchases (and the auth step of any other actions that include one).

bayprogrammer: @curiousepic Doesn't look like the `authorize` method is delegated to from any other methods, so it would just be `authorize` and `purchase` where we add the stored credential options to an outgoing request.

Me moving the `add_stored_credential` call to happen directly from `authorize` and `purchase` (rather `add_payment_method`) sound good?

PR #3244, opened 2019-06-17
Fix ENV based configuration of Net::Http for proxies (again)

georgiev:
Proxy configuration from environment variables is broken again (previous incarnations #2372 / #2800). This time at gateway level - ActiveMerchant::PostsData unconditionally assigns gateway class level proxy_address (defalu nil) to connection.proxy_address (default :ENV) changing the default from **"Use environment"**  to **"Disable proxy"**.  
The proposed fix is to initialize both default values with :ENV


georgiev:
Closing for now - the issue and the fix are valid but the proposed test is failing in Travis environment, I'll have to come up with better test. Suggestions are welcome.

PR #3248, opened 2019-06-19
Fixes checkout v2 integration

filipebarcos:
Formatting + 2 fixes

- Card on File: the way we were sending that value was wrong/misplaced.
- Phone numbers: Checkout.com does not support non-digit chars for the `source: { phone: { number: '' }}}` field

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -32,7 +32,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         post[:capture] = false
         add_invoice(post, amount, options)
-        add_payment_method(post, payment_method, options)
+        add_payment_method(post, payment_method)
         add_customer_data(post, options)
         add_transaction_data(post, options)
         add_3ds(post, options)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -94,35 +94,34 @@ def add_invoice(post, money, options)
...
-        if(address && post[:source])
+        if address && post[:source]
           post[:source][:billing_address] = {}
           post[:source][:billing_address][:address_line1] = address[:address1] unless address[:address1].blank?
           post[:source][:billing_address][:address_line2] = address[:address2] unless address[:address2].blank?
           post[:source][:billing_address][:city] = address[:city] unless address[:city].blank?
           post[:source][:billing_address][:state] = address[:state] unless address[:state].blank?
           post[:source][:billing_address][:country] = address[:country] unless address[:country].blank?
           post[:source][:billing_address][:zip] = address[:zip] unless address[:zip].blank?
-          post[:source][:phone] = { number: address[:phone] } unless address[:phone].blank?
```
elfassy: 👍

PR #3251, opened 2019-06-21
Barclaycard Smartpay: 3DS2 Support

britth:
This preps Barclaycard Smartpay for 3DS2 integration. The api version
is bumped up to 40 to get relevant 3DS2 functionality, and additional 
browser_info is captured. The parse method is also updated so that it 
can return a nested hash with an additionalData field for 3DS2.

Unit:
28 tests, 142 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
34 tests, 76 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.0588% passed
(failed `test_successful_third_party_payout` - unrelated invalid credentials error)

### DIFF
--- a/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
+++ b/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
@@ -13,7 +13,7 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'https://www.barclaycardsmartpay.com/'
       self.display_name = 'Barclaycard Smartpay'
 
-      API_VERSION = 'v30'
+      API_VERSION = 'v40'
 
       def initialize(options = {})
         requires!(options, :company, :merchant, :password)
@@ -37,7 +37,7 @@ module ActiveMerchant #:nodoc:
         post[:card] = credit_card_hash(creditcard)
         post[:billingAddress] = billing_address_hash(options) if options[:billing_address]
         post[:deliveryAddress] = shipping_address_hash(options) if options[:shipping_address]
-        add_3ds(post, options) if options[:execute_threed]
+        add_3ds(post, options)
         commit('authorise', post)
       end
 
@@ -186,7 +186,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def parse_avs_code(response)
-        AVS_MAPPING[response['avsResult'][0..1].strip] if response['avsResult']
+        AVS_MAPPING[response['additionalData']['avsResult'][0..1].strip] if response.dig('additionalData', 'avsResult')
       end
 
       def flatten_hash(hash, prefix = nil)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -37,7 +37,7 @@ def authorize(money, creditcard, options = {})
         post[:card] = credit_card_hash(creditcard)
         post[:billingAddress] = billing_address_hash(options) if options[:billing_address]
         post[:deliveryAddress] = shipping_address_hash(options) if options[:shipping_address]
-        add_3ds(post, options) if options[:execute_threed]
+        add_3ds(post, options)
```
nfarve: We don't need the `:execute_threed`?

britth: good question - i was mostly leaning on what's happening for adyen, which, in the `add_3ds` method will add 3ds2 browser info if that gets passed in (whether it receives the `execute_threed` flag or not), or will check that a 3ds flag is present before falling back to adding 3ds1 data

nfarve: 👍 Might be something to monitor when it goes in production.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -210,12 +210,18 @@ def headers(account, password)
       end
 
       def parse(response)
-        Hash[
-          response.split('&').map do |x|
-            key, val = x.split('=', 2)
-            [key.split('.').last, CGI.unescape(val)]
+        parsed_response = {}
+        response.split('&').map do |x|
```
jeremywrowe: It's really confusing what this is doing - can't we just use: http://ruby-doc.org/stdlib-2.6.3/libdoc/cgi/rdoc/CGI.html#method-c-parse?

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
```diff
@@ -343,8 +349,31 @@ def store_request(options)
...
+              screenHeight: browser_info[:height],
+              screenWidth: browser_info[:width],
+              timeZoneOffset: browser_info[:timezone],
+              userAgent: browser_info[:user_agent]
+            }
+
+            if device_channel = three_ds_2_options[:channel]
+              post[:threeDS2RequestData] = {
+                deviceChannel: device_channel,
+                notificationURL: three_ds_2_options[:notification_url] || 'https://example.com/notification'
```
jeremywrowe: The fallback URL is problematic - we should remove it from Adyen also. This is being done for tests as a convenience.

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
britth: good call - updated 👍 Outside of this spot, does the rest of the method seem overly complex (creating the nested hash)?

lib/active_merchant/billing/gateways/barclaycard_smartpay.rb
britth: removed fallback and added value within 3ds2_options hash passed in tests

jeremywrowe: Sounds good :D

PR #3252, opened 2019-06-21
Adyen: Replace empty state string with N/A

therufs:
[ECS-367](https://spreedly.atlassian.net/browse/ECS-367)

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -232,9 +232,13 @@ module ActiveMerchant #:nodoc:
           post[:billingAddress][:houseNumberOrName] = address[:address2] || 'N/A'
           post[:billingAddress][:postalCode] = address[:zip] if address[:zip]
           post[:billingAddress][:city] = address[:city] || 'N/A'
-          post[:billingAddress][:stateOrProvince] = address[:state] || 'N/A'
+          post[:billingAddress][:stateOrProvince] = get_state(address)
           post[:billingAddress][:country] = address[:country] if address[:country]
         end
+      end
+
+      def get_state(address)
+        address[:state] && !address[:state].blank? ? address[:state] : 'N/A'
       end
 
       def add_invoice(post, money, options)
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -605,11 +605,16 @@ class RemoteAdyenTest < Test::Unit::TestCase
     assert_match Gateway::STANDARD_ERROR_CODE[:incorrect_address], response.error_code
   end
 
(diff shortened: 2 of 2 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -605,6 +605,13 @@ def test_blank_country_for_purchase
     assert_match Gateway::STANDARD_ERROR_CODE[:incorrect_address], response.error_code
   end
 
+  def test_nil_state_for_purchase
+    @options[:billing_address][:state] = nil
+    response = @gateway.authorize(@amount, @credit_card, @options)
+    assert_failure response
+    assert_match Gateway::STANDARD_ERROR_CODE[:incorrect_address], response.error_code
+  end
+
```
molbrown: 🤔 Wait a minute. Don't we want both these state tests to succeed? Are we sure N/A is what Adyen wants here?

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -231,11 +231,15 @@ def add_address(post, options)
...
-          post[:billingAddress][:stateOrProvince] = address[:state] || 'N/A'
+          post[:billingAddress][:stateOrProvince] = get_state(address)
           post[:billingAddress][:country] = address[:country] if address[:country]
         end
       end
 
+      def get_state(address)
+        address[:state] && !address[:state].blank? ? address[:state] : 'N/A'
+      end
+
```
molbrown: 😻

test/remote/gateways/remote_adyen_test.rb
therufs: *facepalm* yes I suppose we do want that

therufs:
Unit tests:
--------------------------------------------------------------------------------------------------------------------------------
38 tests, 183 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------

Remote tests:
--------------------------------------------------------------------------------------------------------------------------------
61 tests, 187 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.4426% passed
--------------------------------------------------------------------------------------------------------------------------------

Fixes for failing tests anticipated pursuant to https://spreedly.slack.com/archives/CF080246R/p1561051065014400

test/remote/gateways/remote_adyen_test.rb
molbrown: I'm curious. How does this just work now? Did you have to change an account setting or something?

therufs: Hmm, the test has been flipped to expect success but the update doesn't seem to be self-aware in Github 🤔

PR #3254, opened 2019-06-24
Decidir: Add new gateway

jknipp:
Add support for the Decidir gateway. This implementation sends card data
inline with purchase and authorize requests rather than creating payment
method tokens.

ECS-374

Unit:
18 tests, 104 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
17 tests, 60 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -1,0 +1,232 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class DecidirGateway < Gateway
+      self.test_url = 'https://developers.decidir.com/api/v2'
+      self.live_url = 'https://live.decidir.com/api/v2'
+
+      self.supported_countries = ['AR']
+      self.money_format = :cents
+      self.default_currency = 'ARS'
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club]
+
+      self.homepage_url = 'http://www.decidir.com'
+      self.display_name = 'Decidir'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        1 => STANDARD_ERROR_CODE[:call_issuer],
+        2 => STANDARD_ERROR_CODE[:call_issuer],
+        3 => STANDARD_ERROR_CODE[:config_error],
+        4 => STANDARD_ERROR_CODE[:pickup_card],
+        5 => STANDARD_ERROR_CODE[:card_declined],
+        7 => STANDARD_ERROR_CODE[:pickup_card],
+        12 => STANDARD_ERROR_CODE[:processing_error],
+        14 => STANDARD_ERROR_CODE[:invalid_number],
+        28 => STANDARD_ERROR_CODE[:processing_error],
+        38 => STANDARD_ERROR_CODE[:incorrect_pin],
+        39 => STANDARD_ERROR_CODE[:invalid_number],
+        43 => STANDARD_ERROR_CODE[:pickup_card],
+        45 => STANDARD_ERROR_CODE[:card_declined],
+        46 => STANDARD_ERROR_CODE[:invalid_number],
+        47 => STANDARD_ERROR_CODE[:card_declined],
+        48 => STANDARD_ERROR_CODE[:card_declined],
+        49 => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        51 => STANDARD_ERROR_CODE[:card_declined],
+        53 => STANDARD_ERROR_CODE[:card_declined],
+        54 => STANDARD_ERROR_CODE[:expired_card],
+        55 => STANDARD_ERROR_CODE[:incorrect_pin],
+        56 => STANDARD_ERROR_CODE[:card_declined],
+        57 => STANDARD_ERROR_CODE[:card_declined],
+        76 => STANDARD_ERROR_CODE[:call_issuer],
+        96 => STANDARD_ERROR_CODE[:processing_error],
+        97 => STANDARD_ERROR_CODE[:processing_error],
+      }
+
+      def initialize(options={})
+        requires!(options, :api_key)
+        super
+        @options[:preauth_mode] ||= false
+      end
+
+      def purchase(money, payment, options={})
+        raise ArgumentError, 'Purchase is not supported on Decidir gateways configured with the preauth_mode option' if @options[:preauth_mode]
+
+        post = {}
+        add_auth_purchase_params(post, money, payment, options)
+        commit(:post, 'payments', post)
+      end
+
+      def authorize(money, payment, options={})
+        raise ArgumentError, 'Authorize is not supported on Decidir gateways unless the preauth_mode option is enabled' unless @options[:preauth_mode]
+
+        post = {}
+        add_auth_purchase_params(post, money, payment, options)
+        commit(:post, 'payments', post)
+      end
+
+      def capture(money, authorization, options={})
+        raise ArgumentError, 'Capture is not supported on Decidir gateways unless the preauth_mode option is enabled' unless @options[:preauth_mode]
+
+        post = {}
+        add_amount(post, money, options)
+        commit(:put, "payments/#{authorization}", post)
+      end
+
+      def refund(money, authorization, options={})
+        post = {}
+        add_amount(post, money, options)
+        commit(:post, "payments/#{authorization}/refunds", post)
+      end
+
+      def void(authorization, options={})
+        post = {}
+        commit(:post, "payments/#{authorization}/refunds", post)
+      end
+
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -0,0 +1,220 @@
...
+
+        commit(:post, 'payments', post)
+      end
+
+      def authorize(money, payment, options={})
+        post = {}
+        add_auth_purchase_params(post, money, payment, options)
+
+        commit(:post, 'payments', post)
+      end
```
curiousepic: Am I missing something or is there nothing distinguishing auth and purchase?

curiousepic: Oh they use different creds... hmm.

curiousepic: I vaguely recall discussing this, but what are the downsides of having one set of credentials, and distinguishing them by type (`auth_api_key` and `purchase_api_key`)?

jknipp: @curiousepic You are not deceived. The API endpoints for `auth` and `purchase` are identical, as are those for `void` and `refund`. The difference is the `api_key` credential used.

jknipp: The downside rests in how we would determine which credential to use. We could have `auth_api_key` and `purchase_api_key` and require both. We would need to update the transaction methods in AM to pass the `api_key` based on the transaction being performed. I think this is going to be confusing for customers. 

The `api_key` is presented as a single value - at least for the flow we are using. From what I can tell, the `api_key` is based on settings at Decidir.

jknipp: Attaching the email thread for context

> De: Jared Knipp [mailto:jared@spreedly.com] 
> Enviado el: viernes, 7 de junio de 2019 16:46
> Para: Ignacio Castro Ulzurrun <iulzurrun@prismamp.com>
> CC: Lee Jacobs <lee@spreedly.com>; Branden Roden <branden@paymentsfn.com>; Alejandro Garcia <AlGarcia@prismamp.com>; Facundo Islas <FIslas@prismamp.com>
> Asunto: Re: Spreedly - Decidir Integration
> 
> Hi Ignacio, 
> 
> I'm the engineer at Spreedly working on the Decidir gateway integration. 
> 
> Will we receive a sandbox apikey to test all API requests? Or will we need to test with individual apikey's provided in the API documentation for each request type?
> 
> I have a few questions regarding purchase/pre-auth requests that skip tokenization and send payment information inline. The PDF supplement for sending payment info inline with purchase requests does not include an apikey to test with. My attempt to reuse the apikey provided in the API docs for a purchase with a token were unsuccessful. Can you provide an apikey we can use to test this functionality? Is there an account setting required to perform a purchase/pre-auth without first creating a token?
> 
> Thanks,
> 
> Jared
> 
>

curiousepic: I see; I'm definitely on board with a singular cred since it's tied to a mode. However, I'm wondering if we should distinguish them somehow, such as a gateway level field for which type and throw an error if an unsupported one is requested. In any case, some way to identify an existing gateway as one or the other without having to know which mode a certain api_key is for...

jknipp: 🤔It does get weird because the endpoints are the same for `purchase` and `auth`. Nothing prevents a user from making a `purchase` request when they should be doing an `auth`. The request and response look the same to the user.

I'd prefer to add a gateway field for the mode and reserve the using the adapter inheritance model in case we need to support creating payment method tokens prior to purchase/authorize.

jknipp: There is a disclaimer in the [Decidir documentation](https://decidirv2.api-docs.io/1.0/transacciones-en-dos-pasos/flujo-de-una-transaccion-en-dos-pasos) that this must be enabled

> IMPORTANT! 
> Transactions in two steps are included in the special transactions offered by DECIDE. For this reason, in order to enable them, the client must inform  about its use, in order to configure the necessary parameters at the ID Site level. Note that an ID Site that uses this type of Transaction can not implement the classic operation.

curiousepic: @jknipp 👍 Sounds good. Let's actually not use the term "mode" since it implies it can be toggled... maybe something like "auth_type"?

jknipp: The `auth_type` is a binary option, so wouldn't it be toggled? Single step (purchase) would be the default (false), auth and capture would need to be enabled (true).

curiousepic: That does make sense, I'll leave it to you.

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -0,0 +1,220 @@
...
+        {
+          'message' => "A non-JSON response was received from Decidir where one was expected. The raw response was:\n\n#{body}"
+        }
+      end
+
+      def message_from(success, response)
+        return response['status'] if success
+        return response['message'] if response['message']
+
+        msg = nil
```
curiousepic: Nitpick: not a fan of abrvs (though I sympathize with the probable intent of distinguishing it from the response fields)

curiousepic:
Otherwise this looks good! Nice work, grats on your first gateway 😄

lib/active_merchant/billing/gateways/decidir.rb
jknipp: 😄I waffled on the abbreviation - started with `msg`, changed to `message` and then back to `msg` after seeing the abbreviation elsewhere in AM. Won't hurt my feelings to update it.

jknipp:
It looks like partial captures (variable captures) are allowed but they must be configured at Decidir ([referenced here](https://decidirv2.api-docs.io/1.0/transacciones-en-dos-pasos/flujo-de-una-transaccion-en-dos-pasos)). 

I included a test for failed partial captures, but the implementation should not preclude users from attempting and completing a partial capture if their account is appropriately configured. I will include a note in docs about enabling partial captures.

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -44,23 +44,28 @@ class DecidirGateway < Gateway
       def initialize(options={})
         requires!(options, :api_key)
         super
+        @options[:two_phase_capture] ||= false
```
jknipp: @curiousepic Open to suggestions on naming. This is the flag to determine single phase (purchase) vs. two phase (auth + capture).

curiousepic: The term "Preauth" to refer to auth+capture is more common. Since the default is one-phase, the flag might simply be called "Preauth" or "Preauth_mode" (I came around on "mode').

curiousepic: It's also usually good to reflect the gateway's particular nomenclature - exactly how do they refer to this mode?

jknipp: I wasn't sure if `preauth_*` would be descriptive enough to match the docs. They gateway refers to this as `Transactions in two steps` (untranslated `Transacciones en Dos Pasos`). Aka pre-authorization and capture/confirmation. I'm ok changing this to `preauth_mode`.

curiousepic: @jknipp Yeah I think that's probably most suitable.

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -44,27 +44,27 @@ class DecidirGateway < Gateway
       def initialize(options={})
         requires!(options, :api_key)
         super
-        @options[:two_phase_capture] ||= false
+        @options[:preauth_mode] ||= false
       end
 
       def purchase(money, payment, options={})
-        raise ArgumentError, 'Purchase is not supported for two-phase transactions' if @options[:two_phase_capture]
+        raise ArgumentError, 'Purchase is not supported for two-step transactions' if @options[:preauth_mode]
```
curiousepic: Let's also make this a little more descriptive, something like "Purchase is not supported on Decidir gateways configured with the preauth_mode option"?

jknipp: @curiousepic Should I update the message for `auth`, `capture` and `verify`?

curiousepic: Yes plz.

curiousepic:
Nice. Good to go if you're ready 👍

PR #3261, opened 2019-06-27
[Orbital] Add support for 3DS

pi3r:
Adds support for 3ds to Orbital. All remote tests are passing using the test cards provided by Orbital.

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -465,14 +465,60 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def add_cdpt_eci_and_xid(xml, creditcard)
-        xml.tag! :AuthenticationECIInd, creditcard.eci
-        xml.tag! :XID, creditcard.transaction_id if creditcard.transaction_id
+      def add_eci(xml, creditcard, three_d_secure)
+        eci = if three_d_secure
+                three_d_secure[:eci]
+              elsif creditcard.is_a?(NetworkTokenizationCreditCard)
+                creditcard.eci
+              end
+
+        xml.tag!(:AuthenticationECIInd, eci) if eci
       end
 
-      def add_cdpt_payment_cryptogram(xml, creditcard)
+      def add_xid(xml, creditcard, three_d_secure)
+        xid = if three_d_secure && creditcard.brand == 'visa'
+                three_d_secure[:xid]
+              elsif creditcard.is_a?(NetworkTokenizationCreditCard)
+                creditcard.transaction_id
+              end
+
+        xml.tag!(:XID, xid) if xid
+      end
+
+      def add_cavv(xml, creditcard, three_d_secure)
+        return unless three_d_secure && creditcard.brand == 'visa'
+
+        xml.tag!(:CAVV, three_d_secure[:cavv])
+      end
+
+      def add_aav(xml, creditcard, three_d_secure)
+        return unless three_d_secure && creditcard.brand == 'master'
+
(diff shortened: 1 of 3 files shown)


pi3r:
> I do not know about network tokenization much but I noticed the existing `add_cdpt_eci_and_xid`:
> 
> ```
>       def add_cdpt_eci_and_xid(xml, creditcard)
>         xml.tag! :AuthenticationECIInd, creditcard.eci
>         xml.tag! :XID, creditcard.transaction_id if creditcard.transaction_id
>       end
> ```
> 
> that is used when the card is a `NetworkTokenizationCreditCard`. Is the plan that we address network tokenization later?

@bbraschi a network tokenized card usually comes with an XID because it goes through the 3ds 1.0 flow. I didn't reuse those because the schema needs the CAVV to be present in between the ECI and XID. It's a rigide schema where the order of the fields matter... Thank you for the thorough  review!

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -421,6 +425,14 @@ def add_destination_address(xml, address)
         end
       end
 
+      def add_3ds(xml, options)
+        return unless three_d_secure = options[:three_d_secure]
+
+        xml.tag!(:AuthenticationECIInd, three_d_secure[:eci]) if three_d_secure[:eci]
+        xml.tag!(:CAVV, three_d_secure[:cavv]) if three_d_secure[:cavv]
```
manuelclaveras: What about AAV? It's in the schema, not sure why they make a difference between the two but it seems that they are expecting this when the card is a mastercard.

manuelclaveras:
> > I do not know about network tokenization much but I noticed the existing `add_cdpt_eci_and_xid`:
> > ```
> >       def add_cdpt_eci_and_xid(xml, creditcard)
> >         xml.tag! :AuthenticationECIInd, creditcard.eci
> >         xml.tag! :XID, creditcard.transaction_id if creditcard.transaction_id
> >       end
> > ```
> > 
> > 
> > that is used when the card is a `NetworkTokenizationCreditCard`. Is the plan that we address network tokenization later?
> 
> @bbraschi a network tokenized card usually comes with an XID because it goes through the 3ds 1.0 flow. I didn't reuse those because the schema needs the CAVV to be present in between the ECI and XID. It's a rigide schema where the order of the fields matter... Thank you for the thorough review!

You could get rid of it. Your `add_3ds` would do the trick since you don't add the tag CAVV if it's not present.

pi3r:
> > > I do not know about network tokenization much but I noticed the existing `add_cdpt_eci_and_xid`:
> > > ```
> > >       def add_cdpt_eci_and_xid(xml, creditcard)
> > >         xml.tag! :AuthenticationECIInd, creditcard.eci
> > >         xml.tag! :XID, creditcard.transaction_id if creditcard.transaction_id
> > >       end
> > > ```
> > > 
> > > 
> > > that is used when the card is a `NetworkTokenizationCreditCard`. Is the plan that we address network tokenization later?
> > 
> > 
> > @bbraschi a network tokenized card usually comes with an XID because it goes through the 3ds 1.0 flow. I didn't reuse those because the schema needs the CAVV to be present in between the ECI and XID. It's a rigide schema where the order of the fields matter... Thank you for the thorough review!
> 
> You could get rid of it. Your `add_3ds` would do the trick since you don't add the tag CAVV if it's not present. Not that it matters a lot 😄

That's fair, I wanted to avoid changing too much of the previous logic and match what was done with previous gateway classes.

pi3r:
This is ready for 3DS 1.0. I will do a follow up for 2.0

PR #3264, opened 2019-07-01
MiGS: Add tx_source gateway specific field

leila-alderman:
Added the `tx_source` gateway specific field to the MiGS gateway along
with remote and unit tests.

CE-34

Unit:
9 tests, 21 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
14 tests, 44 assertions, 0 failures, 1 errors, 0 pendings, 0 omissions,
0 notifications
92.8571% passed
Undefined method `request_uri` error in `test_server_purchase_url`;
error unrelated to change

### DIFF
--- a/lib/active_merchant/billing/gateways/migs.rb
+++ b/lib/active_merchant/billing/gateways/migs.rb
@@ -63,6 +63,7 @@ module ActiveMerchant #:nodoc:
         add_creditcard(post, creditcard)
         add_standard_parameters('pay', post, options[:unique_id])
         add_3ds(post, options)
+        add_tx_source(post, options)
 
         commit(post)
       end
@@ -83,6 +84,7 @@ module ActiveMerchant #:nodoc:
         add_amount(post, money, options)
         add_advanced_user(post)
         add_standard_parameters('capture', post, options[:unique_id])
+        add_tx_source(post, options)
 
         commit(post)
       end
@@ -99,6 +101,7 @@ module ActiveMerchant #:nodoc:
         add_amount(post, money, options)
         add_advanced_user(post)
         add_standard_parameters('refund', post, options[:unique_id])
+        add_tx_source(post, options)
 
         commit(post)
       end
@@ -110,6 +113,7 @@ module ActiveMerchant #:nodoc:
 
         add_advanced_user(post)
         add_standard_parameters('voidAuthorisation', post, options[:unique_id])
+        add_tx_source(post, options)
 
         commit(post)
       end
@@ -239,6 +243,10 @@ module ActiveMerchant #:nodoc:
         post['3DSECI'] = options[:three_ds_eci] if options[:three_ds_eci]
         post['3DSenrolled'] = options[:three_ds_enrolled] if options[:three_ds_enrolled]
         post['3DSstatus'] = options[:three_ds_status] if options[:three_ds_status]
+      end
+
+      def add_tx_source(post, options)
+        post[:TxSource] = options[:tx_source] if options[:tx_source]
       end
 
       def add_creditcard(post, creditcard)
--- a/test/remote/gateways/remote_migs_test.rb
+++ b/test/remote/gateways/remote_migs_test.rb
@@ -16,6 +16,9 @@ class RemoteMigsTest < Test::Unit::TestCase
     @diners = credit_card('30123456789019',   :month => 5, :year => 2021, :brand => 'diners_club')
     @credit_card = @visa
 
+    @valid_tx_source = 'MOTO'
+    @invalid_tx_source = 'penguin'
+
     @options = {
       :order_id => '1',
       :currency => 'SAR'
@@ -102,6 +105,48 @@ class RemoteMigsTest < Test::Unit::TestCase
     # assert response = @gateway.refund(@amount, payment_response.authorization, @options)
     # refute_success response
     # assert_equal 'Approved', response.message
+  end
+
+  def test_purchase_passes_tx_source
+    # returns a successful response when a valid tx_source parameter is sent
+    assert good_response = @gateway.purchase(@amount, @credit_card, @options.merge(tx_source: @valid_tx_source))
+    assert_success good_response
+    assert_equal 'Approved', good_response.message
+
+    # returns a failed response when an invalid tx_source parameter is sent
+    assert bad_response = @gateway.purchase(@amount, @credit_card, @options.merge(tx_source: @invalid_tx_source))
+    assert_failure bad_response
+  end
+
+  def test_capture_passes_tx_source
+    # authorize the credit card in order to then run capture
+    assert auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+    assert_equal 'Approved', auth.message
+
+    # returns a successful response when a valid tx_source paramater is sent
+    assert good_response = @gateway.capture(@amount, auth.authorization, @options.merge(tx_source: @valid_tx_source))
+    assert_success good_response
+
+    # returns a failed response when an invalid tx_source parameter is sent
+    assert bad_response = @gateway.capture(@amount, auth.authorization, @options.merge(tx_source: @invalid_tx_source))
+    assert_failure bad_response
+  end
+
+  def test_void_passes_tx_source
+    # authorize the credit card in order to then run capture
+    assert auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+    assert_equal 'Approved', auth.message
+
+    # returns a successful response when a valid tx_source paramater is sent
+    assert good_response = @gateway.void(auth.authorization, @options.merge(tx_source: @valid_tx_source))
+    assert_success good_response
+    assert_equal 'Approved', good_response.message
+
+    # returns a failed response when an invalid tx_source parameter is sent
+    assert bad_response = @gateway.void(auth.authorization, @options.merge(tx_source: @invalid_tx_source))
+    assert_failure bad_response
   end
 
   def test_status
--- a/test/unit/gateways/migs_test.rb
+++ b/test/unit/gateways/migs_test.rb
@@ -3,18 +3,21 @@ require 'test_helper'
 class MigsTest < Test::Unit::TestCase
   def setup
     @gateway = MigsGateway.new(
-                 :login => 'login',
-                 :password => 'password',
-                 :secure_hash => '76AF3392002D202A60D0AB5F9D81653C'
+                 login: 'login',
+                 password: 'password',
+                 secure_hash: '76AF3392002D202A60D0AB5F9D81653C',
+                 advanced_login: 'advlogin',
+                 advanced_password: 'advpass'
                )
-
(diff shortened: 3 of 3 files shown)


test/unit/gateways/migs_test.rb
```diff
@@ -74,6 +76,26 @@ def test_scrub
     assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
   end
 
+  def test_purchase_passes_tx_source
+    @gateway.expects(:commit).with(has_entries(:TxSource => 'MOTO'))
```
jasonxp: We could make these expectations more DRY by factoring them out into a method. That way, if additional methods need this test in the future, we can just call that expectation method. Likewise, if we need to change how the expectation works, we'll be able to do it in one place.

For example, something like:

```
def expect_gateway_commit_with_tx_source
  @gateway.expects(:commit).with(has_entries(:TxSource => @tx_source))
end
```
with ```@tx_source``` defined once in the ```setup``` method, which is also more DRY since we could change the value in a single place if we needed to.

lib/active_merchant/billing/gateways/migs.rb
```diff
@@ -173,6 +177,7 @@ def purchase_offsite_url(money, options = {})
         post[:ReturnURL] = options[:return_url]
 
         add_standard_parameters('pay', post, options[:unique_id])
+        add_tx_source(post, options)
```
jasonxp: It looks like we don't have a test for this method (```purchase_offsite_url```). The test to add would just be that ```tx_source``` is added to the URL, which the method returns.

Once a test is added, you can print out the return value from ```purchase_offsite_url``` to see what the result should look like, and test that ```vpc_TxSource``` is present in that string. The return value will look something like this:

```
https://migs.mastercard.com.au/vpcpay?vpc_Amount=100&vpc_OrderInfo=1&vpc_Locale=en&vpc_ReturnURL=http%3A%2F%2Fwww.aol.com%2F&vpc_Version=1&vpc_Merchant=login&vpc_AccessCode=password&vpc_Command=pay&vpc_MerchTxnRef=fe9f392f1ea3e2fba612dd176acfd7ee&vpc_TxSource=MOTO&vpc_SecureHash=06355CAF2AB6AA8754583456F400926FAA1911EA4575E4743E85C8DA66FF4B71&vpc_SecureHashType=SHA256
```

leila-alderman: Do we know whether the `purchase_offsite_url` actually needs the `tx_source` parameter passed? I added it in, but I'm still not sure whether it should be there.

leila-alderman: Currently, my remote test only checks that the `purchase` action passes `tx_source`. Since I'm adding a unit test to ensure that every method can pass `tx_source`, should I also have remote tests for each method?

jasonxp: Based on the code comments, ```purchase_offsite_url``` appears to be used for a payment flow where the user clicks a link that goes to MiGS for the payment processing. It makes sense that all parameters for the transaction would have to appear in that URL (where they are query parameters). So, I think ```purchase_offsite_url``` will need to include the options in ```purchase``` (except for the credit card itself).

The remote tests call out to gateways to test how they respond to our requests. ```purchase_offsite_url``` just generates a URL string with no external communication, so a unit test is all we need there.

jasonxp: After digging slightly deeper, I see that there is an existing remote test that simulates clicking on the URL that ```purchase_offsite_url``` generates: https://github.com/activemerchant/active_merchant/blob/547d0ebb419f2070f8327ae30d1b0dc7187c7b94/test/remote/gateways/remote_migs_test.rb#L34. This is pretty cool. We could use that test as the basis for an additional one that tests the offsite URL with the ```vpc_TxSource``` option. The assertion would likely be the same as the existing test (i.e., we just need to confirm that a successful response still comes back when the new option is present).

leila-alderman: The `purchase_offsite_url` method, in addition to using `add_creditcard_type` instead of the `add_creditcard` method used by the `purchase` action, does not include the `add_3ds` method, so it's already not receiving all of the parameters that the `purchase` action does. 

Since the `purchase_offsite_url` method states that it "Generates a URL to redirect user to MiGS to process payment", it seems as though the source of the transaction would be readily apparent and potentially would not need to be specified. Is there a way to determine whether this method requires `tx_source` to be passed?

leila-alderman: > Currently, my remote test only checks that the `purchase` action passes `tx_source`.

My question here was not specifically in regards to the `purchase_offsite_url` method. Should I add separate remote tests for all of the core actions to which I added `add_tx_source`, i.e., `purchase`, `capture`, `refund`, and `void`?

jasonxp: Yep -- testing each of the calls against the real gateway would be good to confirm that they behave as expected. For the offsite URL, I'd expect the documentation to have this information (somewhere). If there's still uncertainty about it, I'd see including it as low risk assuming the remote test verifies that the gateway doesn't produce an error when it is there. The code as you wrote it only will include the parameter if the user has specified it, and the user  would likely only include it if is needed, so that makes it even less of a risk if we're not able to pinpoint that in the docs.

test/remote/gateways/remote_migs_test.rb
```diff
@@ -104,6 +104,16 @@ def test_refund
     # assert_equal 'Approved', response.message
   end
 
+  def test_passes_tx_source
+    # Purchase returns a successful response when a valid tx_source parameter is sent.
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(tx_source: 'MOTO'))
```
leila-alderman: As I'm going through and writing more remote tests, I was looking more closely at these lines. Why is this written as an `assert` statement? Should this `assert` be removed?

curiousepic: This is a fairly common pattern, but I haven't really examined why. It would give us more "resolution" for failures, since checking that this step fails is distinct from the logic of the `assert_success` check.

test/unit/gateways/migs_test.rb
```diff
@@ -5,16 +5,20 @@ def setup
     @gateway = MigsGateway.new(
                  :login => 'login',
                  :password => 'password',
-                 :secure_hash => '76AF3392002D202A60D0AB5F9D81653C'
+                 :secure_hash => '76AF3392002D202A60D0AB5F9D81653C',
+                 advanced_login: 'advlogin',
+                 advanced_password: 'advpass'
```
curiousepic: I'm not sure if Rubocop would catch this, but it's idiomatic to keep the same kind of notation here (either `asdf: 'qwer'` or :`asdf => 'qwer'` but not both.

leila-alderman: Thanks for catching that! I copied those from another part of the code. Is there any preference between keeping the hash rockets or changing over to the more modern syntax?

curiousepic: Probably prefer modern over rockets, but nothing concrete.

test/remote/gateways/remote_migs_test.rb
```diff
@@ -104,6 +104,16 @@ def test_refund
     # assert_equal 'Approved', response.message
```
leila-alderman: The remote test for the `refund` action was previously commented out and labelled as "not working." I uncommented out this test and confirmed that it still isn't working. Should I attempt to resolve this problem or simply forego writing a remote test for `refund` passing `tx_source`?

lib/active_merchant/billing/gateways/migs.rb
leila-alderman: @jasonxp The `test_server_purchase_url` method, which you posted above and suggested for use as the basis of a test for the `purchase_offsite_url` action passing `tx_source`, is the method that is returning the error that I've been overlooking when running the full remote test file. Since that method is failing, can you think of another way to write a test for `purchase_offsite_url` passing `tx_source`?

test/remote/gateways/remote_migs_test.rb
leila-alderman: Interesting; thanks for the insight. I'll leave these in for more test granularity.

test/unit/gateways/migs_test.rb
leila-alderman: Changed all of the unit test hash rockets to the modern syntax. ✅

lib/active_merchant/billing/gateways/migs.rb
jasonxp: In that case: @curiousepic mentioned that this particular flow isn't supported by Spreedly. I'd thus be inclined to not bother adding the option to ```purchase_offsite_url```, since the request to add ```vpc_TxSource``` originated with a Spreedly customer, and the additional testing would be inconvenient. Would you agree, @curiousepic?

test/remote/gateways/remote_migs_test.rb
jasonxp: Since the comment says that refunds were not working in the test environment, it sounds like we likely won't be able to do anything to fix it, as we are indeed limited by the environment settings. Foregoing the remote test sounds fine.

jasonxp:
Just a heads up: the latest changes are failing CI due to some linting complaints, which you can view by clicking through the Details of the checks.

lib/active_merchant/billing/gateways/migs.rb
curiousepic: I think that's the better of two suboptimal situations, yep 👍

leila-alderman: I've submitted a new commit that removes `add_tx_source` from the `purchase_offsite_url` method.

test/remote/gateways/remote_migs_test.rb
```diff
@@ -16,6 +16,9 @@ def setup
     @diners = credit_card('30123456789019',   :month => 5, :year => 2021, :brand => 'diners_club')
     @credit_card = @visa
 
+    @valid_tx_source = 'MOTO'
```
jasonxp: 👍

test/unit/gateways/migs_test.rb
```diff
@@ -3,18 +3,22 @@
...
+    @tx_source = 'MOTO'
 
     @options = {
-      :order_id => '1',
-      :billing_address => address,
-      :description => 'Store Purchase'
+      order_id: '1',
+      billing_address: address,
+      description: 'Store Purchase',
+      return_url: 'http://localhost:8080/payments/return'
```
jasonxp: I think you added ```return_url``` previously because you were testing ```purchase_offsite_url```. Since we are no longer modifying that method and ```return_url``` isn't used anywhere else, we should remove it from the ```@options``` hash here.

leila-alderman: 👍 Done!

PR #3270, opened 2019-07-10
BlueSnap: Default to not send amount on capture

molbrown:
Including an amount on capture can cause an issue if attempting to
capture full amount authorized and there are slight differences in
amount requested and amount actually authorized due to currency
exchange. BlueSnap will capture full amount authorized if no amount is
included in the capture request. Adds a flag for including the amount
only in the case of partial capture.

Continues ECS-454
Relates to ECS-430

Unit:
27 tests, 110 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
35 tests, 109 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -95,7 +95,7 @@ module ActiveMerchant
         commit(:capture, :put) do |doc|
           add_authorization(doc, authorization)
           add_order(doc, options)
-          add_amount(doc, money, options)
+          add_amount(doc, money, options) if options[:include_capture_amount] == true
         end
       end
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -35,6 +35,7 @@
 * CardConnect: Fix parsing of level 3 fields [hdeters] #3273
 * TrustCommerce: Support void after purchase [jknipp] #3265
 * Payflow: Support arbitrary level 2 + level 3 fields [therufs] #3272
+* BlueSnap: Default to not send amount on capture [molbrown] #3270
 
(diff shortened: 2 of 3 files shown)


curiousepic:
Unless we are defining the word differently, there doesn't currently seem to be a flag involved here. I was expecting a new `option` that activates the conditional for only including it in partial captures. I think we would want an actual flag, since AFAICT most adapters' `capture` methods seem to expect that an amount argument is present.

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -95,7 +95,7 @@ def capture(money, authorization, options={})
         commit(:capture, :put) do |doc|
           add_authorization(doc, authorization)
           add_order(doc, options)
-          add_amount(doc, money, options)
+          add_amount(doc, money, options) if options[:partial_amount] == true
```
molbrown: @curiousepic I believe I addressed what you were talking about, let me know if I missed what you meant.

curiousepic: Wow, I'm sorry, in fact I just misunderstood the original, it was fine. Though I think the option name could be more explicit, since "amount" kinda implies it should be a the actual amount, rather than a boolean flag. Perhaps "partial_capture"  or "capture_partial_amount", or even "include_capture_amount" (since it doesn't necessarily need to be partial)...

molbrown: Confusion is never a good sign. I think I like `include_capture_amount` since it is the most descriptive of what is happening here. I have re-named it so.

PR #3272, opened 2019-07-11
Payflow: support arbitrary L2/L3 fields

therufs:
Still needed: 

~[x] L2 implementation or refactor~
~[x] Line item management~ -- For expediency's sake, this can be done as per the original "like Stripe" idea, only passing a blob of XML instead of JSON.
~[3/3] Remote tests~
[] Unit tests (make sure we're sending all the XML we think we are)

Known issues: 

Not obvious whether all L2/L3 fields are supported or only those documented in XMLPay.
Remote tests pass, but don't (so far) include any info about whether the fields were received/understood.

### DIFF
--- a/lib/active_merchant/billing/gateways/payflow.rb
+++ b/lib/active_merchant/billing/gateways/payflow.rb
@@ -1,3 +1,4 @@
+require 'nokogiri'
 require 'active_merchant/billing/gateways/payflow/payflow_common_api'
 require 'active_merchant/billing/gateways/payflow/payflow_response'
 require 'active_merchant/billing/gateways/payflow_express'
@@ -187,7 +188,44 @@ module ActiveMerchant #:nodoc:
             end
           end
         end
-        xml.target!
+        add_level_two_three_fields(xml.target!, options)
+      end
+
+      def add_level_two_three_fields(xml_string, options)
+        if options[:level_two_fields] || options[:level_three_fields]
+          xml_doc = Nokogiri::XML.parse(xml_string)
+          %i[level_two_fields level_three_fields].each do |fields|
+            xml_string = add_fields(xml_doc, options[fields]) if options[fields]
+          end
+        end
+        xml_string
+      end
+
+      def check_fields(parent, fields, xml_doc)
+        fields.each do |k, v|
+          if v.is_a? String
+            new_node = Nokogiri::XML::Node.new(k, xml_doc)
+            new_node.add_child(v)
+            xml_doc.at_css(parent).add_child(new_node)
+          else
+            check_subparent_before_continuing(parent, k, xml_doc)
+            check_fields(k, v, xml_doc)
+          end
+        end
+        xml_doc
+      end
+
+      def check_subparent_before_continuing(parent, subparent, xml_doc)
+        unless xml_doc.at_css(subparent)
+          subparent_node = Nokogiri::XML::Node.new(subparent, xml_doc)
+          xml_doc.at_css(parent).add_child(subparent_node)
+        end
+      end
+
+      def add_fields(xml_doc, options_fields)
+        fields_to_add = JSON.parse(options_fields)
+        check_fields('Invoice', fields_to_add, xml_doc)
+        xml_doc.root.to_s
       end
 
       def build_check_request(action, money, check, options)
@@ -213,7 +251,7 @@ module ActiveMerchant #:nodoc:
             end
           end
         end
-        xml.target!
+        add_level_two_three_fields(xml.target!, options)
       end
 
       def add_credit_card(xml, credit_card, options = {})
--- a/test/remote/gateways/remote_payflow_test.rb
+++ b/test/remote/gateways/remote_payflow_test.rb
@@ -29,6 +29,27 @@ class RemotePayflowTest < Test::Unit::TestCase
       :routing_number => '111111118',
       :account_number => '1111111111'
     )
+
+    @l2_json = '{
+        "Tender": {
+          "ACH": {
+            "AcctType": "C",
+            "AcctNum": "6355059797",
+            "ABA": "021000021"
+          }
+        }
+      }'
+
+    @l3_json = '{
+        "Invoice": {
+          "Date": "20190104",
+          "Level3Invoice": {
+            "CountyTax": {"Amount": "3.23"}
+          },
+          "Items":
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -188,6 +189,53 @@ def build_credit_card_request(action, money, credit_card, options)
...
+            xml_doc.at_css(parent).add_child(new_node)
+          else
+            check_subparent_before_continuing(parent, k, xml_doc)
+            check_fields(k, v, xml_doc)
+          end
+        end
+        xml_doc
+      end
+
+      def check_subparent_before_continuing(parent, subparent, xml_doc)
```
therufs: this could probably have a somewhat shorter name :D

bayprogrammer: I like the name. Much merit to keeping things explicit. The computer doesn't mind the longer name, and humans who aren't familiar with the code will probably appreciate the extra detail. I recommend keeping it! 👍

lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -188,6 +189,53 @@ def build_credit_card_request(action, money, credit_card, options)
           end
         end
         xml.target!
```
bayprogrammer: So I'm not super familiar with `Builder`, but it looks like what `target!` does is return the built XML as a string so far? Since this call to `target!` is no longer the return value, unless there is some necessary side-effect from calling it before here before the final expression, the value returned from `xml.target!` would just be discarded. Am I reading things right?

lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -188,6 +189,53 @@ def build_credit_card_request(action, money, credit_card, options)
...
+      end
+
+      def add_level_two_three_fields(xml, options)
+        if options[:level_two_fields] || options[:level_three_fields]
+          xml_doc = Nokogiri::XML.parse(xml.target!) ## so it doesn't put on an xml instruction
+          %i[level_two_fields level_three_fields].each do |fields|
+            xml = add_fields(xml_doc, options[fields]) if options[fields]
+          end
+        end
+        xml
```
bayprogrammer: If `options[:level_two_fields]` and `options[:level_three_fields]` were omitted, then we'd jump straight down to here and return the Builder `xml` argument we were called with. In that case, we didn't convert the builder object to the string explicitly the same way we used to from `build_credit_card_request` with the final Builder `xml.target!` call. And thus the type of our return value changes for `build_credit_card_request` in the cases neither l2 or l3 data is used.

When this is called, we pretty much only care about the XML string built up by Builder so far, and want to further manipulate it via Nokogiri in the case these optional options are passed, right? Would it make more sense to then pass the result of `xml.target!` directly to this method, and then we know that whether we further modify it via Nokogiri or not, we can safely pass through our argument?

```
      def build_credit_card_request(action, money, credit_card, options)
        xml = Builder::XmlMarkup.new
        xml.tag! TRANSACTIONS[action] do
          # ...
        end
        add_level_two_three_fields(xml.target!, options)
      end

      def add_level_two_three_fields(xml_string, options)
        if options[:level_two_fields] || options[:level_three_fields]
          xml_doc = Nokogiri::XML.parse(xml_string)
          %i[level_two_fields level_three_fields].each do |fields|
            xml_string = add_fields(xml_doc, options[fields]) if options[fields]
          end
        end
        xml_string
      end
```

lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -188,6 +189,53 @@ def build_credit_card_request(action, money, credit_card, options)
...
         end
         xml.target!
+        add_level_two_three_fields(xml, options)
+      end
+
+      def add_level_two_three_fields(xml, options)
+        if options[:level_two_fields] || options[:level_three_fields]
+          xml_doc = Nokogiri::XML.parse(xml.target!) ## so it doesn't put on an xml instruction
+          %i[level_two_fields level_three_fields].each do |fields|
+            xml = add_fields(xml_doc, options[fields]) if options[fields]
```
bayprogrammer: If both l2 and l3 field options are passed, won't this effectively discard the former after we finish processing the latter? Is that intentional? (Like, are they mutually exclusive?) Or does the call to `add_fields` mutate the passed `xml_doc` such that the second time through the resultant string is built up in addition to the first pass (so that the final result includes the processing from both passes)? (I am suspecting the latter, but just want to make sure)

lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -188,6 +189,53 @@ def build_credit_card_request(action, money, credit_card, options)
...
+        check_fields('Invoice', fields, xml_doc)
+        xml_doc.root.to_s
+      end
+
+      def add_fields(xml_doc, options_fields)
+        if options_fields.is_a? Hash
+          fields = options_fields
+        elsif options_fields.is_a? String
+          fields = JSON.parse(options_fields)
+        end
```
bayprogrammer: Do we support the l2 and l3 fields being passed either as a Ruby Hash or a JSON string? I saw the latter in the remote tests, but not the former. I think maybe you said you were going to add some more tests; wouldn't hurt to add a test or two exercising the Hash path.

lib/active_merchant/billing/gateways/payflow.rb
bayprogrammer: Nevermind on this. It's clear to me upon further reading that we are updating the `xml_doc` with each pass, so the return value is always the built-up value regardless of whether we call `add_fields` once or twice (depending on the passed options). I see your tests also verify this behavior. 👍

lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -188,6 +189,53 @@ def build_credit_card_request(action, money, credit_card, options)
...
+        return unless fields
+        string_on_fields(xml_doc, fields)
```
bayprogrammer: Sorry if you get this comment twice, I think something got messed up on my end.

I would recommend simplifying this `add_fields` method if possible. Since `options_fields` is either a Hash or a String, and won't be `nil` since you guard against that before the call from `add_level_two_three_fields`, we can just use `options_fields` directly, parsing it first into a Hash if necessary.

One idea:

```
      def add_fields(xml_doc, options_fields)
        if options_fields.is_a? String
          options_fields = JSON.parse(options_fields)
        end
        string_on_fields(xml_doc, options_fields)
      end
```

lib/active_merchant/billing/gateways/payflow.rb
therufs: That is indeed what Builder is doing here. The secret side effect is that Nokogiri, which we're about to use to disassemble and reassemble the XML, doesn't know what to do with a Builder XML object. I didn't feel super enthusiastic about converting the whole method to Nokogiri given time constraints, so turning it into a string and passing it along seemed simplest.

lib/active_merchant/billing/gateways/payflow.rb
therufs: As it stands, this receives the string output of `xml.target!` and returns either that string or the string as manipulated by Nokogiri, so we do end up with a string of XML either way. I think this makes a lot more sense, though!

lib/active_merchant/billing/gateways/payflow.rb
therufs: Having messed around with the tests a bit, I think it is most straightforward to insist on them being a string. Have cleaned this up by getting rid of the ambiguously-named `string_on_fields` method.

therufs:
@bayprogrammer should be good to go! cc @curiousepic if interested :)

therufs:
Unit tests: 48 tests, 234 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests: 35 tests, 152 assertions, 9 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
74.2857% passed 
Failures seem to occur around recurring billing and fraud filters, both of which (per the tests) require special account setup.

PR #3278, opened 2019-07-16
CECAbank: Update encryption to SHA2

leila-alderman:
Updated the encryption method for the CECAbank gateway from SHA1 to
SHA2. This change also included updating the URLs for the gateway and
revising the remote tests.

CE-23

Unit:
7 tests, 36 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
5 tests, 17 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cecabank.rb
+++ b/lib/active_merchant/billing/gateways/cecabank.rb
@@ -1,7 +1,7 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class CecabankGateway < Gateway
-      self.test_url = 'http://tpv.ceca.es:8000'
+      self.test_url = 'https://tpv.ceca.es'
       self.live_url = 'https://pgw.ceca.es'
 
       self.supported_countries = ['ES']
@@ -13,14 +13,14 @@ module ActiveMerchant #:nodoc:
 
       #### CECA's MAGIC NUMBERS
       CECA_NOTIFICATIONS_URL = 'NONE'
-      CECA_ENCRIPTION = 'SHA1'
+      CECA_ENCRIPTION = 'SHA2'
       CECA_DECIMALS = '2'
       CECA_MODE = 'SSL'
       CECA_UI_LESS_LANGUAGE = 'XML'
       CECA_UI_LESS_LANGUAGE_REFUND = '1'
       CECA_UI_LESS_REFUND_PAGE = 'anulacion_xml'
-      CECA_ACTION_REFUND   = 'tpvanularparcialmente' # use partial refund's URL to avoid time frame limitations and decision logic on client side
-      CECA_ACTION_PURCHASE = 'tpv'
+      CECA_ACTION_REFUND   = 'anulaciones/anularParcial' # use partial refund's URL to avoid time frame limitations and decision logic on client side
+      CECA_ACTION_PURCHASE = 'tpv/compra'
       CECA_CURRENCIES_DICTIONARY = {'EUR' => 978, 'USD' => 840, 'GBP' => 826}
 
       # Creates a new CecabankGateway
@@ -168,8 +168,8 @@ module ActiveMerchant #:nodoc:
           'AcquirerBIN' => options[:acquirer_bin],
           'TerminalID' => options[:terminal_id]
         )
-        url = (test? ? self.test_url : self.live_url) + "/cgi-bin/#{action}"
-        xml = ssl_post(url, post_data(parameters))
+        url = (test? ? self.test_url : self.live_url) + "/tpvweb/#{action}.action"
+        xml = ssl_post("#{url}?", post_data(parameters))
         response = parse(xml)
         Response.new(
           response[:success],
@@ -242,7 +242,7 @@ module ActiveMerchant #:nodoc:
           CECA_NOTIFICATIONS_URL +
           CECA_NOTIFICATIONS_URL
         end
-        Digest::SHA1.hexdigest(signature_fields)
+        Digest::SHA2.hexdigest(signature_fields)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_cecabank_test.rb
```diff
@@ -36,20 +36,19 @@ def test_successful_refund
...
 
   def test_unsuccessful_refund
-    assert response = @gateway.refund(@amount, 'wrongreference', @options)
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success purchase
+
+    assert response = @gateway.refund(@amount, purchase.authorization, @options.merge(currency: 'USD'))
     assert_failure response
-    assert_equal 'ERROR', response.message
+    assert_match 'Error de comunicaciones', response.message
```
leila-alderman: I went with testing for the "Communications error" when sending incorrect information in the refund request. This seemed to me to be a much better alternative than testing for the "Unable to parse response" error that is received when the response is HTML (which could indicate any number of problems) caused by sending an incorrect `amount` or `authorization`.

test/remote/gateways/remote_cecabank_test.rb
```diff
@@ -36,20 +36,19 @@ def test_successful_refund
...
   end
 
   def test_invalid_login
-    gateway = CecabankGateway.new(
-      :merchant_id => '',
-      :acquirer_bin => '',
-      :terminal_id => '',
-      :key => ''
-    )
+    gateway = CecabankGateway.new(fixtures(:cecabank).merge(key: 'invalid'))
```
leila-alderman: Here, I updated the `gateway` definition to supply only an invalid `key` because when all of the other parameters were invalid, the response was ""Unable to parse the response. #<ArgumentError: invalid byte sequence in UTF-8>" along with a lot of other HTML. 

This way, the error message is "ERROR AL CALCULAR FIRMA." ("Error in calculating the signature."), which seems much more helpful and appropriate.

lib/active_merchant/billing/gateways/cecabank.rb
```diff
@@ -168,7 +168,7 @@ def commit(action, parameters)
           'AcquirerBIN' => options[:acquirer_bin],
           'TerminalID' => options[:terminal_id]
         )
-        url = (test? ? self.test_url : self.live_url) + "/cgi-bin/#{action}"
+        url = (test? ? self.test_url : self.live_url) + "#{action}.action?"
```
curiousepic: Minor nitpick, but I think technically the question mark is not part of the url, but rather initiates the query? Maybe we should add it as part of the xml below: `xml = ssl_post("#{url}?", post_data(parameters))` (if that actually works).

leila-alderman: Good point! I made that update, and all of the tests are still working fine.

lib/active_merchant/billing/gateways/cecabank.rb
```diff
@@ -1,8 +1,8 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class CecabankGateway < Gateway
-      self.test_url = 'http://tpv.ceca.es:8000'
-      self.live_url = 'https://pgw.ceca.es'
+      self.test_url = 'https://tpv.ceca.es/tpvweb/'
+      self.live_url = 'https://pgw.ceca.es/tpvweb/'
```
jasonxp: Just a heads up - I think we'll need to update the docs page (https://docs.spreedly.com/payment-gateways/cecabank/) with this new production URL.

leila-alderman: @jasonxp I'm confused about what exactly you think needs to be updated in the documentation.

The CECAbank documentation currently states that the API endpoint URL is `https://pgw.ceca.es`: 
image

The code above states that the production endpoint URL is still 'https://pgw.ceca.es'. I didn't think that we included paths as part of the documented production URL, especially considering that the previous code still relied on paths to get to the appropriate URL but were not explicitly documented as the endpoint URL:
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cecabank.rb#L171

curiousepic:
Good point on monitoring Lava when deploying. Get familiar with [how to roll forward](https://paper.dropbox.com/doc/How-to-Deploy-Core-and-Id--AhHRxow6zFvUlBIuDO30IT9NAg-1cGXDYWTwDm2wIFf3M8Wx#:uid=092356818449007780704863&h2=Rolling-Forward-in-case-of-adv ) in case 😄

lib/active_merchant/billing/gateways/cecabank.rb
jasonxp: From a cursory look, I saw the documented production URLs matching the live URLs in ActiveMerchant. For example, Adyen includes the path in the docs: https://docs.spreedly.com/payment-gateways/adyen/. @curiousepic might have some insight since he worked on that gateway recently.

leila-alderman: I see what you're saying now. Perhaps it makes more sense to instead move the `/tpvweb/` part of the URL down to line 171 where the rest of the path gets added to the URL?

jasonxp: I think it's fine either way in Active Merchant. Whether having the path would be beneficial in the documentation depends mostly on how customers use that piece of information. Feel free to :shipit:, as we could always tweak the docs later if we find out that would be useful.

curiousepic: There are some cases where we would want more than the root to show up in our Docs, such as when there might be versioning via url path. That means an issue in Docs rather than here though, I think.

PR #3279, opened 2019-07-16
Add 3DS values to paypal

nebdil:
Paypal currently does not accept 3DS values in AM. This PR passes the 3DS values, 1.x to Paypal.

Example usage from the Paypal [documentation](https://developer.paypal.com/docs/classic/paypal-payments-pro/integration-guide/3d-secure/) for 1.0

```
<Version>59.0</Version>
.
.
.
<ThreeDSecureRequest>
    <MpiVendor3ds>Y</MpiVendor3ds>
    <AuthStatus3ds>Y</AuthStatus3ds>
    <Cavv>jMKEKlqlJGiJARAbxMDZ5+fnFeg=</Cavv>
    <Eci3ds>02</Eci3ds>
    <Xid>TTVmdlFxbERYVXo5R1hrVUY5bjY=</Xid>
</ThreeDSecureRequest>
```

Next up this [PR](https://github.com/activemerchant/active_merchant/pull/3285) will be merged once Paypal is ready to accept 2.0 parameters

### DIFF
--- a/lib/active_merchant/billing/gateways/paypal.rb
+++ b/lib/active_merchant/billing/gateways/paypal.rb
@@ -90,12 +90,25 @@ module ActiveMerchant #:nodoc:
             xml.tag! 'n2:Payer', options[:email]
             add_address(xml, 'n2:Address', address)
           end
+
+          add_three_d_secure(xml, options) if options[:three_d_secure]
         end
       end
 
       def add_descriptors(xml, options)
         xml.tag! 'n2:SoftDescriptor', options[:soft_descriptor] unless options[:soft_descriptor].blank?
         xml.tag! 'n2:SoftDescriptorCity', options[:soft_descriptor_city] unless options[:soft_descriptor_city].blank?
+      end
+
+      def add_three_d_secure(xml, options)
+        three_d_secure = options[:three_d_secure]
+        xml.tag! 'ThreeDSecureRequest' do
+          xml.tag! 'MpiVendor3ds', 'Y'
+          xml.tag! 'AuthStatus3ds', three_d_secure[:trans_status] unless three_d_secure[:trans_status].blank?
+          xml.tag! 'Cavv', three_d_secure[:cavv] unless three_d_secure[:cavv].blank?
+          xml.tag! 'Eci3ds', three_d_secure[:eci] unless three_d_secure[:eci].blank?
+          xml.tag! 'Xid', three_d_secure[:xid] unless three_d_secure[:xid].blank?
+        end
       end
 
       def credit_card_type(type)
--- a/test/unit/gateways/paypal_test.rb
+++ b/test/unit/gateways/paypal_test.rb
@@ -611,6 +611,18 @@ class PaypalTest < Test::Unit::TestCase
     assert_equal(:processing_error, response.error_code)
   end
 
+  def test_3ds_version_1_request
+    stub_comms do
+      @gateway.purchase(@amount, @credit_card, @options.merge(three_d_secure_option))
+    end.check_request do |endpoint, data, headers|
+      assert_match %r{<n2:Version>124</n2:Version>}, data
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/paypal.rb
```diff
@@ -73,6 +73,11 @@ def build_sale_or_authorization_request(action, money, credit_card_or_referenced
         xml.target!
       end
 
+      def api_version(options)
+        return 214 if options.dig(:three_d_secure, :version) =~ /^2/
```
bbraschi: - I do not have much context but I feel misleading that the `API_VERSION = '124'` is alone, it give the impression that is *the* version.
- [Paypal's doc](https://developer.paypal.com/docs/classic/paypal-payments-pro/integration-guide/3d-secure/#integration-steps) uses `214.0` with decimal. Any significance?
- the request does not change otherwise between the two versions we have?

lib/active_merchant/billing/gateways/paypal.rb
```diff
@@ -98,6 +105,19 @@ def add_descriptors(xml, options)
         xml.tag! 'n2:SoftDescriptorCity', options[:soft_descriptor_city] unless options[:soft_descriptor_city].blank?
       end
 
+      def add_three_d_secure(xml, options)
+        three_d_secure = options[:three_d_secure]
+        xml.tag! 'ThreeDSecureRequest' do
+          xml.tag! 'MpiVendor3ds', 'Y'
+          xml.tag! 'AuthStatus3ds', three_d_secure[:trans_status] unless three_d_secure[:trans_status].blank?
```
bbraschi: Discussed with Filipe: `trans_status` is [required by Paypal](https://developer.paypal.com/docs/classic/paypal-payments-pro/integration-guide/3d-secure/#integration-steps).

lib/active_merchant/billing/gateways/paypal.rb
nebdil: - added a new constant for the version
- i originally did not think it's significant because in the docs they say to set that field to `214` and then in the example they use `214.0`, but it's better to pass whatever's used in the example. so thank you, i changed that
- as far as i can tell, request does not change, but it's required to add more fields if 3DS 1.x and even more fields if 2.x

lib/active_merchant/billing/gateways/paypal.rb
```diff
@@ -73,6 +73,11 @@ def build_sale_or_authorization_request(action, money, credit_card_or_referenced
         xml.target!
       end
 
+      def api_version(options)
+        return API_VERSION_3DS2 if options.dig(:three_d_secure, :version) =~ /^2/
```
elfassy: could we pass them the real version? (aka just `options.dig(:three_d_secure, :version)`?)

test/unit/gateways/paypal_test.rb
```diff
@@ -611,6 +611,31 @@ def test_error_code_with_no_mapping_returns_standardized_processing_error
     assert_equal(:processing_error, response.error_code)
   end
 
+  def test_3ds_version_1_request
+    stub_comms do
```
elfassy: can we do remote tests?

lib/active_merchant/billing/gateways/paypal.rb
elfassy: or
```ruby
options.dig(:three_d_secure, :version) || API_VERSION
```

elfassy: never mind, it's API versions, not 3DS version 👓

elfassy:
@nebdil please add the PayPal docs outlining the API to the description of this PR

test/unit/gateways/paypal_test.rb
nebdil: can do

PR #3280, opened 2019-07-17
Realex: Implement credit

leila-alderman:
Implemented support for `credit` (also known as `general_credit` in
Spreedly parlance) for the Realex gateway, including remote and unit
tests.

CE-52 / CE-58

Unit:
27 tests, 844 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
26 tests, 135 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@
 == HEAD
 * CardConnect: Move domain from gateway specific to gateway field [hdeters] #3283
 * Braintree Blue: Support for stored credentials [hdeters] #3286
+* Realex: Re-implement credit as general credit [leila-alderman] #3280
 
 == Version 1.96.0 (Jul 26, 2019)
 * Bluesnap: Omit state codes for unsupported countries [therufs] #3229
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -43,6 +43,7 @@ module ActiveMerchant
       def initialize(options = {})
         requires!(options, :login, :password)
         options[:refund_hash] = Digest::SHA1.hexdigest(options[:rebate_secret]) if options.has_key?(:rebate_secret)
+        options[:credit_hash] = Digest::SHA1.hexdigest(options[:refund_secret]) if options.has_key?(:refund_secret)
         super
       end
 
@@ -70,9 +71,9 @@ module ActiveMerchant
         commit(request)
       end
 
-      def credit(money, authorization, options = {})
-        ActiveMerchant.deprecated CREDIT_DEPRECATION_MESSAGE
-        refund(money, authorization, options)
+      def credit(money, creditcard, options = {})
+        request = build_credit_request(money, creditcard, options)
+        commit(request)
       end
 
       def void(authorization, options = {})
@@ -180,6 +181,22 @@ module ActiveMerchant
           xml.tag! 'autosettle', 'flag' => 1
           add_comments(xml, options)
           add_signed_digest(xml, timestamp, @options[:login], sanitize_order_id(options[:order_id]), amount(money), (options[:currency] || currency(money)), nil)
+        end
+        xml.target!
+      end
+
+      def build_credit_request(money, credit_card, options)
+        timestamp = new_timestamp
+        xml = Builder::XmlMarkup.new :indent => 2
+        xml.tag! 'request', 'timestamp' => timestamp, 'type' => 'credit' do
+          add_merchant_details(xml, options)
+          xml.tag! 'orderid', sanitize_order_id(options[:order_id])
+          add_amount(xml, money, options)
+          add_card(xml, credit_card)
+          xml.tag! 'refundhash', @options[:credit_hash] if @options[:credit_hash]
+          xml.tag! 'autosettle', 'flag' => 1
+          add_comments(xml, options)
+          add_signed_digest(xml, timestamp, @options[:login], sanitize_order_id(options[:order_id]), amount(money), (options[:currency] || currency(money)), credit_card.number)
         end
         xml.target!
       end
--- a/test/remote/gateways/remote_realex_test.rb
+++ b/test/remote/gateways/remote_realex_test.rb
@@ -75,15 +75,15 @@ class RemoteRealexTest < Test::Unit::TestCase
   end
 
   def test_realex_purchase_with_invalid_account
-    response = RealexGateway.new(fixtures(:realex_with_account)).purchase(@amount, @visa,
+    response = RealexGateway.new(fixtures(:realex_with_account).merge(account: 'invalid')).purchase(@amount, @visa,
       :order_id => generate_unique_id,
-      :description => 'Test Realex purchase with invalid acocunt'
+      :description => 'Test Realex purchase with invalid account'
     )
 
     assert_not_nil response
     assert_failure response
 
-    assert_equal '504', response.params['result']
+    assert_equal '506', response.params['result']
     assert_match %r{no such}i, response.message
   end
 
@@ -388,6 +388,40 @@ class RemoteRealexTest < Test::Unit::TestCase
     assert response.test?
     assert_equal '101', response.params['result']
     assert_match %r{DECLINED}i, response.message
+  end
+
+  def test_successful_credit
+    gateway_with_refund_password = RealexGateway.new(fixtures(:realex).merge(:refund_secret => 'refund'))
+
+    credit_response = gateway_with_refund_password.credit(@amount, @visa,
+      :order_id => generate_unique_id,
+      :description => 'Test Realex Credit',
+      :billing_address => {
+        :zip => '90210',
+        :country => 'US'
+      }
+    )
+
+    assert_not_nil credit_response
+    assert_success credit_response
+    assert credit_response.authorization.length > 0
+    assert_equal 'Successful', credit_response.message
+  end
+
+  def test_failed_credit
+    credit_response = @gateway.credit(@amount, @visa,
+      :order_id => generate_unique_id,
+      :description => 'Test Realex Credit',
+      :billing_address => {
+        :zip => '90210',
+        :country => 'US'
+      }
+    )
+
+    assert_not_nil credit_response
+    assert_failure credit_response
+    assert credit_response.authorization.length > 0
+    assert_equal 'Refund Hash not present.', credit_response.message
   end
 
   def test_maps_avs_and_cvv_response_codes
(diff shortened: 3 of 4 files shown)


test/remote/gateways/remote_realex_test.rb
```diff
@@ -82,7 +82,7 @@ def test_realex_purchase_with_invalid_account
     assert_not_nil response
     assert_failure response
 
-    assert_equal '504', response.params['result']
+    assert_equal '506', response.params['result']
```
leila-alderman: This test was failing before I made any changes. I just updated it the actual error response code that I was getting back.

lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -184,6 +185,28 @@ def build_refund_request(money, authorization, options)
...
+        xml = Builder::XmlMarkup.new :indent => 2
+        xml.tag! 'request', 'timestamp' => timestamp, 'type' => 'credit' do
+          add_merchant_details(xml, options)
+          xml.tag! 'orderid', sanitize_order_id(options[:order_id])
+          add_amount(xml, money, options)
+          add_card(xml, credit_card)
+          xml.tag! 'refundhash', @options[:credit_hash] if @options[:credit_hash]
+          xml.tag! 'autosettle', 'flag' => 1
+          add_signed_digest(xml, timestamp, @options[:login], sanitize_order_id(options[:order_id]), amount(money), (options[:currency] || currency(money)), credit_card.number)
+          if credit_card.is_a?(NetworkTokenizationCreditCard)
```
leila-alderman: I'm not sure whether I need to keep this `if` statement. It was part of the `purchase` request that I modeled this method after, but I don't know whether it's needed for `credit`.

lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -184,6 +185,28 @@ def build_refund_request(money, authorization, options)
...
+            add_network_tokenization_card(xml, credit_card)
+          else
+            add_three_d_secure(xml, options)
+          end
+          add_comments(xml, options)
+          add_address_and_customer_info(xml, options)
```
leila-alderman: I'm not sure whether I need to keep this line. It was part of the `purchase` request that I modeled this method after, but I don't know whether it's needed for `credit`.

lib/active_merchant/billing/gateways/realex.rb
molbrown: Based on the [Realex docs](https://developer.globalpay.com/#!/api/card-payments#api-credit), it doesn't appear this is needed. I would expect the `build_credit_request` is most similar to the `build_refund_request`, and I don't see it in the sample xml request in that doc.

I also don't believe 3DS is ever used for requests other than authorize/ purchase.

lib/active_merchant/billing/gateways/realex.rb
molbrown: Same as above comment, this might not be needed, according to what I see in the docs.

lib/active_merchant/billing/gateways/realex.rb
leila-alderman: Thanks for checking into this for me! I've removed this now.

lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -70,9 +71,9 @@ def refund(money, authorization, options = {})
         commit(request)
       end
 
-      def credit(money, authorization, options = {})
-        ActiveMerchant.deprecated CREDIT_DEPRECATION_MESSAGE
```
jasonxp: Have you looked yet into whether we need to do anything special now that we're removing support for the way this deprecated request originally operated? (i.e., it's changing from a refund for a particular authorization to a general credit).

test/remote/gateways/remote_realex_test.rb
jasonxp: Are you still getting that 506? On master, I'm getting a 504, so my test example fails when it checks for 506.

```
     82:     assert_not_nil response
     83:     assert_failure response
     84: 
  => 85:     assert_equal '506', response.params['result']
     86:     assert_match %r{no such}i, response.message
     87:   end
     88: 
test/remote/gateways/remote_realex_test.rb:85:in `test_realex_purchase_with_invalid_account'
<"506"> expected but was
<"504">
```

lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -43,6 +43,7 @@ class RealexGateway < Gateway
       def initialize(options = {})
         requires!(options, :login, :password)
         options[:refund_hash] = Digest::SHA1.hexdigest(options[:rebate_secret]) if options.has_key?(:rebate_secret)
+        options[:credit_hash] = Digest::SHA1.hexdigest(options[:credit_secret]) if options.has_key?(:credit_secret)
```
jasonxp: I've identified what I believe is a complication here. The user has to pass these values in as credentials in the Spreedly API call (see https://docs.spreedly.com/payment-gateways/realex/). In core, this corresponds to the code here: https://github.com/spreedly/core/blob/master/app/models/gateways/realex_gateway.rb#L10 and here: https://github.com/spreedly/core/blob/master/app/models/gateways/realex_gateway.rb#L18. There, ```refund_password``` is setup as the field that the user passes in to Spreedly, and that field gets converted into the ```rebate_secret``` credential. This seems to be incorrect, because the refund password and rebate password are separate credentials with Realex. In other words, we're calling the Spreedly "rebate" password a "refund" password, when it should really just be the "rebate" password.

So, the issue is what to call the new parameter that the user passes into Spreedly. If not for the existing implementation, I think it would be called ```refund_password```, since that's the name of the field with the user's Realex account. Since that name has been taken, one idea is to call it ```credit_password```, giving us something in core like:

```gateway_field :credit_password, :crypted => true```
```mode.credential :credit_secret, field: :credit_password```

So then the Spreedly user would pass the rebate password into ```<refund_password>``` (which it looks like they must already be doing), and the refund password into ```<credit_password>```. Not ideal, but would be documented, and would avoid breaking current implementations.

Can you please take a look to check my logic, and see if you have any other ideas?

File: test/remote/gateways/remote_realex_test.rb (hunk 1 above)
Comment (jasonxp) [reply test/remote/gateways/remote_realex_test.rb]: We determined that I was getting a 504 because I had the default credentials. So, 506 should indeed be the appropriate response code expectation.

File: lib/active_merchant/billing/gateways/realex.rb (hunk 4 above)
Comment (leila-alderman) [reply lib/active_merchant/billing/gateways/realex.rb]: Yes, I have asked Connect about this issue. https://spreedly.slack.com/archives/CF080246R/p1563548513083000

> IMO it's safe enough, nothing special needed except a mention in the commit, maybe make the commit title descriptive that it's a re-implementation (Re-implement credit as general credit, or something)

Comment (jasonxp) [reply lib/active_merchant/billing/gateways/realex.rb]: 👍

File: lib/active_merchant/billing/gateways/realex.rb (hunk 5 above)
Comment (leila-alderman) [reply lib/active_merchant/billing/gateways/realex.rb]: Slack conversation: https://spreedly.slack.com/archives/CJ2MQCAJY/p1563808343004700

You seem to be getting a little entangled in the terminology, so I've tried to break it down a bit more here:
Spreedly term: refund  == Realex term: rebate
  refund_password (in Core) == rebate_secret (in Active Merchant) == Rebate Password (in the Realex docs)
Spreedly term: general_credit == Realex term: credit
  credit_password (in Core) == refund_secret (in Active Merchant) == Refund Password (in the Realex docs)
I was previously using credit_secret in Active Merchant, which I have now changed to refund_secret since that seems to paradoxically be less confusing.

File: lib/active_merchant/billing/gateways/realex.rb [hunk 6]
```diff
@@ -43,6 +43,7 @@ class RealexGateway < Gateway
       def initialize(options = {})
         requires!(options, :login, :password)
         options[:refund_hash] = Digest::SHA1.hexdigest(options[:rebate_secret]) if options.has_key?(:rebate_secret)
+        options[:credit_hash] = Digest::SHA1.hexdigest(options[:refund_secret]) if options.has_key?(:refund_secret)
```
jasonxp: I think the inconsistencies that we discussed between the use of the terms refund/rebate/credit at Spreedly and Realex makes the naming here unavoidably confusing. I scratched my head a bit after not looking at this for a few days, trying to remember the logic behind this naming iteration. It would be helpful to have a comment here so that someone looking at this code a year down the road isn't confused by ```refund_secret``` being in ```credit_hash``` rather than ```refund_hash```, which seems counterintuitive. Can you please add one?

jasonxp:
Looks good. I added one note requesting a comment to help explain the refund/rebate/credit variable/naming situation for posterity. Feel free to :shipit: after that.

lib/active_merchant/billing/gateways/realex.rb
leila-alderman: @jasonxp I see what you're saying about the terminology being confusing, but I'm unsure of what sort of comment I can add to help with the confusion beyond saying "yep, it's weird, but it's right". This is my draft of a comment, but I'm not sure if it's what you're looking for or whether it would actually be helpful. 

```
        # Although this looks odd, according to the Realex gateway documentation, the `refund_secret` is indeed used by the `credit` action and not the `refund` action.
```

What do you think?

jasonxp: I looked at the docs again to make sure I had my head wrapped around this correctly, and I see the symmetry in your revised naming. Feel free to leave it as-is without the comment. Thanks for taking a stab at it.

CHANGELOG
```diff
@@ -46,6 +46,7 @@
 * Add Kosovo to the list of countries [AnotherJoSmith] #3226
 * Realex: Adds 3DS 1&2 support through external MPI [filipebarcos] #3284
 * PayPal: Adds 3DS 1 support through external MPI [nebdil] #3279
+* Realex: Re-implement credit as general credit [leila-alderman] #3280
```
molbrown: Move your changelog entry up to HEAD, since this was not included in the July 26 version. (Easy to miss)

test/remote/gateways/remote_realex_test.rb
```diff
@@ -75,7 +75,7 @@ def test_realex_purchase_with_invalid_login
   end
 
   def test_realex_purchase_with_invalid_account
-    response = RealexGateway.new(fixtures(:realex_with_account)).purchase(@amount, @visa,
+    response = RealexGateway.new(fixtures(:realex_with_account).merge(account: 'invalid')).purchase(@amount, @visa,
       :order_id => generate_unique_id,
       :description => 'Test Realex purchase with invalid acocunt'
```
molbrown: invalid acocunt 😆

leila-alderman: Ahh, good catch! Apparently that's been in the code for 12 years now. 🤣

CHANGELOG
leila-alderman: Thanks for catching that! So grateful for your re-review!

PR #3283, opened 2019-07-18
CardConnect: fix domain GSF

hdeters:
Retrieve domain from gateway specific fields rather than at the
top gateway level

ECS-448

Unit:
22 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
23 tests, 55 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* CardConnect: Move domain from gateway spefici to gateway field
 
 == Version 1.96.0 (Jul 26, 2019)
 * Bluesnap: Omit state codes for unsupported countries [therufs] #3229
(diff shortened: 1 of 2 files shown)


hdeters:
Yeah, that's definitely doable. I don't have a ton of context around GSF vs gateway_field so I went with the path of least resistance. Happy to change it up if that's the consensus.

molbrown:
Based on what I understand `domain` is, I agree that it should instead be changed to a gateway field in Core.

For a little context, GSFs must get passed with every transaction, so are useful for things that could be unique for every transaction or only required for some types of transactions. `gateway_field`s are saved on the gateway, and passed along with every transaction. Much easier UX for things that are unchanging.

hdeters:
@molbrown sounds good. Do you think we should make it a required field?

PR #3289, opened 2019-07-31
Paymill: Add currency and amount to store requests

jasonxp:
Paymill now requires a currency and amount to be specified for tokenization. These are used
for an authorization. We now set this authorization amount to $1 USD -- the minimum allowed -- in our ```store``` requests. 

The ```currency``` and ```money``` options  are translated to ```presentation.amount3D``` and ```presentation.currency3D``` in the request made to Paymill.

### DIFF
--- a/lib/active_merchant/billing/gateways/paymill.rb
+++ b/lib/active_merchant/billing/gateways/paymill.rb
@@ -48,6 +48,11 @@ module ActiveMerchant #:nodoc:
       end
 
       def store(credit_card, options={})
+        # The store request requires a currency and amount of at least $1 USD.
+        # This is used for an authorization that is handled internally by Paymill.
+        options[:currency] = 'USD'
+        options[:money] = 100
+
         save_card(credit_card, options)
       end
 
--- a/test/unit/gateways/paymill_test.rb
+++ b/test/unit/gateways/paymill_test.rb
@@ -181,6 +181,20 @@ class PaymillTest < Test::Unit::TestCase
     assert response.test?
   end
 
+  def test_store_includes_currency_and_amount
+    expected_currency = 'USD'
+    expected_amount = 100
+
+    @gateway.expects(:raw_ssl_request).with(
+      :get,
+      store_endpoint_url(@credit_card, expected_currency, expected_amount),
+      nil,
+      {}
+    ).returns(successful_store_response, successful_purchase_response)
+
+    @gateway.store(@credit_card)
+  end
+
   def test_failed_store_with_invalid_credit_card
     @gateway.expects(:raw_ssl_request).returns(failed_store_response)
     response = @gateway.store(@credit_card)
@@ -224,6 +238,10 @@ class PaymillTest < Test::Unit::TestCase
   end
 
   private
+
+  def store_endpoint_url(credit_card, currency, amount)
+    "https://test-token.paymill.com?account.holder=#{credit_card.first_name}+#{credit_card.last_name}&account.number=#{credit_card.number}&account.expiry.month=#{'%02d' % credit_card.month}&account.expiry.year=#{credit_card.year}&account.verification=#{credit_card.verification_value}&presentation.amount3D=#{amount}&presentation.currency3D=#{currency}&channel.id=PUBLIC&jsonPFunction=jsonPFunction&transaction.mode=CONNECTOR_TEST"
+  end
 
   def successful_store_response
     MockResponse.new 200, %[jsonPFunction({"transaction":{"mode":"CONNECTOR_TEST","channel":"57313835619696ac361dc591bc973626","response":"SYNC","payment":{"code":"CC.DB"},"processing":{"code":"CC.DB.90.00","reason":{"code":"00","message":"Successful Processing"},"result":"ACK","return":{"code":"000.100.112","message":"Request successfully processed in 'Merchant in Connector Test Mode'"},"timestamp":"2013-02-12 21:33:43"},"identification":{"shortId":"1998.1832.1612","uniqueId":"tok_4f9a571b39bd8d0b4db5"}}})]
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 * CardConnect: Move domain from gateway specific to gateway field [hdeters] #3283
 * Braintree Blue: Support for stored credentials [hdeters] #3286
 * Realex: Re-implement credit as general credit [leila-alderman] #3280
+* Paymill: Add currency and amount to store requests [jasonxp] #3289
 
 == Version 1.96.0 (Jul 26, 2019)
 * Bluesnap: Omit state codes for unsupported countries [therufs] #3229


curiousepic:
Can/should the amount be voided afterward? If so, we can use a `MultiResponse` to perform a void after the store.

molbrown:
Was about to ask the same question ☝️

test/unit/gateways/paymill_test.rb
```diff
@@ -181,6 +181,32 @@ def test_successful_store
     assert response.test?
   end
 
+  def test_store_includes_currency
+    expected_currency = 'USD'
+
+    @gateway.expects(:raw_ssl_request).with(
```
leila-alderman: Looking through the other unit tests in this file, none of them use `.with`. Why is it needed for this unit test?

test/unit/gateways/paymill_test.rb
```diff
@@ -181,6 +181,32 @@ def test_successful_store
...
+      :get,
```
leila-alderman: What does this line do?

test/unit/gateways/paymill_test.rb
```diff
@@ -181,6 +181,32 @@ def test_successful_store
...
+      store_endpoint(@credit_card, expected_currency, 100),
+      nil,
+      {}
+    ).returns(successful_store_response, successful_purchase_response)
```
leila-alderman: Since it's been determined that the `store` call no longer works without passing both `currency` and `amount`, what is the reason for creating two separate tests ensuring that these values are passed as opposed to just testing that the `store` calls are successful?

test/unit/gateways/paymill_test.rb
```diff
@@ -181,6 +181,32 @@ def test_successful_store
...
+
+    @gateway.store(@credit_card)
+  end
+
+  def test_store_includes_amount
```
leila-alderman: Why have these set up as two separate tests? It looks like both of these tests are passing in both `currency` and `amount`, so I'm not seeing the difference between these tests.

test/unit/gateways/paymill_test.rb
```diff
@@ -225,6 +251,10 @@ def test_transcript_scrubbing
 
   private
 
+  def store_endpoint(credit_card, currency, amount)
```
leila-alderman: Why is this endpoint needed? It looks like there were already unit tests set up to test the `store` action, so why is this needed now?

lib/active_merchant/billing/gateways/paymill.rb
```diff
@@ -48,6 +48,11 @@ def void(authorization, options={})
       end
 
       def store(credit_card, options={})
+        # The store request requires a currency and amount of at least $1 USD.
+        # This is used for an authorization.
```
leila-alderman: I don't understand why this comment is talking about authorization within the `store` method. Is the issue that a `store` transaction for PayMill now runs an authorization first before storing the card? Do all `store` methods work that way?

jasonxp:
@curiousepic and @molbrown My understanding is that the "store" authorization is something that Paymill does internally. It doesn't appear to be possible to void it from our end, since we don't receive an authorization ID back in response. This is what the store response looks like: https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/paymill_test.rb#L229.

lib/active_merchant/billing/gateways/paymill.rb
jasonxp: Yep, that is the issue. The ```store``` requests for Paymill started resulting in errors when Paymill added ```currency``` and ```amount``` as required fields for the ```store``` operation. I can tweak the comment to be more explicit in saying that the authorization is something Paymill does internally. I'm not aware of any other gateway ```store```s that work this way.

test/unit/gateways/paymill_test.rb
jasonxp: This is a method that generates the endpoint URL that the ```store``` operation makes its request to. It's used by the new tests added here. I named the method as a noun for consistency with the existing helper methods like ```successful_store_response```, but it could also be named something like ```generate_store_endpoint```. I'll go with ```store_endpoint_url``` since that seems like it would be more clear.

test/unit/gateways/paymill_test.rb
jasonxp: The purpose of the test is to ensure that the currency gets included in the request sent to Paymill. So, the expectation is that when we call the method to make an SSL request, it's done ```with``` the currency parameter set to the right value.

test/unit/gateways/paymill_test.rb
jasonxp: ```store_endpoint``` is just a method that returns a string. That string is the second expected argument in the ```raw_ssl_request``` call.

leila-alderman: Oooh, I think I see now. So the `.with()`method requires _all_ of the parameters that are passed to the given function, meaning that you needed to pass in everything here:
```
def raw_ssl_request(method, endpoint, data, headers = {})
```
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/posts_data.rb#L47

That's where I had been confused. I didn't know what the four different parameters listed out here were doing.

test/unit/gateways/paymill_test.rb
leila-alderman: Okay, that makes sense. Why was this check added as a unit test but not as a remote test?

test/unit/gateways/paymill_test.rb
leila-alderman: Nice, thanks!

curiousepic:
Oh duh, it likely doesn't actually capture, and they just let the auth timeout. 👍

molbrown:
I do wonder why they ask for an amount though, if that is the case. You would think they could just do an internal $1 auth on any store request. I am skeptical that this could be for a purchase+store action, and I find their docs really challenging to navigate.

Do we have confirmation from Paymill/ other confidence that the store amount does what we think it does?

test/unit/gateways/paymill_test.rb
jasonxp: Good question. When Paymill responds to a ```store``` request, it doesn't return anything different in the case of a currency and amount being sent vs. not being sent. So, there's nothing that we can check in a remote test. The options are being set in the code that implements the request, but a user of that code cannot see those details. The remote test is performing its testing as if it were a user of the code, only calling the public methods (e.g., ```gateway.store```) and checking the return values. The more general theme here is unit testing vs. integration testing.

leila-alderman: I remember that there was a similar issue when trying to test the new `tx_source` GSF for the MiGS gateway. Would replicating the approach I took for those remote make sense here?
https://github.com/activemerchant/active_merchant/commit/4493b55b5bf49f854bcd43e08cae1b9e30b62eb2#diff-a1a05cf6a8356f8c6539f34382a90bf6R110

jasonxp: That is indeed a similar case. I don't think a test like that is possible in this case, though, because the options being passed are hardcoded in the ```store``` implementation. There isn't a way for the caller (i.e. our remote test) to change them to something invalid. FWIW, I did manually test that.

PR #3290, opened 2019-07-31
Stripe Payment Intents - Add new gateway

britth:
In preparation for Stripe 3DS2 support, this PR adds support for Stripe Payment Intents.
Stripe Payment Intents is added as a gateway, with functionality to create new PaymentIntents 
and Stripe PaymentMethods. PaymentIntents can also be confirmed, captured, cancelled, and 
updated. This gateway is implemented as a subclass of the Stripe gateway. To take advantage of 
the latest features, the API for this gateway is bumped up to the latest version, `2019-05-16`.

Remote:
23 tests, 99 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
100% passed

===

_Confirming that original Stripe tests also passing:_
_Remote:_
_67 tests, 313 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications_
_100% passed_

_Unit:_
_135 tests, 722 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications_
_100% passed_

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -1,0 +1,239 @@
+require 'active_support/core_ext/hash/slice'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    # This gateway uses the current Stripe {Payment Intents API}[https://stripe.com/docs/api/payment_intents].
+    # For the legacy API, see the Stripe gateway
+    class StripePaymentIntentsGateway < StripeGateway
+      ALLOWED_METHOD_STATES = %w[automatic manual].freeze
+      ALLOWED_CANCELLATION_REASONS = %w[duplicate fraudulent requested_by_customer abandoned].freeze
+      CREATE_INTENT_ATTRIBUTES = %i[description statement_descriptor receipt_email save_payment_method]
+      CONFIRM_INTENT_ATTRIBUTES = %i[receipt_email return_url save_payment_method setup_future_usage off_session]
+      UPDATE_INTENT_ATTRIBUTES = %i[description statement_descriptor receipt_email setup_future_usage]
+      DEFAULT_API_VERSION = '2019-05-16'
+
+      def create_intent(money, payment_method, options = {})
+        post = {}
+        add_amount(post, money, options, true)
+        add_capture_method(post, options)
+        add_confirmation_method(post, options)
+        add_customer(post, options)
+        add_payment_method_token(post, payment_method, options)
+        add_metadata(post, options)
+        add_return_url(post, options)
+        add_connected_account(post, options)
+        add_shipping_address(post, options)
+        setup_future_usage(post, options)
+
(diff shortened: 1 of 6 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -0,0 +1,176 @@
...
+        post = {}
+        add_payment_method_token(post, payment_method, options)
+        CONFIRM_INTENT_ATTRIBUTES.each do |attribute|
+          add_whitelisted_attribute(post, options, attribute)
+        end
+
+        commit(:post, "payment_intents/#{intent_id}/confirm", post, options)
+      end
+
+      # TODO: update to allow for card_present
```
jeremywrowe: We should probably not leave this in here

dinophile:
Hey @britth! This is awesome! We had connected with Stripe to see if someone out there had some work already done (and were trying to get something done on our end but you're waaay farther ahead than we got!). Closed our initial issue (3288) and will let this PR speak for itself!

Stripe will be super happy to see this!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -0,0 +1,185 @@
...
+        post
+      end
+
+      def add_confirmation_method(post, options)
+        confirmation_method = options[:confirmation_method].to_s
+        post[:confirmation_method] = confirmation_method if ALLOWED_METHOD_STATES.include?(confirmation_method)
+        post
+      end
+
+      def add_customer(post, options)
```
adamgrad: This method does not go along with `#Authorize` and will fail. Authorize calls `#create_post_for_auth_or_purchase`
and in there `#add_customer` is called: `add_customer(post, payment, options)`. Your method definition will override current implementation resulting in `wrong number of arguments (given 3, expected 2)` error

nfarve: @britth We might want to make sure the Stripe tests pass as well as the ones you added.

britth:
thanks all! just pushed up some updates for Payment Intents authorize/purchase actions, and reran Stripe tests to ensure that any Payment Intents updates didn't cause any failures for existing integrations

dinophile:
Could I make one more suggestion? Someone at Stripe suggested adding a comment and a console warning to the 'old' gateway to just let them know there are now two gateways that use different endpoints and that the new one uses the updated API. Just to make sure any smaller companies are aware of that and that if they have any need to switch to Payment Intents then they need to switch gateways! 👍 

Nothing major but they just wanted to make sure all the bases were covered for new or old users! What do you all think?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
nfarve: @britth were you able to fix this?

britth: yep, updates to authorize and purchase methods mean that this should be fixed

PR #3291, opened 2019-08-01
Realex: Fix an error that occurs when the rebate_secret or refund_secret is nil

jasonxp:
Fix an error that occurs when the ```rebate_secret``` or ```refund_secret``` option is set to ```nil```. In that case, the Realex gateway initializer attempts to calculate the refund hash or credit hash, passing ```nil``` to ```Digest::SHA1.hexdigest```, causing a type error. The fix is to only compute each hash if its secret has actually been provided.

### DIFF
--- a/test/unit/gateways/realex_test.rb
+++ b/test/unit/gateways/realex_test.rb
@@ -51,6 +51,43 @@ class RealexTest < Test::Unit::TestCase
     @amount = 100
   end
 
+  def test_initialize_sets_refund_and_credit_hashes
+    refund_secret = 'refund'
+    rebate_secret = 'rebate'
+
+    gateway = RealexGateway.new(
+      login: @login,
+      password: @password,
+      rebate_secret: rebate_secret,
+      refund_secret: refund_secret
+    )
+
+    assert gateway.options[:refund_hash] == Digest::SHA1.hexdigest(rebate_secret)
+    assert gateway.options[:credit_hash] == Digest::SHA1.hexdigest(refund_secret)
+  end
+
+  def test_initialize_with_nil_refund_and_rebate_secrets
+    gateway = RealexGateway.new(
+      login: @login,
+      password: @password,
+      rebate_secret: nil,
+      refund_secret: nil
+    )
+
+    assert_false gateway.options.key?(:refund_hash)
+    assert_false gateway.options.key?(:credit_hash)
+  end
+
+  def test_initialize_without_refund_and_rebate_secrets
+    gateway = RealexGateway.new(
+      login: @login,
+      password: @password
+    )
+
+    assert_false gateway.options.key?(:refund_hash)
+    assert_false gateway.options.key?(:credit_hash)
+  end
+
   def test_hash
     gateway = RealexGateway.new(
       :login => 'thestore',
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,7 @@
 * Braintree Blue: Support for stored credentials [hdeters] #3286
 * Realex: Re-implement credit as general credit [leila-alderman] #3280
 * Paymill: Add currency and amount to store requests [jasonxp] #3289
+* Realex: Prevent error calculating `refund_hash` or `credit_hash` when the secret is nil [jasonxp] #3291
 
 == Version 1.96.0 (Jul 26, 2019)
 * Bluesnap: Omit state codes for unsupported countries [therufs] #3229
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -42,8 +42,8 @@ module ActiveMerchant
 
       def initialize(options = {})
         requires!(options, :login, :password)
-        options[:refund_hash] = Digest::SHA1.hexdigest(options[:rebate_secret]) if options.has_key?(:rebate_secret)
-        options[:credit_hash] = Digest::SHA1.hexdigest(options[:refund_secret]) if options.has_key?(:refund_secret)
+        options[:refund_hash] = Digest::SHA1.hexdigest(options[:rebate_secret]) if options[:rebate_secret].present?
+        options[:credit_hash] = Digest::SHA1.hexdigest(options[:refund_secret]) if options[:refund_secret].present?
         super
       end
 


test/unit/gateways/realex_test.rb
```diff
@@ -51,6 +51,41 @@ def setup
...
+    rebate_secret = 'rebate'
+
+    options = {
+      login: @login,
+      password: @password,
+      rebate_secret: rebate_secret,
+      refund_secret: refund_secret
+    }
+
+    gateway = RealexGateway.new(options)
```
leila-alderman: The other gateways in this file and in other places that I've seen pass in the variables directly to the `Gateway.new` call, which I think looks cleaner and is less confusing. It's also nice to be consistent.

```suggestion
    gateway = RealexGateway.new(
      login: @login,
      password: @password,
      rebate_secret: rebate_secret,
      refund_secret: refund_secret
    )
```

test/unit/gateways/realex_test.rb
```diff
@@ -51,6 +51,41 @@ def setup
...
+
+    set_options = gateway.instance_variable_get(:@options)
+
+    assert set_options[:refund_hash] == Digest::SHA1.hexdigest(rebate_secret)
```
leila-alderman: I did some testing of this, and the extra step in line 67 isn't needed since you can actually call `options` directly.

```suggestion
    assert gateway.options[:refund_hash] == Digest::SHA1.hexdigest(rebate_secret)
```

test/unit/gateways/realex_test.rb
```diff
@@ -51,6 +51,41 @@ def setup
...
+
+  def test_initialize_with_nil_refund_and_rebate_secrets
+    options = {
+      login: @login,
+      password: @password,
+      rebate_secret: nil,
+      refund_secret: nil
+    }
+
+    gateway = RealexGateway.new(options)
```
leila-alderman: Should this made into an `assert` statement since we just want to test whether we can successfully create a new gateway object when passing in `nil` for the secrets?

test/unit/gateways/realex_test.rb
```diff
@@ -51,6 +51,41 @@ def setup
...
+
+    set_options = gateway.instance_variable_get(:@options)
+
+    assert_false set_options.key?(:refund_hash)
+    assert_false set_options.key(:credit_hash)
```
leila-alderman: If the above gateway creation is made into an assert statement, then are these tests still needed? I'm not sure that they really get at the behavior that we're trying to test here.

test/unit/gateways/realex_test.rb
jasonxp: Great catch -- I hadn't noticed the accessor here: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L144.

test/unit/gateways/realex_test.rb
jasonxp: Done.

test/unit/gateways/realex_test.rb
jasonxp: Another way to think about unit tests is that they define the expected behavior of the method. Failing to create a gateway object (due to a type error) when passing in ```nil``` for the secrets led us to add tests here, but the test more generally tests all of the expected behavior. So from a definitional standpoint, someone reading this test can see that a key for the each option won't be added to the options hash (rather than, say, a key being added and set to ```nil```). The asserts themselves also enforce that behavior, so there would be a failure if someone refactored ```initialize``` later such that it that added keys with ```nil``` or any other value.

test/unit/gateways/realex_test.rb
```diff
@@ -51,6 +51,33 @@ def setup
     @amount = 100
   end
 
+  def test_initialize_sets_refund_and_credit_hashes
+    refund_secret = 'refund'
+    rebate_secret = 'rebate'
+
+    gateway = RealexGateway.new({
```
leila-alderman: The curly braces aren't needed inside the `Gateway.new` calls.

test/unit/gateways/realex_test.rb
leila-alderman: This seems to be an answer to my other question.

This question was simply whether it be good to add an `assert` at the beginning of this line to create a test that confirms that the gateway is successfully created since that was the specific behavior that was breaking previously.

test/unit/gateways/realex_test.rb
leila-alderman: Why is it important for the behavior of this gateway that these hashes aren't added as keys to `options`? In my mind, this is an implementation detail rather than a description of the overall desired behavior.

test/unit/gateways/realex_test.rb
jasonxp: Hmm. Nothing comes to mind for me as far as a case where ```initialize``` could return ```nil```. In the scenario where we had it breaking before, you would just have the test fail due to the type error, so the ```assert``` would never get tested.

test/unit/gateways/realex_test.rb
jasonxp: Hmm. I think we only want these hashes added if the corresponding secrets are provided. If the implementation of ```initialize``` were to add keys with some bad value when the secrets are not provided, the rest of the code would behave as if a ```refundhash``` is available when it's not.

test/unit/gateways/realex_test.rb
jasonxp: Removed.

test/unit/gateways/realex_test.rb
leila-alderman: Adding the `assert` in at this point does return the `TypeError` and fails the test:
image

image

test/unit/gateways/realex_test.rb
leila-alderman: Maybe my question is more of a higher level one. Theoretically, my understanding is that tests should favor testing behavior over implementation. If I were to describe what this test is designed to check it would be "the gateway is successfully created even when `nil` is passed in as the values for the secrets." Based on that description, it seems to me that testing whether the `refund_hash` and `credit_hash` do not exist as keys in `@options` is more of an implementation detail that shouldn't necessarily be tested.

From an actual code perspective, it doesn't seem to be an issue at all if the hashes do get added with a `nil` value:
image

jasonxp: I designed the test to test the behavior of the ```initialize``` method. I'd say that setting/not setting fields relates to behavior, rather than implementation. The behavior is "what" the code does, whereas the implementation is "how" the code does it. So, you wouldn't want to have your test check that your method is calling some private method, because that's an implementation detail. The current code does pass when passing in a ```nil``` secret, but the test is covering the broader result of calling ```initialize```. So going back to my previous example, if someone were to come in and refactor ```initialize``` in such a way that a bad value were added for ```options[:refund_hash]``` instead of not adding a key/value pair at all, the test would fail.

leila-alderman: Could you explain what you mean by "a bad value"? I had thought that you meant `nil`, but it sounds like you might mean something else.

jasonxp: Sure. Any value calculated without a secret would be a "bad" value, because presumably there's no valid way to calculate a hash without a secret. So, hypothetically, if we switched to a different method for calculating the hash and it returned something (instead of erroring out like with the current method) when given ```nil```, we'd have a bogus hash, and the test would catch that unexpected behavior.

test/unit/gateways/realex_test.rb
bayprogrammer: For this test, for the original bug, you get that exception thrown whether the assert is added or not. The assert doesn't capture the exception, so it doesn't fail the assert it's just a normal unhandled exception getting surfaced.

Unless `RealexGateway.new` is redefined to do something very non-standard, it won't return `nil`. The constructor will always return some kind of object, or raise an error, in the common case.

I don't think the extra assert would add value in this specific case.

test/unit/gateways/realex_test.rb
bayprogrammer: I tend to agree that testing the observable behavior when initialized with bad input is useful here. Basically this test, with these assertions, is defining what the original intended behavior was: we only try to compute and store the hash values if the correct inputs are passed.

test/unit/gateways/realex_test.rb
```diff
@@ -51,6 +51,33 @@ def setup
...
+  def test_initialize_with_nil_refund_and_rebate_secrets
+    gateway = RealexGateway.new(
+      login: @login,
+      password: @password,
+      rebate_secret: nil,
+      refund_secret: nil
+    )
+
+    assert_false gateway.options.key?(:refund_hash)
+    assert_false gateway.options.key(:credit_hash)
```
bayprogrammer: `key?` vs `key` typo (should probably use `key?` in both cases)

test/unit/gateways/realex_test.rb
```diff
@@ -51,6 +51,33 @@ def setup
...
+      password: @password,
+      rebate_secret: rebate_secret,
+      refund_secret: refund_secret
+    )
+
+    assert gateway.options[:refund_hash] == Digest::SHA1.hexdigest(rebate_secret)
+    assert gateway.options[:credit_hash] == Digest::SHA1.hexdigest(refund_secret)
+  end
+
+  def test_initialize_with_nil_refund_and_rebate_secrets
```
bayprogrammer: Would it make sense to do one more test variation here where we also assert that the hash computations are omitted when the `:rebate_secret` and `:refund_secret` options are actually omitted entirely (what the original guard did do correctly)?

test/unit/gateways/realex_test.rb
jasonxp: Good catch - fixed.

test/unit/gateways/realex_test.rb
jasonxp: Good point. Done.

test/unit/gateways/realex_test.rb
leila-alderman: Gotcha, thanks for the clarifications @bayprogrammer!

test/unit/gateways/realex_test.rb
leila-alderman: Thanks for clarifying what behavior we're trying to test @bayprogrammer!

test/unit/gateways/realex_test.rb
leila-alderman: Thanks!

test/unit/gateways/realex_test.rb
leila-alderman: Thanks for making this change!

PR #3295, opened 2019-08-06
MercadoPago: Add Cabal card type

leila-alderman:
Added the Cabal card type to Active Merchant and to the MercadoPago
gateway.

CE-64

Unit:
25 tests, 129 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
26 tests, 76 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
92.3077% passed

The `test_partial_capture` and
`test_successful_purchase_with_processing_mode_gateway` remote tests are
currently failing and were failing prior to making any of these changes.


lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -54,7 +54,7 @@ def void(authorization, options={})
 
       def verify(credit_card, options={})
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
+          r.process { authorize(500, credit_card, options) }
```
leila-alderman: When I started working on this gateway, the `test_successful_verify` and `test_failed_verify` remote tests were also failing. Through trial and error, I determined that both of these tests failed whenever the `authorize` amount here was set to anything less than `500` and succeeded otherwise. 

I'm very uncertain of whether this is an okay change to make and how to test whether the `verify` action is working as currently implemented.

jasonxp: The two tests you mentioned actually pass for me. I suspect that we must be using different values for ```access_token``` in ```fixtures.yml```. Mine is just the default, and has a comment that says "Working credentials, no need to replace." It is different from the one in our team password vault. Can you give it a shot with the access token in ```test/fixtures.yml``` if yours is different?


```
Jasons-MacBook-Pro:active_merchant jasonnappier$ ruby -I test test/remote/gateways/remote_mercado_pago_test.rb  -n test_successful_verify
Loaded suite test/remote/gateways/remote_mercado_pago_test
Started
.
Finished in 3.634304 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Jasons-MacBook-Pro:active_merchant jasonnappier$ ruby -I test test/remote/gateways/remote_mercado_pago_test.rb  -n test_failed_verify
Loaded suite test/remote/gateways/remote_mercado_pago_test
Started
.
Finished in 3.060677 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

curiousepic: Each MP account is tied to a specific country and currency. Each may have a minimum amount required for an auth.

leila-alderman: I've confirmed that I'm using a different access token than Jason. The access token that's included in the Active Merchant fixtures file differs from the access token that's in my local fixtures. Oddly, the access token that's in Active Merchant isn't included in the list of tokens saved in 1Password. Moreover, I don't know how to determine which country (and currency) that access token is associated with. 

Since the `verify` tests are passing for some tokens/countries, should I leave the implementation alone and update my fixtures file or should I leave this update to the `verify` method so that it also works for the token/country that I was testing against?

jasonxp: I don't think we should change the ```verify``` implementation in order to get tests to pass, since that could affect use in production in a way that is outside the scope of this request. It sounds like the ideal solution would be to add the default Mercado Pago access token from ```test/fixtures.yml``` to your fixtures, and to use the other access token in new Cabal tests that require it. It might also be helpful to add a comment to the ```verify``` tests indicating that they fail with some MP accounts when the amount is less than $5. That way, if someone is working with MP in the future and uses a different access token in ```fixtures.yml``` like you did, that can save him/her some time and head scratching.

jasonxp:
LGTM except for the amount change in ```verify```.

lib/active_merchant/billing/gateways/mercado_pago.rb
curiousepic: Agree with Jason 👍

leila-alderman: Change reverted.

PR #3296, opened 2019-08-08
CyberSource: Add support for issuer additionalData gateway-specific field

jasonxp:
Added support for the issuer additionalData CyberSource-specific field. The CyberSource XML schema has also been updated. Per the gateway's instructions, we are now using version 1.156 in the test environment, and version 1.155 in production.

### DIFF
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -69,6 +69,9 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
         :amount => 100
       }
     }
+
+    @issuer_additional_data = 'PR25000000000011111111111112222222sk111111111111111111111111111'
+    + '1111111115555555222233101abcdefghijkl7777777777777777777777777promotionCde'
   end
 
   def test_transcript_scrubbing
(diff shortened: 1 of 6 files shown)


test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -101,9 +104,15 @@ def test_network_tokenization_transcript_scrubbing
 
   def test_successful_authorization
     assert response = @gateway.authorize(@amount, @credit_card, @options)
-    assert_equal 'Successful transaction', response.message
-    assert_success response
-    assert response.test?
+    assert_successful_response(response)
+    assert !response.authorization.blank?
```
leila-alderman: Consider changing this to `assert_not` instead of `assert !`.

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -590,4 +604,11 @@ def test_verify_credentials
     assert !gateway.verify_credentials
   end
 
+  private
+
+  def assert_successful_response(response)
```
leila-alderman: It looks like this method wasn't extracted out from all of the other tests, leading to some tests still including all three lines (e.g., `test_successful_authorization_with_elo`) and others using this new private method. This seems like a good thing to make consistent.

test/remote/gateways/remote_cyber_source_test.rb
jasonxp: That's actually not code that I added, so I'll leave it alone for now since ```assert !``` is used consistently (15 times) throughout the tests.

test/remote/gateways/remote_cyber_source_test.rb
jasonxp: True, there were a lot of them. I refactored them all to DRY things out.

test/remote/gateways/remote_cyber_source_test.rb
jasonxp: (I'm slightly bewildered as to why GitHub shows this with a green highlight, but it is indeed in master here: https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_cyber_source_test.rb#L107).

leila-alderman: Is this something worth refactoring for readability?

bayprogrammer: I do like `assert_not` better in general too, but I think consistency within the same test case is more important, so I would stick with `assert !` if that's what's used throughout the rest of the nearby code.

leila-alderman: Yes, agreed that consistency is most important! My question was whether this is something that would be worth changing throughout this test file.

bayprogrammer: Ah, gotcha. Doesn't hurt either way in my opinion, but probably in this case it is a matter of deciding how far to go for the immediate ticket vs. leaving otherwise unrelated changes to a different refactor.

I know I always see more I want to do, but often have to determine what is necessary for the task at hand and constrain myself to that. 🙂

PR #3297, opened 2019-08-12
Add transfer_data and application_fee_amount to stripe gateway

GrzegorzStochel:
Add transfer_data and application_fee_amount to follow latest changes in creating destination charges: https://stripe.com/docs/connect/destination-charges


glongman:
👍

Carrigan:
@GrzegorzStochel thank you for your contribution here, and our apologies for not getting to this sooner! 

I reviewed this earlier today and found that there was already support for these fields in the `stripe_gateway_intents` gateway. That implementation had a couple of extra fields and different names, so we opted to use that implementation here in order to make sure that the two are in sync. The changes are live in master now from PR #3535 if you'd like to use them.

PR #3300, opened 2019-08-13
Stripe: Send customer name and address with 3DS 1.0

jknipp:
Add the cardholder name, address, email and phone to the card source
created for 3DS 1.0 transactions.

ECS-455

Unit:
135 tests, 722 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
67 tests, 313 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote 3DS:
12 tests, 88 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,4 +7,5 @@ gem 'rubocop', '~> 0.60.0', require: false
 group :test, :remote_test do
   # gateway-specific dependencies, keeping these gems out of the gemspec
   gem 'braintree', '>= 2.93.0'
+  gem 'mechanize'
 end
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -304,8 +304,8 @@ module ActiveMerchant #:nodoc:
         add_amount(post, money, options, true)
         post[:type] = type
         if type == 'card'
-          add_creditcard(post, payment, options)
-          post[:card].delete(:name)
+          add_creditcard(post, payment, options, true)
+          add_source_owner(post, payment, options)
         elsif type == 'three_d_secure'
           post[:three_d_secure] = {card: payment}
           post[:redirect] = {return_url: options[:redirect_url]}
@@ -351,6 +351,12 @@ module ActiveMerchant #:nodoc:
           add_creditcard(post, payment, options)
         end
 
+        add_charge_details(post, money, payment, options)
+        post
+      end
+
+      # Used internally by Spreedly to populate the charge object for 3DS 1.0 transactions
+      def add_charge_details(post, money, payment, options)
         if emv_payment?(payment)
           add_statement_address(post, options)
           add_emv_metadata(post, payment)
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -7,4 +7,5 @@ gem 'rubocop', '~> 0.60.0', require: false
 group :test, :remote_test do
   # gateway-specific dependencies, keeping these gems out of the gemspec
   gem 'braintree', '>= 2.93.0'
+  gem 'mechanize'
```
jknipp: We need mechanize so that we can complete the Stripe 3DS 1.0 authentication step for purchases. This is done through URL provide by Stripe and requires a button to be clicked to change the 3DS source from pending to chargeable.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -347,6 +347,12 @@ def create_post_for_auth_or_purchase(money, payment, options)
           add_creditcard(post, payment, options)
         end
 
+        add_charge_details(post, money, payment, options)
+        post
+      end
+
+      # Used internally by Spreedly to populate the charge object for 3DS 1.0 transactions
+      def add_charge_details(post, money, payment, options)
```
jknipp: I plan to simplify the code in Core by calling this method, rather than duplicating the logic. I felt a comment was necessary to indicate this shouldn't be renamed/removed.

test/remote/gateways/remote_stripe_3ds_test.rb
```diff
@@ -23,42 +26,29 @@ def setup
 
   def test_create_3ds_card_source
     assert response = @gateway.send(:create_source, @amount, @credit_card, 'card', @options)
-    assert_success response
-    assert_equal 'source', response.params['object']
-    assert_equal 'chargeable', response.params['status']
-    assert_equal 'required', response.params['card']['three_d_secure']
-    assert_equal 'card', response.params['type']
```
jknipp: I consolidated the common sections of assertions to their own methods.

lib/active_merchant/billing/gateways/stripe.rb
bayprogrammer: Just curious, because I was getting confused trying to understand the existing code. Because `create_source` it's not called internally in the class (unless I missed it), I am not 100% why it's a private method. Is `create_source` also a Spreedly-specific method?

This isn't really a review comment as much as a question as I try to understand what's going on. Still reviewing the PR.

jknipp: @bayprogrammer Yes, it is Spreedly specific. Our implementation of 3DS 1.0 uses the sources API. As you might have gathered from this PR, using a `source` with the `charges` API is different than the gateway adapter implementation of a charge. I was not the original author of `create_source`, but I can imagine it was a conscious decision to make it private.

https://github.com/spreedly/core/blob/master/app/models/gateways/stripe_gateway.rb#L88

PR #3302, opened 2019-08-14
Credorax: Enable selecting a processor

leila-alderman:
Adds gateway specific fields to the Credorax gateway to enable the user
to select a different processor, including adding a unit test and a
(failing) remote test.

Gateway specific fields added:
 - R1: processor name
 - R2: processor merchant ID

CE-35

Unit:
23 tests, 106 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
26 tests, 37 assertions, 21 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
19.2308% passed

The Credorax gateway integration was recently updated, and now many of
the remote tests that were previously passing are failing. It is
currently unknown why these tests are all failing. We have reached
out to Credorax to inquire about this new behavior.

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -135,6 +135,7 @@ module ActiveMerchant #:nodoc:
         add_echo(post, options)
         add_submerchant_id(post, options)
         add_transaction_type(post, options)
+        add_processor(post, options)
 
         commit(:purchase, post)
       end
@@ -149,6 +150,7 @@ module ActiveMerchant #:nodoc:
         add_echo(post, options)
         add_submerchant_id(post, options)
         add_transaction_type(post, options)
+        add_processor(post, options)
 
         commit(:authorize, post)
       end
@@ -160,6 +162,7 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, options)
         add_echo(post, options)
         add_submerchant_id(post, options)
+        add_processor(post, options)
 
         commit(:capture, post)
       end
@@ -171,6 +174,7 @@ module ActiveMerchant #:nodoc:
         add_echo(post, options)
         add_submerchant_id(post, options)
         post[:a1] = generate_unique_id
+        add_processor(post, options)
 
         commit(:void, post, reference_action)
       end
@@ -182,6 +186,7 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, options)
         add_echo(post, options)
         add_submerchant_id(post, options)
+        add_processor(post, options)
 
         commit(:refund, post)
       end
@@ -195,6 +200,7 @@ module ActiveMerchant #:nodoc:
         add_echo(post, options)
         add_submerchant_id(post, options)
         add_transaction_type(post, options)
+        add_processor(post, options)
 
         commit(:credit, post)
       end
@@ -303,7 +309,6 @@ module ActiveMerchant #:nodoc:
 
       def add_3d_secure_1_data(post, options)
         post[:i8] = build_i8(options[:eci], options[:cavv], options[:xid])
-        post[:r1] = 'CREDORAX'
       end
 
       def add_normalized_3d_secure_2_data(post, options)
@@ -315,7 +320,6 @@ module ActiveMerchant #:nodoc:
         )
         post[:'3ds_version'] = three_d_secure_options[:version]
         post[:'3ds_dstrxid'] = three_d_secure_options[:ds_transaction_id]
-        post[:r1] = 'CREDORAX'
       end
 
       def build_i8(eci, cavv=nil, xid=nil)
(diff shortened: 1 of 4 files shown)


test/unit/gateways/credorax_test.rb
```diff
@@ -276,6 +276,18 @@ def test_supports_billing_descriptor
     end.respond_with(successful_purchase_response)
   end
 
+  def test_adds_processor_fields
```
jasonxp: Since this test seems to be specific to the ```purchase``` operation, should the test method be named ```test_purchase_adds_processor_fields``` rather than ```test_adds_processor_fields```?

leila-alderman: I was just following the convention in this test file, which also includes `test_adds_3d_secure_fields`, `test_adds_3ds2_fields_via_normalized_hash`, `test_adds_default_cavv_when_omitted_from_normalized_hash`, `test_adds_a9_field`, `test_adds_submerchant_id`, and `test_supports_billing_descriptor`, all of which test only the `purchase` transaction type.

jasonxp:
One general comment about the test coverage. The implementation covers ```purchase```, ```authorize```, ```capture```, ```void```, ```refund```, and ```credit```, but the tests appear to only cover ```purchase```. As a litmus test for coverage, I would expect test failures if we commented out any of the ```add_processor``` calls to intentionally break support for one of the other operations. If that doesn't happen, it would imply that there is no test coverage for the other methods to guard against regressions. The common code for the tests can be factored out into a helper test method, similar to what we did in the CyberSource tests that you reviewed recently @leila-alderman.

leila-alderman:
Similar to my comment above, I was trying to follow the convention in this test file, which only tests `purchase` transactions in the tests for the gateway specific fields. 

I have the same questions about this recommendation as I wrote above. Perhaps this is a higher level discussion that would be beneficial to have outside of GitHub.

test/unit/gateways/credorax_test.rb
jasonxp: In this case, I see the lack of test coverage for the non-```purchase``` methods as technical debt rather than a convention that we'd want to follow. Perhaps it was originally an oversight or a conscious decision based on time constraints, but the impact is that we don't have tests to catch regressions that would prevent the GSFs from being added to the request for those operations. Having those kinds of things break isn't uncommon when refactoring code. So, having tests to cover the new code can be quite valuable in allowing us to refactor or make other changes in the future quickly and with confidence.

It's always nice to chip away at technical debt when we can. Virtually any codebase will have some, and Active Merchant is certainly no exception. That said, deciding the best way to spend your time is more of an art. This would be my personal take for this scenario if I were working on the ticket:

* Definitely add the tests to cover the new code that we're adding in this PR, since they are valuable in their own right.

* As an optional bonus, change the test method names that you mentioned to include the word ```purchase```. This makes things better by indicating that they're really only testing the ```purchase``` method, which can be helpful to someone skimming the code, or someone who runs into a test failure. It also addresses part of the inconsistency you noticed, and would be a quick win.

* Skip the task of going back to add test coverage for all of the other GSFs in non-```purchase``` methods. This is out of scope for the original ticket you're working on, and  seems like it would be time-consuming. You likely also wouldn't have the complete context that you have for the new code you actually added here, and there are probably more valuable things that you could do with your time at the moment.

When deciding whether to tackle tech debt that I run into, I usually consider those two factors of context and time. If I can't knock it out relatively quickly given the context that I have from the original work I was tasked with, I tend to leave it for another day. Another thing to keep in mind is that you can always enter a ticket for tech debt tasks that you believe are valuable. That way, we're able to prioritize the time-consuming ones against our other work.

leila-alderman: The additional unit tests have been added for the gateway specific fields. I haven't added in any additional remote tests since they are still not working and therefore can't be tested.

leila-alderman:
I'm re-requesting review for this one now that the tests are working. (!!!!!!) With the 3DS2 updates for Credorax, the implementation is slightly different. 

The fixture for `credorax_with_processor` has been added to the Spreedly 1Password note for Credorax.

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -334,6 +338,11 @@ def add_transaction_type(post, options)
         post[:a9] = options[:transaction_type] if options[:transaction_type]
       end
 
+      def add_processor(post, options)
+        post[:r1] = options[:processor] || 'CREDORAX'
```
jasonxp: It looks like with the previous implementation, no ```r1``` parameter would have been set for a non-3DS transaction, whereas now ```r1``` will be set to ```CREDORAX``` for a non-3DS transaction that does not have the new ```processor``` option. This seems like it would be okay, but I'm not specifically familiar with the option. Is that difference fine?

leila-alderman: My understanding is that `r1` needs to be sent through for all transactions, not just 3DS transactions, whenever selecting a processor is enabled in the gateway settings. 

What I don't know (and sort of assume is okay) is the effect of sending `r1` when selecting a processor is *not* enabled in the gateway settings.

PR #3303, opened 2019-08-14
Checkout_v2: Support for native 3DS2.0

nfarve:
Adds a method to verify a payment after its gone through 3DS
authentication.

Loaded suite test/unit/gateways/checkout_v2_test
..........................

26 tests, 116 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/remote/gateways/remote_checkout_v2_test

29 tests, 70 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -68,6 +68,10 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def verify_payment(authorization, option={})
+        commit(:verify_payment, authorization)
+      end
+
       def supports_scrubbing?
         true
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -128,19 +132,23 @@ def add_transaction_data(post, options = {})
       end
 
       def add_3ds(post, options)
-        if options[:three_d_secure]
+        if options[:three_d_secure] || options[:execute_threed]
```
jeremywrowe: Suggestion: For clarity remove the nested conditional putting the condition that applies to both options[:three_d_secure] and options[:execute_threed] at the top of the method. Placing the specific configuration of  options[:three_d_secure] below it. Something like -

```rb
def add_3ds(post, options)
  if options[:three_d_secure] || options[:execute_threed]
    post[:'3ds'] = {}
    post[:'3ds'][:enabled] = true
    post[:success_url] = options[:callback_url] if options[:callback_url]
    post[:failure_url] = options[:callback_url] if options[:callback_url]
  end

  if options[:three_d_secure]
    post[:'3ds'][:eci] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
    post[:'3ds'][:cryptogram] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
    post[:'3ds'][:version] = options[:three_d_secure][:version] if options[:three_d_secure][:version]
    post[:'3ds'][:xid] = options[:three_d_secure][:ds_transaction_id] || options[:three_d_secure][:xid]
  end
end
```

Thoughts?

nfarve: Definitely makes things more explicit

PR #3307, opened 2019-08-21
Decidir: Add Naranja Card

hdeters:
EVS-181

Adds Naranha card to Decidir gateway
Updated two tests to correspond to changed Decidir responses

Unit:
0 notifications
100% passed

Remote:
17 tests, 60 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -7,7 +7,7 @@ module ActiveMerchant #:nodoc:
       self.supported_countries = ['AR']
       self.money_format = :cents
       self.default_currency = 'ARS'
-      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club]
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :naranja]
 
       self.homepage_url = 'http://www.decidir.com'
       self.display_name = 'Decidir'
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -99,7 +99,8 @@ def scrub(transcript)
         transcript.
           gsub(%r((apikey: )\w+)i, '\1[FILTERED]').
           gsub(%r((\"card_number\\\":\\\")\d+), '\1[FILTERED]').
-          gsub(%r((\"security_code\\\":\\\")\d+), '\1[FILTERED]')
+          gsub(%r((\"security_code\\\":\\\")\d+), '\1[FILTERED]').
+          gsub(%r((\"emv_issuer_data\\\":\\\")\d+), '\1[FILTERED]')
```
dtykocki: Is the `emv_issuer_data` something new that comes through in the response?

hdeters: I'm not sure. I think so -- it seems to just be the security code repeated which causes the existing scrub test to fail. However, I didn't see anything about it in the docs when I looked. So not exactly sure what the reasoning behind it is.

PR #3317, opened 2019-08-29
Stripe Payment Intents - Update transfer_data option

britth:
Rather than requiring a user to send in a hash with transfer data, this PR changes
to use individual `transfer_destination` and `transfer_amount` fields instead.

Remote:
24 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -207,10 +207,10 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_connected_account(post, options = {})
-        return unless transfer_data = options[:transfer_data]
+        return unless options[:transfer_destination]
         post[:transfer_data] = {}
-        post[:transfer_data][:destination] = transfer_data[:destination] if transfer_data[:destination]
-        post[:transfer_data][:amount] = transfer_data[:amount] if transfer_data[:amount]
+        post[:transfer_data][:destination] = options[:transfer_destination]
+        post[:transfer_data][:amount] = options[:transfer_amount] if options[:transfer_amount]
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -207,10 +207,10 @@ def setup_future_usage(post, options = {})
       end
 
       def add_connected_account(post, options = {})
-        return unless transfer_data = options[:transfer_data]
```
bayprogrammer: Sorry if this was already discussed elsewhere, but wouldn't this be a backwards-incompatible change for other users of this gem? What's the driving need for us to change the option keys here?

britth: I noticed as I was about to add this as a GSF that it seems pretty rare for other gateways to have a hash as an option for something pretty simple like this set of fields, so I thought it might be good to simplify it. It _seems_ low risk since I just added this gateway within the last 2 weeks or so, but you're right that this would be backwards incompatible. To be safe, I can close this out and just handle on our end though in case anyone is already using this if you all think that's the better approach!

bayprogrammer: Ah, didn't realize it was a change to that recent an addition. If there hadn't been a gem release yet for that code I would say we're probably safe, but if it had already been released with that as part of the public API we'd probably want to err on the safe side.

britth: Yeah, doublechecked, and the gateway addition is at the very top of the changelog HEAD, so doesn't look like it's been released in a version yet

britth:
thanks @bayprogrammer !

PR #3318, opened 2019-08-29
BlueSnap: add standardized 3DS 2 auth fields

bayprogrammer:
Unit:
30 tests, 142 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
37 tests, 115 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-552

Closes #3318

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -223,6 +223,7 @@ module ActiveMerchant
         doc.send('merchant-transaction-id', truncate(options[:order_id], 50)) if options[:order_id]
         doc.send('soft-descriptor', options[:soft_descriptor]) if options[:soft_descriptor]
         add_description(doc, options[:description]) if options[:description]
+        add_3ds(doc, options[:three_d_secure]) if options[:three_d_secure]
         add_level_3_data(doc, options)
       end
 
@@ -235,6 +236,22 @@ module ActiveMerchant
         doc.address(address[:address]) if address[:address]
         doc.city(address[:city]) if address[:city]
         doc.zip(address[:zip]) if address[:zip]
+      end
+
+      def add_3ds(doc, three_d_secure_options)
+        eci = three_d_secure_options[:eci]
+        cavv = three_d_secure_options[:cavv]
+        xid = three_d_secure_options[:xid]
+        ds_transaction_id = three_d_secure_options[:ds_transaction_id]
+        version = three_d_secure_options[:version]
+
+        doc.send('three-d-secure') do
+          doc.eci(eci) if eci
+          doc.cavv(cavv) if cavv
+          doc.xid(xid) if xid
+          doc.send('three-d-secure-version', version) if version
+          doc.send('ds-transaction-id', ds_transaction_id) if ds_transaction_id
+        end
       end
 
       def add_level_3_data(doc, options)
--- a/test/unit/gateways/blue_snap_test.rb
+++ b/test/unit/gateways/blue_snap_test.rb
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 require 'test_helper'
 
 class BlueSnapTest < Test::Unit::TestCase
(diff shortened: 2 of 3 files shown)


test/unit/gateways/blue_snap_test.rb
```diff
@@ -55,6 +65,37 @@ def test_successful_echeck_purchase
...
+      assert_match(/<cavv>#{Regexp.quote(@options_3ds2[:three_d_secure][:cavv])}<\/cavv>/, data)
+      assert_match(/<xid>#{Regexp.quote(@options_3ds2[:three_d_secure][:xid])}<\/xid>/, data)
+      assert_match(/<three-d-secure-version>#{Regexp.quote(@options_3ds2[:three_d_secure][:version])}<\/three-d-secure-version>/, data)
+      assert_match(/<ds-transaction-id>#{Regexp.quote(@options_3ds2[:three_d_secure][:ds_transaction_id])}<\/ds-transaction-id>/, data)
+    end.respond_with(successful_purchase_with_3ds_auth_response)
+
+    assert_success response
+    assert_equal '1024951831', response.authorization
+    assert_equal '019082915501456', response.params['original-network-transaction-id']
+    assert_equal '019082915501456', response.params['network-transaction-id']
```
bayprogrammer: I've observed that the network transaction id values are included in the responses to the purchase and auth requests which include third-party 3DS auth data. These fields are duplicated in different places in the response XML document that is returned, but they seem to be referring to the same value.

Since our current response parsing strategy flattens out the returned XML document, _if_ the fields with the same names but at different paths ever did need to be distinguished, we will need to do more work. Decided it would be better to wait until we find we need to do that extra work rather than try to preemptively do it now.

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -237,6 +238,22 @@ def add_address(doc, options)
...
       end
 
+      def add_3ds(doc, three_d_secure_options)
+        eci = three_d_secure_options[:eci]
+        cavv = three_d_secure_options[:cavv]
+        xid = three_d_secure_options[:xid]
+        ds_transaction_id = three_d_secure_options[:ds_transaction_id]
+        version = three_d_secure_options[:version]
+
+        doc.send('three-d-secure') do
```
molbrown: 👍  the conditional when `add_3ds` is called prevents sending an empty `three-d-secure` element.

test/unit/gateways/blue_snap_test.rb
molbrown: Will there be a situation that they would have different values? If they are referring to the same value, I agree that it doesn't seem like a problem.

bayprogrammer: @molbrown Yah, that was sort of bothering me, so I just dug a little deeper into BlueSnap's documentation to verify. It does appear to be the case that `network-transaction-id` is just duplicated in two different places in the XML response. From [their documentation](https://developers.bluesnap.com/v8976-JSON/docs/auth-capture#section--3-d-secure-examples-):

>### 3-D Secure Examples
>#### 3DS with shopper-initiated transaction
>
>The request includes `"transactionInitiator":"SHOPPER"` to indicate that this is a Customer-Initiated Transaction. The response includes the same `networkTransactionId` in both the `networkTransactionInfo` and the `processingInfo` sections. The merchant keeps this `networkTransactionId` for future transactions.
>
>#### 3DS with MIT
>
>The request includes `"transactionInitiator":"MERCHANT"` to indicate that this is a Merchant-Initiated Transaction. The merchant also includes the `originalNetworkTransactionId` to link the transaction to an existing transaction on which SCA was performed so SCA does not need to be performed again. In the `networkTransactionInfo` section, the response includes both the `originalNetworkTransactionId` and the new `networkTransactionId`. The merchant continues using the `originalNetworkTransactionId` for future transactions.

So I think we're good to go on this point. One more thing that occurs to me is: we may soon wish to implement the Stored Credentials framework for BlueSnap and as part of that ensure that original network transaction id is being returned with the AM response for follow-up requests.

PR #3319, opened 2019-08-30
Barclaycard Smartpay: Add support for shopper_statement gateway-specific field

jasonxp:


test/unit/gateways/barclaycard_smartpay_test.rb
```diff
@@ -182,11 +182,24 @@ def test_successful_authorize_with_shipping_house_number_and_street
...
+          shopper_interaction: shopper_interaction,
+          device_fingerprint: device_fingerprint,
+          shopper_statement: shopper_statement
+        )
+      )
     end.check_request do |endpoint, data, headers|
-      assert_match(/shopperInteraction=ContAuth/, data)
-      assert_match(/deviceFingerprint=abcde123/, data)
+      assert_match(/shopperInteraction=#{shopper_interaction}/, data)
+      assert_match(/shopperStatement=#{Regexp.quote(CGI.escape(shopper_statement))}/, data)
```
jasonxp: We escape all strings sent to the gateway here: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/barclaycard_smartpay.rb#L229.

PR #3321, opened 2019-09-03
Bambora Online ePay 3DS support

AllanWLie:
Adds 3DS support to the epay payment method.
---- Unit Test ----
Loaded suite test/unit/gateways/epay_test.rb
16 tests, 55 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
---- Remote Test ----
Loaded suite test/remote/gateways/remote_epay_test.rb
9 tests, 15 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/epay.rb
+++ b/lib/active_merchant/billing/gateways/epay.rb
@@ -63,6 +63,7 @@ module ActiveMerchant #:nodoc:
         add_invoice(post, options)
         add_creditcard_or_reference(post, credit_card_or_reference)
         add_instant_capture(post, false)
+        add_3ds_auth(post, options)
 
         commit(:authorize, post)
       end
@@ -74,6 +75,7 @@ module ActiveMerchant #:nodoc:
         add_creditcard_or_reference(post, credit_card_or_reference)
         add_invoice(post, options)
         add_instant_capture(post, true)
+        add_3ds_auth(post, options)
 
         commit(:authorize, post)
       end
@@ -158,6 +160,16 @@ module ActiveMerchant #:nodoc:
         post[:instantcapture] = option ? 1 : 0
       end
 
+      def add_3ds_auth(post, options)
+        if options[:three_d_secure]
+          post[:eci] = options.dig(:three_d_secure, :eci)
+          post[:xid] = options.dig(:three_d_secure, :xid)
+          post[:cavv] = options.dig(:three_d_secure, :cavv)
+          post[:threeds_version] = options.dig(:three_d_secure, :version)
+          post[:ds_transaction_id] = options.dig(:three_d_secure, :ds_transaction_id)
+        end
+      end
+
       def commit(action, params)
         response = send("do_#{action}", params)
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/epay.rb
```diff
@@ -158,6 +160,16 @@ def add_instant_capture(post, option)
         post[:instantcapture] = option ? 1 : 0
       end
 
+      def add_3ds_auth(post, options)
+        if options[:three_d_secure]
+          post[:eci] = options.dig(:three_d_secure, :eci)
+          post[:xid] = options.dig(:three_d_secure, :xid)
+          post[:cavv] = options.dig(:three_d_secure, :cavv)
+          post[:threeds_version] = options.dig(:three_d_secure, :version)
+          post[:ds_transaction_id] = options.dig(:three_d_secure, :ds_transaction_id)
```
filipebarcos: In many cases `xid` and `ds_transaction_id` are exclusive. Would it be ok to have one of these values being `nil`?

lib/active_merchant/billing/gateways/epay.rb
```diff
@@ -268,7 +279,7 @@ def xml_builder(params, soap_call)
 
       def authorize_post_data(params = {})
         params[:language] = '2'
-        params[:cms] = 'activemerchant'
+        params[:cms] = 'activemerchant_3ds'
```
filipebarcos: can we get the reasoning behind this change?

filipebarcos: Also, any chance we could support something like `application_id || 'activemerchant'`?https://github.com/activemerchant/active_merchant/blob/341b53eb0b3a129808389c5b4f7714b6a80d06cd/lib/active_merchant/billing/gateways/checkout_v2.rb#L98

test/remote/gateways/remote_epay_test.rb
```diff
@@ -5,12 +5,10 @@ def setup
     Base.mode = :test
 
     @gateway = EpayGateway.new(fixtures(:epay))
-
     @credit_card = credit_card('3333333333333000')
     @credit_card_declined = credit_card('3333333333333102')
-
     @amount = 100
-    @options = {order_id: '1'}
+    @options = {order_id: '1', three_d_secure: { eci: '7', xid: '123', cavv: '456', version: '2', ds_transaction_id: '798' }}
```
filipebarcos: we mostly never have xid and ds_transaction_id on the same request (same as before)

AllanWLie: eci, cavv and xid are required for 3ds-V1. The ds_transaction_id is only required when 3DS-V2 kicks in. We can handle empty values.

lib/active_merchant/billing/gateways/epay.rb
AllanWLie: Yes.

lib/active_merchant/billing/gateways/epay.rb
filipebarcos: `application_id` could be set by a user of Active Merchant to identify their platform. With a static value all requests using active merchant are grouped on the same bucket. If you had support for `application_id`, you could differentiate between "shopping cart" A and B

filipebarcos:
@AllanWLie you go approvals and unless you're planning to change anything I'll go ahead and merge these changes

PR #3322, opened 2019-09-03
Decidir: Add Cabal card

leila-alderman:
Adds the Cabal card type to the Decidir gateway.

With our current test gateway implementation, it does not seem possible
to test Cabal cards. When attempting to run a remote test with a Cabal
test card number, the error message states that the request contains an
invalid BIN (first 6 digits of the card number). Therefore, the addition
of Cabal cards has not been tested on the Decidir gateway.

CE-94 / CE-100

Unit:
0 notifications
100% passed

Remote:
17 tests, 60 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,7 @@
 * Stripe Payment Intents, Checkout V2: Add support for `MOTO` flagging [britth] #3323
 * Global Collect: Add Cabal card [leila-alderman] #3310
 * WorldPay: Add Cabal card [leila-alderman] #3316
+* Decidir: Add Cabal card [leila-alderman] #3322
 
 == Version 1.97.0 (Aug 15, 2019)
 * CyberSource: Add issuer `additionalData` gateway-specific field [jasonxp] #3296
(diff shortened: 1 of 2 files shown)


jknipp:
@leila-alderman Did you reach out to Decidir to see if they could enable Cabal test cards? They may be able to give us an API key to test this functionality.

jasonxp:
@jknipp Good question. Since we already launched Cabal support for a subset of gateways, Production is currently in an inconsistent state in that we have successful Cabal transactions with only some of the gateways that support the card. The current docs do not account for card types not working on all gateways. As a result, we decided to go with a strategy of getting support for Cabal released on the remaining gateways as quickly as possible, and following up with the gateways after release to get our test accounts working properly for additional remote tests. Does that seem reasonable?

jknipp:
@jasonxp Makes sense, thanks for the clarification that the decision to skip tests was intentional 👍

PR #3325, opened 2019-09-03
dLocal: Add Cabal card

leila-alderman:
Adds the Cabal card type to the dLocal gateway, including adding remote
tests to confirm that Cabal cards work on this gateway. The remote tests
include a failed purchase test to ensure that dLocal is validating the
test card.

CE-94 / CE-102

Unit:
17 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
26 tests, 72 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@
 * WorldPay: Add Cabal card [leila-alderman] #3316
 * Decidir: Add Cabal card [leila-alderman] #3322
 * PayU Latam: Add Cabal card [leila-alderman] #3324
+* dLocal: Add Cabal card [leila-alderman] #3325
 
 == Version 1.97.0 (Aug 15, 2019)
 * CyberSource: Add issuer `additionalData` gateway-specific field [jasonxp] #3296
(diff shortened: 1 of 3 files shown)


jasonxp:
@leila-alderman Is there any reason you chose to add unit tests for the card in gateways like PayU Latam, but not dLocal?

leila-alderman:
@jasonxp 
The main reason is that I was following the established conventions for the gateway integrations. 
The other reason was that as I was looking at the dLocal unit tests, I wasn't sure whether adding unit tests for the Cabal card type would actually add anything of value. Specifically, because the different transaction methods are already being tested by the existing unit tests, does writing additional tests with a different card number test anything new?

jasonxp:
> @jasonxp
> The main reason is that I was following the established conventions for the gateway integrations.
> The other reason was that as I was looking at the dLocal unit tests, I wasn't sure whether adding unit tests for the Cabal card type would actually add anything of value. Specifically, because the different transaction methods are already being tested by the existing unit tests, does writing additional tests with a different card number test anything new?

👍After looking at it more closely, I see that there is no "brand map" for this gateway as there is for PayU Latam, so I don't see anything different that would be unit tested here.

PR #3312, opened 2019-08-26
Adding 3DS2 passthru support to Braintree Blue

filipebarcos:

PR #3328, opened 2019-09-04
Braintree Blue: Adding 3DS2 passthru support

molbrown:
Continued from #3312 

Unit:
74 tests, 175 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
80 tests, 435 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -647,26 +647,23 @@ module ActiveMerchant #:nodoc:
 
       def add_3ds_info(parameters, three_d_secure_opts)
         return if empty?(three_d_secure_opts)
+        pass_thru = {}
 
-        parameters[:three_d_secure_pass_thru] = {
-          three_d_secure_version: three_d_secure_opts[:version],
-          eci_flag: three_d_secure_opts[:eci],
-          cavv_algorithm: three_d_secure_opts[:cavv_algorithm],
-          cavv: three_d_secure_opts[:cavv],
-          directory_response: three_d_secure_opts[:directory_response_status],
-          authentication_response: three_d_secure_opts[:authentication_response_status],
-        }.merge(xid_or_ds_trans_id(three_d_secure_opts))
+        pass_thru[:three_d_secure_version] = three_d_secure_opts[:version] if three_d_secure_opts[:version]
+        pass_thru[:eci_flag] = three_d_secure_opts[:eci] if three_d_secure_opts[:eci]
+        pass_thru[:cavv_algorithm] = three_d_secure_opts[:cavv_algorithm] if three_d_secure_opts[:cavv_algorithm]
+        pass_thru[:cavv] = three_d_secure_opts[:cavv] if three_d_secure_opts[:cavv]
+        pass_thru[:directory_response] = three_d_secure_opts[:directory_response_status] if three_d_secure_opts[:directory_response_status]
+        pass_thru[:authentication_response] = three_d_secure_opts[:authentication_response_status] if three_d_secure_opts[:authentication_response_status]
+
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -651,6 +645,32 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
...
+        pass_thru[:cavv] = three_d_secure_opts[:cavv] if three_d_secure_opts[:cavv]
+        pass_thru[:directory_response] = three_d_secure_opts[:directory_response_status] if three_d_secure_opts[:directory_response_status]
+        pass_thru[:authentication_response] = three_d_secure_opts[:authentication_response_status] if three_d_secure_opts[:authentication_response_status]
+
+        parameters[:three_d_secure_pass_thru] = pass_thru.merge(xid_or_ds_trans_id(three_d_secure_opts))
+      end
+
+      def xid_or_ds_trans_id(three_d_secure_opts)
+        if three_d_secure_opts[:version].to_f >= 2
+          {
```
filipebarcos: tiny nitpick, we should use one line hashes here `{ ds_transaction_id: three_d_secure_opts[:ds_transaction_id] }`

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -651,6 +645,28 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
...
+        pass_thru[:eci_flag] = three_d_secure_opts[:eci] if three_d_secure_opts[:eci]
+        pass_thru[:cavv_algorithm] = three_d_secure_opts[:cavv_algorithm] if three_d_secure_opts[:cavv_algorithm]
+        pass_thru[:cavv] = three_d_secure_opts[:cavv] if three_d_secure_opts[:cavv]
+        pass_thru[:directory_response] = three_d_secure_opts[:directory_response_status] if three_d_secure_opts[:directory_response_status]
+        pass_thru[:authentication_response] = three_d_secure_opts[:authentication_response_status] if three_d_secure_opts[:authentication_response_status]
+
+        parameters[:three_d_secure_pass_thru] = pass_thru.merge(xid_or_ds_trans_id(three_d_secure_opts))
+      end
+
+      def xid_or_ds_trans_id(three_d_secure_opts)
```
bayprogrammer: Does Braintree Blue treat `xid` and `ds_transaction_id` as mutually exclusive for passing 3DS2 auth data? I only ask because I've learned that BlueSnap requires both (yet I'm pretty sure I've seen other gateways that do treat them as mutually exclusive, so I just want to double-check). If Braintree /is/ treating as mutually exclusive, can we add a test that exercises this conditional logic both both cases (when `xid` wins, and when `ds_transaction_id` wins)?

molbrown: Unfortunately the sandbox environment is not informative in this way, but here's what I have from Braintree docs, which seems to follow the mutually exclusive pattern: 
> xid:      type: string      description: Transaction identifier resulting from 3D Secure authentication. Required on the transaction if returned by the MPI provider. This value is not used in 3D Secure 2 authentications. 
> ds_transaction_id:      type: string      description: The directory server transaction ID is used by the card brand's 3DS directory server. Required for 3D Secure 2 authentications.  
Adding a test to confirm that conditional behavior is a good idea. I have added a unit test. 👍

filipebarcos: That's what I heard from them, at least.

>  This value is not used in 3D Secure 2 authentications.

bayprogrammer: Ok, that reads to me like they're mutually exclusive for Braintree also. Thanks for double-checking! 🎉

PR #3329, opened 2019-09-04
Worldpay: Exemption flagging (MOTO)

britth:
This PR adds the necessary fields to mark transactions as MOTO 
by setting `dynamicInteractionType` to `MOTO` when the 
payment is marked as a manual entry ([requires configuration on 
the account level to be able to use](http://support.worldpay.com/support/kb/gg/corporate-gateway-guide-beta/content/directintegration/paymentrequests.htm#Flexible)).

Remote:
51 tests, 217 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
90.1961% passed

Unit:
66 tests, 392 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -207,6 +207,7 @@ module ActiveMerchant #:nodoc:
               if options[:instalments]
                 add_instalments_data(xml, options)
               end
+              add_moto_flag(xml, options) if options.dig(:metadata, :manual_entry)
             end
           end
         end
@@ -421,6 +422,10 @@ module ActiveMerchant #:nodoc:
           xml.tag! 'instalments', options[:instalments]
           xml.tag! 'cpf', options[:cpf] if options[:cpf]
         end
+      end
+
+      def add_moto_flag(xml, options)
+        xml.tag! 'dynamicInteractionType', 'type' => 'MOTO'
       end
 
       def address_with_defaults(address)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -23,6 +23,7 @@
 * PayU Latam: Add Cabal card [leila-alderman] #3324
 * dLocal: Add Cabal card [leila-alderman] #3325
 * BlueSnap: Add Cabal card [leila-alderman] #3326
+* Worldpay: Add support for MOTO flagging [britth] #3329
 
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -422,6 +428,14 @@ def add_instalments_data(xml, options)
         end
       end
 
+      def add_dynamic_interaction_type(xml, options)
+        xml.tag! 'dynamicInteractionType', 'type' => options[:dynamic_interaction_type]
+      end
+
+      def add_exemption(xml, options)
+        xml.tag! 'exemption', 'type' => options[:exemption], 'placement'=>'AUTHORISATION'
```
nfarve: I had something similar, but I think `placement` can be a lot of different things: https://github.com/activemerchant/active_merchant/compare/worldpay_3ds2?expand=1

britth: ah, nice! maybe I'll just remove the exemption field from this PR since you've got it covered there with the other 3ds2 additions, and just do MOTO in this PR

filipebarcos:
I see that we use different options to flag moto transactions across the code base? Some places we have `credit_card.manual_entry` like for https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/authorize_net.rb#L530-L532

I recently opened a PR to use `options[:metadata][:manual_entry]` but I could also change that for `credit_card.manual_entry` -- in our case those two places hold the same value (we do this on purpose for different reasons)

We could try to standardize this too, like we did for the other 3DS fields https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields

britth:
@filipebarcos that makes sense! I can update here to use `options[:metadata][:manual_entry]` if we want to standardize on that for MOTO flagging 👍

filipebarcos:
I'd love to have that standardized too. We're doing some of the MOTO work now like you can see here: https://github.com/activemerchant/active_merchant/pull/3332

pi3r:
Heads up this seems to be causing issues since some Worldpay account can't accept this parameter. Apparently `MOTO` doesn't work with their Business Gateway. We'll update this issue with more information once we figure out to fix it.

PR #3330, opened 2019-09-04
[Checkout.com] Using options[:metadata][:manual_entry] for Moto transactions

filipebarcos:
##### Unit tests
```
...........................
Finished in 0.012471 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
27 tests, 120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

##### Remote tests
```
...............................
Finished in 30.509284 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
31 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -128,7 +128,7 @@ module ActiveMerchant #:nodoc:
       def add_transaction_data(post, options = {})
         post[:payment_type] = 'Regular' if options[:transaction_indicator] == 1
         post[:payment_type] = 'Recurring' if options[:transaction_indicator] == 2
-        post[:payment_type] = 'MOTO' if options[:transaction_indicator] == 3
+        post[:payment_type] = 'MOTO' if options[:transaction_indicator] == 3 || options.dig(:metadata, :manual_entry)
         post[:previous_payment_id] = options[:previous_charge_id] if options[:previous_charge_id]
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -128,7 +128,7 @@ def add_customer_data(post, options)
       def add_transaction_data(post, options = {})
         post[:payment_type] = 'Regular' if options[:transaction_indicator] == 1
         post[:payment_type] = 'Recurring' if options[:transaction_indicator] == 2
-        post[:payment_type] = 'MOTO' if options[:transaction_indicator] == 3
+        post[:payment_type] = 'MOTO' if options[:transaction_indicator] == 3 || options[:moto]
```
manuelclaveras: Right now I think we only set moto in the stripe interactor. Looking at the code in core and cardserver, I think we should instead use `manual_entry` in `metadata`, this is set to true when it is a manual (draft) order.

elfassy: ` we only set moto in the stripe interactor` could we set it in the base interactor instead?

elfassy: either way it should be consistent across all gateways IMO

manuelclaveras: We could but that's what `manual_entry` is for IMHO

PR #3331, opened 2019-09-04
Adyen: Set 3DS exemptions via API

britth:
There are a few different ways to [request exemptions via Adyen](https://docs.adyen.com/payments-essentials/psd2-sca-compliance-and-implementation-guide?query=additionalData.executeThreeD#managingpsd2scacompliance).
Users can let Adyen handle compliance by default, they can
configure specific rules using dynamic 3dsecure, or they can set
exemption flags via the API. This PR updates the Adyen gateway to
allow users to set specific exemptions using the sca_exemption
field, used in conjunction with the execute_threed field. Users
can also now set execute_threed to false if they want to bypass
3DS on a transaction.

Remote:
64 tests, 207 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
39 tests, 188 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -336,6 +336,11 @@ module ActiveMerchant #:nodoc:
             add_browser_info(three_ds_2_options[:browser_info], post)
             post[:threeDS2RequestData] = { deviceChannel: device_channel, notificationURL: three_ds_2_options[:notification_url] }
           end
+
+          if options.has_key?(:execute_threed)
+            post[:additionalData][:executeThreeD] = options[:execute_threed]
+            post[:additionalData][:scaExemption] = options[:sca_exemption] if options[:sca_exemption]
+          end
         else
           return unless options[:execute_threed] || options[:threed_dynamic]
           post[:browserInfo] = { userAgent: options[:user_agent], acceptHeader: options[:accept_header] }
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -340,6 +340,10 @@ def add_3ds(post, options)
           post[:browserInfo] = { userAgent: options[:user_agent], acceptHeader: options[:accept_header] }
           post[:additionalData] = { executeThreeD: 'true' } if options[:execute_threed]
         end
+        if options[:sca_exemption] && !options[:execute_threed].nil?
```
bayprogrammer: This logic has some subtleties that might be worth locking down in the unit test suite. Could we add a couple tests which assert the expected shape of the outgoing request via a `check_request` block (a la `test_update_shopper_statement_and_industry_usage_sent`) given various combinations of `:sca_exemption` and `:execute_threed` options?

jeremywrowe: This can be simplified by using `Hash#has_key?` eg:

```rb
if options.has_key?(:sca_exception) && options.has_key?(:execute_threed)
  ...
end
```

britth: Good call 👍 Since from what i've seen, these exemptions are only relevant for 3ds2, moved this snippet of logic up to the conditional related to 3ds2 and added unit tests to check that values are sent as expected

PR #3337, opened 2019-09-09
Decidir: Add remote tests for Cabal and Naranja

leila-alderman:
Added remote tests for the new card types Cabal and Naranja now that
the Decidir test gateway has been updated to accept these card types.

In addition, added new logic to determine the correct
`payment_method_id` based on the card type.

With the update of the test gateway, some of the error messages changed,
so those were also updated in the remote tests.

CE-94 / CE-110

Unit:
0 notifications
100% passed

Remote:
19 tests, 66 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -106,7 +106,7 @@ module ActiveMerchant #:nodoc:
       private
 
       def add_auth_purchase_params(post, money, credit_card, options)
-        post[:payment_method_id] =  options[:payment_method_id] ? options[:payment_method_id].to_i : 1
+        post[:payment_method_id] = add_payment_method_id(credit_card)
         post[:site_transaction_id] = options[:order_id]
         post[:bin] = credit_card.number[0..5]
         post[:payment_type] = options[:payment_type] || 'single'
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -119,6 +119,18 @@ def add_auth_purchase_params(post, money, credit_card, options)
...
 
+      def add_payment_method_id(credit_card)
+        if options[:payment_method_id]
+          options[:payment_method_id].to_i
+        elsif CreditCard.brand?(credit_card.number) == 'cabal'
+          63
+        elsif CreditCard.brand?(credit_card.number) == 'naranja'
+          24
+        else
+          1
```
therufs: do i correctly infer that 1 means "use the Luhn algorithm" and other numbers mean to use some other algorithm?

leila-alderman: Not to my knowledge, no. Although Naranja doesn't use Luhn, Cabal cards do follow Luhn. I honestly don't know what the significance of these numbers are; they are just what the rep at Decidir told me to use in an email.

therufs: Herm, weird! Is the email in Zendesk? If so (or if in other easily accessible place) might be good to link to it from this PR, to provide context for posterity.

leila-alderman: Very excited to say that I was able to track this down in Decidir's documentation, which has been added to the commit message: https://decidirv2.api-docs.io/1.0/tablas-de-referencia-e-informacion-para-el-implementador/medios-de-pago-disponibles

PR #3343, opened 2019-09-10
Update stripe_payment_intents.rb

waaux:
A bug on Stripe Payment Intents gateway file preventing from storing a customer credit card


curiousepic:
@waaux Right, so you should be able to test a store with another payment method type such as a check. See https://github.com/curiousepic/active_merchant/blame/master/test/remote/gateways/remote_stripe_test.rb#L505

waaux:
Can someone fix this spelling mistake on the Stripe Payment Intents gateway line #137
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe_payment_intents.rb#L137 

Don't think a remote test is needed in this case because this line is calling the legacy Stripe store method already tested.

curiousepic:
@waaux Ah, you're right about the test 👍  I'll clean up history, add changelog entry and merge it in soon.

PR #3344, opened 2019-09-11
Worldpay: Add 3DS2 Support

nfarve:
Adds 3DS support to worldpay gateway. Two remote tests failing that are
unlreated.

.................................................
51 tests, 221 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.0784% passed

Loaded suite test/unit/gateways/worldpay_test
Started
...................................................................

67 tests, 397 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -339,7 +339,8 @@ module ActiveMerchant #:nodoc:
           xml.tag! 'date', 'month' => format(payment_method.month, :two_digits), 'year' => format(payment_method.year, :four_digits)
         end
 
-        xml.tag! 'cardHolderName', options[:execute_threed] && (options[:three_ds_version] =~ /[^2]/).nil? ? '3D' : payment_method.name
+        three_ds_version = options[:three_ds_version] || options[:three_d_secure][:version]
+        xml.tag! 'cardHolderName', options[:execute_threed] && (three_ds_version =~ /[^2]/).nil? ? '3D' : payment_method.name
         xml.tag! 'cvc', payment_method.verification_value
 
         add_address(xml, (options[:billing_address] || options[:address]))


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -208,6 +208,8 @@ def build_authorization_request(money, payment_method, options)
                 add_instalments_data(xml, options)
               end
               add_moto_flag(xml, options) if options.dig(:metadata, :manual_entry)
+              add_additional_3ds_data(xml, options) if options[:execute_threed] && options[:three_ds_version] && options[:three_ds_version].include?('2')
```
curiousepic: This is probably ridiculous but theoretically `three_ds_version` could be `1.2`? So ideally this might be something like `options[:three_ds_version][0] == '2'`

curiousepic: Also shouldn't this be `options[:three_d_secure][:version]`?

curiousepic:
Does their sandbox not handle 3ds2 properly or can we add remote tests for that flow?

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -330,6 +340,7 @@ def add_card(xml, payment_method, options)
         end
 
         xml.tag! 'cardHolderName', options[:execute_threed] ? '3D' : payment_method.name
+        # xml.tag! 'cardHolderName', payment_method.name #When testing 3DS 2.0 uncomment and comment above
```
britth: should we remove this comment?

nfarve: No we need to do this to test our remote tests from core unfortunately.

lib/active_merchant/billing/gateways/worldpay.rb
nfarve: So I believe we only send `options[:three_d_secure][:version]` when we pass the authenticated fields. For regular 3DS fields we send it in the options directly. Will update though so it checks it if starts with 2. Good catch!

curiousepic: Oh I see, I'm conflating this with pass-through. Not seeing any remote tests, are we doing that external to AM?

nfarve:
@curiousepic add a remote test!

lib/active_merchant/billing/gateways/worldpay.rb
nfarve: Actually thought of a way to get rid of this.

PR #3353, opened 2019-09-16
Mundipagg: Append errors to message response field

jasonxp:
For ```purchase```, ```authorize```, and ```store```, extract any error messages from the gateway response and append them to the "message" response feld. Mundipagg can return errors at either the top level of the response object, or under the ```last_transaction``` ```gateway_response``` key.

### DIFF
--- a/test/unit/gateways/mundipagg_test.rb
+++ b/test/unit/gateways/mundipagg_test.rb
@@ -38,6 +38,8 @@ class MundipaggTest < Test::Unit::TestCase
       billing_address: address,
       description: 'Store Purchase'
     }
+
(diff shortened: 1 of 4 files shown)


test/unit/gateways/mundipagg_test.rb
```diff
@@ -38,6 +38,8 @@ def setup
       billing_address: address,
       description: 'Store Purchase'
     }
+
+    @gateway_response_error = 'Esta loja n??o possui um meio de pagamento configurado para a bandeira VR'
```
jasonxp: I haven't been able to reproduce one of these errors, as they likely require a configuration discrepancy. There is thus no corresponding remote test. However, the code and test were written based on examination of a real such error encountered by a user.

PR #3354, opened 2019-09-16
Redsys: Add exemptions

britth:
Sets sca_exemption when available for 3ds transactions. Also sets
moto flag when requested.

Remote:
22 tests, 68 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
36 tests, 111 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys.rb
+++ b/lib/active_merchant/billing/gateways/redsys.rb
@@ -200,6 +200,7 @@ module ActiveMerchant #:nodoc:
         add_threeds(data, options) if options[:execute_threed]
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
+        data[:sca_exemption] = options[:sca_exemption]
 
         commit data, options
       end
@@ -215,6 +216,7 @@ module ActiveMerchant #:nodoc:
         add_threeds(data, options) if options[:execute_threed]
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
+        data[:sca_exemption] = options[:sca_exemption]
 
         commit data, options
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -441,8 +441,15 @@ def build_merchant_data(xml, data, options = {})
             xml.DS_MERCHANT_DIRECTPAYMENT 'true'
           end
 
+          # Set moto flag only if explicitly requested via moto field
+          # Requires account configuration to be able to use
+          if options.dig(:moto) && options.dig(:metadata, :manual_entry)
```
britth: I added a `moto` field in addition to the standard metadata[:manual_entry] since requests will fail with the error `Merchant has not configured the payment method "MOTO/Manual payment" ` if you pass this field without your account being configured. Wondering if I should just remove the check for `options.dig(:metadata, :manual_entry)` in this case?

curiousepic: Where exactly is this `:metadata, :manual_entry` coming from?

britth: per [here](https://github.com/activemerchant/active_merchant/pull/3329#issuecomment-528576484), that's what I believe shopify has used for determining moto transactions, and was suggested as the standardized field for marking moto. I don't want to use just that field though, since if the account doesn't support moto but this field is passed in, it'll result in a possibly unintended error (but perhaps that means I should just use the separate field altogether?)

curiousepic: I'm a little confused; it doesn't seem like the `moto` field would help much in ensuring that's the case. Are you picturing it being an extra verification that the account is flagged to support moto? If so, I don't feel like that's necessary.

britth: My thought was, even though we're standardizing on the manual_entry flag, since it's already being used by Shopify I don't want to unintentionally request this exemption for someone if their account isn't configured. 

So basically trying to prevent a case where you're marking your payment method as manual_entry as you normally would but you're not actually intending to request an exemption, and the transaction fails because your account isn't set up - does that make sense?

curiousepic: Ah I see, I didn't quite understand its interaction with the exemption piece. Thanks for the clarification. I don't have a strong opinion about whether to keep the second condition, I think you're right that it should work as intended with only the first.

PR #3355, opened 2019-09-16
TNS: Update verison and support pay mode

curiousepic:
This is an account-level flag that allows single-step purchases.
Test cards were updated to use expiry date to invoke response codes
according to https://na-gateway.mastercard.com/api/documentation/integrationGuidelines/supportedFeatures/testAndGoLive.html?locale=en_US

Remote (1 unrelated failure due to regional test account):
14 tests, 48 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92.8571% passed

Unit:
17 tests, 78 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/mastercard.rb
+++ b/lib/active_merchant/billing/gateways/mastercard.rb
@@ -7,9 +7,20 @@ module ActiveMerchant
       end
 
       def purchase(amount, payment_method, options={})
-        MultiResponse.run do |r|
-          r.process { authorize(amount, payment_method, options) }
-          r.process { capture(amount, r.authorization, options) }
+        if options[:pay_mode]
+          post = new_post
+          add_invoice(post, amount, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/mastercard.rb
```diff
@@ -7,9 +7,20 @@ def initialize(options={})
       end
 
       def purchase(amount, payment_method, options={})
-        MultiResponse.run do |r|
-          r.process { authorize(amount, payment_method, options) }
-          r.process { capture(amount, r.authorization, options) }
+        if options[:pay_mode]
```
molbrown: Is this flag available for all the regional endpoints?

curiousepic: Yep

PR #3360, opened 2019-09-19
Credorax: Add A Mandatory 3DS field

nfarve:
Credorax now requires `3ds_transtype` is required. This passes a default
value if none is passed. Also passes optional field `3ds_version`.

Loaded suite test/unit/gateways/credorax_test
Started
.......................

Finished in 0.020239 seconds.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
23 tests, 125 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -278,7 +278,9 @@ module ActiveMerchant #:nodoc:
           post[:'3ds_channel'] = '02'
           post[:'3ds_redirect_url'] = three_ds_2_options[:notification_url]
           post[:'3ds_challengewindowsize'] = options[:three_ds_challenge_window_size] || '03'
+          post[:'3ds_version'] = options[:three_ds_version] if options[:three_ds_version]
           post[:d5] = browser_info[:user_agent]
+          post[:'3ds_transtype'] = options[:transaction_type] || '01'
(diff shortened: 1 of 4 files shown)


nfarve:
Credorax sandbox seems to be having trouble. About 10 tests are failing on master.

nfarve:
Remote tests working again!

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -278,7 +278,9 @@ def add_3d_secure(post, options)
           post[:'3ds_channel'] = '02'
           post[:'3ds_redirect_url'] = three_ds_2_options[:notification_url]
           post[:'3ds_challengewindowsize'] = options[:three_ds_challenge_window_size] || '03'
+          post[:'3ds_version'] = options[:three_ds_version] if options[:three_ds_version]
           post[:d5] = browser_info[:user_agent]
+          post[:'3ds_transtype'] = '01' || options[:transaction_type]
```
britth: should we set this as `options[:transaction_type] || '01'` so that if they actually pass in a value it'll use that over '01'

nfarve: yup 🤦‍♀

PR #3361, opened 2019-09-19
Adyen: Add Cabal card

leila-alderman:
Added the Cabal card to the Adyen gateway, including adding unit and
remote tests.

Currently, Adyen does not support recurring transactions for Cabal
cards, so additional logic was added to the `store` method to return an
error message when attempting to store a Cabal card. Since the `store`
method for Adyen uses the `authorise` action, even though Cabal cards
cannot be successfully stored, the response from the gateway is
successful and does not return any sort of error message.

CE-99

Unit:
40 tests, 193 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
68 tests, 214 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -10,7 +10,7 @@ module ActiveMerchant #:nodoc:
       self.supported_countries = ['AT', 'AU', 'BE', 'BG', 'BR', 'CH', 'CY', 'CZ', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GB', 'GI', 'GR', 'HK', 'HU', 'IE', 'IS', 'IT', 'LI', 'LT', 'LU', 'LV', 'MC', 'MT', 'MX', 'NL', 'NO', 'PL', 'PT', 'RO', 'SE', 'SG', 'SK', 'SI', 'US']
       self.default_currency = 'USD'
       self.currencies_without_fractions = %w(CVE DJF GNF IDR JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
-      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :jcb, :dankort, :maestro,  :discover, :elo, :naranja]
+      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :jcb, :dankort, :maestro,  :discover, :elo, :naranja, :cabal]
 
       self.money_format = :cents
 
@@ -97,7 +97,14 @@ module ActiveMerchant #:nodoc:
         add_stored_credentials(post, credit_card, options)
         add_recurring_contract(post, options)
         add_address(post, options)
-        commit('authorise', post, options)
+
+        initial_response = commit('authorise', post, options)
+
+        if initial_response.success? && card_not_stored?(initial_response)
+          unsupported_failure_response(initial_response)
+        else
+          initial_response
+        end
       end
 
       def verify(credit_card, options={})
@@ -502,6 +509,23 @@ module ActiveMerchant #:nodoc:
           timeZoneOffset: browser_info[:timezone],
           userAgent: browser_info[:user_agent]
         }
+      end
+
+      def unsupported_failure_response(initial_response)
+        Response.new(
+          false,
+          'Recurring transactions are not supported for this card type.',
+          initial_response.params,
+          authorization: initial_response.authorization,
+          test: initial_response.test,
+          error_code: initial_response.error_code,
+          avs_result: initial_response.avs_result,
+          cvv_result: initial_response.cvv_result[:code]
+        )
+      end
+
+      def card_not_stored?(response)
+        response.authorization ? response.authorization.split('#')[2].nil? : true
       end
     end
   end
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -35,6 +35,24 @@ class RemoteAdyenTest < Test::Unit::TestCase
 
     @three_ds_enrolled_card = credit_card('4917610000000000', month: 10, year: 2020, verification_value: '737', brand: :visa)
 
+    @cabal_credit_card = credit_card('6035 2277 1642 7021',
+      :month => 10,
+      :year => 2020,
+      :first_name => 'John',
+      :last_name => 'Smith',
+      :verification_value => '737',
+      :brand => 'cabal'
+    )
+
+    @invalid_cabal_credit_card = credit_card('6035 2200 0000 0006',
+      :month => 10,
+      :year => 2020,
+      :first_name => 'John',
+      :last_name => 'Smith',
+      :verification_value => '737',
+      :brand => 'cabal'
+    )
+
     @declined_card = credit_card('4000300011112220')
 
     @improperly_branded_maestro = credit_card(
@@ -376,10 +394,22 @@ class RemoteAdyenTest < Test::Unit::TestCase
     assert_equal '[capture-received]', response.message
   end
 
+  def test_successful_purchase_with_cabal_card
+    response = @gateway.purchase(@amount, @cabal_credit_card, @options.merge(currency: 'ARS'))
+    assert_success response
+    assert_equal '[capture-received]', response.message
+  end
+
   def test_failed_purchase
     response = @gateway.purchase(@amount, @declined_card, @options)
     assert_failure response
     assert_equal 'CVC Declined', response.message
+  end
+
+  def test_failed_purchase_with_invalid_cabal_card
+    response = @gateway.purchase(@amount, @invalid_cabal_credit_card, @options)
+    assert_failure response
+    assert_equal 'Invalid card number', response.message
   end
 
   def test_successful_authorize_and_capture
@@ -393,6 +423,15 @@ class RemoteAdyenTest < Test::Unit::TestCase
 
   def test_successful_authorize_and_capture_with_elo_card
     auth = @gateway.authorize(@amount, @elo_credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount, auth.authorization)
+    assert_success capture
+    assert_equal '[capture-received]', capture.message
+  end
+
+  def test_successful_authorize_and_capture_with_cabal_card
+    auth = @gateway.authorize(@amount, @cabal_credit_card, @options)
     assert_success auth
 
     assert capture = @gateway.capture(@amount, auth.authorization)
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -96,7 +97,28 @@ def store(credit_card, options={})
         add_stored_credentials(post, credit_card, options)
         add_recurring_contract(post, options)
         add_address(post, options)
-        commit('authorise', post, options)
+
+        initial_response = commit('authorise', post, options)
+	if initial_response.success? == false
+	  initial_response
+        elsif initial_response.authorization.split('#')[2].nil? 
+	  # create failure response
```
therufs: How do you feel about pulling this out into a helper method? I think it'd be a bit more graceful/easier to read.

leila-alderman: Done! Thanks for the suggestion!

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -96,7 +96,16 @@ def store(credit_card, options={})
         add_stored_credentials(post, credit_card, options)
         add_recurring_contract(post, options)
         add_address(post, options)
-        commit('authorise', post, options)
+
+        initial_response = commit('authorise', post, options)
+
+        if initial_response.success? == false
```
jasonxp: This conditional looks redundant because ```initial_response``` is returned in by the first ```if``` clause and the ```else```. Is this intentional? If it's possible for the response to be successful without a ```pspReference```, it looks like we also need to make the ```split``` null-safe.

I'm looking at this case, where ```pspReference``` is nil: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/adyen.rb#L473

So if that null case is a possibility, I think this block could be refactored to something like:

```
if initial_response.success? && initial_response.authorization && initial_response.authorization.split('#')[2].nil? 
  failure_response(post)
else
  initial_response
end
```

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -459,6 +468,22 @@ def add_browser_info(browser_info, post)
           userAgent: browser_info[:user_agent]
         }
       end
+
+      def failure_response(post)
+        action = 'authorise'
+        raw_response = ssl_post("#{url}/#{action}", post_data(action, post), request_headers(options))
```
jasonxp: Do we have to make a second request to the gateway in order to return this failure response, after already determining that there was a failure after the initial request?

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -96,7 +96,16 @@ def store(credit_card, options={})
         add_stored_credentials(post, credit_card, options)
         add_recurring_contract(post, options)
         add_address(post, options)
-        commit('authorise', post, options)
+
+        initial_response = commit('authorise', post, options)
+
+        if initial_response.success? == false
+          initial_response
+        elsif initial_response.authorization.split('#')[2].nil?
```
jasonxp: What does this check? Is ```initial_response.authorization.split('#')[2].nil?``` true when this is a recurring transaction? (Looking at ```authorization_from``` linked above, it looked like it would be the opposite -- non-nil when the transaction is recurring -- but I could be interpreting that incorrectly).

jasonxp:
```
def card_not_successfully_stored?(response)
  response.authorization.split('#')[2].nil?
end
```

lib/active_merchant/billing/gateways/adyen.rb
leila-alderman: I've refactored this conditional statement to have only `if` and `else` statements to remove the redundancy.

lib/active_merchant/billing/gateways/adyen.rb
leila-alderman: It turns out we do not! Based on our chat yesterday, this has now been refactored to just replace the `success` and `message` parameters of the initial response without sending a new request.

lib/active_merchant/billing/gateways/adyen.rb
leila-alderman: This has now been refactored to create a new helper method as you suggested above.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -96,7 +96,14 @@ def store(credit_card, options={})
         add_stored_credentials(post, credit_card, options)
         add_recurring_contract(post, options)
         add_address(post, options)
-        commit('authorise', post, options)
+
+        initial_response = commit('authorise', post, options)
+
+        if initial_response.success? && card_not_stored?(initial_response)
```
jasonxp: 👍Nice refactor. This is a lot easier to read/understand.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -459,6 +466,23 @@ def add_browser_info(browser_info, post)
...
+          authorization: initial_response.authorization,
+          test: initial_response.test,
+          error_code: initial_response.error_code,
+          avs_result: initial_response.avs_result,
+          cvv_result: initial_response.cvv_result[:code]
+        )
+      end
+
+      def card_not_stored?(response)
+        response.authorization.split('#')[2].nil?
```
jasonxp: Is this null-safe? In other words, is it always guaranteed that ```response.authorization``` will be a string when this line executes? I think this question translates to, will the ```pspReference``` in the initial response always be present? (Traces down to here: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/adyen.rb#L473). If not or if it depends on what the remote server does, it would be best to add a null check to be safe.

So something like ```response.authorization && response.authorization.split('#')[2].nil?``` if that conditional retains the implication of the card not being stored.

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -33,6 +33,24 @@ def setup
       :brand => 'elo'
     )
```
jasonxp: Just a heads up in case you didn't notice -- there appear to be conflicts in this file now.

leila-alderman: Yep, I'm aware, but I generally hold off on rebasing until I'm all set to go.

lib/active_merchant/billing/gateways/adyen.rb
leila-alderman: From what I've seen `pspReference` always exists; that seems to be the transaction ID. It's just the `recurring.recurringDetailReference` that doesn't always get returned. So from what I know and what I've seen, this should be fine. Is there some other check that I can/should do to confirm that though?

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -459,6 +466,23 @@ def add_browser_info(browser_info, post)
...
+      def failure_response(initial_response)
+        Response.new(
+          false,
+          'Recurring transactions are not supported for this card type.',
+          initial_response.params,
+          authorization: initial_response.authorization,
+          test: initial_response.test,
+          error_code: initial_response.error_code,
+          avs_result: initial_response.avs_result,
+          cvv_result: initial_response.cvv_result[:code]
```
jasonxp: 👍

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -459,6 +466,23 @@ def add_browser_info(browser_info, post)
           userAgent: browser_info[:user_agent]
         }
       end
+
+      def failure_response(initial_response)
```
jasonxp: Since the failure message is hardcoded for the case of recurring transactions not being supported, do you think this method name should be more specific? (Or alternatively, the message could be a parameter).

lib/active_merchant/billing/gateways/adyen.rb
jasonxp: I would just check for null (like in ```response.authorization && ...```). That would make the code more robust, preventing an error in the case that the gateway does something different, someone uses the method for some other case where the authorization could be ```nil```, or that it's possible here for some reason we're unaware of. Looking at ```authorization_from``` again, I realized that the early return of ```nil`` ` in the case of ```pspReference``` not being present is itself a null check. So, we'd actually just be cascading that null check to this method.

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -475,6 +532,13 @@ def test_successful_store_with_elo_card
     assert_equal 'Authorised', response.message
   end
 
+  # Adyen does not currently support recurring transactions with Cabal cards
+  def test_failed_store_with_cabal_card
+    assert response = @gateway.store(@cabal_credit_card, @options)
+    assert_failure response
+    assert_equal 'Recurring transactions are not supported for this card type.', response.message
```
jasonxp: 👍

jasonxp:
Added a couple of minor recommendations, but looks good!

lib/active_merchant/billing/gateways/adyen.rb
leila-alderman: I went with `unsupported_failure_response` so that it's a bit more descriptive.

lib/active_merchant/billing/gateways/adyen.rb
leila-alderman: I went with a ternary to check to see if `response.authorization` exists that returns `true`, i.e., the card was not successfully stored, if it doesn't exist.

`response.authorization ? response.authorization.split('#')[2].nil? : true`

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -459,6 +466,23 @@ def add_browser_info(browser_info, post)
...
+          authorization: initial_response.authorization,
+          test: initial_response.test,
+          error_code: initial_response.error_code,
+          avs_result: initial_response.avs_result,
+          cvv_result: initial_response.cvv_result[:code]
+        )
+      end
+
+      def card_not_stored?(response)
+        response.authorization ? response.authorization.split('#')[2].nil? : true
```
jasonxp: 👍

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -459,6 +466,23 @@ def add_browser_info(browser_info, post)
           userAgent: browser_info[:user_agent]
         }
       end
+
+      def unsupported_failure_response(initial_response)
```
jasonxp: 👍

PR #3368, opened 2019-09-24
Quickpay: Remove testmode flag from v4-v7 platform

wsmoak:
According to Quickpay support, the testmode flag is no longer supported on the old platform.

See: https://github.com/activemerchant/active_merchant/pull/3367


wsmoak:
This was not meant to be opened against upstream activemerchant/active_merchant but against our fork.  sorry for the noise.

PR #3369, opened 2019-09-24
[Orbital] Add support for 3DS 2.0

cgriego:
The `version` and `ds_trasactioni_id` fields are [standardized](https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields). The `ucaf_collection_ind` field is supported as a top-level option by the `MerchantESolutionsGateway` class, which hasn't been updated to support the `:three_d_secure` options hash.

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -30,7 +30,7 @@ module ActiveMerchant #:nodoc:
     class OrbitalGateway < Gateway
       include Empty
 
-      API_VERSION = '7.7'
+      API_VERSION = '7.9'
 
       POST_HEADERS = {
         'MIME-Version' => '1.1',
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -521,6 +521,24 @@ module ActiveMerchant #:nodoc:
         xml.tag!(:PymtBrandProgramCode, 'ASK')
       end
 
+      def add_mc_program_protocol(xml, creditcard, three_d_secure)
+        return unless three_d_secure && creditcard.brand == 'master'
+
+        xml.tag!(:MCProgramProtocol, three_d_secure[:version].to_i) if three_d_secure[:version]
+      end
+
+      def add_mc_directory_trans_id(xml, creditcard, three_d_secure)
+        return unless three_d_secure && creditcard.brand == 'master'
+
+        xml.tag!(:MCDirectoryTransID, three_d_secure[:ds_transaction_id]) if three_d_secure[:ds_transaction_id]
+      end
+
+      def add_ucaf_ind(xml, creditcard, three_d_secure)
+        return unless three_d_secure && creditcard.brand == 'master'
+
+        xml.tag!(:UCAFInd, three_d_secure[:ucaf_collection_ind]) if three_d_secure[:ucaf_collection_ind]
+      end
+
       def add_refund(xml, currency=nil)
         xml.tag! :AccountNum, nil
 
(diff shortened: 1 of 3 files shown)


cgriego:
@pi3r Care to take a look since you did 3DS 1.0?

pi3r:
@cgriego Isn't the `ucaf_collection_ind` the `cavv`? [standardized](https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields) reads `Authentication Value (CAVV for Visa, AAV/UCAF for Mastercard)`. Would that make sense to use the `cavv` here? @filipebarcos any insight?

Otherwise the code looks good to me.

filipebarcos:
Agree with @pi3r here. We should leverage those standardized fields. Cavv was chosen as key name since it was more commonly used, when compared to the MC naming

cgriego:
The UCAF is put in the AAV element from the standardized cavv option already. The ucaf_collection_ind is a separate UCAF Collection Indicator from MasterCard which is a 1-7 single digit code.

pi3r:
@filipebarcos Do we have this information from our 3DS provider?

cgriego:
If it helps, we are receiving it from our 3DS provider, Cardinal Commerce, as a `UCAFIndicator` field in [their payload](https://cardinaldocs.atlassian.net/wiki/spaces/CC/pages/98315/Response+Objects).

nebdil:
@cgriego  UCAF indicator is the ECI for MasterCard. The possible MasterCard `ECIFlag`s we receive from the 3DS provider are: `00`, `01` and `02`. And the possible `UCAFIndicator`s are `0`, `1` and `2`
So you can make use of `three_d_secure[:eci]`

cgriego:
That isn’t going to work, the values mean different things and we get both. This is from a production transaction. `"ECIFlag"=>"07", "UCAFIndicator"=>"0"`

nebdil:
According to Cardinal, if the `ECIFlag` is `07` then it's not MasterCard and if it's not MasterCard `UCAFIndicator` is not returned.

Regarding the `ECIFlag`:
```
Possible Values:
02 or 05 - Fully Authenticated Transaction
01 or 06 - Attempted Authentication Transaction 
00 or 07 - Non 3-D Secure Transaction
Mastercard - 02, 01, 00
VISA - 05, 06, 07
AMEX - 05, 06, 07
JCB - 05, 06, 07
DINERS CLUB - 05, 06, 07
```
and regarding `UCAFIndicator`:
```
Possible Values:
0 - Non-SecureCode transaction, bypassed by the Merchant
1 - Merchant-Only SecureCode transaction
2 - Fully authenticated SecureCode transaction
NOTE: This field is only returned for Mastercard transactions.
```

[resource](https://cardinaldocs.atlassian.net/wiki/spaces/CCen/pages/905478154/Authenticate+Request+Response)

cgriego:
I double-checked that the transaction I referenced was a MasterCard transaction. It also had `Enrolled"=>"B", ThreeDSVersion"=>"1.0.2"`.
> B - Bypass, Merchant authentication rule is triggered to bypass authentication in this use case

The Orbital docs have this to say about ECI and UCAFInd

```
AuthenticationECIInd
  2 - recurring transaction with a consumer digital payment token
  5 - 3DS authenticated transaction
  6 - 3DS attempted authentication
  7 - authentication not performed

UCAFInd
  1 - Attempted authentication
  2 - Full authentication
  3 - Static authentication
  4 - Not authenticated
  5 - Issuer Risk Based Decisioning
  6 - Merchant Risk Based Decisioning
  7 - Partial shipment or recurring payment
```

Might have to involve our Chase contact to sort this out.

nebdil:
Do let me know if I can look into anything else, but what I referenced above is **exactly** what we receive from the 3DS provider and can add to the `three_d_secure` hash at the moment

cgriego:
@nebdil Would it be possible to revive this discussion? We have been successfully running with our fork with this code for the past 2 years. Merchant e-Solutions gateway supports `ucaf_collection_ind` since 2016. https://github.com/activemerchant/active_merchant/commit/a6b2c4bfb29987e6a9dba2babc8ef4ae4ff2ca82

If so then I can work on fixing the conflicts.

nebdil:
We can revive the discussion yes :) if you run this code successfully and there is never `ucaf_collection_ind` since it's not in the standardized `three_d_secure` fields, would that not mean that it's not necessary?

cgriego:
@nebdil Any input on the above questions and how to handle the functionality or backwards compatibility?

curiousepic:

PR #3374, opened 2019-09-26
Adyen: Fix message for authorise3d

jeremywrowe:
When making an `authorise3d` request on the Adyen gateway, there was not a proper
mapping for refusal response messages. This looks to be an oversight in the
original addition of `authorise3d` as the support was added to `success_from`
but not to `message_from`.

2 remote failures are unrelated to this change and are known to fail in master.

Unit:

49 tests, 237 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:

72 tests, 232 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.2222% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
 == HEAD
 * Revert "Revert "Worldpay: Switch to Nokogiri"" [curiousepic] #3373
+* Adyen: Fix `authorise3d` message for refusals [jeremywrowe] #3374
 
 == Version 1.99.0 (Sep 26, 2019)
 * Adyen: Add functionality to set 3DS exemptions via API [britth] #3331
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -464,7 +464,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def message_from(action, response)
-        return authorize_message_from(response) if action.to_s == 'authorise'
+        return authorize_message_from(response) if action.to_s == 'authorise' || action.to_s == 'authorise3d'
         response['response'] || response['message']
       end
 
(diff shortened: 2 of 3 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -246,62 +246,6 @@ def test_successful_authorize_with_3ds_dynamic_rule_broken
...
-
-    auth = @gateway.authorize(@amount, @credit_card, options)
-    assert_success auth
-    assert_equal 'Authorised', auth.message
-    assert_equal 'true', auth.params['additionalData']['liabilityShift']
-
-    response = @gateway.purchase(@amount, @credit_card, options)
-    assert_success response
-  end
-
```
molbrown: Should these tests instead be improved (not as part of this commit)? It doesn't seem there are any other remote tests that support the 3DS auth data fields.

jeremywrowe: 🤔 Probably worth adding back, I am happy to revert the deletion, but I hate to see failing tests in master.

molbrown: Or instead commented-out with the note that they require certain account capabilities?

jeremywrowe: I'd opt for removal over comment out since the only way to run them would be to uncomment them :)

molbrown: Commenting out remote tests that only work _sometimes_ is a pattern I occasionally see in Active Merchant. Example: https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_worldpay_test.rb

Would this be a matter of getting Spreedly's test account the right capabilities in order for it to pass?

molbrown: We support Adyen 3DS auth data fields, so we should have (working) tests for them.

jeremywrowe: I'll let them fail for now, and circle back in another PR and fix them.

PR #3380, opened 2019-10-03
Worldpay: Handle parse errors gracefully

curiousepic:
We have experienced instances of Worldpay responses consisting of a
plain text string, which threw a NoMethodError when attempting to parse.
It will now return a falilure response with the relevant parsing error.
This also replaces the problematic & character in responses.

Remote (6 unrelated failures):
54 tests, 231 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
88.8889% passed

Unit:
68 tests, 401 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -457,7 +457,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def parse(action, xml)
-        doc = Nokogiri::XML(xml)
+        xml = xml.strip.gsub(/\&/, '&amp;')
+        doc = Nokogiri::XML(xml, &:strict)
         doc.remove_namespaces!
         resp_params = {:action => action}
 
@@ -512,6 +513,8 @@ module ActiveMerchant #:nodoc:
           :avs_result => AVSResult.new(code: AVS_CODE_MAP[raw[:avs_result_code_description]]),
           :cvv_result => CVVResult.new(CVC_CODE_MAP[raw[:cvc_result_code_description]])
         )
+      rescue Nokogiri::SyntaxError => e
+        unparsable_response(xml)
       rescue ActiveMerchant::ResponseError => e
         if e.response.code.to_s == '401'
           return Response.new(false, 'Invalid credentials', {}, :test => test?)
@@ -522,6 +525,12 @@ module ActiveMerchant #:nodoc:
 
       def url
         test? ? self.test_url : self.live_url
+      end
+
+      def unparsable_response(raw_response)
+        message = 'Unparsable response received from Worldpay. Please contact Worldpay if you continue to receive this message.'
+        message += " (The raw response returned by the API was: #{raw_response.inspect})"
+        return Response.new(false, message)
       end
 
       # Override the regular handle response so we can access the headers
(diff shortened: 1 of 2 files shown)


curiousepic:
I'm open to other methods here. Ideally this would result in a `gateway_processing_result_unknown` response on the Spreedly end, since we don't technically know the result if the response is unparesable. I'm not yet sure if this method is compatible with that.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -457,7 +457,8 @@ def default_address
       end
 
       def parse(action, xml)
-        doc = Nokogiri::XML(xml)
+        xml = xml.strip.gsub(/\&/, '&amp;')
+        doc = Nokogiri::XML(xml, &:strict)
```
molbrown: 👍

molbrown:
Is there a way to test for the `gateway_processing_result_unknown` result locally? Can you force a string response to see how this error is interpreted by Spreedly?

curiousepic:
I've been rethinking this approach. I think it might be more appropriate to raise a proper AM-level error instead of returning a response, and Spreedly will treat it as desired. Here is a separate PR for consideration: https://github.com/activemerchant/active_merchant/pull/3381

curiousepic:
Since the other approach seemed suboptimal for AM, I'm reexamining this one. My primary concern  now is one of identifying the response as not knowing state of the transaction at the gateway. We have a concept of that at Spreedly, given certain circumstances, but the response here as-is will not trigger those circumstances. Furthermore, I'm not sure if AM has an existing approach for this situation. I'll continue to dig along those lines.

curiousepic:
I think my concern is stemming from the fact that AM just doesn't inherently distinguish between transaction success/failure/unknown and response success/failure. But it's a known shortcoming, so we'll handle it on our end for now.

I did find a common pattern for similar situations with JSON responses, so I'll this to conform to that pattern.

jeremywrowe:
Sounds good, thanks for the care and thought put into this change.


Not for now, but perhaps long term:

> I think my concern is stemming from the fact that AM just doesn't inherently distinguish between transaction success/failure/unknown

Right now we're returning a `Response` that is marked as a failure:

```rb
return Response.new(false, "Error parsing response: #{e}", {}, :test => test?)
```

I wonder if we could add another attribute to Response `unknown` long term. We'd want to ensure that we mark the response as error and unknown if we don't know the result of the response. Something like:

```rb
class Response
 # ... existing implimentation omitted for brevity 

  # add a default attribute for uknown responses
  def unknown?
    false
  end
end
```

And then introduce an UnknownResponse class
```rb
class UnknownResponse < Response
  def success?
    false
  end

  def error?
    true
  end

  def unknown?
    true
  end
end
```

and then replace the current unknown result with:

```rb
return UnknownResponse.new("Error parsing response: #{e}", {}, :test => test?)
```

What I like about it is that it removes the chance for error since we'd be adding a third state. We could also do the same with success and error by exposing `SuccessResponse` and `ErrorResponse` classes as well.

curiousepic:
@jeremywrowe I like that potential approach. We should consider it when we tackle higher level guards against unparsable responses.

curiousepic:
Updated.

PR #3381, opened 2019-10-04
Worldpay: Raise appropriate error when parsing

curiousepic:
We have experienced instances of Worldpay responses consisting of a
plain text string, which threw a NoMethodError when attempting to parse.
It will now throw an InvalidResponseError with details.
This also replaces the problematic & character in responses.

Remote (3 unrelated failures):
54 tests, 235 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.4444% passed

Unit:
68 tests, 399 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -457,7 +457,12 @@ module ActiveMerchant #:nodoc:
       end
 
       def parse(action, xml)
-        doc = Nokogiri::XML(xml)
+        xml = xml.strip.gsub(/\&/, '&amp;')
+        begin
+          doc = Nokogiri::XML(xml, &:strict)
+        rescue Nokogiri::XML::SyntaxError => e
+          raise InvalidResponseError, "There was an error parsing the response: #{e} | The original response body was: #{xml}"
+        end
         doc.remove_namespaces!
         resp_params = {:action => action}
 
--- a/test/unit/gateways/worldpay_test.rb
+++ b/test/unit/gateways/worldpay_test.rb
@@ -836,6 +836,14 @@ class WorldpayTest < Test::Unit::TestCase
     assert_equal '3d4187536044bd39ad6a289c4339c41c', response.authorization
   end
 
+  def test_parsing_error
+    assert_raise ActiveMerchant::InvalidResponseError do
+      stub_comms do
+        @gateway.authorize(@amount, @credit_card, @options)
+      end.respond_with('string')
+    end
+  end
+
   private
 
   def assert_tag_with_attributes(tag, attributes, string)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -457,7 +457,12 @@ def default_address
       end
 
       def parse(action, xml)
-        doc = Nokogiri::XML(xml)
+        xml = xml.strip.gsub(/\&/, '&amp;')
+        begin
+          doc = Nokogiri::XML(xml, &:strict)
+        rescue Nokogiri::XML::SyntaxError => e
+          raise InvalidResponseError, "There was an error parsing the response: #{e} | The original response body was: #{xml}"
```
curiousepic: Honestly this feels a little weird; is this pattern of `raise`ing within a `rescue` bad? Acceptable? Good? Is there a more idiomatic way to convert one type of error into another?

curiousepic: Also, `InvalidResponseError` is only used one other place in AM, for catching a certain type of error thrown for any gateway connection: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/network_connection_retries.rb#L31-L32

jeremywrowe: Exceptions as control flow is typically a "no no", but often folks take this approach. I guess the thing to weigh here is:

1) is it hard to not use an exception as control flow
* if so, why?
* if not, another pattern should probably be used.
2) are exceptions as control flow already present in the system, if so it might be worth continuing the pattern.

curiousepic: @jeremywrowe Hmm, ok. The answer to 1. is not particularly obvious to me (nor is 2. really). 

My initial thinking was, we have a library (Nokogiri) in this library (AM) that throws an exception in certain  apparently-not-rare circumstances. It seems better to (A) wrap it in an abstracted AM-level exception rather than (B) having the original bubble all the way to the top. Would B be the alternative here, or is there an option I'm not seeing (very likely)? Or is B actually favorable?

jeremywrowe: If we need to raise an exception it's probably best to wrap it and include the original context. This changes depending on if the code being described is a library or not. Since this is library code, it's probably best to have a single exception to catch rather than catching everything.

The favorable thing here is to not throw an exception at all, but instead, mark the response as an error and provide a reason why. I think this boils down to making a judgment call though - Since I haven't taken a look at all of what it might take to do so.

The paths that (in my opinion) are ranked below:

1) Respond with a failure response with no exception
2) raise a library-specific exception with the original exception/context included
3) raise the original exception

curiousepic: Thanks, that helps. I moved to this approach primarily because Spreedly would already be treating the outcome  appropriately, but we should probably accommodate the change there rather than take an approach which is suboptimal for AM. There are some problems remaining with the original approach however. I'm going to move discussion back there https://github.com/activemerchant/active_merchant/pull/3380

PR #3384, opened 2019-10-07
Moneris: Add support for stored credentials

chinhle23:
https://usa.visa.com/dam/VCOM/global/support-legal/documents/stored-credential-transaction-framework-vbs-10-may-17.pdf
https://developer.moneris.com/Documentation/NA/E-Commerce%20Solutions/API/Purchase

This change allows for the standardization of utilizing the stored credentials framework

1. VISA's stored credential transaction framework does not support 'recurring' nor 'installment' transactions
   for cardholder-initiated, thus passing the ':cardholder' option will default ':payment_indicator' to 'C' for
   initial transactions and to 'Z' for subsequent transactions.

2. The ':issuer_id', ':payment_indicator', and ':payment_information' options can still be passed directly and
   will override any values set by the ':stored_credential' options.

Unit:
50 tests, 266 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
38 tests, 199 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-422/ECS-780

### DIFF
--- a/lib/active_merchant/billing/gateways/moneris.rb
+++ b/lib/active_merchant/billing/gateways/moneris.rb
@@ -50,7 +50,7 @@ module ActiveMerchant #:nodoc:
         post[:order_id] = options[:order_id]
         post[:address] = options[:billing_address] || options[:address]
         post[:crypt_type] = options[:crypt_type] || @options[:crypt_type]
-        add_cof(post, options)
+        add_stored_credential(post, options)
         action = if post[:cavv]
                    'cavv_preauth'
                  elsif post[:data_key].blank?
@@ -73,7 +73,7 @@ module ActiveMerchant #:nodoc:
         post[:order_id] = options[:order_id]
         post[:address] = options[:billing_address] || options[:address]
         post[:crypt_type] = options[:crypt_type] || @options[:crypt_type]
-        add_cof(post, options)
+        add_stored_credential(post, options)
         action = if post[:cavv]
                    'cavv_purchase'
                  elsif post[:data_key].blank?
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -208,6 +208,52 @@ def add_cof(post, options)
...
+            post[:payment_indicator] ||= 'C'
+          end
+        else
+          post[:payment_indicator] ||= 'C'
+        end
+      end
+
+      def add_stored_credential_used(post, options)
+        post[:payment_information] ||= '2'
+        post[:issuer_id] = options[:stored_credential][:network_transaction_id] if post[:issuer_id] == ''
```
molbrown: Should this be `if options[:issuer_id] == ''`? If so, what about when `:issuer_id` is nil?

lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -208,6 +208,52 @@ def add_cof(post, options)
         post[:payment_information] = options[:payment_information] if options[:payment_information]
       end
 
+      def add_stored_credential(post, options)
+        add_cof(post, options)
+        # if any of :issuer_id, :payment_information, or :payment_indicator is not passed,
+        # then check for :stored credential options
+        return unless (stored_credential = options[:stored_credential]) && cof_details_not_present?(options)
```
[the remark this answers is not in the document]
chinhle23: Yes, I don't know why I didn't think using the existing method was going to work. Thanks!

lib/active_merchant/billing/gateways/moneris.rb
chinhle23: Yes good catch, I should have used `options` instead of `post`. Thank you! `auth_purchase_params` in core will convert all nil `:issuer_id` to `''`. Hmm now I'm wondering if there's something else I'm failing to consider?

chinhle23: Got it. Thanks for the tip. The changes have been pushed.

PR #3392, opened 2019-10-14
Global Collect: Check for more signs of failure

curiousepic:
Previously, the success method was only checking one potential response
status value (REJECTED). Transactions can fail at the gateway and be
returned with other statuses, such as CANCELLED. There is also a
status_category field which can contain success information which we now
also check.

Remote:
17 tests, 40 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
21 tests, 93 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -283,7 +283,7 @@ POST
       end
 
       def success_from(response)
-        !response['errorId'] && response['status'] != 'REJECTED'
+        !response['errorId'] && !['REJECTED', 'CANCELLED'].include?(response['status']) && response.dig('status_category') != 'UNSUCCESSFUL'
       end
 
       def message_from(succeeded, response)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -283,7 +283,7 @@ def content_type
       end
 
       def success_from(response)
-        !response['errorId'] && response['status'] != 'REJECTED'
+        !response['errorId'] && !['REJECTED', 'CANCELLED'].include?(response['status']) && response.dig('status_category') != 'UNSUCCESSFUL'
```
curiousepic: Just realized I don't actually have a test that excercises getting to the final check, I'll add one tomorrow.

jeremywrowe: Would it be better to invert this and check all known success states? https://epayments-api.developer-ingenico.com/s2sapi/v1/en_US/java/statuses.html?paymentPlatform=ALL

curiousepic: Yeah I think we'll have to do something like that. Their status system is terribly unintuitive. I'm going to close this out for another approach.

PR #3394, opened 2019-10-15
Add 3DS2 Support to Netbanx

Jujhar:
Adds Netbanx threeDSecure options to _add payment_ and _add credit card,_ as per request from Paysafe.

### DIFF
--- a/lib/active_merchant/billing/gateways/netbanx.rb
+++ b/lib/active_merchant/billing/gateways/netbanx.rb
@@ -132,6 +132,11 @@ module ActiveMerchant #:nodoc:
         if options[:billing_address]
           post[:card][:billingAddress]  = map_address(options[:billing_address])
         end
+
+        if options[:authentication]
+          post[:authentication]  = map_3ds(options[:authentication]) if options[:authentication]
+        end
+
       end
 
       def add_invoice(post, money, options)
@@ -151,6 +156,7 @@ module ActiveMerchant #:nodoc:
 
         post[:currencyCode] = options[:currency] if options[:currency]
         post[:billingDetails]  = map_address(options[:billing_address]) if options[:billing_address]
+        post[:authentication]  = map_3ds(options[:authentication]) if options[:authentication]
       end
 
       def expdate(credit_card)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/netbanx.rb
```diff
@@ -132,6 +132,11 @@ def add_credit_card(post, credit_card, options = {})
         if options[:billing_address]
           post[:card][:billingAddress]  = map_address(options[:billing_address])
         end
+
+        if options[:authentication]
```
nebdil:
```suggestion
        if options[:three_d_secure]
```

You can refer to the 3DS fields from this doc:
https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields

Jujhar: Updated

lib/active_merchant/billing/gateways/netbanx.rb
```diff
@@ -132,6 +132,10 @@ def add_credit_card(post, credit_card, options = {})
         if options[:billing_address]
           post[:card][:billingAddress]  = map_address(options[:billing_address])
         end
+
+        if options[:three_d_secure]
+          post[:authentication]  = map_3ds(options[:three_d_secure]) if options[:three_d_secure]
```
bbraschi: redundant  `if options[:three_d_secure]`?

lib/active_merchant/billing/gateways/netbanx.rb
```diff
@@ -179,6 +184,18 @@ def map_address(address)
         mapped
       end
 
+      def map_3ds(three_d_secure_options)
+        mapped = {
+          :eci => three_d_secure_options[:eci],
```
bbraschi: The `xid` would be required in order for 3DS 1.x to be supported.

lib/active_merchant/billing/gateways/netbanx.rb
Jujhar: Fixed

lib/active_merchant/billing/gateways/netbanx.rb
Jujhar: Fixed now, remembered seeing this.

PR #3396, opened 2019-10-17
CE-136: Ixopay purchase

jasonxp:
This pull request implements the Ixopay purchase action. It also includes skeleton code for additional work.

What's working:
* ```ruby -I test test/remote/gateways/remote_ixopay_test.rb -n test_successful_purchase```
* ```ruby -I test test/remote/gateways/remote_ixopay_test.rb -n test_failed_purchase```
* ```ruby -I test test/unit/gateways/ixopay_test.rb -n test_successful_purchase```
* ```ruby -I test test/unit/gateways/ixopay_test.rb -n test_failed_purchase```

The diff displayed here is relative to the initial code provided by running ```script/generate gateway ixopay```.

I'm planning a live review session for this code, but am opening the PR now for visibility.

Baseline: https://github.com/activemerchant/active_merchant/compare/ixopay_integration#diff-51a89ecca18e8620ad6cb8e059787833

### DIFF
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -1,3 +1,5 @@
+require 'nokogiri'
+
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class IxopayGateway < Gateway
@@ -18,7 +20,12 @@ module ActiveMerchant #:nodoc:
       end
 
       def purchase(money, payment_method, options={})
-        commit('purchase', build_purchase_request(money, payment_method, options), options)
+        request = build_xml_request do |xml|
+          add_card_data(xml, payment_method)
+          add_debit(xml, money, options)
+        end
+
+        commit(request)
       end
 
       def authorize(money, payment_method, options={})
@@ -78,31 +85,23 @@ module ActiveMerchant #:nodoc:
         Base64.encode64(hmac).delete("\n")
       end
 
-      def parse(action, xml)
-        parse_element({:action => action}, REXML::Document.new(xml))
+      def parse(body)
+        xml = Nokogiri::XML(body)
+        response = Hash.from_xml(xml.to_s)['result']
+
+        response.deep_transform_keys(&:underscore).transform_keys(&:to_sym)
       end
 
-      # todo
-      # This generic method appears in a number of gateways that parse XML.
-      # In the future, we should investigate finding a library method to
-      # drop in, or factoring it out to a module or base class.
-      def parse_element(raw, node)
-        node_name = node.name.underscore
-
-        node.attributes.each do |k, v|
-          raw["#{node_name}_#{k.underscore}".to_sym] = v
+      def build_xml_request
+        builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
+          xml.transactionWithCard 'xmlns' => 'http://secure.ixopay.com/Schema/V2/TransactionWithCard' do
+            xml.username @options[:username]
+            xml.password Digest::SHA1.hexdigest(@options[:password])
+            yield(xml)
+          end
         end
 
-        if node.has_elements?
-          raw[node_name.to_sym] = true unless node.name.blank?
-          node.elements.each { |e| parse_element(raw, e) }
-        elsif node.children.count > 1
-          raw[node_name.to_sym] = node.children.join(' ').strip
-        else
-          raw[node_name.to_sym] = node.text unless node.text.nil?
-        end
-
-        raw
+        builder.to_xml
       end
 
       def build_purchase_request(money, payment_method, options)
@@ -193,16 +192,16 @@ module ActiveMerchant #:nodoc:
         SecureRandom.uuid
       end
 
-      def commit(action, request, options={})
+      def commit(request)
         url = (test? ? test_url : live_url)
 
         begin
           raw_response = ssl_post(url, request, headers(request))
         rescue StandardError => error
-          return response_from_request_error(action, error)
+          return response_from_request_error(error)
         end
 
-        response = parse(action, raw_response)
+        response = parse(raw_response)
 
         Response.new(
           success_from(response),
@@ -214,8 +213,8 @@ module ActiveMerchant #:nodoc:
         )
       end
 
-      def response_from_request_error(action, error)
-        response = parse(action, error.response.body)
+      def response_from_request_error(error)
+        response = parse(error.response.body)
 
         Response.new(
           success_from(response),
@@ -232,7 +231,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def message_from(response)
-        response[:message] || response[:return_type]
+        response.dig(:errors, 'error', 'message') || response[:return_type]
       end
 
       def authorization_from(response)
@@ -241,7 +240,7 @@ module ActiveMerchant #:nodoc:
 
       def error_code_from(response)
         unless success_from(response)
-          response[:code]
+          response.dig(:errors, 'error', 'code')
         end
       end
     end
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -6,7 +6,7 @@ module ActiveMerchant #:nodoc:
 
       self.supported_countries = %w(AO AQ AR AS AT AU AW AX AZ BA BB BD BE BF BG BH BI BJ BL BM BN BO BQ BQ BR BS BT BV BW BY BZ CA CC CD CF CG CH CI CK CL CM CN CO CR CU CV CW CX CY CZ DE DJ DK DM DO DZ EC EE EG EH ER ES ET FI FJ FK FM FO FR GA GB GD GE GF GG GH GI GL GM GN GP GQ GR GS GT GU GW GY HK HM HN HR HT HU ID IE IL IM IN IO IQ IR IS IT JE JM JO JP KE KG KH KI KM KN KP KR KW KY KZ LA LB LC LI LK LR LS LT LU LV LY MA MC MD ME MF MG MH MK ML MM MN MO MP MQ MR MS MT MU MV MW MX MY MZ NA NC NE NF NG NI NL NO NP NR NU NZ OM PA PE PF PG PH PK PL PM PN PR PS PT PW PY QA RE RO RS RU RW SA SB SC SD SE SG SH SI SJ SK SL SM SN SO SR SS ST SV SX SY SZ TC TD TF TG TH TJ TK TL TM TN TO TR TT TV TW TZ UA UG UM US UY UZ VA VC VE VG VI VN VU WF WS YE YT ZA ZM ZW)
       self.default_currency = 'EUR'
-      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb, :maestro]
 
       self.homepage_url = 'https://www.ixopay.com'
       self.display_name = 'Ixopay'
@@ -23,18 +23,22 @@ module ActiveMerchant #:nodoc:
 
       def authorize(money, payment_method, options={})
         # todo
+        # commit('authorize', build_authorize_request(money, payment_method, options), options)
       end
 
       def capture(money, authorization, options={})
         # todo
+        # commit('capture', build_capture_request(money, payment_method, options), options)
       end
 
       def refund(money, authorization, options={})
         # todo
+        # commit('refund', build_refund_request(money, payment_method, options), options)
       end
 
       def void(authorization, options={})
         # todo
+        # commit('void', build_void_request(money, payment_method, options), options)
       end
 
       def verify(credit_card, options={})
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -82,6 +82,7 @@ module ActiveMerchant #:nodoc:
         parse_element({:action => action}, REXML::Document.new(xml))
       end
 
+      # todo
       # This generic method appears in a number of gateways that parse XML.
       # In the future, we should investigate finding a library method to
       # drop in, or factoring it out to a module or base class.
@@ -235,7 +236,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def authorization_from(response)
-        response[:reference_id] ? response[:reference_id] + '|' + response[:purchase_id] : nil
+        response[:reference_id] ? "#{response[:reference_id]}|#{response[:purchase_id]}" : nil
       end
 
       def error_code_from(response)
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -104,18 +104,6 @@ module ActiveMerchant #:nodoc:
         builder.to_xml
       end
 
-      def build_purchase_request(money, payment_method, options)
-        xml = Builder::XmlMarkup.new(indent: 2)
-
-        xml.transactionWithCard 'xmlns' => 'http://secure.ixopay.com/Schema/V2/TransactionWithCard' do
-          xml.username @options[:username]
-          xml.password Digest::SHA1.hexdigest(@options[:password])
-
-          add_card_data(xml, payment_method)
-          add_debit(xml, money, options)
-        end
-      end
-
       def add_card_data(xml, payment_method)
         xml.cardData do
           xml.cardHolder      payment_method.name
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -1,19 +1,15 @@
-require 'pp'
-
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class IxopayGateway < Gateway
       self.test_url = 'https://secure.ixopay.com/transaction'
       self.live_url = 'https://secure.ixopay.com/transaction'
 
-      self.supported_countries = ['US']
+      self.supported_countries = %w(AO AQ AR AS AT AU AW AX AZ BA BB BD BE BF BG BH BI BJ BL BM BN BO BQ BQ BR BS BT BV BW BY BZ CA CC CD CF CG CH CI CK CL CM CN CO CR CU CV CW CX CY CZ DE DJ DK DM DO DZ EC EE EG EH ER ES ET FI FJ FK FM FO FR GA GB GD GE GF GG GH GI GL GM GN GP GQ GR GS GT GU GW GY HK HM HN HR HT HU ID IE IL IM IN IO IQ IR IS IT JE JM JO JP KE KG KH KI KM KN KP KR KW KY KZ LA LB LC LI LK LR LS LT LU LV LY MA MC MD ME MF MG MH MK ML MM MN MO MP MQ MR MS MT MU MV MW MX MY MZ NA NC NE NF NG NI NL NO NP NR NU NZ OM PA PE PF PG PH PK PL PM PN PR PS PT PW PY QA RE RO RS RU RW SA SB SC SD SE SG SH SI SJ SK SL SM SN SO SR SS ST SV SX SY SZ TC TD TF TG TH TJ TK TL TM TN TO TR TT TV TW TZ UA UG UM US UY UZ VA VC VE VG VI VN VU WF WS YE YT ZA ZM ZW)
       self.default_currency = 'EUR'
       self.supported_cardtypes = [:visa, :master, :american_express, :discover]
 
       self.homepage_url = 'https://www.ixopay.com'
       self.display_name = 'Ixopay'
-
-      STANDARD_ERROR_CODE_MAPPING = {}
 
       def initialize(options={})
         requires!(options, :username, :password, :secret)
@@ -22,31 +18,23 @@ module ActiveMerchant #:nodoc:
       end
 
       def purchase(money, payment_method, options={})
-        puts build_purchase_request(money, payment_method, options), options
-
         commit('purchase', build_purchase_request(money, payment_method, options), options)
       end
 
-      def authorize(money, payment, options={})
-        post = {}
-        add_invoice(post, money, options)
-        add_payment(post, payment)
-        add_address(post, payment, options)
-        add_customer_data(post, options)
-
-        commit('authonly', post)
+      def authorize(money, payment_method, options={})
+        # todo
       end
 
       def capture(money, authorization, options={})
-        commit('capture', post)
+        # todo
       end
 
       def refund(money, authorization, options={})
-        commit('refund', post)
+        # todo
       end
 
       def void(authorization, options={})
-        commit('void', post)
+        # todo
       end
 
       def verify(credit_card, options={})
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+          raw[node_name.to_sym] = node.text unless node.text.nil?
+        end
+
+        raw
+      end
+
+      def build_purchase_request(money, payment_method, options)
+        xml = Builder::XmlMarkup.new(indent: 2)
+
+        xml.instruct! :xml, encoding: 'utf-8'
```
leila-alderman: It appears that `xml.instruct!` includes UTF-8 encoding by default and doesn't need to be specified: https://apidock.com/rails/v1.0.0/Builder/XmlMarkup/instruct%21

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+
+        xml.tag! 'transactionWithCard', 'xmlns' => 'http://secure.ixopay.com/Schema/V2/TransactionWithCard' do
```
leila-alderman: Rather than creating a `do...end` block here, it seems to make more sense in the context to use `{...}`. (Similar to examples [in the API docs](https://apidock.com/rails/Builder/XmlMarkup))


You could also consider refactoring this to use the block structure:
```
xml.transactionWithCard { |xml| 
  xml.username(@options[:username]
  xml.password(Digest::SHA1.hexdigest(@options[:password])
  add_card_data(xml, payment_method)
  add_debit(xml, money, options)
}
```

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+      end
+
+      def add_card_data(xml, payment_method)
+        xml.tag! 'cardData' do
+          xml.tag! 'cardHolder', payment_method.name
+          xml.tag! 'pan', payment_method.number
+          xml.tag! 'cvv', payment_method.verification_value
+          xml.tag! 'expirationMonth', format(payment_method.month, :two_digits)
+          xml.tag! 'expirationYear', format(payment_method.year, :four_digits)
+        end
```
leila-alderman: Consider refactoring:

```
xml.cardData { |xml|
  xml.cardHolder(payment_method.name)
  xml.pan(payment_method.number)
  xml.cvv(payment_method.verfication_value)
  xml.expirationMonth(format(payment_method.month, :two_digits))
  xml.expirationYear(format(payment_method.year, :four_digits))
}

File: lib/active_merchant/billing/gateways/ixopay.rb [hunk 4]
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+        xml.tag! 'shippingCity',      address[:city]
+        xml.tag! 'shippingPostcode',  address[:zip]
+        xml.tag! 'shippingState',     address[:state]
+        xml.tag! 'shippingCountry',   address[:country]
+        xml.tag! 'shippingPhone',     address[:phone]
       end
 
-      def parse(body)
-        {}
+      def new_transaction_id
```
leila-alderman: It doesn't look like this is used more than once. Why was it pulled out into a new method?

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -5,36 +5,43 @@ def setup
...
   end
 
   def test_successful_purchase
     response = @gateway.purchase(@amount, @credit_card, @options)
+
     assert_success response
-    assert_equal 'REPLACE WITH SUCCESS MESSAGE', response.message
-  end
+    assert_equal 'FINISHED', response.message
+    assert_match(/[0-9a-zA-Z]+(|[0-9a-zA-Z]+)*/, response.authorization)
```
leila-alderman: What is this intended to check? Would it be just as effective to say something like `assert_not response.authorization.blank?` ?

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -5,36 +5,43 @@ def setup
...
+    assert_equal @credit_card.name,           response.params['card_holder']
+    assert_equal '%02d' % @credit_card.month, response.params['expiry_month']
+    assert_equal @credit_card.year.to_s,      response.params['expiry_year']
+    assert_equal @credit_card.number[0..5],   response.params['first_six_digits']
+    assert_equal 'FINISHED',                  response.params['return_type']
 
-    response = @gateway.purchase(@amount, @credit_card, options)
-    assert_success response
-    assert_equal 'REPLACE WITH SUCCESS MESSAGE', response.message
+    assert_equal @credit_card.number.split(//).last(4).join, response.params['last_four_digits']
```
leila-alderman: These seem far more in depth than usual for remote tests. What is the advantage of moving these sorts of checks out of unit tests and into remote tests?

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -5,36 +5,43 @@ def setup
...
   end
 
   def test_failed_purchase
-    response = @gateway.purchase(@amount, @declined_card, @options)
+    response = @gateway.purchase(@amount, @declined_card, {})
+
     assert_failure response
-    assert_equal 'REPLACE WITH FAILED PURCHASE MESSAGE', response.message
+    assert_equal 'The transaction was declined', response.message
+    assert_equal '2003', response.params['code']
```
leila-alderman: Should this be checking `response.error_code`?

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+          + timestamp + "\n\n" + '/transaction'
+
+        digest = OpenSSL::Digest.new('sha512')
+        hmac = OpenSSL::HMAC.digest(digest, @secret, message)
+
+        Base64.encode64(hmac).split.join
+      end
+
+      def parse(action, xml)
+        parse_element({:action => action}, REXML::Document.new(xml))
```
leila-alderman: Consider whether is can be replaced with the default function: https://github.com/spreedly/internal-docs/blob/master/gateway_templates/example.rb#L194

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -1,53 +1,40 @@
...
-      def purchase(money, payment, options={})
-        post = {}
-        add_invoice(post, money, options)
-        add_payment(post, payment)
-        add_address(post, payment, options)
-        add_customer_data(post, options)
-
-        commit('sale', post)
+      def purchase(money, payment_method, options={})
+        commit('purchase', build_purchase_request(money, payment_method, options), options)
```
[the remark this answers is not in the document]
leila-alderman: We talked about this being a difference between JSON and XML gateways. 

The example gateway in our docs does include a version of `purchase` for XML: https://github.com/spreedly/internal-docs/blob/master/gateway_templates/example.rb#L29
(I'm not sure what the pros and cons of implementing it one way or the other might be; I just think it's valuable to note that we do have a default for XML.)

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: To answer the question of whether this something that is done elsewhere in Active Merchant, I found several examples across gateways:

- [Authorize.net](https://github.com/activemerchant/active_merchant/blob/2821fbdd5941af0880716aa43e23a18f3f26945e/lib/active_merchant/billing/gateways/authorize_net.rb#L166)
- [S5](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/s5.rb#L130) (I'm a big fan of this example: the code looks super slick, and it seems to have been added by Spreedly.)
- [Optimal Payment](https://github.com/activemerchant/active_merchant/blob/35e2bd338c2145ba3b6d26c3e4e473e23a5dc019/lib/active_merchant/billing/gateways/optimal_payment.rb#L210)
- [IPP](https://github.com/activemerchant/active_merchant/blob/776f0735d81d69ec46d943644dd65b175ba7ea10/lib/active_merchant/billing/gateways/ipp.rb#L101)
- [Bambora Apac](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/s5.rb#L112)

Based on these examples, the above code would more accurately be written:
```
xml.transactionWithCard do
  xml.username    @options[:username]
  xml.password    Digest::SHA1.hexdigest(@options[:password]
  add_card_data(xml, payment_method)
  add_debit(xml, money, options)
end
```

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+        response = parse(action, error.response.body)
 
         Response.new(
           success_from(response),
           message_from(response),
           response,
           authorization: authorization_from(response),
-          avs_result: AVSResult.new(code: response["some_avs_response_key"]),
-          cvv_result: CVVResult.new(response["some_cvv_response_key"]),
           test: test?,
```
leila-alderman: Why were the AVS and CVV deleted? Are they just not included in the response from the gateway?

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
 
       def message_from(response)
+        response[:message] || response[:return_type]
       end
 
       def authorization_from(response)
-      end
-
-      def post_data(action, parameters = {})
+        response[:reference_id] ? response[:reference_id] + '|' + response[:purchase_id] : nil
```
leila-alderman: Looking at [the Ixopay docs](https://gateway.ixopay.com/documentation/api#transaction-response), it seems that `reference_id` is always returned, so I'm not sure whether this conditional is needed.

Perhaps also consider updating this to use either string interpolation or `.join`.

lib/active_merchant/billing/gateways/ixopay.rb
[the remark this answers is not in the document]
leila-alderman: The S5 gateway is an example of this approach in action for an XML gateway: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/s5.rb#L30

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -1,53 +1,40 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class IxopayGateway < Gateway
```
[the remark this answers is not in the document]
jasonxp: (Discussed offline) The diff is against the raw output of the gateway generator. That code forms the baseline of our ```ixopay_integration``` branch; there is no PR for it since we don't want to merge it until Ixopay is in an appropriate state to be added to Active Merchant. The baseline code can be viewed here: https://github.com/activemerchant/active_merchant/compare/ixopay_integration#diff-51a89ecca18e8620ad6cb8e059787833

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -1,53 +1,40 @@
...
-      self.test_url = 'https://example.com/test'
-      self.live_url = 'https://example.com/live'
+      self.test_url = 'https://secure.ixopay.com/transaction'
+      self.live_url = 'https://secure.ixopay.com/transaction'
 
-      self.supported_countries = ['US']
-      self.default_currency = 'USD'
+      self.supported_countries = %w(AO AQ AR AS AT AU AW AX AZ BA BB BD BE BF BG BH BI BJ BL BM BN BO BQ BQ BR BS BT BV BW BY BZ CA CC CD CF CG CH CI CK CL CM CN CO CR CU CV CW CX CY CZ DE DJ DK DM DO DZ EC EE EG EH ER ES ET FI FJ FK FM FO FR GA GB GD GE GF GG GH GI GL GM GN GP GQ GR GS GT GU GW GY HK HM HN HR HT HU ID IE IL IM IN IO IQ IR IS IT JE JM JO JP KE KG KH KI KM KN KP KR KW KY KZ LA LB LC LI LK LR LS LT LU LV LY MA MC MD ME MF MG MH MK ML MM MN MO MP MQ MR MS MT MU MV MW MX MY MZ NA NC NE NF NG NI NL NO NP NR NU NZ OM PA PE PF PG PH PK PL PM PN PR PS PT PW PY QA RE RO RS RU RW SA SB SC SD SE SG SH SI SJ SK SL SM SN SO SR SS ST SV SX SY SZ TC TD TF TG TH TJ TK TL TM TN TO TR TT TV TW TZ UA UG UM US UY UZ VA VC VE VG VI VN VU WF WS YE YT ZA ZM ZW)
+      self.default_currency = 'EUR'
       self.supported_cardtypes = [:visa, :master, :american_express, :discover]
```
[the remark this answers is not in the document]
jasonxp: We discussed this and decided to add all supported Ixopay cards that intersect with Spreedly's supported cards. This provides the consistency that we maintain with gateway card support. The ones being added are  ```:diners_club```, ```:jcb```, and ```:maestro```.

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -1,53 +1,40 @@
...
-      def authorize(money, payment, options={})
-        post = {}
-        add_invoice(post, money, options)
-        add_payment(post, payment)
-        add_address(post, payment, options)
-        add_customer_data(post, options)
-
-        commit('authonly', post)
+      def authorize(money, payment_method, options={})
+        # todo
```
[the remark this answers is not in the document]
jasonxp: Added a commented out likely implementation.

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+          'Date' => timestamp,
+          'Content-Type' => 'text/xml; charset=utf-8'
+        }
+      end
+
+      def generate_signature(http_method, xml, timestamp)
+        content_type = 'text/xml; charset=utf-8'
+
+        message = http_method + "\n" + Digest::MD5.hexdigest(xml) + "\n" + content_type + "\n" \
+          + timestamp + "\n\n" + '/transaction'
```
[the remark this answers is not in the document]
jasonxp: Done.

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +54,193 @@ def scrub(transcript)
...
+
+        digest = OpenSSL::Digest.new('sha512')
+        hmac = OpenSSL::HMAC.digest(digest, @secret, message)
+
+        Base64.encode64(hmac).split.join
```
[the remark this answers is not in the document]
jasonxp: Done.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Removed.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Done.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Done.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: These assertions are just checking the raw response elements returned by the gateway. So, the tests will fail either if our code starts passing the wrong data, or if the Ixopay response format changes. They weren't really moved out of unit tests (writing the remote tests first is a good strategy that I followed for the new gateway implementation). It would be possible, albeit more complicated, to write unit tests to check the data we're passing to the gateway (perhaps similar to what do for GSFs). However, it wouldn't be possible to check in unit tests that the gateway returns the expected data.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: It could be. The response ```error_code``` is a field that we extract from the response and name like so. ```response.params['code']``` is the raw code before extraction. I think I had been focusing the remote tests on verifying that the remote server returned the expected data, whereas the corresponding unit test checks the extracted field ```error_code```. We could do both here though.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: I think that this method should probably be replaced with a library method altogether. I'll leave it as a todo for the moment, for the sake of getting this PR merged into the integration branch ASAP since it's needed for other work on Ixopay.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Correct -- they are not present in the gateway response.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: In general, I'd say it's a good idea to check for null instead of relying on the remote API, which is out of our control. If there were an unexpected response, which could always change in the future, the result would be a runtime error without the check. In this case I think a case where the reference ID is not present is the error case (we use ```authorization_from``` in that case as well).

Revised to use string interpolation.

test/unit/gateways/ixopay_test.rb
```diff
@@ -2,33 +2,39 @@
...
-    response = @gateway.purchase(@amount, @credit_card, @options)
+    response = @gateway.purchase(@amount, @declined_card, @options)
+
     assert_failure response
-    assert_equal Gateway::STANDARD_ERROR_CODE[:card_declined], response.error_code
+    assert_equal 'The transaction was declined', response.message
+    assert_equal '2003', response.error_code
   end
 
   def test_successful_authorize
```
leila-alderman: All of the unit tests that are not yet implemented are blank with the exception of `test_scrub`. Please consider commenting out the two assert statements in that test so that the unit tests can be run cleanly.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -5,36 +5,43 @@ def setup
...
   end
 
   def test_successful_authorize_and_capture
```
leila-alderman: To enable the remote test suite to run cleanly, please add 
```
omit "Not yet implemented"
```
underneath the `def` lines for all of the tests that are not yet implemented.

jasonxp: Good idea. Done.

test/unit/gateways/ixopay_test.rb
jasonxp: Done.

test/remote/gateways/remote_ixopay_test.rb
leila-alderman: I thought that the idea of the test pyramid was that you start writing your tests with unit tests and then work your way out to more complicated things like remote tests.

If the intent of these additional `assert` statements is to ensure that the structure of the gateway response is what we expect, then perhaps it would be worthwhile to extract these out into a new `test_response_format` remote test since these seem to be beyond the scope of what is needed to ensure that a purchase request was successful.

test/remote/gateways/remote_ixopay_test.rb
leila-alderman: The line above this checks `response.message`, which is something that was extracted out from the response. I figured the intent of these remote tests was to ensure that the response that we're returning is what we expect it to be, which would include testing the `response_from_request_action` method (and therefore the `error_code_from` method).

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: Neither the `reference_id` nor the `purchase_id` is a nested parameter, so I don't believe that an error would be raised if either was missing. It just wouldn't show up in the response. 

I tested this by removing the conditional from the `authorization_from` method:
```
  "#{response[:reference_id]}|#{response[:purchase_id]}"
end
```
I then altered the `successful_purchase_response` method in the unit test file and deleted the `reference_id` line. When I then ran the `test_successful_purchase` unit test, the `response.authorization` was `"|20191016-b2bef23a30b537b90fbe"`, which indicates that the `authorization_from` method doesn't need the conditional.

leila-alderman:
Now that the test suite has been cleaned up, Rubocop is showing 35 offenses, which is why the Travis CI says it failed.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: Added that additional assertion.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: I don't, but can look into it after this branch is merged into the integration branch.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: I was actually referring to the original implementation; it had ```response[:reference_id] + '|'```, so there would be a runtime error if ```response[:reference_id]``` were nil since we would have tried to add nil to a string. With the latest version, the intent is to return ```nil``` in the case of an error (signaled by the reference ID not being present) rather than a string containing nothing except a vertical pipe.

test/remote/gateways/remote_ixopay_test.rb
leila-alderman: Why would this be an additional assertion instead of replacing `response.params['code']`? What is the value in checking both within the same test? 

I thought that the translation of `response.params['code']` to `response.error_code` was already being tested in the unit tests here: https://github.com/activemerchant/active_merchant/pull/3396/files?file-filters%5B%5D=.rb#diff-332f420dc6b50a19dc1f41508e903506R37

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -1,53 +1,44 @@
...
+      self.live_url = 'https://secure.ixopay.com/transaction'
 
-      self.supported_countries = ['US']
-      self.default_currency = 'USD'
-      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+      self.supported_countries = %w(AO AQ AR AS AT AU AW AX AZ BA BB BD BE BF BG BH BI BJ BL BM BN BO BQ BQ BR BS BT BV BW BY BZ CA CC CD CF CG CH CI CK CL CM CN CO CR CU CV CW CX CY CZ DE DJ DK DM DO DZ EC EE EG EH ER ES ET FI FJ FK FM FO FR GA GB GD GE GF GG GH GI GL GM GN GP GQ GR GS GT GU GW GY HK HM HN HR HT HU ID IE IL IM IN IO IQ IR IS IT JE JM JO JP KE KG KH KI KM KN KP KR KW KY KZ LA LB LC LI LK LR LS LT LU LV LY MA MC MD ME MF MG MH MK ML MM MN MO MP MQ MR MS MT MU MV MW MX MY MZ NA NC NE NF NG NI NL NO NP NR NU NZ OM PA PE PF PG PH PK PL PM PN PR PS PT PW PY QA RE RO RS RU RW SA SB SC SD SE SG SH SI SJ SK SL SM SN SO SR SS ST SV SX SY SZ TC TD TF TG TH TJ TK TL TM TN TO TR TT TV TW TZ UA UG UM US UY UZ VA VC VE VG VI VN VU WF WS YE YT ZA ZM ZW)
+      self.default_currency = 'EUR'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb, :maestro]
```
leila-alderman: Now that these additional card types have been added, should the corresponding remote tests be added?

jasonxp: We don't have any card-specific tests at this point. I'll add that as a follow-up item since it's a much lower priority than getting the core functionality developed before the deadline.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: Removed the initial assertion.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Added to CE-159.

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -67,55 +65,170 @@ def scrub(transcript)
...
+
+      def commit(request)
         url = (test? ? test_url : live_url)
-        response = parse(ssl_post(url, post_data(action, parameters)))
+
+        begin
+          raw_response = ssl_post(url, request, headers(request))
+        rescue StandardError => error
+          return response_from_request_error(error)
+        end
```
leila-alderman: What is the intent of this code? I'm able to delete all of this out along with the `response_from_request_error` method below, and all of the tests still pass.

jasonxp:
Thanks for all of the review @leila-alderman. I incorporated the major changes from your refactoring branch. There are a few smaller ones that I didn't totally grok, so let's plan to review them together next week. In the meantime, I'm merging this initial PR to the integration branch so we can begin the other tickets.

PR #3397, opened 2019-10-17
Quickbooks: Add OAuth 2.0 support, support void

britth:
Quickbooks is deprecating OAuth 1.0 support, and requiring OAuth
2.0 no later than December 17, 2019. This PR adds support for
OAuth 2.0, requiring the `access_token`, `refresh_token`, `client_id`
and `client_secret` when creating a gateway. When refresh_token is
present, OAuth 2.0 authentication will be attempted. If not
included, OAuth 1.0 will be attempted and should still work until
the deprecation date.

In the Quickbooks implementation, the access_token expires every
60 minutes and you must use the refresh_token to generate a new
one. There's no way to proactively check if the token is still
valid; you'll just see that the request fails with an
`AuthenticationFailed` error. This PR adds logic to each action to
check for this error, and attempt to refresh the token and reissue
the request when it happens.

This gateway also did not previously implement the void method;
support is added here. It required a change to authorization_from
method to include the request-id since that is needed in the void
request. The authorization value is split appropriately to parse out
the value needed for a given method.

Remote:
14 tests, 35 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
15 tests, 52 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -10,13 +10,10 @@ module ActiveMerchant #:nodoc:
 
       self.homepage_url = 'http://payments.intuit.com'
       self.display_name = 'QuickBooks Payments'
-      ENDPOINT =  '/quickbooks/v4/payments/charges'
-      OAUTH_ENDPOINTS = {
-        site: 'https://oauth.intuit.com',
-        request_token_path: '/oauth/v1/get_request_token',
-        authorize_url: 'https://appcenter.intuit.com/Connect/Begin',
-        access_token_path: '/oauth/v1/get_access_token'
-      }
+      BASE = '/quickbooks/v4/payments'
+      ENDPOINT =  "#{BASE}/charges"
+      VOID_ENDPOINT = "#{BASE}/txn-requests"
+      REFRESH_URI = 'https://oauth.platform.intuit.com/oauth2/v1/tokens/bearer'
 
       # https://developer.intuit.com/docs/0150_payments/0300_developer_guides/error_handling
 
@@ -51,7 +48,15 @@ module ActiveMerchant #:nodoc:
       FRAUD_WARNING_CODES = ['PMT-1000', 'PMT-1001', 'PMT-1002', 'PMT-1003']
 
       def initialize(options = {})
-        requires!(options, :consumer_key, :consumer_secret, :access_token, :token_secret, :realm)
+        # Quickbooks is deprecating OAuth 1.0 on December 17, 2019.
+        # OAuth 2.0 requires a client_id, client_secret, access_token, and refresh_token
+        # To maintain backwards compatibility, check for the presence of a refresh_token (only specified for OAuth 2.0)
+        # When present, validate that all OAuth 2.0 options are present
+        if options[:refresh_token]
+          requires!(options, :client_id, :client_secret, :access_token, :refresh_token)
+        else
+          requires!(options, :consumer_key, :consumer_secret, :access_token, :token_secret, :realm)
+        end
         @options = options
         super
       end
@@ -62,7 +67,8 @@ module ActiveMerchant #:nodoc:
         add_charge_data(post, payment, options)
         post[:capture] = 'true'
 
-        commit(ENDPOINT, post)
+        response = commit(ENDPOINT, post)
+        check_token_response(response, ENDPOINT, post)
       end
 
       def authorize(money, payment, options = {})
@@ -71,24 +77,35 @@ module ActiveMerchant #:nodoc:
         add_charge_data(post, payment, options)
         post[:capture] = 'false'
 
-        commit(ENDPOINT, post)
+        response = commit(ENDPOINT, post)
+        check_token_response(response, ENDPOINT, post)
       end
 
       def capture(money, authorization, options = {})
         post = {}
-        capture_uri = "#{ENDPOINT}/#{CGI.escape(authorization)}/capture"
+        authorization, _ = split_authorization(authorization)
         post[:amount] = localized_amount(money, currency(money))
         add_context(post, options)
 
-        commit(capture_uri, post)
+        response = commit(capture_uri(authorization), post)
+        check_token_response(response, capture_uri(authorization), post)
       end
 
       def refund(money, authorization, options = {})
         post = {}
         post[:amount] = localized_amount(money, currency(money))
         add_context(post, options)
+        authorization, _ = split_authorization(authorization)
 
-        commit(refund_uri(authorization), post)
+        response = commit(refund_uri(authorization), post)
+        check_token_response(response, refund_uri(authorization), post)
+      end
+
+      def void(authorization, options = {})
+        _, request_id = split_authorization(authorization)
+
+        response = commit(void_uri(request_id))
+        check_token_response(response, void_uri(request_id))
       end
 
       def verify(credit_card, options = {})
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -170,10 +189,12 @@ def commit(uri, body = {}, method = :post)
...
         # a legitimate HTTP error and an actual gateway transactional error.
+        headers = {}
         response = begin
           case method
           when :post
-            ssl_post(endpoint, post_data(body), headers(:post, endpoint))
+            headers = headers(:post, endpoint)
+            ssl_post(endpoint, post_data(body), headers)
           when :get
             ssl_request(:get, endpoint, nil, headers(:get, endpoint))
```
bayprogrammer: Just want to confirm, we don't need to stash the results of `headers(:get, endpoint)`, only for `headers(:post, endpoint)` above?

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -51,7 +48,11 @@ class QuickbooksGateway < Gateway
       FRAUD_WARNING_CODES = ['PMT-1000', 'PMT-1001', 'PMT-1002', 'PMT-1003']
 
       def initialize(options = {})
-        requires!(options, :consumer_key, :consumer_secret, :access_token, :token_secret, :realm)
+        if options[:refresh_token]
```
bayprogrammer: I know we generally avoid leaving comments in the source directly, but I'm wondering if this might be a good time to break that convention and leave a comment explicitly noting that the else condition is for backwards compatibility for OAuth 1? Took me a bit to realize that's what was going on.

britth: updated! let me know if that seems sufficient

lib/active_merchant/billing/gateways/quickbooks.rb
britth: it doesn't look like we're actually doing any get requests currently, but I think we should probably stash them too so we can send them to the `response_object` if needed

britth:
updated unit tests to further test oauth 2.0 changes

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -170,30 +192,37 @@ def commit(uri, body = {}, method = :post)
...
-        response_object(response)
+        response_object(response, headers)
       end
 
-      def response_object(raw_response)
+      def response_object(raw_response, headers = {})
         parsed_response = parse(raw_response)
 
+        if @options[:refresh_token]
+          parsed_response.merge!({ 'access_token' => @options[:access_token], 'refresh_token' => @options[:refresh_token] })
```
britth: After some end-to-end testing of the integration, made an update here to include refresh_token and access_token on the params object so that they can be updated on the gateway integration side of things as they change. Since access_tokens expire every hour, it's probably infeasible for a customer to update them on their end, so we need a way to actually save these updated values rather than continually refreshing with each unique transaction.

I am wondering if there are any concerns around saving these values on the response object though? Technically, without the client_id and client_secret, you couldn't use them to regenerate a token, and the access_token is only useful for a short span of time, but not sure if I'm overlooking any other possibilities. The other way I can think to get these values (for updating gateway settings on subsequent txns) is to grab them out of the txn's transcript before it's scrubbed.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -170,30 +192,37 @@ def commit(uri, body = {}, method = :post)
...
+        response_object(response, headers)
       end
 
-      def response_object(raw_response)
+      def response_object(raw_response, headers = {})
         parsed_response = parse(raw_response)
 
+        # Include access_token and refresh_token in params for OAuth 2.0
+        parsed_response['access_token'] = @options[:access_token] if @options[:refresh_token]
+        parsed_response['refresh_token'] = @options[:refresh_token] if @options[:refresh_token]
```
britth: After some end-to-end testing of the integration, made an update here to include refresh_token and access_token on the params object so that they can be updated on the gateway integration side of things as they change. Since access_tokens expire every hour, it's probably infeasible for a customer to update them on their end, so we need a way to actually save these updated values rather than continually refreshing with each unique transaction.

I am wondering if there are any concerns around saving these values on the response object though? Technically, without the client_id and client_secret, you couldn't use them to regenerate a token, and the access_token is only useful for a short span of time, but not sure if I'm overlooking any other possibilities. The other way I can think to get these values (for updating gateway settings on subsequent txns) is to grab them out of the txn's transcript before it's scrubbed.

britth:
@bayprogrammer no problem at all, thank you!

lib/active_merchant/billing/gateways/quickbooks.rb
jeremywrowe: I think it should be fine to save the tokens with the transaction.

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -170,31 +192,38 @@ def commit(uri, body = {}, method = :post)
         # The QuickBooks API returns HTTP 4xx on failed transactions, which causes a
         # ResponseError raise, so we have to inspect the response and discern between
         # a legitimate HTTP error and an actual gateway transactional error.
+        headers = {}
         response =
           begin
             case method
             when :post
-              ssl_post(endpoint, post_data(body), headers(:post, endpoint))
+              headers = headers(:post, endpoint)
```
jeremywrowe: When squinting at this logic, the get and post logic do almost the same thing. Is there a little bit of clean up that could be done here to get rid of the conditional for get and post?

lib/active_merchant/billing/gateways/quickbooks.rb
bayprogrammer: Agreed. The user of AM has to be careful to not leak those secrets (to the extent they should be kept secret), but presumably the user-caller here has access to those already themselves, so I don't think them receiving the refreshed tokens in the response object is a problem.

May be worth scrubbing out of the transcripts though (if we're not already doing that).

lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -266,8 +333,13 @@ def errors_from(response)
...
       end
 
-      def authorization_from(response)
-        response['id']
+      def authorization_from(response, headers = {})
+        [response['id'], headers['Request-Id']].join('|')
+      end
+
+      def split_authorization(authorization)
+        authorization, request_id = authorization.split('|')
```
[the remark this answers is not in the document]
britth: yep, what @bayprogrammer said!

lib/active_merchant/billing/gateways/quickbooks.rb
britth: updated!

test/remote/gateways/remote_quickbooks_test.rb
```diff
@@ -105,7 +107,24 @@ def test_invalid_login
...
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+    transcript = @gateway.scrub(transcript)
+
+    assert_scrubbed(@credit_card.number, transcript)
+    assert_scrubbed(@credit_card.verification_value, transcript)
+    assert_scrubbed(@gateway.options[:access_token], transcript)
+    assert_scrubbed(@gateway.options[:refresh_token], transcript)
```
bayprogrammer: 👍

lib/active_merchant/billing/gateways/quickbooks.rb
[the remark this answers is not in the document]
jeremywrowe: Is there a test to cover that case?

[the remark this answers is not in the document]
britth: @jeremywrowe Just added these two - do they seem sufficient?
https://github.com/activemerchant/active_merchant/pull/3397/files#diff-32c3623a9ad4afd6268002d9342b3cdeR86-R92

https://github.com/activemerchant/active_merchant/pull/3397/files#diff-32c3623a9ad4afd6268002d9342b3cdeR113-R120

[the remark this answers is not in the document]
jeremywrowe: yep!

PR #3401, opened 2019-10-19
Stripe: Remove outdated 'customer options' deprecation

### LINKED ISSUE #2421: [STRIPE] Passing the customer in the options is deprecated. Just use the response.authorization instead.
We don't have a Stripe use case, but this started popping up recently in the test suite for [Spree Gateway](https://github.com/spree/spree_gateway): 
```
/.rvm/gems/ruby-2.4.1/gems/activemerchant-1.66.0/lib/active_merchant/billing/gateways/stripe.rb:312:in `create_post_for_auth_or_purchase': Passing the customer in the options is deprecated. Just use the response.authorization instead.
```
related to the options param in the [`add_customer`](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L312) method

alexdunae:
This is no longer deprecated; in fact it's actually required.

From the Stripe docs:

**source**: A payment source to be charged. This can be the ID of a card
(i.e., credit or debit card), a bank account, a source, a token, or a
connected account.  For certain sources—namely, cards, bank accounts,
and attached sources—**you must also pass the ID of the associated customer.**

https://stripe.com/docs/api/charges/create#create_charge-source

Closes #2421 (and removes a ton of of deprecation notices from log files
all across the world)

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -516,10 +516,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_customer(post, payment, options)
-        if options[:customer] && !payment.respond_to?(:number)
-          ActiveMerchant.deprecated 'Passing the customer in the options is deprecated. Just use the response.authorization instead.'
-          post[:customer] = options[:customer]
-        end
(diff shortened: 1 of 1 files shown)


alexdunae:
@jknipp any chance of getting a review on this?

alexdunae:
Thanks!

PR #3402, opened 2019-10-23
VisaNet Peru: Pass amount in capture requests

curiousepic:
Passing amounts in capture requests will allow for partial captures.

Remote: Current test credentials are expired so this will need to be
confirmed in production environment and any issues soon addressed.

Unit:
15 tests, 75 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/visanet_peru.rb
+++ b/lib/active_merchant/billing/gateways/visanet_peru.rb
@@ -39,6 +39,7 @@ module ActiveMerchant #:nodoc:
 
       def capture(amount, authorization, options={})
         params = {}
+        params[:amount] = amount(amount) if amount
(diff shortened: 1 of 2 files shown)


curiousepic:
The requesting customer is the only user of this gateway and is willing to test in production.

molbrown:
It looks like this gateway only uses one currency? I'm remembering the complication caused by always passing `amount` on capture for the BlueSnap gateway, when trying to support partial capture and the conversion annoyance it caused for attempted full captures. 

Would be good to see them test full captures as well.

curiousepic:
Hoping Bluesnap is an outlier. Otherwise we should probably a start an initiative to address those situations at a higher level.

curiousepic:

PR #3405, opened 2019-10-24
[WIP] quickbooks update

rymwoo:
Notes for changes required to update Quickbooks from OAuth1 to OAuth2

Will be split into two PRs
first will make a non-breaking change that will select between 1.0 and 2.0 flows
second will be a breaking change that ONLY supports OAuth2.0 flow


rymwoo:
work already done.

PR #3406, opened 2019-10-25
CE-142: Ixopay transaction scrubbing

jasonxp:


lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -60,11 +60,22 @@ def supports_scrubbing?
...
+          gsub(%r((<username>)(.*)(</username>)), '\1[FILTERED]\3').
+          gsub(%r((<password>)(.*)(</password>)), '\1[FILTERED]\3').
+          gsub(%r((<pan>)(.*)(</pan>)), '\1[FILTERED]\3').
+          gsub(%r((<cvv>)\d+(</cvv>)), '\1[FILTERED]\2')
       end
 
       private
 
+      def remove_invalid_utf_8_byte_sequences(text)
+        text.encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
```
leila-alderman: Why is this single line its own method? Consider moving this single line back into the `scrub` method since it's not being reused.

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -60,11 +60,22 @@ def supports_scrubbing?
       end
 
       def scrub(transcript)
-        transcript
+        clean_transcript = remove_invalid_utf_8_byte_sequences(transcript)
+
+        clean_transcript.
+          gsub(%r((Authorization: Gateway )(.*)(:)), '\1[FILTERED]\3').
+          gsub(%r((<username>)(.*)(</username>)), '\1[FILTERED]\3').
```
leila-alderman: Is it necessary to scrub out the username? Looking at another gateway, the username was left in.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: This is an idiom common in Ruby that makes the code self-documenting. The method name tells you what the code is doing.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: I don't think there is any particular reason it needs to be scrubbed. Unscrubbed it.

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: The code that you're using feels relatively straightforward without explicitly naming it in a method. (I've never worked with the `.encode` method myself, but reading it makes it fairly apparent what's happening.) 

It feels anti-Metz to me, but I will also grant that it's a nitpick.

PR #3407, opened 2019-10-25
Allow setup of extra outside of active_merchant for Fat Zebra

montdidier:
Fat Zebra use `extra` for a myriad of additional flags and settings that are not necessarily broadly applicable. It would be nice to handle setting up extra outside of active_merchant. 

This PR will allows the original behaviour to continue while also allowing the flexibility of API consumers to pass in many complex options via extra.

One can see [here](https://docs.cloudpayments.io/reference#create-a-purchase_extra__title) the public parts of extra pertain to 3D secure passthrough data and hence I have adjusted the FZ tests to pass in this data directly using the new extra handling.


montdidier:
@jknipp Just wondering if there will be any movement on this?

jknipp:
@curiousepic Can you add this to your teams backlog?

curiousepic:
@montdidier Looking at the docs for the values possible in `extra`, it appears that there are only 3 additional values beyond what is already supported, all of them fairly standard for 3DS transactions. I think in this case it's preferred to add them individually rather than permitting arbitrary fields to be passed, particuarly since we want to move toward more standardization of 3DS data fields. We're happy to accept a PR using that approach.

Also, Cortex Command was pretty great!

montdidier:
@curiousepic Ok. I don't know if you saw my other PR https://github.com/activemerchant/active_merchant/pull/3409 where I start to do what you suggest - it could be better to just start there and discard this one, I just wasn't sure which would be more digestible upfront.

In this PR there are yet unspoken requirements. There are actually other fields that are not documented on the FZ site that we use and we are hoping to be able to discard our active_merchant fork and still support the features we currently use at FZ.

Now, if we choose to stick close to the paradigm of being explicit about the fields passed through I probably need to open a conversation with FZ to see if they will be ok with making these fields known. I think they will be ok with that but I didn't want to assume which is why I approached this piece in this fashion I did. 

I can open a dialog with them immediately if the preference is to be explicit?

I am glad you enjoyed Cortex Command!

curiousepic:
@montdidier I see, good to know. I don't think we're inherently against fields that aren't documented by a gateway, as long as the gateway is comfortable with it. #3409 is looking good so far; you can add me as a reviewer when it's ready.

PR #3408, opened 2019-10-26
Use current Mastercard URLs

ktully:
- MiGS, TNSPay and DataCash are all now under Mastercard Payment Gateway umbrella
- *.tnspayments.com is deprecated in favour of *-gateway.mastercard.com


ktully:
I've updated documentation URLs where the old ones no longer resolve appropriately.

I've updated the old TNSPay hostnames to the go-forward Mastercard hostnames (which resolve to the same IP addresses underneath).

PR #3409, opened 2019-10-28
Fat Zebra can use standardized 3ds fields

montdidier:
The Fat Zebra adapter does not currently support the standard presentation of 3DS fields as documented [here](https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields). 

This PR allows the adapter to pass on the standard fields at least for 3DS v1. FZ does not at this juncture support 3DS v2.

This PR also allows an extras field with additional keys and values to be passed in unmolested to support various other functions that might be available to the account holder.

~~This PR builds on https://github.com/activemerchant/active_merchant/pull/3407 and I would imagine be merged afterwards.~~

### DIFF
--- a/lib/active_merchant/billing/gateways/fat_zebra.rb
+++ b/lib/active_merchant/billing/gateways/fat_zebra.rb
@@ -125,11 +125,13 @@ module ActiveMerchant #:nodoc:
       def add_extra_options(post, options)
         extra = {}
         extra[:ecm] = '32' if options[:recurring]
-        extra[:cavv] = options[:cavv] if options[:cavv]
-        extra[:xid] = options[:xid] if options[:xid]
-        extra[:sli] = options[:sli] if options[:sli]
+        extra[:cavv] = options[:cavv] || options.dig(:three_d_secure, :cavv) if options[:cavv] || options.dig(:three_d_secure, :cavv)
+        extra[:xid] = options[:xid] || options.dig(:three_d_secure, :xid) if options[:xid] || options.dig(:three_d_secure, :xid)
+        extra[:sli] = options[:sli] || options.dig(:three_d_secure, :eci) if options[:sli] || options.dig(:three_d_secure, :eci)
         extra[:name] = options[:merchant] if options[:merchant]
         extra[:location] = options[:merchant_location] if options[:merchant_location]
+        extra[:card_on_file] = options.dig(:extra, :card_on_file) if options.dig(:extra, :card_on_file)
+        extra[:auth_reason]  = options.dig(:extra, :auth_reason) if options.dig(:extra, :auth_reason)
         post[:extra] = extra if extra.any?
       end
 
--- a/test/remote/gateways/remote_fat_zebra_test.rb
+++ b/test/remote/gateways/remote_fat_zebra_test.rb
@@ -188,6 +188,24 @@ class RemoteFatZebraTest < Test::Unit::TestCase
     assert_match %r{Extra/cavv is required for SLI 05}, response.message
   end
 
+  def test_successful_purchase_with_3DS_information_using_standard_fields
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(three_d_secure: { cavv: 'MDRjN2MxZTAxYjllNTBkNmM2MTA=', xid: 'MGVmMmNlMzI4NjAyOWU2ZDgwNTQ=', eci: '05' }))
+    assert_success response
+    assert_equal 'Approved', response.message
+  end
+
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/fat_zebra.rb
```diff
@@ -122,11 +122,11 @@ def add_creditcard(post, creditcard, options = {})
       end
 
       def add_extra_options(post, options)
-        extra = {}
+        extra = options.fetch(:extra, {})
         extra[:ecm] = '32' if options[:recurring]
-        extra[:cavv] = options[:cavv] if options[:cavv]
-        extra[:xid] = options[:xid] if options[:xid]
-        extra[:sli] = options[:sli] if options[:sli]
+        extra[:cavv] = options[:cavv] || options.dig(:three_d_secure, :cavv) if options[:cavv] || options.dig(:three_d_secure, :cavv)
```
montdidier, line 127: The intent of this is to support both the old passthrough method and the new [standardised way of doing it](https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields).

lib/active_merchant/billing/gateways/fat_zebra.rb
```diff
@@ -122,11 +122,11 @@ def add_creditcard(post, creditcard, options = {})
...
+        extra[:xid] = options[:xid] || options.dig(:three_d_secure, :xid) if options[:xid] || options.dig(:three_d_secure, :xid)
+        extra[:sli] = options[:sli] || options.dig(:three_d_secure, :eci) if options[:sli] || options.dig(:three_d_secure, :eci)
```
montdidier, line 130: FZ acknowledge they have used the Visa specific term here but the value is the same.

montdidier:
@curiousepic what do you think?

I am closing #3407  and focussing on this PR.

curiousepic:
@montdidier Apologies if it was unclear, but I was expecting the new fields to be explicitly added in this approach (in the manner than xid/cavv/sli are), rather than passing through anything that's in the `extra` hash. So my original comment on #3407 still applies here. As far as I could tell, the primary reason not to do so is that the undocumented fields would be exposed to the public, but it sounds like FZ are ok with that, so I think we should opt to map them explicitly.

montdidier:
@curiousepic ok, will update with the field explicitly.

montdidier:
@curiousepic There we go. That might do it. Let me know what you think.

test/remote/gateways/remote_fat_zebra_test.rb
```diff
@@ -177,17 +177,23 @@ def test_successful_purchase_with_metadata
   end
 
   def test_successful_purchase_with_3DS_information
-    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(cavv: 'MDRjN2MxZTAxYjllNTBkNmM2MTA=', xid: 'MGVmMmNlMzI4NjAyOWU2ZDgwNTQ=', sli: '05'))
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(three_d_secure: { cavv: 'MDRjN2MxZTAxYjllNTBkNmM2MTA=', xid: 'MGVmMmNlMzI4NjAyOWU2ZDgwNTQ=', eci: '05' }))
```
curiousepic, line 180: If we're retaining the old form of passing in these fields, this should probably be a new test.

curiousepic: I think that's my only quibble though; once these are tested individually, please squash into one commit and we'll be able to bring it in soon!

montdidier: Good point, i'll add a test for the old way too.

montdidier:
@curiousepic there we go. extra test and squashed.

curiousepic:
@montdidier Are you also getting 4 remote test failures for assertions of transaction failure?

montdidier:
> @montdidier Are you also getting 4 remote test failures for assertions of transaction failure?

no. when I run the below locally it runs fine? 

`rake test TEST=test/remote/gateways/remote_fat_zebra_test.rb`

montdidier:
running the whole test suite seems fine too?

curiousepic:
@montdidier I see what's going on, no worries. I'll bring this in soon, thanks for the changes!

montdidier:
@curiousepic Thank you. Much appreciated, this will help us come closer to using vanilla active_merchant.

curiousepic:
Made a couple tweaks to the tests to prevent generated order_id collisions, merged!

PR #3410, opened 2019-10-28
Add tests for Ixopay exception handling

jasonxp:
The Active Merchant network methods raise an exception for any non-2xx status code returned from the gateway. There is exception handling in place in the Ixopay gateway to handle this. This pull request adds test coverage for that exception handling.

### DIFF
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -184,26 +184,12 @@ module ActiveMerchant #:nodoc:
         url = (test? ? test_url : live_url)
 
         # ssl_post raises an exception for any non-2xx HTTP status from the gateway
-        begin
-          raw_response = ssl_post(url, request, headers(request))
-        rescue StandardError => error
-          return response_from_request_error(error)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -183,6 +183,7 @@ def new_transaction_id
       def commit(request)
         url = (test? ? test_url : live_url)
 
+        # ssl_post raises an exception for any non-2xx HTTP status from the gateway
         begin
```
leila-alderman: Consider refactoring the `commit` method to use a rescue statement similar to this example: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/kushki.rb#L137

By bringing more logic into the `rescue` statement, you should be able to delete the `response_from_request_error` since it's so similar to the response statement contained inside `commit`.

jasonxp: I like that - good find. Done.

jasonxp:
Merged.

PR #3412, opened 2019-10-28
Implement authorize method for Ixopay

therufs:
Remote: 
16 tests, 21 assertions, 0 failures, 0 errors, 0 pendings, 12 omissions, 0 notifications
100% passed

Unit: 
14 tests, 19 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -43,23 +43,23 @@ def test_failed_purchase
     assert_equal '2003', response.error_code
   end
 
-  def test_successful_authorize_and_capture
-    omit 'Not yet implemented'
-
+  def test_successful_authorize#_and_capture
```
therufs: Happy to handle this in some less goofy way :D didn't want to lose info about the standard being to test these together.

therufs: Happy to handle this in some less goofy way :D didn't want to lose info about the standard being to test these together.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -43,23 +43,23 @@ def test_failed_purchase
...
   end
 
   def test_failed_authorize
-    omit 'Not yet implemented'
-
     response = @gateway.authorize(@amount, @declined_card, @options)
     assert_failure response
-    assert_equal 'REPLACE WITH FAILED AUTHORIZE MESSAGE', response.message
+    assert_equal 'The transaction was declined', response.message
+    assert_equal 'ERROR',                        response.params['return_type']
```
leila-alderman: There might not be, but I'm wondering whether if there's an `error_code` returned that we could check here instead.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -43,23 +43,23 @@ def test_failed_purchase
...
 
-  def test_successful_authorize_and_capture
-    omit 'Not yet implemented'
-
+  def test_successful_authorize#_and_capture
     auth = @gateway.authorize(@amount, @credit_card, @options)
     assert_success auth
+    assert_equal 'FINISHED', auth.message
+    assert_not_nil auth.params['purchase_id']
+    assert_not_nil auth.params['reference_id']
```
leila-alderman: I'm wondering whether these two lines could be replaced with something that tests `auth.authorization`.

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -130,6 +134,22 @@ def add_debit(xml, money, options)
...
+
+        xml.preauthorize do
+          xml.transactionId new_transaction_id
+
+          add_customer_data(xml, options)
+
+          xml.amount      money
+          xml.currency    currency
+          xml.description description
+          xml.callbackUrl(options[:callback_url] || 'http://example.com')
```
leila-alderman: Are the parentheses here needed? (I have no idea.)

jasonxp: I think Ruthan followed my example here. I used parentheses for the method call for clarity. Without them, I think the reader needs more knowledge of operator precedence to know whether the statement gets evaluated as

```xml.callbackUrl(options[:callback_url] || 'http://example.com')```
or 
```xml.callbackUrl(options[:callback_url]) || http://example.com```

File: test/remote/gateways/remote_ixopay_test.rb (hunk 1 above)
Comment (jasonxp) [reply test/remote/gateways/remote_ixopay_test.rb]: Looks fine to me. 👍

File: lib/active_merchant/billing/gateways/ixopay.rb (hunk 4 above)
Comment (leila-alderman) [reply lib/active_merchant/billing/gateways/ixopay.rb]: My thought was to write this line more like this:
`xml.callbackUrl   options[:callback_url] || 'http://example.com'`

File: lib/active_merchant/billing/gateways/ixopay.rb [hunk 5]
```diff
@@ -130,6 +134,22 @@ def add_debit(xml, money, options)
         end
       end
 
+      def add_preauth(xml, money, options)
+        currency    = options[:currency] || currency(money)
+        description = options[:description].blank? ? 'Preauthorize' : options[:description]
```
leila-alderman: I don't think that this is necessary. The transactions seem to work fine without sending anything for the `description` parameter. Instead, consider replacing line 148 below with
`xml. description options[:description]`

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -130,6 +134,22 @@ def add_debit(xml, money, options)
...
+
+        xml.preauthorize do
+          xml.transactionId new_transaction_id
+
+          add_customer_data(xml, options)
+
+          xml.amount      money
+          xml.currency    currency
```
leila-alderman: Optionally, to make this more concise, you could delete line 138 and replace this line with
`xml.currency options[:currency] || currency(money)`

test/remote/gateways/remote_ixopay_test.rb
therufs: Happy to handle this in some less goofy way :D didn't want to lose info about the standard being to test these together.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Yep. I mean that I find it ambiguous as to whether ```xml.callbackUrl options[:callback_url] || 'http://example.com'``` is equivalent to ```xml.callbackUrl(options[:callback_url] || 'http://example.com')``` or ```xml.callbackUrl(options[:callback_url]) || 'http://example.com'``` without the reader having good knowledge of Ruby's precedence evaluation.

lib/active_merchant/billing/gateways/ixopay.rb
therufs: Do we want to get rid of it in the `purchase` method too? (or has that maybe already happened and my branch was not up to date?)

jasonxp: I believe I placed the hardcoded fallback in place because the Ixopay docs indicated that the description is required under certain conditions. I did not find details on what those conditions are, and had also seen this pattern with Worldpay. That said, I'll leave it up to you @therufs.

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: Hmm, okay. This seems quite straightforward to me, so I'm not seeing the confusion here. It's a pretty minor point, though, so either way is fine; I was mostly curious.

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: Thanks for clarifying @jasonxp. The Ixopay docs do indeed indicate that the `description` field is "conditionally mandatory". (https://gateway.ixopay.com/documentation/direct-pci-enabled-api?xml#transaction-data)

If that's something that we want to enforce to be on the safe side, then it should probably be wrapped in a test for both the `authorize` and `purchase` methods.

therufs: I would assume that in the cases when the field is mandatory, a specific value will be required, and we want it to be the user's responsibility to populate the field accordingly, rather than ours -- kind of a built-in GSF. I think it's unlikely that leaving the field in will cause any problems, so I'm not averse to leaving it; and I'm going to go ahead and updated the unit tests to check for it, but am not sure those updates actually give us information about whether we can expect transactions sent to the gateway to succeed.

lib/active_merchant/billing/gateways/ixopay.rb
therufs: After much contemplation, I initially opting in favor of the parens because I feel like they make the delineation between the method and the argument a little more legible. 

However, this made my attempt at combining the line formerly known as 138 with the line currently known as 148 look funny, so I've now tried something possibly worse. :D I'm not super attached to it, though, so whatever we decide we like best is good.

test/remote/gateways/remote_ixopay_test.rb
therufs: We can definitely check for presence; I couldn't find any claims in the doc about how the authorization will be constituted, so I'd like to keep the existing checks, since if they do at some point diverge that may be something we'd want to know for (maybe among other things) a 3DS2 implementation.

test/remote/gateways/remote_ixopay_test.rb
therufs: we can check it also!

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: Not to make your call on this harder, but I took the approach of omitting the parens on my `refund` PR that was approved by @jasonxp and has been merged: 
https://github.com/activemerchant/active_merchant/pull/3413/files#diff-51a89ecca18e8620ad6cb8e059787833R141

Barring that, though, I think that the approach you took is a good one @therufs.

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: Hmm, that's an interesting point. We don't actually have any idea of what Ixopay would require that field to be, much less when it's required. Maybe this is a good thing to leave up to our customers instead of instilling a false sense of security. 🤷‍♀

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -53,23 +53,25 @@ def test_failed_authentication
...
 
   def test_failed_authorize
-    omit 'Not yet implemented'
-
     response = @gateway.authorize(@amount, @declined_card, @options)
     assert_failure response
-    assert_equal 'REPLACE WITH FAILED AUTHORIZE MESSAGE', response.message
+    assert_equal 'The transaction was declined', response.message
+    assert_equal 'ERROR',                        response.params['return_type']
+    assert_equal response.error_code, "2003"
```
leila-alderman: Very minor nitpick, but this assert statement has the inverse format of the others above it.
```suggestion
    assert_equal "2003", response.error_code
```

jasonxp:
@leila-alderman @therufs Since we need this PR to unblock other dependencies, I'll address the outstanding comments since Ruthan is traveling today. Since I can't push to Ruthan's fork, I'll merge this and open a new PR with the changes.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: Good catch, since the order actually effects how the test results are stated (expected vs. actual). Updated in follow-up branch.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: I actually would expect a description field to be free-form, rather than provide a specific value. We can follow up with Ixopay to find out specifics around the "conditionally mandatory" statement.

therufs: Right -- I was thinking in cases where it's something like "provide the authorization from the initial transaction" or something for recurring charges.

PR #3413, opened 2019-10-28
Ixopay: Implement refund

leila-alderman:
Implements the `refund` transaction type for the new Ixopay gateway.

CE-140

Unit:
14 tests, 16 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
16 tests, 23 assertions, 0 failures, 0 errors, 0 pendings, 11 omissions,
0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -39,8 +39,11 @@ module ActiveMerchant #:nodoc:
       end
 
       def refund(money, authorization, options={})
-        # todo
-        # commit('refund', build_refund_request(money, payment_method, options), options)
+        request = build_xml_request do |xml|
+          add_refund(xml, money, authorization)
+        end
+
+        commit(request)
       end
 
       def void(authorization, options={})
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -130,6 +133,15 @@ def add_debit(xml, money, options)
         end
       end
 
+      def add_refund(xml, money, authorization)
+        xml.refund do
+          xml.transactionId           new_transaction_id
+          xml.referenceTransactionId  authorization.split('|')[0]
```
jasonxp: It looks like if a ```nil``` authorization is passed in, we'll generate a ```NoMethodError``` here. It would probably be more helpful for the code to survive to where we make the request to the gateway. That way, we'd ultimately end up with a user-visible error that indicates the authorization (or ```referenceTransactionId```) was invalid. So, a more robust statement would just be null-safe, perhaps ```authorization&.split.try(:[], 0)``` or just a ternary checking ```.present?```.

jasonxp:
Added a thought on null safety, but otherwise looks good. 👍

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: Thanks for pointing this out! I've updated the code as well as both the unit and remote tests to ensure that the code is able to handle `nil` for the authorization.

PR #3415, opened 2019-10-30
Credorax: Stop always sending r1 parameter

molbrown:
Credorax users must be enabled in their Credorax account to specify
their processor (via the r1 parameter). Maintains ability to pass in a
value for r1 via the processor option, and stops default sending it
with the value of CREDORAX when none is specified.

Necessary change due to an issue with a Spreedly customer having
rejected 3DS1 transactions due to inclusion of an r1 parameter in their
requests.

ECS-778

Unit:
58 tests, 264 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
35 tests, 128 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_credorax_test.rb
+++ b/test/remote/gateways/remote_credorax_test.rb
@@ -64,11 +64,13 @@ class RemoteCredoraxTest < Test::Unit::TestCase
     assert_equal 'Succeeded', response.message
   end
 
+  # Having processor-specification enabled in Credorax test account causes 3DS tests to fail without an r1 parameter.
   def test_successful_purchase_with_auth_data_via_3ds1_fields
     options = @options.merge(
       eci: '02',
       cavv: 'jJ81HADVRtXfCBATEp01CJUAAAA=',
-      xid: '00000000000000000501'
+      xid: '00000000000000000501',
+      processor: 'CREDORAX'
     )
 
     response = @gateway.purchase(@amount, @fully_auth_card, options)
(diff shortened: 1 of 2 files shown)


test/remote/gateways/remote_credorax_test.rb
```diff
@@ -64,11 +64,13 @@ def test_successful_purchase_with_extra_options
     assert_equal 'Succeeded', response.message
   end
 
+  # Having processor-specification enabled in Spreedly's test account causes 3DS tests to fail without an r1 parameter.
```
jeremywrowe: The wording here should be a little more generic since this applies to more than just Spreedly.

test/remote/gateways/remote_credorax_test.rb
```diff
@@ -64,11 +64,13 @@ def test_successful_purchase_with_extra_options
     assert_equal 'Succeeded', response.message
   end
 
+  # Having processor-specification enabled in Credorax test account causes 3DS tests to fail without an r1 parameter.
```
leila-alderman: Consider moving these comments down to just above the `processor` line to help draw the line between `r1` and `processor`.

molbrown:
Unfortunately I don't think I'm able to match this change in a test, because of how our test account is configured. Our customer has been asked by Credorax to remove `r1` from their 3DS transactions in order to stop their declines, but because we were sending a default value, this was not possible for them to do without a change here. (I get the opposite behavior in these 3DS remote tests when removing the default processor, which is why I needed to pass it directly.)

PR #3416, opened 2019-10-31
Add UYI to list of currencies without fractions

curiousepic:
This brings the default list of currencies without fractions to include
all currencies noted as such in https://en.wikipedia.org/wiki/ISO_4217

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -123,8 +123,9 @@ module ActiveMerchant #:nodoc:
       class_attribute :supported_cardtypes
       self.supported_cardtypes = []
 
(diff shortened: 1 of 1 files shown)


curiousepic:
No Spreedly customers currently transact in UYI.

lib/active_merchant/billing/gateway.rb
```diff
@@ -123,8 +123,9 @@ def generate_unique_id
       class_attribute :supported_cardtypes
       self.supported_cardtypes = []
 
+      # This default list of currencies without fractions are from https://en.wikipedia.org/wiki/ISO_4217
```
bdewater: Nitpick: the primary source is https://www.currency-iso.org/en/home/tables/table-a1.html

curiousepic:

PR #3418, opened 2019-10-31
CE-138: Ixopay capture

jasonxp:


test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -53,43 +53,45 @@ def test_failed_authentication
...
+  def test_failed_authorize
+    response = @gateway.authorize(@amount, @declined_card, @options)
+    assert_failure response
+    assert_equal 'The transaction was declined', response.message
+    assert_equal 'ERROR',                        response.params['return_type']
+    assert_equal response.error_code, '2003'
+  end
 
+  def test_failed_capture
     response = @gateway.capture(@amount, '')
```
leila-alderman: To test the worst case scenario here, this should test sending in `nil` as the authorization.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -53,43 +53,45 @@ def test_failed_authentication
...
-    #assert_equal 'REPLACE WITH SUCCESS MESSAGE', capture.message
-  end
-
-  def test_failed_authorize
-    response = @gateway.authorize(@amount, @declined_card, @options)
-    assert_failure response
-    assert_equal 'The transaction was declined', response.message
-    assert_equal 'ERROR',                        response.params['return_type']
-    assert_equal response.error_code, "2003"
+    assert capture = @gateway.capture(@amount, auth.params['reference_id'])
```
leila-alderman: Looking through other gateways, it should be standard to pass in `auth.authorization` here and parse out the reference, as suggested above.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -53,43 +53,45 @@ def test_failed_authentication
...
-    omit 'Not yet implemented'
+    auth_amount = 500
+    capture_amount = 400
 
-    auth = @gateway.authorize(@amount, @credit_card, @options)
+    auth = @gateway.authorize(auth_amount, @credit_card, @options)
     assert_success auth
 
-    assert capture = @gateway.capture(@amount-1, auth.authorization)
+    assert capture = @gateway.capture(capture_amount, auth.params['reference_id'])
```
leila-alderman: Looking through other gateways, it should be standard to pass in `auth.authorization` here and parse out the reference, as suggested above.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -53,43 +53,45 @@ def test_failed_authentication
...
-    assert_equal response.error_code, "2003"
+    assert capture = @gateway.capture(@amount, auth.params['reference_id'])
+    assert_success capture
+    assert_equal 'FINISHED', capture.message
   end
 
   def test_partial_capture
-    omit 'Not yet implemented'
+    auth_amount = 500
+    capture_amount = 400
```
leila-alderman: What is the advantage of adding these additional lines of code instead of following the default of using `@amount-1`?

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -173,6 +176,17 @@ def add_preauth(xml, money, options)
         end
       end
 
+      def add_capture(xml, money, authorization, options)
+        currency = options[:currency] || currency(money)
+
+        xml.capture do
+          xml.transactionId          new_transaction_id
+          xml.referenceTransactionId authorization
```
leila-alderman: I think that this should be
```
xml.referenceTransactionId  authorization&.split('|')&.first
```
as done in the `refund` method above so that the customer is able to just pass in the same `authorization` that we returned to them.

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -173,6 +176,17 @@ def add_preauth(xml, money, options)
...
+          xml.amount                 localized_amount(money, currency)
```
leila-alderman: Is calling `localized_amount` here necessary? If so, then should this also be used on all of the other transaction types?

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -173,6 +176,17 @@ def add_preauth(xml, money, options)
...
+          xml.currency               currency
```
leila-alderman: This can be made more concise by deleting line 180 and then copying line 158 above.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Makes sense. 👍

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Yes to both, I believe. ```localized_amount``` gives you an amount in (for example) dollars, converting e.g. 100 to 1.00. That is the format specified in the docs, and I think we just missed this conversion initially. I'll update all of the other instances in this PR.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: I wrote this before I noticed that there was already a ```test_partial_capture```, so I wasn't going for a particular advantage.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: I think the ```currency``` definition you mentioned deleting is needed since it is used for the amount as well.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: Done.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: Done.

lib/active_merchant/billing/gateways/ixopay.rb
jasonxp: Done.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: Done.

test/remote/gateways/remote_ixopay_test.rb
jasonxp: (Also, changed it to the subtraction version)

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -14,7 +14,7 @@ class IxopayGateway < Gateway
       self.display_name = 'Ixopay'
 
       def initialize(options={})
-        requires!(options, :username, :password, :secret)
+        requires!(options, :username, :password, :secret, :api_key)
```
jasonxp: We noticed that the ```api_key``` was missing as a requirement here, so went ahead and fixed that in this pull request.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -44,53 +44,57 @@ def test_failed_purchase
   end
 
   def test_failed_authentication
-    gateway = IxopayGateway.new(username: 'baduser', password: 'badpass', secret: 'badsecret')
+    gateway = IxopayGateway.new(
+      username: 'baduser',
+      password: 'badpass',
+      secret:   'badsecret',
+      api_key:  'badapikey'
```
jasonxp: Just added the newly-required API key here.

test/unit/gateways/ixopay_test.rb
```diff
@@ -4,7 +4,12 @@ class IxopayTest < Test::Unit::TestCase
   include CommStub
 
   def setup
-    @gateway = IxopayGateway.new(username: 'username', password: 'password', secret: 'secret')
+    @gateway = IxopayGateway.new(
+      username: 'username',
+      password: 'password',
+      secret:   'secret',
+      api_key:  'api_key'
```
jasonxp: Just added the newly-required API key here.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -44,53 +44,57 @@ def test_failed_purchase
...
+    )
+
     response = gateway.purchase(@amount, @credit_card, {})
 
     assert_failure response
 
-    assert_equal 'Invalid Signature: Invalid authorization header', response.message
```
jasonxp: Just updated the error message here now that the API key is passed in for this test as well.

test/unit/gateways/ixopay_test.rb
leila-alderman: Now that this has been updated, the error message included in `test_failed_authentication` below should also be updated.

test/unit/gateways/ixopay_test.rb
```diff
@@ -77,9 +82,26 @@ def test_failed_authorize
     assert_equal '2003', response.error_code
   end
 
-  def test_successful_capture; end
+  def test_successful_capture
+    @gateway.expects(:ssl_post).returns(successful_capture_response)
 
-  def test_failed_capture; end
+    response = @gateway.capture(@amount, 'auth-123')
```
leila-alderman: Consider changing this to something more similar to the actual authorization value passed in, e.g., `'00eb44f8f0382443cce5|20191028-00eb44f8f0382443cce5'`.

test/unit/gateways/ixopay_test.rb
```diff
@@ -77,9 +82,26 @@ def test_failed_authorize
...
+    assert_success response
+    assert_equal 'FINISHED', response.message
+    assert_equal '17dd1e0b09221e9db038|20191031-17dd1e0b09221e9db038', response.authorization
+    assert response.test?
+  end
+
+  def test_failed_capture
+    @gateway.expects(:ssl_post).returns(failed_capture_response)
+
+    response = @gateway.capture(@amount, 'auth-123')
```
leila-alderman: Consider changing the authorization value here to `nil` or something that more clearly illustrates that the value is invalid.

leila-alderman:
Looking at the CI build, the tests are no longer passing after the recent commits.

test/unit/gateways/ixopay_test.rb
jasonxp: Done.

test/unit/gateways/ixopay_test.rb
jasonxp: Done.

test/unit/gateways/ixopay_test.rb
jasonxp: Done.

jasonxp:
@leila-alderman Those tests were actually failing before that commit. I can't reproduce locally, and the errors don't make sense to me. (It looks like it's saying that ```@gateway.capture``` is being invoked with 0 arguments, but requires 1, and neither of those things seems true). Let me know if you can reproduce or have ideas.

leila-alderman:
@jasonxp I'm not entirely sure what's causing the issue with some of the CI tests failing, but they all seem to be referencing a commit that isn't actually showing up in the list of commits for this PR: https://github.com/activemerchant/active_merchant/commit/5a53cab1186858e4467e6cef2b7e463764b2f585. Maybe doing a rebase, squash, and force-push would calm things down?

jasonxp:
@leila-alderman That was a good idea, but didn't work, either. I think since we're not going to ```master``` yet, we could merge and see if this builds properly in the integration branch. Thoughts?

leila-alderman:
@jasonxp I felt really uncomfortable ignoring the CI failures, so I dug deeper into the problem and have found the solution. 

I created a new PR (https://github.com/activemerchant/active_merchant/pull/3422) to test the solution, and that is now showing all the Travis CI checks as passing.

I changed `xml.capture` in the `add_capture` method to `xml.tag! 'capture'` instead since it turns out that `capture` was the name of an actual method in Active Support version 4.2. When running the gem using this older version of Active Support, this syntax was calling this deprecated method instead of simply adding an xml tag.

See [the deprecated Active Support
method](https://github.com/rails/rails/blob/4-2-stable/activesupport/lib/active_support/core_ext/kernel/reporting.rb#L89).

The [stack trace from this PR](https://travis-ci.org/activemerchant/active_merchant/jobs/606153338?utm_medium=notification&utm_source=github_status) also hints at the problem:
```
/home/travis/build/activemerchant/active_merchant/lib/active_merchant/billing/gateways/ixopay.rb:194:in `add_capture'
/home/travis/build/activemerchant/active_merchant/gemfiles/vendor/bundle/ruby/2.3.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext/kernel/reporting.rb:89:in `capture'
Error: test_failed_capture(IxopayTest): ArgumentError: wrong number of arguments (given 0, expected 1)
```

jasonxp:
@leila-alderman Ha. I was going to tell you I had found that root cause to, but hadn't yet figured out the alternative since my original use of ```tag!``` didn't work the same with the new parsing. Perhaps this is the historical reason that ActiveMerchant uses ```tag!``` instead of our format. Thanks for digging into it!

jasonxp:
To fix the aforementioned issue, I just had to use ```tag!``` to build the entire ```capture``` request instead of mixing it with the ```build_xml_request``` helper. All unit and remote tests are passing now, so we should be good to go.

jasonxp:
> I'm super curious about why the `.tag!` has to be used for the entire request to work...
> 
> This looks good to go!

I think because we call ```tag!``` on instances of ```Builder::XmlMarkup```, while ```build_xml_request``` works with instances of ```Nokogiri::XML::Builder```. The latter adds an actual tag to the XML named ```tag``` when it receives the ```tag!``` message.

jasonxp:
Merged.

PR #3419, opened 2019-10-31
Quickbooks: Make refresh optional

britth:
While access_tokens expire every hour, there are potential
complications in refreshing them in certain workflows. Adds a flag,
`allow_refresh` to optionally refresh the access_token after making
a request. Also marks as a failure if authentication failed.

Remote:
16 tests, 40 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
21 tests, 114 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -68,7 +68,7 @@ module ActiveMerchant #:nodoc:
         post[:capture] = 'true'
 
         response = commit(ENDPOINT, post)
-        check_token_response(response, ENDPOINT, post)
+        check_token_response(response, ENDPOINT, post, options)
       end
 
       def authorize(money, payment, options = {})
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -78,7 +78,7 @@ def authorize(money, payment, options = {})
         post[:capture] = 'false'
 
         response = commit(ENDPOINT, post)
-        check_token_response(response, ENDPOINT, post)
+        check_token_response(response, ENDPOINT, options, post)
```
curiousepic: Minor quibble but the usual argument pattern has `options`  last.

britth: good call, updated! 👍

PR #3421, opened 2019-11-01
Mf/activesupport 5

ScottDuke:
Updated fork to work for rails 5.1.7. Rebased onto the current version of active merchant


ScottDuke:
Wrongly created PR.

PR #3423, opened 2019-11-01
Ixopay: Implement verify action

leila-alderman:
Implement the `verify` action for the new Ixopay gateway, including
remote and unit tests.

CE-135 / CE-141

Unit:
15 tests, 59 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
16 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 3 omissions,
0 notifications
100% passed

### DIFF
--- a/test/unit/gateways/ixopay_test.rb
+++ b/test/unit/gateways/ixopay_test.rb
@@ -113,11 +113,27 @@ class IxopayTest < Test::Unit::TestCase
     assert_equal 'Transaction of type "void" requires a referenceTransactionId', response.message
   end
 
-  def test_successful_verify; end
+  def test_successful_verify
+    @gateway.expects(:ssl_post).times(2).returns(successful_authorize_response, successful_void_response)
+    response = @gateway.verify(credit_card('4111111111111111'), @options)
 
-  def test_successful_verify_with_failed_void; end
+    assert_success response
+    assert_equal 'FINISHED', response.message
+  end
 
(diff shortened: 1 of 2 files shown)


test/unit/gateways/ixopay_test.rb
```diff
@@ -113,11 +113,27 @@ def test_failed_void
...
+  def test_successful_verify
+    @gateway.expects(:ssl_post).times(2).returns(successful_authorize_response, successful_void_response)
+    response = @gateway.verify(credit_card('4111111111111111'), @options)
 
-  def test_successful_verify_with_failed_void; end
+    assert_success response
+    assert_equal 'FINISHED', response.message
+  end
+
+  def test_successful_verify_with_failed_void
```
jasonxp: It wasn't in the generated template, but do you think there would be value in testing this scenario in the remote tests?

leila-alderman: I don't think so. It's a pretty well established pattern in the tests, and it's also just testing `r.process(:ignore_result)` in the gateway code. If that works for a unit test, I don't see the value in also testing it as part of a remote test. On a practical level, I'm also not even sure how I would force a failed void response from within the `verify` method since it pulls the authorization value directly from the `authorize` response.

jasonxp: Makes sense. Thanks 👍

PR #3424, opened 2019-11-04
Update supported countries list. Add currencies without fractions / with 3 decimal places

nicolas-maalouf-cko:
* Updated list of supported countries
* Specify currencies without fractions / with 3 decimal places according to [this](https://docs.checkout.com/docs/calculating-the-value)


nicolas-maalouf-cko:
@bbraschi should be good now!

PR #3432, opened 2019-11-05
Credorax: Add referral CTF transactions

leila-alderman:
Added support for the referral CFT transaction type on the Credorax
gateway (transaction type [34]).

To simplify this new transaction type, it is selected by passing a new
`referral_cft` gateway specific field while requesting a `refund`
transaction. Although the referral CFT transaction defaults to a credit
request when it has not been enabled on the Credorax gateway
integration, the referral CFT requires passing a `g2` value, i.e.,
"Previous Response ID". Therefore, because this transaction type
requires a reference authorization from a previous transaction, its
implmentation was more similar to a refund than a credit.

Because Spreedly's current Credorax gateway integration does not have
the ability to make referral CFT transactions, the remote and unit tests
for this new transaction type are currently failing. These have been
left as failing to indicate that they have not been confirmed to work.
Although an attempt could be made to guess the gateway's response for
the unit tests, the tests have been left with 'RESPONSE NEEDED' to
better and more clearly indicate the unknowns.

CE-203

Unit:
60 tests, 285 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
96.6667% passed

Remote:
37 tests, 133 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 
0 notifications
94.5946% passed

The remote tests `test_failed_referral_cft` and 
`test_successful_referral_cft` are currently failing with the error message
"Operation code \"34\" not supported for the given merchant" since 
the current gateway integration does not support referrals.

### DIFF
--- a/test/remote/gateways/remote_credorax_test.rb
+++ b/test/remote/gateways/remote_credorax_test.rb
@@ -285,6 +285,32 @@ class RemoteCredoraxTest < Test::Unit::TestCase
     assert_equal 'Referred to transaction has not been found.', response.message
   end
 
+  def test_successful_referral_cft
+    options = @options.merge(@normalized_3ds_2_options)
+    response = @gateway.purchase(@amount, @three_ds_card, options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+
+    cft_options = { referral_cft: true, email: 'john.smith@test.com' }
+    referral_cft = @gateway.refund(@amount, response.authorization, cft_options)
+    assert_success referral_cft
+    assert_equal 'Succeeded', referral_cft.message
+    # Confirm that the operation code was `referral_cft`
+    assert_equal '34', referral_cft.params['O']
+  end
+
+  def test_failed_referral_cft
+    options = @options.merge(@normalized_3ds_2_options)
+    response = @gateway.purchase(@amount, @three_ds_card, options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+
+    cft_options = { referral_cft: true, email: 'john.smith@test.com' }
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_credorax_test.rb
```diff
@@ -285,6 +285,30 @@ def test_failed_refund
...
+  def test_successful_referral_cft
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+
+    options = {referral_cft: true}
+    response = @gateway.refund(@amount, response.authorization, options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+    # Confirm that the transaction type was `referral_cft`
+    assert_equal '34', response.params['O']
```
leila-alderman: For our current Credorax gateway integration, these tests were passing but were using the `credit` transaction type instead of the `referral_cft` transaction type. I added this line to ensure that the transactions are going through as intended, but these would fail for us if the tests were actually working.

therufs:
I'm assuming you are well aware of the linter complaints and have made your decisions advisedly. :D

leila-alderman:
@therufs I've fixed the 4 RuboCop errors, but the Travis CI is still failing since I have failing unit tests. Is that alright to merge?

jasonxp:
@leila-alderman I think merging failing unit tests would be problematic, because then all subsequent Travis builds for any branch would fail with them in place. Would marking them pending and commenting out the implementation work?

leila-alderman:
@jasonxp Our customer was able to get Credorax to turn this setting on for us quickly. All of the unit and remote tests are now passing!

PR #3433, opened 2019-11-05
Ixopay: Add support for currency option to refund method

jasonxp:
For refunds, the currency in the ```options``` hash was not passed to the ```refund``` method. This resulted in the default currency being used. I discovered the problem while testing with the new gateway in core. I specified USD for the currency for the purchase, but the credit (refund) defaulted to EUR, which caused a mismatch error at the gateway.

When I posted in Slack, I was under the impression that core was not passing the currency from the transaction for refunds. I was mistaken: the problem was just that we weren't propagating it to the ```add_refund``` helper in Active Merchant.

In Slack, we also spoke briefly about whether the latter part of the ```options[:currency] || currency(money)``` idiom was needed. I took a look and decided to leave that in for now, since the ```currency``` method on the gateway base class results in the default currency, and we're following that everywhere else. Some other non-Spreedly users of the gateway adapters might be passing in money objects that support the ```currency``` method, so there's potentially some value in maintaining consistency there.

### DIFF
--- a/test/unit/gateways/ixopay_test.rb
+++ b/test/unit/gateways/ixopay_test.rb
@@ -24,12 +24,10 @@ class IxopayTest < Test::Unit::TestCase
   end
 
   def test_successful_purchase
-    @gateway.expects(:ssl_post).returns(successful_purchase_response)
-
     response = stub_comms do
       @gateway.purchase(@amount, @credit_card, @options)
     end.check_request do |endpoint, data, headers|
-      assert match(/<description>.+<\/description>/, data)
+      assert_match(/<description>.+<\/description>/, data)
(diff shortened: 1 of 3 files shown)


test/unit/gateways/ixopay_test.rb
```diff
@@ -24,12 +24,10 @@ def setup
   end
 
   def test_successful_purchase
-    @gateway.expects(:ssl_post).returns(successful_purchase_response)
```
jasonxp: This is cleanup that I spotted while working on the ```refund``` fix. We don't need to stub here since stubbing is handled by ```stub_comms```.

PR #3438, opened 2019-11-11
Adyen: Adds Unstore

deedeelavinder:
This adds an `unstore` method using the `disable` action. Adyen provides
a different API for `disable` and so this also adjusts the url to
use `/Recurring`, instead of `/Payment`, when `disable` is called. It
also adds separate versioning for each, as recommended by Adyen's support team.

In addition to `merchantAccount`, there are two additional fields
required for this action: `shopperReference` and
`recurringDetailReference`.

This also pulls in `davidsantoso`'s work from #3451 which adds the `storeToken` action. 

Unit tests:
54 tests, 254 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests:
75 tests, 248 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.3333% passed

(These two failing tests are unrelated to the proposed changes. They
both fail on master. They are the tests ending with
`threeds1_standalone` and `threeds2_standalone`.)

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -4,10 +4,10 @@ module ActiveMerchant #:nodoc:
 
       # we recommend setting up merchant-specific endpoints.
       # https://docs.adyen.com/developers/api-manual#apiendpoints
-      self.test_url = 'https://pal-test.adyen.com/pal/servlet/Payment/'
-      self.live_url = 'https://pal-live.adyen.com/pal/servlet/Payment/'
+      self.test_url = 'https://pal-test.adyen.com/pal/servlet/'
+      self.live_url = 'https://pal-live.adyen.com/pal/servlet/'
 
-      self.supported_countries = ['AT', 'AU', 'BE', 'BG', 'BR', 'CH', 'CY', 'CZ', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GB', 'GI', 'GR', 'HK', 'HU', 'IE', 'IS', 'IT', 'LI', 'LT', 'LU', 'LV', 'MC', 'MT', 'MX', 'NL', 'NO', 'PL', 'PT', 'RO', 'SE', 'SG', 'SK', 'SI', 'US']
+      self.supported_countries = %w(AT AU BE BG BR CH CY CZ DE DK EE ES FI FR GB GI GR HK HU IE IS IT LI LT LU LV MC MT MX NL NO PL PT RO SE SG SK SI US)
       self.default_currency = 'USD'
       self.currencies_without_fractions = %w(CVE DJF GNF IDR JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
       self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :jcb, :dankort, :maestro, :discover, :elo, :naranja, :cabal]
@@ -17,7 +17,8 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'https://www.adyen.com/'
       self.display_name = 'Adyen'
 
-      API_VERSION = 'v40'
+      PAYMENT_API_VERSION = 'v40'
+      RECURRING_API_VERSION = 'v30'
 
       STANDARD_ERROR_CODE_MAPPING = {
         '101' => STANDARD_ERROR_CODE[:incorrect_number],
@@ -110,6 +111,17 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def unstore(options={})
+        requires!(options, :shopper_reference, :recurring_detail_reference)
+        post = {}
+
+        add_shopper_reference(post, options)
+        add_merchant_account(post, options)
+        post[:recurringDetailReference] = options[:recurring_detail_reference]
+
+        commit('disable', post, options)
+      end
+
       def verify(credit_card, options={})
         MultiResponse.run(:use_first_response) do |r|
           r.process { authorize(0, credit_card, options) }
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -399,7 +409,12 @@ def parse(body)
...
       end
 
+      def endpoint(action)
+        action == 'disable' ? 'Recurring/' : 'Payment/'
+      end
+
       def commit(action, parameters, options)
+        url = url(action)
         begin
           raw_response = ssl_post("#{url}/#{action}", post_data(action, parameters), request_headers(options))
```
shasum: Nitpick: Since we now pass in `action` to the `url` method, could we have it return the complete URL including the specific action endpoint, such that the line here reduces to,
```ruby
raw_response = ssl_post(url(action), post_data(action, parameters), request_headers(options))
```

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -107,6 +107,16 @@ def store(credit_card, options={})
         end
       end
 
+      def unstore(options={})
+        requires!(options, :shopper_reference)
+        post = init_post(options)
+
+        post[:shopperReference] = options[:shopper_reference]
```
shasum: Does `shopperReference` allow to disable specific payment methods or would it disable all stored PMs associated with the shopper.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -428,13 +443,14 @@ def cvv_result_from(response)
         CVC_MAPPING[response['additionalData']['cvcResult'][0]] if response.dig('additionalData', 'cvcResult')
       end
 
-      def url
+      def url(action)
+        url_ending = "#{endpoint(action)}#{API_VERSION}"
```
tanyajajodia: Do we need to track different API versions now, each for the `Recurring/` and the `Payment/` endpoint? 

The `Recurring/` endpoint does not seem to have a `v40` that is currently defaulted to for `API_VERSION`.  Don't know how accurate this documentation that I'm referring to is: https://docs.adyen.com/api-explorer/#/Recurring/v30/overview

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -107,6 +107,16 @@ def store(credit_card, options={})
         end
       end
 
+      def unstore(options={})
+        requires!(options, :shopper_reference)
+        post = init_post(options)
+
+        post[:shopperReference] = options[:shopper_reference]
+        post[:merchantAccount] = options[:merchant_account] || @merchant_account
```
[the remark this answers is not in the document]
deedeelavinder: Good catch -- we don't want `init_post` here. 👍

lib/active_merchant/billing/gateways/adyen.rb
deedeelavinder: I like this -- thanks for the suggestion!

lib/active_merchant/billing/gateways/adyen.rb
deedeelavinder: It disables all stored PMs associated with the shopper.

lib/active_merchant/billing/gateways/adyen.rb
deedeelavinder: Thanks, Tanya. I had been looking at this page: https://docs.adyen.com/classic-integration/recurring-payments/disable-stored-details

I used that url in communication with Adyen support and sent them examples using `/Recurring/v40/disable`. 

I have sent another email to support asking explicitly about the url you shared and versions supported for `/Recurring`.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -4,13 +4,13 @@ class AdyenGateway < Gateway
...
-      self.live_url = 'https://pal-live.adyen.com/pal/servlet/Payment/'
+      self.test_url = 'https://pal-test.adyen.com/pal/servlet/'
+      self.live_url = 'https://pal-live.adyen.com/pal/servlet/'
 
-      self.supported_countries = ['AT', 'AU', 'BE', 'BG', 'BR', 'CH', 'CY', 'CZ', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GB', 'GI', 'GR', 'HK', 'HU', 'IE', 'IS', 'IT', 'LI', 'LT', 'LU', 'LV', 'MC', 'MT', 'MX', 'NL', 'NO', 'PL', 'PT', 'RO', 'SE', 'SG', 'SK', 'SI', 'US']
+      self.supported_countries = %w(AT AU BE BG BR CH CY CZ DE DK EE ES FI FR GB GI GR HK HU IE IS IT LI LT LU LV MC MT MX NL NO PL PT RO SE SG SK SI US)
       self.default_currency = 'USD'
       self.currencies_without_fractions = %w(CVE DJF GNF IDR JPY KMF KRW PYG RWF UGX VND VUV XAF XOF XPF)
-      self.supported_cardtypes = [:visa, :master, :american_express, :diners_club, :jcb, :dankort, :maestro, :discover, :elo, :naranja, :cabal]
+      self.supported_cardtypes = %w(:visa :master :american_express :diners_club :jcb :dankort :maestro :discover :elo :naranja :cabal)
```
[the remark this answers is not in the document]
deedeelavinder: In keeping with the patterns of AM, this is back to its original state.

lib/active_merchant/billing/gateways/adyen.rb
tanyajajodia: Ah, okay thanks. The example curls commands they show on the page you shared also use `v25` and `v46` for `Recurring/` and `Payments/`. I wonder if `Recurring/` still accepts `v40` as testing these against their test server has worked?

deedeelavinder: @tanyajajodia I heard back from Adyen:
> V40 is working on test because there is a "back-up" making sure that it is correctly sent to the proper version. I would therefore advice you to use v30 to be able to enjoy all our updates and features improvement on LIVE.

I will update the url-creation to account for this. Great catch!!

lib/active_merchant/billing/gateways/adyen.rb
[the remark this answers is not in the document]
kheang: Not a big deal to me, but I'm wondering if we should have methods for `add_shopper_reference` and `add_merchant_account` since we are setting those here, in `init_post`, and in `add_extra_data`.

lib/active_merchant/billing/gateways/adyen.rb
deedeelavinder: This is now updated to include the `recurringDetailReference` as it has been determined that it does indeed map, one-to-one, to a payment method.

deedeelavinder:
This code was merged with #3451 and deployed in commit f35cce634f9874aa29c709824c97b83e94ae94cb.

PR #3448, opened 2019-11-15
Adyen: Add gateway specific field for splits

leila-alderman:
Added the `splits` gateway specific field to the Adyen gateway.

For more details, see the Adyen documentation:
[Adyen API Reference](https://docs.adyen.com/api-explorer/#/PaymentSetupAndVerificationService/v50/payments__reqParam_splits)
[Adyen MarketPlace information about splits](https://docs.adyen.com/marketpay/processing-payments#providing-split-information)

CE-220

Unit:
51 tests, 243 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
73 tests, 234 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
97.2603% passed

The two failing remote tests are
 - `test_successful_purchase_with_auth_data_via_threeds1_standalone`
 - `test_successful_purchase_with_auth_data_via_threeds2_standalone`
These have been failing for a while and are unrelated.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * CyberSource: Send MDD on capture [leila-alderman] #3453
 * Ixopay: Include extra_data gateway specific field [therufs] #3450
 * CyberSource: Fix XML error on capture [leila-alderman] #3454
+* Adyen: Add gateway specific field for splits [leila-alderman] #3448
 
 == Version 1.102.0 (Nov 14, 2019)
 * Quickbooks: Make token refresh optional with allow_refresh flag [britth] #3419
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -57,6 +57,7 @@ module ActiveMerchant #:nodoc:
         add_installments(post, options) if options[:installments]
         add_3ds(post, options)
         add_3ds_authenticated_data(post, options)
+        add_splits(post, options)
         commit('authorise', post, options)
       end
 
@@ -64,6 +65,7 @@ module ActiveMerchant #:nodoc:
         post = init_post(options)
         add_invoice_for_modification(post, money, options)
         add_reference(post, authorization, options)
+        add_splits(post, options)
         commit('capture', post, options)
       end
 
@@ -71,6 +73,7 @@ module ActiveMerchant #:nodoc:
         post = init_post(options)
         add_invoice_for_modification(post, money, options)
         add_original_reference(post, authorization, options)
+        add_splits(post, options)
         commit('refund', post, options)
       end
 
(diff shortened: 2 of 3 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -363,6 +363,30 @@ def test_successful_purchase_with_risk_data
...
+        'value' => 50
+        },
+      'type' => 'Commission',
+      'reference' => 'THVjYXNCbGVkc29lCg'
+    }]
+
+    options = @options.merge({ splits: split_data })
+    response = @gateway.purchase(@amount, @credit_card, options)
+    assert_failure response
+    assert_match 'Invalid additionalData: MarketPlace account is not configured for this merchant', response.message
```
leila-alderman: The Spreedly test account is not set up to allow MarketPlace transactions, which is what splits are. 

From [Adyen's documentation](https://docs.adyen.com/api-explorer/#/PaymentSetupAndVerificationService/v50/payments__reqParam_splits):
> Splits: The details of how the payment should be split when distributing a payment to a MarketPay Marketplace and its Accounts.

Therefore, I wrote the remote test to fail when passing in splits with the response that it's not configured, which at least confirms that the field was successfully passed to the gateway. 

I'm open to suggestions on improving this, including just deleting this test altogether.

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -363,6 +363,30 @@ def test_successful_purchase_with_risk_data
     assert_equal '[capture-received]', response.message
   end
 
+  def test_successful_purchase_with_splits
+    split_data = [{
```
leila-alderman: I'm still feeling a bit concerned about this test and whether the format I'm using for this test is accurate, i.e., using strings instead of symbols.

therufs: One way to possibly do this is to pin the AM version in your Gemfile to your local repo, then go through "Testing production-like calls". Happy to work through this maybe this afternoon?

test/remote/gateways/remote_adyen_test.rb
therufs: 🤔 Would we need to set up a new test account to have this enabled?

leila-alderman: I'm not sure. Is it worth requesting that Adyen updates our account / adds a new account with this feature? The requested deadline for this from our customer is Friday.

therufs: In general, it seems like a successful test should ... well, succeed? So if we can't actually test that, I'm in favor of getting rid of it (and maybe creating a GH issue to fill in the test that someone someday might pick up.)

I think there may be a bigger conversation to have here about test fixtures -- it seems like it happens not infrequently that account-side modifications need to be made that might not be reflected in the setup of every AM developer ever. That is definitely beyond the scope of this add, though :D

leila-alderman: That makes sense. Remote tests are not ideal in many ways...

I've gone ahead and deleted this remote test for now. We definitely don't have complete test coverage for all of the GSFs in Active Merchant, so I don't think it's critical that we update our Adyen gateway integration just for this.

test/remote/gateways/remote_adyen_test.rb
leila-alderman: I've tested a local production-like purchase request through Core and Active Merchant, and the syntax seems to be working correctly! :tada:

PR #3449, opened 2019-11-15
DLocal: Updates for version 2.1

molbrown:
DLocal has updated its Auth/Capture flow, now accepting captures at a
new endpoint. It has also changed to use an authorization_id parameter
for referencing Authorize transactions (needed in capture and void
requests).

Remote:
26 tests, 69 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
100% passed

ECS-565

### DIFF
--- a/test/remote/gateways/remote_d_local_test.rb
+++ b/test/remote/gateways/remote_d_local_test.rb
@@ -194,15 +194,9 @@ class RemoteDLocalTest < Test::Unit::TestCase
   end
 
   def test_failed_void
-    auth = @gateway.authorize(@amount, @credit_card, @options)
-    assert_success auth
-
-    assert capture = @gateway.capture(@amount, auth.authorization, @options)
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_d_local_test.rb
```diff
@@ -194,15 +194,9 @@ def test_successful_void
   end
 
   def test_failed_void
-    auth = @gateway.authorize(@amount, @credit_card, @options)
-    assert_success auth
-
-    assert capture = @gateway.capture(@amount, auth.authorization, @options)
-    assert_success capture
-
-    response = @gateway.void(auth.authorization)
```
molbrown: This failure case is not longer true. In this updated flow, DLocal will allow a void of the authorization that has already been captured. They are aware of this issue, but I decided it was low risk.

PR #3450, opened 2019-11-15
Ixopay: Implement extraData GSF

therufs:
CE-144

Also filled out the `test_invalid_login` remote test. 

Remote: 23 tests, 95 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 22 tests, 108 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_ixopay_test.rb
+++ b/test/remote/gateways/remote_ixopay_test.rb
@@ -15,11 +15,30 @@ class RemoteIxopayTest < Test::Unit::TestCase
       description: 'Store Purchase',
       ip: '192.168.1.1',
     }
+
+    @extra_data = {extra_data: { customData1: 'some data', customData2: 'Can be anything really' }}
   end
 
   def test_successful_purchase
     response = @gateway.purchase(@amount, @credit_card, @options)
 
+    assert_success response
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -143,20 +213,24 @@ def test_successful_verify
...
+    assert_match %r{FINISHED}, response.message
+  end
+
   def test_failed_verify
     response = @gateway.verify(@declined_card, @options)
     assert_failure response
     assert_match %r{The transaction was declined}, response.message
   end
 
   def test_invalid_login
```
jasonxp: 👍

therufs:
@leila-alderman Thank you! I definitely cribbed it from somewhere. :D 

Unfortunately I can't easily change where the method call is, as the XML elements need to be ordered the way they currently are. (It is technically possible, but would involve re-parsing the XML in the add_extra_data helper method to actually put the extra data where it belongs, and where it belongs depends on what type of transaction it is, so that's a bonus complication.) But I did adjust the whitespace, and (at least locally) Rubocop seems to have been appeased.

PR #3451, opened 2019-11-15
Adyen: Allow the use of the storeToken endpoint for store

davidsantoso:
The store action currently uses the /authorise endpoint when executing
the store call. While Adyen does return the recurringDetailReference in
the authorize response, there are some drawbacks to using this endpoint
if you are looking for a pure tokenize call to store payment details at
Adyen with no additional side effects.

1. The /authorise endpoint will run a full authorization at the issuing
bank meaning the authorize request itself could fail preventing a full
tokenize at Adyen despite there being nothing inherently wrong with the
card itself.

2. It's possible for a response from the authorize call to respond with
a 3DS flow which would be suboptimal if you're not interested in doing
that at the particular time of storing the card.

This PR is meant to be a starting point for iteration on adding support
for the use of storeToken. As always there are always backwards
compatibility considerations and ensuring that both forms of
tokenizing at Adyen can work together seamlessly. My hope is that this
gives a good foundation with a real remote test for introspecting and
determining a best path forward.

As a side note: storeToken is only available if a specific role is added
to your user in Adyen so it's not universally available out of the box.

A couple notes about this particular implementation:

* Because this is a feature Adyen switches on a per account basis, it's
opt-in with the default being the current implementation then the added
ability to pass in a new option for using /storeToken instead.

* The url building had to be altered a bit since there's no version in
the URL for storeToken AND it's a different sub resource ("Recurring")

* I tried to ensure there was backwards compatibility will all other
actions by only appending new checks. But I think this needs an extra
close look.

* There are 2 failing remote tests around 3DS that I don't believe are
related to this in any way. But worth a second set of eyes.

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -4,8 +4,8 @@ module ActiveMerchant #:nodoc:
 
       # we recommend setting up merchant-specific endpoints.
       # https://docs.adyen.com/developers/api-manual#apiendpoints
-      self.test_url = 'https://pal-test.adyen.com/pal/servlet/Payment/'
-      self.live_url = 'https://pal-live.adyen.com/pal/servlet/Payment/'
+      self.test_url = 'https://pal-test.adyen.com/pal/servlet/'
+      self.live_url = 'https://pal-live.adyen.com/pal/servlet/'
 
       self.supported_countries = ['AT', 'AU', 'BE', 'BG', 'BR', 'CH', 'CY', 'CZ', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GB', 'GI', 'GR', 'HK', 'HU', 'IE', 'IS', 'IT', 'LI', 'LT', 'LU', 'LV', 'MC', 'MT', 'MX', 'NL', 'NO', 'PL', 'PT', 'RO', 'SE', 'SG', 'SK', 'SI', 'US']
       self.default_currency = 'USD'
@@ -98,7 +98,9 @@ module ActiveMerchant #:nodoc:
         add_recurring_contract(post, options)
         add_address(post, options)
 
-        initial_response = commit('authorise', post, options)
+        action = options[:tokenize_only] ? 'storeToken' : 'authorise'
+
+        initial_response = commit(action, post, options)
 
         if initial_response.success? && card_not_stored?(initial_response)
           unsupported_failure_response(initial_response)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -459,27 +465,35 @@ def success_from(action, response)
...
       def authorization_from(action, parameters, response)
         return nil if response['pspReference'].nil?
-        recurring = response['additionalData']['recurring.recurringDetailReference'] if response['additionalData']
+
+        if action == 'storeToken'
+          recurring = response['recurringDetailReference']
+        else
+          recurring = response['additionalData']['recurring.recurringDetailReference'] if response['additionalData']
+        end
+
```
deedeelavinder: The `if` on line 494 is triggering a Rubocop error because it is inside an `if/else` block. Variables with a ternary operator is one solution.

```
recurring = response['additionalData']['recurring.recurringDetailReference'] if response['additionalData']
action == 'storeToken' ? recurring = response['recurringDetailReference'] : recurring = recurring
```

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -401,7 +403,7 @@ def parse(body)
 
       def commit(action, parameters, options)
         begin
-          raw_response = ssl_post("#{url}/#{action}", post_data(action, parameters), request_headers(options))
+          raw_response = ssl_post("#{base_url(action)}/#{action}", post_data(action, parameters), request_headers(options))
```
shasum: Would it improve readability to have `base_url` return the complete URL including the action endpoint?

```ruby
raw_response = ssl_post(base_url(action), post_data(action, parameters), request_headers(options))
```

lib/active_merchant/billing/gateways/adyen.rb
shasum: I like DeeDee's suggestion of reversing the order (maybe a simple conditional would suffice then).

```ruby
recurring = response['additionalData']['recurring.recurringDetailReference'] if response['additionalData']
recurring = response['recurringDetailReference'] if action == 'storeToken'
```

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -607,6 +607,14 @@ def test_successful_store
     assert_equal 'Authorised', response.message
   end
 
+  def test_successful_tokenize_only_store
```
shasum: It's great to have a validating remote test!

In keeping with Active Merchant guidelines, could we also add an equivalent unit test. For folks who may not have sandbox credentials, the canned response recorded in unit tests could serve as a useful reference at some point.

PR #3455, opened 2019-11-22
Borgun: Update card for remote tests

molbrown:
Takes failures from 9 down to 1, failing test follows expected behavior
described in the comment.

Remote:
20 tests, 45 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95% passed

### DIFF
--- a/test/remote/gateways/remote_borgun_test.rb
+++ b/test/remote/gateways/remote_borgun_test.rb
@@ -8,7 +8,7 @@ class RemoteBorgunTest < Test::Unit::TestCase
     @gateway = BorgunGateway.new(fixtures(:borgun))
 
     @amount = 100
-    @credit_card = credit_card('5587402000012011', year: 2018, month: 9, verification_value: 415)
+    @credit_card = credit_card('5587402000012011', year: 2022, month: 9, verification_value: 415)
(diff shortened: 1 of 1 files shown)


test/remote/gateways/remote_borgun_test.rb
```diff
@@ -8,7 +8,7 @@ def setup
     @gateway = BorgunGateway.new(fixtures(:borgun))
 
     @amount = 100
-    @credit_card = credit_card('5587402000012011', year: 2018, month: 9, verification_value: 415)
+    @credit_card = credit_card('5587402000012011', year: 2022, month: 9, verification_value: 415)
```
jeremywrowe: Would `Time.now.year + 2` work? That we don't have to update the year every couple of years?

molbrown: Tried that as well as `+ 1`, and kept getting an invalid card for some reason. Decided it wasn't worth the time to figure out why... _bought us a few years_

jeremywrowe: :/ That's unfortunate. Thanks for trying 😄

PR #3458, opened 2019-11-26
Credorax: Don't pass 3ds_version in initial 3ds request

britth:
Credorax 3DS requests are erroring with the message "At least 
one of input parameters is malformed. Parameter 3ds_version is
invalid.". We've been informed that you should not pass
`3ds_version` as part of the original transaction request, and
that the correct version will be returned in your response. You
only need to send it when a 92 needs to be triggered (three ds
completion) or when you're sending an i8 param in your original
request. This PR removes 3ds_version from the original request
where i8 is not passed.

Some updates to tests were also required. It appears that this
behavior is somewhat new, as the 3DS2 tests now fail without this
change. Additionally, the test card doesn't proceed to direct 
success as seemed to have been the case before based on the 
existing test; instead it returns a message that authentication is 
required.

Unit:
60 tests, 289 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
38 tests, 141 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -305,7 +305,6 @@ module ActiveMerchant #:nodoc:
           options.dig(:stored_credential, :initiator) == 'merchant' ? post[:'3ds_channel'] = '03' : post[:'3ds_channel'] = '02'
           post[:'3ds_redirect_url'] = three_ds_2_options[:notification_url]
           post[:'3ds_challengewindowsize'] = options[:three_ds_challenge_window_size] || '03'
-          post[:'3ds_version'] = options[:three_ds_version] if options[:three_ds_version]
           post[:d5] = browser_info[:user_agent]
           post[:'3ds_transtype'] = options[:transaction_type] || '01'
           post[:'3ds_browsertz'] = browser_info[:timezone]
@@ -330,6 +329,7 @@ module ActiveMerchant #:nodoc:
 
       def add_3d_secure_1_data(post, options)
         post[:i8] = build_i8(options[:eci], options[:cavv], options[:xid])
+        post[:'3ds_version'] = options[:three_ds_version].nil? || options[:three_ds_version] == '1' ? '1.0' : options[:three_ds_version]
       end
 
       def add_normalized_3d_secure_2_data(post, options)
@@ -339,7 +339,7 @@ module ActiveMerchant #:nodoc:
           three_d_secure_options[:eci],
           three_d_secure_options[:cavv]
         )
-        post[:'3ds_version'] = three_d_secure_options[:version]
+        post[:'3ds_version'] = three_d_secure_options[:version] == '2' ? '2.0' : three_d_secure_options[:version]
         post[:'3ds_dstrxid'] = three_d_secure_options[:ds_transaction_id]
       end
 
--- a/test/remote/gateways/remote_credorax_test.rb
+++ b/test/remote/gateways/remote_credorax_test.rb
@@ -8,7 +8,7 @@ class RemoteCredoraxTest < Test::Unit::TestCase
     @credit_card = credit_card('4176661000001015', verification_value: '281', month: '12', year: '2022')
     @fully_auth_card = credit_card('5223450000000007', brand: 'mastercard', verification_value: '090', month: '12', year: '2025')
     @declined_card = credit_card('4176661000001111', verification_value: '681', month: '12', year: '2022')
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -330,6 +329,10 @@ def add_3d_secure(post, options)
 
       def add_3d_secure_1_data(post, options)
         post[:i8] = build_i8(options[:eci], options[:cavv], options[:xid])
+        if options[:three_ds_version]
+          three_ds_version = options[:three_ds_version] == '1' ? '1.0' : options[:three_ds_version]
+          post[:'3ds_version'] = three_ds_version
+        end
```
britth: This behavior was a bit unexpected - based on what we heard from the gateway, it seemed like any request where i8 was passed would need to pass version as well. However, 3ds1 transactions seem to work with or without it. Added this logic to include it when present, but open to other thoughts.

test/remote/gateways/remote_credorax_test.rb
```diff
@@ -79,11 +79,31 @@ def test_successful_purchase_with_auth_data_via_3ds1_fields
...
     options = @options.merge(@normalized_3ds_2_options)
     response = @gateway.purchase(@amount, @three_ds_card, options)
-    assert_success response
-    assert_equal 'Succeeded', response.message
+    assert response.test?
+    refute response.authorization.blank?
+
+    refute response.params['3DS_METHOD'].blank?
+    assert_equal '2.1.0', response.params['3DS_VERSION']
+    assert_equal 'Transaction pending cardholder authentication.', response.message
```
britth: not sure why this worked as a direct auth before with this card...the behavior now observed seems logical (auth required for 3ds enabled card), but I'm unsure why it changed.

test/remote/gateways/remote_credorax_test.rb
```diff
@@ -286,8 +306,7 @@ def test_failed_refund
   end
 
   def test_successful_referral_cft
-    options = @options.merge(@normalized_3ds_2_options)
-    response = @gateway.purchase(@amount, @three_ds_card, options)
+    response = @gateway.purchase(@amount, @credit_card, @options)
```
britth: this did not seem dependent on 3ds, and using the 3ds card doesn't result in a completed transaction now, so updated these to use the regular credit card. if my assumption is wrong, definitely let me know! cc @leila-alderman

leila-alderman: @britth I haven't actually tested this transaction type without sending 3DS. I went with the 3DS version based on Credorax's documentation for `referral_cft` transactions:

> For transactions made with a Mastercard payment card, the original transaction to which the CFT refers must include 3D secure data in order for the referral CFT transaction to be processed successfully.

I'm puzzled as to why these transactions would no longer be successful. 🤔

test/remote/gateways/remote_credorax_test.rb
leila-alderman: I'm not sure if this is helpful or not, but in checking Credorax's documentation online, it shows that they just updated their API docs on November 26: https://epower.credorax.com/

Sadly, I don't see a changelog anywhere.

lib/active_merchant/billing/gateways/credorax.rb
molbrown: Based on what the gateway says, do you think we should send a default value if `i8` is included and no `options[:three_ds_version]` is specified?

test/remote/gateways/remote_credorax_test.rb
britth: interesting, thanks for that info! yeah, it's very weird - we can get to the point where 3ds is triggered and awaiting authorization, but it's a bit tricky to try to simulate an actual 3ds auth; but it's very interesting that the 3ds card was working frictionlessly before. i'll check with credorax to see if there's any updated test card data 👍

test/remote/gateways/remote_credorax_test.rb
britth: good find, thank you!!

lib/active_merchant/billing/gateways/credorax.rb
britth: yeah, i think you're right - even though it seems to work in the test env the new docs explicitly say it's mandatory with i8. will update, thank you!

test/remote/gateways/remote_credorax_test.rb
britth: aha, the test cards did change - updated the card number and changed these tests back to their original state

PR #3461, opened 2019-11-27
Ingenico Global Collect: support airline_data GSFs

therufs:
CE-252

Unit:
21 tests, 96 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
18 tests, 42 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -100,6 +100,35 @@ module ActiveMerchant #:nodoc:
         post['order']['references']['invoiceData'] = {
           'invoiceNumber' => options[:invoice]
         }
+        add_airline_data(post, options) if options[:airline_data]
+      end
+
+      def add_airline_data(post, options)
+        airline_data = {}
+
+        flight_date = options[:airline_data][:flight_date]
+        passenger_name = options[:airline_data][:passenger_name]
+        code = options[:airline_data][:code]
+        name = options[:airline_data][:name]
+
+        airline_data['flightDate'] = flight_date if flight_date
+        airline_data['passengerName'] = passenger_name if passenger_name
+        airline_data['code'] = code if code
+        airline_data['name'] = name if name
+
+        flight_legs = []
+        options[:airline_data][:flight_legs]&.each do |fl|
+          leg = {}
+          leg['arrivalAirport'] = fl[:arrival_airport] if fl[:arrival_airport]
+          leg['originAirport'] = fl[:origin_airport] if fl[:origin_airport]
+          leg['date'] = fl[:date] if fl[:date]
+          leg['number'] = fl[:number] if fl[:number]
+          leg['carrierCode'] = fl[:carrier_code] if fl[:carrier_code]
+          leg['airlineClass'] = fl[:carrier_code] if fl[:airline_class]
+          flight_legs << leg
+        end
+        airline_data['flightLegs'] = flight_legs
+        post['order']['additionalInput']['airlineData'] = airline_data
       end
 
       def add_creator_info(post, options)
--- a/test/remote/gateways/remote_global_collect_test.rb
+++ b/test/remote/gateways/remote_global_collect_test.rb
@@ -55,6 +55,51 @@ class RemoteGlobalCollectTest < Test::Unit::TestCase
     assert_equal 'Succeeded', response.message
   end
 
+  def test_successful_purchase_with_airline_data
+    options = @options.merge(
+      airline_data: {
+        code: 111,
+        name: 'Spreedly Airlines',
+        flight_date: '20190810',
+        passenger_name: 'Randi Smith',
+        flight_legs: [
+          { arrival_airport: 'BDL',
+            origin_airport: 'RDU',
+            date: '20190810',
+            carrier_code: 'SA',
+            number: 596,
+            airline_class: 'ZZ'},
+          { arrival_airport: 'RDU',
+            origin_airport: 'BDL',
+            date: '20190817',
+            carrier_code: 'SA',
+            number: 597,
+            airline_class: 'ZZ'}
+        ]
+      }
+    )
+
+    response = @gateway.purchase(@amount, @credit_card, options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+  end
+
+  def test_failed_purchase_with_insufficient_airline_data
+    options = @options.merge(
+      airline_data: {
+        flight_date: '20190810',
+        passenger_name: 'Randi Smith'
+      }
+    )
+
+    response = @gateway.purchase(@amount, @credit_card, options)
+    assert_failure response
+    assert_equal 'PARAMETER_NOT_FOUND_IN_REQUEST', response.message
+    property_names = response.params['errors'].collect { |e| e['propertyName'] }
+    assert property_names.include? 'order.additionalInput.airlineData.code'
+    assert property_names.include? 'order.additionalInput.airlineData.name'
+  end
+
   def test_successful_purchase_with_very_long_name
     credit_card = credit_card('4567350000427977', { first_name: 'thisisaverylongfirstname'})
 
--- a/test/unit/gateways/global_collect_test.rb
+++ b/test/unit/gateways/global_collect_test.rb
@@ -35,6 +35,38 @@ class GlobalCollectTest < Test::Unit::TestCase
     assert_success capture
   end
 
+  def test_successful_purchase_airline_fields
+    options = @options.merge(
+      airline_data: {
+        code: 111,
+        name: 'Spreedly Airlines',
+        flight_date: '20190810',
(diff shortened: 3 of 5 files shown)


test/unit/gateways/global_collect_test.rb
```diff
@@ -35,6 +35,38 @@ def test_successful_authorize_and_capture
     assert_success capture
   end
 
+  def test_successful_authorize_and_capture_airline_fields
```
leila-alderman: Minor suggestion:
Consider renaming this test to `test_successful_purchase_with_airline_fields` since it uses the `purchase` call. (I realize that purchase is equivalent to authorize+capture on this gateway, but I found the naming to be initially confusing.)

test/unit/gateways/global_collect_test.rb
```diff
@@ -362,6 +394,10 @@ def successful_authorize_response
...
   end
 
+  def successful_authorize_response_airline_fields
```
leila-alderman: These two new responses appear to be very similar to the existing successful authorize and capture responses. The only difference that I'm seeing is the fraud fields. 

Is it necessary to add these two new responses? Does it make sense to instead update the two existing responses to include the (new?) fraud fields?

test/unit/gateways/global_collect_test.rb
```diff
@@ -35,6 +35,38 @@ def test_successful_authorize_and_capture
...
+            airline_class: 'ZZ'}
+        ]
+      }
+    )
+    stub_comms do
+      @gateway.purchase(@accepted_amount, @credit_card, options)
+    end.check_request do |endpoint, data, headers|
+      assert_equal 111, JSON.parse(data)['order']['additionalInput']['airlineData']['code']
+      assert_equal '20190810', JSON.parse(data)['order']['additionalInput']['airlineData']['flightDate']
+      assert_equal 2, JSON.parse(data)['order']['additionalInput']['airlineData']['flightLegs'].length
```
leila-alderman: These are some really nice `assert` statements! I don't think I've seen this structure before, but I quite like it. 🎉

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -100,6 +100,29 @@ def add_order(post, money, options, capture: false)
...
+      end
+
+      def add_airline_data(post, options)
+        airline_data = {}
+        airline_data['flightDate'] = options[:airline_data][:flight_date]
+        airline_data['passengerName'] = options[:airline_data][:passenger_name]
+        airline_data['code'] = options[:airline_data][:code]
+        airline_data['name'] = options[:airline_data][:name]
+
+        items = []
```
leila-alderman: Minor suggestion:
Consider renaming this variable to something like `flight_legs` to be more descriptive.

molbrown: What happens if only some of these `airline_data` fields are included? Do we send fields with `nil` values and can the gateway handle that?

test/remote/gateways/remote_global_collect_test.rb
```diff
@@ -55,6 +55,35 @@ def test_successful_purchase_with_more_options
...
+            number: 597,
+            airline_class: 'ZZ'}
+        ]
+      }
+    )
+
+    response = @gateway.purchase(@amount, @credit_card, options)
+    assert_success response
+    assert_equal 'Succeeded', response.message
+  end
```
molbrown: Related to my question above, consider adding a test where some of this data is missing, and see that everything is still handled as expected by the adapter as well as the gateway sandbox.

lib/active_merchant/billing/gateways/global_collect.rb
leila-alderman: That's a good point. It does look like the other methods use the format
```
post['parameter'] = options[:parameter] if options[:parameter]
```

It would probably be good to follow that format here as well.

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -100,6 +100,29 @@ def add_order(post, money, options, capture: false)
...
+        options[:airline_data][:flight_legs].each do |item|
+          leg = {}
+          leg['arrivalAirport'] = item[:arrival_airport]
+          leg['originAirport'] = item[:origin_airport]
+          leg['date'] = item[:date]
+          leg['number'] = item[:number]
+          leg['carrierCode'] = item[:carrier_code]
+          leg['airlineClass'] = item[:airline_class]
```
molbrown: Same question here.

test/unit/gateways/global_collect_test.rb
therufs: THANKS I cribbed it from Adyen https://github.com/activemerchant/active_merchant/pull/3460/files#diff-f2e3fbe51e77715bca1b237e7cb5e5bbR550 :D 
It seemed like a good fit for "make sure byzantine JSON is behaving itself".

lib/active_merchant/billing/gateways/global_collect.rb
therufs: This will work for some fields and I'll figure out which ones. The API Formerly Known As Global Collect has fairly picky requirements about what fields must or conditionally must be included and how they're validated (see https://epayments-api.developer-ingenico.com/s2sapi/v1/en_US/ruby/payments/create.html?paymentPlatform=GLOBALCOLLECT#payments-create-payload Order > additionalInput > airlineData, which will go in the docs) so I can't do this unilaterally.  I also didn't feel like we could improve on the error messages the gateway generates, so I decided to just pass theirs along; I'll write a test to demonstrate this.

therufs: Actually it looks like the errors are the same whether we pass the missing field without a value or don't pass it at all, so maybe I can just do it unilaterally 🤔

test/unit/gateways/global_collect_test.rb
therufs: The main difference I saw was the inclusion of `statusCategory` but it looks to me like that's now included in standard purchases too, so I'll collapse them.

lib/active_merchant/billing/gateways/global_collect.rb
therufs: Okay, I have given this a whack!

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -100,6 +100,35 @@ def add_order(post, money, options, capture: false)
...
+          leg['arrivalAirport'] = fl[:arrival_airport] if fl[:arrival_airport]
+          leg['originAirport'] = fl[:origin_airport] if fl[:origin_airport]
+          leg['date'] = fl[:date] if fl[:date]
+          leg['number'] = fl[:number] if fl[:number]
+          leg['carrierCode'] = fl[:carrier_code] if fl[:carrier_code]
+          leg['airlineClass'] = fl[:carrier_code] if fl[:airline_class]
+          flight_legs << leg
+        end
+        airline_data['flightLegs'] = flight_legs
+        post['order']['additionalInput']['airlineData'] = airline_data
```
leila-alderman: Because all this method is really doing is converting a Ruby hash into JSON, I think that the whole method can be replaced with the following:

```
def add_airline_data(post, options)
  airline_data = options[:airline_data].deep_stringify_keys.deep_transform_keys { |key| key.camelize(:lower) }
  post['order']['additionalInput']['airlineData'] = airline_data
end
```

I've tested this with the unit and remote tests, and they all still pass. What do you think?

therufs: I thought about trying something like this but feel anxious about allowing users to send arbitrary data to an endpoint via our servers. 😬

therufs: Which I realize we're kind of doing anyway, but it still makes me 😬

leila-alderman: That's a good point. Do you know if there's a way to apply strong parameters to something like this to apply whitelisting?

therufs: 🤔 I think of `strong params` as being a Rails (or maybe general DB-backed app) thing that controllers handle when creating DB records, which isn't what we're doing here. So on the one hand, I don't know of a way. On the other hand, hopefully this _is_ being validated on the DB side, so maybe my concerns are overblown. (But on the third hand, I'd just as soon not find out?) 

In the future case, I continue to muse intermittently on the possibility of some generalizable solution for GSFs, which means this is a problem we may still want to solve eventually.

leila-alderman: Yeah, it's something that I've been mulling over for a while as I keep thinking that there has to be a better way... Thanks for sharing your thoughts! :grin:

PR #3463, opened 2019-12-05
Ebanx: Update capture method

chinhle23:
ECS-872/ECS-929
 
https://developers.ebanx.com/api-reference/ebanx-payment-api/payment-reference/reference-capture-operation/
Per Ebanx documentation: “You can make only one partial capture per authorized
payment and this feature is only available in Brazil. If you want to enable
this feature, please contact our Integration Team.”
 
 1. Ebanx will capture full amount authorized if no amount is included in
    the capture request. Adds a flag for including the amount only in the
    case of partial capture.
 
 2. Update tests for new error messages and handling of partial capture
 
Unit:
100% passed
 
Remote:
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -73,7 +73,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_integration_key(post)
         post[:hash] = authorization
-        post[:amount] = amount(money)
+        post[:amount] = amount(money) if options[:include_capture_amount].to_s == 'true'
 
         commit(:capture, post)
       end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,7 @@
 * Ingenico GlobalCollect: support `airline_data` and related GSFs [therufs] #3461
 * Add UnionPay card type [leila-alderman] #3464
 * Braintree: Fix add_credit_card_to_customer in Store [molbrown] #3466
+* EBANX: Default to not send amount on capture [chinhle23] #3463
 
 == Version 1.103.0 (Dec 2, 2019)
 * Quickbooks: Mark transactions that returned `AuthorizationFailed` as failures [britth] #3447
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -73,7 +73,7 @@ def capture(money, authorization, options={})
         post = {}
         add_integration_key(post)
         post[:hash] = authorization
-        post[:amount] = amount(money)
+        post[:amount] = amount(money) if customer_country(options) == 'br'
```
chinhle23: The drawback is if the merchant never intends to transact partial captures in Brazil, the transaction will always fail...thus requiring the merchant to contact the Ebanx Integration Team to enable partial captures.

britth: I wonder if rather than checking country and always adding in that case, it makes sense to add a flag so that the merchant can specify whether or not they want to do a partial capture to avoid instances where it may fail unexpectedly? looks like we do something similar for bluesnap: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/blue_snap.rb#L98

then we could just document the circumstances where partial captures work. what do you think?

chinhle23: @britth That sounds like a much better approach. I'll look into the bluesnap implementation. Thanks!

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -73,7 +73,7 @@ def capture(money, authorization, options={})
         post = {}
         add_integration_key(post)
         post[:hash] = authorization
-        post[:amount] = amount(money)
+        post[:amount] = amount(money) if options[:include_capture_amount] == true
```
curiousepic: We probably want `.to_s == 'true'` here. If this is sent as a string `"false"`, it's still going to trigger, but `to_s` is still true for a bool.

chinhle23: That makes sense. Thanks for the review @curiousepic , the changes have been pushed.

PR #3471, opened 2019-12-17
Adyen: add recurring_contract_type to auth

therufs:
Unit: 
57 tests, 267 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
81 tests, 271 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.5309% passed

Failures on 3DS (1 and 2) purchase_with_auth_data attempts.

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -51,7 +51,7 @@ module ActiveMerchant #:nodoc:
         requires!(options, :order_id)
         post = init_post(options)
         add_invoice(post, money, options)
-        add_payment(post, payment)
+        add_payment(post, payment, options)
         add_extra_data(post, payment, options)
         add_stored_credentials(post, payment, options)
         add_address(post, options)
@@ -59,6 +59,7 @@ module ActiveMerchant #:nodoc:
         add_3ds(post, options)
         add_3ds_authenticated_data(post, options)
         add_splits(post, options)
+        add_recurring_contract(post, options)
         commit('authorise', post, options)
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -317,7 +318,6 @@ def add_payment(post, payment)
         if payment.is_a?(String)
           _, _, recurring_detail_reference = payment.split('#')
           post[:selectedRecurringDetailReference] = recurring_detail_reference
-          add_recurring_contract(post, options)
```
leila-alderman: Why is this not needed anymore? I got the impression that we needed to add this method to `authorize`, but I'm not sure why it's been removed here.

therufs: `add_payment` only gets called in `authorize` and `store`; since `add_recurring_contract` is now included directly in those methods, we didn't need it here (and it seemed to me like `add_recurring_contract` made sense being its own entity, rather than part of `add_payment`.)

therufs: But I probably need to handle it better, one moment

therufs: Okay, two moments 😅

therufs:
Final test output: 
Unit:
58 tests, 269 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 
81 tests, 271 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.5309% passed (failures: test_successful_purchase_with_auth_data_via_threeds1_standalone, test_successful_purchase_with_auth_data_via_threeds2_standalone)

PR #3473, opened 2019-12-18
Cardstream: Localize zero-decimal currencies

britth:
Zero-decimal currencies were being handled in an unexpected way on
Cardstream compared to other gateways. For instance, whereas we
expect a merchant to pass in an amount like 100 to be charged 1
JPY, this gateway processes this as though it's a charge of 100 JPY. 
This PR uses localized_amount to divide the value by 100 before 
passing along to the gateway. The gateway does not support any 
three-decimal currencies, so this PR does not try to test those 
scenarios.

This gateway also had quite a few failing tests due to cc year,
updated error messages and some duplicate transaction handling.
These are also updated in this PR. The tests are still somewhat
temperamental in that you'll sometimes get a timeout, but that
does not appear to be related to these updates as it's a different
test or two each time (and sometimes none at all).

Remote:
31 tests, 204 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
26 tests, 132 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_card_stream_test.rb
+++ b/test/remote/gateways/remote_card_stream_test.rb
@@ -8,35 +8,35 @@ class RemoteCardStreamTest < Test::Unit::TestCase
 
     @amex = credit_card('374245455400001',
       :month => '12',
-      :year => '2014',
+      :year => Time.now.year + 1,
       :verification_value => '4887',
       :brand => :american_express
     )
 
     @mastercard = credit_card('5301250070000191',
       :month => '12',
-      :year => '2014',
+      :year => Time.now.year + 1,
       :verification_value => '419',
       :brand => :master
     )
 
     @visacreditcard = credit_card('4929421234600821',
       :month => '12',
-      :year => '2014',
+      :year => Time.now.year + 1,
       :verification_value => '356',
       :brand => :visa
     )
 
     @visadebitcard = credit_card('4539791001730106',
       :month => '12',
-      :year => '2014',
+      :year => Time.now.year + 1,
       :verification_value => '289',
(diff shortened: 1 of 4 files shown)


britth:
@molbrown good question! I just emailed their support, and they said they can process all zero and two decimal currencies. The only one our test account supports is JPY though. Testing with that, I noticed that the response returns a currencyExponent field that corresponds to the decimal type (so JPY is 0, GBP is 2, etc...). Do you know off-hand if the active merchant default contains all zero decimal currencies?

molbrown:
This is the default, there are some cases where a gateway differs from this. For example when I made a currency change for Adyen, I found that it considers the CLP a fractional currency, even though it is recognized as a non-decimal in other places.- https://github.com/activemerchant/active_merchant/blob/341b1bafd8abb1452ca9018d180805824dea1223/lib/active_merchant/billing/gateway.rb#L126

molbrown:
Currencies apparently have a variety of interpretations 🌈

britth:
got confirmation from gateway of the exact set of zero-decimal currencies supported and updated PR 👍

test/remote/gateways/remote_card_stream_test.rb
```diff
@@ -8,35 +8,35 @@ def setup
 
     @amex = credit_card('374245455400001',
       :month => '12',
-      :year => '2014',
+      :year => Time.now.year,
```
molbrown: Small suggestion, consider changing these to `Time.now.year + 1`?

PR #3480, opened 2020-01-02
Decidir: Add debit card payment method IDs

leila-alderman:
Adds the correct `payment_method_id` value for the four debit cards that
Decidir supports. These are selected by passing the new `debit` GSF.

CE-308

Unit:
32 tests, 136 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
21 tests, 70 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
95.2381% passed

The `test_successful_purchase_with_amex` remote test is failing with a
gateway timeout error.

### DIFF
--- a/test/unit/gateways/decidir_test.rb
+++ b/test/unit/gateways/decidir_test.rb
@@ -278,6 +278,53 @@ class DecidirTest < Test::Unit::TestCase
     assert_equal 24, post[:payment_method_id]
   end
 
+  def test_payment_method_id_with_visa_debit
+    visa_debit_card = credit_card('4517721004856075')
+    debit_options = @options.merge(debit: true)
+
+    stub_comms(@gateway_for_purchase, :ssl_request) do
+      @gateway_for_purchase.purchase(@amount, visa_debit_card, debit_options)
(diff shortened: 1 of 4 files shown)


test/unit/gateways/decidir_test.rb
```diff
@@ -278,6 +278,53 @@ def test_payment_method_id_with_naranja
     assert_equal 24, post[:payment_method_id]
   end
 
+  def test_payment_method_id_with_visa_debit
+    visa_debit_card = credit_card('4517721004856075')
+    debit_options = @options.merge(payment_method_id: 31)
```
therufs: I wonder if we should we write a separate test for passing a visa with `debit: true` for completeness' sake, but do feel like the code path is adequately tested without it. Up to you!

leila-alderman: Shoot! That was definitely supposed to be `debit: true`! 

While I was working on adding this, I discovered that the existing GSF for `payment_method_id` didn't work, so I was troubleshooting that and forgot to change it back. Thanks for pointing that out!

PR #3494, opened 2020-01-10
CardConnect: Remove domain validation

leila-alderman:
CardConnect no longer requires ports to be included in their valid
domains, so this change removes that requirement, allowing customers to
use CardConnect domains without specifying a port.

[CE-209](https://spreedly.atlassian.net/browse/CE-209)

Unit:
22 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
25 tests, 58 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
92% passed

Two failing remote tests:
 - `test_successful_refund`
 - `test_partial_refund`
Both of these tests are failing with the message "Txn not settled" and
are failing on master as well. The last commit to CardConnect (Oct 2019)
shows fully passing remote tests, so something appears to have changed
in the gateway integration since then. These new failures seem to be
unrelated to the changes made in this commit.

All unit tests:
4414 tests, 71334 assertions, 0 failures, 0 errors, 0 pendings, 2
omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/card_connect.rb
+++ b/lib/active_merchant/billing/gateways/card_connect.rb
@@ -63,14 +63,7 @@ module ActiveMerchant #:nodoc:
 
       def initialize(options = {})
         requires!(options, :merchant_id, :username, :password)
-        require_valid_domain!(options, :domain)
         super
-      end
-
-      def require_valid_domain!(options, param)
-        if options[param]
-          raise ArgumentError.new('not a valid cardconnect domain') unless /\Dcardconnect.com:\d{1,}\D/ =~ options[param]
-        end
       end
 
       def purchase(money, payment, options = {})
--- a/test/unit/gateways/card_connect_test.rb
+++ b/test/unit/gateways/card_connect_test.rb
@@ -16,10 +16,8 @@ class CardConnectTest < Test::Unit::TestCase
     }
   end
 
-  def test_incorrect_domain
-    assert_raise(ArgumentError) {
-      CardConnectGateway.new(username: 'username', password: 'password', merchant_id: 'merchand_id', domain: 'www.google.com')
-    }
+  def test_allow_domains_without_ports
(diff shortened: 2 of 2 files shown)


lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -63,16 +63,9 @@ class CardConnectGateway < Gateway
 
       def initialize(options = {})
         requires!(options, :merchant_id, :username, :password)
-        require_valid_domain!(options, :domain)
         super
       end
 
-      def require_valid_domain!(options, param)
```
therufs: This looks to me like it both checks for a port and enforces that the URL contains `cardconnect.com`. It looks to me like the card only wants us to not check for the inclusion of a port anymore. Are we sure we want to get rid of both of these checks?

leila-alderman: Interesting question! My assumption had been that the method was built around requiring a domain, but looking at things more closely, I think that you're right that it's worth continuing to enforce the `cardconnect.com` bit. I've fixed that up now.

therufs: It looks like this check was specifically included on implementation so I'm in favor of keeping it until specifically told to do otherwise: https://github.com/activemerchant/active_merchant/pull/2706#discussion_r160954583

leila-alderman: Oooh, nice digging! Thanks for uncovering that! I went ahead and pushed some changes so that this just removes the port portion.

redbeardaz:
Can we also add a check to be sure that the `https` part is valid?

Example: https://lava.spreedly.com/search?_utf8=%E2%9C%93&query=1BN9Myxkw4KrvTUkyNtfnVzs3s9

https://spreedly.slack.com/archives/CH7B6NWA0/p1578952942056600

leila-alderman:
@redbeardaz @therufs I've updated this PR to include a requirement for `https://`.

PR #3495, opened 2020-01-13
Remove customer hash when using a payment_method_nonce

jeremywrowe:
When processing 3DS based requests there is a need to create a transaction
using a payment method nonce that is generated on the client. To facilitate that
need and keep existing behaviors with payment method nonces, a check for a
string based payment method nonce was added right before finalizing sale
parameters. If the payment method nonce is a string, the customer hash will be
removed since there should already be a customer in Braintree's system.

For further information regarding the need to use a client payment method nonce,
please check here: https://developers.braintreepayments.com/guides/3d-secure/server-side/ruby

Unit:

77 tests, 178 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:

80 tests, 435 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -628,6 +628,11 @@ module ActiveMerchant #:nodoc:
 
         parameters[:line_items] = options[:line_items] if options[:line_items]
 
+        if options[:payment_method_nonce].is_a?(String)
+          parameters.delete(:customer)
+          parameters[:payment_method_nonce] = options[:payment_method_nonce]
+        end
+
         parameters
       end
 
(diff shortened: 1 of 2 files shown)


jeremywrowe:
@molbrown - There is not a good way to test this with remote tests, that's why I had opted for unit tests. The customer hash is only removed when the payment method nonce option is a string. I think I'll expand upon this test (https://github.com/activemerchant/active_merchant/blob/7d92f9923f77b784425d97eb43fddc849b0aa89c/test/unit/gateways/braintree_blue_test.rb#L57-L65) to show that the customer hash is intact.

jeremywrowe:
@molbrown Thanks for the review, I've updated based on the rubocop suggestions on Travis.

molbrown:
👍

britth:
@jeremywrowe should we add a changelog entry for this commit?

jeremywrowe:
@britth yes, I missed it sorry.

PR #3500, opened 2020-01-14
HPS: support check payments

therufs:
CE-205

Unit:
36 tests, 179 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

All remote tests pass between existing credentials and new credentials, but some things aren't yet turned on for new credentials (I've sent an email).

### DIFF
--- a/lib/active_merchant/billing/gateways/hps.rb
+++ b/lib/active_merchant/billing/gateways/hps.rb
@@ -32,10 +32,10 @@ module ActiveMerchant #:nodoc:
         commit('CreditAuth') do |xml|
           add_amount(xml, money)
           add_allow_dup(xml)
-          add_customer_data(xml, card_or_token, options)
+          add_card_or_token_customer_data(xml, card_or_token, options)
           add_details(xml, options)
           add_descriptor_name(xml, options)
-          add_payment(xml, card_or_token, options)
+          add_card_or_token_payment(xml, card_or_token, options)
           add_three_d_secure(xml, card_or_token, options)
         end
       end
@@ -47,15 +47,11 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def purchase(money, card_or_token, options={})
-        commit('CreditSale') do |xml|
-          add_amount(xml, money)
-          add_allow_dup(xml)
-          add_customer_data(xml, card_or_token, options)
-          add_details(xml, options)
-          add_descriptor_name(xml, options)
-          add_payment(xml, card_or_token, options)
-          add_three_d_secure(xml, card_or_token, options)
+      def purchase(money, payment_method, options={})
+        if payment_method.is_a?(Check)
+          commit_check_sale(money, payment_method, options)
+        else
+          commit_credit_sale(money, payment_method, options)
         end
       end
 
@@ -64,22 +60,28 @@ module ActiveMerchant #:nodoc:
           add_amount(xml, money)
           add_allow_dup(xml)
           add_reference(xml, transaction_id)
-          add_customer_data(xml, transaction_id, options)
+          add_card_or_token_customer_data(xml, transaction_id, options)
           add_details(xml, options)
         end
       end
 
       def verify(card_or_token, options={})
         commit('CreditAccountVerify') do |xml|
-          add_customer_data(xml, card_or_token, options)
+          add_card_or_token_customer_data(xml, card_or_token, options)
           add_descriptor_name(xml, options)
-          add_payment(xml, card_or_token, options)
+          add_card_or_token_payment(xml, card_or_token, options)
         end
       end
 
       def void(transaction_id, options={})
-        commit('CreditVoid') do |xml|
-          add_reference(xml, transaction_id)
+        if options[:check_void]
+          commit('CheckVoid') do |xml|
+            add_reference(xml, transaction_id)
+          end
+        else
+          commit('CreditVoid') do |xml|
+            add_reference(xml, transaction_id)
+          end
         end
       end
 
@@ -92,10 +94,34 @@ module ActiveMerchant #:nodoc:
           gsub(%r((<hps:CardNbr>)[^<]*(<\/hps:CardNbr>))i, '\1[FILTERED]\2').
           gsub(%r((<hps:CVV2>)[^<]*(<\/hps:CVV2>))i, '\1[FILTERED]\2').
           gsub(%r((<hps:SecretAPIKey>)[^<]*(<\/hps:SecretAPIKey>))i, '\1[FILTERED]\2').
-          gsub(%r((<hps:PaymentData>)[^<]*(<\/hps:PaymentData>))i, '\1[FILTERED]\2')
+          gsub(%r((<hps:PaymentData>)[^<]*(<\/hps:PaymentData>))i, '\1[FILTERED]\2').
+          gsub(%r((<hps:RoutingNumber>)[^<]*(<\/hps:RoutingNumber>))i, '\1[FILTERED]\2').
+          gsub(%r((<hps:AccountNumber>)[^<]*(<\/hps:AccountNumber>))i, '\1[FILTERED]\2')
       end
 
       private
+
+      def commit_check_sale(money, check, options)
+        commit('CheckSale') do |xml|
+          add_check_payment(xml, check, options)
+          add_amount(xml, money)
+          add_sec_code(xml, options)
+          add_check_customer_data(xml, check, options)
+          add_details(xml, options)
+        end
+      end
+
+      def commit_credit_sale(money, card_or_token, options)
+        commit('CreditSale') do |xml|
+          add_amount(xml, money)
+          add_allow_dup(xml)
+          add_card_or_token_customer_data(xml, card_or_token, options)
+          add_details(xml, options)
+          add_descriptor_name(xml, options)
+          add_card_or_token_payment(xml, card_or_token, options)
+          add_three_d_secure(xml, card_or_token, options)
+        end
+      end
 
       def add_reference(xml, transaction_id)
         xml.hps :GatewayTxnId, transaction_id
@@ -105,7 +131,7 @@ module ActiveMerchant #:nodoc:
         xml.hps :Amt, amount(money) if money
       end
 
-      def add_customer_data(xml, credit_card, options)
+      def add_card_or_token_customer_data(xml, credit_card, options)
         xml.hps :CardHolderData do
           if credit_card.respond_to?(:number)
             xml.hps :CardHolderFirstName, credit_card.first_name if credit_card.first_name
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -47,15 +47,25 @@ def capture(money, transaction_id, options={})
...
-          add_details(xml, options)
-          add_descriptor_name(xml, options)
-          add_payment(xml, card_or_token, options)
-          add_three_d_secure(xml, card_or_token, options)
+      def purchase(money, payment_method, options={})
+        if payment_method.is_a?(Check)
+          commit('CheckSale') do |xml|
+            add_check_payment(xml, payment_method, options)
+            add_amount(xml, money)
+            xml.hps :SECCode, 'WEB'
```
therufs: Thinking of making this a GSF, since we already need to implement one for `check_void`.

therufs: Also thinking about making it a helper method for uniformity, but this is literally all that will ever go in it, so I'm 50/50 on that.

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -159,6 +185,17 @@ def add_payment(xml, card_or_token, options)
...
 
+      def add_check_payment(xml, check, options)
+        xml.hps :CheckAction, 'SALE'
+        xml.hps :AccountInfo do
+          xml.hps :RoutingNumber, check.routing_number
+          xml.hps :AccountNumber, check.account_number
+          xml.hps :CheckNumber, check.number
+          xml.hps :AccountType, check.account_type.upcase
+        end
+        xml.hps :CheckType, check.account_holder_type.upcase
```
therufs: These fields are all required by HPS; attempting a payment without them will result in a gateway processing error.

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -295,18 +332,19 @@ def commit(action, &request)
         )
       end
 
+      SUCCESSFUL_RESPONSE_CODES = %w(0 00 85)
       def successful?(response)
         (
           (response['GatewayRspCode'] == '0') &&
-          ((response['RspCode'] || '00') == '00' || response['RspCode'] == '85')
```
therufs: It looks to me like this says "if there's no response code, pretend we got a response code of 00"; this is a little perplexing, but I wasn't convinced that trying to dig into this would be a good use of time.

test/remote/gateways/remote_hps_test.rb
```diff
@@ -5,8 +5,10 @@ def setup
     @gateway = HpsGateway.new(fixtures(:hps))
 
     @amount = 100
+    @check_amount = 2000
```
therufs: HPS requires a minimum value for a check transaction.

test/unit/gateways/hps_test.rb
```diff
@@ -114,6 +124,14 @@ def test_successful_void
     assert_success void
   end
 
+  def test_successful_check_void
```
therufs: I really wish any of the tests were actually checking the relevant XML for correctness! I may still do that here, since we'd like to make sure that we're sending check txns marked as such.

lib/active_merchant/billing/gateways/hps.rb
leila-alderman: I agree with being 50/50 on making this a helper method. There are a good number of single line methods in this gateway, though, so that might tip the scales sufficiently into making it into a method.

I do think that it would be worthwhile to make this a GSF. I can certainly see a use case for specifying telephone orders.

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -47,15 +47,25 @@ def capture(money, transaction_id, options={})
...
-          add_payment(xml, card_or_token, options)
-          add_three_d_secure(xml, card_or_token, options)
+      def purchase(money, payment_method, options={})
+        if payment_method.is_a?(Check)
+          commit('CheckSale') do |xml|
+            add_check_payment(xml, payment_method, options)
+            add_amount(xml, money)
+            xml.hps :SECCode, 'WEB'
+            add_details(xml, options)
+            add_check_customer_data(xml, payment_method, options)
```
leila-alderman: Looking [at the docs](https://cert.api2.heartlandportico.com/Gateway/PorticoSOAPSchema/build/Default/webframe.html#Portico_xsd~e-PosRequest~e-Ver1.0~e-Transaction~e-CheckSale.html), the order of these methods should be switched.

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -124,7 +142,15 @@ def add_customer_data(xml, credit_card, options)
         end
       end
 
-      def add_payment(xml, card_or_token, options)
+      def add_check_customer_data(xml, check, options)
+        xml.hps :ConsumerInfo do
+          xml.hps :FirstName, check.first_name
+          xml.hps :LastName, check.last_name
+          xml.hps :CheckName, check.bank_name
```
leila-alderman: I might be reading this incorrectly, but this seems to be an option for setting a business name, not the name of the bank:
image

lib/active_merchant/billing/gateways/hps.rb
leila-alderman: This is indeed a bit perplexing. The approach you took makes enough sense to me. 👍

test/unit/gateways/hps_test.rb
leila-alderman: Are you still working on getting the remote tests to work to verify that the XML is correct?

therufs: I am; all the different tests do pass, but I have to use two different sets of creds to get them to do so. Since the provided creds won't have echeck support, it seems like it'll be especially useful.

lib/active_merchant/billing/gateways/hps.rb
therufs: Possibly alarmingly, the tests seem to run either way, and HPS's test suite has them in the specified order. But the remote tests pass either way, so I've switched them.

lib/active_merchant/billing/gateways/hps.rb
therufs: I definitely thought I had seen evidence to the contrary elsewhere, but now I can't find it! 🤷‍♀

therufs:
!!!!!!

Remote:
42 tests, 108 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 36 tests, 179 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -47,15 +47,25 @@ def capture(money, transaction_id, options={})
...
-        commit('CreditSale') do |xml|
-          add_amount(xml, money)
-          add_allow_dup(xml)
-          add_customer_data(xml, card_or_token, options)
-          add_details(xml, options)
-          add_descriptor_name(xml, options)
-          add_payment(xml, card_or_token, options)
-          add_three_d_secure(xml, card_or_token, options)
+      def purchase(money, payment_method, options={})
+        if payment_method.is_a?(Check)
```
fatcatt316: Just a nitpicky style idea, but might be nice to break these two different paths up into two methods (maybe like `def purchase_with_check` and `def purchase_with_credit`). It'd shorten this method up a bit. What do you think?

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -92,7 +108,9 @@ def scrub(transcript)
           gsub(%r((<hps:CardNbr>)[^<]*(<\/hps:CardNbr>))i, '\1[FILTERED]\2').
           gsub(%r((<hps:CVV2>)[^<]*(<\/hps:CVV2>))i, '\1[FILTERED]\2').
           gsub(%r((<hps:SecretAPIKey>)[^<]*(<\/hps:SecretAPIKey>))i, '\1[FILTERED]\2').
-          gsub(%r((<hps:PaymentData>)[^<]*(<\/hps:PaymentData>))i, '\1[FILTERED]\2')
+          gsub(%r((<hps:PaymentData>)[^<]*(<\/hps:PaymentData>))i, '\1[FILTERED]\2').
+          gsub(%r((<hps:RoutingNumber>)[^<]*(<\/hps:RoutingNumber>))i, '\1[FILTERED]\2').
```
fatcatt316: Nice thought to add these here 👍

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -296,21 +337,22 @@ def commit(action, &request)
...
       end
 
       def message_from(response)
         if response['Fault']
           response['Fault']
         elsif response['GatewayRspCode'] == '0'
-          if response['RspCode'] != '00' && response['RspCode'] != '85'
-            issuer_message(response['RspCode'])
-          else
+          if SUCCESSFUL_RESPONSE_CODES.include? response['RspCode']
```
fatcatt316: Setting `SUCCESSFUL_RESPONSE_CODES` in one place makes this all a lil' more readable 👍

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -296,21 +337,22 @@ def commit(action, &request)
         )
       end
 
+      SUCCESSFUL_RESPONSE_CODES = %w(0 00 85)
```
fatcatt316: Where'd the `0` come from? I see where the `00` and `85` were before.

test/remote/gateways/remote_hps_test.rb
```diff
@@ -3,10 +3,13 @@
 class RemoteHpsTest < Test::Unit::TestCase
   def setup
     @gateway = HpsGateway.new(fixtures(:hps))
+    @check_gateway = HpsGateway.new(fixtures(:hps_echeck))
 
     @amount = 100
+    @check_amount = 2000
     @declined_amount = 1034
     @credit_card =   credit_card('4000100011112224')
```
fatcatt316: Interesting spacing on this line :) (I know your PR doesn't touch this, just jumped out at me).

test/remote/gateways/remote_hps_test.rb
```diff
@@ -21,6 +24,13 @@ def test_successful_purchase
     assert_equal 'Success', response.message
   end
 
+  def test_successful_check_purchase
+    options = @options.merge(company_name: 'Hot Buttered Toast Incorporated')
```
fatcatt316: 🍞 🔥

lib/active_merchant/billing/gateways/hps.rb
therufs: Because of reasons, 0 is the successful return code for a check transaction -- https://github.com/activemerchant/active_merchant/pull/3500/files/675919e04c8c71b93cd524e2a7a5b7c50bc11c25#diff-cc2188012eda9d8edc721798125c486fR505 is a sample of such a response.

test/remote/gateways/remote_hps_test.rb
therufs: Rubocop will come for it one of these days, I'm sure :D

lib/active_merchant/billing/gateways/hps.rb
therufs: Seems reasonable; now done :)

test/unit/gateways/hps_test.rb
therufs: Resolving this because XML checks have been added AND the test creds have been managed!

PR #3504, opened 2020-01-17
Merchant Warrior: Add recurringFlag to purchase & authorize

Carrigan:
## Why?

The Merchant Warrior gateway does not support the `recurringFlag` field which is used to indicate if a transaction will be marked as recurring.

- https://dox.merchantwarrior.com/#processcard3162
- https://dox.merchantwarrior.com/#processauth3163
- CE-348

## What Changed?

- Added support for the `recurringFlag` field via a `recurring_flag` option on the `process` and `authorize` functions
- Added missing unit test for Merchant Warrior `authorize` function

## Test Suite Results

Unit:
```
17 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
```

Remote:
```
15 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed
```

### DIFF
--- a/lib/active_merchant/billing/gateways/merchant_warrior.rb
+++ b/lib/active_merchant/billing/gateways/merchant_warrior.rb
@@ -30,6 +30,7 @@ module ActiveMerchant #:nodoc:
         add_order_id(post, options)
         add_address(post, options)
         add_payment_method(post, payment_method)
+        add_recurring_flag(post, options)
         commit('processAuth', post)
       end
 
@@ -39,6 +40,7 @@ module ActiveMerchant #:nodoc:
         add_order_id(post, options)
         add_address(post, options)
         add_payment_method(post, payment_method)
+        add_recurring_flag(post, options)
         commit('processCard', post)
       end
 
@@ -140,6 +142,12 @@ module ActiveMerchant #:nodoc:
         post['transactionAmount'] = amount(money)
         post['transactionCurrency'] = currency
         post['hash'] = verification_hash(amount(money), currency)
+      end
+
+      def add_recurring_flag(post, options)
+        return if options[:recurring_flag].nil?
+
+        post['recurringFlag'] = options[:recurring_flag]
       end
 
       def verification_hash(money, currency)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/merchant_warrior.rb
```diff
@@ -142,6 +144,12 @@ def add_amount(post, money, options)
         post['hash'] = verification_hash(amount(money), currency)
       end
 
+      def add_recurring_flag(post, options)
+        return if options[:recurring_flag].nil?
+
+        post['recurringFlag'] = options[:recurring_flag]
```
leila-alderman: Very minor suggestion to consider:

```
  post['recurringFlag'] = options[:recurring_flag] if options[:recurring_flag]
```

test/remote/gateways/remote_merchant_warrior_test.rb
```diff
@@ -137,6 +137,16 @@ def test_successful_purchase_with_funky_names
     assert_success purchase
   end
 
+  def test_successful_purchase_with_recurring_flag
+    @options[:recurring_flag] = 1
+    test_successful_purchase
```
leila-alderman: 😮 This approach has never occurred to me... Brilliant!!

test/unit/gateways/merchant_warrior_test.rb
```diff
@@ -21,6 +21,16 @@ def setup
     }
   end
 
+  def test_successful_authorize
```
leila-alderman: Thanks for adding in this missing unit test! 🎉

lib/active_merchant/billing/gateways/merchant_warrior.rb
Carrigan: Does this provide anything that the guard clause on line 148 doesn't do?

leila-alderman: Nope! It's the same thing just written slightly differently. Either way works. If you prefer the guard clause approach, that's more than fine, although I might tweak it a bit to
```
return unless options[:recurring_flag]
```

Carrigan: Got it, thanks! I think I will leave as is so that it has the option of being a boolean flag in the future if needed.

PR #3512, opened 2020-01-27
Mercado Pago: Add taxes and net_amount gateway specific fields

Carrigan:
## Why?

- For Colombian users to supply IVA tax information to Mercado Pago, support is required for the `taxes` and `net_amount` fields on purchase requests. 
- https://www.mercadopago.com.co/developers/en/guides/localization/iva-colombia/
- CE-365

## What Changed?

- Added support for `taxes` and `net_amount` fields on the Mercado Pago gateway. These optional fields allow Colombian users to add IVA tax information to transactions. 
- The `taxes` field has support for a single tax object or an array of tax objects to better support upstream XML representation of this field.

## Test Suite

Unit:

```
4433 tests, 71395 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed
```

Remote:

```
31 tests, 93 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.5484% passed
```

Remote Test Failures (also on master, unrelated):
- test_partial_capture
- test_successful_purchase_with_processing_mode_gateway

### DIFF
--- a/lib/active_merchant/billing/gateways/mercado_pago.rb
+++ b/lib/active_merchant/billing/gateways/mercado_pago.rb
@@ -96,6 +96,8 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, payment, options)
         add_address(post, options)
         add_processing_mode(post, options)
+        add_net_amount(post, options)
+        add_taxes(post, options)
         post[:binary_mode] = (options[:binary_mode].nil? ? true : options[:binary_mode])
         post
       end
@@ -194,6 +196,47 @@ module ActiveMerchant #:nodoc:
         post[:token] = options[:card_token]
         post[:issuer_id] = options[:issuer_id] if options[:issuer_id]
         post[:payment_method_id] = options[:payment_method_id] if options[:payment_method_id]
+      end
+
+      def add_net_amount(post, options)
+        post[:net_amount] = Float(options[:net_amount]) if options[:net_amount]
+      end
+
+      def add_taxes(post, options)
+        return unless (tax_object = options[:taxes])
+
+        if tax_object.is_a?(Array)
+          post[:taxes] = process_taxes_array(tax_object)
+        elsif tax_object.is_a?(Hash)
+          post[:taxes] = process_taxes_hash(tax_object)
+        else
+          raise taxes_error
+        end
+      end
+
+      def process_taxes_hash(tax_object)
+        [sanitize_taxes_hash(tax_object)]
+      end
+
+      def process_taxes_array(taxes_array)
+        taxes_array.map do |tax_object|
+          raise taxes_error unless tax_object.is_a?(Hash)
+
+          sanitize_taxes_hash(tax_object)
+        end
+      end
+
+      def sanitize_taxes_hash(tax_object)
+        tax_value = tax_object['value'] || tax_object[:value]
+        tax_type = tax_object['type'] || tax_object[:type]
+
+        raise taxes_error if tax_value.nil? || tax_type.nil?
+
+        { value: Float(tax_value), type: tax_type }
+      end
+
+      def taxes_error
+        ArgumentError.new("Taxes should be a single object or array of objects with the shape: { value: 500, type: 'IVA' }")
       end
 
       def parse(body)
--- a/test/remote/gateways/remote_mercado_pago_test.rb
+++ b/test/remote/gateways/remote_mercado_pago_test.rb
@@ -4,9 +4,11 @@ class RemoteMercadoPagoTest < Test::Unit::TestCase
   def setup
     @gateway = MercadoPagoGateway.new(fixtures(:mercado_pago))
     @argentina_gateway = MercadoPagoGateway.new(fixtures(:mercado_pago_argentina))
+    @colombian_gateway = MercadoPagoGateway.new(fixtures(:mercado_pago_colombia))
 
     @amount = 500
     @credit_card = credit_card('4509953566233704')
+    @colombian_card = credit_card('4013540682746260')
     @elo_credit_card = credit_card('5067268650517446',
       :month => 10,
       :year => 2020,
@@ -86,6 +88,21 @@ class RemoteMercadoPagoTest < Test::Unit::TestCase
     amex_card = credit_card('375365153556885', brand: 'american_express', verification_value: '1234')
 
     response = @gateway.purchase(@amount, amex_card, @options)
+    assert_success response
+    assert_equal 'accredited', response.message
+  end
+
+  def test_successful_purchase_with_taxes_and_net_amount
+    # Minimum transaction amount is 0.30 EUR or ~1112 $COL on 1/27/20.
+    # This value must exceed that
+    amount = 10000_00
+
+    # These values need to be represented as dollars, so divide them by 100
+    tax_amount = amount * 0.19
+    @options[:net_amount] = (amount - tax_amount) / 100
+    @options[:taxes] = [{ value: tax_amount / 100, type: 'IVA' }]
+
+    response = @colombian_gateway.purchase(amount, @colombian_card, @options)
     assert_success response
     assert_equal 'accredited', response.message
   end
--- a/test/unit/gateways/mercado_pago_test.rb
+++ b/test/unit/gateways/mercado_pago_test.rb
@@ -329,6 +329,106 @@ class MercadoPagoTest < Test::Unit::TestCase
     assert_equal '4141491|1.0', response.authorization
   end
 
+  def test_purchase_includes_taxes_array
+    taxes_type = 'IVA'
+    taxes_value = 500.0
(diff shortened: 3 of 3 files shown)


lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -196,6 +198,46 @@ def add_payment(post, options)
         post[:payment_method_id] = options[:payment_method_id] if options[:payment_method_id]
       end
 
+      def add_net_amount(post, options)
+        post[:net_amount] = options[:net_amount].to_f if options[:net_amount]
```
leila-alderman: Since we're adding this field to a JSON object, does the `.to_f` here make a difference?

lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -196,6 +198,46 @@ def add_payment(post, options)
...
+        tax_value = tax_object['value'] || tax_object[:value]
+        tax_type = tax_object['type'] || tax_object[:type]
+
+        raise taxes_error if tax_value.nil? || tax_type.nil?
+
+        { value: tax_value.to_f, type: tax_type }
+      end
+
+      def taxes_error
+        ArgumentError.new("Taxes should be a single object or array of objects with the shape: { value: 500, type: 'IVA' }")
```
leila-alderman: Minor nitpick to add "following" before the colon here:

```suggestion
        ArgumentError.new("Taxes should be a single object or array of objects with the following shape: { value: 500, type: 'IVA' }")
```

lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -196,6 +198,46 @@ def add_payment(post, options)
...
+        end
+      end
+
+      def sanitize_taxes_hash(tax_object)
+        tax_value = tax_object['value'] || tax_object[:value]
+        tax_type = tax_object['type'] || tax_object[:type]
+
+        raise taxes_error if tax_value.nil? || tax_type.nil?
+
+        { value: tax_value.to_f, type: tax_type }
```
leila-alderman: Same question here as above about whether `.to_f` is needed.

lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -196,6 +198,46 @@ def add_payment(post, options)
...
+      def process_taxes_array(taxes_array)
+        taxes_array.map do |tax_object|
+          raise taxes_error unless tax_object.is_a?(Hash)
+          sanitize_taxes_hash(tax_object)
+        end
+      end
+
+      def sanitize_taxes_hash(tax_object)
+        tax_value = tax_object['value'] || tax_object[:value]
+        tax_type = tax_object['type'] || tax_object[:type]
```
leila-alderman: Did you notice in your tests that these were inconsistent in whether they were symbols or strings?

test/remote/gateways/remote_mercado_pago_test.rb
```diff
@@ -90,6 +92,20 @@ def test_successful_purchase_with_american_express
     assert_equal 'accredited', response.message
   end
 
+  def test_successful_purchase_with_taxes_and_net_amount
+    # Minimum transaction amount is 0.30 EUR or ~1112 $COL on 1/27/20.
+    # This value must exceed that
+    @amount = 10000_00
```
leila-alderman: Should this be a normal variable and not an instance variable?

What does the underscore in this value indicate? Is it different than a decimal?

test/remote/gateways/remote_mercado_pago_test.rb
```diff
@@ -280,4 +296,10 @@ def test_transcript_scrubbing
     assert_scrubbed(@credit_card.verification_value, transcript)
     assert_scrubbed(@gateway.options[:access_token], transcript)
   end
+
+  private
+
+  def format_dollars_amount(amount)
+    sprintf('%.2f', amount.to_f / 100).to_f
```
leila-alderman: Why is this method needed? Would it perhaps be simpler to just directly define values for `amount` and `tax_amount`?

test/unit/gateways/mercado_pago_test.rb
```diff
@@ -329,6 +329,104 @@ def test_includes_issuer_id
     assert_equal '4141491|1.0', response.authorization
   end
 
+  def test_purchase_includes_taxes_array
```
leila-alderman: Do you think it would be worthwhile to add a test for an array containing multiple objects? That seems like the most complicated use case.

lib/active_merchant/billing/gateways/mercado_pago.rb
Carrigan: I added this for the case where a user submits the field as a string containing a floating point, i.e. `"net_amount": "5.60"` would be converted to `"net_amount": 5.6` which is what Mercado Pago expects from testing.

lib/active_merchant/billing/gateways/mercado_pago.rb
Carrigan: Same answer as above: this is just extra sanitation.

lib/active_merchant/billing/gateways/mercado_pago.rb
Carrigan: Since these fields are nested inside of the hash, they will come in as strings from an upstream API. For example,  the Adyen `split` field takes these only as strings:

https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/adyen_test.rb#L342-L366

This is a bit confusing since your test payload alternates between strings and symbols. I thought that accepting either could save some grief in the future but am open to suggestions here.

test/remote/gateways/remote_mercado_pago_test.rb
Carrigan: You are right, this could be an instance variable since no other places are using it. The underscore is a visual helper only where `10000_00 == 1000000`. Normally it is used to break up large numbers where commas would go, i.e. `10_000_000` makes easily see that it is 10 million. Similarly since this is on cents, `10000_00` makes it easy to quickly see that this would be 10 thousand Colombian pesos.

test/unit/gateways/mercado_pago_test.rb
Carrigan: One of the failure cases has multiple objects, but I will add that in the successful case as well :+1:

test/remote/gateways/remote_mercado_pago_test.rb
Carrigan: I grabbed this from the `amount` function in gateway.rb to be consistent on money formatting: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L250-L268

It looks like it is used to truncate the decimal portion of the amount, but since we are supplying the values here rather than a user, I agree that it is not needed.

lib/active_merchant/billing/gateways/mercado_pago.rb
leila-alderman: I was mostly wondering if you saw something in particular. The strings vs. symbols was very confusing for me in that code, and this seems like a great way to ensure that the correct values are coming through!

leila-alderman:
I forgot to note this earlier, but there are also some RuboCop complaints that you'll need to clean up: https://travis-ci.org/activemerchant/active_merchant/jobs/642517823?utm_medium=notification&utm_source=github_status

To run RuboCop (and all of the unit tests) locally, I like to run `rake test:local` before I push up my branch.

test/remote/gateways/remote_mercado_pago_test.rb
```diff
@@ -90,6 +92,21 @@ def test_successful_purchase_with_american_express
     assert_equal 'accredited', response.message
   end
 
+  def test_successful_purchase_with_taxes_and_net_amount
+    # Minimum transaction amount is 0.30 EUR or ~1112 $COL on 1/27/20.
+    # This value must exceed that
+    amount = 10000_00
+
+    # These values need to be represented as dollars, so divide them by 100
```
leila-alderman: Instead of taking the extra action to divide by 100 here, what about changing the amount to `10000`?

Carrigan: Since this is a remote test, it has to abide by the logic `transaction_amount = net_amount + taxes` or the transaction will fail. This is a bit confusing but amount is in cents because it is a supported ActiveMerchant field and the `net_amount` and `taxes` are in dollars because we just pass them through and that is what the gateway accepts.

leila-alderman: Ah, that makes sense! Thanks for explaining!

Carrigan:
Thank you for the review @leila-alderman 😄

PR #3515, opened 2020-01-29
Allow optional fields in 3DS 2 based transactions

jeremywrowe:
This is the start of an effort to create an escape hatch to allow folks to add
additional request data when creating 3DS 2 based transactions. Credorax is the
first gateway to leverage the change. I had added more gateways, but complexity
grew quickly when doing so.

usage summary:

When creating a 3DS 2 based authorization or purchase, if the option `{ three_ds_2: { optional: { ... } }`
is provided, all fields inside of the optional hash will be applied to the post
body root.

example

```rb
options = {
  three_ds_2: {
    optional: {
      a: 1,
      b: 2
    }
  }
}

credorax.purchase(amount, payment_method, options)

...

post data will look like:

{
  ...existing_post_data,
  a: 1,
  b: 2
}
```

In addition to the change, the signing of the credorax request body no longer fails with frozen strings. Checkout unit tests on master for credorax vs this branch.

```
Unit Tests (1 unrelated failure)
----------------------------------------------------------------------------------
4440 tests, 71132 assertions, 1 failures, 71 errors, 0 pendings, 0 omissions, 0 notifications
98.3784% passed
----------------------------------------------------------------------------------
```

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -132,6 +132,7 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, options)
         add_email(post, options)
         add_3d_secure(post, options)
+        add_3ds_2_optional_fields(post, options)
         add_echo(post, options)
         add_submerchant_id(post, options)
         add_stored_credential(post, options)
@@ -147,6 +148,7 @@ module ActiveMerchant #:nodoc:
         add_customer_data(post, options)
         add_email(post, options)
         add_3d_secure(post, options)
+        add_3ds_2_optional_fields(post, options)
         add_echo(post, options)
         add_submerchant_id(post, options)
         add_stored_credential(post, options)
@@ -226,6 +228,21 @@ module ActiveMerchant #:nodoc:
         transcript.
           gsub(%r((b1=)\d+), '\1[FILTERED]').
           gsub(%r((b5=)\d+), '\1[FILTERED]')
+      end
+
+      def add_3ds_2_optional_fields(post, options)
+        three_ds = options[:three_ds_2] || {}
+
+        if three_ds.has_key?(:optional)
+          three_ds[:optional].each do |key, value|
+            normalized_value = normalize(value)
+            next if normalized_value.nil?
+
+            post[key] = normalized_value unless post[key]
+          end
+        end
+
+        post
       end
 
       private
@@ -405,8 +422,10 @@ module ActiveMerchant #:nodoc:
 
       def sign_request(params)
         params = params.sort
-        params.each { |param| param[1].gsub!(/[<>()\\]/, ' ') }
-        values = params.map { |param| param[1].strip }
+        values = params.map do |param|
+          value = param[1].gsub(/[<>()\\]/, ' ')
+          value.strip
+        end
         Digest::MD5.hexdigest(values.join + @options[:cipher_key])
       end
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* Credorax: Allow optional 3DS 2 fields [jeremywrowe] #3515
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -405,8 +407,10 @@ def commit(action, params, reference_action = nil)
 
       def sign_request(params)
         params = params.sort
-        params.each { |param| param[1].gsub!(/[<>()\\]/, ' ') }
-        values = params.map { |param| param[1].strip }
+        values = params.map do |param|
+          value = param[1].gsub(/[<>()\\]/, ' ')
+          value.strip
+        end
```
britth: this change is totally separate from the 3DS related update, correct?

lib/active_merchant/billing/gateway.rb
```diff
@@ -209,6 +209,20 @@ def add_field_to_post_if_present(post, options, field)
...
+        if three_ds_2.has_key?(:optional)
+          three_ds_2[:optional].each do |key, value|
+            normalized_value = normalize(value)
+            next if normalized_value.nil?
+            post[key] = normalized_value unless post[key]
+          end
+        end
+
+        post
+      end
```
britth: neat 😎 I guess technically anyone could end up using this and adding top level fields to the post if they knew to add this param; are there any risks there? should we maybe require that they also have included options[:execute_threed]? maybe not, just curious

lib/active_merchant/billing/gateways/credorax.rb
jeremywrowe: Yeah, it's to fix an error in tests (and application code if an developer were to freeze attribute values). You'll see the failures if you run unit tests for Credorax on master.

lib/active_merchant/billing/gateway.rb
jeremywrowe: I didn't want to over complicate the logic here. three_ds_2 options come in with execute_threed, so I didn't see too much of a risk when I was implementing. I had considered the change but elected not to do so. That all being said, if someone has a strong opinion on the matter, I'm happy to make the change.

lib/active_merchant/billing/gateway.rb
```diff
@@ -209,6 +209,20 @@ def add_field_to_post_if_present(post, options, field)
         post[field] = options[field] if options[field]
       end
 
+      def add_3ds_2_optional_fields(post, options)
+        three_ds_2 = options[:three_ds_2] || {}
+
+        if three_ds_2.has_key?(:optional)
+          three_ds_2[:optional].each do |key, value|
+            normalized_value = normalize(value)
+            next if normalized_value.nil?
```
Carrigan: Could there be value in passing along `nil` values for some fields? I am thinking of APIs where some value has a default that you'd like to clear by sending `some_value: nil` or where the key is required but you have no value to pass along.

lib/active_merchant/billing/gateway.rb
```diff
@@ -209,6 +209,20 @@ def add_field_to_post_if_present(post, options, field)
...
+            post[key] = normalized_value unless post[key]
```
Carrigan: I like the protection here from hijacking other parts of the post :+1:

lib/active_merchant/billing/gateway.rb
jeremywrowe: it's better to not pass nil along. Since we're using normalize here, a '' or 'null' can turn into a nil.

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -228,6 +230,20 @@ def scrub(transcript)
...
+        if three_ds_2.has_key?(:optional)
+          three_ds_2[:optional].each do |key, value|
+            normalized_value = normalize(value)
+            next if normalized_value.nil?
+            post[key] = normalized_value unless post[key]
+          end
+        end
+
+        post
+      end
```
britth: cool, this feels a little less risky and we can move to gateway.rb later on 👍

britth:
merged on behalf of @jeremywrowe

PR #3517, opened 2020-02-01
Paypal: Fix OrderTotal elements in `add_payment_details`

chinhle23:
ECS-986

Possible `nil` or blank values in the `subtotal`, `shipping`, `handling`,
or `tax` fields will cause the `localized_amount` method to break causing
`NoMethodError`.

This change requires for the affected fields to have a value before being
evaluated.

Unit:
19 tests, 38 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
29 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
+++ b/lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
@@ -586,11 +586,11 @@ module ActiveMerchant #:nodoc:
           xml.tag! 'n2:OrderTotal', localized_amount(money, currency_code), 'currencyID' => currency_code
 
           # All of the values must be included together and add up to the order total
-          if [:subtotal, :shipping, :handling, :tax].all?{ |o| options.has_key?(o) }
-            xml.tag! 'n2:ItemTotal', localized_amount(options[:subtotal], currency_code), 'currencyID' => currency_code
-            xml.tag! 'n2:ShippingTotal', localized_amount(options[:shipping], currency_code),'currencyID' => currency_code
-            xml.tag! 'n2:HandlingTotal', localized_amount(options[:handling], currency_code),'currencyID' => currency_code
-            xml.tag! 'n2:TaxTotal', localized_amount(options[:tax], currency_code), 'currencyID' => currency_code
+          if [:subtotal, :shipping, :handling, :tax].all?{ |o| options[o].present?}
+            xml.tag! 'n2:ItemTotal', localized_amount(options[:subtotal].to_i, currency_code), 'currencyID' => currency_code
+            xml.tag! 'n2:ShippingTotal', localized_amount(options[:shipping].to_i, currency_code),'currencyID' => currency_code
+            xml.tag! 'n2:HandlingTotal', localized_amount(options[:handling].to_i, currency_code),'currencyID' => currency_code
+            xml.tag! 'n2:TaxTotal', localized_amount(options[:tax].to_i, currency_code), 'currencyID' => currency_code
           end
 
           xml.tag! 'n2:InsuranceTotal', localized_amount(options[:insurance_total], currency_code),'currencyID' => currency_code unless options[:insurance_total].blank?
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * Worldpay: Remove unnecessary .tag! methods [leila-alderman] #3519
 * EBANX: Fix `scrub` [chinhle23] #3521
 * Worldpay: Add `riskData` GSF [fatcatt316] #3514
+* Paypal: Fix OrderTotal elements in `add_payment_details` [chinhle23] #3517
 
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
```diff
@@ -586,11 +586,11 @@ def add_payment_details(xml, money, currency_code, options = {})
           xml.tag! 'n2:OrderTotal', localized_amount(money, currency_code), 'currencyID' => currency_code
 
           # All of the values must be included together and add up to the order total
-          if [:subtotal, :shipping, :handling, :tax].all?{ |o| options.has_key?(o) }
-            xml.tag! 'n2:ItemTotal', localized_amount(options[:subtotal], currency_code), 'currencyID' => currency_code
-            xml.tag! 'n2:ShippingTotal', localized_amount(options[:shipping], currency_code),'currencyID' => currency_code
-            xml.tag! 'n2:HandlingTotal', localized_amount(options[:handling], currency_code),'currencyID' => currency_code
-            xml.tag! 'n2:TaxTotal', localized_amount(options[:tax], currency_code), 'currencyID' => currency_code
+          if [:subtotal, :shipping, :handling, :tax].all?{ |o| options[o].present?}
```
britth: 😎

lib/active_merchant/billing/gateways/paypal/paypal_common_api.rb
```diff
@@ -586,11 +586,11 @@ def add_payment_details(xml, money, currency_code, options = {})
...
+            xml.tag! 'n2:ItemTotal', localized_amount(options[:subtotal].to_i, currency_code), 'currencyID' => currency_code
+            xml.tag! 'n2:ShippingTotal', localized_amount(options[:shipping].to_i, currency_code),'currencyID' => currency_code
+            xml.tag! 'n2:HandlingTotal', localized_amount(options[:handling].to_i, currency_code),'currencyID' => currency_code
+            xml.tag! 'n2:TaxTotal', localized_amount(options[:tax].to_i, currency_code), 'currencyID' => currency_code
```
britth: is there ever a chance that these values will need to be more precise than an integer? like for say, a three decimal currency?

chinhle23: Good question. It looks like Paypal currently doesn't support three-decimal currencies (BHD, IQD, JOD, KWD, LYD, OMR, TND): https://developer.paypal.com/docs/api/reference/currency-codes/.  Also, the `localized_amount` method is expecting an integer here according to this line: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L261. It'll convert those values back to a decimal before being sent to the Paypal gateway, so I think we're okay here in case they ever do decide to start supporting three-decimal currencies.

elfassy:
@chinhle23 @britth This PR breaks support for money objects
```ruby
Money.new(1, 'USD').to_i #=> 1
# vs how amount used to handle money objects
Money.new(1, 'USD').cents #=> 100 
```

rymwoo:
@chinhle23 @britth opening a revert PR to discuss

chinhle23:
@elfassy @rymwoo, after looking at this further, I'm not sure if this is a real concern here. We currently set `money_format` to `:dollars` here: https://github.com/activemerchant/active_merchant/blob/601b9acff4049e151ed9430051b77a973da94d8b/lib/active_merchant/billing/gateway.rb#L117 and this change only affects the Paypal gateway. Can you let me know if I'm missing something?

chinhle23:
Same structure is occurring here as well on the CyberSource gateway: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L441

elfassy:
Localized amount simply calls amount ([here](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L279)) which has support for both Money objects and numeric values
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L253-L259

When calling `to_i` on the money object, you only get the dollar amount
```ruby
> Money.new(1.23).to_i
=> 1
```
vs what used to happen, which would call `.cents`
```ruby
> Money.new(1.23).cents
=> 123
```

granted support for money objects is deprecated but removing support should be done in a major release. 
We can add a failing test for this use case.

elfassy:
@chinhle23 cybersource is likely also broken for money objects... we should probably speed up removing support for Money objects as it's been deprecated for 10 years! 😅 https://github.com/activemerchant/active_merchant/commit/0ae2fd69e11deb172711137b8aae7b22234a3e8c

chinhle23:
Ah okay, thank you for the explanation @elfassy! How would we go about removing support for Money objects?

PR #3518, opened 2020-02-03
BPoint: Remove amount from void requests

leila-alderman:
The `void` method for the BPoint gateway was previously implemented
incorrectly. Throughout Active Merchant, gateway `void` requests are
passed only an authorization number for the original transaction and an
options hash. However, for BPoint, the `void` method was also passed an
explicit `amount`. Due to the implementations on top of Active Merchant,
this was causing issues for Spreedly customers.

The BPoint `void` method has been rewritten to take in only
`authorization` and an `options` hash. Because the BPoint gateway

CE-338

Unit:
19 tests, 63 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
18 tests, 41 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

All unit tests:
4425 tests, 71371 assertions, 0 failures, 0 errors, 0 pendings, 2
omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/bpoint.rb
+++ b/lib/active_merchant/billing/gateways/bpoint.rb
@@ -61,10 +61,10 @@ module ActiveMerchant #:nodoc:
         commit(request_body)
       end
 
-      def void(amount, authorization, options={})
+      def void(authorization, options={})
         request_body = soap_request do |xml|
           process_payment(xml) do |payment_xml|
-            add_void(payment_xml, amount, authorization, options)
+            add_void(payment_xml, authorization, options)
           end
         end
         commit(request_body)
@@ -73,7 +73,7 @@ module ActiveMerchant #:nodoc:
       def verify(credit_card, options={})
         MultiResponse.run(:use_first_response) do |r|
           r.process { authorize(100, credit_card, options) }
-          r.process(:ignore_result) { void(100, r.authorization, options) }
+          r.process(:ignore_result) { void(r.authorization, options.merge(amount: 100)) }
         end
       end
 
@@ -154,7 +154,9 @@ module ActiveMerchant #:nodoc:
         transaction_number_xml(xml, transaction_number)
       end
 
-      def add_void(xml, amount, transaction_number, options)
+      def add_void(xml, transaction_number, options)
+        # The amount parameter is required for void requests on BPoint.
+        amount = options[:amount]
         payment_xml(xml, 'REVERSAL', amount, options)
         transaction_number_xml(xml, transaction_number)
       end
--- a/test/unit/gateways/bpoint_test.rb
+++ b/test/unit/gateways/bpoint_test.rb
@@ -88,13 +88,23 @@ class BpointTest < Test::Unit::TestCase
 
   def test_successful_void
     @gateway.expects(:ssl_post).returns(successful_void_response)
-    response = @gateway.void(@amount, '')
+    response = @gateway.void('', amount: 300)
     assert_success response
+  end
+
+  def test_void_passes_correct_transaction_reference
+    stub_comms do
+      # transaction number from successful authorize response
+      @gateway.void('219388558', amount: 300)
+    end.check_request do |endpoint, data, headers|
+      assert_match(%r(<OriginalTransactionNumber>219388558</OriginalTransactionNumber>)m, data)
+      assert_match(%r(<Amount>300</Amount>)m, data)
+    end.respond_with(successful_void_response)
   end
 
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/bpoint.rb
```diff
@@ -154,7 +154,8 @@ def add_refund(xml, amount, transaction_number, options)
         transaction_number_xml(xml, transaction_number)
       end
 
-      def add_void(xml, amount, transaction_number, options)
+      def add_void(xml, transaction_number, options)
+        amount = options[:amount]
```
fatcatt316: In this case, is this a required option? I think so, just wanted to verify that.

test/unit/gateways/bpoint_test.rb
```diff
@@ -88,13 +88,23 @@ def test_failed_refund
...
   def test_successful_void
     @gateway.expects(:ssl_post).returns(successful_void_response)
-    response = @gateway.void(@amount, '')
+    response = @gateway.void('', amount: 300)
     assert_success response
   end
 
+  def test_void_passes_correct_transaction_reference
+    stub_comms do
+      # transaction number from successful authorize response
```
fatcatt316: Helpful comment! 👍

lib/active_merchant/billing/gateways/bpoint.rb
```diff
@@ -61,10 +61,10 @@ def refund(amount, authorization, options={})
         commit(request_body)
       end
 
-      def void(amount, authorization, options={})
+      def void(authorization, options={})
```
fatcatt316: Do we need to update any code that was previously calling `void` with `amount` as the first argument?

lib/active_merchant/billing/gateways/bpoint.rb
leila-alderman: Yes, for the BPoint gateway, the `amount` is a required field for void transactions. This has been confirmed both through remote tests and an inquiry to BPoint's support staff. It's unusual for a void transaction to require this field, but BPoint apparently does. 🤷‍♀ 

Maybe a comment to confirm that would be helpful here?

lib/active_merchant/billing/gateways/bpoint.rb
leila-alderman: So the entire reason for this bug is that Core never includes `amount` when making a `void` request; it just sends the authorization and the options hash. I have a separate PR out for the Core half of this problem, which gets around this issue by passing the required `amount` field within the `options` hash. 

I'm not aware of anywhere else that this method is called.

lib/active_merchant/billing/gateways/bpoint.rb
fatcatt316: It doesn't hurt! But if it feels like overkill, I wouldn't sweat it. Thanks for that answer.

lib/active_merchant/billing/gateways/bpoint.rb
fatcatt316: Yep, I checked out that other PR, and kinda realized this, too. Interesting that voids never worked for BPoint 🤔

leila-alderman: Sadly, BPoint is not alone in this... In investigating this, Chinh and I realized that Bambora Apac and Merchant Warrior have the same problem for their `void` transactions. We created new bug tickets to resolve those while we're at it. 

This problem seems to stem from the fact that we don't have any automated end-to-end testing that runs through both Core and Active Merchant.

leila-alderman:
@fatcatt316 A lot of the credit for diagnosing and solving this one goes to @chinhle23! 

@chinhle23 asked me the same thing about my branch! There's a comment [in our internal documentation here](https://paper.dropbox.com/doc/Active-Merchant-Merging-and-Updating-it-in-Core--Atpyd5rMPTNBP~ZnBPLO3bypAg-u5a9PskAKQrM4RiRagLCd#:uid=870165264084605343588833&h2=Committing-to-and-opening-a-PR) that is why I push directly to an upstream branch. I've seen it both ways, so I don't believe that there is a preference.

lib/active_merchant/billing/gateways/bpoint.rb
fatcatt316: Setting up that kind of automated integration tests would be a big effort, but I wonder if the payoff would be worth it. I _kinda_ think it would.

PR #3519, opened 2020-02-03
Worldpay: Remove unnecessary .tag! methods

leila-alderman:
When building an XML request, the `.tag!` method is only required if the
XML tag is a non-standard string, e.g., if it contains a hyphen or a
colon. Therefore, this method is often not needed and can be replaced
with a far more concise version.

This commit removes all unnecessary `.tag!` method calls for the
Worldpay gateway to improve the readability of the code.

Unit:
69 tests, 411 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
56 tests, 240 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
96.4286% passed

All unit tests:
4426 tests, 71371 assertions, 0 failures, 0 errors, 0 pendings, 2
omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -166,7 +166,7 @@ module ActiveMerchant #:nodoc:
         xml = Builder::XmlMarkup.new :indent => 2
         xml.instruct! :xml, :encoding => 'UTF-8'
         xml.declare! :DOCTYPE, :paymentService, :PUBLIC, '-//WorldPay//DTD WorldPay PaymentService v1//EN', 'http://dtd.worldpay.com/paymentService_v1.dtd'
-        xml.tag! 'paymentService', 'version' => '1.4', 'merchantCode' => @options[:login] do
+        xml.paymentService 'version' => '1.4', 'merchantCode' => @options[:login] do
           yield xml
         end
         xml.target!
@@ -174,8 +174,8 @@ module ActiveMerchant #:nodoc:
 
       def build_order_modify_request(authorization)
         build_request do |xml|
-          xml.tag! 'modify' do
-            xml.tag! 'orderModification', 'orderCode' => authorization do
+          xml.modify do
+            xml.orderModification 'orderCode' => authorization do
               yield xml
             end
           end
@@ -184,20 +184,20 @@ module ActiveMerchant #:nodoc:
 
       def build_order_inquiry_request(authorization, options)
         build_request do |xml|
-          xml.tag! 'inquiry' do
-            xml.tag! 'orderInquiry', 'orderCode' => authorization
+          xml.inquiry do
+            xml.orderInquiry 'orderCode' => authorization
           end
         end
       end
 
       def build_authorization_request(money, payment_method, options)
         build_request do |xml|
-          xml.tag! 'submit' do
-            xml.tag! 'order', order_tag_attributes(options) do
+          xml.submit do
+            xml.order order_tag_attributes(options) do
               xml.description(options[:description].blank? ? 'Purchase' : options[:description])
               add_amount(xml, money, options)
               if options[:order_content]
-                xml.tag! 'orderContent' do
+                xml.orderContent do
                   xml.cdata! options[:order_content]
                 end
               end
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -166,16 +166,16 @@ def build_request
         xml = Builder::XmlMarkup.new :indent => 2
         xml.instruct! :xml, :encoding => 'UTF-8'
         xml.declare! :DOCTYPE, :paymentService, :PUBLIC, '-//WorldPay//DTD WorldPay PaymentService v1//EN', 'http://dtd.worldpay.com/paymentService_v1.dtd'
-        xml.tag! 'paymentService', 'version' => '1.4', 'merchantCode' => @options[:login] do
+        xml.paymentService 'version' => '1.4', 'merchantCode' => @options[:login] do
```
leila-alderman: It turns out that these hashes don't need to be enclosed by parentheses after all, so I removed them anywhere they weren't needed.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -219,36 +219,34 @@ def order_tag_attributes(options)
...
             add_amount(xml, money, options)
           end
         end
       end
 
       def build_void_request(authorization, options)
-        build_order_modify_request(authorization) do |xml|
-          xml.tag! 'cancel'
-        end
+        build_order_modify_request(authorization, &:cancel)
```
leila-alderman: This is a situation where RuboCop was complaining about the syntax; this seemed to make both RuboCop and the tests happy.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -317,26 +315,33 @@ def add_payment_method(xml, amount, payment_method, options)
...
+        xml.expiryDate do
+          xml.date(
+            'month' => format(payment_method.month, :two_digits),
+            'year' => format(payment_method.year, :four_digits)
+          )
         end
 
-        xml.tag! 'cardHolderName', options[:execute_threed] && !options[:three_ds_version]&.start_with?('2') ? '3D' : payment_method.name
-        xml.tag! 'cvc', payment_method.verification_value
+        if options[:execute_threed] && !options[:three_ds_version]&.start_with?('2')
```
leila-alderman: This was another weird RuboCop case where it would complain that the complicated logic needed to be surrounded by parentheses but then when I added them, it would complain that you shouldn't enclose a ternary condition with parentheses... This seemed like a reasonable way to appease the RuboCop overlords.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -317,26 +315,33 @@ def add_payment_method(xml, amount, payment_method, options)
...
         end
       end
 
       def add_card(xml, payment_method, options)
-        xml.tag! 'cardNumber', payment_method.number
-        xml.tag! 'expiryDate' do
-          xml.tag! 'date', 'month' => format(payment_method.month, :two_digits), 'year' => format(payment_method.year, :four_digits)
+        xml.cardNumber payment_method.number
+        xml.expiryDate do
+          xml.date(
```
fatcatt316: Hooray for shortening up these lines

lib/active_merchant/billing/gateways/worldpay.rb
fatcatt316: Would this work? It seems to work on my laptop (and RuboCop isn't screaming about it):
```
xml.cardHolderName(options[:execute_threed] && !options[:three_ds_version]&.start_with?('2') ? '3D' : payment_method.name)
```
Although it might be more readable to break it up like maybe:
```
card_holder_name = options[:execute_threed] && !options[:three_ds_version]&.start_with?('2') ? '3D' : payment_method.name
xml.cardHolderName(card_holder_name)
```

leila-alderman: That does look much better, especially breaking it out to two lines! Thanks for the suggestion!

PR #3521, opened 2020-02-05
EBANX: Fix `scrub`

chinhle23:
Per EBANX, the `integration_key` can include digits and letters. This
change ensures the full value is properly scrubbed from the transcript.

Unit:
100% passed

Remote:
100% passed


chinhle23:
Hey @jeremywrowe, there's already a test for this: https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_ebanx_test.rb#L226 Do you think that's good enough?

curiousepic:
Does that test actually exercise letters in the field?

chinhle23:
Thanks for the review @jeremywrowe & @curiousepic. I see your point. I pushed an update to include letters and digits in the `integration_key` of `pre_scrubbed`. Does that seem adequate?

PR #3524, opened 2020-02-10
Revert Stripe Payment Intents GSF "cross_border_classification"

fatcatt316:
This parameter wasn't able to be tested through our test account,
and the merchant who was going to test it in production is
no longer with our customer.

Also, I was unable to find documentation of this parameter
beyond a Word doc that the customer sent us.

CE-196

### Local

```
rake TEST=test/unit/gateways/stripe_payment_intents_test.rb

100% passed
```

### Remote

```
rake TEST=test/remote/gateways/remote_stripe_payment_intents_test.rb

30 tests, 123 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.6667% passed
```
Failing test also fails on `master`:
* test_create_payment_intent_that_saves_payment_method

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,7 +5,6 @@
 * Stripe: Remove outdated 'customer options' deprecation [alexdunae] #3401
 * BPoint: Remove amount from void requests [leila-alderman] #3518
 * Worldpay: Remove unnecessary .tag! methods [leila-alderman] #3519
-* Stripe Payment Intents: Allow cross_border_classification parameter [fatcatt316] #3508
 * EBANX: Fix `scrub` [chinhle23] #3521
 * Worldpay: Add `riskData` GSF [fatcatt316] #3514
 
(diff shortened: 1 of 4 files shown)


CHANGELOG
```diff
@@ -8,6 +8,7 @@
 * Stripe Payment Intents: Allow cross_border_classification parameter [fatcatt316] #3508
 * EBANX: Fix `scrub` [chinhle23] #3521
 * Worldpay: Add `riskData` GSF [fatcatt316] #3514
+* Stripe Payment Intents: Revert "Allow cross_border_classification parameter" [fatcatt316] #3524
```
Carrigan: Since these are released on minor version updates, what do you think about just removing line 8 above instead of having the changelog say "Add X, remove X"? The history of this being added and removed will be preserved by git.

fatcatt316: @Carrigan I agree! I thought about that some, too, and will go ahead and make that change now

PR #3525, opened 2020-02-11
Merchant Warrior: Send void amount in options

leila-alderman:
The `void` method for the Merchant Warrior gateway was previously
implemented incorrectly. Throughout Active Merchant, gateway `void`
requests are passed only an authorization number for the original
transaction and an options hash. However, for Merchant Warrior, the void
method was also passed an explicit amount. Due to the implementations on
top of Active Merchant, this was causing issues for Spreedly customers.

The Merchant Warrior `void` method has been rewritten to take in only
authorization and an options hash. Because the Merchant Warrior gateway

In addition, the required verification hash for `void` transactions
should consist of the original transaction ID, not the amount and the
currency. This change also adds a new `void_verification_hash` method
that calculates the correct hash for the `void` method.

CE-379

Unit:
17 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
15 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

All unit tests:
4442 tests, 71468 assertions, 0 failures, 0 errors, 0 pendings, 2
omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/merchant_warrior.rb
+++ b/lib/active_merchant/billing/gateways/merchant_warrior.rb
@@ -60,9 +60,12 @@ module ActiveMerchant #:nodoc:
         commit('refundCard', post)
       end
 
-      def void(money, identification, options = {})
+      def void(identification, options = {})
         post = {}
-        add_amount(post, money, options)
+        # The amount parameter is required for void transactions
+        # on the Merchant Warrior gateway.
+        post['transactionAmount'] = options[:amount]
+        post['hash'] = void_verification_hash(identification)
         add_transaction(post, identification)
         commit('processVoid', post)
       end
@@ -157,6 +160,16 @@ module ActiveMerchant #:nodoc:
             @options[:merchant_uuid].to_s +
             money.to_s +
             currency
+          ).downcase
+        )
+      end
+
+      def void_verification_hash(transaction_id)
+        Digest::MD5.hexdigest(
+          (
+            @options[:api_passphrase].to_s +
+            @options[:merchant_uuid].to_s +
+            transaction_id
           ).downcase
         )
       end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
 * EBANX: Add metadata information in post [miguelxpn] #3522
 * HPS: support eCheck [therufs] #3500
 * Decidir: Add support for fraud_detection, site_id, and establishment_name [fatcatt316] #3527
+* Merchant Warrior: Send void amount in options [leila-alderman] #3525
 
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/merchant_warrior.rb
```diff
@@ -161,6 +164,16 @@ def verification_hash(money, currency)
         )
       end
 
+      def void_verification_hash(transaction_id)
+        Digest::MD5.hexdigest(
+          (
+            @options[:api_passphrase].to_s +
```
fatcatt316: You mentioned `the required verification hash for void transactions should consist of the original transaction ID`. Is it fine also adding in the `api_passphrase` and `merchant_uuid` like what's being done here? I assume so, but figured I'd ask.

lib/active_merchant/billing/gateways/merchant_warrior.rb
```diff
@@ -60,9 +60,12 @@ def refund(money, identification, options = {})
         commit('refundCard', post)
       end
 
-      def void(money, identification, options = {})
+      def void(identification, options = {})
         post = {}
-        add_amount(post, money, options)
+        # The amount parameter is required for void transactions
+        # on the Merchant Warrior gateway.
+        post['transactionAmount'] = options[:amount]
```
fatcatt316: Is it okay that we're not setting `transactionCurrency` anymore in here?

lib/active_merchant/billing/gateways/merchant_warrior.rb
leila-alderman: Yes, those fields are also needed. This was the response from Merchant Warrior to our customer about how to build the hash for a void transaction:
```
Note, the hash is calculated by: strtolower(md5(md5(api_passphrase).uuid.transactionid))
```

lib/active_merchant/billing/gateways/merchant_warrior.rb
leila-alderman: I believe so. When I initially added the void transaction to Merchant Warrior, they sent us this information for the fields that are needed for void:
image

lib/active_merchant/billing/gateways/merchant_warrior.rb
fatcatt316: Ah, great to know.

lib/active_merchant/billing/gateways/merchant_warrior.rb
fatcatt316: Interesting, not a `currency` in sight.

PR #3530, opened 2020-02-17
Authorize.Net: Pass `account_type` to `check` payment types

chinhle23:
https://developer.authorize.net/api/reference/features/echeck.html

A Spreedly customer reported notifications from Authorize.Net that a
savings account transaction was processed as a checking account transaction.
Authorize.Net requires `accountType` to be passed for `eCheck` transactions
in order for proper reporting and to avoid these type of notifications.

Though these `eCheck` transactions have been successful without the `accountType`
field being passed, the customer would like the notifications to stop.

Unit:
98 tests, 584 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
70 tests, 242 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

All unit tests:
4447 tests, 71494 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -544,6 +544,7 @@ module ActiveMerchant
       def add_check(xml, check)
         xml.payment do
           xml.bankAccount do
+            xml.accountType(check.account_type)
             xml.routingNumber(check.routing_number)
             xml.accountNumber(check.account_number)
             xml.nameOnAccount(truncate(check.name, 22))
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -544,6 +544,7 @@ def valid_track_data
       def add_check(xml, check)
         xml.payment do
           xml.bankAccount do
+            xml.accountType(check.account_type)
```
molbrown: Is there ever a situation where there wouldn't be an account_type for a check or is it always defined? (I'm not super familiar with this payment method)

molbrown: Looks like it is required https://github.com/activemerchant/active_merchant/blob/1b2fce718fe211865297805ebe634e5764a8d777/lib/active_merchant/billing/check.rb#L40

PR #3531, opened 2020-02-18
Credorax: Fix `3ds_transtype` setting in post

chinhle23:
ECS-938

https://epower.credorax.com/wp-content/uploads/2019/11/Credorax-Source-Payment-API-Specifications-v1.3-.pdf
The `3ds_transtype` parameter is incorrectly being set by the `transaction_type`
parameter which is meant for setting the `a9` parameter used for stored
credentials transactions and thus causing failed transactions with "At
least one of input parameters is malformed. Parameter 3ds_transtype is
invalid." error.

This change allows the `3ds_transtype` parameter to be explicitly set, otherwise
it defaults to "01"

Unit:
66 tests, 307 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
40 tests, 147 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

All unit tests:
4446 tests, 71485 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -16,6 +16,7 @@
 * Redsys: Update scrub method to account for 3DS error responses [britth] #3534
 * RuboCop: Fix Layout/IndentHash [leila-alderman] #3529
 * Bambora Apac: Send void amount in options [leila-alderman] #3532
+* Credorax: Fix `3ds_transtype` setting in post [chinhle23] #3531
 
 == Version 1.104.0 (Jan 29, 2020)
 * Adyen: add `recurring_contract_type` GSF [therufs] #3460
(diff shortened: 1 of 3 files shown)


britth:
Is there any risk that someone could currently be using transaction_type to set this value, and this would cause issues? I'm guessing no, since if they were using it correctly for 3DS it would probably not work for the a9 param, or vice-versa, so it's not working properly either way?

chinhle23:
> Is there any risk that someone could currently be using transaction_type to set this value, and this would cause issues? I'm guessing no, since if they were using it correctly for 3DS it would probably not work for the a9 param, or vice-versa, so it's not working properly either way?

Yeah, it's not working properly either way...a user would have to pick which field they intend to get  right.

chinhle23:
Thanks for the suggestion @molbrown, I pushed an update to include a unit test. Let me know what you think.

PR #3532, opened 2020-02-18
Bambora Apac: Send void amount in options

leila-alderman:
The `void` method for the Bambora Apac gateway was previously
implemented incorrectly. Throughout Active Merchant, gateway `void`
requests are passed only an authorization number for the original
transaction and an options hash. However, for Bambora Apac, the void
method was also passed an explicit amount. This break in the established
pattern was causing issues for implementations on top of Active
Merchant.

The Bambora Apac `void` method has been rewritten to take in only
authorization and an options hash. Because the Bambora Apac gateway

CE-389

Unit:
7 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0
notifications
100% passed

Remote:
15 tests, 30 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

All unit tests:
4444 tests, 71480 assertions, 0 failures, 0 errors, 0 pendings, 2
omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/bambora_apac.rb
+++ b/lib/active_merchant/billing/gateways/bambora_apac.rb
@@ -30,7 +30,7 @@ module ActiveMerchant #:nodoc:
         commit('SubmitSinglePayment') do |xml|
           xml.Transaction do
             xml.CustRef options[:order_id]
-            add_amount(xml, money)
+            xml.Amount amount(money)
             xml.TrnType '1'
             add_payment(xml, payment)
             add_credentials(xml, options)
@@ -43,7 +43,7 @@ module ActiveMerchant #:nodoc:
         commit('SubmitSinglePayment') do |xml|
           xml.Transaction do
             xml.CustRef options[:order_id]
-            add_amount(xml, money)
+            xml.Amount amount(money)
             xml.TrnType '2'
             add_payment(xml, payment)
             add_credentials(xml, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/bambora_apac.rb
```diff
@@ -72,11 +72,11 @@ def refund(money, authorization, options={})
         end
       end
 
-      def void(money, authorization, options={})
+      def void(authorization, options={})
         commit('SubmitSingleVoid') do |xml|
           xml.Void do
             xml.Receipt authorization
-            add_amount(xml, money)
```
fatcatt316: I like this change! Any reason you spared the other `add_amount` calls in this gateway? It seems like it'd be good to either not that method at all in here, or _only_ use it. What think ye?

leila-alderman: Good point! Since the method was just a single line, I went for the option of deleting the method and replacing it throughout the other transaction methods.

lib/active_merchant/billing/gateways/bambora_apac.rb
```diff
@@ -116,10 +116,6 @@ def add_credentials(xml, options)
         end
       end
 
-      def add_amount(xml, money)
```
fatcatt316: 👋

PR #3536, opened 2020-02-19
Added support for fraud review in CyberSource gateway

greg-burgoon:
…, but flagged for fraud review

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -49,6 +49,11 @@ module ActiveMerchant #:nodoc:
         :elo => '054'
       }
 
+      @@decision_codes = {
+        :accept => 'ACCEPT',
+        :review => 'REVIEW'
+      }
+
       @@response_codes = {
         :r100 => 'Successful transaction',
         :r101 => 'Request is missing one or more required fields',
@@ -828,7 +833,7 @@ module ActiveMerchant #:nodoc:
           response = { message: e.to_s }
         end
 
-        success = response[:decision] == 'ACCEPT'
+        success = success?(response)
         message = response[:message]
 
         authorization = success ? authorization_from(response, action, amount, options) : nil
@@ -836,6 +841,7 @@ module ActiveMerchant #:nodoc:
         Response.new(success, message, response,
           :test => test?,
           :authorization => authorization,
+          :fraud_review => in_fraud_review?(response),
           :avs_result => { :code => response[:avsCode] },
           :cvv_result => response[:cvCode]
         )
@@ -886,6 +892,14 @@ module ActiveMerchant #:nodoc:
       def authorization_from(response, action, amount, options)
         [options[:order_id], response[:requestID], response[:requestToken], action, amount,
          options[:currency], response[:subscriptionID]].join(';')
+      end
+
+      def in_fraud_review?(response)
+        response[:decision] == @@decision_codes[:review]
+      end
+
+      def success?(response)
+        response[:decision] == @@decision_codes[:accept]
       end
     end
   end
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -411,6 +411,13 @@ class CyberSourceTest < Test::Unit::TestCase
     assert_success(@gateway.credit(@amount, @credit_card, @options))
   end
 
+  def test_authorization_under_review_request
+    @gateway.stubs(:ssl_post).returns(authorization_review_response)
+
+    assert_failure(response = @gateway.authorize(@amount, @credit_card, @options))
+    assert response.fraud_review?
+  end
+
   def test_successful_credit_to_subscription_request
     @gateway.stubs(:ssl_post).returns(successful_create_subscription_response, successful_subscription_credit_response)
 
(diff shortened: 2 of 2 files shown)


greg-burgoon:
This is in reference to this (closed) issue: https://github.com/activemerchant/active_merchant/issues/1681.

greg-burgoon:
I've updated this to better align with how other gateways handle fraud review

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -887,6 +895,14 @@ def authorization_from(response, action, amount, options)
         [options[:order_id], response[:requestID], response[:requestToken], action, amount,
          options[:currency], response[:subscriptionID]].join(';')
       end
+
+      def in_fraud_review?(response)
+        response[:decision] == @@decision_codes[:review] && response[:ccAuthReply_reasonCode] == '100'
```
pi3r: What does `100` means? Does it mean that the review was successful?

greg-burgoon: It means the credit card was authorized. The decision = REVIEW means that although it has been authorized, CyberSource has flagged the payment for fraud review based on its Decision Manager rules.

pi3r: What happens if review doesn't happen? What I'm getting at is that most system that rely on ActiveMerchant are not ready to handle a fraud_review state. Like nothing will happened down the line. Does this mean that the payment can be safely captured later on when the decision = REVIEW and code "100"?

greg-burgoon: You raise a good point.  I believe payment can still be successfully captured (I've tested manually settling a transaction that is "in review" in Cybersource, and it still allows me to settle the payment).  

Cybersource flags the request for review, but my tests, there is nothing stopping someone from capturing the payment.  The credit card has been authorized and payment should be able to be captured.  I think this makes sense.

molbrown: @greg-burgoon based on my understanding of this flag, wouldn't it be better to fail the authorization and the merchant can take action to perform the review/ capture the payment based on the returned `fraud_review` value? It seems more risky to mark this as successful, when Cybersource did not ACCEPT the charge. I echo @pi3r's concern about downstream actions not set up to consider the fraud_review state at this time, and hiding fraud risk payments in success state.

Just because it allows you to settle the payment in REVIEW, I'm not convinced a user would always want to.  How are you testing capturing an in-review payment? Is that against a sandbox or production endpoint? I wonder if the behavior is different between them.

greg-burgoon:
@molbrown @pi3r Fair points.  I've updated the logic based on your comments. Key highlights: 1) When `decision = REVIEW` I simply update the `fraud_review` flag for the response.  2) I updated the success logic to be based on when `reasonCode = 100` rather than the decision field (since this is what cybersource documentation recommends.  The decision field is actually a fallback for when the reasonCode is unrecognized). '100' => Successful Transaction

pi3r:
@greg-burgoon Thank you, that looks much better. Do you have a link to the documentation that explains that the decision field is a fallback for the reasonCode?

greg-burgoon:
@pi3r Just re-read the documentation.  Seems a little less clear what should take precedence....http://apps.cybersource.com/library/documentation/dev_guides/Getting_Started_SO/Getting_Started_SO_API.pdf

The paragraph that originally led me to this understanding is here: 
Screen Shot 2020-02-20 at 4 47 08 PM

Thoughts?

molbrown:
Also, based on the recommended options in the **Review** section, it seems defaulting to failure is the proper direction, especially considering Active Merchant use cases:

Screen Shot 2020-02-20 at 5 49 02 PM

greg-burgoon:
@molbrown Yeah, I think the wording is stronger in what you pasted and I agree with the direction, but it raises a question:  If `decision = REVIEW` and `reasonCode = 100`, it will result in a failure....but should it fail?  It just seems slightly odd to me since the resulting message would be "Successful transaction" for this case (and this isn't just an edge case.  I have cybersource configured right now to return things in this manner).  IMO the documentation isn't sufficient enough to describe how to handle this case, so I understand the desire to be safe and just fail the call, but it results in a weird state (failure, with success message) that gets propagated to other systems. 

Anyways,  just wanted to get your thoughts.  If we still feel failure is the right call for this case, I'll update the PR to just set the `fraud_review` flag so we can close out this thread.

pi3r:
The main issue is that the current ActiveMerchant design is dated and doesn't match what gateways are doing nowadays. I totally agree that it should have a way to represent a fraud review case so that the state is pending until review is done. The only issue with this is, systems relying on ActiveMerchant would have to handle this new state, and be able to poll the gateway for the state of the review, which is not at all present right now. As an example, to this day we still see some orders being treated as success even thought the review failed on Authorize.net due to this code change https://github.com/activemerchant/active_merchant/commit/5c479110b2b55807b1d1bbd281a4ab28f1d249f0. 

Do you actually need the `reasonCode` change? Isn' the `fraud_review` attribute in the response enough?

molbrown:
I took a look again at the reasonCode information in the docs, I wonder if it is only an informative field for ERROR or REJECT cases? I'm thinking it would make sense for REVIEW to always be accompanied by a `100`, because if the payment isn't authorized, it would have a REJECT/ ERROR decision.

Screen Shot 2020-02-21 at 1 33 50 PM

It seems Cybersource intends for the merchant to take action before converting the charge to an ACCEPT, and this could be aided through the use of the `fraud_review` flag you are adding here. I'm interpreting the REVIEW decision as a temporary non-success state, until such action is taken. It seems that if a merchant has set review rules to catch certain transactions, they intend to address those transactions before completing them? What do you think?

Changing success logic is a risky change, just trying to make sure we understand as much as we can before deciding to do so.

greg-burgoon:
@pi3r @molbrown I've updated the PR and kept it simple/safe.  It *only* updates the `fraud_review` flag at this point.  Thanks for the feedback.  Let me know if there's any other concerns.

molbrown:
Merged with Changelog entry 34e73c677a928a58510b6cc0dc99354d7df441d5

PR #3538, opened 2020-02-20
pay_junction_v2: Passing billing address to auth and purchase methods

naashton:
Added support to PayJunction v2 gateway to send billing address
information with auth and purchase transactions.

CE-261

Unit: 4451 tests, 71520 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

Remote: 20 tests, 57 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/pay_junction_v2.rb
+++ b/lib/active_merchant/billing/gateways/pay_junction_v2.rb
@@ -21,6 +21,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_invoice(post, amount, options)
         add_payment_method(post, payment_method)
+        add_address(post, options)
 
         commit('purchase', post)
       end
@@ -30,6 +31,7 @@ module ActiveMerchant #:nodoc:
         post[:status] = 'HOLD'
         add_invoice(post, amount, options)
         add_payment_method(post, payment_method)
+        add_address(post, options)
 
         commit('authorize', post)
       end
@@ -108,6 +110,27 @@ module ActiveMerchant #:nodoc:
           post[:cardExpYear] = format(payment_method.year, :four_digits)
           post[:cardCvv] = payment_method.verification_value
         end
+      end
+
+      def add_address(post, options)
+        if (address = options[:billing_address])
+          post[:billing] = {}
+          post[:billing][:firstName] = address[:first_name] if address[:first_name]
+          post[:billing][:lastName] = address[:last_name] if address[:last_name]
+          post[:billing][:companyName] = address[:company] if address[:company]
+          post[:billing][:phone] = address[:phone_number] if address[:phone_number]
+
+          post[:billing][:address] = {}
+          post[:billing][:address][:address] = address[:address1] if address[:address1]
+          post[:billing][:address][:city] = address[:city] if address[:city]
+          post[:billing][:address][:state] = address[:state] if address[:state]
+          post[:billing][:address][:country] = address[:country] if address[:country]
+          post[:billing][:address][:zip] = address[:zip] if address[:zip]
+        end
+      end
+
+      def get_state(address)
+        address[:state] && !address[:state].blank? ? address[:state] : 'NA'
       end
 
       def commit(action, params)
--- a/lib/active_merchant/billing/gateways/pay_junction_v2.rb
+++ b/lib/active_merchant/billing/gateways/pay_junction_v2.rb
@@ -129,10 +129,6 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def get_state(address)
-        address[:state] && !address[:state].blank? ? address[:state] : 'NA'
-      end
-
       def commit(action, params)
         response =
           begin
--- a/test/remote/gateways/remote_pay_junction_v2_test.rb
+++ b/test/remote/gateways/remote_pay_junction_v2_test.rb
@@ -5,9 +5,10 @@ class RemotePayJunctionV2Test < Test::Unit::TestCase
     @gateway = PayJunctionV2Gateway.new(fixtures(:pay_junction_v2))
 
     @amount = 99
-    @credit_card = credit_card('4444333322221111', month: 01, year: 2020, verification_value: 999)
+    @credit_card = credit_card('4444333322221111', month: 01, year: 2021, verification_value: 999)
     @options = {
-      order_id: generate_unique_id
+      order_id: generate_unique_id,
+      billing_address: address()
     }
   end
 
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/pay_junction_v2.rb
```diff
@@ -110,6 +112,27 @@ def add_payment_method(post, payment_method)
...
+          post[:billing][:address] = {}
+          post[:billing][:address][:address] = address[:address1] if address[:address1]
+          post[:billing][:address][:city] = address[:city] if address[:city]
+          post[:billing][:address][:state] = address[:state] if address[:state]
+          post[:billing][:address][:country] = address[:country] if address[:country]
+          post[:billing][:address][:zip] = address[:zip] if address[:zip]
+        end
+      end
+
+      def get_state(address)
```
fatcatt316: Hmm, is this method called anywhere?  I'm _guessing_ it should be called up on line 126.

test/remote/gateways/remote_pay_junction_v2_test.rb
```diff
@@ -5,9 +5,10 @@ def setup
     @gateway = PayJunctionV2Gateway.new(fixtures(:pay_junction_v2))
 
     @amount = 99
-    @credit_card = credit_card('4444333322221111', month: 01, year: 2020, verification_value: 999)
+    @credit_card = credit_card('4444333322221111', month: 01, year: 2021, verification_value: 999)
```
fatcatt316: Nice catch on this! What do you think about bumping it up even further into the future? 🔮

lib/active_merchant/billing/gateways/pay_junction_v2.rb
Carrigan: Not sure if you meant to use this or delete it, but it is currently not used

test/remote/gateways/remote_pay_junction_v2_test.rb
```diff
@@ -5,9 +5,10 @@ def setup
...
     @options = {
-      order_id: generate_unique_id
+      order_id: generate_unique_id,
+      billing_address: address()
```
fatcatt316: It's a wee more Rubyish if you want to leave off the `()` for this method call. I think Python requires `()` to be there for a function call, yeah?

test/unit/gateways/pay_junction_v2_test.rb
```diff
@@ -205,6 +206,21 @@ def test_failed_store
     assert_match %r{Card Number is not a valid card number}, response.message
   end
 
+  def test_add_address
+    post = {:card => {:billingAddress => {}}}
```
fatcatt316: Just a style preference, but since these keys are symbols, this line could be a bit shorter if you wanted:
```
post = {card: {billingAddress: {}}}
```

test/unit/gateways/pay_junction_v2_test.rb
```diff
@@ -205,6 +206,21 @@ def test_failed_store
     assert_match %r{Card Number is not a valid card number}, response.message
   end
 
+  def test_add_address
```
Carrigan: This is a pretty cool way to test this!

test/remote/gateways/remote_pay_junction_v2_test.rb
naashton: My thoughts on this are to keep it a year out so that some basic hygiene can be performed on the test data we are sending to the gateway. I'm not sure how often test cards might change, but it'll at least get whoever is working on it to go peek at the gateway's API and check for this stuff.

lib/active_merchant/billing/gateways/pay_junction_v2.rb
naashton: For the `X-PJ-Application-Key`, that is the PayJunction specific access key that is added to the headers. You'll find it in Spreedly's Onepass and you should just add it to your fixtures assigned to the `api_key`.

naashton: Good catch. I'm not sure I actually want that method. Was following Leila's Adyen example and I think that default to 'NA' is an Adyen gateway thing.

test/remote/gateways/remote_pay_junction_v2_test.rb
naashton: Yes, Python require's that, so this is definitely habit. But I did see this being used elsewhere in AM.

test/unit/gateways/pay_junction_v2_test.rb
naashton: This is another thing I copied from the Adyen gateway implementation. I'm not sure what is the convention across AM, but I'd prefer to do it however Spreedling's are doing it.

test/unit/gateways/pay_junction_v2_test.rb
naashton: Credit goes to Leila! I just copied they way she implemented this in the Adyen gateway :)

lib/active_merchant/billing/gateways/pay_junction_v2.rb
naashton: Thanks! Meant to delete it!

test/remote/gateways/remote_pay_junction_v2_test.rb
fatcatt316: Ah, yeah, I see it all over... welp shucks.

lib/active_merchant/billing/gateways/pay_junction_v2.rb
leila-alderman: Yes, confirming that Adyen is a bit finicky about requiring non-nil fields (e.g., "NA") to be sent for certain address information even if that information isn't available. Most gateways don't have those restrictions.

test/remote/gateways/remote_pay_junction_v2_test.rb
leila-alderman: @naashton The "official" test cards from PayJunction now have an expiration date of `01/2022`, and I think it's worth matching the values provided in their documentation. https://developer.payjunction.com/hc/en-us/articles/216477397-Development-Sandbox

test/unit/gateways/pay_junction_v2_test.rb
leila-alderman: This is most likely a case where older implementation are using older `=>` syntax and RuboCop simply hasn't caught up yet. I would support switching to the new `card:` hash syntax here.

leila-alderman:
Closing this PR since this change was committed in https://github.com/activemerchant/active_merchant/commit/728330b64d6873010e395af95a91a83c450631fe.

PR #3545, opened 2020-02-25
CheckoutV2: Use status as message for 3DS txns in progress

britth:
The response from Checkout direct 3DS requests often includes
the message "Unable to parse error message" because while the
transaction is pending, the existing fields we try to parse a
message from are not present. This PR updates the message_from
field to also check in the status field for a possible message
before using the default.

Remote:
32 tests, 77 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
27 tests, 120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


britth:
thanks @Carrigan !

PR #3546, opened 2020-02-25
Adyen: Fail unexpected 3DS responses

curiousepic:
There are certain situations where the initiator of a transaction is not
expecting or prepared to complete a 3DS challenge flow, yet an authorize
transaction triggers a 3DS auth response from Adyen.

Previously, if a 3DS flow was not expected, a Purchase was initiated,
and the auth step received a 3ds authentication response, the adapter
would assume the auth step succeeded, and perform the capture, for which
Adyen's response is not sufficient to indicate the money was actually
moved. This means the Purchase would be assumed to succeed when in fact
it was stalled at the 3DS auth step.

This change recognizes these situations and fails the transaction with a
meaningful message.

Remote:
82 tests, 276 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
59 tests, 279 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -471,7 +471,7 @@ module ActiveMerchant #:nodoc:
           raw_response = e.response.body
           response = parse(raw_response)
         end
-        success = success_from(action, response)
+        success = success_from(action, response, options)
         Response.new(
           success,
           message_from(action, response),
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -520,7 +520,12 @@ def request_headers(options)
         headers
       end
 
-      def success_from(action, response)
+      def success_from(action, response, options)
+        if response.dig('resultCode') == 'RedirectShopper' && !options[:execute_threed] && !options[:threed_dynamic]
```
britth: I know it's possible sometimes to have a direct challenge that skips the `RedirectShooper` result code and goes straight to `ChallengeShopper` - would that be possible in the dynamic 3ds scenario and do we need to check for that as well?

curiousepic: Great question, I'm not sure if it's possible but it seems like a good idea to add it to the criteria.

PR #3549, opened 2020-02-26
Ixopay: Support stored credentials

leila-alderman:
Implemented support for stored credentials on the Ixopay gateway.

Ixopay is slightly unusual in that it does not return a true
`network_transaction_id` and therefore this value cannot be sent on
subsequent transactions.

See [Ixopay's documentation on recurring transactions](https://gateway.ixopay.com/documentation/gateway#transaction-types)
as well as [an example recurring debit transaction](https://gateway.ixopay.com/documentation/direct-pci-enabled-api?xml#transactionwithcard-request-debit).

CE-161

Unit:
28 tests, 138 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
23 tests, 95 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

All unit tests:
4458 tests, 71557 assertions, 0 failures, 0 errors, 0 pendings, 2
omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ixopay.rb
+++ b/lib/active_merchant/billing/gateways/ixopay.rb
@@ -151,6 +151,7 @@ module ActiveMerchant #:nodoc:
           xml.currency    currency
           xml.description description
           xml.callbackUrl(options[:callback_url])
+          add_stored_credentials(xml, options)
         end
       end
 
@@ -169,6 +170,7 @@ module ActiveMerchant #:nodoc:
           xml.currency    currency
           xml.description description
           xml.callbackUrl callback_url
+          add_stored_credentials(xml, options)
         end
       end
 
@@ -249,6 +251,25 @@ module ActiveMerchant #:nodoc:
 
       def new_transaction_id
         SecureRandom.uuid
+      end
+
+      # Ixopay does not pass any parameters for cardholder/merchant initiated.
+      # Ixopay also doesn't support installment transactions, only recurring
+      # ("RECURRING") and unscheduled ("CARDONFILE").
+      #
+      # Furthermore, Ixopay is slightly unusual in its application of stored
+      # credentials in that the gateway does not return a true
+      # network_transaction_id that can be sent on subsequent transactions.
+      def add_stored_credentials(xml, options)
+        return unless stored_credential = options[:stored_credential]
+
+        if stored_credential[:initial_transaction]
+          xml.transactionIndicator 'INITIAL'
+        elsif stored_credential[:reason_type] == 'recurring'
+          xml.transactionIndicator 'RECURRING'
+        elsif stored_credential[:reason_type] == 'unscheduled'
+          xml.transactionIndicator 'CARDONFILE'
+        end
       end
 
       def add_extra_data(xml, extra_data)
--- a/test/remote/gateways/remote_ixopay_test.rb
+++ b/test/remote/gateways/remote_ixopay_test.rb
@@ -14,6 +14,7 @@ class RemoteIxopayTest < Test::Unit::TestCase
       email: 'test@example.com',
       description: 'Store Purchase',
       ip: '192.168.1.1',
+      stored_credential: stored_credential(:initial)
     }
 
     @extra_data = {extra_data: { customData1: 'some data', customData2: 'Can be anything really' }}
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
 * Rubocop corrections for space around operators [cdmackeyfree] #3543
 * Fat Zebra: Add `is_billing` in post for `store` call [chinhle23] #3551
 * SafeCharge: Adds four supported countries [carrigan] #3550
+* Ixopay: Support stored credentials [leila-alderman] #3549
 
 == Version 1.105.0 (Feb 20, 2020)
 * Credorax: Fix `3ds_transtype` setting in post [chinhle23] #3531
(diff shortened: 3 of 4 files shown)


lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -251,6 +253,29 @@ def new_transaction_id
...
+      # credentials in that the gateway does not return a true
+      # network_transaction_id that can be sent on subsequent transactions.
+
+      def add_stored_credentials(xml, options)
+        return unless stored_credential = options[:stored_credential]
+
+        if stored_credential[:initial_transaction]
+          xml.transactionIndicator 'INITIAL'
+          return
+        end
```
fatcatt316: Just another approach, but what do you think of something like this here?
```
if stored_credential[:initial_transaction]
  xml.transactionIndicator 'INITIAL'
elsif stored_credential[:reason_type] == 'recurring'
  xml.transactionIndicator 'RECURRING'
elsif stored_credential[:reason_type] == 'unscheduled'
  xml.transactionIndicator 'CARDONFILE'
end
```

lib/active_merchant/billing/gateways/ixopay.rb
```diff
@@ -251,6 +253,29 @@ def new_transaction_id
...
       end
 
+      # Ixopay does not pass any parameters for cardholder/merchant initiated.
+      # Ixopay also doesn't support installment transactions, only recurring
+      # ("RECURRING") and unscheduled ("CARDONFILE").
+      #
+      # Furthermore, Ixopay is slightly unusual in its application of stored
+      # credentials in that the gateway does not return a true
+      # network_transaction_id that can be sent on subsequent transactions.
+
```
fatcatt316: As always, I'm a fan of these helpful comments! One tiny nitpick is there probably doesn't need to be a blank space on this line, but that may be personal preference.

test/remote/gateways/remote_ixopay_test.rb
```diff
@@ -14,6 +14,7 @@ def setup
       email: 'test@example.com',
       description: 'Store Purchase',
       ip: '192.168.1.1',
+      stored_credential: stored_credential(:recurring)
```
fatcatt316: Let me know if I'm wrong, but it looks like this method sets `network_transaction_id`. I _think_ you said something in a comment that Ixopay doesn't return a `network_transaction_id`, though; is that right? And does it affect if this test should be used like this?

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: I wasn't sure what the convention was, but happy to get rid of it if that's preferred.

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: So my issue here (and the reason for the weird implementation) is that we can take in both `initial_transaction` and `reason_type` but that they both write to the same Ixopay field of `transactionIndicator`. I want to be sure that regardless of anything that the customer sends in for the `reason_type` that if `initial_transaction` is sent that that field will overwrite everything to be certain that we send `INITIAL`. Is there a cleaner way of ensuring that behavior?

test/remote/gateways/remote_ixopay_test.rb
leila-alderman: It looks to me like the helper method only supplies a `network_transaction_id` if `id` is specified: https://github.com/activemerchant/active_merchant/blob/master/test/test_helper.rb#L249

Am I reading that correctly?

fatcatt316: I think in this case, it'll end up setting `id` to `nil` on that line, and then assigning that value to `network_transaction_id` [here](https://github.com/activemerchant/active_merchant/blob/master/test/test_helper.rb#L252). Maybe that's okay? I'm honestly not sure, but figured it was worth pointing out.

leila-alderman: I think that a value of `nil` is fine. In truth, Core will send almost every GSF with a value of `nil` if a specific value isn't provided. We just don't want to send any sort of `network_transaction_id` to Ixopay or save a value for it within our vault (which is handled in Core).

fatcatt316: Ah, okie doke. Thanks for explaining that, then!

lib/active_merchant/billing/gateways/ixopay.rb
fatcatt316: Hmm, I think it'll do that in the code I wrote, if I understand right. I'm down to talk it over right quick if I'm missing something, though!

lib/active_merchant/billing/gateways/ixopay.rb
fatcatt316: It could easily just be my personal preference, and it's honestly not a big deal to me either way :)

So I'm fine if you leave it as-is here, too.

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: You're completely right! I've simplified this method now and updated [the unit tests](https://github.com/activemerchant/active_merchant/pull/3549/files#diff-332f420dc6b50a19dc1f41508e903506R271) to confirm that this field is getting passed correctly. Thanks for the suggestion!

lib/active_merchant/billing/gateways/ixopay.rb
leila-alderman: It's dead now. 💀

lib/active_merchant/billing/gateways/ixopay.rb
therufs: FWIW I like that this makes it a little easier to understand that it's the same field getting set in the XML regardless of what's put into it. :)

PR #3552, opened 2020-02-26
BlueSnap: Adds localized currency support

Carrigan:
## Why?

BlueSnap currently only has support for currencies with two decimals.

## What Changed

This commit adds localized support for 0-decimal currencies and 3-decimal currencies as noted in their documentation.

## Test Results

Unit Tests:

```
4456 tests, 71555 assertions, 0 failures, 0 errors, 0 pendings,
2 omissions, 0 notifications
100% passed
```

Bluesnap remote tests:

```
43 tests, 131 assertions, 0 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/test/remote/gateways/remote_blue_snap_test.rb
+++ b/test/remote/gateways/remote_blue_snap_test.rb
@@ -52,6 +52,20 @@ class RemoteBlueSnapTest < Test::Unit::TestCase
     assert_equal 'Success', response.message
   end
 
+  def test_successful_fractionless_currency_purchase
+    options = @options.merge(currency: 'JPY')
+    response = @gateway.purchase(12300, @credit_card, options)
+    assert_success response
+    assert_equal 'Success', response.message
+  end
+
+  def test_successful_three_decimal_currency_purchase
+    options = @options.merge(currency: 'BHD')
+    response = @gateway.purchase(1234, @credit_card, options)
+    assert_success response
+    assert_equal 'Success', response.message
+  end
+
   def test_successful_purchase_with_cabal_card
     options = @options.merge({
       email: 'joe@example.com'
--- a/test/unit/gateways/blue_snap_test.rb
+++ b/test/unit/gateways/blue_snap_test.rb
@@ -2,6 +2,16 @@
 
 require 'test_helper'
 
+class BlueSnapCurrencyDocMock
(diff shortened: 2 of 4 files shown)


test/unit/gateways/blue_snap_test.rb
```diff
@@ -334,8 +344,33 @@ def test_echeck_scrub
...
+    ActiveMerchant::Billing::BlueSnapGateway.currencies_without_fractions.each do |currency|
+      assert_equal '12', check_amount_registered(amount, currency)
+    end
+
+    # Check all 3 decimal currencies
+    ActiveMerchant::Billing::BlueSnapGateway.currencies_with_three_decimal_places.each do |currency|
+      assert_equal '1.234', check_amount_registered(amount, currency)
+    end
+  end
+
```
chinhle23: Nice!

test/remote/gateways/remote_blue_snap_test.rb
```diff
@@ -52,6 +52,20 @@ def test_successful_purchase
     assert_equal 'Success', response.message
   end
 
+  def test_successful_fractionless_currency_purchase
+    options = @options.merge(currency: 'BYR')
+    response = @gateway.purchase(1234, @credit_card, options)
+    assert_success response
+    assert_equal 'Success', response.message
+  end
```
chinhle23: I printed the response on this and got: `..."amount"=>"0.00", "usd-amount"=>"0.00", "currency"=>"USD"...`, so I bumped the amount to `123400` on this test and got `Transaction failed  because of payment processing failure.: 1000 - There are no E-Commerce merchant accounts configured to process Visa transactions with currency 974.` I'm guessing BlueSnap does not fully support the `BYR` currency? Instead of `BYR`, maybe we can use `JPY` (as it's a more widely used non-fractional currency)? We'll get a more consistent response from the gateway. Also, maybe we should test with an amount that ends in two zeros (since they'll be dropped off in the request anyway)?

Carrigan: Great catch! Thank you for running that locally. I changed it over to JPY and tested locally and it is now behaving as we would expect.

Carrigan:
Thanks @chinhle23!

PR #3554, opened 2020-02-27
Stripe Payment Intents: Updates for idempotency key support

britth:
Stripe Payment Intents supports the use of an idempotency key, but
that key can only be used against a single endpoint. Because the
Stripe PI flow can touch multiple different endpoints, we need to
ensure that it is slightly different across the different types
of requests to prevent errors. For instance, if the payment method
provided is a credit card, the adapter hits the payment method endpoint
to create that object before hitting the payment intent endpoint to
create the intent. This PR adds a method to add a suffix aligning to
request endpoints other than /payment_intents (e.g. a post to /payment_methods
would be <provided-key-pm> or a post to /payment_intents/intent-id/capture
would be <provided-key-capture>) to ensure that keys are unique but stable
across a request type. Also adds a remote test to ensure idempotency key
behavior remains as expected when a credit card payment type is used.

Unit:
9 tests, 55 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
35 tests, 167 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -45,7 +45,6 @@ module ActiveMerchant #:nodoc:
         CONFIRM_INTENT_ATTRIBUTES.each do |attribute|
           add_whitelisted_attribute(post, options, attribute)
         end
-
         commit(:post, "payment_intents/#{intent_id}/confirm", post, options)
       end
 
@@ -58,7 +57,7 @@ module ActiveMerchant #:nodoc:
         post[:card][:exp_year] = payment_method.year
         post[:card][:cvc] = payment_method.verification_value if payment_method.verification_value
         add_billing_address(post, options)
-
+        options = format_idempotency_key(options, 'pm')
         commit(:post, 'payment_methods', post, options)
       end
 
@@ -76,7 +75,6 @@ module ActiveMerchant #:nodoc:
         UPDATE_INTENT_ATTRIBUTES.each do |attribute|
           add_whitelisted_attribute(post, options, attribute)
         end
-
         commit(:post, "payment_intents/#{intent_id}", post, options)
       end
 
@@ -97,6 +95,7 @@ module ActiveMerchant #:nodoc:
           post[:transfer_data][:amount] = options[:transfer_amount]
         end
         post[:application_fee_amount] = options[:application_fee] if options[:application_fee]
+        options = format_idempotency_key(options, 'capture')
         commit(:post, "payment_intents/#{intent_id}/capture", post, options)
       end
 
@@ -129,9 +128,11 @@ module ActiveMerchant #:nodoc:
             post[:validate] = options[:validate] unless options[:validate].nil?
             post[:description] = options[:description] if options[:description]
             post[:email] = options[:email] if options[:email]
+            options = format_idempotency_key(options, 'customer')
             customer = commit(:post, 'customers', post, options)
             customer_id = customer.params['id']
           end
+          options = format_idempotency_key(options, 'attach')
           commit(:post, "payment_methods/#{params[:payment_method]}/attach", { customer: customer_id }, options)
         else
           super(payment_method, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -97,12 +96,14 @@ def capture(money, intent_id, options = {})
...
         end
         post[:application_fee_amount] = options[:application_fee] if options[:application_fee]
+        options = format_idempotency_key(options, 'capture')
         commit(:post, "payment_intents/#{intent_id}/capture", post, options)
       end
 
       def void(intent_id, options = {})
         post = {}
         post[:cancellation_reason] = options[:cancellation_reason] if ALLOWED_CANCELLATION_REASONS.include?(options[:cancellation_reason])
+        options = format_idempotency_key(options, 'cancel')
```
jknipp: 🤔I was planning to handle this within Core failure cleanup since `void` is usually an independent action. Should we limit the scope to actions that are compound in AM or Core?

britth: ah yes, that makes sense - i will update 👍

jknipp:
@britth How would you feel about merging my PR https://github.com/activemerchant/active_merchant/pull/3555 into this one?

britth:
yep, that makes sense - i'll merge it in, thanks!

jknipp:
Thanks!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -129,9 +128,11 @@ def store(payment_method, options = {})
             post[:validate] = options[:validate] unless options[:validate].nil?
             post[:description] = options[:description] if options[:description]
             post[:email] = options[:email] if options[:email]
+            options = format_idempotency_key(options, 'customer')
             customer = commit(:post, 'customers', post, options)
             customer_id = customer.params['id']
           end
+          options = format_idempotency_key(options, 'attach')
```
jknipp: Do we need to add a test for `store`?

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -418,6 +450,31 @@ def test_create_a_payment_intent_and_void
     assert_equal 'requested_by_customer', cancel_response.params['cancellation_reason']
   end
 
+  def test_create_a_payment_intent_and_void_requires_unique_idempotency_key
+    idempotency_key = SecureRandom.hex
```
fatcatt316: Tiniest nitpick, but it might be worth using only `SecureRandom.hex` or `SecureRandom.uuid` when setting `idempotency_key` in this test file. I'm guessing it doesn't matter, but a future developer might wonder why different methods were used. What do you think?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
britth: good call! test added

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -264,6 +265,12 @@ def add_shipping_address(post, options = {})
         post[:shipping][:tracking_number] = shipping[:tracking_number] if shipping[:tracking_number]
         post
       end
+
+      def format_idempotency_key(options, suffix='')
+        return options unless options[:idempotency_key]
+
+        return options.merge(idempotency_key: "#{options[:idempotency_key]}-#{suffix}")
```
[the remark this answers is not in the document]
jknipp: Yep, let's nix the last return.

britth:
@fatcatt316 all good ideas! updated 👍

test/remote/gateways/remote_stripe_payment_intents_test.rb
jknipp: The use of `SecureRandom.hex` originated from a similar Adyen test. Ultimately it doesn't matter because both are generating random characters. I do agree we should standardize within this test to avoid confusion.  `SecureRandom.hex` is used more heavily in AM fwiw.

PR #3559, opened 2020-03-02
Update BIN ranges for Alelo and Maestro cards

leila-alderman:
Updated the BIN ranges for the Alelo and Maestro cards to accurately
reflect ranges for those cards.

CE-363
CE-364
CE-438

All unit tests:
4479 tests, 71674 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -13,7 +13,12 @@ module ActiveMerchant #:nodoc:
         'diners_club'        => ->(num) { num =~ /^3(0[0-5]|[68]\d)\d{11}$/ },
         'jcb'                => ->(num) { num =~ /^35(28|29|[3-8]\d)\d{12}$/ },
         'dankort'            => ->(num) { num =~ /^5019\d{12}$/ },
-        'maestro'            => ->(num) { (12..19).cover?(num&.size) && in_bin_range?(num.slice(0, 6), MAESTRO_RANGES) },
+        'maestro'            => lambda { |num|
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -13,7 +13,12 @@ module CreditCardMethods
         'diners_club'        => ->(num) { num =~ /^3(0[0-5]|[68]\d)\d{11}$/ },
         'jcb'                => ->(num) { num =~ /^35(28|29|[3-8]\d)\d{12}$/ },
         'dankort'            => ->(num) { num =~ /^5019\d{12}$/ },
-        'maestro'            => ->(num) { (12..19).cover?(num&.size) && in_bin_range?(num.slice(0, 6), MAESTRO_RANGES) },
+        'maestro'            => lambda { |num|
```
therufs: Why is this a `lambda` instead of a `->` ?

fatcatt316: @therufs I wondered the same thing! If you look a little further down, there's another one using `lambda`, so I think it's just to continue that style.

leila-alderman: Just to illuminate this, I needed to switch over to using the `lambda` because I needed to pull from both `MAESTRO_RANGES` and the new `MAESTRO_BINS`. Since it's a more complicated comparison, the `->` was no longer sufficient. 

(And yep, copied the syntax from the `carnet` example already present.)

PR #3564, opened 2020-03-09
Decidir: Improving the response message when encountering errors

naashton:
Appending error type to error description to provide a better response
message when a transaction is declined or rejected.

Decidir also changed the error message returned which resulted in some
of the remote tests failing and included a new error code (91) that was
not in the STANDARD_ERROR_CODE_MAPPING array.

CE-444

Unit: 32 tests, 139 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 21 tests, 75 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -37,7 +37,7 @@ module ActiveMerchant #:nodoc:
         56 => STANDARD_ERROR_CODE[:card_declined],
         57 => STANDARD_ERROR_CODE[:card_declined],
         76 => STANDARD_ERROR_CODE[:call_issuer],
-        91 => STANDARD_ERROR_CODE[:invalid_number],
+        91 => STANDARD_ERROR_CODE[:call_issuer],
         96 => STANDARD_ERROR_CODE[:processing_error],
         97 => STANDARD_ERROR_CODE[:processing_error],
       }
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -242,7 +242,7 @@ module ActiveMerchant #:nodoc:
         message = nil
         if error = response.dig('status_details', 'error')
           message = error.dig('reason', 'description')
-          if detail = error.dig('type')
+          if detail = error['type']
             message += ' | ' + detail
           end
         elsif response['error_type']
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -37,6 +37,7 @@ module ActiveMerchant #:nodoc:
         56 => STANDARD_ERROR_CODE[:card_declined],
         57 => STANDARD_ERROR_CODE[:card_declined],
         76 => STANDARD_ERROR_CODE[:call_issuer],
+        91 => STANDARD_ERROR_CODE[:invalid_number],
         96 => STANDARD_ERROR_CODE[:processing_error],
         97 => STANDARD_ERROR_CODE[:processing_error],
       }
@@ -239,9 +240,11 @@ module ActiveMerchant #:nodoc:
         return response['message'] if response['message']
 
         message = nil
-
         if error = response.dig('status_details', 'error')
           message = error.dig('reason', 'description')
+          if detail = error.dig('type')
+            message += ' | ' + detail
+          end
         elsif response['error_type']
           message = response['validation_errors'].map { |errors| "#{errors['code']}: #{errors['param']}" }.join(', ') if response['validation_errors']
           message ||= response['error_type']
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -241,10 +241,7 @@ module ActiveMerchant #:nodoc:
 
         message = nil
         if error = response.dig('status_details', 'error')
-          message = error.dig('reason', 'description')
-          if detail = error['type']
-            message += ' | ' + detail
-          end
+          message = "#{error.dig('reason', 'description')} | #{error['type']}"
         elsif response['error_type']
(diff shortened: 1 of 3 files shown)


redbeardaz:
@naashton This looks great.

One thing I noticed, it seems like the `error_code` (even the standard error codes) are not getting pulled into Core.

Example: https://lava.spreedly.com/core_transactions/WxaIJTxsXHyq1iL9T5faUqjU15A

I would expect `error_code: 5`. Thoughts?

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -37,6 +37,7 @@ class DecidirGateway < Gateway
         56 => STANDARD_ERROR_CODE[:card_declined],
         57 => STANDARD_ERROR_CODE[:card_declined],
         76 => STANDARD_ERROR_CODE[:call_issuer],
+        91 => STANDARD_ERROR_CODE[:invalid_number],
```
fatcatt316: Nice catch noticing `91` wasn't handled! You've thought about this more than me, but is `invalid_number` the right message for this? My Spanish is rusty, but it looks like error 91 translates roughly to:

```
OUTLINE LINE ISSUER

Request telephone authorization, if approved, load the code obtained and leave the operation in
```

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -239,9 +240,11 @@ def message_from(success, response)
         return response['message'] if response['message']
 
         message = nil
-
         if error = response.dig('status_details', 'error')
           message = error.dig('reason', 'description')
+          if detail = error.dig('type')
```
fatcatt316: Nitpick, but since it's only one layer into the hash, I think this could just be `if detail = error['type']` if you wanted.

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -239,9 +240,11 @@ def message_from(success, response)
...
+            message += ' | ' + detail
```
fatcatt316: Just wondering, could the code get to this point if `error.dig('reason', 'description')` returned `nil`? If so, this line would blow up.

Would it be worth doing something like `message = [message, detail].compact.join(' | ')` to guard against that?

lib/active_merchant/billing/gateways/decidir.rb
naashton: I think this might be a part of a discussion in the Jira ticket that Branden and I were having about error_code's, using the int versus a standardized string: https://spreedly.atlassian.net/secure/RapidBoard.jspa?rapidView=52&projectKey=CE&modal=detail&selectedIssue=CE-378&assignee=5e26fb3aef603b0ca6f6f25a

Seems like there is inconsistent use across the board as to whether or not we return the error code as we get it from the gateway or try to map it to something Spreedly understands. On one hand, it makes sense that our response is standardized, but on the other, that `91` error code maps to something internally understood by Decidir and as you highlighted, it has a much more specific meaning and probably holds more value than the standardized value we are mapping it to.

That being said, I'm not sure if I have enough information as to why we standardize in some gateways and why we don't in others.

lib/active_merchant/billing/gateways/decidir.rb
naashton: Since the `if` statement is within the scope of `if error = response.dig('status_details', 'error')`, it would not get to this point if that is  `nil`. I imagine there is a cleaner way to do this, but without knowing a lot of tricks in ruby, I just took the easy route 😄

lib/active_merchant/billing/gateways/decidir.rb
fatcatt316: @naashton Ah okay. I was just wondering about this one since Decidir's error message didn't seem to match up to "invalid number". If it's part of another conversation I'll leave it alone here.

lib/active_merchant/billing/gateways/decidir.rb
fatcatt316: Ah no problem. I was saying if `message` was set to `nil` on line 244, which I don't think any `if` statements catch.

Another idea might be to just do `message = "#{message} | #{detail}", which'd handle if `message` was `nil` (I think that's kinda like f-strings in Python, if I remember that right 😄 )

naashton:
Made a couple changes. The remote tests keep failing with respect to the `response.mesage` we are expecting to get back. There doesn't appear to be any particular reason as to why the message keeps changing from one to another.

lib/active_merchant/billing/gateways/decidir.rb
fatcatt316: @naashton Still kinda wondering about this part. If the response had this kinda shape:
```
{"status_details"=>{"error"=>{"type"=>"typy"}}}
```
it'd end up raising an error on line 246, since `message` would be `nil` there.

What do you think about something like this (replacing lines 244-247)?
```
message = [
  error.dig('reason', 'description'),
  error['type']
].compact.join(' | ')
```
Just a thought. What think ye?

naashton: Ah, I see what you are saying now. I'll make this change.

naashton: Just pushed that change up. I'm using the fancy string format method: `message = "#{msg} | #{detail}"` 😄

naashton:
> Nice job hunting this one down!
> 
> I still have this question (https://github.com/activemerchant/active_merchant/pull/3564/files#r389922241), but it sounds like you and Branden are talking it over as a separate kind of thing, so I'm a-okay with this PR!

I actually tried to start a discussion in slack regarding this, here: https://spreedly.slack.com/archives/C024YVAN6/p1583846823184200

I know what you are asking is more of a mapping issue, but those Decidir error's seem a bit more specific than the standardized ones we are mapping to. Based on what David said, I think there is a bit of freedom for us engineer's to make the decision one way or the other, but I would prefer to get more insights from people with more experience before making major changes. 

That being said, I'm going to change it to `call_issuer` because that seems to be closer to what this error is asking for.

PR #3570, opened 2020-03-11
Stripe Payment Intents: Early return failed `payment_methods` response

chinhle23:
ECS-858

When a user creates a new Stripe account and fail to verify their phone
number before attempting transactions, they will get a `You cannot confirm
this PaymentIntent because it's missing a payment method. You can either
update the PaymentIntent with a payment method and then confirm it again,
or confirm it again directly with a payment method.` error message when
attempting to run any transactions with raw credit card numbers.

This change will return early when a `payment_methods` call fails and
provide a more useful error message, instead of continuing on to the
`payment_intents` call.

Unit:
10 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
36 tests, 174 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

All unit tests:
4446 tests, 71510 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -20,7 +20,9 @@ module ActiveMerchant #:nodoc:
         add_capture_method(post, options)
         add_confirmation_method(post, options)
         add_customer(post, options)
-        add_payment_method_token(post, payment_method, options)
+        payment_method = add_payment_method_token(post, payment_method, options)
+        return payment_method if payment_method.is_a?(ActiveMerchant::Billing::Response)
+
         add_metadata(post, options)
         add_return_url(post, options)
         add_connected_account(post, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -191,6 +199,8 @@ def add_payment_method_token(post, payment_method, options)
 
         if payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
           p = create_payment_method(payment_method, options)
+          return p if p.success? == false
```
britth: nitpick, but might do `return p unless p.success?` here

chinhle23: Cool, that does look better. Thanks again for the review, @britth

PR #3572, opened 2020-03-11
Borgun: support airline data

therufs:
Support passing airline data fields pursuant to https://docs.borgun.com/paymentgateways/bgateway/#sending-passenger-itinerary. 

Some fields are required; we don't validate those, but the adapter will pass generated error messages back to the user. Of particular note, the English documentation above says that Version is required, but requests that include Version as part of PassengerItineraryData are rejected. The documentation for the corresponding Icelandic implementation of the API does not appear to include Version (https://docs.borgun.is/formats/mci10/ section A1). 

CE-385

Remote: 21 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit: 9 tests, 41 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/borgun.rb
+++ b/lib/active_merchant/billing/gateways/borgun.rb
@@ -34,7 +34,7 @@ module ActiveMerchant #:nodoc:
         post[:TransType] = '5'
         add_invoice(post, money, options)
         add_payment_method(post, payment)
-        commit('authonly', post)
+        commit('authonly', post, options)
       end
 
       def capture(money, authorization, options={})
@@ -125,12 +125,12 @@ module ActiveMerchant #:nodoc:
         response
       end
 
-      def commit(action, post)
+      def commit(action, post, options={})
         post[:Version] = '1000'
         post[:Processor] = @options[:processor]
         post[:MerchantID] = @options[:merchant_id]
 
-        request = build_request(action, post)
+        request = build_request(action, post, options)
         raw = ssl_post(url(action), request, headers)
         pairs = parse(raw)
         success = success_from(pairs)
@@ -180,7 +180,7 @@ module ActiveMerchant #:nodoc:
         }
       end
 
-      def build_request(action, post)
+      def build_request(action, post, options={})
         mode = action == 'void' ? 'cancel' : 'get'
         xml = Builder::XmlMarkup.new indent: 18
         xml.instruct!(:xml, version: '1.0', encoding: 'utf-8')
@@ -188,9 +188,20 @@ module ActiveMerchant #:nodoc:
           post.each do |field, value|
             xml.tag!(field, value)
           end
+          build_airline_xml(xml, options[:passenger_itinerary_data]) if options[:passenger_itinerary_data]
         end
         inner = CGI.escapeHTML(xml.target!)
         envelope(mode).sub(/{{ :body }}/, inner)
+      end
+
+      def build_airline_xml(xml, airline_data)
+        xml.tag!('PassengerItineraryData') do
+          xml.tag!('A1') do
+            airline_data.each do |field, value|
+              xml.tag!(field, value)
+            end
+          end
+        end
       end
 
       def envelope(mode)
--- a/test/remote/gateways/remote_borgun_test.rb
+++ b/test/remote/gateways/remote_borgun_test.rb
@@ -54,6 +54,27 @@ class RemoteBorgunTest < Test::Unit::TestCase
     assert_success capture
   end
 
+  def test_successful_authorize_airline_data
+    passenger_itinerary_data = {
+      'MessageNumber' => '1111111',
+      'TrDate' => '20120222',
+      'TrTime' => '151515',
+      'PassengerName' => 'Jane Doe',
+      'ServiceClassCode_1' => '100',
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/borgun.rb
```diff
@@ -180,19 +180,30 @@ def headers
...
-      def build_request(action, post)
+      def build_request(action, post, options={})
         mode = action == 'void' ? 'cancel' : 'get'
         xml = Builder::XmlMarkup.new indent: 18
         xml.instruct!(:xml, version: '1.0', encoding: 'utf-8')
         xml.tag!("#{mode}Authorization") do
           post.each do |field, value|
             xml.tag!(field, value)
           end
+          build_airline_xml(xml, options[:passenger_itinerary_data]) if options[:passenger_itinerary_data]
```
therufs: This seems a little hinky, but refactoring how XML is generated felt out of scope.

lib/active_merchant/billing/gateways/borgun.rb
```diff
@@ -180,19 +180,30 @@ def headers
...
         end
         inner = CGI.escapeHTML(xml.target!)
         envelope(mode).sub(/{{ :body }}/, inner)
       end
 
+      def build_airline_xml(xml, airline_data)
+        xml.tag!('PassengerItineraryData') do
```
fatcatt316: One little less "hinky" way to do this might be to just call `xml.PassengerItineraryData` (I think the `tag` is only needed if the element name is a reserved word). For that reason, might be worth keeping the `tag!` on line 201 (since that element name comes from user input).

test/unit/gateways/borgun_test.rb
```diff
@@ -56,6 +56,23 @@ def test_authorize_and_capture
     assert_success capture
   end
 
+  def test_authorize_airline_data
+    # itinerary data abbreviated for brevity
```
fatcatt316: Maybe it's just me, but I don't worry too much about brevity in tests, if that was the only thing holding you back from adding more! That being said, this much is probably fine, too. What do you think?

therufs: If I were actually testing that each item was making it into the XML, I think it'd make sense to include all of them. However, since the API seems lmore of ike it doesn't follow a strict ordering (today in history!!) I didn't feel it was necessary to test the inclusions specifically. 

This might be an argument for eliminating even more of the included data! But I didn't want it to be totally unrecognizable either. 🤷‍♀

lib/active_merchant/billing/gateways/borgun.rb
therufs: So the hinkiness I had in mind in particular was having a random extra helper buried 3 methods away from where it actually gets called (because of the iterative `xml.tag!` in `build_request` that doesn't allow for nesting). 

I'd be happy to swap out the `tag!`s but am a little 🤔 about having two different functions that serve the same purpose; what do you think?

test/unit/gateways/borgun_test.rb
fatcatt316: Yep, I think it's fine here. Just didn't know if brevity was the _only_ reason it was being shortened. Huzzah for any-order XML!

lib/active_merchant/billing/gateways/borgun.rb
fatcatt316: Ah, I see what you're saying. I think it's fine to have one method.

lib/active_merchant/billing/gateways/borgun.rb
fatcatt316: I'm okay with this hinkiness 👍

test/remote/gateways/remote_borgun_test.rb
```diff
@@ -54,6 +54,27 @@ def test_successful_authorize_and_capture
...
+      'TravelDate_1' => '20120222',
+      'DepartureAirport_1' => 'KEF',
+      'CarrierCode_1' => 'CC',
+      'TravelAgencyCode' => 'A7654321',
+      'TravelAgencyName' => 'Spreedly Inc',
+      'TicketNumber' => '900.123.222'
+    }
+
+    options = @options.merge(passenger_itinerary_data: passenger_itinerary_data)
+    auth = @gateway.authorize(@amount, @credit_card, options)
```
fatcatt316: Is there anything else that'd be worth testing there? Does it respond different if itinerary data is passed in? If so, it might be worth checking for some of it here ([like was done here](https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_decidir_test.rb#L88-L90)). ¿What think ye?

therufs: No, as best I can tell the response is the same -- it doesn't echo any of the fields back or even acknowledge them (other than not erroring out when required fields are missing).

therufs: Also now I'm hungry for Heavenly Buffaloes, thanks for that :P

fatcatt316: Hahaha. It felt nice to get a little Durham into this repo :)

PR #3574, opened 2020-03-12
Acapture(OPP): Update gateway credentials

molbrown:
Acapture notified us that they will be deprecating the user_id/password
request authentication method in favor of an access_token header. This
change includes:
- Adding the access_token required credentials and removing the old
user_id and password fields. The entity_id field will continue to be
used.
- Updating request building logic around those changes, introducing the
Bearer token header.
- Updating the scrub method.

ECS-863

Remote:
100% passed

Unit:
14 tests, 73 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/opp.rb
+++ b/lib/active_merchant/billing/gateways/opp.rb
@@ -119,7 +119,7 @@ module ActiveMerchant #:nodoc:
       self.display_name = 'Open Payment Platform'
 
       def initialize(options={})
-        requires!(options, :user_id, :password, :entity_id)
+        requires!(options, :access_token, :entity_id)
         super
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/opp.rb
```diff
@@ -335,6 +335,13 @@ def commit(post, authorization, options)
         )
       end
 
+      def headers
+        {
```
chinhle23: Rubocop is telling me there's a trailing whitespace here.

molbrown: Updated 👍

PR #3577, opened 2020-03-18
Ingenico/GlobalCollect: 'requires_approval' support

fatcatt316:
## What changed?

Add support for optional "requiresApproval"
gateway-specific field for Ingenico/GlobalCollect gateway.

If `requires_approval` is sent in as the boolean `false`, then a purchase
will only make an `auth` call (and not a subsequent `capture` call).  The
'auth' call should request capture in this case.

If `requires_approval` is not sent, or is sent in as `true`,
then a purchase will possibly call both `auth` and `capture` like
before this change.

Also, a small Rubocop fix in a CenPOS test.

## Why?

There was a misunderstanding in #3571 .

Please see [Ingenico/GlobalCollect's documentation](https://epayments-api.developer-ingenico.com/s2sapi/v1/en_US/ruby/payments/create.html?paymentPlatform=GLOBALCOLLECT#payments-create-payload)

Note: you have to click on "Group: cardPaymentMethodSpecificInput"
in order to get to info about the "requiresApproval" field.

```
requiresApproval

Description
	•	true = the payment requires approval before the funds will be captured using the Approve payment or Capture payment API
	•	false = the payment does not require approval, and the funds will be captured automatically

```

CE-464

## Testing

Unit
```
ruby -Ilib:test test/unit/gateways/global_collect_test.rb

23 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

All local
```
rake test:local

4449 tests, 71521 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed
```

Remote
```
ruby -Ilib:test test/remote/gateways/remote_global_collect_test.rb

21 tests, 53 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```


test/unit/gateways/global_collect_test.rb
```diff
@@ -36,23 +36,22 @@ def test_successful_authorize_and_capture
...
+    call_count = 0
     stub_comms do
       @gateway.purchase(@accepted_amount, @credit_card, @options.merge(requires_approval: true))
     end.check_request do |endpoint, data, headers|
-      assert_equal true, JSON.parse(data)['cardPaymentMethodSpecificInput']['requiresApproval']
-    end.respond_with(successful_authorize_response)
+      call_count += 1
+      assert_equal(true, JSON.parse(data)['cardPaymentMethodSpecificInput']['requiresApproval']) if call_count == 1
+    end.respond_with(successful_authorize_response, successful_capture_response)
+    assert call_count == 2 # auth and capture
```
fatcatt316: Is there a less janky way to do this line and line 44? 🤔

test/remote/gateways/remote_global_collect_test.rb
```diff
@@ -55,18 +55,17 @@ def test_successful_purchase_with_more_options
     assert_equal 'Succeeded', response.message
   end
 
-  # When approval is required, purchase will not make a capture request
+  # When approval is required, `purchase` will also make a `capture` request
```
leila-alderman: Yay for comments! Could you perhaps add something in here that states that the `purchase` request will also send an explicit `capture` request to the gateway?

test/remote/gateways/remote_cenpos_test.rb
```diff
@@ -113,7 +113,7 @@ def test_failed_capture
     @gateway.capture(@amount, response.authorization)
     capture = @gateway.capture(@amount, response.authorization)
     assert_failure capture
-    assert_match /Duplicated.*transaction/, capture.message
+    assert_match(/Duplicated.*transaction/, capture.message)
```
leila-alderman: Thanks for fixing this RuboCop complaint!

test/remote/gateways/remote_global_collect_test.rb
```diff
@@ -55,18 +55,17 @@ def test_successful_purchase_with_more_options
...
     response = @gateway.purchase(@amount, @credit_card, options)
     assert_success response
     assert_equal 'Succeeded', response.message
-    assert_equal 'PENDING_APPROVAL', response.params['payment']['status']
-    assert_equal 'PENDING_MERCHANT', response.params['payment']['statusOutput']['statusCategory']
+    assert_equal 'CAPTURE_REQUESTED', response.params['payment']['status']
   end
 
-  # When approval is not required, purchase will make a capture request
+  # When approval is required, `purchase` will request capture via `auth`
```
leila-alderman: Shouldn't this still say "not required"?

Perhaps add some more clarification here saying that the `purchase` request will send an `auth` request and then perform an automatic capture that does not require an explicit `capture` request?

test/unit/gateways/global_collect_test.rb
leila-alderman: Just a suggestion, but I think the fact that it's returning both the `successful_authorize_response` and the `successful_capture_response` is sufficient proof that it's behaving the way it should even without explicitly checking for the `requiresApproval` field. It would certainly be less janky although a tad less thorough.

test/remote/gateways/remote_global_collect_test.rb
fatcatt316: Whoops, good catch on that comment. I've clarified that comment, feel free to have another looksie.

test/remote/gateways/remote_global_collect_test.rb
fatcatt316: @leila-alderman Yep! I updated this comment to be more explicit, lemme know your thoughts 🙇

test/remote/gateways/remote_global_collect_test.rb
leila-alderman: @fatcatt316 Hmm, I'm not seeing any changes yet. Maybe try pushing again?

fatcatt316: @leila-alderman 🤦‍♂ Pushed now.

test/remote/gateways/remote_global_collect_test.rb
```diff
@@ -55,18 +55,19 @@ def test_successful_purchase_with_more_options
     assert_equal 'Succeeded', response.message
   end
 
-  # When approval is required, purchase will not make a capture request
+  # When requires_approval is true (or not present),
+  # `purchase` will make both an `auth` and # a `capture` call
```
leila-alderman: Stray octothorpe:
```suggestion
  # `purchase` will make both an `auth` and a `capture` call
```

fatcatt316: 🦅 👁 I appreciate the thorough reviewa lot! Thanks, that pound sign is outta here.

karlitos83:
I'm trying to apply this feature adding requires_approval field to payment options, but I'm getting both process `authorize` and  `capture` with `"message": "Succeeded"`, `"statusCode": 800` and  `"status": "CAPTURE_REQUESTED"`. Regardless I use true or false.

I don't know if I'm introducing the Boolean field correctly. I'm sending purchase options with a structure like this in OpenStruct form.

```
  "options": {
    "requires_approval": false,
    "currency": "USD",
    "amount": 3500,
    "order_id": 100000014,
    "billing_address": {
      "country": "US"
    }
```
I would really appreciate a little help.

Thank you for develop this functionality.




``
```

PR #3579, opened 2020-03-19
Orbital: Fix stored credentials

chinhle23:
ECS-1076

https://secure.paymentech.com/developercenter/files/file?fid=J4p8Vagjmx4%3D
Subsequent cardholder initiated stored credential transactions are failing
due to the `get_msg_type` method looking for the `customer` parameter.

This change corrects the method to look for the `cardholder` parameter
instead in order to return the correct `MITMsgType` required by the gateway.

Unit:
91 tests, 527 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
34 tests, 188 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

All unit tests:
4459 tests, 71574 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -569,7 +569,7 @@ module ActiveMerchant #:nodoc:
 
         initiator =
           case parameters[:stored_credential][:initiator]
-          when 'customer' then 'C'
+          when 'cardholder', 'customer' then 'C'
           when 'merchant' then 'M'
           end
         reason =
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -569,7 +569,7 @@ def get_msg_type(parameters)
 
         initiator =
           case parameters[:stored_credential][:initiator]
-          when 'customer' then 'C'
+          when 'cardholder' then 'C'
```
britth: i don't know how likely it is, but in case someone is using 'customer' in this field, wonder if we should do:
`when 'cardholder', 'customer' then 'C'`
to keep it backwards compatible

curiousepic: I agree.

chinhle23: Great point. Thanks for the reviews @britth and @curiousepic! This has been updated.

PR #3580, opened 2020-03-20
Checkout V2: Purchase using Network Token

arbianchi:
Make Purchase call on Checkout V2 gateway using a network token. Add
`use_network_token` to the response to indicate NT was used for the
transaction

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -102,13 +102,21 @@ module ActiveMerchant #:nodoc:
 
       def add_payment_method(post, payment_method, options)
         post[:source] = {}
-        post[:source][:type] = 'card'
-        post[:source][:name] = payment_method.name
-        post[:source][:number] = payment_method.number
-        post[:source][:cvv] = payment_method.verification_value
+        if payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source == :network_token
+          post[:source][:type] = 'network_token'
+          post[:source][:token] = payment_method.number
+          post[:source][:token_type] = payment_method.brand == 'visa' ? 'vts' : 'mdes'
+          post[:source][:cryptogram] = payment_method.payment_cryptogram
+          post[:source][:eci] = options[:eci] || '05'
+        else
+          post[:source][:type] = 'card'
+          post[:source][:name] = payment_method.name
+          post[:source][:number] = payment_method.number
+          post[:source][:cvv] = payment_method.verification_value
+          post[:source][:stored] = 'true' if options[:card_on_file] == true
+        end
         post[:source][:expiry_year] = format(payment_method.year, :four_digits)
         post[:source][:expiry_month] = format(payment_method.month, :two_digits)
-        post[:source][:stored] = 'true' if options[:card_on_file] == true
       end
 
       def add_customer_data(post, options)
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -179,7 +188,8 @@ def response(action, succeeded, response)
           error_code: error_code_from(succeeded, response),
           test: test?,
           avs_result: avs_result,
-          cvv_result: cvv_result
+          cvv_result: cvv_result,
+          use_network_token: use_network_token
```
arbianchi: @tanyajajodia @shasum This indicates weather NT was used for the transaction. An open question is weather we also want to include a separate field that indicates whether NT _was supposed to be used_

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -216,6 +226,10 @@ def cvv_result(response)
         response['source'] && response['source']['cvv_check'] ? CVVResult.new(response['source']['cvv_check']) : nil
       end
 
+      def use_network_token(response)
+        response['source'] && response['source']['payment_account_reference'].present?
```
arbianchi: `payment_account_reference` is a response field unique to a network token source payment method.

lib/active_merchant/billing/gateways/checkout_v2.rb
shasum: @arbianchi Is this needed, when caller knows it passed in a `NetworkTokenizationCreditCard` as the payment method.

tanyajajodia: I'm not sure if need these attributes in the AM side of things. Probably just tracking them in Core?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -179,7 +187,7 @@ def response(action, succeeded, response)
           error_code: error_code_from(succeeded, response),
           test: test?,
           avs_result: avs_result,
-          cvv_result: cvv_result
+          cvv_result: cvv_result,
```
shasum: Nitpick: Redundant comma at the end?

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -183,14 +197,6 @@ def test_successful_authorize_and_capture_with_3ds2
     assert_success capture
   end
 
-  def test_direct_3ds_authorize
-    auth = @gateway.authorize(@amount, @threeds_card, @options.merge(execute_threed: true))
-
-    assert_equal 'Pending', auth.message
-    assert_equal 'Y', auth.params['3ds']['enrolled']
-    assert auth.params['_links']['redirect']
-  end
```
shasum: Is this something you found no longer relevant to what the gateway supported?

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -58,6 +65,13 @@ def test_successful_purchase
     assert_equal 'Succeeded', response.message
   end
 
+  def test_successful_purchase_with_network_token
```
shasum: Nice! Could you also add one to the non-remote unit tests for consistency sake. The stubbed responses there sometimes act as a useful reference.

arbianchi:
Thanks @shasum. Could you take a quick look at the unit test I added please?

britth:
@arbianchi could you also [add a changelog entry](https://github.com/activemerchant/active_merchant/blob/master/CHANGELOG) for this update - thanks!

PR #3581, opened 2020-03-21
CyberSource: Fix `void` for `purchase` transactions

chinhle23:
ECS-1080

Voiding `purchase` transactions were failing due to `<ccAuthReversalService run=\"true\">`
being sent in the request instead of the required `<voidService run=\"true\">`.

This change corrects this issue.

Unit:
76 tests, 340 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
75 tests, 385 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.3333% passed

All unit tests:
4460 tests, 71572 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

The 5 remote failures seem to be previously existing and unrelated:

 - test_successful_3ds_validate_authorize_request
 - test_successful_3ds_validate_purchase_request
 - test_successful_pinless_debit_card_puchase
 - test_successful_tax_calculation
 - test_successful_validate_pinless_debit_card

### DIFF
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -451,22 +451,34 @@ class CyberSourceTest < Test::Unit::TestCase
     end.respond_with(successful_card_credit_response)
   end
 
+  def test_successful_void_purchase_request
+    purchase = '1000;1842651133440156177166;AP4JY+Or4xRonEAOERAyMzQzOTEzMEM0MFZaNUZCBgDH3fgJ8AEGAMfd+AnwAwzRpAAA7RT/;purchase;100;USD;'
+
+    stub_comms do
+      @gateway.void(purchase, @options)
+    end.check_request do |endpoint, data, headers|
+      assert_match(%r(<voidService run=\"true\"), data)
+    end.respond_with(successful_void_response)
+  end
+
   def test_successful_void_capture_request
-    @gateway.stubs(:ssl_post).returns(successful_capture_response, successful_auth_reversal_response)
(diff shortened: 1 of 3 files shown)


test/unit/gateways/cyber_source_test.rb
```diff
@@ -451,22 +451,34 @@ def test_credit_includes_mdd_fields
     end.respond_with(successful_card_credit_response)
   end
 
+  def test_successful_void_purchase_request
+    authorization = '1000;1842651133440156177166;AP4JY+Or4xRonEAOERAyMzQzOTEzMEM0MFZaNUZCBgDH3fgJ8AEGAMfd+AnwAwzRpAAA7RT/;purchase;100;USD;'
```
Carrigan: Do you think that you could change this to `purchase` to reflect what type of action it is? Same for the capture on line 465.

test/unit/gateways/cyber_source_test.rb
```diff
@@ -451,22 +451,34 @@ def test_credit_includes_mdd_fields
...
+
+    stub_comms do
+      @gateway.void(authorization, @options)
+    end.check_request do |endpoint, data, headers|
+      assert_match(%r(<voidService run=\"true\"), data)
```
Carrigan: Thank you for upgrading these tests to check the values on these!

test/unit/gateways/cyber_source_test.rb
chinhle23: Yeah, that makes more sense. Thanks for the review @Carrigan!

PR #3585, opened 2020-03-30
Optimal Payment: Add store method

britth:
Add support for store on Optimal Payment, which calls verify with
the cc and any options passed. You can use the authorization value
returned from the successful transaction to issue a purchase or auth,
which will utilize that same payment method.

Remote:
18 tests, 85 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
19 tests, 75 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_optimal_payment_test.rb
+++ b/test/remote/gateways/remote_optimal_payment_test.rb
@@ -7,6 +7,7 @@ class RemoteOptimalPaymentTest < Test::Unit::TestCase
     @amount = 100
     @declined_amount = 5
     @credit_card = credit_card('4387751111011')
+    @expired_card = credit_card('4387751111011', month: 12, year: 2019)
 
     @options = {
       order_id: '1',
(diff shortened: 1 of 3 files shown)


curiousepic:
Weird. This makes me wonder if there's some option we should be setting for normal Verify calls to prevent the data being stored.

curiousepic:
Oh, I would add failed case tests as well, if it makes sense.

test/remote/gateways/remote_optimal_payment_test.rb
```diff
@@ -57,6 +58,52 @@ def test_successful_verify
...
+    response = @gateway.store(@invalid_card, @options)
+    assert_failure response
+    assert_not_nil response.authorization
+    assert_equal 'ERROR', response.params['decision']
+
+
+    assert stored_purchase = @gateway.stored_purchase(@amount, response.authorization)
+    assert_failure stored_purchase
+    assert_equal 'ERROR', stored_purchase.params['decision']
+  end
```
britth: @curiousepic how do these last two seem for failed test cases? I don't see a specific invalid test card, so just created a card with a blank PAN for these two

curiousepic: 👍

PR #3586, opened 2020-04-01
Update SecurePay Australia test URL

bbraschi:


jimmijazz:
Message from securepay:

> 
> When using the test Merchant ID credentials in test mode (suggested by SecurePay) and placing the order with the test credit card, we get the error message "Invalid Response"
> I have checked the test environment for your account and cannot see any transaction activity from shopify. This could be caused by an incorrect test URL in the plugin. Please contact the developer of the plugin to ensure the correct test URL is being used.
> XML - https://test.api.securepay.com.au/xmlapi/payment
> DP- https://test.api.securepay.com.au/directpost/authorise
> SF- https://test.payment.securepay.com.au/secureframe/invoice

This came through their regular support channels (Nas was agent we were speaking with) support@securepay.com.au

Will get better contact and context this week and add here.

PR #3587, opened 2020-04-01
Iridium: Localize zero-decimal currencies

chinhle23:
ECS-1119

Iridium (PayVector) requires the transaction amount to be sent in "minor
currency". This change ensures that zero-decimal currencies are properly
sent (i.e. 14200 JPY is sent as 142)

A new remote test was not written because Iridium (PayVector) does not
return the amount in the response.

Unit:
13 tests, 56 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
21 tests, 84 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

All unit tests:
4462 tests, 71592 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/iridium.rb
+++ b/lib/active_merchant/billing/gateways/iridium.rb
@@ -291,7 +291,8 @@ module ActiveMerchant #:nodoc:
         order_id, cross_reference, _ = authorization.split(';')
         build_request(options) do |xml|
           if money
-            details = {'CurrencyCode' => currency_code(options[:currency] || default_currency), 'Amount' => amount(money)}
+            currency = options[:currency] || currency(money)
+            details = {'CurrencyCode' => currency_code(currency), 'Amount' => localized_amount(money, currency)}
           else
             details = {'CurrencyCode' => currency_code(default_currency), 'Amount' => '0'}
           end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/iridium.rb
```diff
@@ -291,7 +291,8 @@ def build_reference_request(type, money, authorization, options)
         order_id, cross_reference, _ = authorization.split(';')
         build_request(options) do |xml|
           if money
-            details = {'CurrencyCode' => currency_code(options[:currency] || default_currency), 'Amount' => amount(money)}
+            currency = options[:currency] || currency(money)
+            details = {'CurrencyCode' => currency_code(options[:currency] || default_currency), 'Amount' => localized_amount(money, currency)}
```
molbrown, line 295: Would it make sense to use the `currency` variable in the currency_code method?

chinhle23: Thanks @molbrown, that does make more sense. This has been updated.

molbrown: 👍

PR #3591, opened 2020-04-07
Kushki: support preAuthorize and capture

therufs:
Some notes inline. :)

Unit: 15 tests, 98 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 13 tests, 42 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

CE-463

### DIFF
--- a/lib/active_merchant/billing/gateways/kushki.rb
+++ b/lib/active_merchant/billing/gateways/kushki.rb
@@ -4,8 +4,8 @@ module ActiveMerchant #:nodoc:
       self.display_name = 'Kushki'
       self.homepage_url = 'https://www.kushkipagos.com'
 
-      self.test_url = 'https://api-uat.kushkipagos.com/v1/'
-      self.live_url = 'https://api.kushkipagos.com/v1/'
+      self.test_url = 'https://api-uat.kushkipagos.com/'
+      self.live_url = 'https://api.kushkipagos.com/'
 
       self.supported_countries = ['CL', 'CO', 'EC', 'MX', 'PE']
       self.default_currency = 'USD'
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -4,8 +4,8 @@ class KushkiGateway < Gateway
       self.display_name = 'Kushki'
       self.homepage_url = 'https://www.kushkipagos.com'
 
-      self.test_url = 'https://api-uat.kushkipagos.com/v1/'
-      self.live_url = 'https://api.kushkipagos.com/v1/'
+      self.test_url = 'https://api-uat.kushkipagos.com/'
+      self.live_url = 'https://api.kushkipagos.com/'
```
therufs, line 8: Per Kushki (https://docs.kushkipagos.com/api/card/requestcardtoken), some API calls go to `card/v1` whereas some just go to `v1`. I refactored this while attempting to fix something that turned out to be unrelated, but left the changes to more closely adhere to Kushki's docs.

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -24,6 +24,23 @@ def purchase(amount, payment_method, options={})
         end
       end
 
+      def authorize(amount, payment_method, options={})
+        MultiResponse.run() do |r|
+          r.process { tokenize(amount, payment_method, options) }
```
therufs, line 29: Other than tokenization, Kushki only operates on tokens, not directly on payment methods; since this integration doesn't currently support the use of tokens, we tokenize a payment method each time.

test/remote/gateways/remote_kushki_test.rb
```diff
@@ -51,6 +51,50 @@ def test_failed_purchase
     assert_equal 'Monto de la transacción es diferente al monto de la venta inicial', response.message
   end
 
+  def test_successful_authorize
+    # Kushki only allows preauthorization for PEN, CLP, and UF.
```
therufs, line 55: This doesn't seem to actually be true in practice; the `test_successful_capture` runs with the default USD currency and everything goes fine. 🤷‍♀

test/remote/gateways/remote_kushki_test.rb
```diff
@@ -81,15 +125,15 @@ def test_successful_void
...
+    assert_equal 'Cuerpo de la petición inválido.', response.message
   end
 
   def test_invalid_login
     gateway = KushkiGateway.new(public_merchant_id: '', private_merchant_id: '')
 
     response = gateway.purchase(@amount, @credit_card)
     assert_failure response
-    assert_match %r{ID de comercio no válido}, response.message
+    assert_match %r{Unauthorized}, response.message
```
therufs, line 136: Couple of pre-existing failing tests here that had changed output.

lib/active_merchant/billing/gateways/kushki.rb
naashton: That's only slightly annoying.

PR #3592, opened 2020-04-08
Added support for MerchantInformation CyberSource-specific fields

apfranzen:
Added support for MerchantInformation CyberSource-specific fields. The CyberSource XML schema has also been updated. Per the gateway's instructions, we are now using version 1.164.

CE-189

77 tests, 345 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

test/remote/gateways/remote_cyber_source_test.rb contains 5 pre-existing, unrelated test failures:
- test_successful_3ds_validate_authorize_request
- test_successful_3ds_validate_purchase_request
- test_successful_pinless_debit_card_puchase
- test_successful_tax_calculation
- test_successful_validate_pinless_debit_card

76 tests, 389 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.4211% passed


test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -370,6 +370,21 @@ def test_successful_capture_with_mdd_fields
...
+        merchantDescriptor: {
+          name: 'Test Name',
+          address1: '123 Main Dr',
+          locality: 'Durham',
+        }
+      }
+    }
+
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(merchant_options))
+    assert_successful_response(response)
```
apfranzen, line 385: The fields added in this PR: `name`, `address1` and `locality` are not returned from CyberSource with a successful transaction from what I read, but please weigh in if there are other cases that should be tested for.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -25,8 +25,8 @@ class CyberSourceGateway < Gateway
       self.live_url = 'https://ics2wsa.ic3.com/commerce/1.x/transactionProcessor'
 
       # Schema files can be found here: https://ics2ws.ic3.com/commerce/1.x/transactionProcessor/
-      TEST_XSD_VERSION = '1.159'
-      PRODUCTION_XSD_VERSION = '1.159'
+      TEST_XSD_VERSION = '1.164'
+      PRODUCTION_XSD_VERSION = '1.164'
```
apfranzen, line 29: The work related to this PR doesn't require an upgrade of the the XSD version, but I saw there was a later version available. Please weigh in if you have a sense of if this should be left as-is, or if upgrading is helpful.

test/remote/gateways/remote_cyber_source_test.rb
therufs: It seems pretty common for gateways to not return information about passed fields; I don't think you need to worry about it.

lib/active_merchant/billing/gateways/cyber_source.rb
therufs: Upgrading is good 👍

therufs: Occurs to me it might be good to get rid of the old XSD, just for tidiness.

PR #3593, opened 2020-04-09
ePay: Send unique order ids for remote tests

curiousepic:
A number of remote tests were failing due to the gateway rejecting non-
unique order ids.

Remote:
15 tests, 27 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
16 tests, 55 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_epay_test.rb
+++ b/test/remote/gateways/remote_epay_test.rb
@@ -8,8 +8,8 @@ class RemoteEpayTest < Test::Unit::TestCase
     @credit_card = credit_card('3333333333333000')
     @credit_card_declined = credit_card('3333333333333102')
     @amount = 100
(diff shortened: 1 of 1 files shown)


chinhle23:
I see that some tests were failing with the `Order number already exists.` error on master, but while testing on this branch the following tests fail with error message: `ePay: -1003 PBS: -1`
`test_successful_authorize_and_capture_ds_transaction_id`
`test_successful_authorize_and_capture_xid`
`test_successful_refund_ds_transaction_id`
`test_successful_refund_xid`

...and these tests fail with error message: `ePay: -1003`
`test_successful_void_ds_transaction_id`
`test_successful_void_xid`

Are there additional requirements for my environment?

curiousepic:
@chinhle23 make sure a password isn't specified with your fixtures.

chinhle23:
Hmm...the password isn't present and I'm still getting those failures.

PR #3594, opened 2020-04-09
Decidir: Handle unique error response

naashton:
Decidir, in some odd cases, is returning a json response that was not
being caught by the `error_code_from` method

CE-378

Unit: 33 tests, 143 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 21 tests, 75 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -266,6 +266,11 @@ module ActiveMerchant #:nodoc:
           error_code ||= error['type']
         elsif response['error_type']
           error_code = response['error_type'] if response['validation_errors']
+        elsif error = response.dig('error')
+          validation_errors = error.dig('validation_errors', 0)
+          code = validation_errors['code'] if validation_errors && validation_errors['code']
+          param = validation_errors['param'] if validation_errors && validation_errors['param']
+          error_code = "#{error['error_type']} | #{code} | #{param}" if error['error_type']
         end
 
         error_code || STANDARD_ERROR_CODE[:processing_error]
--- a/lib/active_merchant/billing/gateways/kushki.rb
+++ b/lib/active_merchant/billing/gateways/kushki.rb
@@ -209,7 +209,6 @@ module ActiveMerchant #:nodoc:
         if ['void', 'refund'].include?(action)
           base_url + 'v1/' + ENDPOINT[action] + '/' + params[:ticketNumber].to_s
         else
-          #require 'pry-byebug'; byebug
           base_url + 'card/v1/' + ENDPOINT[action]
         end
       end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * Ixopay: Properly support three-decimal currencies [chinhle23] #3589
 * Kushki: support `auth` and `capture` [therufs] #3591
 * PaymentExpress: Update references to Windcave to reflect rebranding [britth] #3595
+* Decidir: Improve handling of error responses from the gateway [naashton] #3594
 
 == Version 1.107.1 (Apr 1, 2020)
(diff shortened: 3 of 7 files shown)


naashton:
There was some cleanup to please rubocop as well

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -209,7 +209,6 @@ def url(action, params)
         if ['void', 'refund'].include?(action)
           base_url + 'v1/' + ENDPOINT[action] + '/' + params[:ticketNumber].to_s
         else
-          #require 'pry-byebug'; byebug
```
therufs, line 212: _oops_

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -266,6 +266,11 @@ def error_code_from(response)
           error_code ||= error['type']
         elsif response['error_type']
           error_code = response['error_type'] if response['validation_errors']
+        elsif error = response.dig('error')
+          validation_errors = error.dig('validation_errors', 0)
```
therufs, line 270: Do we need to provide 0 here? Wouldn't `validation_errors` be set to nil otherwise?

lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -266,6 +266,11 @@ def error_code_from(response)
...
+          code = validation_errors['code'] if validation_errors && validation_errors['code']
+          param = validation_errors['param'] if validation_errors && validation_errors['param']
+          error_code = "#{error['error_type']} | #{code} | #{param}" if error['error_type']
```
therufs, line 273: It looks like we won't return information about the code or the param if there's no `'error_type'` -- it's possible that this is what we want, but I wanted to double check.

therufs:
Sorry and thanks about the rubocop business 😬

therufs:
Also sorry about the accidental close, yikes

lib/active_merchant/billing/gateways/decidir.rb
naashton: So, the response that I'm working from is this:
`{\"error\":{\"error_type\":\"invalid_request_error\",\"validation_errors\":[{\"code\":\"invalid_param\",\"param\":\"payment_type\"}]}}`
And that ends up being a nested array. I'm only grabbing it at index 0 because I'm not sure what a response with additional indexes would look like.

lib/active_merchant/billing/gateways/decidir.rb
naashton: Hard to say, being that this is the only example of this type of response from Decidir, but I think we want to just resort to the last part which will return the `processor_error` incase the `error_code` is `nil`. Since this is such a unique case, I think the majority of error's will be handled by the first 2 conditionals.

naashton: The other way I was thinking of doing it, was to just grab that `error_type` and `validation_errors`, which would have ended up looking a little sloppy: 

`invalid_request_error | [{"code"=>"invalid_param", "param"=>"payment_type"}]`

I feel like that might end up being more difficult for the users to parse, downstream.

lib/active_merchant/billing/gateways/decidir.rb
therufs: Ohhhh I misunderstood what was going on here. Seems sensible 👍

lib/active_merchant/billing/gateways/decidir.rb
therufs: I'm okay with leaving it how it is -- I'm sure someone will let us know if they're still not getting useful info 😅

naashton:
Thanks!

PR #3597, opened 2020-04-10
Add google pay to hps gateway

MSmedal:


olleolleolle:
Most of the tests keep on following the established pattern, so that repetition is of no concern.

Perhaps the PR needs a Description - otherwise it seems complete.

Two questions that the Description could answer: "What does it add? How does it teach that to the user?"

PR #3603, opened 2020-04-22
EBANX: Fix transaction amount for verify transaction

miguelxpn:


chinhle23:
Hi @miguelxpn, can you explain why this change is needed? I changed the Chile verify amount to 100 and the `test_successful_verify_for_chile` test still works. Does verify work differently in production than it does on the sandbox?

miguelxpn:
When using local currencies we end up doing the verify using the local currency, and in those cases 100 may be less than the minimum value for a valid transaction in some countries.

PR #3605, opened 2020-04-29
Forte: Use underscore for unused argument in test

wsmoak:
Working with the Forte gateway in a fork, I received this error from Codacy:

```
test/unit/gateways/forte_test.rb
Unused block argument - `headers`. If it's necessary, use `_` or `_headers` as an argument name to indicate that it won't be used.
end.check_request do |type, url, parameters, headers|
```

This will mark the `headers` argument as unused so that the quality review will pass.


wsmoak:
Okay, apparently y'all don't use Codacy here on master.  Someone must have configured it for our fork and I misunderstood what I was seeing.

Feel free to close this if you don't want to merge it!

PR #3620, opened 2020-05-06
Cybersource additions

pi3r:
1. Send a specific card brand commerceIndicator for 3DS to comply with following doc

image

2. Pass application_id as partnerSolutionID

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -471,6 +471,8 @@ module ActiveMerchant #:nodoc:
         xml.tag! 'clientLibrary', 'Ruby Active Merchant'
         xml.tag! 'clientLibraryVersion', VERSION
         xml.tag! 'clientEnvironment', RUBY_PLATFORM
+        xml.tag!('partnerSolutionID', application_id) if application_id
+
         add_merchant_descriptor(xml, options)
       end
 
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -27,6 +27,13 @@ module ActiveMerchant #:nodoc:
       # Schema files can be found here: https://ics2ws.ic3.com/commerce/1.x/transactionProcessor/
       TEST_XSD_VERSION = '1.164'
       PRODUCTION_XSD_VERSION = '1.164'
+      ECI_BRAND_MAPPING = {
+        visa: 'vbv',
+        master: 'spa',
+        american_express: 'aesk',
+        jcb: 'js',
+        discover: 'pb',
+      }.freeze
 
       self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb, :dankort, :maestro, :elo]
       self.supported_countries = %w(US BR CA CN DK FI FR DE IN JP MX NO SE GB SG LB PK)
@@ -598,7 +605,7 @@ module ActiveMerchant #:nodoc:
         xml.tag!('cavvAlgorithm', threeds_2_options[:cavv_algorithm]) if threeds_2_options[:cavv_algorithm]
         xml.tag!('paSpecificationVersion', threeds_2_options[:version]) if threeds_2_options[:version]
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
-        xml.tag!('commerceIndicator', options[:commerce_indicator]) if options[:commerce_indicator]
+        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
         xml.tag!('eciRaw', threeds_2_options[:eci]) if threeds_2_options[:eci]
         xml.tag!('xid', threeds_2_options[:xid]) if threeds_2_options[:xid]
         xml.tag!('veresEnrolled', threeds_2_options[:enrolled]) if threeds_2_options[:enrolled]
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -27,6 +27,13 @@ class CyberSourceGateway < Gateway
       # Schema files can be found here: https://ics2ws.ic3.com/commerce/1.x/transactionProcessor/
       TEST_XSD_VERSION = '1.164'
       PRODUCTION_XSD_VERSION = '1.164'
+      ECI_BRAND_MAPPING = {
```
bbraschi, line 30: There are more values in [app_commerce_indicators](https://developer.cybersource.com/library/documentation/dev_guides/CC_Svcs_SO_API/html/Topics/app_commerce_indicators.htm) than in `ECI_BRAND_MAPPING`. Is it because you are only mapping the non 3-DS cases (ie `vbv` but not `vbv_attempted` or `vbv_failure`), because the `eci` is passed in `options[:three_d_secure]` in the 3DS cases?

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -598,7 +607,7 @@ def add_normalized_threeds_2_data(xml, payment_method, options)
         xml.tag!('cavvAlgorithm', threeds_2_options[:cavv_algorithm]) if threeds_2_options[:cavv_algorithm]
         xml.tag!('paSpecificationVersion', threeds_2_options[:version]) if threeds_2_options[:version]
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
-        xml.tag!('commerceIndicator', options[:commerce_indicator]) if options[:commerce_indicator]
+        xml.tag!('commerceIndicator', ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
```
bbraschi, line 610: _CyberSource: Pass commerce indicator if present_ #3350 made it possible for the caller to inject `options[:commerce_indicator]`and your PR removes that. Is it intentional? Would it make sense to give precedence to `options[:commerce_indicator]` if it is present?

lib/active_merchant/billing/gateways/cyber_source.rb
nebdil: it looks like cybersource has `eciRaw` and `commerceIndicator` as separate variables and they treat `eciRaw` as 3DS specific?

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -464,6 +471,8 @@ def add_merchant_data(xml, options)
         xml.tag! 'clientLibrary', 'Ruby Active Merchant'
         xml.tag! 'clientLibraryVersion', VERSION
         xml.tag! 'clientEnvironment', RUBY_PLATFORM
+        xml.tag!('partnerSolutionID', application_id) if application_id
```
nebdil, line 474: what is the `application_id`?

pi3r: fyi https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L141

nebdil: thank you!

lib/active_merchant/billing/gateways/cyber_source.rb
pi3r: Yes, because we don't send any 3ds value in case of attempt or failure, do we?

lib/active_merchant/billing/gateways/cyber_source.rb
pi3r: Thanks for catching that, I addressed this, could you review again please?

lib/active_merchant/billing/gateways/cyber_source.rb
bbraschi: Shopify does not attempt the payment if the buyer failed the challenge. In the other cases, Shopify provides 3DS data in case the 3DS authentication was attempted, even if it errorred.

pi3r: How could we know that the authentication was attempted and not successful? Are there anything from the three_d_secure option that would hint to that?

bbraschi: For example, in the Cardinal [Test Case 13: Error on Authentication](https://cardinaldocs.atlassian.net/wiki/spaces/CCen/pages/903577725/EMV+3DS+2.0+Test+Cases#EMV3DS2.0TestCases-TestCase13:ErroronAuthentication)


ECI Value 07 | Visa, American Express, Discover (Diners Club), CB Visa, ELO
-- | --
ECI Value 00 | Mastercard, CB Mastercard

pi3r: Should I append an `attempted` to the commerceIndicator when the eci is 00 or 07 then? Seems like it would make sense. What do you think?

bbraschi: I do not know without reading cybersource doc more.
We already provide the `eci`
```
```
so I do not understand yet why we have to also provide `commerceIndicator`

pi3r: Yeah... me neither... anyway, I'll merge that for now and talk with them a bit more, this is still a good change to merge.

curiousepic: If it helps, `commerceIndicator` is also utilized for specifying certain COF transaction types, i.e. `recurring` `install` etc.

PR #3623, opened 2020-05-11
Element: Add card_present_code field

schwarzgeist:
CE-376, CE-584

Remote:
20 tests, 57 assertions, 1 failures, 0 errors,
95% passed

Unit:
20 tests, 53 assertions, 0 failures, 0 errors,
100% passed

Remote test failure seems to be pending an update with an updated
value containing an actual CardPresentCode value. This
modification to remote unit tests may need to be removed.

### DIFF
--- a/lib/active_merchant/billing/gateways/element.rb
+++ b/lib/active_merchant/billing/gateways/element.rb
@@ -194,7 +194,7 @@ module ActiveMerchant #:nodoc:
       def add_terminal(xml, options)
         xml.terminal do
           xml.TerminalID '01'
-          xml.CardPresentCode 'UseDefault'
+          xml.CardPresentCode options[:card_present_code] || 'UseDefault'
           xml.CardholderPresentCode 'UseDefault'
           xml.CardInputCode 'UseDefault'
           xml.CVVPresenceCode 'UseDefault'
--- a/test/remote/gateways/remote_element_test.rb
+++ b/test/remote/gateways/remote_element_test.rb
@@ -50,6 +50,12 @@ class RemoteElementTest < Test::Unit::TestCase
     assert_equal 'Approved', response.message
   end
 
+  def test_successful_purchase_with_card_present_code
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(card_present_code: 'Present'))
+    assert_success response
+    assert_equal 'Approved', response.message
+  end
+
   def test_successful_authorize_and_capture
     auth = @gateway.authorize(@amount, @credit_card, @options)
     assert_success auth
--- a/test/unit/gateways/element_test.rb
+++ b/test/unit/gateways/element_test.rb
@@ -142,6 +142,11 @@ class ElementTest < Test::Unit::TestCase
     assert_failure response
   end
 
+  def test_successful_purchase_with_card_present_code
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(card_present_code: 'Present'))
+    assert_equal 'Present', response.params['terminal']['cardpresentcode']
+  end
+
   def test_scrub
     assert @gateway.supports_scrubbing?
     assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@
 * Mercado Pago: Add payment_method_option_id field [schwarzgeist] #3635
 * Stripe: Provide error when attempting an authorize with ACH [britth] #3633
 * EBANX: Send original order id as merchant_payment_code metadata [miguelxpn] #3637 
+* Element: Add card_present_code field [schwarzgeist] #3623
 
 == Version 1.107.3 (May 8, 2020)
 * Realex: Ignore IPv6 unsupported addresses [elfassy] #3622


lib/active_merchant/billing/gateways/element.rb
```diff
@@ -194,7 +194,7 @@ def add_transaction(xml, money, options = {})
       def add_terminal(xml, options)
         xml.terminal do
           xml.TerminalID '01'
-          xml.CardPresentCode 'UseDefault'
+          xml.CardPresentCode options[:card_present_code] if options[:card_present_code]
```
therufs, line 197: Since the remote tests still pass, I assume that Element doesn't care if we're not passing this tag at all?

schwarzgeist: That's a really good point. I like the idea of updating with a fallback case. The only argument for **not** using "UseDefault" as a fallback is that there seems to be a discrepancy for what the default value should be. There is some recent documentation for the Element gateway that suggests "Default" is the default value. (see below)

Screen Shot 2020-05-12 at 8 26 04 PM

I'm going to try to find a ground truth as to what exactly the default value should be (and if it changes), and update with that value as a fallback. (I imagine it's _UseDefault_, since that's what seems to work at the moment)

 While I'm at it, going to update the commit message according to [this](https://gist.github.com/jknipp/dbe9896a7fd74a0138218ae875243512)

schwarzgeist: Well, it's definitely UseDefault

schwarzgeist: Made "UseDefault" fallback case: https://github.com/activemerchant/active_merchant/blob/8a5cdb9329e2ec7d38acdd23e83c5d680a4db2f5/lib/active_merchant/billing/gateways/element.rb#L197

I moved the entry in CHANGELOG for this issue back into HEAD and updated the Pull Request to track this one. Is moving it into HEAD okay here?

Also updated the commit message to conform to styling.

leila-alderman: Changelog entries should always be located under the `HEAD` section. After you rebase this branch, you'll need to resolve conflicts within the CHANGELOG file and ensure that your entry is still located at the bottom of the `HEAD` section.

test/remote/gateways/remote_element_test.rb
```diff
@@ -159,4 +159,13 @@ def test_transcript_scrubbing_with_echeck
     assert_scrubbed(@check.routing_number, transcript)
     assert_scrubbed(@gateway.options[:account_token], transcript)
   end
+
+  def test_successful_purchase_with_card_present_code
```
leila-alderman, line 163: It looks like this method might more logically be located and modeled after `test_successful_purchase_with_shipping_address`.

test/remote/gateways/remote_element_test.rb
```diff
@@ -159,4 +159,13 @@ def test_transcript_scrubbing_with_echeck
...
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(card_present_code: 'Present'))
+    assert_success response
+    assert_equal 'Approved', response.message
+    assert_match %r{Street address and postal code do not match}, response.avs_result['message']
+    assert_match %r{CVV matches}, response.cvv_result['message']
+    assert_equal 'Present', response.params["terminal"]["cardpresentcode"]
```
leila-alderman, line 169: When I run the tests locally, this line fails because the response contains `<CardPresentCode>UseDefault</CardPresentCode>` despite the request containing `<CardPresentCode>Present</CardPresentCode>`. Does this pass for you?

test/remote/gateways/remote_element_test.rb
```diff
@@ -159,4 +159,13 @@ def test_transcript_scrubbing_with_echeck
     assert_scrubbed(@check.routing_number, transcript)
     assert_scrubbed(@gateway.options[:account_token], transcript)
   end
+
+  def test_successful_purchase_with_card_present_code
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(card_present_code: 'Present'))
+    assert_success response
+    assert_equal 'Approved', response.message
+    assert_match %r{Street address and postal code do not match}, response.avs_result['message']
+    assert_match %r{CVV matches}, response.cvv_result['message']
```
leila-alderman, line 168: Since these two lines are checking things that are unrelated to this new field and are already covered by another test, I would remove them.

test/remote/gateways/remote_element_test.rb
schwarzgeist: https://github.com/activemerchant/active_merchant/pull/3623/commits/548f0082dd3b100808794f0895231bcfb804d56f#

Moved the test case immediately after `test_successful_purchase_with_shipping_address`. Also modeled after that method.

test/remote/gateways/remote_element_test.rb
schwarzgeist: It did not pass for me actually, I think the test itself was incorrect to perform a comparison of `cardpresentcode`. That comparison has been removed to make it behave like the `test_successful_purchase_with_shipping_address` method.

test/remote/gateways/remote_element_test.rb
schwarzgeist: The method now looks like this:
https://github.com/activemerchant/active_merchant/blob/548f0082dd3b100808794f0895231bcfb804d56f/test/remote/gateways/remote_element_test.rb#L53-L57

test/unit/gateways/element_test.rb
```diff
@@ -142,6 +142,11 @@ def test_handles_error_response
     assert_failure response
   end
 
+  def test_successful_purchase_with_card_present_code
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(card_present_code: 'Present'))
+    assert_equal 'Present', response.params["terminal"]["cardpresentcode"]
```
leila-alderman, line 147: Be sure to use single quotation marks here to keep RuboCop happy. 😁 
```suggestion
    assert_equal 'Present', response.params['terminal']['cardpresentcode']
```

schwarzgeist: Whoops. Fixed:
https://github.com/activemerchant/active_merchant/blob/1aeb26ca1b256c188f8d483ed7ec31ee6cf37c45/test/unit/gateways/element_test.rb#L147

PR #3625, opened 2020-05-12
Stripe Payment Intents: Utilize execute_threed flag to determine success

britth:
Currently Stripe Payment Intents will return a successful response
if a user tries to initiate a purchase or auth and the api responds
with a status of `required_action` (i.e. if additional steps are
required to authenticate with 3ds and proceed with the transaction).
In other adapters that implement 3DS, we use an `execute_threed` flag
to either signify that we should hit a 3DS specific endpoint, or
classify a response as success/failure depending on whether or not
3DS was requested. For Stripe, there's no separate 3DS endpoint; rather,
if a transaction requires 3DS the response will just signify that in the
status field and by having redirect urls. This PR updates to use the
presence of an `execute_threed` option to either classify the response
as a failure should that field not be present and authentication was
required, or default to the normal success determination

Questions:
This will have potential impacts on backwards compatibility. Would it be
better to have a flag specifically saying you do not want to request 3DS
instead of the normal execute_threed flag. That would go against all other
implementations though...

Remote:
37 tests, 177 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
10 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -295,6 +295,16 @@ module ActiveMerchant #:nodoc:
 
         options.merge(idempotency_key: "#{options[:idempotency_key]}-#{suffix}")
       end
+
+      def success_from(response, options)
+        if response['status'] == 'requires_action' && !options[:execute_threed]
+          response['error'] = {}
+          response['error']['message'] = 'Received unexpected 3DS authentication response. Use the execute_threed option to initiate a proper 3DS flow.'
+          return false
+        end
+
+        super(response, options)
+      end
     end
   end
 end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,8 +5,9 @@
 * Vantiv Express: Implement true verify [leila-alderman] #3617
 * Litle: Pass expiration data for basis payment method [therufs] #3606
 * Stripe Payment Intents: Error handling and backwards compatibility within refund [britth] #3627
-* HPS: Prevent errors when account_type or account_holder_type nil are nil [britth] #3628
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -281,6 +281,16 @@ def format_idempotency_key(options, suffix)
 
         options.merge(idempotency_key: "#{options[:idempotency_key]}-#{suffix}")
       end
+
+      def success_from(response, options)
+        if response['status'] == 'requires_action' && !options[:execute_threed]
```
britth, line 286: another thought related to question in PR description would be to add one more flag so that we can still standardize around `execute_threed` but explicitly request to require it, e.g. something like `if response['status'] == 'requires_action' && options[:require_threed_flag] && !options[:execute_threed]`...but that seems maybe overly complicated?

chinhle23: To me, this does seem overly complicated. Spreedly currently does not use the `execute_threed` flag for Stripe Payment Intents implementation. This is something we'll need to add in core, right?

chinhle23: n/m just saw your core update.

britth: yep correct, to take advantage of this, any integrations will need to pass in an `execute_threed` flag in their options params. This approach matches what we do in other AM adapters that allow for 3DS auth, like adyen: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/adyen.rb#L530

PR #3626, opened 2020-05-12
D Local: Handle invalid country code errors

curiousepic:
Implemented handling similar to other gateways.

Unit:
18 tests, 71 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
27 tests, 71 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -93,8 +93,10 @@ module ActiveMerchant #:nodoc:
         post[:country] = lookup_country_code(address[:country])
       end
 
-      def lookup_country_code(country)
-        Country.find(country).code(:alpha2).value
+      def lookup_country_code(country_field)
+        Country.find(country_field).code(:alpha2).value
+      rescue InvalidCountryCodeError
+        nil
       end
 
       def add_payer(post, card, options)
--- a/test/unit/gateways/d_local_test.rb
+++ b/test/unit/gateways/d_local_test.rb
@@ -70,6 +70,14 @@ class DLocalTest < Test::Unit::TestCase
     end.respond_with(successful_authorize_response)
   end
 
+  def test_invalid_country
+    stub_comms(@gateway, :ssl_request) do
+      @gateway.authorize(@amount, @credit_card, @options.merge(billing_address: address(country: 'INVALID')))
+    end.check_request do |method, endpoint, data, headers|
+      assert_match(/\"country\":null/, data)
+    end.respond_with(successful_authorize_response)
+  end
+
   def test_failed_authorize
     @gateway.expects(:ssl_post).returns(failed_authorize_response)
 


Carrigan:
I was having a tough time copying my error before from tmux, but it is:
```
test/comm_stub.rb:22:in `block in respond_with'
     66:     stub_comms(@gateway, :ssl_request) do
     67:       @gateway.authorize(@amount, @credit_card, @options)
     68:     end.check_request do |method, endpoint, data, headers|
  => 69:       assert_match(/"country\":\"CA\"/, data)
     70:     end.respond_with(successful_authorize_response)
     71:   end
     72:

test/unit/gateways/d_local_test.rb:69:in `block in test_passing_country_as_string'
Failure: test_passing_country_as_string(DLocalTest):
  </"country\":\"CA\"/> was expected to be =~
  <"{\"amount\":\"1.00\",\"currency\":\"USD\",\"payment_method_id\":\"CARD\",\"payment_method_flow\":\"DIRECT\",\"country\":{\"value\":\"CA\",\"format\":\"alpha2\"},\"payer\":{\"name\":\"Longbob Longsen\",\"phone\":\"(555)555-5555\",\"address\":{\"state\":\"ON\",\"city\":\"Ottawa\"}},\"card\":{\"holder_name\":\"Longbob Longsen\",\"expiration_month\":9,\"expiration_year\":2021,\"number\":\"4242424242424242\",\"cvv\":\"123\",\"capture\":false},\"order_id\":\"1\"}">.
```
Looks like this changed the country from the string "CA" to a hash of `{"value":"CA","format":"alpha2"}` for me.

curiousepic:
How strange 🤔 Travis is also now failing with that error... I have no idea what's going on.

Carrigan:
It looks like other gateways that use this code are XML based. I wonder if the XML code explicitely calls `.to_s` on the `Country` object while the JSON version does not. 

Either way, it looks like in the original code, `.value` was called at the end of the chain. I imagine that adding `&.value` to the end of your return value will fix this.

Carrigan:
Your latest changes fix the test for me 🎉 

I get two local linting errors, one of which is still caused by this change:

```
Offenses:

lib/active_merchant/billing/gateways/d_local.rb:97:24: C: Style/RescueModifier: Avoid using rescue in its modifier form. (https://github.com/rubocop-hq/ruby-style-guide#no-rescue-modifiers)
        country_code = Country.find(country_field) rescue nil
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

curiousepic:
@Carrigan Odd that rubocop isn't detecting the same pattern in CyberSource and elsewhere...

Carrigan:
I think that rewriting the function as shown below would be best since it reduces the number of `try` statements and fixes the Rubocop linter:

```ruby
def lookup_country_code(country_field)
  Country.find(country_field).code(:alpha2).value
rescue InvalidCountryCodeError
  nil
end
```

If we want consistency, it might also be good to copy this back to the other offending files and remove them from from the rubocop todo. What do you think?

curiousepic:
Sounds good. So much for a 10 minute fix 😄

curiousepic:
@Carrigan Wait, does that code without the `&.` safe navigation not trigger the implicit string conversion issue?

PR #3631, opened 2020-05-13
Update supported Ruby and Rails versions

leila-alderman:
Updated the supported Ruby and Rails versions to remove versions that
have officially been moved to end of life while maintaining support for
versions 5.0 and 5.1 for Rails.

 - Removed Ruby versions 2.3 and 2.4
 - Added Ruby versions 2.6 and 2.7
 - Removed Rails version 4.2
 - Added Rails version 6.0

In addition, the following changes were made as part of this update:

 - RuboCop target Ruby version was updated to 2.5, which necessitated
 some RuboCop fixes in various files
 - The RuboCop version was upgraded to 0.83, which necessitated even
 more RuboCop fixes in various files
 - [Fix an REXML error](https://github.com/kbeckman/omniauth-wsfed/pull/30)
for the So Easy Pay gateway
 - Add in missing Active Support core extensions to the test helper file

All unit tests:
4497 tests, 71905 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/check.rb
+++ b/lib/active_merchant/billing/check.rb
@@ -7,8 +7,8 @@ module ActiveMerchant #:nodoc:
     # You may use Check in place of CreditCard with any gateway that supports it.
     class Check < Model
       attr_accessor :first_name, :last_name,
-        :bank_name, :routing_number, :account_number,
-        :account_holder_type, :account_type, :number
+                    :bank_name, :routing_number, :account_number,
+                    :account_holder_type, :account_type, :number
 
       # Used for Canadian bank accounts
       attr_accessor :institution_number, :transit_number
--- a/lib/active_merchant/billing/gateways/opp.rb
+++ b/lib/active_merchant/billing/gateways/opp.rb
@@ -126,7 +126,7 @@ module ActiveMerchant #:nodoc:
         # debit
         options[:registrationId] = payment if payment.is_a?(String)
         execute_dbpa(options[:risk_workflow] ? 'PA.CP' : 'DB',
-          money, payment, options)
+                     money, payment, options)
       end
 
       def authorize(money, payment, options={})
@@ -357,7 +357,7 @@ module ActiveMerchant #:nodoc:
 
         success_regex = /^(000\.000\.|000\.100\.1|000\.[36])/
 
-        if success_regex =~ response['result']['code']
+        if success_regex&.match?(response['result']['code'])
           true
         else
           false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,7 +15,11 @@ AllCops:
     - "lib/active_merchant/billing/gateways/paypal_express.rb"
     - "vendor/**/*"
   ExtraDetails: false
-  TargetRubyVersion: 2.3
+  # This setting disables any new cops in updated versions of RuboCop.
+  # To complete cleaning up the AM code, this should be removed to enable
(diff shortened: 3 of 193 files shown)


lib/active_merchant/billing/gateways/opp.rb
```diff
@@ -357,7 +357,7 @@ def success_from(response)
 
         success_regex = /^(000\.000\.|000\.100\.1|000\.[36])/
 
-        if success_regex =~ response['result']['code']
+        if success_regex&.match?(response['result']['code'])
```
InPermutation, line 360: What does the `&` signify here?

leila-alderman: So in all honesty, that was one of the changes that RuboCop made automatically for me. However, [this issue](https://github.com/rubocop-hq/rubocop/issues/5026) states that the `&` there provides a `nil` check so that in the event that `success_regex` is `nil` this won't error out thanks to [the safe navigation operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/).

therufs: I'm good with the use of `match` but think we should get rid of this `&`; if `success_regex` is nil, something has gone very wrong, but the presence of the `&` will result in the check returning the `else` clause (thereby obscuring that something has gone very wrong.) 

I wonder if Rubocop would be okay with not `&`ing a constant (i.e., changing `success_regex` to `SUCCESS_REGEX`.) Otherwise, I think we should consider telling Rubocop to be mad better. :)

therufs:
Local tests all pass for me; looks good other than the `&` thing :)

lib/active_merchant/billing/check.rb
```diff
@@ -7,8 +7,8 @@ module Billing #:nodoc:
     # You may use Check in place of CreditCard with any gateway that supports it.
     class Check < Model
       attr_accessor :first_name, :last_name,
-        :bank_name, :routing_number, :account_number,
-        :account_holder_type, :account_type, :number
+                    :bank_name, :routing_number, :account_number,
+                    :account_holder_type, :account_type, :number
```
bdewater, line 11: To be honest, I'm not really a fan of all these changes (nor Rubocop's pedantic standards around this). It makes this PR pretty huge. Not sure if we ever discussed which rules we want to adopt or not - but the default settings don't seem to apply to the existing code style.

lib/active_merchant/billing/gateways/opp.rb
bdewater: This is Rubocop being defensive on the autocorrect: https://github.com/rubocop-hq/rubocop-performance/blob/7caf6f53e782d1bc889b69fdbf8116632d4e678c/lib/rubocop/cop/performance/regexp_match.rb#L261-L270

It doesn't analyze whether that local variable is always set or not, since for most cases that would be nearly impossible to reliable do anyway with Ruby :) `TYPES_IMPLEMENTING_MATCH` contains constants, which this should've been anyway IMO, and will autocorrect without the operator.

casperisfine:
👋 can I help with this? activemerchant will soon be the last gem causing warnings in our app.

leila-alderman:
@casperisfine If you'd like to help out by reviewing [the new version of this PR](https://github.com/activemerchant/active_merchant/pull/3656), I would really appreciate it. Thanks for offering to help!

PR #3633, opened 2020-05-14
Stripe: Provide error when attempting an authorize with ACH

britth:
Currently, the Stripe adapter can be used to attempt an authorize
with a check object. However, this causes an application error,
because if you pass a check directly in an authorize, we don't try
to tokenize it in any way before processing, so it errors out with
a NoMethodError (we're trying to process it as though it's a credit
card, and in this case, there's no `month` method on the check,
causing an error). Additionally, ACH simply does not seem to be
supported by Stripe for revenue transfers where funds aren't captured - 
after storing a check as you'd have to do for a purchase, you'll
end up with an error related to capture being false if you attempt
the auth. This PR updates the logic in authorize to check if the
payment method is a check before processing the auth, and if so, we
return an error the way we do for purchases when the check has not
been previously stored. Also adds a test showing that auths fail
anyway if attempted using a stored check, but don't result in any
unexpected application errors.

Further reading: https://stripe.com/docs/ach

Unit:
138 tests, 738 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
71 tests, 325 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@
 * CyberSource: Fix invalid and missing field tests [curiousepic] #3634
 * CyberSource: Pass stored credentials with purchase [curiousepic] #3636
 * Mercado Pago: Add payment_method_option_id field [schwarzgeist] #3635
+* Stripe: Provide error when attempting an authorize with ACH [britth] #3633
 
 == Version 1.107.3 (May 8, 2020)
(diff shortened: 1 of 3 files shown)


britth:
thanks @Carrigan !

PR #3635, opened 2020-05-15
Mercado Pago: Add payment_method_option_id field

schwarzgeist:
CE-562, CE-593

Unit:
38 tests, 174 assertions, 0 failures, 0 errors,
100% passed

Remote:
32 tests, 82 assertions, 8 failures, 1 errors,
71.875% passed

All Unit Tests:
4501 tests, 71919 assertions, 0 failures, 0 errors,
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/mercado_pago.rb
+++ b/lib/active_merchant/billing/gateways/mercado_pago.rb
@@ -113,6 +113,7 @@ module ActiveMerchant #:nodoc:
 
         post[:processing_mode] = options[:processing_mode]
         post[:merchant_account_id] = options[:merchant_account_id] if options[:merchant_account_id]
+        post[:payment_method_option_id] = options[:payment_method_option_id] if options[:payment_method_option_id]
         add_merchant_services(post, options)
       end
 
--- a/test/remote/gateways/remote_mercado_pago_test.rb
+++ b/test/remote/gateways/remote_mercado_pago_test.rb
@@ -42,7 +42,8 @@ class RemoteMercadoPagoTest < Test::Unit::TestCase
       processing_mode: 'gateway',
       merchant_account_id: fixtures(:mercado_pago)[:merchant_account_id],
       fraud_scoring: true,
-      fraud_manual_review: true
+      fraud_manual_review: true,
+      payment_method_option_id: '123abc'
     }
   end
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@
 * CyberSource: Stored Credential fixes [curiousepic] #3624
 * CyberSource: Fix invalid and missing field tests [curiousepic] #3634
 * CyberSource: Pass stored credentials with purchase [curiousepic] #3636
+* Mercado Pago: Add payment_method_option_id field [schwarzgeist] #3635
 
 == Version 1.107.3 (May 8, 2020)
 * Realex: Ignore IPv6 unsupported addresses [elfassy] #3622


schwarzgeist:
The one comment I should point out is this line (and the function is lives in) https://github.com/activemerchant/active_merchant/blob/26af60cec60dc223cf29a9b0e77bb399434fc560/test/remote/gateways/remote_mercado_pago_test.rb#L93

is kind-of a placeholder. I wasn't able to get a response that pointed me in a direction of where to look for this exact comparison. Not sure what the most appropriate way of handling this is. Committed anyway in hopes someone else would have a more sound idea.

lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -214,6 +215,10 @@ def add_taxes(post, options)
         end
       end
 
+      def add_payment_method_option_id(post, options)
```
leila-alderman, line 218: Because this new field is so closely related to using [the processing mode parameters for Mercado Pago](https://www.mercadopago.com.co/developers/en/guides/gateway/api/receiving-payments), this field would make more logical sense being added to the existing `add_processing_mode` method.

test/remote/gateways/remote_mercado_pago_test.rb
```diff
@@ -84,6 +84,15 @@ def test_successful_purchase_with_processing_mode_gateway
     assert_equal 'accredited', response.message
   end
 
+  def test_successful_purchase_with_payment_method_option_id
```
leila-alderman, line 87: Again, because this new parameter is so closely tied into the existing fields for processing mode, I would suggest not writing a new test to cover this test case but to instead expand the existing `test_successful_purchase_with_processing_mode_gateway` by adding the new `payment_method_option_id` to the `@processing_options` hash.

test/remote/gateways/remote_mercado_pago_test.rb
```diff
@@ -84,6 +84,15 @@ def test_successful_purchase_with_processing_mode_gateway
...
+    payment_method_option_id = '1234'
+
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(payment_method_option_id: payment_method_option_id))
+    assert_success response
+    assert_equal 'accredited', response.message
+    assert_equal payment_method_option_id, response
```
leila-alderman, line 93: It's rare for gateways to return all of the request parameters in their response. Therefore, when it comes to testing gateway specific fields in a remote test, we typically just send off the request with the fields and then look for a successful response. Therefore, you don't need to look for any specific response parameters.

test/unit/gateways/mercado_pago_test.rb
```diff
@@ -429,6 +429,16 @@ def test_invalid_taxes_shape
     end
   end
 
+  def test_successful_purchase_with_payment_method_option_id
+    payment_method_option_id = '1234'
+
+    stub_comms do
+      @gateway.purchase(@amount, @credit_card, @options.merge(payment_method_option_id: payment_method_option_id))
+    end.check_request do |endpoint, data, headers|
+      assert_match("\"payment_method_option_id\":\"#{payment_method_option_id}\"", data) if endpoint =~ /payments/
```
leila-alderman, line 438: Not sure that I understand what the `if` statement is doing here. Could you explain your thoughts on that?

schwarzgeist: That began as an artifact of duplicating code. It stayed from a misunderstanding of the documentation. It is probably not appropriate here.
Screen Shot 2020-05-15 at 10 41 26 AM

Regardless, I think this entire test case would go away. There's not currently anything in the unit test that deals with processing options, and adding them may increase the scope of this pull request.

lib/active_merchant/billing/gateways/mercado_pago.rb
schwarzgeist: https://github.com/activemerchant/active_merchant/pull/3635/commits/007ab496410a32775742ba9d92258cbc61952732

I've moved _ payment_method_option_id_ to the _add_processing_mode__ method and removed the _add_payment_method_option_id_ method

test/remote/gateways/remote_mercado_pago_test.rb
schwarzgeist: 007ab496410a32775742ba9d92258cbc61952732

I've added `payment_method_option_id` to the `@processing_options` hash, but with one caveat, it's value. I believe it's real-word value would a hash, would it be worth just creating a random hash for this test?

test/unit/gateways/mercado_pago_test.rb
schwarzgeist: 007ab496410a32775742ba9d92258cbc61952732

I've removed this test case for now

test/remote/gateways/remote_mercado_pago_test.rb
schwarzgeist: https://github.com/activemerchant/active_merchant/blob/e0395f63e520c902caa2985b5dd7523ee19ead38/test/remote/gateways/remote_mercado_pago_test.rb#L46

PR #3641, opened 2020-05-21
[Cybersource] Set a default collectionIndicator

pi3r:
For mastercard cards, we need to always be passing a collectionIndicator for the request to succeed.

I decided to use `2` as the default collectionIndicator, as it was already the one used for tokenized card and
it seems to match our usecase from all the pther choices:

- 0: UCAF collection is not supported at your web
site.
- 1: UCAF collection is supported at your web
site, and the UCAF was populated.
- 2: UCAF collection is supported at your web
site and the UCAF was populated. This value
indicates a successful Mastercard Identity
Check transaction.
- 5: UCAF collection is supported at your web
site, and the UCAF was populated based on the
risk assessment that the issuer performed. This
value is supported only for Masterpass
transactions.
- 6: UCAF collection is supported at your web
site, and the UCAF was populated based on the
risk assessment that you performed. This value
is supported only for Masterpass transactions.

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -34,6 +34,7 @@ module ActiveMerchant #:nodoc:
         jcb: 'js',
         discover: 'pb',
       }.freeze
+      DEFAULT_COLLECTION_INDICATOR = 2
 
       self.supported_cardtypes = [:visa, :master, :american_express, :discover, :diners_club, :jcb, :dankort, :maestro, :elo]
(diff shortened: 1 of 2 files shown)


pi3r:
Seems to be another issue that's happening on master too, I'm looking into it

pi3r:
@nebdil this pr https://github.com/activemerchant/active_merchant/pull/3631/files#diff-e82b302c845780c96d5c9f7a6a925482R160 will address the failures

pi3r:
```
4505 tests, 71986 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

PR #3643, opened 2020-05-23
Implement change to formatting of VerificationStr1 to strip control characters

bittercoder:
This pull request implements a change to strip out \r  and \n characters from address components in the VerificationStr1.
This is to conform with tighter verification on VerificationStr1 values introduced by FirstData in May 2020.


curiousepic:
@bittercoder FYI for the future, this adapter is locked to v11 of the FD API. The updated version which supports v27+ is https://github.com/activemerchant/active_merchant/blob/3b7a6b6189f24a3268d0313546b26436038b94df/lib/active_merchant/billing/gateways/firstdata_e4_v27.rb (the file name is misleading)
We will be looking into making this same change in that adapter.

PR #3646, opened 2020-05-29
Fix Rubocop UnusedMethodArgument

cdmackeyfree:
4466 tests, 71579 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/moneris_us.rb
```diff
@@ -0,0 +1,352 @@
+require 'rexml/document'
```
therufs, line 1: ... where did this file come from??? :confused_england:

lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -129,8 +129,13 @@ def credit(money, authorization, options = {})
         refund(money, authorization, options)
       end
 
+<<<<<<< HEAD
```
therufs, line 132: a small thing to clean up here

PR #3651, opened 2020-06-01
DLocal: Fix address field names

molbrown:
DLocal was previously only passing state and country in the billing
address due to checking for incorrect address fields. DLocal also
separates number and street fields, so adds address parser for address1.

Unit:
20 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
27 tests, 71 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-1267

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -118,10 +118,20 @@ module ActiveMerchant #:nodoc:
         address_object = {}
         address_object[:state] = address[:state] if address[:state]
         address_object[:city] = address[:city] if address[:city]
-        address_object[:zip_code] = address[:zip_code] if address[:zip_code]
-        address_object[:street] = address[:street] if address[:street]
-        address_object[:number] = address[:number] if address[:number]
+        address_object[:zip_code] = address[:zip] if address[:zip]
+        address_object[:street] = address[:street] || parse_street(address) if parse_street(address)
+        address_object[:number] = address[:number] || parse_house_number(address) if parse_house_number(address)
         address_object
+      end
+
+      def parse_street(address)
+        street = address[:address1].split(/\s+/).keep_if { |x| x !~ /\d/ }.join(' ')
+        street.empty? ? nil : street
+      end
+
+      def parse_house_number(address)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -118,12 +118,22 @@ def add_address(post, card, options)
         address_object = {}
         address_object[:state] = address[:state] if address[:state]
         address_object[:city] = address[:city] if address[:city]
-        address_object[:zip_code] = address[:zip_code] if address[:zip_code]
-        address_object[:street] = address[:street] if address[:street]
-        address_object[:number] = address[:number] if address[:number]
+        address_object[:zip_code] = address[:zip] if address[:zip]
+        address_object[:street] = parse_street(address) if address[:address1]
+        address_object[:number] = parse_house_number(address) if address[:address1]
```
britth, line 123: do we care about backwards compatibility? or since the existing code is very non-standard compared to others, do we feel fine removing support for those old fields

test/unit/gateways/d_local_test.rb
```diff
@@ -62,6 +62,18 @@ def test_successful_authorize_without_address
     assert_equal 'D-15104-be03e883-3e6b-497d-840e-54c8b6209bc3', response.authorization
   end
 
+  def test_passing_billing_address
```
britth, line 65: 😎
What do you think about another test where street or number isn't present in the address field? I don't know how likely that'd actually be but might be worth just exercising that bit of code too

molbrown:
Thanks for that feedback, @britth I added a remote and unit test to test out an `address1` without both components, and also changed the handling to not include the field if there is no value for it. 

DLocal only requires those fields are sent in India, which is why I went this direction: https://docs.dlocal.com/api-documentation/payins-api-reference/payments#the-address-object

Also, we have had payments succeeding with partial addresses up to this point, so 🤷‍♀️

lib/active_merchant/billing/gateways/d_local.rb
molbrown: 👍 I have added `address[:street]` and `address[:number]` back in for backwards compatibility/ allowing direct assignment.

PR #3653, opened 2020-06-02
Forte: Change default sec_code value to PPD

molbrown:
The sec_code field for echeck transactions was recently added to the
Forte gateway in 1c4b10818a780d0c80e4554e0843741193c6584a. After
customer issues and digging into Forte value definitions, it has become
clear that 'PPD' should be the default value when none is specified.
Forte currently assigns PPD to a transaction where no sec_code is specified,
so we should follow that pattern here.

Forte will soon be requiring a sec_code is sent on any echeck
transaction, which is why a default is being set.

Remote:
100% passed

Unit:
20 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

CE-532

### DIFF
--- a/lib/active_merchant/billing/gateways/forte.rb
+++ b/lib/active_merchant/billing/gateways/forte.rb
@@ -166,7 +166,7 @@ module ActiveMerchant #:nodoc:
         post[:echeck][:routing_number] = payment.routing_number
         post[:echeck][:account_type] = payment.account_type
         post[:echeck][:check_number] = payment.number
-        post[:echeck][:sec_code] = options[:sec_code] || "WEB"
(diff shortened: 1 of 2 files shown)


redbeardaz:
@molbrown Just to be sure we're not jumping the gun... Fattmerchant provided this:
>When reading through Forte's documentation it appears that sec code may be optional, would it be safest for you to not pass any value if none is provided?

molbrown:
@redbeardaz Forte has started requiring the sec_code in their sandbox (`MANDATORY FIELD MISSING:echeck.sec_code`), ahead of requiring it in production at some point his year.  The default is an effort to avoid a flood of failures at that point. Users can pass another `options[:sec_code]` value if this default is not what they need.

PR #3655, opened 2020-06-04
Default executeThreeD to false if it's not set

naashton:
This seems like what we want to do, but the output of the remote test for `test_successful_authorize_with_3ds2_app_based_request` and `test_successful_authorize_with_3ds2_browser_client_data` is a bit concerning as I'm not sure how it will affect downstream users. 

`bundle exec rake test:remote TEST=test/remote/gateways/remote_adyen_test.rb`

The ticket states that Adyen does not have a default behavior if the field is passed as null, but passing in false seems to completely change the response for these requests.

I think the alternative here, implementing it as a GSF and allowing the individual users to decide whether or not they want to explicitly set `executeThreeD` to false, may be safer.

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -413,8 +413,12 @@ module ActiveMerchant #:nodoc:
             post[:threeDS2RequestData] = { deviceChannel: device_channel, notificationURL: three_ds_2_options[:notification_url] }
           end
 
-          post[:additionalData][:executeThreeD] = options[:execute_threed] || false
+          if options.has_key?(:execute_threed)
+            post[:additionalData][:executeThreeD] = options[:execute_threed]
+          end
+          
           post[:additionalData][:scaExemption] = options[:sca_exemption] if options[:sca_exemption]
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -413,10 +413,9 @@ def add_3ds(post, options)
             post[:threeDS2RequestData] = { deviceChannel: device_channel, notificationURL: three_ds_2_options[:notification_url] }
           end
 
-          if options.has_key?(:execute_threed)
-            post[:additionalData][:executeThreeD] = options[:execute_threed]
-            post[:additionalData][:scaExemption] = options[:sca_exemption] if options[:sca_exemption]
-          end
+          post[:additionalData][:executeThreeD] = options[:execute_threed] || false
```
molbrown, line 416: If `executeThreeD` is sent to Adyen as false, it will override the `:threed_dynamic` choice made by the customer, and 3DS will not be performed. Up to this point, this has functioned like so:
- `execute_threed: true` definitely do 3DS
-  `threed_dynamic: true` include some additonal fields to indicate to Adyen to check the transaction against my preset rules. Perform 3DS only if it breaks my rules.
- neither true = the additional fields are not included, and `executeThreeD` is null. This third time is the only time we should be setting `executeThreeD` as false.

PR #3657, opened 2020-06-05
Payu Latam: Prevent blank country in billing_address

britth:
In PayU Latam requests, passing in certain billing_address fields
as empty strings (specifically `country`) will result in an error.
This PR updates the logic to not pass a billing_address country field
that's blank.

Remote:
34 tests, 83 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.0588% passed
unrelated failure, also on master: `test_well_formed_refund_fails_as_expected`

Unit:
32 tests, 122 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * Vantiv: Vantiv(Element): add option to send terminal id in transactions [cdmackeyfree] #3654
 * Update supported Ruby and Rails versions [leila-alderman] #3656
 * CI: Drop unused sudo: false Travis directive [olleolleolle] #3616
+* PayU Latam: Prevent blank country in billing_address [britth] #3657
 
 == Version 1.107.4 (Jun 2, 2020)
 * Elavon: Implement true verify action [leila-alderman] #3610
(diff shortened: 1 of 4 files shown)


curiousepic:
@britth Do we happen to know for sure that the gateway will accept a request with certain billing address fields not present at all (and the sandbox is just permissive)?

britth:
@curiousepic great question - I know in certain cases we definitely don't want to send an empty string, and I know many requests are fine when no address is passed at all, but the docs aren't super clear on strict requirements if some fields are present and some aren't. This came up because a request sent an empty string for country, causing the API to return the error: `property: transaction.payer.billingAddress.country, message: size must be between 2 and 2\n`

curiousepic:
@britth I guess I'm wondering if switching to not sending the field at all could end up causing more error responses than we fix 🤔 
Maybe we start with just the country and then expand if these requests without a country have an ok result?

britth:
@curiousepic i like that, thanks for the suggestion 👍 updated to only prevent blank country

PR #3658, opened 2020-06-08
[Cybersource] Send cavv as xid when xid is missing

pi3r:
We've been experiencing an error with 3DS 2.0 where Cybersource still
requires an xid even though this field isn't present for 3DS 2.0 transactions.

This is the exact error `<c:missingField>c:xid</c:missingField>`

After discussing with Cybersource and Cardinal it was decided that we
will send the cavv as the xid for 3DS 2.0 until they sort out their api.


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -617,7 +617,7 @@ def add_normalized_threeds_2_data(xml, payment_method, options)
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
         xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
         xml.tag!('eciRaw', threeds_2_options[:eci]) if threeds_2_options[:eci]
-        xml.tag!('xid', threeds_2_options[:xid]) if threeds_2_options[:xid]
+        xml.tag!('xid', threeds_2_options[:xid] || threeds_2_options[:cavv])
```
curiousepic, line 620: It should probably check for presence of `cavv` before trying to add.

PR #3661, opened 2020-06-11
DLocal: Handle nil address1

molbrown:
Recent address parsing addition in 6419a3474fb87a0d02d75d8312de61e823e8beec
did not account for nil address1. Adds handling to not attempt parsing
when there is no address1.

Unit:
22 tests, 90 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
28 tests, 73 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-1267

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -125,11 +125,15 @@ module ActiveMerchant #:nodoc:
       end
 
       def parse_street(address)
+        return unless address[:address1]
+
         street = address[:address1].split(/\s+/).keep_if { |x| x !~ /\d/ }.join(' ')
         street.empty? ? nil : street
       end
 
       def parse_house_number(address)
+        return unless address[:address1]
+
         house = address[:address1].split(/\s+/).keep_if { |x| x =~ /\d/ }.join(' ')
         house.empty? ? nil : house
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -119,8 +119,8 @@ def add_address(post, card, options)
         address_object[:state] = address[:state] if address[:state]
         address_object[:city] = address[:city] if address[:city]
         address_object[:zip_code] = address[:zip] if address[:zip]
-        address_object[:street] = address[:street] || parse_street(address) if parse_street(address)
-        address_object[:number] = address[:number] || parse_house_number(address) if parse_house_number(address)
+        address_object[:street] = address[:street] || parse_street(address) if ( address[:address1] && parse_street(address) )
+        address_object[:number] = address[:number] || parse_house_number(address) if ( address[:address1] && parse_house_number(address) )
```
curiousepic, line 123: I feel like these parse methods should include a guard clause instead of running it as part of the conditional here.

molbrown: Thanks, updated 👍

PR #3662, opened 2020-06-11
Stripe PI: Implement verify action

leila-alderman:
Added the `verify` action to the Stripe Payment Intents gateway using
the [SetupIntents endpoint](https://stripe.com/docs/api/setup_intents/create).

CE-554

Unit:
13 tests, 77 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
41 tests, 188 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

All unit tests:
ÿ7 tests, 72072 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_stripe_payment_intents_test.rb
+++ b/test/remote/gateways/remote_stripe_payment_intents_test.rb
@@ -592,6 +592,24 @@ class RemoteStripeIntentsTest < Test::Unit::TestCase
     assert_equal store1.params['id'], store2.params['id']
   end
 
+  def test_successful_verify
+    options = {
+      customer: @customer
+    }
+    assert verify = @gateway.verify(@visa_payment_method, options)
+
+    assert_equal 'succeeded', verify.params['status']
+  end
+
+  def test_failed_verify
+    options = {
+      customer: @customer
+    }
+    assert verify = @gateway.verify(@declined_payment_method, options)
+
+    assert_equal 'Your card was declined.', verify.message
+  end
+
   def test_moto_enabled_card_requires_action_when_not_marked
     options = {
       currency: 'GBP',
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -631,7 +649,6 @@ def test_transcript_scrubbing
       currency: 'GBP',
       customer: @customer,
       confirmation_method: 'manual',
-      capture_method: 'manual',
```
leila-alderman, line 634: This is a weird thing to delete, I know, but this test was consistently failing because the byte size of the response just _happened_ to be `1737 bytes`. Because the verification value for the test card is `737`, this test kept thinking that the CVV hadn't been correctly filtered out even though it had. I deleted this line just to change up the size of the response.

I'd really love to see a better implementation of these scrubbing tests that isn't so finicky.

therufs: Might be worth creating an issue for, for posterity?

test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -348,4 +355,38 @@ def failed_service_response
       {"error": {"message": "Error while communicating with one of our backends.  Sorry about that!  We have been notified of the problem.  If you have any questions, we can help at https://support.stripe.com/.", "type": "api_error"  }}
     RESPONSE
   end
+
+  def successful_verify_response
+    <<-RESPONSE
+      {
+        "id": "seti_1Gsw0aAWOtgoysog0XjSBPVX",
```
therufs, line 362: 📡 👾 👾 👾

leila-alderman:
@therufs Thanks for reviewing this!

I just rebased and re-ran the remote tests for Stripe PI, but I'm still not seeing any errors. I also ran the test you called out above by itself, and it still passed. Not sure what's going on there...

image

PR #3606, opened 2020-04-29
Litle: pass expiration data for basis payment method

therufs:
Litle (Vantiv Ecommerce) allows (and may require) expiration date information to be sent when a `litleToken` is used as the payment method.

Tests currently in extreme disarray.

PR #3667, opened 2020-06-12
AAFES: Integrating purchase method for the AAFES gateway

naashton:
First iteration of integrating AAFES gateway into AM starting with the
basic purchase method.

Some nuance for how this gateway works means there is no auth method
(unless otherwise approved by AAFES - an arm of the DoD). What this
means for anyone who wants to use this gateway is that they must become
an approved partner with AAFES and implement the Authentication portion
on their own internal systems.

Additionally, AAFES is both the card issuer and the gateway provider for
the MilStar card. The MilStar card is a credit card only available to
members of the US armed forces. A part of the authentication step, is a
vaulting process, where a MilStar card holder and merchant customer is
redirected from the merchant checkout to the AAFES Stargate (AAFES
gateway vault) and will redirect a tokenized version of the credit card
along with an expiration date (YYMM) and zipcode for verification in
subsequent transactions with the gateway.

On the topic of transactions. AAFES identifies each transaction by an
RRN value that is required to be 12 characters long and base64 encoded.
This value MUST be unique for each transaction and has a life cycle of
30 days on the AAFFES gateway. Capture, Refund, and Void transactions
all require a unique RRN for the transaction as well as an original RRN
to identify the initial Purchase transaction.

It is also worth noting, this gateway requires the currency value to be
passed as a base 10 decimal value as opposed to an integer.
CE-636

Unit: 2 tests, 8 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 2 tests, 4 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/aafes.rb
+++ b/lib/active_merchant/billing/gateways/aafes.rb
@@ -74,13 +74,13 @@ module ActiveMerchant #:nodoc:
       end
 
       def parse(body)
-        puts '---- parse -----'
-        puts body
+        xml = Nokogiri::XML(body)
+        response = Hash.from_xml(xml.to_s)['Message']
       end
 
       def commit(request)
         url = (test? ? test_url : live_url)
-        puts '-------------------'
+        puts '---------- request ---------'
         puts request
         response =
           begin
@@ -88,7 +88,7 @@ module ActiveMerchant #:nodoc:
           rescue StandardError => error
             parse(error.response.body)
           end
-
+        puts response
         Response.new(
           success_from(response),
           message_from(response),
@@ -137,11 +137,11 @@ module ActiveMerchant #:nodoc:
       def add_headers(xml, options)
         xml['cm'].Header do
           xml['cm'].IdentityUUID(@identity_uuid)
-          xml['cm'].LocalDateTime('2019-09-04T22:26:02') #Time.now.getutc
+          xml['cm'].LocalDateTime(Time.now.utc.iso8601)
           xml['cm'].SettleIndicator(false)
           xml['cm'].OrderNumber(options[:order_id])
-          xml['cm'].TransactionId(options[:transaction_id])
-          xml['cm'].TermId(options[:term_id])
+          xml['cm'].transactionId(options[:transaction_id])
+          xml['cm'].termId(options[:term_id])
           xml['cm'].Comment(options[:comment])
           xml['cm'].CustomerID(options[:customer_id])
         end
@@ -155,7 +155,7 @@ module ActiveMerchant #:nodoc:
           xml['cm'].Token('Token')
           xml['cm'].Account(payment.payment_data)
           xml['cm'].Expiration(payment.metadata[:expiration])
-          xml['cm'].AmountField(amount)
+          xml['cm'].AmountField('50.00')
           xml['cm'].PlanNumbers do
             xml['cm'].PlanNumber(options[:plan_number])
           end
--- a/lib/active_merchant/billing/gateways/aafes.rb
+++ b/lib/active_merchant/billing/gateways/aafes.rb
@@ -75,20 +75,18 @@ module ActiveMerchant #:nodoc:
 
       def parse(body)
         xml = Nokogiri::XML(body)
-        response = Hash.from_xml(xml.to_s)['Message']
+        Hash.from_xml(xml.to_s)['Message']
       end
 
       def commit(request)
         url = (test? ? test_url : live_url)
-        puts '---------- request ---------'
-        puts request
         response =
           begin
             parse(ssl_post(url, request, headers))
           rescue StandardError => error
             parse(error.response.body)
           end
-        puts response
+
         Response.new(
           success_from(response),
           message_from(response),
@@ -106,9 +104,12 @@ module ActiveMerchant #:nodoc:
       end
       
       def success_from(response)
+        return true if response.dig('Response', 'ReasonCode').to_i.between?(0,100) unless response.nil?
+        return false
       end
 
       def message_from(response)
+        response.dig('Response', 'ResponseType')
       end
 
       def authorization_from(response)
@@ -155,7 +156,7 @@ module ActiveMerchant #:nodoc:
           xml['cm'].Token('Token')
           xml['cm'].Account(payment.payment_data)
           xml['cm'].Expiration(payment.metadata[:expiration])
-          xml['cm'].AmountField('50.00')
+          xml['cm'].AmountField(amount)
           xml['cm'].PlanNumbers do
             xml['cm'].PlanNumber(options[:plan_number])
           end
--- a/lib/active_merchant/billing/gateways/aafes.rb
+++ b/lib/active_merchant/billing/gateways/aafes.rb
@@ -1,0 +1,129 @@
+require 'nokogiri'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class AafesGateway < Gateway
+      self.test_url = 'https://uat-stargate.aafes.com:1009/stargate/1/creditmessage'
+      self.live_url = ''
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      # TODO: Not sure if AAFES supports traditional cards
+      # 
+    #   self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.homepage_url = ''
+      self.display_name = 'AAFES'
+
+      STANDARD_ERROR_CODE_MAPPING = {}
+
+      def initialize(options={})
+        requires!(options, :some_credential, :another_credential)
+        super
+      end
+
+      def purchase(money, payment, options={})
+        post = {}
+        add_invoice(post, money, options)
+        add_payment(post, payment)
+        add_address(post, payment, options)
+        add_customer_data(post, options)
+
+        commit('sale', post)
+      end
+
+      # AAFES is structured in such a way that the auth should be performed before AM can execute
+      # any transactions
+    #   def authorize(money, payment, options={})
+    #     post = {}
+    #     add_invoice(post, money, options)
+    #     add_payment(post, payment)
+    #     add_address(post, payment, options)
+    #     add_customer_data(post, options)
+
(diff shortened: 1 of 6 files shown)


test/remote/gateways/remote_aafes_test.rb
```diff
@@ -0,0 +1,167 @@
...
+    @amount = '%.2f' % 100
+    @metadata = {
+      :zip => 75236,
+      :expiration => 2210
+    }
+    
+    @milstar_card = ActiveMerchant::Billing::PaymentToken.new(
+      '900PRPYIGCWDS4O2615',
+      @metadata
+    )
```
naashton, line 16: I want to point this out, in case this is the incorrect use of `PaymentToken`.

test/remote/gateways/remote_aafes_test.rb
```diff
@@ -0,0 +1,166 @@
+require 'test_helper'
+
+class RemoteAafesTest < Test::Unit::TestCase
+  def setup
+    @gateway = AafesGateway.new(fixtures(:aafes))
+    @gateway.ssl_strict = false # Fails to send request to gateway otherwise :shrug
```
therufs, line 6: 😬

test/remote/gateways/remote_aafes_test.rb
therufs: It doesn't look like there are any other examples of `PaymentToken` getting used as a top-level citizen; I wonder if we need to make a fresh token type?

lib/active_merchant/billing/gateways/aafes.rb
```diff
@@ -0,0 +1,172 @@
...
+          xml['cm'].OrderNumber(options[:order_id])
+          xml['cm'].transactionId(options[:transaction_id])
+          xml['cm'].termId(options[:term_id])
+          xml['cm'].Comment(options[:comment])
+          xml['cm'].CustomerID(options[:customer_id])
+        end
+      end
+
+      def add_milstar_auth(xml, amount, payment, options)
+        xml['cm'].Request('RRN' => options[:rrn]) do
```
leila-alderman, line 153: Should we have some sort of check in place to ensure that the RRN is indeed unique?

lib/active_merchant/billing/gateways/aafes.rb
```diff
@@ -0,0 +1,172 @@
...
+          xml['cm'].Media('Milstar')
+          xml['cm'].RequestType('Sale')
+          xml['cm'].InputType('Keyed')
+          xml['cm'].Token('Token')
+          xml['cm'].Account(payment.payment_data)
+          xml['cm'].Expiration(payment.metadata[:expiration])
```
leila-alderman, line 159: I'm not entirely sure where this expiration date information is coming from, but would it be appropriate to follow [the way that Ruthan implemented adding expiration dates to third-party vaulted payment methods for Litle](https://github.com/activemerchant/active_merchant/pull/3606/files)?

test/remote/gateways/remote_aafes_test.rb
```diff
@@ -0,0 +1,148 @@
...
+    @amount = '%.2f' % 100
```
leila-alderman, line 7: This seems like an odd workaround. I would expect this logic to be included in the gateway code, not within the remote test code.

test/unit/gateways/aafes_test.rb
```diff
@@ -0,0 +1,186 @@
...
+  # end
+
+  # def test_scrub
+  #   # assert @gateway.supports_scrubbing?
+  #   # assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
+  # end
+
+  private
+
+  def pre_scrubbed
```
leila-alderman, line 95: Is this needed? Based on what I know for AAFES, we won't need to scrub the transcript, so I'm not sure this will even be needed down the road.

test/unit/gateways/aafes_test.rb
```diff
@@ -0,0 +1,186 @@
...
+      read 925 bytes
+      Conn close
+    TRANSCRIPT
+  end
+
+  # def post_scrubbed
+  # end
+
+  def successful_authorize_response
+    <<-XML
```
leila-alderman, line 122:
```suggestion
    <<~XML
```

test/unit/gateways/aafes_test.rb
```diff
@@ -0,0 +1,186 @@
...
+    <Message TypeCode="Response" MajorVersion="3" MinorVersion="4" FixVersion="0" xmlns="http://www.aafes.com/credit">
```
leila-alderman, line 123: This response should all be indented an additional tab to get ahead of [my RuboCop PR](https://github.com/activemerchant/active_merchant/pull/3685).

test/unit/gateways/aafes_test.rb
```diff
@@ -0,0 +1,186 @@
...
+          <DescriptionField>APPROVED  </DescriptionField>
+          <origReqType>Sale</origReqType>
+      </Response>
+    </Message>
+    XML
+  end
+
+  def failed_authorize_response
+    <<-XML
+    <Message TypeCode="Response" MajorVersion="3" MinorVersion="4" FixVersion="0" xmlns="http://www.aafes.com/credit">
```
leila-alderman, line 149: This response should all be indented an additional tab to get ahead of [my RuboCop PR](https://github.com/activemerchant/active_merchant/pull/3685).

test/unit/gateways/aafes_test.rb
```diff
@@ -0,0 +1,186 @@
...
+          <PlanNumber>10001</PlanNumber>
+          <DescriptionField>APPROVED  </DescriptionField>
+          <origReqType>Sale</origReqType>
+      </Response>
+    </Message>
+    XML
+  end
+
+  def failed_authorize_response
+    <<-XML
```
leila-alderman, line 148:
```suggestion
    <<~XML
```

lib/active_merchant/billing/gateways/aafes.rb
```diff
@@ -0,0 +1,172 @@
...
+      end
+
+      def add_headers(xml, options)
+        xml['cm'].Header do
+          xml['cm'].IdentityUUID(@identity_uuid)
+          xml['cm'].LocalDateTime(Time.now.utc.iso8601)
+          xml['cm'].SettleIndicator(false)
+          xml['cm'].OrderNumber(options[:order_id])
+          xml['cm'].transactionId(options[:transaction_id])
+          xml['cm'].termId(options[:term_id])
```
leila-alderman, line 146: Not at all required to change, but the parentheses here are all optional:

```suggestion
          xml['cm'].IdentityUUID         @identity_uuid
          xml['cm'].LocalDateTime      Time.now.utc.iso8601
          xml['cm'].SettleIndicator       false
          xml['cm'].OrderNumber         options[:order_id]
          xml['cm'].transactionId          options[:transaction_id]
          xml['cm'].termId                     options[:term_id]
```

lib/active_merchant/billing/gateways/aafes.rb
naashton: Soo, I'm not actually sure how we can check that, because it's got to be unique based on what AAFES has in their database (and I assume it's per `identity_uuid`). This is going to be the tricky part of this implementation. Jacob's recommendation was that we can just use a Spreedly transaction token, trim it down to 12 chars, and use that. As far as verifying goes, I'm unsure of how we would do that without creating some database table that keeps track of them for 30 days.

naashton: My thought on this, is this is a Core side issue.

lib/active_merchant/billing/gateways/aafes.rb
naashton: The expiration date will be coming from Priceline, along with the Milstar token and billing zip. And Priceline gets the expiration from AAFES in the format of `YYMM`. I don't think we need to do any tinkering with it, since AAFES gives it to us in the format they expect it.

lib/active_merchant/billing/gateways/aafes.rb
```diff
@@ -0,0 +1,172 @@
...
+          xml['cm'].InputType('Keyed')
+          xml['cm'].Token('Token')
+          xml['cm'].Account(payment.payment_data)
+          xml['cm'].Expiration(payment.metadata[:expiration])
+          xml['cm'].AmountField(amount)
+          xml['cm'].PlanNumbers do
+            xml['cm'].PlanNumber(options[:plan_number])
+          end
+          xml['cm'].DescriptionField(options[:description])
+          xml['cm'].AddressVerificationService do
```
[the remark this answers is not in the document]
naashton: So, that first link you provided, is actually the one AAFES stepped through with us. I believe that the XML needs to be formatted this way as it is the parent tag to the address data.

lib/active_merchant/billing/gateways/aafes.rb
```diff
@@ -0,0 +1,172 @@
...
+            xml['cm'].BillingZipCode(payment.metadata[:zip])
```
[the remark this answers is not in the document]
naashton: So, I just added a test for this in the gateway extensions gem. I'm fairly certain billing zip is required and I wrote a test to reflect that.

lib/active_merchant/billing/payment_token.rb
```diff
@@ -7,6 +7,7 @@ module Billing #:nodoc:
     # the token's type, like 'apple_pay' or 'stripe' should your target payment gateway process these tokens.
     class PaymentToken
       attr_reader :payment_data
+      attr_reader :metadata
```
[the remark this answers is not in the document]
naashton: This was really a hacky way of not being able to directly mimic a third party payment token. I'm not sure I should be doing this, but given this work is now isolated in its own gem, I don't feel it's too much of a risk.

test/remote/gateways/remote_aafes_test.rb
naashton: I took this out, not sure what was going on when I initially tested this.

PR #3671, opened 2020-06-15
Adyen: Set executeThreeD to false as default in certain situations 

naashton:
Give a default behavior to executeThreeD by setting it to false if
threed_dynamic is false and execute_threed is nil

CE-550

Unit: 67 tests, 319 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 90 tests, 350 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -413,8 +413,12 @@ module ActiveMerchant #:nodoc:
             post[:threeDS2RequestData] = { deviceChannel: device_channel, notificationURL: three_ds_2_options[:notification_url] }
           end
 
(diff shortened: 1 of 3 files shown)


naashton:
I think it would still come back to making it a GSF. Considering `execute_threed` will always be `nil`, if `threed_dynamic` is set to true, according to this line:
https://github.com/spreedly/core/blob/8699d488f63650255969a1c0438c9078e79486a6/app/models/gateway_extensions/threeds/adyen.rb#L91

So, in any case, it looks like one of those gets passed in as true. And given that, I think this PR is useless, unless I'm missing something.

PR #3678, opened 2020-06-18
Mercado-Pago: Notification url GSF

cdmackeyfree:
Add notification_url GSF, also update Mercado-Pago test card values

Unit:
Finished in 0.033803 seconds.
37 tests, 171 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote 
 Finished in 67.755391 seconds.
31 tests, 85 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
83.871% passed
0.46 tests/s, 1.25 assertions/s

All Unit tests:
  Finished in 11.617083 seconds.
4517 tests, 72070 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
388.82 tests/s, 6203.79 assertions/s

Tests unrelated to this work are still failing, so that should be addressed in another ticket

### DIFF
--- a/lib/active_merchant/billing/gateways/mercado_pago.rb
+++ b/lib/active_merchant/billing/gateways/mercado_pago.rb
@@ -98,6 +98,7 @@ module ActiveMerchant #:nodoc:
         add_processing_mode(post, options)
         add_net_amount(post, options)
         add_taxes(post, options)
+        add_notification_url(post, options)
         post[:binary_mode] = (options[:binary_mode].nil? ? true : options[:binary_mode])
         post
       end
@@ -201,6 +202,10 @@ module ActiveMerchant #:nodoc:
 
       def add_net_amount(post, options)
         post[:net_amount] = Float(options[:net_amount]) if options[:net_amount]
+      end
+
+      def add_notification_url(post, options)
+        post[:notification_url] = options[:notification_url] if options[:notification_url]
       end
 
       def add_taxes(post, options)
--- a/test/remote/gateways/remote_mercado_pago_test.rb
+++ b/test/remote/gateways/remote_mercado_pago_test.rb
@@ -7,7 +7,7 @@ class RemoteMercadoPagoTest < Test::Unit::TestCase
     @colombian_gateway = MercadoPagoGateway.new(fixtures(:mercado_pago_colombia))
 
     @amount = 500
-    @credit_card = credit_card('4509953566233704')
+    @credit_card = credit_card('5031433215406351')
     @colombian_card = credit_card('4013540682746260')
     @elo_credit_card = credit_card('5067268650517446',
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -114,6 +114,7 @@ def add_processing_mode(post, options)
         post[:processing_mode] = options[:processing_mode]
         post[:merchant_account_id] = options[:merchant_account_id] if options[:merchant_account_id]
         post[:payment_method_option_id] = options[:payment_method_option_id] if options[:payment_method_option_id]
+        post[:notification_url] = options[:notification_url] if options[:notification_url]
```
leila-alderman, line 117: Looking at [Mercado Pago's documentation for the notification URL](https://www.mercadopago.com.co/developers/es/guides/notifications/webhooks/), this field isn't at all related to the processing mode. Therefore, this field is not dependent on the `processing_mode` option being set, which is the requirement for adding the fields in this method.

This field can either be added directly to the `purchase_request`, etc. methods above or be added as its own new method and then that method can be added to the `purchase_request`, etc. methods above.

test/remote/gateways/remote_mercado_pago_test.rb
```diff
@@ -43,7 +44,8 @@ def setup
       merchant_account_id: fixtures(:mercado_pago)[:merchant_account_id],
       fraud_scoring: true,
       fraud_manual_review: true,
-      payment_method_option_id: '123abc'
+      payment_method_option_id: '123abc',
+      notification_url: 'www.mercado-pago.com'
```
leila-alderman, line 48: Because this field isn't dependent on the processing mode (and because that's one of the remote tests that's currently failing), could you write a fresh remote test to just test this one new parameter?

test/unit/gateways/mercado_pago_test.rb
```diff
@@ -80,6 +81,16 @@ def test_successful_purchase_with_naranja
     assert response.test?
   end
 
+  # def test_successful_purchase_with_notification_url
```
leila-alderman, line 84: It seems like the new unit test for this new field shouldn't be commented out?

Also, because this is a unit test for a GSF, you will likely need to use the `stub_comms` method to test what is being sent in the request instead of what's coming back in the response since GSFs are often not sent back in the response from the gateway.

See the [other unit test examples](https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/mercado_pago_test.rb#L288).

cdmackeyfree: Oops that was definitely not intentional

PR #3680, opened 2020-06-19
NMI: Add support for collect.js generated payment tokens

david-gamar:
NMI: Add support for tokenized card data generated with Collect.js as documented here:
https://secure.networkmerchants.com/gw/merchants/resources/integration/integration_portal.php#cjs_methodology.


curiousepic:

PR #3686, opened 2020-06-23
Worldpay: exclude empty inst_id

bbraschi:
Exclude empty `inst_id` as per recommendation from Worldpay:

> \<order orderCode="c12999949615176.2" installationId=""\>
> If there is not installation ID, you should not be including this data. You can instead just submit:
> \<order orderCode="c12999949615176.2"\>
> The response with NMTOKEN must be a name token usually means a value is missing. Anything with "" and nothing is in the middle of the quotes typically will produce this error.

### DIFF
--- a/test/unit/gateways/worldpay_test.rb
+++ b/test/unit/gateways/worldpay_test.rb
@@ -649,6 +649,11 @@ class WorldpayTest < Test::Unit::TestCase
     assert_failure response
   end
 
+  def test_empty_inst_id_is_stripped
+    @gateway.expects(:ssl_post).with { |_, request, _| request !~ /installationId/ }.returns(successful_authorize_response)
+    @gateway.authorize(@amount, @credit_card, @options.merge({ inst_id: '' }))
+  end
+
   def test_3ds_name_coersion
     @options[:execute_threed] = true
     response = stub_comms do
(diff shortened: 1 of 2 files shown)


test/unit/gateways/worldpay_test.rb
```diff
@@ -649,6 +649,11 @@ def test_failed_verify
     assert_failure response
   end
 
+  def test_empty_inst_id_is_stripped
+    @gateway.expects(:ssl_post).with { |_, request, _| request !~ /installationId/ }.returns(successful_authorize_response)
```
nebdil, line 653: TIL

bbraschi: Tx Dilan.

Not the clearest... If you know a better way, I take it! I wanted to use something like
```
with(anything, regexp_matches(/NOT installationId/), anything)
``` 
but did not find a way to inject a regular expression that would match if a string is NOT present. The 
[Regexp doc](https://ruby-doc.org/core-2.5.1/Regexp.html) has `Negative lookahead` etc., but it was even more cryptic, and I did not make work anyway :-).


I made sure the test failed before the fix.

leila-alderman: @bbraschi It's not required, but if you'd like to implement this in a cleaner way, I would recommend using the `stub_comms do` combined with `assert_not_match`. You can find examples of this approach in other unit tests in this file for reference.

PR #3695, opened 2020-07-08
FirstdataE4V27Gateway: Fix strip_line_breaks method

Carrigan:
## Why?

The recently added strip_line_breaks method added the requirement that all values in the shipping hash be strings. It is often the case though that nil values are supplied, especially for fields like address2. 

## What Changed?

This fix changes the function to allow for nil values.

No test changes were made since the function is already under test with test_requests_scrub_newline_and_return_characters_from_verification_string_components

## Test Suite

4533 tests, 72185 assertions, 0 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications
100% passed
692 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
+++ b/lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
@@ -304,7 +304,7 @@ module ActiveMerchant #:nodoc:
       def strip_line_breaks(address)
         return unless address.is_a?(Hash)
 
-        Hash[address.map { |k, s| [k, s.tr("\r\n", ' ').strip] }]
+        Hash[address.map { |k, s| [k, s&.tr("\r\n", ' ')&.strip] }]
       end
 
       def add_invoice(xml, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/firstdata_e4_v27.rb
```diff
@@ -304,7 +304,7 @@ def add_address(xml, options)
       def strip_line_breaks(address)
         return unless address.is_a?(Hash)
 
-        Hash[address.map { |k, s| [k, s.tr("\r\n", ' ').strip] }]
+        Hash[address.map { |k, s| [k, s.try(:tr, "\r\n", ' ').strip] }]
```
fatcatt316, line 307: I think the `strip` here will blow up if `s.try...` returns `nil`

Carrigan: Good catch! I added a nil value in the strip line test and it failed. Thanks for the 🦅 👀 

Using your recommendation and modifying the test to make sure it doesn't regress.

PR #3699, opened 2020-07-15
Update logic to send cavv as xid for 3DS2

douglas:
### What are you trying to accomplish?

Update logic to send `cavv` as `xid` for 3DS2 according to `Cybersource` recommendations.

### Context

Pierre (@pi3r) was talking with Cybersource about an error that we had related to the `xid` field and we got this response from them:

```
1.x
  if xid is present, send it
  do not send the cavv in the xid XML element
2.x
  if xid is present, send it
  else if cavv is present, send cavv in the xid XML element for all card brands except :master
```

So this change intends to make the code behave in the recommended way.

### What should reviewers focus on?

Is there anything that I'm missing? Any improvement suggestion?

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -611,16 +611,22 @@ module ActiveMerchant #:nodoc:
 
       def add_normalized_threeds_2_data(xml, payment_method, options)
         threeds_2_options = options[:three_d_secure]
+        cc_brand = card_brand(payment_method).to_sym
+        xid = threeds_2_options[:xid]
 
-        xml.tag!('cavv', threeds_2_options[:cavv]) if threeds_2_options[:cavv] && card_brand(payment_method).to_sym != :master
+        xml.tag!('cavv', threeds_2_options[:cavv]) if threeds_2_options[:cavv] && cc_brand != :master
         xml.tag!('cavvAlgorithm', threeds_2_options[:cavv_algorithm]) if threeds_2_options[:cavv_algorithm]
         xml.tag!('paSpecificationVersion', threeds_2_options[:version]) if threeds_2_options[:version]
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
-        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
+        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[cc_brand])
         xml.tag!('eciRaw', threeds_2_options[:eci]) if threeds_2_options[:eci]
 
-        xid = threeds_2_options[:xid] || threeds_2_options[:cavv]
-        xml.tag!('xid', xid) if xid
+        if xid.present?
+          xml.tag!('xid', xid)
+        elsif threeds_2_options[:version]&.start_with?('2') && cc_brand != :master
+          cavv = threeds_2_options[:cavv]
+          xml.tag!('xid', cavv) if cavv.present?
+        end
 
         xml.tag!('veresEnrolled', threeds_2_options[:enrolled]) if threeds_2_options[:enrolled]
         xml.tag!('paresStatus', threeds_2_options[:authentication_response_status]) if threeds_2_options[:authentication_response_status]
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -927,6 +927,43 @@ class CyberSourceTest < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
+  def test_send_xid_for_3ds_1_regardless_of_cc_brand
+    options_with_normalized_3ds = @options.merge(
+      three_d_secure: {
+        eci: '05',
+        cavv: '637574652070757070792026206b697474656e73',
+        xid: 'this-is-an-xid',
+        ds_transaction_id: '97267598-FAE6-48F2-8083-C23433990FBC',
+        cavv_algorithm: 'vbv'
+      }
+    )
+
+    stub_comms do
+      @gateway.purchase(@amount, @elo_credit_card, options_with_normalized_3ds)
+    end.check_request do |endpoint, data, headers|
+      assert_match(/<xid\>this-is-an-xid/, data)
+    end.respond_with(successful_purchase_response)
+  end
+
+  def test_dont_send_cavv_as_xid_in_3ds2_for_mastercard
+    options_with_normalized_3ds = @options.merge(
+      three_d_secure: {
+        version: '2.0',
+        eci: '05',
+        cavv: '637574652070757070792026206b697474656e73',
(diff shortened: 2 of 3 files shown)


test/unit/gateways/cyber_source_test.rb
```diff
@@ -941,7 +941,7 @@ def test_adds_cavv_as_xid_for_3ds2
     )
 
     stub_comms do
-      @gateway.purchase(@amount, @master_credit_card, options_with_normalized_3ds)
+      @gateway.purchase(@amount, @credit_card, options_with_normalized_3ds)
```
douglas, line 944: For additional context, `@credit_card` is a `visa` card.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -611,16 +611,19 @@ def add_auth_service(xml, payment_method, options)
...
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
-        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
+        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[cc_brand])
         xml.tag!('eciRaw', threeds_2_options[:eci]) if threeds_2_options[:eci]
 
-        xid = threeds_2_options[:xid] || threeds_2_options[:cavv]
-        xml.tag!('xid', xid) if xid
+        if cc_brand == :visa
+          xid = threeds_2_options[:xid] || threeds_2_options[:cavv]
+          xml.tag!('xid', xid) if xid
```
bbraschi, line 625: `xid` must be set regardless of the brand, even if it is not `visa`

bbraschi: After chat IRL - I think we need to check what is the expectation fro 3DS 1.x

douglas: Yup, while we wait for their response I added a commit to ensure that we send `xid` regardless of the card brand but allowing following Pierre's previous logic when it is visa and 3DS version 2.

douglas:
Hello @pi3r and @bbraschi! I updated the code with our latest understanding about it and added some comments as well.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -611,15 +611,25 @@ def add_auth_service(xml, payment_method, options)
...
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
-        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
+        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[cc_brand])
         xml.tag!('eciRaw', threeds_2_options[:eci]) if threeds_2_options[:eci]
 
-        xid = threeds_2_options[:xid] || threeds_2_options[:cavv]
+        xid = case cc_brand
+              when :master
+                # Only send xid for mastercard on 3ds 1.x
+                threeds_2_options[:xid] if threeds_2_options[:version].nil?
```
bbraschi, line 626: My understanding is:
- 1.x 
   - if `xid` is present, send it
   - do not send the `cavv` in the `xid` XML element
- 2.x 
   - if `xid` is present, send it
   - else if `cavv`  is present, send `cavv` in the `xid` XML element for all card brands except `:master`


Does that match yours? 
If it does, I read the code is doing something else.

bbraschi, line 626: `version` is not `nil` for 1.x, at least not always.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -611,15 +611,25 @@ def add_auth_service(xml, payment_method, options)
...
+              when :visa
+                # Visa can use :cavv for xid when the later is not present
+                threeds_2_options[:xid] || threeds_2_options[:cavv]
```
bbraschi, line 629: That would send `cavv` in the `xid` XML element
- in `1.x` also
- only to `visa`

no?

pi3r:
What Bertrand said ^

lib/active_merchant/billing/gateways/cyber_source.rb
douglas: Oh, I misunderstood the logic then, thanks for the clarification!

lib/active_merchant/billing/gateways/cyber_source.rb
douglas: Yes, I misunderstood that the change in the behaviour was for 3DS2 - good catch 🎣

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -611,16 +611,22 @@ def add_auth_service(xml, payment_method, options)
...
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
-        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
+        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[cc_brand])
         xml.tag!('eciRaw', threeds_2_options[:eci]) if threeds_2_options[:eci]
 
-        xid = threeds_2_options[:xid] || threeds_2_options[:cavv]
-        xml.tag!('xid', xid) if xid
+        if xid.present?
+          xml.tag!('xid', xid)
+        elsif threeds_2_options[:version] == '2.0' && cc_brand != :master
```
pi3r, line 626: I think you should regex on `^2`

bbraschi: or `start_with?('2')` because we have 2.1, 2.2 etc?

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -611,16 +611,22 @@ def add_auth_service(xml, payment_method, options)
...
         xml.tag!('cavvAlgorithm', threeds_2_options[:cavv_algorithm]) if threeds_2_options[:cavv_algorithm]
         xml.tag!('paSpecificationVersion', threeds_2_options[:version]) if threeds_2_options[:version]
         xml.tag!('directoryServerTransactionID', threeds_2_options[:ds_transaction_id]) if threeds_2_options[:ds_transaction_id]
-        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[card_brand(payment_method).to_sym])
+        xml.tag!('commerceIndicator', options[:commerce_indicator] || ECI_BRAND_MAPPING[cc_brand])
         xml.tag!('eciRaw', threeds_2_options[:eci]) if threeds_2_options[:eci]
 
-        xid = threeds_2_options[:xid] || threeds_2_options[:cavv]
-        xml.tag!('xid', xid) if xid
+        if xid.present?
```
bbraschi, line 624: ❤️

PR #3701, opened 2020-07-22
[Cybersource] Set authorization on the response when in Fraud Review

pi3r:
We use the `authorization` as a reference when voiding/refunding a transaction. `authorization` isn't set when the transaction is in review which prevents later void/refund.


pi3r:
@bbraschi Indeed!

PR #3702, opened 2020-07-23
Realex: Change 3DSecure v1 message_version to a valid format

Shuhala:
Major/minor versions (ex: 1.0.2) in the `message_version` tag is not accepted for 3D Secure v1 for the Realex gateway. For a valid 3D Secure v1 request `message_version` must be either 1 or blank.

### DIFF
--- a/lib/active_merchant/billing/gateways/realex.rb
+++ b/lib/active_merchant/billing/gateways/realex.rb
@@ -316,6 +316,7 @@ module ActiveMerchant
           else
             xml.tag! 'cavv', three_d_secure[:cavv]
             xml.tag! 'xid', three_d_secure[:xid]
+            version = '1'
           end
           xml.tag! 'eci', three_d_secure[:eci]
           xml.tag! 'message_version', version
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/realex.rb
```diff
@@ -318,7 +318,7 @@ def add_three_d_secure(xml, options)
             xml.tag! 'xid', three_d_secure[:xid]
           end
           xml.tag! 'eci', three_d_secure[:eci]
-          xml.tag! 'message_version', version
+          xml.tag! 'message_version', version.sub(/^1.*/, '1')
```
bbraschi, line 321: I think that updating `version` within the branch that is specific to 1.x  helps because you do not have to question yourself about the validity of the `version.sub(/^1.*/, '1')` statement for 2.x.

Eg:
```ruby
          if /^2/.match?(version)
            xml.tag! 'authentication_value', three_d_secure[:cavv]
            xml.tag! 'ds_trans_id', three_d_secure[:ds_transaction_id]
          else
            xml.tag! 'cavv', three_d_secure[:cavv]
            xml.tag! 'xid', three_d_secure[:xid]
            version = '1'
          end
          xml.tag! 'eci', three_d_secure[:eci]
          xml.tag! 'message_version', version
```

PR #2383, opened 2017-03-30
Paypal Express: support order actions

whitby3001:
Adds support for setting up and processing the order action

PR #3703, opened 2020-07-25
Paypal Express: support for "Order"  PaymentAction

rmatejacabello:
Paypal Express supports 3 different payment actions, but active merchant only includes 2: Sale and Authorization. 
This is adding support for the 3rd payment action "Order". 

You can find detailed info from paypal in the following sources:
https://developer.paypal.com/docs/archive/express-checkout/integration-guide/ECSettlements/#order-payment-action
https://www.paypal.com/uk/smarthelp/article/what-are-the-differences-between-the-express-checkout-payment-actions-ts1501

There was an original pull request from @whitby3001 that got never included and was closed by David, but it's still relevant.
https://github.com/activemerchant/active_merchant/pull/2383

### DIFF
--- a/lib/active_merchant/billing/gateways/paypal/paypal_express_response.rb
+++ b/lib/active_merchant/billing/gateways/paypal/paypal_express_response.rb
@@ -60,6 +60,22 @@ module ActiveMerchant #:nodoc:
       def note
         @params['note_text']
       end
+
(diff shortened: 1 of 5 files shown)


kenikadawan0129 (approved):
I'm not understanding this part?

PR #3705, opened 2020-07-28
Realex: add missing change log for #3702

bbraschi:


CHANGELOG
```diff
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* Realex: Change 3DSecure v1 message_version to a valid format #3702
```
bbraschi, line 4: Not clear what is the expected thing to do: add a line in the `CHANGELOG` for each commit or update `CHANGELOG` only at release time. `CHANGELOG` includes the PR number, so it would need to be updated once the PR number is known.
Will update CONTRIBUTING.md if necessary.


@pi3r ?

pi3r: It's better to do it for every contribution but we should still double check its accuracy will creating a new release.

PR #3706, opened 2020-07-28
Adyen 3DS 1.x: use directory_response if no authentication_response

bbraschi:
WIP


bbraschi:
From Ayden:
> The fallback to directory_response will not work as this only gives an indicator about enrolment of the card but no indicator about the actual authentication result.
3d
What we are now looking is internally to fallback to ECI value as there we can retrieve the authentication result as well. This will require no change in active merchants

-> closing

PR #3710, opened 2020-08-03
Add pry-byebug development dependency

douglas:
### What are you trying to accomplish?

While debugging recent PRs we (@bbraschi  and I) noticed that it was not possible to do step-by-step debugging and stack navigation, so this PR adds it.

### How did you accomplish this and why did you choose this approach?

Just added `pry-byebug` as a development dependency.

### How can this PR be tophatted:

Just running the tests with a breakpoint set `require 'pry'; binding.pry` and then use the commands `step` and `continue`.

### What should reviewers focus on?

Any other way of doing this? Any suggestions?

### DIFF
--- a/activemerchant.gemspec
+++ b/activemerchant.gemspec
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency('mocha', '~> 1')
   s.add_development_dependency('pry')
+  s.add_development_dependency('pry-byebug')
   s.add_development_dependency('rake')
   s.add_development_dependency('test-unit', '~> 3')
   s.add_development_dependency('thor')


activemerchant.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency('mocha', '~> 1')
   s.add_development_dependency('pry')
+  s.add_development_dependency('pry-byebug')
```
bbraschi, line 32: does the lock file not need to be updated?

douglas: It is not needed since it is defined in the `gemspec` - also, there is not `lock` file in AM.

bbraschi: I did not notice that the `Gemfile.lock` was not in GH, just local. Tx so much!

PR #3712, opened 2020-08-03
Cybersource: do not send 3DS fields if 'cavv` is missing and `commerceIndicator` is inferred

bbraschi:
### Problem
A merchant can attempt to authorize a non-3DS-authenticated transaction but the liability shift may not occur under the same conditions as it would for a fully 3DS-authenticated transaction.

This is the case, for example, in the `Failed Frictionless Authentication` test with the `4000000000001018` credit card, [here](https://developer.cybersource.com/library/documentation/dev_guides/Payer_Authentication_SO_API/html/index.html#t=Topics%2FTest_Cases_for_3D_Secure_2_x.htm).

If a merchant does not want to attempt to authorize a non-3DS-authenticated transaction, then the call to authorize the transaction through `active_merchant` should not be made and there won't be a problem.

On the other hand, a merchant may want to authorize a non-3DS-authenticated transaction and a corresponding call to `active_merchant` should be made. If no `commerceIndicator` is specified by the caller, `active_merchant` infers a `commerceIndicator` that indicates the authentication has occurred, that is the problem.
In the `4000000000001018` example, `active_merchant` infers `vbv` for the `commerceIndicator`:

```xml
      <ccAuthService run="true">
        <paSpecificationVersion>2.2.0</paSpecificationVersion>
        <directoryServerTransactionID>...</directoryServerTransactionID>
        <commerceIndicator>vbv</commerceIndicator>
        <eciRaw>07</eciRaw>
        <veresEnrolled>Y</veresEnrolled>
      </ccAuthService>
```
As explained by Michael T. from cybersource
> when you pass in “vbv” as the commerceIndicator [the cybersource backend] looks for the presence of a CAVV, since “vbv” means the request was successfully authenticated

and the payment fails with
```xml
    <c:replyMessage xmlns:c="urn:schemas-cybersource-com:transaction-data-1.164">
      ...
      <c:decision>REJECT</c:decision>
      <c:reasonCode>101</c:reasonCode>
      <c:missingField>c:xid</c:missingField>
      ...
      <c:ccAuthReply>
        <c:reasonCode>101</c:reasonCode>
      </c:ccAuthReply>
    </c:replyMessage>
```

### Fix

If the
1. `cavv` is absent and
1. `commerceIndicator` is absent and is inferred by `active_merchant`

then do not add the 3DS fields to the `ccAuthService` xml element.

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -612,6 +612,9 @@ module ActiveMerchant #:nodoc:
       def add_normalized_threeds_2_data(xml, payment_method, options)
         threeds_2_options = options[:three_d_secure]
         cc_brand = card_brand(payment_method).to_sym
+
+        return if threeds_2_options[:cavv].blank? && infer_commerce_indicator?(options, cc_brand)
+
         xid = threeds_2_options[:xid]
 
         xml.tag!('cavv', threeds_2_options[:cavv]) if threeds_2_options[:cavv] && cc_brand != :master
@@ -630,6 +633,10 @@ module ActiveMerchant #:nodoc:
 
         xml.tag!('veresEnrolled', threeds_2_options[:enrolled]) if threeds_2_options[:enrolled]
         xml.tag!('paresStatus', threeds_2_options[:authentication_response_status]) if threeds_2_options[:authentication_response_status]
+      end
+
+      def infer_commerce_indicator?(options, cc_brand)
+        options[:commerce_indicator].blank? && ECI_BRAND_MAPPING[cc_brand].present?
       end
 
       def add_threeds_2_ucaf_data(xml, payment_method, options)
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -836,7 +836,7 @@ class CyberSourceTest < Test::Unit::TestCase
       stub_comms do
         @gateway.purchase(@amount, @credit_card, @options.merge(three_d_secure: { cavv: 'anything but empty' }))
       end.check_request do |endpoint, data, headers|
-        assert_match(/commerceIndicator\>#{CyberSourceGateway::ECI_BRAND_MAPPING[brand.to_sym]}/, data)
+        assert_match(/commerceIndicator\>#{CyberSourceGateway::ECI_BRAND_MAPPING[brand.to_sym]}</, data)
       end.respond_with(successful_purchase_response)
     end
   end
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -612,6 +612,9 @@ def add_auth_service(xml, payment_method, options)
       def add_normalized_threeds_2_data(xml, payment_method, options)
         threeds_2_options = options[:three_d_secure]
         cc_brand = card_brand(payment_method).to_sym
+
+        return if !threeds_2_options[:cavv].present? && infer_commerce_indicator?(options, cc_brand)
```
douglas, line 616: Perhaps use `empty?` instead of `!present?` to increase legibility?

```suggestion
        return if threeds_2_options[:cavv].empty? && infer_commerce_indicator?(options, cc_brand)
```

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -632,6 +635,10 @@ def add_normalized_threeds_2_data(xml, payment_method, options)
         xml.tag!('paresStatus', threeds_2_options[:authentication_response_status]) if threeds_2_options[:authentication_response_status]
       end
 
+      def infer_commerce_indicator?(options, cc_brand)
+        !options[:commerce_indicator].present? && ECI_BRAND_MAPPING[cc_brand].present?
```
douglas, line 639: Same here:

```suggestion
        options[:commerce_indicator].empty? && ECI_BRAND_MAPPING[cc_brand].present?
```

test/unit/gateways/cyber_source_test.rb
```diff
@@ -877,6 +877,69 @@ def test_adds_3ds2_fields_via_normalized_hash
...
+      assert_not_match(/<cavvAlgorithm\>#{options[:three_d_secure][:cavv_algorithm]}</, data)
+      assert_not_match(/<veresEnrolled\>#{options[:three_d_secure][:enrolled]}</, data)
+      assert_not_match(/<commerceIndicator\>#{options[:commerce_indicator]}</, data)
+    end.respond_with(successful_purchase_response)
+  end
+
+  def test_adds_3ds2_fields_via_normalized_hash_when_cavv_and_commerce_indicator_absent_and_commerce_indicator_not_inferred
+    @credit_card.brand =
+      (ActiveMerchant::Billing::CyberSourceGateway.supported_cardtypes -
+        ActiveMerchant::Billing::CyberSourceGateway::ECI_BRAND_MAPPING.keys).first
```
douglas, line 899: Hello Bertrand! Could you add some context on why we need to do this? When I debugged locally the card brand that was returned was `diners_club`, is this expected?

bbraschi: done

bbraschi: done

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -632,6 +635,10 @@ def add_normalized_threeds_2_data(xml, payment_method, options)
         xml.tag!('paresStatus', threeds_2_options[:authentication_response_status]) if threeds_2_options[:authentication_response_status]
       end
 
+      def infer_commerce_indicator?(options, cc_brand)
+        options[:commerce_indicator].blank? && ECI_BRAND_MAPPING[cc_brand].present?
```
bbraschi, line 639: Asked cybersource what are the commerce indicators for the supported brands absent from `ECI_BRAND_MAPPING`. If a commerce indicator could be inferred for all the supported brands, this code could be simplified.

PR #3714, opened 2020-08-04
Adyen: Pass `subMerchantId` as `additionalData`

naashton:
Payment facilitators must pass this value for auth/purchase transactions

CE-839

Unit:
69 tests, 325 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
91 tests, 351 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,13 +3,13 @@
 == HEAD
 * Cybersource: add `maestro` and `diners_club` eci brand mapping [bbraschi] #3708
 * Cybersource: Ensure Partner Solution Id placement conforms to schema [britth] #3715
+* Adyen: Adyen: Pass `subMerchantId` as `additionalData` [naashton] #3714
 
 == Version 1.111.0
 * Fat Zebra: standardized 3DS fields and card on file extra data for Visa scheme rules [montdidier] #3409
 * Realex: Change 3DSecure v1 message_version to a valid format [shuhala] #3702
 * Ingenico/ GlobalCollect: Add field for installments [cdmackeyfree] #3707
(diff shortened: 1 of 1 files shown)


naashton:
The random line from `global_collect` was a lint thing.

test/unit/gateways/adyen_test.rb
```diff
@@ -810,6 +810,16 @@ def test_authorize_with_network_tokenization_credit_card
     assert_success response
   end
 
+  def test_authorize_with_sub_merchant_id
```
therufs, line 813: Does this test actually pass `sub_merchant_id`? :confused_england:

naashton: Hah, whoops. It is not actually testing whether `sub_merchant_id` gets passed.

test/unit/gateways/adyen_test.rb
```diff
@@ -810,6 +810,17 @@ def test_authorize_with_network_tokenization_credit_card
...
+    response = stub_comms do
+      @gateway.authorize(@amount, @credit_card, @options.merge(sub_merchant_id: '12345abcde67890'))
+    end.check_request do |endpoint, data, headers|
+      parsed = JSON.parse(data)
+      puts parsed['additionalData']
```
therufs, line 818: Might be good form to remove this `puts`. Otherwise lgtm!

naashton: It's just one of those days

therufs: That's why we have code review. :D

PR #3715, opened 2020-08-05
Cybersource: Ensure Partner Solution Id placement conforms to schema

britth:
The Cybersource XML request must follow the defined [xsd schema format](https://ics2wstest.ic3.com/commerce/1.x/transactionProcessor/CyberSourceTransaction_1.167.xsd) 
or else it will result in an XML parse error. Currently, if someone
tries to use stored credential fields in addition to the partner
solution id, they will receive an XML parse error from the gateway. The 
schema requires `subsequentAuth` to come before the solution id and 
for the other stored credential fields to come after solution id. This PR 
updates the request building methods to ensure that these fields are 
added in the right order, also adding remote tests for this behavior.

Remote:
89 tests, 457 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.382% passed
(pre-existing failures)

Unit:
96 tests, 463 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -280,10 +280,11 @@ module ActiveMerchant #:nodoc:
         add_threeds_services(xml, options)
         add_payment_network_token(xml) if network_tokenization?(creditcard_or_reference)
         add_business_rules_data(xml, creditcard_or_reference, options)
+        add_stored_credential_subsequent_auth(xml, options)
+        add_partner_solution_id(xml)
         add_stored_credential_options(xml, options)
         add_issuer_additional_data(xml, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -843,20 +846,27 @@ def lookup_country_code(country_field)
         country_code&.code(:alpha2)
       end
 
-      def add_stored_credential_options(xml, options={})
+      def add_stored_credential_subsequent_auth(xml, options={})
```
britth, line 849: couldn't really think of a better way to refactor this, so any other ideas welcome!

britth:
thanks @curiousepic !

PR #3723, opened 2020-08-11
RuboCop: Fix Naming/ConstantName

leila-alderman:
Fixes the RuboCop to-do for [Naming/ConstantName](https://docs.rubocop.org/rubocop/0.85/cops_naming.html#namingconstantname).

In this case, RuboCop was incorrectly thinking that `AssertionClass`
should be a constant, likely because of the capitalization used for this
variable. To fix this issue, this variable has been renamed as an
instance variable.

All unit tests:
4543 tests, 72248 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -31,7 +31,7 @@ end
 
 module ActiveMerchant
   module Assertions
-    AssertionClass = defined?(Minitest) ? MiniTest::Assertion : Test::Unit::AssertionFailedError
+    ASSERTION_CLASS = defined?(Minitest) ? MiniTest::Assertion : Test::Unit::AssertionFailedError
 
     def assert_field(field, value)
       clean_backtrace do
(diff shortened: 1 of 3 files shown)


test/test_helper.rb
```diff
@@ -31,7 +31,7 @@ class SubclassGateway < SimpleTestGateway
 
 module ActiveMerchant
   module Assertions
-    AssertionClass = defined?(Minitest) ? MiniTest::Assertion : Test::Unit::AssertionFailedError
+    @assertion_class = defined?(Minitest) ? MiniTest::Assertion : Test::Unit::AssertionFailedError
```
Carrigan, line 34: This actually subtly changes this entity from a constant to a class-level variables, which makes it mutable. Given how it is used, I don't _think_ this matters, but what do you think about changing `AssertionClass` to `ASSERTION_CLASS` in order to make Rubocop happy and keep it as a constant just in case?

leila-alderman: Thanks for the suggestion! That does indeed appear to appease the mighty RuboCop. 🎉

therufs (approved):
woot!

leila-alderman:
Yay! Thanks for the review @therufs!

Carrigan:
Looks great - thanks for changing the name and for getting rid of this warning 🎉

PR #3733, opened 2020-08-14
Cybersource: Ensure issueradditionaldata comes before partnerSolutionId

britth:
XML parse error observed if this field comes after partnerSolutionId. Switch order and add tests. Also check merchant_descriptor in case that could cause issues.

Unit:
96 tests, 467 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
93 tests, 481 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.6237% passed


britth:
thanks @curiousepic !

PR #3734, opened 2020-08-14
Orbital: Add cardIndicators field

meagabeth:
CE-710

Unit:
94 tests, 557 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
36 tests, 185 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
88.8889% passed

Unrelated, previously failing remote tests:
- test_successful_american_express_authorization_with_3ds
- test_successful_american_express_purchase_with_3ds
- test_successful_master_authorization_with_3ds
- test_successful_master_purchase_with_3ds

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -199,6 +199,7 @@ module ActiveMerchant #:nodoc:
             add_customer_data(xml, creditcard, options)
             add_managed_billing(xml, options)
           end
+          add_card_indicators(xml, options) if @options[:card_indicators]
         end
         commit(order, :authorize, options[:trace_number])
       end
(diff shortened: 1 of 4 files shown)


leila-alderman:
@meagabeth I noticed your after-the-fact RuboCop linting fixes, and just in case you aren't already aware of it, I just wanted to share that you can run `bundle exec rake test:local` on your AM branch locally, which includes a RuboCop check.

It took me _forever_ to find that function and I hated waiting for the CI to finish to see if RuboCop approved it, so I try to share that knowledge as much as possible! :grin:

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -219,6 +219,7 @@ def purchase(money, creditcard, options = {})
             add_customer_data(xml, creditcard, options)
             add_managed_billing(xml, options)
           end
+          add_card_indicators(xml, options) if @options[:card_indicators]
```
leila-alderman, line 222: Glancing over the Jira ticket, it seems like this field might also need to be added to the `authorize` action above.

meagabeth: Done! Thank you for catching that.

meagabeth:
Branch successfully merged into master; leaving record of these commits for historical record/reference.

PR #3736, opened 2020-08-18
Orbital: Fix typo in PC3DtlLineTot field

naashton:
PC3Dtllinetot changed to PC3DtlLineTot

CE-762

Unit: 93 tests, 553 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 35 tests, 181 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
85.7143% passed

Failed tests appear to be new and unrelated to changes made in Active
Merchant


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -412,7 +412,7 @@ def add_line_items(xml, options={})
               xml.tag! :PC3DtlIndex,  byte_limit(index + 1, 2)
               line_item.each do |key, value|
                 if key == :line_tot
-                  formatted_key = :PC3Dtllinetot
+                  formatted_key = :PC3DtlLineTot
```
therufs, line 415: It looks to me like we can actually just get rid of this clause entirely, since `LineTot` = `:line_tot.to_s.camelize`?

therufs: otherwise lgtm; same tests seem to fail for me on master as on branch. 👍

PR #3737, opened 2020-08-18
Payflow: Use application_id to set buttonsource

britth:
For rev share flagging purposes, Payflow requires the corresponding
code to be included as ExtData field BUTTONSOURCE, similar to Paypal.
[Example structure as follows](https://stackoverflow.com/a/19414626), confirmed by paypal/payflow support:

```
<Sale>
     <PayData>
          (Invoice)
          (Tender)
     </PayData>
     (ExtData)*
</Sale>
```

This PR updates sale and auth requests to set this field appropriately
when application_id is provided

Unit:
49 tests, 243 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
38 tests, 156 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
68.4211% passed
(unrelated failures preexist on master)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * RuboCop: Fix Style/TrailingCommaInHashLiteral [leila-alderman] #3718
 * RuboCop: Fix Naming/PredicateName [leila-alderman] #3724
 * RuboCop: Fix Style/Attr [leila-alderman] #3728
+* Payflow: Use application_id to set buttonsource [britth] #3737
 
 == Version 1.112.0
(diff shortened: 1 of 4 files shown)


britth:
@leila-alderman thanks for the review! 🎉  that ensure just makes sure that in all the other tests the application_id is set back to `nil`. If you comment out the ensure statement and put a binding in a subsequent test and check the value of ActiveMerchant::Billing::PayflowGateway.application_id, it'll still be `partner_id` in the other tests rather than getting reset.

PR #3738, opened 2020-08-18
HPS: Enable refunds using capture txn

britth:
Currently, if you try to issue a refund using a capture as your
reference transaction, you will get an error. This is because the
authorization value here only refers to the capture, with no
connection to the auth. In other adapters, we solve for this by
building up an authorization value that includes references to the
current transaction as well as reference transactions (and potentially
other data). This allows us to parse this field to return the
correct reference transaction for supplemental actions. This PR
updates HPS to build an authorization value including the authorize
transaction id for captures. It also adds in logic to pluck off
the appropriate value when issuing a refund on a capture txn.

Unit:
54 tests, 267 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
52 tests, 139 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
90.3846% passed
(preexisting failures also on master)

### DIFF
--- a/lib/active_merchant/billing/gateways/hps.rb
+++ b/lib/active_merchant/billing/gateways/hps.rb
@@ -43,7 +43,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def capture(money, transaction_id, options={})
-        commit('CreditAddToBatch') do |xml|
+        commit('CreditAddToBatch', transaction_id) do |xml|
           add_amount(xml, money)
           add_reference(xml, transaction_id)
         end
@@ -126,7 +126,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_reference(xml, transaction_id)
-        xml.hps :GatewayTxnId, transaction_id
+        reference = transaction_id.to_s.include?('|') ? transaction_id.split('|').first : transaction_id
+        xml.hps :GatewayTxnId, reference
       end
 
       def add_amount(xml, money)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -126,7 +126,8 @@ def commit_credit_sale(money, card_or_token, options)
       end
 
       def add_reference(xml, transaction_id)
-        xml.hps :GatewayTxnId, transaction_id
+        reference = transaction_id.to_s.include?(';') ? transaction_id.split(';').first : transaction_id
```
leila-alderman, line 129: A small thing, but I noticed that in [the external docs that are currently in progress](https://paper.dropbox.com/doc/Spreedly-Gateway-Integration-Submission-Guide--A6XiaoDgktOANJkEt0FFIu0BAg-xKEbPETXVm7lslzOVgkYx#:uid=844585903520059154756786&h2=Code-Expectations), David stated a preference for using `|` as the separator.

britth: good call out, thank you!

britth:
@leila-alderman that's so weird - i realized I was missing the `hps_echeck` entry in the fixtures files and adding that resolves the two test errors you mention, but three 3DS test errors are still there for me: `test_three_d_secure_mastercard`, `test_three_d_secure_discover`, `test_three_d_secure_amex`. will have to dig into that at some point 🕵️‍♀️ 

thanks for the review!!

PR #3743, opened 2020-08-24
Quickbooks: Omit empty strings in address

leila-alderman:
This fixes a bug caused by submitting an empty string for the state
field (which is submitted as 'region' to Quickbooks). While `nil` values
for this field are fine because the field is then simply omitted from
the request sent to Quickbooks, sending an empty string results in the
error message `card.address.region is invalid`.

ECS-1355

Unit:
22 tests, 118 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
18 tests, 44 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

All unit tests:
4544 tests, 72251 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_quickbooks_test.rb
+++ b/test/remote/gateways/remote_quickbooks_test.rb
@@ -141,6 +141,25 @@ class RemoteTest < Test::Unit::TestCase
     assert_success response
   end
 
+  def test_successful_purchase_without_state_in_address
+    options = {
+      order_id: '1',
+      billing_address:
+        {
+          zip: 90210,
+          # Submitting a value of an empty string for the `state` field
+          # results in a `region is invalid` error message from Quickbooks.
+          # This test ensures that an empty string is not sent from AM.
+          state: '',
+          country: ''
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_quickbooks_test.rb
```diff
@@ -141,6 +141,24 @@ def test_successful_purchase_with_expired_token
...
   end
 
+  def test_successful_purchase_without_state_address
+    options = {
+      order_id: '1',
+      billing_address:
+        {
+          zip: 90210,
+          # the test will pass if these values are nil
+          # but they fail when passed as empty strings
```
britth, line 151: do we want to keep this comment in? I think now that this change fixes this problem it could be omitted (or changed to something like `...previously it would fail if passed as empty strings`

leila-alderman: My intent was to leave a comment in there to describe that it's the empty string version that fails, but you're right that I need to update the text. Thanks for catching that!

PR #3744, opened 2020-08-26
Orbital: Allow level 3 fields to be passed on capture

naashton:
Add level 3 fields to capture transactions. Revert a previous 'fix'
where PC3DtlLineTot was incorrectly assumed to be the correct tag
(should be `PC3Dtllinetot`).

CE-930

Unit: 95 tests, 561 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 37 tests, 192 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.5946% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,7 +3,6 @@
 == HEAD
 * Orbital: Add cardIndicators field [meagabeth] #3734
 * Openpay: Add Colombia to supported countries [molbrown] #3740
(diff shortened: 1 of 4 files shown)


naashton:
Yes, the failing tests are known. I was trying to resolve those in this PR but wasn't quite able to track them down.

PR #3746, opened 2020-08-27
Update cardIndicators field to fix bug

meagabeth:
Confirm that card_indicators is being passed to the gateway. We have 1 test failing with this message: @message="Request does not adhere to the DTD. Please correct and send again.", but I was able to confirm that card_indicators is being passed with options. See options hash below. 3DS tests on orbital are still failing, as they have been on master.

{:order_id=>"bc8b1fd1ca40f4d8ba6e072cfa11f38f", :address=>{:name=>"Jim Smith", :address1=>"456 My Street", :address2=>"Apt 1", :company=>"Widgets Inc", :city=>"Ottawa", :state=>"ON", :zip=>"K1C2N6", :country=>"CA", :phone=>"(555)555-5555", :fax=>"(555)555-6666"}, :merchant_id=>"merchant1234", :soft_descriptors=>{:merchant_name=>"Merch", :product_description=>"Description", :merchant_email=>"email@example"}, :card_indicators=>"y"}

Unit:
95 tests, 561 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
36 tests, 184 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
86.1111% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -199,7 +199,6 @@ module ActiveMerchant #:nodoc:
             add_customer_data(xml, creditcard, options)
             add_managed_billing(xml, options)
           end
-          add_card_indicators(xml, options) if @options[:card_indicators]
         end
         commit(order, :authorize, options[:trace_number])
       end
@@ -220,7 +219,6 @@ module ActiveMerchant #:nodoc:
             add_customer_data(xml, creditcard, options)
             add_managed_billing(xml, options)
           end
-          add_card_indicators(xml, options) if @options[:card_indicators]
         end
         commit(order, :purchase, options[:trace_number])
       end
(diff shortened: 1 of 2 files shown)


naashton:
I've been trying to figure this one out, because __I think__ the one failure that is `"Request does not adhere to the DTD. Please correct and send again."` means the XML is invalid.

One thing I noticed in the Orbital documentation is that the tag is actually properly capitalized as, `CardIndicators`, not `cardIndicators`. Unfortunately that didn't fix the test.

The other thing I found in the documentation is that the Orbital gateway validates other elements when `CardIndicators` are present. Here is the snippet from the docs:

```
Orbital Gateway performs validations when this value is present. Those validations include:
• The Bin supports Card Type Indicators
• The Message Type is supported 
 - Supported Message Types: A, AC
• The Method of Payment is supported 
 - Supported MOPs: Visa, Mastercard, Discover, Diners, JCB, International Maestro
```

I also tried to line up the XML against the XSD and it appears that everything is in the correct order and all values that need to be there are present, which leads me to believe we don't quite have the test data correct.

I'm leaning toward approving this, but I think it might be worth getting additional eyes on this and potentially even reaching out to Orbital (or the customer) about this particular element.

meagabeth:
Branch successfully merged into master; leaving record of these commits for historical record/reference.

PR #3747, opened 2020-08-27
CyberSource: Always send country code

leila-alderman:
Previously, when an empty string was submitted for the
`[:billing_address][:country]`, CyberSource would response with an error
message since this field is required.

When a payment method that contained only a phone number with no other
address information was used, this resulted in the payment failing. To
resolve this issue, the `setup_address_hash` method was updated to match
the intent of the already existing comment: to always supply the default
country to CyberSource regardless of whether a `nil` value or an empty
string was initially submitted as the value for `country`.

ECS-1363

Unit:
97 tests, 470 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
95 tests, 490 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
94.7368% passed

All unit tests:
4551 tests, 72299 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -426,6 +426,12 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
     ActiveMerchant::Billing::CyberSourceGateway.application_id = nil
   end
 
+  def test_successful_purchase_with_country_submitted_as_empty_string
+    @options[:billing_address] = { country: '' }
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_successful_response(response)
+  end
+
   def test_unsuccessful_purchase
     assert response = @gateway.purchase(@amount, @declined_card, @options)
     assert_equal 'Invalid account number', response.message
@@ -918,22 +924,6 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
     }
     assert response = @gateway.purchase(@amount, @credit_card, @options)
     assert_successful_response(response)
-  end
-
-  def test_missing_field
-    @options = @options.merge({
-      address: {
-        address1: 'Unspecified',
-        city: 'Unspecified',
-        state: 'NC',
-        zip: '00000',
-        country: ''
-      }
-    })
-
-    assert response = @gateway.purchase(@amount, @credit_card, @options)
-    assert_failure response
-    assert_equal 'Request is missing one or more required fields: c:billTo/c:country', response.message
   end
 
   def test_invalid_field
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@
 * Orbital: Fix typo in PC3DtlLineTot field [naashton] #3736
 * Credorax: Send first and last name parameters for CFT transactions [britth] #3748
 * Orbital: Update CardIndicators field to fix bug [meagabeth] #3746
+* CyberSource: Always send country code [leila-alderman] #3747
 
 == Version 1.112.0
 * Cybersource: add `maestro` and `diners_club` eci brand mapping [bbraschi] #3708
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -266,7 +266,9 @@ module ActiveMerchant #:nodoc:
           zip: '00000',
           country: 'US'
         }
-        options[:billing_address] = options[:billing_address] || options[:address] || default_address
+
+        submitted_address = options[:billing_address] || options[:address] || default_address
+        options[:billing_address] = default_address.merge(submitted_address) { |_k, default, submitted| submitted.blank? ? default : submitted }
         options[:shipping_address] = options[:shipping_address] || {}
       end
 
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -1097,6 +1097,14 @@ class CyberSourceTest < Test::Unit::TestCase
     end.respond_with(successful_capture_response)
   end
 
+  def test_country_code_sent_as_default_when_submitted_as_empty_string
+    stub_comms do
+      @gateway.authorize(100, @credit_card, billing_address: { country: '' })
+    end.check_request do |endpoint, data, headers|
+      assert_match('<country>US</country>', data)
+    end.respond_with(successful_capture_response)
+  end
+
   def test_adds_application_id_as_partner_solution_id
     partner_id = 'partner_id'
     CyberSourceGateway.application_id = partner_id


test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -920,22 +926,6 @@ def test_successful_subsequent_unscheduled_cof_purchase
     assert_successful_response(response)
   end
 
-  def test_missing_field
```
leila-alderman, line 923: [This test was added recently](https://github.com/activemerchant/active_merchant/commit/3803514ec9275844d4ff713d815cd8413ebaefc9) with the intent of testing whether we are correctly passing the error message back from CyberSource. 

I don't believe that the intent was ever to ensure that we always pass through an empty string for the `country`, so I've deleted this test to match the new functionality. The test below still ensures that we are correctly passing back the error message from CyberSource.

leila-alderman:
@Carrigan Thanks for looking at this!

My intent with this change was to implement what seemed to be original intent based on the comment 
```
  # Create all address hash key value pairs so that we still function if we
  # were only provided with one or two of them or even none
```
combined with the fact that the `default_address` included all of the parameters that it did. 

My interpretation of that was that these fields are all required, based on [the original commit](https://github.com/activemerchant/active_merchant/commit/2fd2afa8e59a26b1e918bdb43dbcc7cebb7bc0d0) and [the original issue](https://github.com/spreedly/core/issues/1866). This even came up in [a later issue where David noted](https://github.com/spreedly/core/issues/1989#issuecomment-234383862):
> It looks like, as implemented, it doesn't actually do this - instead it's an all-or-nothing thing. Am I reading the code correctly?

and the response was
> yup, all or nothing. needs change to handle when parts of billing_address are not passed in ( like country in this specific error case )

Now, it is of course possible that all of the fields added to the `default_address` are not actually required. However, I did some very basic remote testing by removing all of the fields contained in the `default_address` hash and then adding them back in one by one until the tests passed again, and I got the following errors:
```
 - "Request is missing one or more required fields: c:billTo/c:state"
 - "Request is missing one or more required fields: c:billTo/c:postalCode"
 - "Request is missing one or more required fields: c:billTo/c:city"
 - "Request is missing one or more required fields: c:billTo/c:street1"
```
Based on these results, it seems that CyberSource only complains about one field at a time until all of the fields contained in the `default_address` hash are present, thus indicating that all of these fields are indeed required.

> Alternatively, we could require that the users of ActiveMerchant are passing in whatever default value they find appropriate.

I suppose it's also possible that we could just write back to the customer and tell them that CyberSource requires that they send in the `[:billing_address][:country]` for every transaction and not make any code changes to accommodate them. Looking through when this was originally implemented, though, I noticed [this statement from Duff](https://github.com/spreedly/core/issues/1866#issuecomment-225722998):
> This is an area we help customers when they're dealing with gateways that have requirements that most gateways don't have. It allows them to not have to specify things like billing address when they don't want to. This can be especially helpful if they're interacting with lots of gateways and they don't want to bother with their customers having to enter lots of details.

Therefore, providing these default address values seems like something that we want to do to help our customers run transactions more smoothly across gateways rather than pushing back on the customer to gather this data or provide default values for only certain gateways that require it.

pi3r:
FYI we just tried to deploy this at Shopify and it caused some issues with AVS. With those changes a perfectly fine address would end up being filled with `Unspecified` values. We're still investigating what the nature of the issue is.

Carrigan:
Thanks for letting us know @pi3r - let us know what you find and we'd be happy to help.

pi3r:
What we know is something like this 
```
"billing_address"=>{"name"=>"Pierre Oui", "address1"=>"12 moon lane", "address2"=>"", "phone"=>"", "city"=>"East Brunswick", "zip"=>"08816", "state"=>"NJ", "country"=>"US"}
```
Ended up being sent to Cybersource like this:
```
      <billTo>
        <firstName>Pierre</firstName>
        <lastName>Oui</lastName>
        <street1>Unspecified</street1>
        <city>Unspecified</city>
        <state>NC</state>
        <postalCode>00000</postalCode>
        <country>US</country>
        <email>[REDACTED]</email>
        <ipAddress>[REDACTED]</ipAddress>
      </billTo>
```

Carrigan:
Could it be an issue with String/Symbol conversion? The code that expects those values is looking for symbols. As an example:

```
xml.tag! 'street1, address[:address1]
```

The new merge method here would merge all of the defaults into the hash that you provided as symbol-keys and then the actual XML building code would read from the symbol-key over the string-key.

pi3r:
Yes that's it! I was just coming to the same conclusion :)

pi3r:
What do you think? Calling symbolize_keys on the submitted address should do it?

Carrigan:
Yeah - if you pass in `options[:billing_address] = { name: "Pierre Oui", address1: "12 moon lane", address2: "", phone: "", city: "East Brunswick", zip: "08816",  state: "NJ", country: "US" }` I think it will behave in the expected way :+1:

pi3r:
I more worried about other AM user. I'll make a pr to explicitly symbolize the keys as we can't ask every AM users to change how they pass params.

Carrigan:
Was this integration rendering the right address for a string-keyed input hash before this commit? It looks like this commit only changed the defaults for fields but the `add_address` function has always worked off of symbols.

pi3r:
Yeah, it would point to the fact that is wasn't working properly before. Let me double check something.

pi3r:
Ok, so we figured it out. This is due to the accessor precedence in `ActionController::Parameters`. Example:
```ruby
default_address = {
  address1: 'Unspecified',
  city: 'Unspecified',
  state: 'NC',
  zip: '00000',
  country: 'US'
}
address = ActionController::Parameters.new({ name: "Pierre Oui", address1: "12 moon lane", address2: "", phone: "", city: "East Brunswick", zip: "08816", state: "NJ", country: "US" })
address[:address1] == "12 moon lane" #=> true
address["address1"] == "12 moon lane" #=> true

default_address.merge(submitted_address) { |_k, default, submitted| submitted.blank? ? default : submitted }
default_address[:address1] == "12 moon lane" #=> false
default_address["address1"] == "12 moon lane" #=> true

It was fine before because without the merge there was only one key in the `ActionController::Parameters` instance, now with the merge there are both stored and the default address one takes precedence. Am I making sense?
```

Carrigan:
Yes this makes sense. I can see three paths forward here:

1. Back out this commit and have a different way of populating the defaults
1. Add a new PR in ActiveMerchant to keep these defaults but force the incoming hashes to symbol-keys
1. Have Shopify symbolize its keys before passing them to AM

Given that others may be passing in `ActionController::Parameters` but this PR provides valuable defaults, I think the second option makes the most to me. Which would you prefer to pursue?

pi3r:
I agree, I'm putting pr together for 2. Thank you for you help Brian!

Carrigan:
No problem @pi3r - feel free to tag spreedly-integrating as a reviewer when you are finished and thanks for bringing this issue up!

PR #3748, opened 2020-08-27
Credorax: Send j5 and j13 params for CFT txns

britth:
Credorax now requires the following parameters when issuing CFT payouts:
* j5 - Fund recipient's first name
* j13 - Fund recipient's surname

Updates the ~credit and~ referral_cft action to set these two fields,
and checks for their presence in the tests. _(Updated: credorax confirmed that 
they're not required for general credit [6], only actions [34], [35], [37], and [38] - 
of these, currently AM only supports [[34]](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/credorax.rb#L409))_

Remote:
41 tests, 152 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
69 tests, 331 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
 * Quickbooks: Omit empty strings in address [leila-alderman] #3743
 * BlueSnap: Add transactionMetaData support #3745
 * Orbital: Fix typo in PC3DtlLineTot field [naashton] #3736
+* Credorax: Send first and last name parameters for CFT transactions [britth] #3748
 
 == Version 1.112.0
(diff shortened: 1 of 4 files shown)


britth:
several remote tests are currently failing due to what looks like the 3DS server being down - will try running again later, but failures should be unrelated (and tests all passed yesterday)

britth:
thanks @Carrigan !

PR #3750, opened 2020-09-02
iATS Payments: Pass email address outside of the billing address scope

naashton:
Give ability to pass customer details outside of the scope of the
billing address

CE-795

Unit: 18 tests, 178 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 14 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_iats_payments_test.rb
+++ b/test/remote/gateways/remote_iats_payments_test.rb
@@ -28,6 +28,14 @@ class IatsPaymentsTest < Test::Unit::TestCase
     assert response.authorization
   end
 
+  def test_successful_purchase_with_customer_details
+    assert response = @gateway.purchase(@amount, @credit_card, @options.merge(@customer_details))
+    assert_success response
+    assert response.test?
+    assert_equal 'Success', response.message
+    assert response.authorization
+  end
+
   def test_failed_purchase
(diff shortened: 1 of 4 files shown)


meagabeth (commented):
Tests are passing! Does a unit test need to be written for this?

naashton:
In this particular case, this is a specific user case where the customer wants to be able to pass customer data at the transaction level as opposed to passing it in the billing array. That being said, adding a unit test won't hurt.

test/remote/gateways/remote_iats_payments_test.rb
```diff
@@ -66,7 +74,7 @@ def test_failed_refund
     purchase = @gateway.purchase(@amount, credit_card, @options)
     assert_success purchase
 
-    assert refund = @gateway.refund(@amount, purchase.authorization)
```
therufs, line 69: weird

therufs (approved):
lgtm; all local and remote tests pass!

test/remote/gateways/remote_iats_payments_test.rb
naashton: Super weird

PR #3752, opened 2020-09-03
BlueSnap: Protect against nil metadata

Carrigan:
## Why?

A recent commit added support for receiving meta-data on BlueSnap but did not protect against a `nil` key being submitted for that field. Line 218 with `{ transaction_meta_data: nil }` would raise a NoMethodError.

## What Changed?

This change protects against `{ transaction_meta_data: nil }` for BlueSnap.

## Test Suite

4556 tests, 72358 assertions, 0 failures, 0 errors, 0 pendings,
  0 omissions, 0 notifications
692 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -214,7 +214,7 @@ module ActiveMerchant
       end
 
       def add_metadata(doc, options)
-        transaction_meta_data = options.fetch(:transaction_meta_data, {})
+        transaction_meta_data = options[:transaction_meta_data] || []
         return if transaction_meta_data.empty? && !options[:description]
 
         doc.send('transaction-meta-data') do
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -214,7 +214,7 @@ def add_credit_card(doc, card)
       end
 
       def add_metadata(doc, options)
-        transaction_meta_data = options.fetch(:transaction_meta_data, {})
+        transaction_meta_data = options[:transaction_meta_data] || []
```
fatcatt316, line 217: Nice catch n' fix! Did you check that your new test fails as expected if this code change here isn't in place?

fatcatt316 (approved):
Those pesky `nil`s ✊ 

Diagnosis and fix look good to me!

lib/active_merchant/billing/gateways/blue_snap.rb
Carrigan: Yes - apologies for not putting that in the description.

```
Error: test_successful_purchase_with_metadata_nil(BlueSnapTest): NoMethodError: undefined method `empty?' for nil:NilClass
```

naashton (approved):
Tests pass. Thanks Brian!

Carrigan:
Thanks for the quick review @naashton @fatcatt316 :)

PR #3755, opened 2020-09-03
[CyberSource] Ensure the default address doesn't override `ActionController::Parameters`

pi3r:
When `ActionController::Parameters` is passed to Cybersource, the default address fields are used
instead of the passed parameters. See following comment for more information.

References: https://github.com/activemerchant/active_merchant/pull/3747#issuecomment-686720226

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -268,7 +268,7 @@ module ActiveMerchant #:nodoc:
         }
 
         submitted_address = options[:billing_address] || options[:address] || default_address
-        options[:billing_address] = default_address.merge(submitted_address) { |_k, default, submitted| submitted.blank? ? default : submitted }
+        options[:billing_address] = default_address.merge(submitted_address.symbolize_keys) { |_k, default, submitted| submitted.blank? ? default : submitted }
         options[:shipping_address] = options[:shipping_address] || {}
       end
 
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -1105,6 +1105,22 @@ class CyberSourceTest < Test::Unit::TestCase
     end.respond_with(successful_capture_response)
   end
 
+  def test_default_address_does_not_override_when_hash_keys_are_strings
+    stub_comms do
+      @gateway.authorize(100, @credit_card, billing_address: {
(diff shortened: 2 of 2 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -268,7 +268,7 @@ def setup_address_hash(options)
         }
 
         submitted_address = options[:billing_address] || options[:address] || default_address
-        options[:billing_address] = default_address.merge(submitted_address) { |_k, default, submitted| submitted.blank? ? default : submitted }
+        options[:billing_address] = default_address.merge(submitted_address.symbolize_keys) { |_k, default, submitted| submitted.blank? ? default : submitted }
```
bbraschi, line 271: An alternative would be to create default_address with string keys from the get go.
We could also use `merge!` with the bang, since we instantiate `default_address` on the fly.

bbraschi (approved):

lib/active_merchant/billing/gateways/cyber_source.rb
pi3r: > An alternative would be to create default_address with string keys from the get go.

That wouldn't work since the symbol accessor is used here https://github.com/activemerchant/active_merchant/blob/e049610610d3e20aa0d76188db200976aea97b08/lib/active_merchant/billing/gateways/cyber_source.rb#L526

> We could also use merge! with the bang, since we instantiate default_address on the fly.

Didn't you told that `merge` was faster?

bbraschi: I thought merge! was faster, quickly googled [here](https://www.paperlesspost.com/blog/teams/different-methods-for-merging-ruby-hashes/)
> Using merge! is almost 2 times as fast

Carrigan: Is `symbolize_keys` OK to use here? I believe that is a Rails function and not a Ruby function, and this repo does not have Rails as a dependency.

https://apidock.com/rails/Hash/symbolize_keys

Carrigan: Ahh thanks - the tests all passed for me but I didn't see how it was being included since ActiveSupport isn't in the Gemfile :+1:

Carrigan (approved):
Looks good! Thanks for making this more robust.

PR #3756, opened 2020-09-04
Adyen: Safely add execute_threeds: false

curiousepic:
In df60a2c5512bf43917a8d5d825d6b7c021dbdd1d when execute_threeds is
false it is assigned to the additionalData as a whole instead of
additively, so previously added additionalData fields were destroyed.
This change adds the field safely.

Remote:
92 tests, 353 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
70 tests, 330 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -432,7 +432,8 @@ module ActiveMerchant #:nodoc:
           return unless !options[:execute_threed].nil? || !options[:threed_dynamic].nil?
 
           post[:browserInfo] = { userAgent: options[:user_agent], acceptHeader: options[:accept_header] } if options[:execute_threed] || options[:threed_dynamic]
-          post[:additionalData] = { executeThreeD: options[:execute_threed] } if !options[:execute_threed].nil?
+          post[:additionalData] ||= {}
+          post[:additionalData][:executeThreeD] = options[:execute_threed] if !options[:execute_threed].nil?
         end
       end
 
--- a/test/unit/gateways/adyen_test.rb
+++ b/test/unit/gateways/adyen_test.rb
@@ -376,6 +376,15 @@ class AdyenTest < Test::Unit::TestCase
     end.respond_with(successful_authorize_response)
   end
 
+  def test_execute_threed_false_with_additional_data
+    stub_comms do
+      @gateway.authorize(@amount, @credit_card, @options.merge({execute_threed: false, overwrite_brand: true, selected_brand: 'maestro'}))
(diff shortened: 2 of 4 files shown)


molbrown (approved):
Nice catch and clear solution. I get passing remotes and units as well.

test/unit/gateways/adyen_test.rb
```diff
@@ -376,6 +376,16 @@ def test_splits_sent
     end.respond_with(successful_authorize_response)
   end
 
+  def test_execute_threed_false_with_additional_data
```
Carrigan, line 379: If I revert the changes in the Adyen gateway here, this test still passes. Is there any way to catch a failure in the old code?

curiousepic: Good catch, I'll take a look

curiousepic: Strange, I'm getting inconsistent behavior between virtually identical unit and remote tests. Even as-is here, the unit test doesn't add the executeThreeD flag - (`assert_false` is a bad test).

curiousepic: Nevermind, it was just a typo on my part `execute_threeds` instead of `execute_threed` 😩

curiousepic: Fixed now. If you could verify I'd appreciate it!

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -432,7 +432,8 @@ def add_3ds(post, options)
           return unless !options[:execute_threed].nil? || !options[:threed_dynamic].nil?
 
           post[:browserInfo] = { userAgent: options[:user_agent], acceptHeader: options[:accept_header] } if options[:execute_threed] || options[:threed_dynamic]
-          post[:additionalData] = { executeThreeD: options[:execute_threed] } if !options[:execute_threed].nil?
+          post[:additionalData] ||= {}
```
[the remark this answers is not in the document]
britth: 💥

britth (approved):
looks great, and test update passes with your changes, and fails without them now for me 👍

test/unit/gateways/adyen_test.rb
Carrigan: Worked for me - thanks for fixing this 🎉

Carrigan (approved):
Looks great to me :+1:

PR #3758, opened 2020-09-09
CyberSource: Only include non-nil `mdd_` fields

arbianchi:
When all `mdd_field` options are `nil`, the merchantDefinedData gets
added as an empty XML tag which causes the gateway to return a parse
error.


shasum (approved):
Looks straightforward 👍

PR #3760, opened 2020-09-10
Orbital: Fix line_tot conditional check

naashton:
`if :line_tot` is not resolving to true in some cases because the value
is actually a string. Expanding the check to evaluate against a string
and symlink should resolve this issue.

CE-762

Unit: 95 tests, 560 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 37 tests, 192 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.5946% passed


meagabeth (approved):
Looks good! Unit tests passing, Remote tests have 2 unrelated failures: 
- test_successful_american_express_authorization_with_3ds
- test_successful_american_express_purchase_with_3ds

PR #3761, opened 2020-09-11
Checkout V2: Move to single-transaction Purchases

curiousepic:
When the Checkout V2 adapter was initially implemented, it seems that
the Checkout API couldn't perform a refund on an auto-captured
authorizations (ie, single-transaction purchase), so Purchase used a
multi-response authorization and capture call.

We have recently noticed intermittent Purchases that the adapter
considers successful but in fact the Capture response was a 404.
According to Checkout, this may be due to their id for the authorization
not having yet been instantiated before the Capture reached them. So the
multiresponse is too fast for its own good, and also the adapter's
success criteria isn't sufficient to catch a failed Capture as part of
a Purchase.

However, it seems that the constraint on refunds has since been removed.
So, we can switch to a single-transaction Purchase, by not setting the
capture field as false. There should be no effective difference in terms
of the resulting Response.

Remote:
33 tests, 80 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
29 tests, 129 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_checkout_v2_test.rb
+++ b/test/remote/gateways/remote_checkout_v2_test.rb
@@ -149,9 +149,9 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
   end
 
   def test_failed_purchase
-    response = @gateway.purchase(@amount, @declined_card, @options)
+    response = @gateway.purchase(12305, @credit_card, @options)
     assert_failure response
-    assert_equal 'request_invalid: card_number_invalid', response.message
+    assert_equal 'Declined - Do Not Honour', response.message
   end
 
   def test_avs_failed_purchase
@@ -207,8 +207,9 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
   end
 
   def test_failed_authorize
-    response = @gateway.authorize(@amount, @declined_card, @options)
+    response = @gateway.authorize(12314, @credit_card, @options)
     assert_failure response
+    assert_equal 'Invalid Card Number', response.message
   end
 
   def test_partial_capture


britth (commented):
this change looks great and makes sense to me 👍 
Not sure if this was just a transient thing, but the first time I tried to run remote tests, the `test_successful_refund` test actually failed at the refund stage, though it was a 403 forbidden error - however, I ran twice afterward and it passed both times. I feel like that's unrelated but just noting in case that deserves investigation. Otherwise, LGTM!

molbrown (commented):
Have you determined yet if merchants also need to enable auto-capture in their Checkout admin, or this will work as the default? In [the docs](https://api-reference.checkout.com/#tag/Payments), I see the `capture` field described as-
> Whether to capture the payment (if applicable)

And I wonder what (if applicable) refers to, since the default value is true.

I don't think this is test-able, since captures are completed asynchronously, but it would be good to have confirmation on this from Checkout, as this could be a concern for current integrations to this adapter.

curiousepic:
@molbrown Thanks for the scrutinty - I can't find any reference in Docs that this is would be a merchant-level toggle, nor in their sandbox UI, and Nicolas' communications in our internal ticket doesn't indicate this either. I expect that "if applicable" refers to whether it's a payout instead of a payment.

nicolas-maalouf-cko:
Hi all,

Thanks for your help with this one.

"If applicable" is there because some alternative payment methods as well as payout to cards don't involve captures.

When Capture is set to true (or left to its default value) in a card payment request, we automatically queue a capture after the authorisation is successful. Note that the synchronous response is still the one related to the authorisation and not the capture.

The 403s obtained in some tests are most likely because the automatic capture hadn't been executed yet, which resulted in the refund's failure.

Would this change affect both the Shopify and Spreedly integrations?

Kind regards

curiousepic:
@nicolas-maalouf-cko Thanks for the info! It will affect Shopify, but as a reminder, only if/when they accept it internally, not on merging into this repo.
Tagging @activemerchant/shopify-payments-devs as an FYI

nicolas-maalouf-cko:
One last think I'd recommend is testing declines and other scenarios on a sandbox test store:
https://docs.checkout.com/testing/response-code-testing

Would it be possible?

curiousepic:
It would be good to move to testing failed cases via those amount codes. I assume these were not available when the adapter was first built, so we are primarily testing failed cases via a bad card number, but this actually invokes fail-by-error rather than proper decline/invalid responses. Thanks for pointing that out. I will create a separate PR to add a few of those just so we have at least some coverage in that respect.

In any case, some manual spot checks invoking responses via amount codes do pass well with this change 👍

curiousepic:

PR #3766, opened 2020-09-15
Orbital: Fix schema errors

britth:
After fixing an issue with problematic values being sent for network
tokenized PMs (#3757), errors relating to the request not adhering to the
DTD would sometimes appear. This happens in the existing code when
soft descriptors are passed along with a network tokenized card.
In the schema, the soft descriptor fields need to come before the
network token fields. This PR updates the order of these elements
and adds additional test conditions.

Unit:
96 tests, 563 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
37 tests, 192 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.5946% passed
(same as master)

### DIFF
--- a/test/unit/gateways/orbital_test.rb
+++ b/test/unit/gateways/orbital_test.rb
@@ -174,6 +174,22 @@ class OrbitalGatewayTest < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
+  def test_schema_for_soft_descriptors_with_network_tokenization_credit_card_data
+    options = @options.merge(
+      soft_descriptors: {
+        merchant_name: 'Merch',
+        product_description: 'Description',
+        merchant_email: 'email@example'
+      }
+    )
+    stub_comms do
+      @gateway.purchase(50, network_tokenization_credit_card(nil, eci: '5', transaction_id: 'BwABB4JRdgAAAAAAiFF2AAAAAAA='), options)
+    end.check_request do |_endpoint, data, _headers|
+      # Soft descriptor fields should come before dpan and cryptogram fields
+      assert_match %{<SDMerchantEmail>email@example<\/SDMerchantEmail><DPANInd>Y<\/DPANInd><DigitalTokenCryptogram}, data.gsub(/\s+/, '')
+    end.respond_with(successful_purchase_response)
+  end
+
   def test_three_d_secure_data_on_visa_purchase
     stub_comms do
       @gateway.purchase(50, credit_card, @options.merge(@three_d_secure_options))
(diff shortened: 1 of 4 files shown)


test/unit/gateways/orbital_test.rb
```diff
@@ -174,6 +174,22 @@ def test_network_tokenization_credit_card_data
...
+    options = @options.merge(
+      soft_descriptors: {
+        merchant_name: 'Merch',
+        product_description: 'Description',
+        merchant_email: 'email@example'
+      }
+    )
+    stub_comms do
+      @gateway.purchase(50, network_tokenization_credit_card(nil, eci: '5', transaction_id: 'BwABB4JRdgAAAAAAiFF2AAAAAAA='), options)
+    end.check_request do |endpoint, data, headers|
```
leila-alderman, line 187: With some of the recent RuboCop clean up that I've done, this should be updated to

```suggestion
    end.check_request do |_endpoint, data, _headers|
```

test/unit/gateways/orbital_test.rb
```diff
@@ -174,6 +174,22 @@ def test_network_tokenization_credit_card_data
...
+      # Soft descriptor fields should come before dpan and cryptogram fields
+      assert_match %{<SDMerchantEmail>email@example</SDMerchantEmail> <DPANInd>Y</DPANInd> <DigitalTokenCryptogram}, data.squish
```
leila-alderman, line 189: The Travis CI does not like the use of `squish` here:
image

I'm not entirely sure if the CI will find this acceptable, but I got the test to pass using this format:
```suggestion
      data =~ /<SDMerchantEmail>email@example<\/SDMerchantEmail> <DPANInd>Y<\/DPANInd> <DigitalTokenCryptogram/
```

leila-alderman (changes requested):
Thanks for catching this error and fixing it! Just need to make the CI happy, and then this should be good to go.

test/unit/gateways/orbital_test.rb
britth: tsk tsk tsk, I think I even reviewed that one! 🤦‍♀️  Thanks for the catch!

test/unit/gateways/orbital_test.rb
britth: updated! I ended up subbing all whitespace and then doing an assert_match (I was getting `nil` as the return value for `data =~ /<SDMerchantEmail>email@example<\/SDMerchantEmail> <DPANInd>Y<\/DPANInd> <DigitalTokenCryptogram/` rather than true, I think because there's some weird whitespace and line breaks in that data string)

leila-alderman (approved):
Perfection! The CI is happy, and all of the tests pass for me locally. 🚀

britth:
thanks @leila-alderman !

PR #3767, opened 2020-09-15
Checkout V2: Start testing via amount code

curiousepic:
The Checkout sandbox API supports invocation of response types via
specific amounts:
https://docs.checkout.com/testing/response-code-testing
We're currently testing failed cases via a bad card number, but this
actually invokes fails-by-error rather than proper decline/invalid
number responses. This changes the primary authorize and purchase remote
tests to utilize this method, but is only a start.

Remote:
33 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


leila-alderman (approved):
Awesome update! Thanks for getting progress on this started.

The unit and remote tests all pass for me. 🚀

PR #3774, opened 2020-09-28
PayU Latam: Support partial refunds

leila-alderman:
Added support for partial refunds indicated by passing
`partial_refund: true`
in the options hash.

Previously, all refunds were treated as full refunds, regardless of the
amount sent to the gateway.

ECS-1401

Unit:
34 tests, 137 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
100% passed

Remote:
35 tests, 79 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions,
0 notifications
91.4286% passed

All unit tests:
4563 tests, 72389 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_payu_latam_test.rb
+++ b/test/remote/gateways/remote_payu_latam_test.rb
@@ -232,7 +232,7 @@ class RemotePayuLatamTest < Test::Unit::TestCase
   end
 
   def test_failed_purchase
-    response = @gateway.purchase(@amount, @declined_card, @options)
+    response = @gateway.purchase(@amount, @declined_card)
     assert_failure response
     assert_equal 'DECLINED', response.params['transactionResponse']['state']
   end
@@ -298,7 +298,7 @@ class RemotePayuLatamTest < Test::Unit::TestCase
   end
 
   def test_failed_authorize
-    response = @gateway.authorize(@amount, @pending_card, @options)
+    response = @gateway.authorize(@amount, @declined_card)
     assert_failure response
     assert_equal 'DECLINED', response.params['transactionResponse']['state']
   end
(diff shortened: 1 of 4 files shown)


curiousepic:
I'm wondering if we could simply assume partial refunds by default, and only send a normal/full refund if there is no amount provided, rather than tying it to an option? Do you think that would be unsafe?

leila-alderman:
@curiousepic What you're suggesting was exactly my first approach to solving this. However, it doesn't work because within Core, we _always_ set the `amount` field and sent it into Active Merchant. If the customer doesn't send an amount in the request to Spreedly, then we populate the field with the amount of the original (reference) transaction. This also matches [the method signature that we expect to see in Active Merchant](https://github.com/activemerchant/active_merchant/wiki/contributing#generating-the-gateway-skeleton) for `refund` methods.

leila-alderman:
The 3 failing remote tests:
 - `test_successful_purchase_brazil`
 - `test_successful_purchase_mexico`
 - `test_successful_purchase_with_buyer`

These are all failing with the error message `"INVALID_TRANSACTION | Affiliation: Invalid parameter format."`

Despite poking around the code and the gateway's documentation, I have failed to determine what even `Affiliation` is. I don't believe that any of these failures are related to the changes made here.

test/remote/gateways/remote_payu_latam_test.rb
```diff
@@ -232,7 +232,7 @@ def test_successful_purchase_no_description
   end
 
   def test_failed_purchase
-    response = @gateway.purchase(@amount, @declined_card, @options)
+    response = @gateway.purchase(@amount, @declined_card)
```
leila-alderman, line 235: I removed `@options` here and below to prevent the `billing_address` field used with `@options` from overwriting the "name" field, which is what triggers a success or failure response.

curiousepic (approved):
Looks good, considering! 👍

PR #3781, opened 2020-10-05
RuboCop: Fix Naming/HeredocDelimiterCase & Naming

leila-alderman:
Fixes the RuboCop to-dos for both [Naming/HeredocDelimiterCase](https://docs.rubocop.org/rubocop/0.85/cops_naming.html#namingheredocdelimitercase)
and [Naming/HeredocDelimiterNaming](https://docs.rubocop.org/rubocop/0.85/cops_naming.html#namingheredocdelimiternaming).

All unit tests:
4565 tests, 72404 assertions, 0 failures, 0 errors, 0 pendings, 0
omissions, 0 notifications
100% passed


schwarzgeist:
These look like great suggested changes! Out of curiosity, where did your naming convention come from?

leila-alderman:
@schwarzgeist I searched through the existing Active Merchant code and saw that other files typically used either "RESPONSE" or "XML". [I found an article about what heredocs are intended to achieve in Ruby](https://makandracards.com/makandra/1675-using-heredoc-for-prettier-ruby-code), which convinced me that "XML" was the way to go when appropriate. There were some cases where the text was an HTTP request, not XML, so in those cases, I went with "REQUEST".

schwarzgeist:
That sounds very reasonable. Thanks for clarifying!

schwarzgeist (approved):
With [this explanation](https://github.com/activemerchant/active_merchant/pull/3781#issuecomment-705756887), I think these changes make sense and should clarify things.

Rubocop is also happy for me, no complaints:

> 4565 tests, 72404 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
> 100% passed

PR #3785, opened 2020-10-06
Redsys 3DS2 Integration Support

esmitperez:
Support 3DS2 Fields

ECS-1427

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys.rb
+++ b/lib/active_merchant/billing/gateways/redsys.rb
@@ -197,6 +197,7 @@ module ActiveMerchant #:nodoc:
         add_amount(data, money, options)
         add_order(data, options[:order_id])
         add_payment(data, payment)
+        add_external_mpi_fields(data, options)
         add_threeds(data, options) if options[:execute_threed]
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
@@ -213,6 +214,7 @@ module ActiveMerchant #:nodoc:
         add_amount(data, money, options)
         add_order(data, options[:order_id])
         add_payment(data, payment)
+        add_external_mpi_fields(data, options)
         add_threeds(data, options) if options[:execute_threed]
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -460,6 +480,28 @@ def build_merchant_data(xml, data, options = {})
         end
       end
 
+      def build_merchant_mpi_external(xml, data)
+        return unless data[:txid]
```
leila-alderman, line 484: The `txid` is only a required field for 3DS1, not 3DS2. Maybe something like:

```suggestion
        return unless data[:txid] || data[:threeDSServerTransID]
```

test/remote/gateways/remote_redsys_test.rb
```diff
@@ -18,6 +18,17 @@ def test_successful_purchase
     assert_equal 'Transaction Approved', response.message
   end
 
+  def test_successful_purchase_threeds2
+    response = @gateway.purchase(100, @credit_card, @options.merge(
+      three_d_secure: {
+        xid: '33ds22'
```
leila-alderman, line 24: Could you try exercising more 3DS fields here, [similar to the Adyen remote tests](https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_adyen_test.rb#L336-L343)?

We may also want to consider breaking this out into two separate remote tests:

- 3DS1
  - xid
  - cavv
  - eci
- 3DS2
  - threeDSServerTransID
  - dsTransID
  - authenticacionValue
  - protocolVersion
  - authenticacionMethod
  - authenticacionType
  - authenticacionFlow

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -321,6 +322,22 @@ def add_payment(data, card)
...
+        data[:cavv] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
+        data[:eci] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
+
+        data[:threeDSServerTransID] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid] 
+        data[:dsTransID] = options[:three_d_secure][:ds_transaction_id] if options[:three_d_secure][:ds_transaction_id]
+        data[:protocolVersion] = options[:three_d_secure][:version] if options[:three_d_secure][:version]
+
+        data[:authenticacionMethod] = options[:three_d_secure][:eci] if options[:three_d_secure][:authentication_method]
+        data[:authenticacionType] = options[:three_d_secure][:eci] if options[:three_d_secure][:authentication_type]
+        data[:authenticationFlow] = options[:three_d_secure][:eci] if options[:three_d_secure][:authentication_flow]
```
leila-alderman, line 338: Since these are optional fields for Redsys that do not map to any of our standardized 3DS fields, should we consider allowing the customer to pass these in as standard GSFs that are provided outside of the `options[:three_d_secure]` hash?

PR #3787, opened 2020-10-07
Stripe Payment Intents: Add request_three_d_secure option

molbrown:
Enable override of Stripe's SCA engine and directly request 3DS using
the request_three_d_secure option. When 3DS is not supported for the
card being charged, liability is still shifted, as indicated by the
'attempt_acknowledged' response in
`payment_method_details.three_d_secure.result`.

Details of field use: https://stripe.com/docs/api/payment_intents/create#create_payment_intent-payment_method_options-card-request_three_d_secure

3DS Result values: https://stripe.com/docs/api/charges/object#charge_object-payment_method_details-card-three_d_secure-result

Unit:
14 tests, 84 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
42 tests, 192 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-1455

### DIFF
--- a/test/remote/gateways/remote_stripe_payment_intents_test.rb
+++ b/test/remote/gateways/remote_stripe_payment_intents_test.rb
@@ -13,11 +13,15 @@ class RemoteStripeIntentsTest < Test::Unit::TestCase
     @three_ds_credit_card = credit_card('4000000000003220',
       verification_value: '737',
       month: 10,
-      year: 2020)
+      year: 2021)
+    @three_ds_not_required_card = credit_card('4000000000003055',
+      verification_value: '737',
+      month: 10,
+      year: 2021)
     @visa_card = credit_card('4242424242424242',
       verification_value: '737',
       month: 10,
-      year: 2020)
+      year: 2021)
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -13,11 +13,18 @@ def setup
     @three_ds_credit_card = credit_card('4000000000003220',
       verification_value: '737',
       month: 10,
-      year: 2020)
+      year: 2021
+    )
```
britth, line 17: nitpick - i'm getting a couple of rubocop errors around these closing parens (`Layout/MultilineMethodCallBraceLayout: Closing method call brace must be on the same line as the last argument when opening brace is on the same line as the first argument.`)

britth (approved):
this looks great to me! change makes sense and behavior as exhibited in tests matches what I'd expect based on the description and the docs :shipit:

PR #3788, opened 2020-10-14
Checkout V2: Support ability to pass attempt_n3d 3ds field

naashton:
Give the option to pass attempt_n3d which will attempt non-3d secure if
the card issure is not enrolled

CE-1080

100% passed

Remote: 33 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


meagabeth (approved):
Looks good and tests are passing!

PR #3790, opened 2020-10-16
Checkout V2: Support for attempt_n3d 3DS field

naashton:
This updates the support for the attempt_n3d field when it is passed
outside of the three_d_secure context

CE-1080

100% passed

Remote: 33 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


meagabeth (approved):
Looks good and tests are passing. Good call on noticing that these need to go outside of the standardized three_d_secure hash - this is something I just learned!

PR #3794, opened 2020-10-16
Redsys: Add 3DS2 Integration Support

esmitperez:
Support 3DS2 Fields, adding an embedded JSON document in node DS_MERCHANT_MPIEXTERNAL

ECS-1427

Unit:
40 tests, 140 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
23 tests, 72 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys.rb
+++ b/lib/active_merchant/billing/gateways/redsys.rb
@@ -197,6 +197,7 @@ module ActiveMerchant #:nodoc:
         add_amount(data, money, options)
         add_order(data, options[:order_id])
         add_payment(data, payment)
+        add_external_mpi_fields(data, options)
         add_threeds(data, options) if options[:execute_threed]
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
@@ -213,6 +214,7 @@ module ActiveMerchant #:nodoc:
         add_amount(data, money, options)
         add_order(data, options[:order_id])
         add_payment(data, payment)
+        add_external_mpi_fields(data, options)
         add_threeds(data, options) if options[:execute_threed]
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
@@ -321,6 +323,25 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def add_external_mpi_fields(data, options)
+        return unless options[:three_d_secure]
+
+        if options[:three_d_secure][:version]
+          data[:threeDSServerTransID] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
+          data[:dsTransID] = options[:three_d_secure][:ds_transaction_id] if options[:three_d_secure][:ds_transaction_id]
+          data[:authenticacionValue] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
+          data[:protocolVersion] = options[:three_d_secure][:version] if options[:three_d_secure][:version]
+
+          data[:authenticacionMethod] = options[:authentication_method] if options[:authentication_method]
+          data[:authenticacionType] = options[:authentication_type] if options[:authentication_type]
+          data[:authenticationFlow] = options[:authentication_flow] if options[:authentication_flow]
+        elsif options[:three_d_secure][:version]
+          data[:txid] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
+          data[:cavv] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
+          data[:eci] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
+        end
+      end
+
       def add_threeds(data, options)
         data[:threeds] = { threeDSInfo: 'CardData' } if options[:execute_threed] == true
       end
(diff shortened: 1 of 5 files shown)


test/remote/gateways/remote_redsys_sha256_test.rb
```diff
@@ -18,6 +18,27 @@ def test_successful_purchase
...
+          ds_transaction_id: ds_transaction_id,
+          xid: xid
+        }
+      )
+    )
+
+    assert_success response
+    assert_equal 'Transaction Approved', response.message
+  end
+
```
curiousepic, line 41: Given that we've had various issues with strict ordering of XML requests, I'd suggest testing this with the fields that could be added after the mpi element. Were you able to find a schema?

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -321,6 +323,26 @@ def add_payment(data, card)
...
+      def add_external_mpi_fields(data, options)
+        return unless options[:three_d_secure]
+
+        if options[:three_d_secure][:version]
+          data[:threeDSServerTransID] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
+          data[:dsTransID] = options[:three_d_secure][:ds_transaction_id] if options[:three_d_secure][:ds_transaction_id]
+          data[:authenticacionValue] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
+          data[:protocolVersion] = options[:three_d_secure][:version] if options[:three_d_secure][:version]
+
+          # TODO: keep here?
```
curiousepic, line 335: Were you considering pulling these out to another method? I think they're alright here.

test/remote/gateways/remote_redsys_sha256_test.rb
esmitperez: yes, by adding ?WSDL to the URL it shows the schema, they're all marked "literal", strict ordering/type check is not enforced at the structural level. I'll add some smoke tests for our peace of mind

lib/active_merchant/billing/gateways/redsys.rb
esmitperez: Leila suggested these could actually live as optional via our gateway integration in`core`. I welcome your thoughts around that too

test/remote/gateways/remote_redsys_sha256_test.rb
curiousepic: Nice!

lib/active_merchant/billing/gateways/redsys.rb
curiousepic: I see. Yes, I think these non-standard but 3ds-related fields should be general options rather than assume dto exist in the `three_d_secure` hash.

varyonic:
Which external service are you using?  We are working on this also.

esmitperez:
> Which external service are you using? We are working on this also. #3793

Our wrapper abstracts that away, but it's cool to see concurrent work on the same gateway :)

test/remote/gateways/remote_redsys_sha256_test.rb
```diff
@@ -18,6 +18,30 @@ def test_successful_purchase
...
+      @credit_card,
+      @options.merge(
+        three_d_secure: {
+          version: version,
+          ds_transaction_id: ds_transaction_id,
+          xid: xid
+        },
+        description: 'description',
+        store: 'store',
+        sca_exemption: 'MOTO'
```
varyonic, line 37: I'm curious why an exemption is being requested here.

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -321,6 +323,25 @@ def add_payment(data, card)
...
+
+        if options[:three_d_secure][:version]
+          data[:threeDSServerTransID] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
+          data[:dsTransID] = options[:three_d_secure][:ds_transaction_id] if options[:three_d_secure][:ds_transaction_id]
+          data[:authenticacionValue] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
+          data[:protocolVersion] = options[:three_d_secure][:version] if options[:three_d_secure][:version]
+
+          data[:authenticacionMethod] = options[:authentication_method] if options[:authentication_method]
+          data[:authenticacionType] = options[:authentication_type] if options[:authentication_type]
+          data[:authenticationFlow] = options[:authentication_flow] if options[:authentication_flow]
```
curiousepic, line 337: is this actually `authenticationFlow` and inconsistent?

curiousepic (approved):
Noticed a possible typo, otherwise LGTM!

lib/active_merchant/billing/gateways/redsys.rb
esmitperez: I deliberately held-off on using the misspelling until the only place it's actually "needed", ie: actually talking to the endpoint. Should we consistently misspell it thru all the layers instead?

curiousepic: Ah, though `build_merchant_mpi_external` only picks up `authenticacionFlow`, right? So I don't think it's actually getting passed. In any case I'd suggest consistency with the sibling fields here.

PR #3796, opened 2020-10-20
Add support for Authorize.net CIM gateway's getTransactionDetails request

joeyparis:
Adds support Authorize.net CIM's `getTransactionDetails` method defined here: https://developer.authorize.net/api/reference/index.html#transaction-reporting-get-transaction-details

The method has worked for me without error in my own environment, but I am having trouble wrapping my head around how the remote/unit testing suites work making it difficult for me to write tests for this added functionality, let alone proper tests that confirm everything. Any guidance or assistance on testing would be greatly appreciated!


curiousepic:

PR #3798, opened 2020-10-23
[Cybersource] Use firstname/lastname from the address instead of the payment method

pi3r:
Some merchants reached out as the firstname/lastname don't match in Cybersource.
This change will use the firstname/lastname from the billing/shipping address instead of
the payment method. We have reached to Cybersource to ensure this is a safe change.

Email I got with Scott Griffiths at Cybersource:
```
Since this code is not owned by Cybersource, but rather an active_merchant opensource repository, we are not sure why the payment method first/last name is persisted to the shipTo and billTo elements.  

Yes, you can safely send the actual shipTo and billTo first and last names to Cybersource.
```

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -520,9 +520,11 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_address(xml, payment_method, address, options, shipTo = false)
+        first_name, last_name = address_names(address[:name], payment_method)
+
         xml.tag! shipTo ? 'shipTo' : 'billTo' do
-          xml.tag! 'firstName',             payment_method.first_name if payment_method
-          xml.tag! 'lastName',              payment_method.last_name if payment_method
+          xml.tag! 'firstName',             first_name if first_name
+          xml.tag! 'lastName',              last_name if last_name
           xml.tag! 'street1',               address[:address1]
           xml.tag! 'street2',               address[:address2] unless address[:address2].blank?
           xml.tag! 'city',                  address[:city]
@@ -537,6 +539,16 @@ module ActiveMerchant #:nodoc:
           xml.tag! 'driversLicenseNumber',  options[:drivers_license_number]  unless options[:drivers_license_number].blank?
           xml.tag! 'driversLicenseState',   options[:drivers_license_state]   unless options[:drivers_license_state].blank?
         end
+      end
+
+      def address_names(address_name, payment_method)
+        names = split_names(address_name)
+        return names if names.any?(&:present?)
+
+        [
+          payment_method&.first_name,
+          payment_method&.last_name
+        ]
       end
 
       def add_creditcard(xml, creditcard)
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -121,6 +121,39 @@ class CyberSourceTest < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
+  def test_uses_names_from_billing_address_if_present
+    name = 'Wesley Crusher'
+
+    stub_comms do
+      @gateway.authorize(100, @credit_card, billing_address: { name: name })
+    end.check_request do |_endpoint, data, _headers|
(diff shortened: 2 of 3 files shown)


test/unit/gateways/cyber_source_test.rb
```diff
@@ -121,6 +121,41 @@ def test_merchant_description
...
+    stub_comms do
+      @gateway.authorize(100, @credit_card, shipping_address: { name: name })
+    end.check_request do |_endpoint, data, _headers|
+      assert_match(%r(<shipTo>.*<firstName>Wesley</firstName>.*</shipTo>)m, data)
+      assert_match(%r(<shipTo>.*<lastName>Crusher</lastName>.*</shipTo>)m, data)
+    end.respond_with(successful_purchase_response)
+  end
+
+  def test_uses_names_from_the_payment_method
+    name = 'Wesley Crusher'
```
bbraschi, line 147: `name` does not seem to be used -> remove?

test/unit/gateways/cyber_source_test.rb
```diff
@@ -121,6 +121,41 @@ def test_merchant_description
...
+  def test_uses_names_from_shipping_address_if_present
+    name = 'Wesley Crusher'
+
+    stub_comms do
+      @gateway.authorize(100, @credit_card, shipping_address: { name: name })
+    end.check_request do |_endpoint, data, _headers|
+      assert_match(%r(<shipTo>.*<firstName>Wesley</firstName>.*</shipTo>)m, data)
+      assert_match(%r(<shipTo>.*<lastName>Crusher</lastName>.*</shipTo>)m, data)
+    end.respond_with(successful_purchase_response)
+  end
```
bbraschi, line 144: Do we need to test that, when both billing and shipping address are present, each is sourced from the proper address? Eg:
```suggestion
  def test_uses_names_from_billing_and_shipping_addresses_when_present
    billing_name = 'Wesley Crusher'
    shipping_name = 'Elvis Preisler'

    stub_comms do
      @gateway.authorize(100, @credit_card, billing_address: { name: billing_name }, shipping_address: { name: shipping_name })
    end.check_request do |_endpoint, data, _headers|
      assert_match(%r(<billTo>.*<firstName>Wesley</firstName>.*</billTo>)m, data)
      assert_match(%r(<billTo>.*<lastName>Crusher</lastName>.*</billTo>)m, data)
      assert_match(%r(<shipTo>.*<firstName>Elvis</firstName>.*</shipTo>)m, data)
      assert_match(%r(<shipTo>.*<lastName>Preisler</lastName>.*</shipTo>)m, data)
    end.respond_with(successful_purchase_response)
  end
```

bbraschi (approved):

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -539,6 +541,16 @@ def add_address(xml, payment_method, address, options, shipTo = false)
         end
       end
 
+      def address_names(address_name, payment_method)
+        names = split_names(address_name)
+        return names if names.any?(&:present?)
```
bbraschi, line 546: ❤️ My notes: this makes sure that the first and last name come from the same source, either the provided `address_name` or  `payment_method`.

bbraschi (approved):

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -520,9 +520,11 @@ def add_purchase_data(xml, money = 0, include_grand_total = false, options = {})
       end
 
       def add_address(xml, payment_method, address, options, shipTo = false)
+        first_name, last_name = address_names(address[:name], payment_method)
+
         xml.tag! shipTo ? 'shipTo' : 'billTo' do
-          xml.tag! 'firstName',             payment_method.first_name if payment_method
-          xml.tag! 'lastName',              payment_method.last_name if payment_method
+          xml.tag! 'firstName',             first_name
+          xml.tag! 'lastName',              last_name
```
bbraschi, line 527: Before, the XML `firstName` or `lastName` would not be set if not defined.
Does this set them even if not defined?

pi3r: You're right!

bbraschi (approved):

PR #3801, opened 2020-10-28
GlobalCollect: Add support for Third-party 3DS2 data

molbrown:
Uses standardized three_d_secure data fields to include 3DS2
authenticated data on the Global Collect gateway

Also adds acs_transaction_id to standardized list (change to
contributing wiki pending approval here). This is another common auth
field that has not been required yet, but is fitting for the
standardized fields.

ECS-1435

Unit:
24 tests, 105 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
23 tests, 57 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.6522% passed
One unrelated failure, also failing on master

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,9 +2,10 @@
 
 == HEAD
 * Checkout V2: Pass attempt_n3d along with 3ds enabled [naashton] #3805
+* GlobalCollect: Add support for Third-party 3DS2 data [molbrown] #3801
 
 == Version 1.116.0 (October 28th)
-* Remove Braintree specific version dependency [pi3r] #38000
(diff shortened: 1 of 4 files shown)


britth (approved):
Looks great to me and new test passes, just looks like there's a couple of rubocop complaints :shipit: 

Do you think it's worth adding a unit test to check the request structure when passing the auth fields? I see that with some of the other supported gateways.

molbrown:
Thanks @britth! I also added some unit tests and addressed the Rubocop, let me know how it looks.

britth:
Looks great!

PR #3805, opened 2020-10-30
Checkout V2: Pass attempt_n3d along with 3ds enabled

naashton:
In some cases `three_d_secure` may not be set while `execute_threed` is
leading to `attempt_n3d` not being passed when `[:'3ds'][:enabled]` is

CE-1080

100% passed

Remote:33 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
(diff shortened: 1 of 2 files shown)


meagabeth (approved):
Tests pass - looks good to me

PR #3806, opened 2020-10-30
Authorize.net: Don't pass isFirstSubsequentAuth

therufs:
Though authorize.net [appears to specify in docs](https://developer.authorize.net/api/reference/index.html#payment-transactions) that `isFirstSubsequentAuth` should be passed in `processingOptions`, doing so results in failure. 

    Unit:
    101 tests, 601 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed
    
    Remote:
    73 tests, 269 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed
    
    CE-1026


InPermutation (approved):
swoosh it

PR #3810, opened 2020-11-03
Spreedly: Additional card types for Stripe, Worldpay, Checkout.com

LinTrieu:
## What has been done

Paddle are integrating with a number of payment gateways, via Spreedly. We have noted missing card types with the Stripe, Worldpay and Checkout.com gateways, which should be supported by active_merchant.

We have added the following card types to the relevant gateway class and its `supported_cardtypes` array.

1. StripeGateway: `unionpay` card type 

2. WorldpayGateway: `unionpay` card type

3. CheckoutV2Gateway: `jcb` card type


meagabeth (commented):
This was reviewed, approved and merged into production!

meagabeth (approved):

PR #3811, opened 2020-11-03
Stripe PI: Pass external 3DS auth data

curiousepic:
Support for passing through external 3DS 1 and 2 auth data. Reference:
https://stripe.com/docs/payments/payment-intents/three-d-secure-import

Remote (1 unrelated failure):
46 tests, 198 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.8261% passed

Unit:
16 tests, 104 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -23,6 +23,7 @@ module ActiveMerchant #:nodoc:
         payment_method = add_payment_method_token(post, payment_method, options)
         return payment_method if payment_method.is_a?(ActiveMerchant::Billing::Response)
 
+        add_external_three_d_secure_auth_data(post, options)
         add_metadata(post, options)
         add_return_url(post, options)
         add_connected_account(post, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -276,6 +277,19 @@ def request_three_d_secure(post, options = {})
...
       end
 
+      def add_external_three_d_secure_auth_data(post, options = {})
+        return unless options[:three_d_secure]&.is_a?(Hash)
+
+        three_d_secure = options[:three_d_secure]
+        post[:payment_method_options] ||= {}
+        post[:payment_method_options][:card] ||= {}
+        post[:payment_method_options][:card][:three_d_secure] ||= {}
+        post[:payment_method_options][:card][:three_d_secure][:version] = three_d_secure[:version] || (three_d_secure[:ds_transaction_id] ? '2.1.0' : '1.0.2')
```
britth, line 287: just curious, but how do we determine the minor version here (e.g. `2.1.0` rather than `2.2.0`)

britth (approved):
Looks great! 🙌 All tests pass for me, and logic and new tests all seem to align with docs 🚀

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
curiousepic: Unfortunately I don't think we can, and Stripe requires that a minor version is specified, so this fallback is hard coded. If happy to bump it if you think 2.2.0 is more common, but figured lower is safer.

britth: That makes sense, was just wondering! The lower version you have now seems like a good starting point to me too 👍

curiousepic: I see now that exemptions are the primary change in 2.2.0. 🤔  Perhaps it's it better to side with 2.2.0, given possible discrepancy?

britth: hmmm, maybe we try 2.2.0 to start and then just update if we notice errors?

britth: I guess really for our purposes, we will most likely always have a version directly passed in

curiousepic:
Closed with https://github.com/activemerchant/active_merchant/commit/ed01055f19f8542ef4655b8c52ba8416acd29fff

PR #3812, opened 2020-11-04
Credorax: Allow 3DS1 normalized pass-through, ease version matching

britth:
Allow use of normalized three_d_secure hash with v1 auth data as
with v2. Also use `start_with?` to check the version so that the
request doesn't error on that parameter when passing a specific
minor version.

Remote
42 tests, 155 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
70 tests, 333 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


curiousepic (approved):
Nice! LGTM 👍

PR #3815, opened 2020-11-10
Response: add network_transaction_id attribute

therufs:
What: Add `network_transaction_id` as an attribute to Response

Why: Elavon (and possibly other gateways) require two different fields for a stored credential network_transaction_id (see #3816). Rather than dividing management duties between the host library and AM, provide a unified interface so that concatenation and division can occur within a single codebase.


naashton (approved):
This looks good. I saw this same update in the related PR for Elavon, but I think it's a good idea to merge them separately for commit/changelog history.

PR #3816, opened 2020-11-10
Elavon: support standardized stored creds

therufs:


lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -337,6 +338,27 @@ def add_line_items(xml, level_3_data)
...
+        when network_transaction_id.nil?
+          return
+        when network_transaction_id.to_s.include?('|')
+          oar_data, ps2000_data = options[:stored_credential][:network_transaction_id].split('|')
+          xml.ssl_oar_data oar_data unless oar_data.nil? || oar_data.empty?
+          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty?
+        when network_transaction_id.to_s.length > 22
+          xml.ssl_oar_data options.dig(:stored_credential, :network_transaction_id)
+        else
+          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
```
naashton, line 353: In the Elavon docs 

> ssl_oar_data
20 | alphanumeric

> ssl_ps2000_data
22 | alphanumeric

Does this not hold true? I took a peak at the `network_transaction_id`'s length and they were all 83 characters long.

naashton (approved):
I'm a little confused by the docs for `ssl_oar_data` saying the character limit is 20, but it seems to work when it gets a value with way more than that.

Looks good to ship though.

lib/active_merchant/billing/gateways/elavon.rb
therufs: I'm looking at https://developer.elavon.com/#/api/eb6e9106-0172-4305-bc5a-b3ebe832f823.rcosoomi/documents?converge-integration-guide/book/transaction_types/credit_card/auth_only.html which lists a max of 60, which seems to math out right (60 + pipe + 22 = 83) in addition to having the benefit of matching what they send :)

PR #3818, opened 2020-11-12
Feature/ecp orbital

ajawadmirza:
ECP Chase Orbital-Electronic Check Integration:

1. ECP Standard Processing
2. ECP Authorization Methods
3. ECPSameDay Index Processing


ajawadmirza:
**ECP Chase Orbital-Electronic Check Integration:**

- ECP Standard Processing
- ECP Authorization Methods
- ECPSameDay Index Processing

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -389,6 +490,34 @@ def test_successful_purchase_with_overridden_normalized_stored_credentials
     assert_equal 'Approved', response.message
   end
 
+  # Successful force capture on Electronic Check
+  def test_successful_force_capture_with_echeck
+    @options[:force_capture] = true
+    assert response = @gateway.purchase(@amount, @echeck, @options)
+    assert_success response
+    assert_equal 'APPROVAL        ', response.message
```
jessiagee, line 498: Do you need all of this blank space after APPROVAL?

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -442,6 +599,23 @@ def test_refund
...
+    assert response.authorization
+    assert refund = @gateway.refund(@amount, response.authorization, @options)
+    assert_success refund
+  end
+
+  # Failed refund due to invalid Transaction provided
+  def test_failed_refund_with_echeck_due_to_invalid_authorization
+    assert refund = @gateway.refund(@amount, '123;123', @options)
+    assert_failure refund
+    assert_equal 'The LIDM you supplied (3F3F3F) does not match with any existing transaction', refund.message
```
jessiagee, line 615: Should this be 123;123 instead of 3F3F3F ?

test/unit/gateways/orbital_test.rb
```diff
@@ -108,6 +110,39 @@ def test_successful_purchase
...
+    assert_equal '888', response.params['proc_status']
+  end
+
+  # Successful Force Capture (FC) through Electronic Check
+  def test_successful_force_capture_with_echeck
+    @gateway.expects(:ssl_post).returns(successful_force_capture_with_echeck_response)
+
+    assert response = @gateway.purchase(31, @echeck, order_id: '2', force_capture: true)
+    assert_instance_of Response, response
+    assert_equal 'APPROVAL        ', response.message
```
jessiagee, line 141: Do we need all of this white space behind APPROVAL?

test/unit/gateways/orbital_test.rb
```diff
@@ -882,6 +917,49 @@ def test_currency_code_and_exponent_are_set_for_profile_authorizations
...
+    assert_failure response
+    assert_equal 'Error validating amount. Must be numeric, equal to zero or greater [-1]', response.message
+    assert_equal '885', response.params['proc_status']
+  end
+
+  # Successful refund through Electronic Check
+  def test_successful_refund_with_echeck
+    @gateway.expects(:ssl_post).returns(successful_refund_with_echeck_response)
+
+    assert response = @gateway.refund(50, '1;2', @options)
```
jessiagee, line 946: Is '1;2' the correct param to pass? I don't see the difference between that and the failure below.

jessiagee (changes requested):

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -478,6 +499,32 @@ def add_customer_address(xml, options)
         end
       end
 
+      # Payment can be done through either Credit Card or Electronic Check
+      def add_payment_source(xml, payment_source, options = {})
+        if payment_source.instance_of?(ActiveMerchant::Billing::Check)
+          add_echeck(xml, payment_source, options)
+        else
+          add_creditcard(xml, payment_source, options[:currency])
```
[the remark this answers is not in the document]
ajawadmirza: If we do this in the current code: 
`
        if payment_source.instance_of?(ActiveMerchant::Billing::Check)
          add_echeck(xml, payment_source, options)
        elsif payment_source.instance_of?(ActiveMerchant::Billing::CreditCard) or payment_source.instance_of?(ActiveMerchant::Billing::NetworkTokenizationCreditCard)
          add_creditcard(xml, payment_source, options[:currency])
        end
`

then in that case, Orbital's previously existing two unit test cases are failed as in the previous implementation CurrencyCode needs to be added even when no payment source is passed (please refer to image below):

image

Now, we can have three possibilities here:
1. Keep it as it is so that everything works fine as in the case of no credit card, some required attributes will be added
2. Make changes in previous Orbital code and debug so that it should work even when no payment source is added
3. Make changes in the tests of the previous orbital's implementation

Please suggest on this?

test/remote/gateways/remote_orbital_test.rb
ajawadmirza: Blank spaces are already produced in our response, that's why it was put here to match it exactly. However, we can the condition so that it checks whether APPROVAL is contained in the response.message.

test/remote/gateways/remote_orbital_test.rb
ajawadmirza: Whatever value we provide in the authorization string, server produces response that has 3F3F3F and please note that 123 doesn't mean something special but it just an invalid number passed to check the failure case.

test/unit/gateways/orbital_test.rb
ajawadmirza: Already answered this before.

test/unit/gateways/orbital_test.rb
ajawadmirza: @jessiagee , please note that this is a unit test, where we produce static response to test the correct rendering of the response. Here the response is already given in the **successful_refund_with_echeck_response**. So, the values are just passed so that the method can be called correctly and produce our static response.

Previous one was an actual remote test case where we hit the server and get the response, so in that case we intentionally gave incorrect value to test the failure (negative) case.

lib/active_merchant/billing/gateways/orbital.rb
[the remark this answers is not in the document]
jessiagee: If there's existing code that this check would break than let's avoid adding that in. I don't want to mess with any legacy code that could potentially cause a lot more issues to existing customers. Thanks!

test/remote/gateways/remote_orbital_test.rb
jessiagee: Okay thanks for the information!

test/unit/gateways/orbital_test.rb
jessiagee: Thanks for the clarification!

jessiagee (approved):

PR #3820, opened 2020-11-14
Second iteration of mit gateway development

EsporaInfra:
Second iteration of mit gateway integration, the gateway includes operations for capture, authorization and refund

### DIFF
--- a/lib/active_merchant/billing/gateways/mit.rb
+++ b/lib/active_merchant/billing/gateways/mit.rb
@@ -1,0 +1,258 @@
+require 'json'
+require 'openssl'
+require 'digest'
+require 'base64'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class MitGateway < Gateway
+      self.live_url = 'https://wpy.mitec.com.mx/ModuloUtilWS/activeCDP.htm'
+
+      self.supported_countries = ['MX']
+      self.default_currency = 'MXN'
+      self.supported_cardtypes = %i[visa master]
+
+      self.homepage_url = 'http://www.centrodepagos.com.mx/'
+      self.display_name = 'MIT Centro de pagos'
+
+      self.money_format = :dollars
+
+      def initialize(options = {})
+        requires!(options, :commerce_id, :user, :api_key, :key_session)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        MultiResponse.run do |r|
+          r.process { authorize(money, payment, options) }
+          r.process { capture(money, r.authorization, options) }
+        end
+      end
+
+      def cipher_key
+        @options[:key_session]
+      end
+
+      def decrypt(val, keyinhex)
+        # Splits the first 16 bytes (the IV bytes) in array format
+        unpacked = val.unpack('m')
+        iv_base64 = unpacked[0].bytes.slice(0, 16)
+        # Splits the second bytes (the encrypted text bytes) these would be the
+        # original message
+        full_data = unpacked[0].bytes.slice(16, unpacked[0].bytes.length)
+        # Creates the engine
+        engine = OpenSSL::Cipher::AES128.new(:CBC)
+        # Set engine as decrypt mode
+        engine.decrypt
+        # Converts the key from hex to bytes
+        engine.key = [keyinhex].pack('H*')
+        # Converts the ivBase64 array into bytes
+        engine.iv = iv_base64.pack('c*')
+        # Decrypts the texts and returns the original string
+        engine.update(full_data.pack('c*')) + engine.final
+      end
+
+      def encrypt(val, keyinhex)
+        # Creates the engine motor
+        engine = OpenSSL::Cipher::AES128.new(:CBC)
+        # Set engine as encrypt mode
+        engine.encrypt
+        # Converts the key from hex to bytes
+        engine.key = [keyinhex].pack('H*')
+        # Generates a random iv with this settings
+        iv_rand = engine.random_iv
+        # Packs IV as a Base64 string
+        iv_base64 = [iv_rand].pack('m')
+        # Converts the packed key into bytes
+        unpacked = iv_base64.unpack('m')
+        iv = unpacked[0]
+        # Sets the IV into the engine
+        engine.iv = iv
+        # Encrypts the texts and stores the bytes
+        encrypted_bytes = engine.update(val) + engine.final
+        # Concatenates the (a) IV bytes and (b) the encrypted bytes then returns a
+        # base64 representation
+        [iv << encrypted_bytes].pack('m')
+      end
+
+      def authorize(money, payment, options = {})
+        post = {
+          operation: 'Authorize',
+          commerce_id: @options[:commerce_id],
+          user: @options[:user],
+          apikey: @options[:api_key],
+          testMode: (test? ? 'YES' : 'NO')
+        }
+        add_invoice(post, money, options)
+        # Payments contains the card information
+        add_payment(post, payment)
+        add_customer_data(post, options)
+        post[:key_session] = @options[:key_session]
+
+        post_to_json = post.to_json
+        post_to_json_encrypt = encrypt(post_to_json, @options[:key_session])
+
+        final_post = '<authorization>' + post_to_json_encrypt + '</authorization><dataID>' + @options[:user] + '</dataID>'
+        json_post = {}
+        json_post[:payload] = final_post
+        commit('sale', json_post)
+      end
+
+      def capture(money, authorization, options = {})
+        post = {
+          operation: 'Capture',
+          commerce_id: @options[:commerce_id],
+          user: @options[:user],
+          apikey: @options[:api_key],
+          testMode: (test? ? 'YES' : 'NO'),
+          transaction_id: authorization,
+          amount: amount(money)
+        }
+        post[:key_session] = @options[:key_session]
+
+        post_to_json = post.to_json
+        post_to_json_encrypt = encrypt(post_to_json, @options[:key_session])
+
+        final_post = '<capture>' + post_to_json_encrypt + '</capture><dataID>' + @options[:user] + '</dataID>'
+        json_post = {}
+        json_post[:payload] = final_post
+        commit('capture', json_post)
+      end
+
+      def refund(money, authorization, options = {})
+        post = {
+          operation: 'Refund',
+          commerce_id: @options[:commerce_id],
+          user: @options[:user],
+          apikey: @options[:api_key],
+          testMode: (test? ? 'YES' : 'NO'),
+          transaction_id: authorization,
+          auth: authorization,
+          amount: amount(money)
+        }
+        post[:key_session] = @options[:key_session]
+
+        post_to_json = post.to_json
+        post_to_json_encrypt = encrypt(post_to_json, @options[:key_session])
+
+        final_post = '<refund>' + post_to_json_encrypt + '</refund><dataID>' + @options[:user] + '</dataID>'
+        json_post = {}
+        json_post[:payload] = final_post
+        commit('refund', json_post)
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        ret_transcript = transcript
+        auth_origin = ret_transcript[/<authorization>(.*?)<\/authorization>/, 1]
+        unless auth_origin.nil?
+          auth_decrypted = decrypt(auth_origin, @options[:key_session])
+          auth_json = JSON.parse(auth_decrypted)
+          auth_json['card'] = '[FILTERED]'
+          auth_json['cvv'] = '[FILTERED]'
+          auth_json['apikey'] = '[FILTERED]'
+          auth_json['key_session'] = '[FILTERED]'
+          auth_to_json = auth_json.to_json
+          auth_tagged = '<authorization>' + auth_to_json + '</authorization>'
+          ret_transcript = ret_transcript.gsub(/<authorization>(.*?)<\/authorization>/, auth_tagged)
+        end
+
+        cap_origin = ret_transcript[/<capture>(.*?)<\/capture>/, 1]
+        unless cap_origin.nil?
+          cap_decrypted = decrypt(cap_origin, @options[:key_session])
+          cap_json = JSON.parse(cap_decrypted)
+          cap_json['apikey'] = '[FILTERED]'
+          cap_to_json = cap_json.to_json
+          cap_tagged = '<capture>' + cap_to_json + '</capture>'
+          ret_transcript = ret_transcript.gsub(/<capture>(.*?)<\/capture>/, cap_tagged)
+        end
+
+        ref_origin = ret_transcript[/<refund>(.*?)<\/refund>/, 1]
+        unless ref_origin.nil?
+          ref_decrypted = decrypt(ref_origin, @options[:key_session])
+          ref_json = JSON.parse(ref_decrypted)
+          ref_json['apikey'] = '[FILTERED]'
+          ref_to_json = ref_json.to_json
+          ref_tagged = '<refund>' + ref_to_json + '</refund>'
+          ret_transcript = ret_transcript.gsub(/<refund>(.*?)<\/refund>/, ref_tagged)
+        end
+
+        res_origin = ret_transcript[/#{Regexp.escape('reading ')}(.*?)#{Regexp.escape('read')}/m, 1]
+        loop do
+          break if res_origin.nil?
+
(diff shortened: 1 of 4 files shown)


EsporaInfra (commented):
The cardholder's name for unit tested was changed

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,244 @@
...
+
+        post_to_json = post.to_json
+        puts(post_to_json)
+        post_to_json_encrypt = encrypt(post_to_json, self.options[:key_session])
+
+        final_post = post_to_json_encrypt + '-' + self.options[:user]
+        commit('refund', final_post)
+      end
+
+      # Currently unsupported
```
Carrigan, line 157: Since this is unsupported, I'd recommend that it be removed instead of simply being commented out.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,244 @@
...
+        false
+      end
+
+      private
+
+      def add_customer_data(post, options)
+        post[:email] = options[:email] || 'nadie@mit.test'
+      end
+
+      def add_address(post, creditcard, options); end
```
Carrigan, line 182: There is no reason to keep in methods that were in the template but not used in the adapter. `add_address` and `parse` can both be removed.

test/remote/gateways/remote_mit_test.rb
```diff
@@ -0,0 +1,127 @@
...
+    @options = {
+      order_id: '721',
+      transaccion_id: '721', # unique id for every transaction, needs to be generated for every test
+      billing_address: address,
+      description: 'Store Purchase'
+    }
+  end
+
+  def test_successful_purchase
+    puts('======================================================================================')
```
Carrigan, line 44: Similar to above, we would like to not have `puts` statements in any of our test code.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,244 @@
...
+      def parse(body)
+        {}
+      end
+
+      def commit(action, parameters)
+        url = (test? ? test_url : live_url)
+        raw_response = ssl_post(url, parameters, { 'Content-type' => 'text/plain' })
+        puts(raw_response)
+        response = JSON.parse(decrypt(raw_response, self.options[:key_session]))
+        puts(response)
```
Carrigan, line 209: If you'd like to view the request and response data in your testing, feel free to add `DEBUG_ACTIVE_MERCHANT=true` to your test command.

test/remote/gateways/remote_mit_test.rb
```diff
@@ -0,0 +1,127 @@
+require 'test_helper'
+
+class RemoteMitTest < Test::Unit::TestCase
```
Carrigan, line 3: When running the tests, I get three failing tests in this file:

```
Failure: test_successful_authorize_and_capture(RemoteMitTest)
Failure: test_successful_purchase(RemoteMitTest)
Failure: test_successful_refund(RemoteMitTest)
```

The failures all look similar:
```ruby
{
     "folio_cdp"=>"201397145",
     "auth"=>"",
     "response"=>"denied",
     "message"=>" -15- ",
     "id_comercio"=>"147",
     "reference"=>"TID|1615390643",
     "amount"=>"11.15",
     "time"=>"15:37:24 10:03:2021",
     "operation"=>"Authorize"
}
```

Any idea why these are not passing? I cannot find any documentation on this error code.

Carrigan (changes requested):
Hello and thank you for submitting this new gateway to ActiveMerchant 🎉 

I've reviewed this submission and left some preliminary requests and suggestions below. In addition, I am curious to know if there are any public documents that detail this API.

Please feel free to request a review from `activemerchant/spreedly-integrating` when these changes are implemented and you are ready for us to take another look. You can leave any additional questions you have in this PR and we'd be happy to help. Cheers!

EsporaInfra:
Hello and thank you for reviewing our implementation for this gateway!

The changes you've requested have been implemented and the code updated in
the repository. We are currently working on documentation. In regards to
your questions about the operations that are failing I asked the gateway
and they answered me this:

The gateway does not currently support a sandbox environment, therefore in
order to do a test of authorization and capture that are required to
succeed, data from a real credit card needs to be used. In order for you to
run the remote tests they are willing to supply the information of a test
card you can use for minimal amount purchases.

Regarding the refund operation, the gateway does not support refund unless
there's been a minimum of twenty four hours after the purchase has been
generated.

If you require test card information to run the remote tests I'll come back
to you with it, If not then the implementation is ready for a review.

Thank you so much in advance.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,218 @@
...
+      self.default_currency = 'MXN'
+
+      self.supported_cardtypes = %i[visa master]
+
+      self.homepage_url = 'http://www.centrodepagos.com.mx/'
+      self.display_name = 'MIT Centro de pagos'
+
+      self.money_format = :dollars
+
+      STANDARD_ERROR_CODE_MAPPING = {}
```
Carrigan, line 22: If not used, this can be removed.

Carrigan (commented):
Thanks for the quick changes and info @EsporaInfra!

Per [the contributions guide](https://github.com/activemerchant/active_merchant/wiki/Contributing), ActiveMerchant typically does not accept contributions without a functional remote test suite.  I will check in with our team about what our options are regarding a test suite that relies on live cards. Between this and not having documentation, it is difficult to verify this integration. 

Otherwise I left some additional comments below that still need to be addressed at the code level.

EsporaInfra:
I've talked about the necessity of a test mode for the gateway and they are
implementing it. It may take about two weeks on their end to implement it
and as soon as it;s done I will upload the changes to the repository.

Thank you so much for your support

EsporaInfra:
Hello

After requesting a test mode from the gateway we have implemented the
changes as well as removed the lines you pointed out, the documentation
will soon be public and you can perform the tests with no issues

Thanks in advance

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,215 @@
...
+      end
+
+      def purchase(money, payment, options = {})
+        authorization = authorize(money, payment, options)
+        return authorization unless authorization.success?
+
+        capture(money, payment, options)
+      end
+
+      def cipher
```
Carrigan, line 34: Can this function be removed? It looks like it is currently not being used.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,215 @@
...
+
+      def decrypt(val, keyinhex)
+        # Splits the first 16 bytes (the IV bytes) in array format
+        unpacked = val.unpack('m')
+        iv_base64 = unpacked[0].bytes.slice(0, 16)
+        # Splits the second bytes (the encrypted text bytes) these would be the
+        # original message
+        full_data = unpacked[0].bytes.slice(16, unpacked[0].bytes.length)
+        # Creates the engine
+        engine = OpenSSL::Cipher::Cipher.new('AES-128-CBC')
```
Carrigan, line 50: I receive a warning when running my unit tests: `warning: constant OpenSSL::Cipher::Cipher is deprecated`. Recommend using the new style of Cipher creation here: `OpenSSL::Cipher::AES128.new(:CBC)` here and on line 63.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,215 @@
...
+        [iv << encrypted_bytes].pack('m')
+      end
+
+      def authorize(money, payment, options = {})
+        post = {
+          operation: 'Authorize',
+          id_comercio: self.options[:id_comercio],
+          user: self.options[:user],
+          apikey: self.options[:apikey],
+          testMode: self.options[:testMode]
```
Carrigan, line 90: To keep up our convention of using underscores between words, would it be possible to change this to `options[:test_mode]` instead of `testMode`? It is of course fine to use `testMode` in the payload going to the gateway, but for our internal options hash, underscore separated words are preferred.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,215 @@
...
+        encrypted_bytes = engine.update(val) + engine.final
+        # Concatenates the (a) IV bytes and (b) the encrypted bytes then returns a
+        # base64 representation
+        [iv << encrypted_bytes].pack('m')
+      end
+
+      def authorize(money, payment, options = {})
+        post = {
+          operation: 'Authorize',
+          id_comercio: self.options[:id_comercio],
```
Carrigan, line 87: I believe these initial `post = { ... }` hash options should use just `options[:key]` instead of `self.options[:key]`. This applies for here, capture, and refund.

Carrigan (commented):
Thank you very much! All of the changes look great :+1: I was able to run the remote and unit tests locally without any issues. We appreciate MIT's work to make this easier to test.

Now that my testing can go farther, I identified a couple more issues below that will need to be resolved.

EsporaInfra:
Thank you for your feedback!

I've made most of the changes, with only one exception. In your review you
mention that the post calls need the self options replaced with options to
retrieve the data for the transactions. The reason they are retrieved from
self.options is that the five fields are provided to the gateway upon
creation (through the fixtures in the case of the remote tests) and remain
constant through the different transactions. Is it required to include them
on every transaction call to the gateway as part of the function call?
Would the preferred practice be to remove them from the fixtures and
directly include them in the calls of the transactions?

Thanks in advance

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,223 @@
...
+      end
+
+      def capture(money, authorization, options = {})
+        post = {
+          operation: 'Capture',
+          id_comercio: self.options[:id_comercio],
+          user: self.options[:user],
+          apikey: self.options[:apikey],
+          testMode: self.options[:test_mode],
+          transaccion_id: options[:transaccion_id],
```
tatsianaclifton, line 110: Firs of all, I think it should be named `transaction_id` not `transaccion_id` within option hash.

Additionally, could you please provide clarification on that parameter? I see that for the authorize method it is not needed and standard `:order_id` is used. Is there any reason why it is different for capture and refund methods? 

I also don't see authorization parameter that is passed to the capture method is used. So how is capture linked to the authorize?

EsporaInfra:
Thank you for your review. We have made the requested changes and updated
the repository accordingly.

In regards to your question about the relationship between authorization
and capture: The operations are linked internally by the gateway. When an
authorization request is made, the information is saved at the gateway and
a transaction id is used to link this operation to any further
capture/refund requests. This transaction id is required by the gateway in
capture requests done by businesses.

Thanks in advance.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,262 @@
...
+      end
+
+      def capture(money, authorization, options = {})
+        post = {
+          operation: 'Capture',
+          commerce_id: @options[:commerce_id],
+          user: @options[:user],
+          apikey: @options[:api_key],
+          testMode: (test? ? 'YES' : 'NO'),
+          transaction_id: options[:transaction_id],
```
tatsianaclifton, line 109: The options hash does not have `transaction_id` as mentioned here https://github.com/activemerchant/active_merchant/wiki/Contributing#options-hash. But the capture method is defined with `authorization` parameter that should be used to link the capture request to the authorize request.

lib/active_merchant/billing/gateways/mit.rb
```diff
@@ -0,0 +1,223 @@
...
+      def initialize(options = {})
+        requires!(options, :id_comercio, :user, :apikey, :key_session)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        authorization = authorize(money, payment, options)
+        return authorization unless authorization.success?
+
+        capture(money, payment, options)
```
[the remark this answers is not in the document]
tatsianaclifton: I am still concerned with the way capture is implemented. It is very different that all other Active Merchant gateways are implemented. The standard is that the authorize action is performed and the reference is retrieved and passed to the capture or any other reference actions (refund etc.). I recommend checking the Adyen gateway to see how it is implemented there https://github.com/tatsianaclifton/active_merchant/blob/master/lib/active_merchant/billing/gateways/adyen.rb#L38
If you change this method to retrieve the reference appropriately you can run the MultiResponse with authorize and capture the same way as the Adyen gateway
```
  response['auth']
end 
```

tatsianaclifton (changes requested):
Thank you for making some requested changes and implementing the scrub method. I have left more review comments, some of them are requests for small changes. The main concern is with the capture and the refund methods. You mentioned that the gateway rely on the `transaction_id` as a reference. However, the `transaction_id` is not a standard field in the options hash. I added some explanations as comments. Please let me know if something is unclear there. I'll try my best to clarify.

Also, please share documentation if it is available.

EsporaInfra:
Thank you for reviewing the latest changes to the gateway implementation!

We've made several changes in accordance to your requests. The only exception is that the gateway currently does not support the void operation and thus has been removed from the gateway's code.

We've made changes to the syntax in the conditions, also we changed the way the capture/refund operations receive the information of authorization to be in line with the observations you've made.

test/remote/gateways/remote_mit_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    assert_not_equal 'approved', response.message
+  end
+
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options_success)
+    end
+
+    clean_transcript = @gateway.scrub(transcript)
+    assert_scrubbed(@credit_card.number, transcript)
```
tatsianaclifton, line 129:
```suggestion
    assert_scrubbed(@credit_card.number, clean_transcript)
```

test/remote/gateways/remote_mit_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    assert_scrubbed(@credit_card.verification_value, transcript)
```
tatsianaclifton, line 130:
```suggestion
    assert_scrubbed(@credit_card.verification_value, clean_transcript)
```

test/remote/gateways/remote_mit_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    assert_scrubbed(@gateway.options[:api_key], transcript)
```
tatsianaclifton, line 131:
```suggestion
    assert_scrubbed(@gateway.options[:api_key], clean_transcript)
```

test/unit/gateways/mit_test.rb
```diff
@@ -0,0 +1,218 @@
...
+  end
+
+  def test_failed_refund
+    @gateway.expects(:ssl_post).returns(failed_refund_response)
+    response = @gateway.refund(@amount, 'authorizationtest', @options)
+    assert_failure response
+    assert_not_equal 'approved', response.message
+  end
+
+  def test_transcript_scrubbing
```
tatsianaclifton, line 104: For unit tests, I recommend writing a test that is similar for other gateways where assertion is performed not for individual fields, but for the whole transcript. Here is an example,
  ```
def test_transcript_scrubbing
      assert_equal post_scrubbed, @gateway.scrub(pre_scrubbed)
   end
```

tatsianaclifton (changes requested):
Thanks for making changes. I have just some small suggestions. Once they are addressed please follow instructions here to rebase with master and squash commits https://github.com/activemerchant/active_merchant/wiki/Submitting-Your-Gateway#rebasing-after-review

Thanks again

tatsianaclifton:
@EsporaInfra I think there were some issues while rebasing. I see 161 commits and many changes that are already in master. Please ensure that it is just 1 commit and only needed changes are in this pull requests. Here is the link to instructions on how to squash and rebase https://github.com/activemerchant/active_merchant/wiki/Submitting-Your-Gateway#rebasing-after-review.

EsporaInfra:
@tatsianaclifton Thanks for pointing out the issue with the rebase! We have resubmitted it as a single commit.

tatsianaclifton:
> @tatsianaclifton Thanks for pointing out the issue with the rebase! We have resubmitted it as a single commit.

Thank you. All looks good now. I will just make couple small changes and merge this RP.

tatsianaclifton (approved):
Looks good to me. Thanks

PR #3821, opened 2020-11-16
SafeCharge: 3DS external MPI data refinements

curiousepic:
Prevent xid field from being sent when external MPI data is provided for
3DS version 2. Lock version values to singe digit keyed from presence
of a DS Transaction ID.

Remote (2 unrelated failures):
27 tests, 69 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92.5926% passed

Unit:
23 tests, 120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


Carrigan (approved):
Nice :+1: Makes sense to me and looks good

PR #3834, opened 2020-12-02
Credorax: Add support for 3DS Adviser

meagabeth:
The Credorax 3DS Adviser module depends on the use of the `3ds_initiate` parameter and can be controlled more specifically with the use of the `f23` parameter. If the `3ds_initiate` parameter isn’t received in the original request, it will be sent with a value of ’01’ by default. The `f23` parameter is optional and has no default if it is not sent in the original request.

Unit:
69 tests, 331 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
42 tests, 155 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

CE-1082

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -331,6 +331,7 @@ module ActiveMerchant #:nodoc:
           three_ds_2_options = options[:three_ds_2]
           browser_info = three_ds_2_options[:browser_info]
           post[:'3ds_initiate'] = options[:three_ds_initiate] || '01'
+          post[:f23] = options[:f23] if options[:f23]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -330,6 +330,7 @@ def add_3d_secure(post, options)
           three_ds_2_options = options[:three_ds_2]
           browser_info = three_ds_2_options[:browser_info]
           post[:'3ds_initiate'] = options[:three_ds_initiate] || '01'
+          post[:'f23'] = options[:f23] if options[:f23]
```
naashton, line 333: Take the quotes out for `f23` and you're GTG! 🚢

naashton (approved):
I'll let you fix it during the rebase. 🚀

PR #3847, opened 2020-12-16
add network tokenization support

DStoyanoff:


test/remote/gateways/remote_safe_charge_test.rb
```diff
@@ -19,6 +19,16 @@ def setup
...
     @three_ds_invalid_pa_res_card = credit_card('4012 0010 3749 0006')
+
+    @network_token_credit_card = ActiveMerchant::Billing::NetworkTokenizationCreditCard.new({
+      brand: 'Visa',
+      payment_cryptogram: 'UnVBR0RlYm42S2UzYWJKeWJBdWQ=',
+      number: '4012001037490014',
+      source: :network_token,
+      month: '12',
+      year: 2020,
+      verification_value: '217'
```
jimilpatel24, line 30: Why do we need to set the **month, year** and **verification_value** params? Is the network token number and cryptogram not enough? I am comparing this with [Cybersource's implementation](https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_cyber_source_test.rb#L533-L537) (which already works well) and they don't require these params. Just trying to make sure we can be consistent in supporting NT for different gateways.

lib/active_merchant/billing/gateways/safe_charge.rb
```diff
@@ -171,12 +178,11 @@ def add_customer_details(post, payment, options)
...
         version = options[:three_d_secure][:ds_transaction_id] ? '2' : '1'
-
-        post[:sg_eci] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
-        post[:sg_cavv] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
+        post[:sg_ECI] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
+        post[:sg_CAVV] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
         post[:sg_dsTransID] = options[:three_d_secure][:ds_transaction_id] if version == '2'
         post[:sg_threeDSProtocolVersion] = version
-        post[:sg_xid] = options[:three_d_secure][:xid] if version == '1'
+        post[:sg_Xid] = options[:three_d_secure][:xid] if version == '1'
```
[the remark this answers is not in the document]
DStoyanoff: According to the docs I have from Nuvei, it should be capital

image

test/remote/gateways/remote_safe_charge_test.rb
DStoyanoff: From what I remember, the request was failing if those are not provided (some sort of card validation was failing). As for the CVV, I got an explicit request from Nuvei saying: "Yes – if is provided, forward it." so I believe this answers your question.

lib/active_merchant/billing/gateways/safe_charge.rb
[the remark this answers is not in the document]
jimilpatel24: Thanks Daniel! Can you confirm that you are looking at latest API docs from Nuvei for this? We couldn't find this change in Nuvei's public facing docs.

jimilpatel24 (approved):
Changes look good to me. Approved.

lib/active_merchant/billing/gateways/safe_charge.rb
```diff
@@ -146,12 +146,19 @@ def add_transaction_data(trans_type, post, money, options)
...
-        post[:sg_CardNumber] = payment.number
         post[:sg_ExpMonth] = format(payment.month, :two_digits)
         post[:sg_ExpYear] = format(payment.year, :two_digits)
         post[:sg_CVV2] = payment.verification_value
-        post[:sg_StoredCredentialMode] = (options[:stored_credential_mode] == true ? 1 : 0)
+        post[:sg_CardNumber] = payment.number
+
+        if payment.is_a?(NetworkTokenizationCreditCard) && payment.source == :network_token
+          post[:sg_CAVV] = payment.payment_cryptogram
+          post[:sg_ECI] = options[:three_d_secure][:eci] || '05'
```
jimilpatel24, line 155: @DStoyanoff - I think the logic for the following line of code is wrong:
`post[:sg_ECI] = options[:three_d_secure][:eci] || '05'`

If the value for options[:three_d_secure][:eci] is not set, it returns a Nil error instead of defaulting to a value of '05'

Please correct it asap. Since I have already merged this PR, you will have to create a new branch and then create a new PR for the fix.

DStoyanoff: Hello @jimilpatel24. I have opened a new PR - https://github.com/activemerchant/active_merchant/pull/3920
Strangely it shows more diff then expected, as if this PR was not merged.. Let me know if there is anything else I can do

PR #3850, opened 2020-12-23
Orbital: Update 3DS support for Mastercard 

meagabeth:
WIP - Orbital: Add 3DS2 support

Current status: methods have been built to add `MCProgramProtocol`, `MCDirectoryTransID` and `UCAFInd`. Remote tests are returning the error, "Request does not adhere to the DTD. Please correct and send again." Which indicates a XML schema validation issue. Using `xmllint`, we have locally validated that what we are sending in the request matches their latest published XML request schema architecture. We have a ticket open with Orbital's Support team for this issue.

CE-1242

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@
 * Add 3DS 2.0 values to paypal [nebdil] #3285
 * Redsys: Update Mpi Fields [tatsianaclifton] #3855
 * Paypal: Update AuthStatus3ds MPI field [curiousepic] #3857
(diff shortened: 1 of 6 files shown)


therufs (approved):
looks good to me! two remote test failures, also failing on master:
test_successful_american_express_authorization_with_3ds
test_successful_american_express_purchase_with_3ds

meagabeth:
For posterity: The original issue we were having was based on the XML schema version we were referencing. Orbital has since upgraded from 7.7 and is now using 8.1 or higher.

PR #3855, opened 2020-12-29
Redsys: Update Mpi Fields

tatsianaclifton:
Added support for using three_ds_server_trans_id as threeDSServerTransID if provided
Added support to set Eci when 3ds version 2

ECS-1599

Unit:
33 tests, 104 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
Remote
21 tests, 60 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
90.4762% passed

The following tests failed on master:
test_successful_authorise_using_vault_id
test_successful_purchase_using_vault_id

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys.rb
+++ b/lib/active_merchant/billing/gateways/redsys.rb
@@ -333,18 +333,18 @@ module ActiveMerchant #:nodoc:
         return unless options[:three_d_secure]
 
         if options[:three_d_secure][:version] == THREE_DS_V2
-          data[:threeDSServerTransID] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
+          data[:threeDSServerTransID] = options[:three_d_secure][:three_ds_server_trans_id] if options[:three_d_secure][:three_ds_server_trans_id]
           data[:dsTransID] = options[:three_d_secure][:ds_transaction_id] if options[:three_d_secure][:ds_transaction_id]
           data[:authenticacionValue] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
           data[:protocolVersion] = options[:three_d_secure][:version] if options[:three_d_secure][:version]
-
           data[:authenticacionMethod] = options[:authentication_method] if options[:authentication_method]
           data[:authenticacionType] = options[:authentication_type] if options[:authentication_type]
           data[:authenticacionFlow] = options[:authentication_flow] if options[:authentication_flow]
+          data[:eci_v2] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
         elsif options[:three_d_secure][:version] == THREE_DS_V1
           data[:txid] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
           data[:cavv] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
-          data[:eci] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
+          data[:eci_v1] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
         end
       end
 
@@ -525,13 +525,13 @@ module ActiveMerchant #:nodoc:
         ds_merchant_mpi_external = {}
         ds_merchant_mpi_external[:TXID] = data[:txid] if data[:txid]
         ds_merchant_mpi_external[:CAVV] = data[:cavv] if data[:cavv]
-        ds_merchant_mpi_external[:ECI] = data[:eci] if data[:eci]
+        ds_merchant_mpi_external[:ECI] = data[:eci_v1] if data[:eci_v1]
 
         ds_merchant_mpi_external[:threeDSServerTransID] = data[:threeDSServerTransID] if data[:threeDSServerTransID]
         ds_merchant_mpi_external[:dsTransID] = data[:dsTransID] if data[:dsTransID]
         ds_merchant_mpi_external[:authenticacionValue] = data[:authenticacionValue] if data[:authenticacionValue]
         ds_merchant_mpi_external[:protocolVersion] = data[:protocolVersion] if data[:protocolVersion]
-
+        ds_merchant_mpi_external[:Eci] = data[:eci_v2] if data[:eci_v2]
         ds_merchant_mpi_external[:authenticacionMethod] = data[:authenticacionMethod] if data[:authenticacionMethod]
         ds_merchant_mpi_external[:authenticacionType] = data[:authenticacionType] if data[:authenticacionType]
         ds_merchant_mpi_external[:authenticacionFlow] = data[:authenticacionFlow] if data[:authenticacionFlow]
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,6 +18,7 @@
 * Redsys: add_payment method solution [meagabeth] #3845
 * Stripe Payment Intents: Add support for error_on_requires_action option [tatsianaclifton] #3846
 * Add 3DS 2.0 values to paypal [nebdil] #3285
+* Redsys: Update Mpi Fields [tatsianaclifton] #3855
 
 == Version 1.117.0 (November 13th)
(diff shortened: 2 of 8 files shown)


lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -531,7 +533,7 @@ def build_merchant_mpi_external(xml, data)
         ds_merchant_mpi_external[:dsTransID] = data[:dsTransID] if data[:dsTransID]
         ds_merchant_mpi_external[:authenticacionValue] = data[:authenticacionValue] if data[:authenticacionValue]
         ds_merchant_mpi_external[:protocolVersion] = data[:protocolVersion] if data[:protocolVersion]
-
+        ds_merchant_mpi_external[:Eci] = data[:eci] if data[:eci]
```
Carrigan, line 536: This seems to have a duplicate line a few lines up; is this intentional? 

https://github.com/activemerchant/active_merchant/pull/3855/files#diff-c1465fbdcc1b656226ad4ffdda93044a9b05262de9adce1f7a16d70f1fbdc435R530

Looks like the keys have different casing.

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -333,19 +333,21 @@ def add_external_mpi_fields(data, options)
...
-
           data[:authenticacionMethod] = options[:authentication_method] if options[:authentication_method]
           data[:authenticacionType] = options[:authentication_type] if options[:authentication_type]
           data[:authenticacionFlow] = options[:authentication_flow] if options[:authentication_flow]
         elsif options[:three_d_secure][:version] == THREE_DS_V1
           data[:txid] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
           data[:cavv] = options[:three_d_secure][:cavv] if options[:three_d_secure][:cavv]
-          data[:eci] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
         end
+        data[:eci] = options[:three_d_secure][:eci] if options[:three_d_secure][:eci]
```
Carrigan, line 350: The docs show `Eci` for V2 and `ECI` for V1; do you think its worth being extra cautious and following the casing that they list since the sandbox environment is not providing good validation?

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -333,19 +333,21 @@ def add_external_mpi_fields(data, options)
         return unless options[:three_d_secure]
 
         if options[:three_d_secure][:version] == THREE_DS_V2
-          data[:threeDSServerTransID] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
+          data[:threeDSServerTransID] = options[:three_d_secure][:three_ds_server_trans_id] if options[:three_d_secure][:three_ds_server_trans_id]
```
Carrigan, line 336: 👏 I am glad that this will be a new addition to the standard fields!

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -333,19 +333,21 @@ def add_external_mpi_fields(data, options)
...
+          unless data[:threeDSServerTransID]
```
Carrigan, line 337: 🤔 I'm not sure that we want to default this back to XID since that is specifically a field meant for 3DS V1. Is this primarily for backwards compatibility?

Carrigan (commented):
Couple of questions but the untangling of `three_ds_server_trans_id` and `xid` is looking very good.

lib/active_merchant/billing/gateways/redsys.rb
tatsianaclifton: That's correct the only difference is casting, but I added it intentionally since he documentation states ECI for v1 and Eci for v2.

lib/active_merchant/billing/gateways/redsys.rb
tatsianaclifton: Casing is set up as in documentation later in the build_merchant_mpi_external method.

lib/active_merchant/billing/gateways/redsys.rb
tatsianaclifton: I am not sure if it is needed. I just didn't want to remove it in case somebody already uses this gateway and passes xid. Do you think it is safe to remove xid from 3ds V2?

curiousepic: Agree with Tatsiana, there is no reason to not leave it for backward compatibility.

lib/active_merchant/billing/gateways/redsys.rb
Carrigan: Ahh I see! This sets the data and the other function displays it. Thanks.

lib/active_merchant/billing/gateways/redsys.rb
curiousepic: I might suggest this more explicit structure for reasonability, since the situations are mutually exclusive:
```
if options[:three_d_secure][:three_ds_server_trans_id]
  data[:threeDSServerTransID] = options[:three_d_secure][:three_ds_server_trans_id]
else
   data[:threeDSServerTransID] = options[:three_d_secure][:xid] if options[:three_d_secure][:xid]
end
```

lib/active_merchant/billing/gateways/redsys.rb
Carrigan: If `data[:eci]` is present here it will be added twice under both casing types, which isn't ideal. What do you think about setting `eci_v1` or `eci_v2` in `add_external_mpi_fields` and then setting the proper casing here accordingly instead?

lib/active_merchant/billing/gateways/redsys.rb
curiousepic: Alternately, I think it could be safely condensed into
`data[:threeDSServerTransID] = options.dig(:three_d_secure, :three_ds_server_trans_id) || options.dig(:three_d_secure, :xid)`

tatsianaclifton: I looked more into this. I'll remove assignment of xid to threeDSServerTransID because it doesn't look correct. Firstly, as you mentioned xid is for version 1. Secondly, xid and threeDSServerTransID are different identifiers. xid is for Directory Server, threeDSServerTransID is for 3ds Server.

curiousepic: Void the comments above, I was thinking this was for the `dsTransID` field.

tatsianaclifton: Thanks for suggestion anyway. It will be useful for future because I wasn't sure about more concise way I can do it in Ruby.

lib/active_merchant/billing/gateways/redsys.rb
tatsianaclifton: Great suggestion. I'll make that change

lib/active_merchant/billing/gateways/redsys.rb
curiousepic: `dig` is somewhat new in Ruby. It's a more convenient way to safely reach into a nested hash, so it can replace instances of "options[:field] if options[:field]" etc., but since it is new it is not yet used extensively in AM.

Carrigan (approved):
Looks great 🎉 Thanks for picking up the Rubocop errors as well.

Remote tests have 2 failures (same as master) for me but they were not affected by these changes. Test suite otherwise passes for me.

tatsianaclifton:
@Carrigan I also saw 2 remote test failures for remote_redsys_test.rb and then realized that there are separate remote test for sha256 remote_redsys_sha256_test.rb. So I thought I need just ensure remote_redsys_sha256_test tests pass and didn't leave note in commit message for remote_redsys_test.rb failures. I'll add this note back when merging into master. Thanks

PR #3865, opened 2021-01-12
Elavon: Fix issue with encoding data sent in the request

naashton:
Unicode characters greater than 1 byte appear to result in an increase
in length+1 for each character for each byte greater than 1.

`force_encoding('iso-8859-1')` on the individual piece of data and then
re-encoding the request data to `utf-8` seems to resolve the issue.

CE-1231

Unit: 41 tests, 226 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 35 tests, 165 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -380,6 +380,7 @@ module ActiveMerchant #:nodoc:
 
       def commit(request)
         request = "xmldata=#{request}".delete('&')
+
         response = parse(ssl_post(test? ? self.test_url : self.live_url, request, headers))
 
         Response.new(
@@ -415,6 +416,14 @@ module ActiveMerchant #:nodoc:
 
       def authorization_from(response)
         [response[:approval_code], response[:txn_id]].join(';')
+      end
+
+      def truncate(value, size)
+        return nil unless value
+
+        difference = value.force_encoding('iso-8859-1').length - value.length
+
+        return value.to_s[0, (size - difference)]
       end
     end
   end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
 * Orbital: Update ECP details to use payment source [jessiagee] #3881
 * Alelo: Add additional BIN ranges [meagabeth] #3882
 * HPS: Update Add support for general credit [naashton] #3885
+* Elavon: Fix issue with encoding data sent in the request [naashton] #3865
 
 == Version 1.118.0 (January 22nd, 2021)
 * Worldpay: Add support for challengeWindowSize [carrigan] #3823
--- a/test/remote/gateways/remote_elavon_test.rb
+++ b/test/remote/gateways/remote_elavon_test.rb
@@ -413,6 +413,16 @@ class RemoteElavonTest < Test::Unit::TestCase
     assert response.authorization
   end
 
+  def test_successful_purchase_with_truncated_data
+    credit_card = @credit_card
+    credit_card.first_name = 'Ricky ™ Martínez įncogníto'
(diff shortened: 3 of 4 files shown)


naashton:
Draft request for now because I'm not sure this is the right solution. The initial indicator that this is an encoding issue is this `í` being replaced with `\u00ED` in the transcript, although this could be a result of some bad decoding when the transcript is generated because the hex value is still showing up even with the successful test. 

That said, it does seem like a wonky encoding issue either with the Nokogiri::XML::Builder or the HTTP::Net framework. Leaning toward the HTTP request because everything looks as expected after the XML is built (unicode characters appear normal).

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -379,7 +379,9 @@ def build_xml_request
       end
 
       def commit(request)
-        request = "xmldata=#{request}".delete('&')
+        request = request.force_encoding('iso-8859-1')
+        request = "xmldata=#{request.force_encoding('utf-8')}".delete('&')
```
therufs, line 383: Not super relevant to your point, but we should be able to get rid of this `.delete('&')` if the encoding tweaks work.

therufs:
@naashton Would it help to write a unit test that confirms that the outgoing request fields are the correct length and don't have wonky character mangling?

naashton:
When running tests, I'm checking string lengths as the values get added to the xml and they are showing the correct length. Unless you're suggesting checking those values AFTER the XML is built, which I will try now. 😄

naashton:
Special characters are not properly truncated leading to the request
being rejected

`puts builder.to_xml` shows what looks to be expected values

```
<?xml version="1.0" encoding="UTF-8"?>
<txn>
  <ssl_first_name>Ricky ™ Martínez </ssl_first_name>
  <ssl_last_name>Lesly Andrea Mart™nez estrad</ssl_last_name>
  <ssl_avs_address>Bats &amp; Cats</ssl_avs_address>
</txn>
```

`puts Nokogiri::XML(request)`  in the `commit` method shows that the
values contain hex, which when not properly truncated, end up adding
additional bytes

```
<?xml version="1.0"?>
<txn>
  <ssl_first_name>Ricky &#x2122; Mart&#xED;nez </ssl_first_name>
  <ssl_last_name>Lesly Andrea Mart&#x2122;nez estrad</ssl_last_name>
  <ssl_avs_address>Bats &amp; Cats</ssl_avs_address>
</txn>
```

CE-1231

Unit: 41 tests, 226 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 35 tests, 165 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Delete nonexistent method call

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -379,7 +379,8 @@ def build_xml_request
       end
 
       def commit(request)
-        request = "xmldata=#{request}".delete('&')
+        request = "xmldata=#{request}".scrub.delete('&')
```
therufs, line 382: errrr sorry if dumb q but why are we calling `scrub` here when the new method is `truncate`?

therufs: Update: I figured out what was going on with `truncate` but am still not sure why we're `scrub`bing :D

naashton: Well, I could have sworn there was a reason I was using `scrub`, but after running the tests without it, I can't remember when and why I added it. Just going to remove it 🤷

therufs (approved):
woot! lgtm && tests pass 🎉

PR #3867, opened 2021-01-13
add orbital echeck processing

ajawadmirza:
ECP Chase Orbital-Electronic Check Integration:

  1. ECP Standard Processing
  2. ECP Authorization Methods
  3. ECPSameDay Index Processing

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,6 +26,7 @@
 * Paymentez: Add 3DS MPI field support [carrigan] #3856
 * BlueSnap: Add support `fraud-session-id` field [meagabeth] #3863
 * BlueSnap: Update handling of `transaction-fraud-info` fields [meagabeth] #3866
+* Payeezy: Allow no stored credential transaction id [therufs] #3868
 
 == Version 1.117.0 (November 13th)
 * Checkout V2: Pass attempt_n3d along with 3ds enabled [naashton] #3805
--- a/lib/active_merchant/billing/gateways/payeezy.rb
+++ b/lib/active_merchant/billing/gateways/payeezy.rb
@@ -259,12 +259,12 @@ module ActiveMerchant
 
       def original_transaction_id(options)
         return options[:cardbrand_original_transaction_id] if options[:cardbrand_original_transaction_id]
-        return options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id]
+        return options[:stored_credential][:network_transaction_id] if options.dig(:stored_credential, :network_transaction_id)
       end
 
       def initiator(options)
         return options[:initiator] if options[:initiator]
-        return options[:stored_credential][:initiator].upcase if options[:stored_credential][:initiator]
+        return options[:stored_credential][:initiator].upcase if options.dig(:stored_credential, :initiator)
       end
 
       def sequence(initial_transaction)


jessiagee (approved):

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -478,6 +509,32 @@ def add_customer_address(xml, options)
...
+          add_creditcard(xml, payment_source, options[:currency])
+        end
+      end
+
+      # Adds Electronic Check attributes
+      def add_echeck(xml, check, options = {})
+        xml.tag! :CardBrand, 'EC'
+        xml.tag! :CurrencyCode, currency_code(options[:currency])
+        xml.tag! :CurrencyExponent, currency_exponents(options[:currency])
+        unless check.nil?
```
naashton, line 526: I'm not sure what the general preference is here, but I'm of the opinion that we don't raise exceptions on data we send in the request. Given that AM is an abstraction for the gateways, we should pass data through as it is or isn't received and let the gateway dictate what is and isn't valid.

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -605,6 +662,27 @@ def add_managed_billing(xml, options)
...
+      def validate_ecp_check_serial(serial_number, bin)
+        (bin.eql?('000001') && serial_number.length.eql?(9)) || (bin.eql?('000002') && serial_number.length.eql?(6))
+      end
+
+      # Adds ECP conditional attributes depending on other attribute values
+      def add_ecp_details(xml, parameters = {})
+        requires!(parameters, :check_serial_number) if parameters[:auth_method]&.eql?('A') || parameters[:auth_method]&.eql?('P')
+        xml.tag! :ECPActionCode, parameters[:action_code] if parameters[:action_code] && ECP_ACTION_CODES.include?(parameters[:action_code])
+        if parameters[:auth_method]&.eql?('A') || parameters[:auth_method]&.eql?('P')
+          raise 'Invalid Check Serial Number' unless validate_ecp_check_serial(parameters[:check_serial_number], bin)
```
naashton, line 675: Same as comment on line 527.

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -217,6 +220,104 @@ def test_successful_purchase_with_discover_network_tokenization_credit_card
     assert_false response.authorization.blank?
   end
 
+  # Successful purchase through Electronic Check
```
naashton, line 223: These function level comments aren't necessary since the code is generally self documenting.

lib/active_merchant/billing/gateways/orbital.rb
ajawadmirza: Instead of creating an invalid request where the mandatory parameters are missing and then 
waiting for the server to send an expected error, simple validation has been added to save the unnecessary request
response roundtrip before the error is returned back to the caller. Moreover, validate function exists within active merchant
for multiple payment source objects.

lib/active_merchant/billing/gateways/orbital.rb
ajawadmirza: Reason is similar to the previous comment.

test/remote/gateways/remote_orbital_test.rb
ajawadmirza: The idea was to provide a single line description of the tests in plain text. If required, these can be removed.

lib/active_merchant/billing/gateways/orbital.rb
naashton: It's not necessarily a bad idea to validate data before passing it through, but for payment methods, it can be particularly challenging. Especially since we deal with payment methods from many different countries with their own set of financial standards. 

The `validate` method for check payment method type actually runs the `routing_number` against an algorithm that is specific to US banks. As far as I can tell, `validate` is not used in AM outside of test methods. It is probably better to maintain consistency throughout AM and follow the implementation for adding payment method types.

It's also important to remember that AM is an open source project and taking the agnostic approach gives us the flexibility of implementing a gateway and using it for our specific needs while also allowing other users of this project to build out their own use cases.

test/remote/gateways/remote_orbital_test.rb
naashton: Yes, these comments are not needed.

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -478,6 +509,32 @@ def add_customer_address(xml, options)
...
+
+      # Adds Electronic Check attributes
+      def add_echeck(xml, check, options = {})
+        xml.tag! :CardBrand, 'EC'
+        xml.tag! :CurrencyCode, currency_code(options[:currency])
+        xml.tag! :CurrencyExponent, currency_exponents(options[:currency])
+        unless check.nil?
+          raise "Invalid Check Received. Error: #{check.validate}" unless check.validate.empty?
+
+          xml.tag! :BCRtNum, options[:routing_number]
```
naashton, line 529: `routing_number` is an attribute of the `check` payment method type, so we should expect `check.routing_number` to grab what we need here.

lib/active_merchant/billing/gateways/orbital.rb
naashton: We should remove the `raise` flag and use conditionals where appropriate, let the gateway process the transaction and handle responses accordingly.

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -605,6 +662,27 @@ def add_managed_billing(xml, options)
         end
       end
 
+      # Validates Electronic Check Serial number based on bin value provided
+      def validate_ecp_check_serial(serial_number, bin)
+        (bin.eql?('000001') && serial_number.length.eql?(9)) || (bin.eql?('000002') && serial_number.length.eql?(6))
```
naashton, line 667: I'm actually a little confused here. Are we only accepting payment methods with a bin that is one of these 2 values? @jessiagee?

jessiagee: @naashton From Erik: "Really only 000001.  I know we have the logic in the AM code for looking at those BIN values based on MID length but from a Spreedly perspective we only support the Stratus platform which is 000001."

jessiagee: Also from the spec doc: "The Orbital Gateway supports Electronic Check Processing (ECP) for eligible merchants. This method of payment is only available to Stratus platform merchants (BIN 000001)." So I suppose we could actually just check if BIN == 000001 and if not than we can't do ECP.

lib/active_merchant/billing/gateways/orbital.rb
ajawadmirza: Point taken. We will remove this validation.

lib/active_merchant/billing/gateways/orbital.rb
ajawadmirza: Acknowledged.

test/remote/gateways/remote_orbital_test.rb
ajawadmirza: Acknowledged. We will remove the comments.

lib/active_merchant/billing/gateways/orbital.rb
ajawadmirza: Since the code on line 675 will be removed as per previous comment hence this function is not required anymore so it will be removed.

lib/active_merchant/billing/gateways/orbital.rb
ajawadmirza: The check object is specific to the US banking system, i.e. it only accepts a nine digit routing number. Some countries such as Canada use an eight digit routing number.

naashton: Hmm, I'm not entirely sure what constraints we have internally for ACH payment methods. In Spreedly Core, it looks like we treat all routing numbers as IBAN (which is the largest character count), so I think we could expect all routing numbers to come from the `payment_method` via `routing_number` attribute. Maybe @jessiagee can confirm for me.

naashton: Erik confirmed that we will pass the `routing_number` with the `payment_method`, so I think we only need to be concerned about grabbing `check.routing_number` here.

naashton:
I got one linter issue: 
```
lib/active_merchant/billing/gateways/orbital.rb:668:9: C: Style/IfUnlessModifier: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||. (https://github.com/rubocop-hq/ruby-style-guide#if-as-a-modifier)
        if parameters[:auth_method]&.eql?('A') || parameters[:auth_method]&.eql?('P')
```

Condensing that into `xml.tag! :ECPCheckSerialNumber, parameters[:check_serial_number]` should resolve that.

Otherwise, everything else looks good.

PR #3868, opened 2021-01-15
Payeezy: Allow no stored credential initiator

therufs:


cdmackeyfree (approved):
LGTM and tests pass! 🚢

PR #3872, opened 2021-01-20
[Firstdatae427] Fixes some apple pay transaction issues

pi3r:
Firsdata/Payeezy contacted Shopify to let us know that they need
the ECI for applepay/discover to be set to 5 and not 4.
See: https://support.payeezy.com/hc/en-us/articles/203730589-Ecommerce-Flag-Values

We also need to pass a WalletProviderID of 4 when the card source
is apple pay.
See: https://support.payeezy.com/hc/en-us/articles/206601408-First-Data-Payeezy-Gateway-Web-Service-API-Reference-Guide#3.9


bbraschi (commented):
👍

pi3r:
> Do we understand what changed between needing to set a 04 vs 05?

Those are the changes they asked us to do. Looking at their docs it seems to make sense.

PR #3885, opened 2021-02-03
HPS: Add support for general credit

naashton:
General credit transactions utilize the same endpoint as refund, but
take a `payment_method` instead and do not require a `transaction_id`
from a preceding transaction.

Tests that fail are due to some incorrect configurations for account.

CE-1305

Unit: 57 tests, 278 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 55 tests, 146 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.3636% passed

### DIFF
--- a/test/remote/gateways/remote_hps_test.rb
+++ b/test/remote/gateways/remote_hps_test.rb
@@ -167,6 +167,17 @@ class RemoteHpsTest < Test::Unit::TestCase
     assert_failure response
   end
 
+  def test_successful_credit
+    credit = @gateway.credit(@amount, @credit_card, @options)
+    assert_success credit
+    assert_equal 'Success', credit.params['GatewayRspMsg']
+  end
+
+  def test_failed_credit
+    credit = @gateway.credit(nil, @credit_card)
+    assert_failure credit
+  end
+
   def test_successful_void
     auth = @gateway.authorize(@amount, @credit_card, @options)
     assert_success auth
--- a/test/unit/gateways/hps_test.rb
+++ b/test/unit/gateways/hps_test.rb
@@ -142,6 +142,20 @@ class HpsTest < Test::Unit::TestCase
     assert_failure refund
   end
 
+  def test_successful_credit
+    @gateway.expects(:ssl_post).returns(successful_refund_response)
+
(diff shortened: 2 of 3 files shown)


naashton:
The tests that fail:
* `test_successful_check_purchase(RemoteHpsTest):`
* `test_successful_check_void(RemoteHpsTest)`

test/unit/gateways/hps_test.rb
```diff
@@ -142,6 +142,20 @@ def test_failed_refund
     assert_failure refund
   end
 
+  def test_successful_credit
```
meagabeth, line 145: The majority of the other tests utilize `assert_instance_of` before the `assert_failure` or `assert_success` lines. Would adding that line benefit these new tests?

test/remote/gateways/remote_hps_test.rb
```diff
@@ -167,6 +167,16 @@ def test_failed_refund
     assert_failure response
   end
 
+  def test_successful_credit
```
meagabeth, line 170: Similar observation as for the unit tests; would adding in `assert_equal` or `assert_match` lines benefit these tests?

test/unit/gateways/hps_test.rb
naashton: I saw that and I didn't see any reason for it. It's just asserting that the class is a `Response` object, which is going to always be true.

test/remote/gateways/remote_hps_test.rb
naashton: The response from HPS doesn't have any valuable information besides the response code and message, which end up in the Response object. That said, `assert_success` only checks that what the `Response.code` is and whether or not it's successful. It's conceivable that the `message` could differ from the `code` we get, so I added an `assert_equal` to verify the the `message` is what we expect.

meagabeth (approved):
Same 2 tests fail for me; I say go forth! 🚀

naashton:
Merged. Closing manually since the CHANGELOG was not updated in the same commit.

PR #3890, opened 2021-02-10
Adding retry_logic to params

jessiagee:
rake test:local                                             

Finished in 7.84327 seconds.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
4634 tests, 73065 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
590.83 tests/s, 9315.63 assertions/s

Loaded suite test/unit/gateways/orbital_test
Started
..........................................................................................................
Finished in 0.36499 seconds.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
106 tests, 632 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
290.42 tests/s, 1731.55 assertions/s

Loaded suite test/remote/gateways/remote_orbital_test
Started
.................FF.....................................
Finished in 31.880217 seconds.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
56 tests, 266 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.4286% passed
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1.76 tests/s, 8.34 assertions/s

Two failed remote tests are known issues with 3DS, not impacting this PR.

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -232,7 +232,7 @@ module ActiveMerchant #:nodoc:
             add_managed_billing(xml, options)
           end
         end
-        commit(order, :authorize, options[:trace_number])
+        commit(order, :authorize, options[:retry_logic], options[:trace_number])
       end
 
       def verify(creditcard, options = {})
@@ -252,12 +252,12 @@ module ActiveMerchant #:nodoc:
             add_managed_billing(xml, options)
           end
         end
-        commit(order, :purchase, options[:trace_number])
+        commit(order, :purchase, options[:retry_logic], options[:trace_number])
       end
 
       # MFC - Mark For Capture
       def capture(money, authorization, options = {})
-        commit(build_mark_for_capture_xml(money, authorization, options), :capture)
+        commit(build_mark_for_capture_xml(money, authorization, options), :capture, options[:retry_logic], options[:trace_number])
       end
 
       # R – Refund request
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -753,9 +753,9 @@ def recurring_parse_element(response, node)
         end
       end
 
-      def commit(order, message_type, trace_number = nil)
+      def commit(order, message_type, retry_logic = nil, trace_number = nil)
         headers = POST_HEADERS.merge('Content-length' => order.size.to_s)
-        if @options[:retry_logic] && trace_number
+        if @options[:retry_logic] || retry_logic && trace_number
```
therufs, line 758: it's late, but i _think_ that `&&` takes precedence over `||`. So
`if @options[:retry_logic] || retry_logic && trace_number` 
is not the same as 
`if (@options[:retry_logic] || retry_logic) && trace_number` (which would preserve the current behavior):

```
irb(main):016:0> true || false && nil
=> true
irb(main):017:0> (true || false) && nil
=> nil
```

So if we still only want to set the headers if there's a trace number in either case, we will want some parens here. (It looks like we don't currently have a test case that checks that we don't send the headers if there's no trace number.)

jessiagee: Oh good catch, I meant to wrap them and forgot! And yeah I'll add in a test for "if we don't send retry logic than ignore trace number" and also "if there's no trace number than ignore retry logic check"

therufs (approved):
LGTM && tests pass. Thanks for adding the extra test for trace number :D

PR #3891, opened 2021-02-12
Elavon: Remove ampersand char from fields

naashton:
The `&` char converts to the longer html encoding AFTER truncated,
resulting in the request being rejected for strings that are too long.
Easiest solution is to just remove the `&` from any value added to the
xml.

CE-1231

Unit: 42 tests, 229 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 35 tests, 165 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


therufs (approved):
lgtm && tests pass 🎉

PR #3897, opened 2021-02-17
Mercado Pago: Give ability to pass capture option in authorize txn

naashton:
Users can pass `true` as the `capture` value to execute a force capture
on an authorize transaction, which changes the initial default value for
this option from `false`. Mercado Pago defaults the `capture` value to
`true` unless it is explicitly set to `false`.

CE-1332

Unit: 39 tests, 181 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 33 tests, 96 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
90.9091% passed


jessiagee (approved):
Looks good to me, only 3 test failures that we previously knew about

PR #3898, opened 2021-02-17
New Spreedly gateway for PaywayDotCom

DanAtPayway:
Hello,

I added this pull for the New Spreedly gateway for PayWayDotCom.

Let me know if you need our PaywayWS Document.

The reponse.message will contain our 4 digit paywayCode concatenated with a '-' (dash) character followed by a paywayMessage (text description that is could change in the future).  The 4 digit paywayCode will not change.  Both are described in our PaywayWS Credit Card Integration Guide. 

Dan

### DIFF
--- a/lib/active_merchant/billing/gateways/payway_dot_com.rb
+++ b/lib/active_merchant/billing/gateways/payway_dot_com.rb
@@ -1,0 +1,316 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PaywayDotComGateway < Gateway
+      self.test_url = 'https://devedgilpayway.net/PaywayWS/Payment/'
+      self.live_url = 'https://edgilpayway.com/PaywayWS/Payment/'
+
+      self.supported_countries = ['US', 'CA']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = [:visa, :master, :american_express, :discover]
+
+      self.money_format = :cents
+
+      self.homepage_url = 'http://www.payway.com'
+      self.display_name = 'Payway Gateway'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        '5012' => STANDARD_ERROR_CODE[:card_declined],
+        '5035' => STANDARD_ERROR_CODE[:invalid_number],
+        '5037' => STANDARD_ERROR_CODE[:invalid_expiry_date], # The expiration date is invalid.
+        '5045' => STANDARD_ERROR_CODE[:incorrect_zip] # The zip code or postal code is invalid.
+      }
+
+      # Payway to standard AVSResult codes.
+      AVS_MAPPING = {
+        'N1'  => 'I', #  No address given with order
+        'N2'  => 'I', #  Bill-to address did not pass
+        '““'  => 'R', #  AVS not performed (Blanks returned)
+        'IU'  => 'G', #  AVS not performed by Issuer
+        'ID'  => 'S', #  Issuer does not participate in AVS
+        'IE'  => 'E', #  Edit Error – AVS data is invalid
+        'IS'  => 'R', #  System unavailable or time-out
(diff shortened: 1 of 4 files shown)


tatsianaclifton:
@DanAtPayway I noticed that implementation is for supporting other payment types besides credit cards. However, all tests are written just for credit cards. Could you please add unit and remote tests for all supported payment types or change implementation to support just credit cards? I will continue with code review once this is addressed. Thank you

DanAtPayway:
Tatsiana,

I checked in the updated code,

Dan

lib/active_merchant/billing/gateways/payway_dot_com.rb
```diff
@@ -0,0 +1,268 @@
...
+
+      def error_code_from(response)
+        if success_from(response)
+          ''
+        else
+          error = !STANDARD_ERROR_CODE_MAPPING[response['paywayCode']].nil? ?
+          STANDARD_ERROR_CODE_MAPPING[response['paywayCode']] :
+          STANDARD_ERROR_CODE[:processing_error]
+        end
+      end
```
tatsianaclifton, line 237: Is there a reason to set error_code to empty string versus just keep it nil? Anyway I think the method can be simplified to something like 
```suggestion
      def error_code_from(response)
        return '' if success_from(response)
        
        error = !STANDARD_ERROR_CODE_MAPPING[response['paywayCode']].nil? ? 
        STANDARD_ERROR_CODE_MAPPING[response['paywayCode']] : 
        STANDARD_ERROR_CODE[:processing_error]
        return error
      end
```

Or can just do `return if success_from(response)` if there is no specific reason to set error to empty string

lib/active_merchant/billing/gateways/payway_dot_com.rb
```diff
@@ -0,0 +1,268 @@
...
+      def message_from(success, response)
+        if !response['paywayCode'].nil?
+          if success
+            return response['paywayCode'] + "-" + "success"
+          else
+            return response['paywayCode'] + "-" + response['paywayMessage']
+          end
+        end
+        ""
+      end
```
tatsianaclifton, line 248: Maybe it can be simplified to something like 
```suggestion
      def message_from(success, response)
        return '' if response['paywayCode'].nil?

        return response['paywayCode'] + '-' + 'success' if success

        response['paywayCode'] + '-' + response['paywayMessage']
      end
```
Or can just do `return if response['paywayCode'].nil?` if there is no specific reason to set message to an empty string.

lib/active_merchant/billing/gateways/payway_dot_com.rb
```diff
@@ -0,0 +1,268 @@
...
+
+      def authorization_from(response)
+        if success_from(response)
+          if !response['cardTransaction'].nil?
+            return response['cardTransaction']['name'] if response['cardTransaction']['name']
+          end
+        end
+        ""
+      end
```
tatsianaclifton, line 257: Maybe something simpler will also work
```suggestion
      def authorization_from(response)
        return '' unless success_from(response) && response['cardTransaction'].nil?

        response['cardTransaction']['name']
      end
```

Can be just `return unless success_from(response) && response['cardTransaction'].nil?` if no specific reason to set authorization to an empty string.

tatsianaclifton (changes requested):
Hello,

Thanks for working on that implementation. Overall, it looks very good so far. Great job! Just couple questions and small suggestions. Once fixed I'll look over one more time to make sure we don't have any more questions before merging.

DanAtPayway:
Tatsiana,

I have completed all the changes you requested.

I changed idSource to sourceId, removed unused code and comments and updated code per suggestions.

The Authorization Hold PaywayWS request is "authorization", not "authorizationSale" as you saw in the document/guide.  We have updated our document to reflect this.

Dan

tatsianaclifton (approved):
Looks good overall. Approved, but I will make some styling changes and merge after that.

PR #3900, opened 2021-02-19
Payeezy: Update `error_code_from` method

meagabeth:
CE-1303

Update `error_code_from` method to send `nil` if `bank_resp_code` is present and equal to '100', map the value of `bank_resp_code` to `error_code` if it is not equal to '100'

Unit:
36 tests, 172 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
36 tests, 141 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


cdmackeyfree (approved):
Tests pass and it LGTM!

PR #3902, opened 2021-02-22
WorldPay: remove some defaults in billing address

Carrigan:
## Why?

ECS-14244

ActiveMerchant sends default values to WorldPay for the billing address when no value is given. This can be problematic when the gateway needs to do address verification.

## What Changed?

This commit removes the defaults for billing address state, address1, and city. The WorldPay docs specify that if either field of address1 and city is given, the other must also be, so defaulting was added for these fields when only one or the other is given.

## Test Suite

Worldpay Remote Tests Before:
60 tests, 249 assertions, 5 failures, 0 errors, 0 pendings,
  0 omissions, 0 notifications

Worldpay Remote Tests After:
61 tests, 250 assertions, 5 failures, 0 errors, 0 pendings,
  0 omissions, 0 notifications

Unit Tests:
4652 tests, 73157 assertions, 0 failures, 0 errors, 0 pendings,
  0 omissions, 0 notifications
693 files inspected, no offenses detected

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,7 @@
 * Worldpay: Add support for `statementNarrative` field [meagabeth] #3901
 * Mercado Pago: Give ability to pass capture option in authorize txn field [naashton] #3897
 * Orbital: Ensure correct fields sent in refund [jessiagee] #3903
+* WorldPay: remove some defaults in billing address [carrigan] #3902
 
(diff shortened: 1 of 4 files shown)


curiousepic (approved):
👍  Looks good.

Carrigan:
Thanks @curiousepic!

PR #3906, opened 2021-02-25
vPOS: add gateway

therufs:
Pending: better remote tests, any unit tests, ~implementation of `scrub`~.

Currently, we don't get a successful transaction response, even using their specified test cards, which makes some response attributes challenging to establish. 😅

### DIFF
--- a/test/remote/gateways/remote_vpos_test.rb
+++ b/test/remote/gateways/remote_vpos_test.rb
@@ -1,0 +1,80 @@
+require 'test_helper'
+
+class RemoteVposTest < Test::Unit::TestCase
+  def setup
+    @gateway = VposGateway.new(fixtures(:vpos))
+
+    @amount = 100000
+    @credit_card = credit_card('5418630110000014', month: 8, year: 2021, verification_value: '258')
+    @declined_card = credit_card('4000300011112220')
+    @options = {
+      billing_address: address,
+      description: 'Store Purchase'
+    }
+  end
+
+  def test_successful_purchase
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal 'Transaccion aprobada', response.message
+  end
+
+  def test_failed_purchase
+    response = @gateway.purchase(@amount, @declined_card, @options)
+    assert_failure response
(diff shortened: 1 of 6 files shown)


therufs:
Unit tests: 
5 tests, 15 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests:
7 tests, 21 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed 

Implementation of `credit` forthcoming. 

PS-165

jessiagee (changes requested):
Seeing some rubocop errors on my end after pulling down!

jessiagee (approved):
Looks good and the tests pass! Aside from squashing I think this is g2g from my end!

test/remote/gateways/remote_vpos_test.rb
```diff
@@ -0,0 +1,92 @@
+require 'test_helper'
+
+class RemoteVposTest < Test::Unit::TestCase
+  def setup
+    @gateway = VposGateway.new(fixtures(:vpos))
+
+    @amount = 100000
+    @credit_card = credit_card('5418630110000014', month: 8, year: 2021, verification_value: '258')
```
meagabeth, line 8: Did these test card details come from vPOS specifically? Having the hard coded expiration will cause a hiccup when it expires, but it might just be the way it is if they gave these deets.

test/remote/gateways/remote_vpos_test.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    purchase = @gateway.purchase(@amount, @credit_card, options)
+    assert_success purchase
+
+    assert void = @gateway.void(nil, options)
+    assert_success void
+    assert_equal 'RollbackSuccessful', void.message
+  end
+
+  def test_duplicate_void_fails
```
meagabeth, line 52: Nice! That's a good idea to test for ⭐

meagabeth (approved):
Wow! This is all thoughtfully written code and covers many major bases. Tests pass, go forth!

test/remote/gateways/remote_vpos_test.rb
therufs: Yes, vPOS specifically requires this expiration date. I'll set a reminder to check the test suite in August, though :D

PR #3909, opened 2021-03-01
Stripe PI: Send Validate on Payment Method Attach

tatsianaclifton:
Currently we are passing the validate field to Stripe PI’s /customers endpoint rather than the /payment_methods endpoint. This field is essentially ignored by the /customers endpoint, and all payment methods are being validated, even if the request was sent with ”validate”: false. This commit includes updates to instead pass "validate": false during the attachment step (POST /v1/payment_methods/:id/attach), rather than when the customer is being created. This will guarantee that Stripe does not validate cards when users do not want to do so.

ECS-1681

Unit:
23 tests, 141 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
54 tests, 250 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Local:
4655 tests, 73169 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -183,7 +183,6 @@ module ActiveMerchant #:nodoc:
           if options[:customer]
             customer_id = options[:customer]
           else
-            post[:validate] = options[:validate] unless options[:validate].nil?
             post[:description] = options[:description] if options[:description]
             post[:email] = options[:email] if options[:email]
             options = format_idempotency_key(options, 'customer')
@@ -191,7 +190,9 @@ module ActiveMerchant #:nodoc:
             customer_id = customer.params['id']
           end
           options = format_idempotency_key(options, 'attach')
-          commit(:post, "payment_methods/#{params[:payment_method]}/attach", { customer: customer_id }, options)
+          attach_parameters = { customer: customer_id }
+          attach_parameters[:validate] = options[:validate] unless options[:validate].nil?
+          commit(:post, "payment_methods/#{params[:payment_method]}/attach", attach_parameters, options)
         else
           super(payment_method, options)
         end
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -183,15 +183,15 @@ def store(payment_method, options = {})
...
             customer_id = options[:customer]
           else
-            post[:validate] = options[:validate] unless options[:validate].nil?
             post[:description] = options[:description] if options[:description]
             post[:email] = options[:email] if options[:email]
             options = format_idempotency_key(options, 'customer')
             customer = commit(:post, 'customers', post, options)
             customer_id = customer.params['id']
           end
+          validate = !options[:validate].nil? ? options[:validate] : false
```
curiousepic, line 192: Nitpick, but Ruby norms prefer not using '!' when possible, and here we can use
`options[:validate].present? ? options[:validate] : false`

curiousepic: However, we might want to go with some extra resilience by enforcing `true`/`false`, via something like:
`validate = (options[:validate] == "true").to_s`

curiousepic: Also want to confirm that not validating should be the default case? That seems to be true of the initial implementation but it's something we should be sure about.

tatsianaclifton: The ticket description sounded to me as we want to set to 'false' by default but  now you mentioned I'm not sure and asked there to clarify.

tatsianaclifton: @curiousepic  I think it makes sense to pass `validate` parameter to /attach only if this option was requested by user. Otherwise, we should let Stripe to perform their default actions (currently they do validation by default). So, I made a change to pass a `validate` parameter only if it was provided in `options`.

curiousepic (approved):
That makes sense and this looks good. 💥

PR #3910, opened 2021-03-03
Adyen: Update handling of authorization returned from gateway

meagabeth:
Add test to confirm successful refund possible using `pspReference` from an `authorize` transaction instead of a `capture` transaction. Update test cards that had expired.

CE-803

Local:
4650 tests, 73149 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
71 tests, 372 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
94 tests, 361 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -6,12 +6,12 @@ class RemoteAdyenTest < Test::Unit::TestCase
 
     @amount = 100
 
-    @credit_card = credit_card('3700 0000 0000 002',
+    @credit_card = credit_card('4111111111111111',
       month: 3,
       year: 2030,
       first_name: 'John',
       last_name: 'Smith',
-      verification_value: '7373',
+      verification_value: '737',
       brand: 'visa')
 
     @avs_credit_card = credit_card('4400000000000008',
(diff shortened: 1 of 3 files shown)


meagabeth:
Adyen issues a `pspReference` when they receive an `authorize` or `capture` request. When issuing a refund, Adyen needs the original `pspReference` to be sent in with the request. Sometimes the `pspReference` string begins with `#` and other times it does not, so I modified the method that handles that and is being called in the `refund`.

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -6,7 +6,7 @@ def setup
 
     @amount = 100
 
-    @credit_card = credit_card('3700 0000 0000 002',
```
naashton, line 9: I'm getting a lot of test failures because of the different card number. Might need to change it back.

naashton (changes requested):

test/remote/gateways/remote_adyen_test.rb
naashton: I think if you want to change it to the Visa card, you'll have to also change the CVV, but I'm not sure if there was a particular reason for using AMEX as the default test credit_card or not.

meagabeth: Oops, this was a merge conflict and I didn't choose the right one. Will fix this now - thanks!

naashton (approved):
Good to go!

Issue #3895, opened 2021-02-15
Clean up gateway generator templates to be Rubocop compliant


therufs:
The gateway template files in `/generators/gateway/templates` are used to generate skeleton gateway implementations, but the generated skeletons end up with a _ton_ of linting errors. Rubocop does not seem to be smart enough to operate on the templates, so these may need to be fixed manually.

PR #3912, opened 2021-03-04
Update gateway skeleton templates

therufs:
The existing gateway skeleton templates generated skeletons that are not
Rubocop-compliant as of 19386a2d7c0531ea81eb8a4ef1abb7981cec0c7f.

Compare 203149a19aabe20848b2e716677b1dbbf1b40025 for the updates required by
a generated skeleton.

See also #3895.


naashton (approved):
Looks good!

PR #3914, opened 2021-03-08
Payment Express: standardize verify operation

therufs:
The original signature of `verify` for Payment Express did not match `verify` implementations in other gateway adapters.


naashton (approved):
Looks good! 🚀

PR #3918, opened 2021-03-15
Worldpay: Only override cardholdername for 3ds tests

curiousepic:
After confirming with Worldpay, the cardholder name should only be "3D"
for tests. This adds a check for whether it's a test request, but
retains the override logic for user sandbox testing.

Remote (2 unrelates failures also on master):
61 tests, 254 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.7213% passed

Unit:
78 tests, 504 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -430,7 +430,7 @@ module ActiveMerchant #:nodoc:
           )
         end
 
-        card_holder_name = options[:execute_threed] && !options[:three_ds_version]&.start_with?('2') ? '3D' : payment_method.name
+        card_holder_name = test? && options[:execute_threed] && !options[:three_ds_version]&.start_with?('2') ? '3D' : payment_method.name
         xml.cardHolderName card_holder_name
         xml.cvc payment_method.verification_value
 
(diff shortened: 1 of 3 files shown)


tatsianaclifton:
It looks good to me, but maybe it would be good to add a test to check if name is set correctly when not a test request. What do you think?

curiousepic:
There's the catch 22; there's no easy way to run a test outside of test mode, in the AM test framework. But the change should not be disruptive, so we should be able to safely confirm after deployment.

curiousepic:
There's the catch 22; there's no easy way to run a test outside of test mode, in the AM test framework. However, I can run it without the `test?` check to confirm - this also means the cardholder name in the remote tests should be changed from `3D`...

tatsianaclifton:
> There's the catch 22; there's no easy way to run a test outside of test mode, in the AM test framework. However, I can run it without the `test?` check to confirm - this also means the cardholder name in the remote tests should be changed from `3D`...

I meant unit tests. I though it would be possible to have a test when the gateway is created with `test: false` and then verify that cardHolderName in the request is not set to '3D'

curiousepic:
@tatsianaclifton Unit tests also assume test mode, but in fact there is an instance of forcing the mode to production elsewhere in the tests, so I added one for this. Thanks for the recommendation!

tatsianaclifton (approved):
LGTM

curiousepic:
Closed with https://github.com/activemerchant/active_merchant/commit/3633283fa2606321f87f94bdecacc2307b91ed16

PR #3923, opened 2021-03-18
Banco Sabadell: Ensure sca_exemption field is used

esmitperez:
The use of `sca_exemption` requires passing
`DS_MERCHANT_DIRECTPAYMENT=true` for those terminals where 3DS2 is
enabled.

This commit introduces a new option
`sca_exemption_direct_payment_enabled` that can be used to pass this
value, when needed

ECS-1747

Local:
45 tests, 183 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
30 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
4681 tests, 73300 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Note:

`test/remote/gateways/remote_redsys_test.rb`:
- failing same as current HEAD, 3633283fa2606321f87f94bdecacc2307b91ed16

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys.rb
+++ b/lib/active_merchant/billing/gateways/redsys.rb
@@ -208,6 +208,7 @@ module ActiveMerchant #:nodoc:
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
         data[:sca_exemption] = options[:sca_exemption]
+        data[:sca_exemption_direct_payment_enabled] = options[:sca_exemption_direct_payment_enabled]
 
         commit data, options
       end
@@ -226,6 +227,7 @@ module ActiveMerchant #:nodoc:
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
         data[:sca_exemption] = options[:sca_exemption]
+        data[:sca_exemption_direct_payment_enabled] = options[:sca_exemption_direct_payment_enabled]
 
         commit data, options
       end
@@ -397,7 +399,8 @@ module ActiveMerchant #:nodoc:
           REQUEST
           parse(ssl_post(threeds_url, request, headers(action)), action)
         else
-          parse(ssl_post(url, "entrada=#{CGI.escape(xml_request_from(data, options))}", headers), action)
+          xmlreq = xml_request_from(data, options)
+          parse(ssl_post(url, "entrada=#{CGI.escape(xmlreq)}", headers), action)
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -226,6 +227,7 @@ def authorize(money, payment, options = {})
         data[:description] = options[:description]
         data[:store_in_vault] = options[:store]
         data[:sca_exemption] = options[:sca_exemption]
```
Carrigan, line 229: Does this also need the new `Y` logic?

test/unit/gateways/redsys_sha256_test.rb
```diff
@@ -123,14 +124,49 @@ def test_successful_authorize_with_3ds
...
+  def test_successful_purchase_with_3ds2
+    @gateway.expects(:ssl_post).returns(successful_authorize_with_3ds_response)
+    response = @gateway.purchase(100, @threeds2_credit_card, { execute_threed: true, order_id: '156270437866' })
+    assert response.test?
+    assert response.params['ds_emv3ds']
+    assert_equal response.message, 'CardConfiguration'
+    assert_equal response.authorization, '156270437866||'
+  end
+
+  def test_successful_purchase_with_3ds2_and_mit_exemption
```
Carrigan, line 145: Great new test coverage on the exceptions! Do you think its work adding a test here as well for `DS_MERCHANT_DIRECTPAYMENT`?

test/remote/gateways/remote_redsys_sha256_test.rb
```diff
@@ -133,6 +136,111 @@ def test_successful_3ds_authorize_with_exemption
     assert_equal 'CardConfiguration', response.message
   end
 
+  def test_successful_3ds_purchase_with_exemption
```
Carrigan, line 139: Great remote test coverage 🎉 remote suite passes for me with 2 exceptions (same as master)

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -485,7 +488,14 @@ def build_merchant_data(xml, data, options = {})
...
           xml.DS_MERCHANT_MERCHANTSIGNATURE  build_signature(data) unless sha256_authentication?
-          xml.DS_MERCHANT_EXCEP_SCA          data[:sca_exemption] if data[:sca_exemption]
+
+          action = determine_3ds_action(data[:threeds]) if data[:threeds]
+          if action == 'iniciaPeticion' && data[:sca_exemption]
+            xml.DS_MERCHANT_EXCEP_SCA 'Y'
+          elsif data[:sca_exemption]
+            xml.DS_MERCHANT_EXCEP_SCA data[:sca_exemption]
+          end
+          xml.DS_MERCHANT_DIRECTPAYMENT data[:sca_exemption_direct_payment_enabled] if data[:sca_exemption_direct_payment_enabled]
```
tatsianaclifton, line 498: It seems that it will also be added for 'iniciaPeticion', but I think it was only needed for 'trataPeticion'.

esmitperez: Good catch! will update

test/unit/gateways/redsys_sha256_test.rb
esmitperez: Yes, but a correction Tatsiana and I made makes it available only on `trataPeticion` and I'm trying to figure out the best way to trigger/capture that.

lib/active_merchant/billing/gateways/redsys.rb
Carrigan: Never mind - I realize that this is just setting up the data and not actually appending to the request.

tatsianaclifton:
LGTM and tests pass except those that are failing on master

tatsianaclifton (approved):

PR #3929, opened 2021-03-23
PaywayDotCom: Improve Scrubbing

tatsianaclifton:
The scrub method didn't scrub responses properly as the response has some extra spaces. Additionally, failed transactions include the account number in message field.Even though it is the invalid account number it must be scrubbed as we don't know what part of it is invalid, maybe just 1 didgit. This commit includes changes to fix described issues. Remote tests and unit tests were expanded also.
ECS-1713

Unit:
15 tests, 62 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
16 tests, 43 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payway_dot_com.rb
+++ b/lib/active_merchant/billing/gateways/payway_dot_com.rb
@@ -56,7 +56,10 @@ module ActiveMerchant #:nodoc:
       SCRUB_PATTERNS = [
         %r(("password\\?":\\?")[^\\]+),
         %r(("fsv\\?":\\?")\d+),
-        %r(("accountNumber\\?":\\?")\d+)
+        %r(("fsv\\?": \\?")\d+),
+        %r(("accountNumber\\?":\\?")\d+),
+        %r(("accountNumber\\?": \\?")[^\\]+),
+        %r(("Invalid account number: )\d+)
       ].freeze
 
       SCRUB_REPLACEMENT = '\1[FILTERED]'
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/payway_dot_com.rb
```diff
@@ -56,7 +56,10 @@ class PaywayDotComGateway < Gateway
       SCRUB_PATTERNS = [
         %r(("password\\?":\\?")[^\\]+),
         %r(("fsv\\?":\\?")\d+),
-        %r(("accountNumber\\?":\\?")\d+)
+        %r(("fsv\\?": \\?")\d+),
+        %r(("accountNumber\\?":\\?")\d+),
+        %r(("accountNumber\\?": \\?")[^\\]+),
+        %r(("Invalid account number: )\d+)
```
curiousepic, line 62: I'm not fluent in Regex, but are you sure we don't need operators similar to the others above for this?

curiousepic (approved):
Nice catch, and this looks good - I'm not super fluent in Regex though, so if you'd like a double check on that you might get some more eyes.

tatsianaclifton:
@esmitperez Hi Esmit, this is a PR I wanted for you to take a quick look when you have a minute.

tatsianaclifton:
I think Regex should be good I did a bit more testing, going to merge this PR

PR #3932, opened 2021-03-29
Adyen: Pass Network Tx Reference for MIT

curiousepic:
Add the Network Transaction Reference for MIT requests - see https://docs.adyen.com/api-explorer/#/Payout/v30/post/payout__reqParam_additionalData-AdditionalDataCommon-networkTxReference

The new tests assume the test merchant has toggled on Network
Transaction Reference in the Additional Data Settings of the API URLs
section of the Adyen account dashboard. Otherwise the value is not
returned in initial responses.

Also add this value as a param on the Response itself.

Remote (2 failures on master):
96 tests, 361 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.9167% passed

Unit:
73 tests, 384 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -1040,19 +1040,26 @@ class RemoteAdyenTest < Test::Unit::TestCase
     assert_success response
   end
 
+  # The MIT tests may fail unless the test merchant has toggled on Network
+  # Transaction Reference in the Additional Data Settings of the API URLs
+  # section of the Adyen account dashboard
   def test_purchase_using_stored_credential_recurring_cit
     initial_options = stored_credential_options(:cardholder, :recurring, :initial)
     assert auth = @gateway.authorize(@amount, @credit_card, initial_options)
     assert_success auth
     assert_equal 'Subscription', auth.params['additionalData']['recurringProcessingModel']
+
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -1090,12 +1093,12 @@ def test_purchase_using_stored_credential_unscheduled_mit
...
     assert_success auth
     assert_equal 'UnscheduledCardOnFile', auth.params['additionalData']['recurringProcessingModel']
+
     assert capture = @gateway.capture(@amount, auth.authorization)
     assert_success capture
     assert_equal '[capture-received]', capture.message
-    assert network_transaction_id = 'none'
 
-    used_options = stored_credential_options(:unscheduled, :cardholder, id: network_transaction_id)
+    used_options = stored_credential_options(:unscheduled, :cardholder, id: auth.network_transaction_id)
```
curiousepic, line 1101: I just noticed all of the remote tests are using `cardholder` for the followup transaction, which means the NTID would not actually be sent for any of these. I will add a few for merchant-initiated followups.

tatsianaclifton (approved):
LGTM as per Adyen documentation, tests pass besides 2 remote that fail on master also.

PR #3934, opened 2021-03-29
Checkout: Add support for stored_credential

meagabeth:
CE-1444

Local:
4692 tests, 73353 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
30 tests, 140 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
34 tests, 88 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -79,6 +79,7 @@ module ActiveMerchant #:nodoc:
         add_invoice(post, amount, options)
         add_payment_method(post, payment_method, options)
         add_customer_data(post, options)
+        add_stored_credential_options(post, options)
         add_transaction_data(post, options)
         add_3ds(post, options)
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -138,6 +139,26 @@ def add_transaction_data(post, options = {})
...
       end
 
+      def add_stored_credential_options(post, options = {})
+        return unless options[:stored_credential]
+
+        case options[:stored_credential][:initial_transaction]
+        when true
+          post[:merchant_initiated] = false
+        when false
+          post[:'source.stored'] = true
```
jessiagee, line 149: Is this meant to have quotes? `'source.stored'` ?

meagabeth: Yep! That is the name of the field according to [checkout's docs](https://docs.checkout.com/quickstart/use-an-existing-card/stored-card-details) and without the quotes, tests fail because they search for a method called 'stored'.

jessiagee: Wonder if it's because it uses a `.` instead of a `_` to break up the words

jessiagee (approved):
Looks good and tests pass!

PR #3937, opened 2021-03-31
Adyen: cancelOrRefund endpoint when passed as option

naashton:
Support the cancelOrRefund endpoint when `cancel_or_refund` boolean
option is passed. Void transaction defaults to 'cancel' if the value is
false or non-existent.

Test were failing because of a change in the expected error message from
Adyen

CE-1408

Unit: 74 tests, 387 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 97 tests, 373 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


therufs (approved):
lgtm && tests pass!

PR #3944, opened 2021-04-08
Braintree: Add support for $0 auth verification

meagabeth:
If a merchant sends the field `allow_card_verification` as `true`, the `verify` method will utilize Braintree's credit-card-verification endpoint. If `allow_card_verification` is sent as `false` or not sent at all, it will execute the `verify` with the existing authorize/void transaction flow as it has in the past. To note: `allow_card_verification` is not a gateway supported field so it is not passed on to Braintree during the `verify` process. The purpose of the field is to direct which flow the `verify` method should use.

CE-555

Local:
4697 tests, 73368 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
82 tests, 188 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
83 tests, 442 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -115,10 +115,34 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def verify(credit_card, options = {})
-        MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
-          r.process(:ignore_result) { void(r.authorization, options) }
+      def verify(creditcard, options = {})
+        if options[:allow_card_verification] == true
+          options.delete(:allow_card_verification)
+          exp_month = creditcard.month.to_s
+          exp_year = creditcard.year.to_s
+          expiration = "#{exp_month}/#{exp_year}"
+          payload = {
+            credit_card: {
+              number: creditcard.number,
+              expiration_date: expiration,
+              cvv: creditcard.verification_value,
+              billing_address: {
+                postal_code: options[:billing_address][:zip]
+              }
+            }
+          }
+          commit do
+            result = @braintree_gateway.verification.create(payload)
+            response = Response.new(result.success?, message_from_transaction_result(result), response_options(result))
+            response.cvv_result['message'] = ''
+            response
+          end
+
+        else
+          MultiResponse.run(:use_first_response) do |r|
+            r.process { authorize(100, creditcard, options) }
+            r.process(:ignore_result) { void(r.authorization, options) }
+          end
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -116,9 +116,33 @@ def void(authorization, options = {})
...
-          r.process(:ignore_result) { void(r.authorization, options) }
+        if options[:allow_card_verification] == true
+          options.delete(:allow_card_verification)
+          exp_month = credit_card.month.to_s
+          exp_year = credit_card.year.to_s
+          expiration = "#{exp_month}/#{exp_year}"
+          payload = {
+            creditCard: {
+              number: credit_card.number,
+              expirationDate: expiration,
```
dsmcclain, line 127: should this be `expiration_date`? That's what I'm seeing in the Braintree docs...

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -116,9 +116,33 @@ def void(authorization, options = {})
...
+              cvv: credit_card.verification_value,
+              billing_address: {
+                postalCode: options[:billing_address][:zip]
+              }
+            }
+          }
```
dsmcclain, line 133: Also, out of curiosity, are there plans to add the `risk_data` or `external_vault` objects to this payload at some point? (It looks like maybe there are based on the conversation in Jira?)

dsmcclain (changes requested):
Good work Mo you made that look easy 😃 
Tests pass 👍 

Requesting changes on the `expirationDate` field, unless you tell me it's supposed to be that way.

lib/active_merchant/billing/gateways/braintree_blue.rb
meagabeth: Yes! Good catch - revised!

meagabeth: I also revised `postalCode` to be `postal_code` 👍

lib/active_merchant/billing/gateways/braintree_blue.rb
meagabeth: Great questions! Adding `risk_data` has been made into its own ticket in Jira for future work and we chatted about the `external_vault` situation in yesterday's standup. The end result was that since we are not always vaulting, it doesn't need to be included at this time.

dsmcclain (approved):
All tests passing. Good work Mo!

PR #3946, opened 2021-04-09
Update JCB BIN Range

dsmcclain:
CE-669

Local Tests:
4698 tests, 73371 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -11,7 +11,7 @@ module ActiveMerchant #:nodoc:
         'american_express'   => ->(num) { num =~ /^3[47]\d{13}$/ },
         'naranja'            => ->(num) { num&.size == 16 && in_bin_range?(num.slice(0, 6), NARANJA_RANGES) },
         'diners_club'        => ->(num) { num =~ /^3(0[0-5]|[68]\d)\d{11,16}$/ },
-        'jcb'                => ->(num) { num =~ /^35(28|29|[3-8]\d)\d{12}$/ },
+        'jcb'                => ->(num) { num =~ /^(35(28|29|[3-8]\d)\d{12}|308800\d{10})$/ },
         'dankort'            => ->(num) { num =~ /^5019\d{12}$/ },
         'maestro'            => lambda { |num|
           (12..19).cover?(num&.size) && (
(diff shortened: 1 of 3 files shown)


naashton (approved):
noice

dsmcclain:
> Based on patterning, should this be `\d{12}|308800\d{10}$/` instead of `\d{12}$|308800\d{10}/` ?

Yes, good catch! It should be both. Without the `$` something like `3580000000000000g` would pass the test.

I also missed the carat at the front of the `308800` option. The final version should be:
`/^35(28|29|[3-8]\d)\d{12}$|^308800\d{10}$/`

I will change it.

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -11,7 +11,7 @@ module CreditCardMethods
         'american_express'   => ->(num) { num =~ /^3[47]\d{13}$/ },
         'naranja'            => ->(num) { num&.size == 16 && in_bin_range?(num.slice(0, 6), NARANJA_RANGES) },
         'diners_club'        => ->(num) { num =~ /^3(0[0-5]|[68]\d)\d{11,16}$/ },
-        'jcb'                => ->(num) { num =~ /^35(28|29|[3-8]\d)\d{12}$/ },
+        'jcb'                => ->(num) { num =~ /^35(28|29|[3-8]\d)\d{12}$|^308800\d{10}$/ },
```
meagabeth, line 14: Should the first `$` still be in there now that it was added in towards the end?

dsmcclain: Well, I was thinking this regex was providing two completely separate options, so I needed the beginning-of-line operator and end-of-line operator in there twice, once for each option.

But based on your comment I looked back through the file for similar patterns and realized that it could be simplified as
`/^(35(28|29|[3-8]\d)\d{12}|308800\d{10})$/`

...which is probably better. So I will change it. Thank you for pointing that out.

naashton: The `|` symbol represents an `or` condition for the regex, so it should be the first pattern or the second. The `$` just denotes "end of line" for the pattern we are expecting, meaning the regular expression should match from `^` (beggining) to `$` (end).

meagabeth (approved):
Yahoooo :shipit:

PR #3949, opened 2021-04-14
Map ECI values per brand to brand-independent 3DS-focused values

bbraschi:
Several gateway integrations have their own ECI mapping, per brand, to control 3DS behavior. This PR provides such an ECI mapping so it can be shared by gateway integrations.

Planning to use this ECI mapping in #3950.


lib/active_merchant/billing/three_d_secure_eci_mapper.rb
```diff
@@ -0,0 +1,28 @@
...
+        discover: ECI_05_06_07_MAP,
+        elo: ECI_05_06_07_MAP,
+        jcb: ECI_05_06_07_MAP,
+        maestro: ECI_00_01_02_MAP,
+        master: ECI_00_01_02_MAP,
+        visa: ECI_05_06_07_MAP,
+      }.freeze
+
+      def self.map(brand, eci)
+        BRAND_TO_ECI_MAP[brand].try(:[], eci)
```
alexblackie, line 24: could do this in one call to avoid the `try`:

```suggestion
        BRAND_TO_ECI_MAP.dig(brand, eci)
```

lib/active_merchant/billing/three_d_secure_eci_mapper.rb
```diff
@@ -0,0 +1,28 @@
+module ActiveMerchant
+  module Billing
+    class ThreeDSecureEciMapper
```
pi3r, line 3: Why do you need a class if you're not using anything on the instance? I think you can just declare `module ThreeDSecureEciMapper`.

pi3r (approved):

lib/active_merchant/billing/three_d_secure_eci_mapper.rb
pi3r: Hellooo alexxx! Thank you, and also I removed you from the org :D

lib/active_merchant/billing/three_d_secure_eci_mapper.rb
bbraschi: Changed to using a module

PR #3950, opened 2021-04-14
[Checkout V2] disable 3DS based on ECI

bbraschi:


nicolas-maalouf-cko (commented):
@bbraschi thanks for that, I'm not very familiar with the section in which you used ThreeDSecureEciMapper. I'm not sure it covers cases with a missing xid and cavv from Cardinal, and sets 3ds.enabled to false then

PR #3956, opened 2021-04-26
CyberSource: Adjust Auth

naashton:
Add support for incremental auth (adjust) transactions. In scenarios
where a merchant authorizes payment but needs to make an adjustment to
that auth before capture, they make an adjust request. Currently, a
merchant would have to execute a void and a new auth before capture.

CE-1438

Unit: 102 tests, 499 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 96 tests, 496 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92.7083% passed

Our MID credentials do not support this feature, so the test fails.

### DIFF
--- a/test/fixtures.yml
+++ b/test/fixtures.yml
@@ -246,6 +246,10 @@ cyber_source:
   login: X
   password: Y
 
+cyber_source_latam_pe:
+  login: merchant_id
+  password: soap_key
+
 # Working credentials, no need to replace
 d_local:
   login: aeaf9bbfa1
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -6,6 +6,7 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
     Base.mode = :test
 
     @gateway = CyberSourceGateway.new({ nexus: 'NC' }.merge(fixtures(:cyber_source)))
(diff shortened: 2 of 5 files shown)


test/fixtures.yml
```diff
@@ -246,6 +246,14 @@ cyber_source:
   login: X
   password: Y
 
+cyber_source_latam_co:
```
therufs, line 249: It looks like we don't actually use this fixture anywhere; do we need to keep it?

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -6,8 +6,10 @@ def setup
     Base.mode = :test
 
     @gateway = CyberSourceGateway.new({ nexus: 'NC' }.merge(fixtures(:cyber_source)))
+    @gateway_latam = CyberSourceGateway.new({}.merge(fixtures(:cyber_source_latam_pe)))
 
     @credit_card = credit_card('4111111111111111', verification_value: '987')
+    @mastercard = credit_card('2222420000001113', brand: 'master')
```
therufs, line 12: Similarly, this doesn't look like it gets used anywhere?

therufs (approved):
lgtm && tests pass!

PR #3961, opened 2021-04-29
authorize.net: add settings section to refund xml

sumit-garg:
the xml created by the credit method already has a section called settings (duplicate window, email customer etc). This is missing from the xml created by the refund method


curiousepic:

PR #3965, opened 2021-04-30
Mundipagg: Use `orders` endpoint for authorize and purchase

dsmcclain:
CE-1505

Currently, all requests to Mundipagg use the `charges` endpoint. However, we need to be able to send `SubMerchant` fields on all `authorize` and `purchase` calls, and the `charges` endpoint does not receive those fields.

This PR routes all `authorize` and `purchase` calls to the `orders` endpoint, and adds logic to handle the differences between the two. Requests to `orders` have different requirements than requests to `charges`:

- instead of accepting a `payment` object, `orders` accepts a `payments` array containing one or more `payment` objects.
- `orders` also requires an `items` object with three required subfields: `amount`, `description`, and `quantity`.  These subfields cannot be blank.

The response returned from `orders` is structured differently than the response returned from `charges`. It essentially acts as a wrapper around the standard `charges` response, including one or more charges in a `charges` array, as well as an additional set of root-level fields. For this reason it is necessary to parse responses differently depending on which endpoint the request was hitting.

Rubocop:
Running RuboCop...

Unit tests:
4707 tests, 73409 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests:
39 tests, 97 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/mundipagg.rb
+++ b/lib/active_merchant/billing/gateways/mundipagg.rb
@@ -40,6 +40,7 @@ module ActiveMerchant #:nodoc:
         add_shipping_address(post, options)
         add_payment(post, payment, options)
         add_submerchant(post, options)
+        add_items(post, money, options)
 
         commit('sale', post)
       end
@@ -52,6 +53,8 @@ module ActiveMerchant #:nodoc:
         add_payment(post, payment, options)
         add_capture_flag(post, payment)
         add_submerchant(post, options)
+        add_items(post, money, options)
+
         commit('authonly', post)
       end
 
@@ -229,6 +232,14 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def add_items(post, money, options)
+        post[:items] = [{
+          amount: money,
+          description: options.dig(:items, :description) || 'no description provided',
+          quantity: options.dig(:items, :quantity) || 1
+        }]
+      end
+
       def headers
         {
           'Authorization' => 'Basic ' + Base64.strict_encode64("#{@options[:api_key]}:"),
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -288,14 +304,26 @@ def commit(action, parameters, auth = nil)
...
       end
 
-      def success_from(response)
-        %w[pending paid processing canceled active].include? response['status']
+      def extract_charge_object_from_response(response, url)
+        orders_endpoint?(url) ? charge_object(response) : response
+      end
+
+      def charge_object(response)
+        response.try(:[], 'charges').try(:[], 0)
```
dsmcclain, line 312: The use of `try` here is misleading. If for some reason we cannot find the `charges` array in the response and this method returns `nil`, other methods called in `def commit` will break. This leads to a bad situation where the transaction has already been processed by the gateway but we blow up on the way to returning a response to the merchant.

This shouldn't ever happen as long as the gateway is returning responses the way we expect. But I am open to suggestions about how to protect ourselves in this scenario.

naashton: The only place to consider guarding against the scenario where this function returns `nil` is where the returned value itself is used. So, in `message_from` you could first `return unless charge`, and in the case it is `nil` it should avoid runtime errors.

naashton, line 312: I would consider `response.dig('charges', 0)` here. It should effectively be the same as what you have, but a bit more readable.

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -288,14 +304,26 @@ def commit(action, parameters, auth = nil)
...
       end
 
-      def message_from(response)
-        return gateway_response_errors(response) if gateway_response_errors?(response)
-        return response['message'] if response['message']
-        return response['last_transaction']['acquirer_message'] if response['last_transaction']
+      def message_from(charge)
+        return gateway_response_errors(charge) if gateway_response_errors?(charge)
+        return charge['message'] if charge['message']
+        return charge['last_transaction']['acquirer_message'] if charge['last_transaction']
```
naashton, line 326: I think you can also use `.dig` here (just for the last return stmt). In the case where `acquirer_message` isn't present in the hash, you will end up with a runtime error.

lib/active_merchant/billing/gateways/mundipagg.rb
```diff
@@ -327,10 +355,10 @@ def gateway_response_errors(response)
...
       end
 
-      def authorization_from(response, action)
-        return "#{response['customer']['id']}|#{response['id']}" if action == 'store'
+      def authorization_from(charge, action)
+        return "#{charge['customer']['id']}|#{charge['id']}" if action == 'store'
 
-        response['id']
+        charge['id']
       end
```
naashton, line 362: Like you mentioned, you'll encounter downstream runtime errors if `charge_object` returns `nil`. You can head that off with a simple `return unless charge`.

naashton (changes requested):
Only a few recommended changes.

lib/active_merchant/billing/gateways/mundipagg.rb
dsmcclain: This makes sense. I resisted doing this because the end result would be to return a failure response without a message. I suppose that is preferable to a runtime error, but it still feels icky because it's possible that the transaction was successful, but then because we couldn't parse the response, `success_from` is nil, `message_from` is nil, and `error_code` is nil.

I will change it to include the guard clause like you describe.

dsmcclain: Yes, definitely an improvement. Thank you, I will change it.

lib/active_merchant/billing/gateways/mundipagg.rb
dsmcclain: Good catch!

naashton (approved):
🚢

PR #3970, opened 2021-05-05
dLocal: Add supported cardtypes

dsmcclain:
CE-1511

This PR adds `elo`, `alia`, and `carnet` to the list of cardtypes supported by the dLocal gateway.

It also adds fixes for the remote dLocal tests, most of which were failing. A new `payer.document` was required to get most of the tests passing again.

The test `test_failed_purchase_with_cabal` was removed because dLocal's sandbox does not validate cards and simply returns a success unless a failure is stimulated in the `description` field of the request (see https://docs.dlocal.com/guides/getting-started/sandbox-testing#testing-payments). 

I am seeing two failing remote tests, `test_successful_refund` and `test_partial_refund`, both of which fail with an `insufficient funds` error. This is because the sandbox merchant account has a (fake) funds balance that is currently in the negative, so the refund is rejected when it tries to assess a $20 refund fee to the account. If you are using the same sandbox account as I am when you review this PR, you should expect to see the same failures.

Rubocop:

Local:
4710 tests, 73418 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
27 tests, 70 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92.5926% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -6,7 +6,7 @@ module ActiveMerchant #:nodoc:
 
       self.supported_countries = %w[AR BR CL CO MX PE UY TR]
       self.default_currency = 'USD'
-      self.supported_cardtypes = %i[visa master american_express discover jcb diners_club maestro naranja cabal]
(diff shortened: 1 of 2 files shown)


naashton (approved):
🚢

PR #3973, opened 2021-05-07
Credorax: Add support for `d2` field

meagabeth:
`d2` is known as an echo parameter, which will be returned in the transaction response if the respective parameter is sent in the request

CE-1078

Local:
4716 tests, 73439 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
76 tests, 365 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
44 tests, 161 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


naashton (approved):
🚢

PR #3975, opened 2021-05-11
Worldpay: support cancelOrRefund via options

therufs:
Note: Currently, the cancelOrRefund method behaves in a way opposite to what Worldpay docs claim.  See https://developer.worldpay.com/docs/wpg/manage/modificationrequests#xml-reference: `<cancelOrRefund> ... Must include the <amount> information`, but actually doing this causes an error.

CE-1208

Remote:
------
67 tests, 289 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.0149% passed
Failures: test_3ds_version_2_parameters_pass_thru, test_3ds_version_1_parameters_pass_thru

Unit:
----
85 tests, 530 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -59,6 +59,7 @@
 * CyberSource: Update XML tag for merchantDefinedData [meagabeth] #3969
 * Elavon: Send ssl_vendor_id field [dsmcclain] #3972
 * Credorax: Add support for `echo` field [meagabeth] #3973
+* Worldpay: support cancelOrRefund via options [therufs] #3975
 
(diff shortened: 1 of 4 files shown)


therufs:
@dsmcclain Re the failing tests, the last several Worldpay-related commits mention 2 failures (though not what they are). The same tests fail on master (for me anyway), so I think this is just The Way It Is Right Now.

PR #3976, opened 2021-05-11
Moneris: Add authorization to `store` response

mateosuburu:
Hello! First of all thanks everyone involved in the project support.

When using the `store` method with the Moneris gateway are currently always getting a `nil` authorization in the response. According to [the docs](https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards#store), the token should be returned in `Response#authorization` instead. This commit adds that expected behavior to the Moneris gateway.

Remote:
40 tests, 219 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
51 tests, 281 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


mateosuburu:
Hi! Would be nice if someone could take a look at this. It should be safe because it's just populating a field that otherwise is always `nil`. Thanks!

PR #3978, opened 2021-05-12
Bin Range Update

naashton:
Adding several bins for MC and 1 for Maestro.

Change check on ln:207 from `range.cover?(bin)` to
`range.include?(bin)`.

The reason for this is, upon adding arrays of individual bins (following
existing convention in this file), `cover?` only applies to a range of
values. `include?` allows us to add single value arrays in existing bin
range sets and achieve same outcomes.

`include?` and `cover?` are both inclusive, so:
`(1..2).cover?(1)` returns `true`
`(1..2).include?(1)` returns `true`

CE-1593

Unit: 44 tests, 503 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: No Remote Tests

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -61,6 +61,7 @@
 * Credorax: Add support for `echo` field [meagabeth] #3973
 * Worldpay: support cancelOrRefund via options [therufs] #3975
 * Payeezy: support general credit [cdmackeyfree] #3977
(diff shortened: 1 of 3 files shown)


dsmcclain (approved):
Tests passing.

Thank you for the explanation of `include?` versus `cover?` in the PR description.

PR #3983, opened 2021-05-20
Add 'unmaskExpirationDate' and 'includeIssuerInfo' to customer profile call

dewyze:
Currently the 'unmaskExpirationDate' option only exists when getting a
customer payment profile. If you get a customer profile, which in turn
includes connected customer payment profiles, this option is not
available.

That would mean in cases where you may be trying to get a default card,
or just a list of cards, you need to make 2..n calls in order to get the
expiration date.

Additionally, the Auth.net CIM api allows for getting the issuer info:

- The Issuer Identification Number (IIN)
- Also sometimes known as the Bank Identification Number (BIN)
- Not all issuers are banks

This adds the issuer info option as well to both the customer profile call
and the customer payment profile call.

### DIFF
--- a/test/remote/gateways/remote_authorize_net_cim_test.rb
+++ b/test/remote/gateways/remote_authorize_net_cim_test.rb
@@ -267,7 +267,7 @@ class AuthorizeNetCimTest < Test::Unit::TestCase
       customer_type: 'individual',
       bill_to: @address,
       payment: {
-        credit_card: credit_card('1234123412341234')
+        credit_card: credit_card('4111111111111111')
       }
     }
     assert response = @gateway.create_customer_profile(@options)
@@ -288,8 +288,8 @@ class AuthorizeNetCimTest < Test::Unit::TestCase
 
     assert response = @gateway.get_customer_profile(customer_profile_id: @customer_profile_id)
     assert_equal 2, response.params['profile']['payment_profiles'].size
-    assert_equal 'XXXX1234', response.params['profile']['payment_profiles'][0]['payment']['credit_card']['card_number']
-    assert_equal 'XXXX4242', response.params['profile']['payment_profiles'][1]['payment']['credit_card']['card_number']
+    assert(response.params['profile']['payment_profiles'].one? { |payment| payment['payment']['credit_card']['card_number'] == 'XXXX4242' })
+    assert(response.params['profile']['payment_profiles'].one? { |payment| payment['payment']['credit_card']['card_number'] == 'XXXX1111' })
   end
 
   def test_successful_delete_customer_payment_profile_request
--- a/test/remote/gateways/remote_authorize_net_cim_test.rb
+++ b/test/remote/gateways/remote_authorize_net_cim_test.rb
@@ -73,9 +73,28 @@ class AuthorizeNetCimTest < Test::Unit::TestCase
     assert_success response
     assert_nil response.authorization
     assert response = @gateway.get_customer_profile(customer_profile_id: @customer_profile_id)
-    assert_nil response.params['profile']['merchant_customer_id']
-    assert_nil response.params['profile']['description']
     assert_equal 'new email address', response.params['profile']['email']
+  end
+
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -73,11 +73,30 @@ def test_successful_profile_create_get_update_and_delete
     assert_success response
     assert_nil response.authorization
     assert response = @gateway.get_customer_profile(customer_profile_id: @customer_profile_id)
-    assert_nil response.params['profile']['merchant_customer_id']
```
dewyze, line 76: I am not sure if this is based on previous behavior, but after a get request these are always present.

test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -179,7 +198,7 @@ def test_successful_create_customer_payment_profile_request
 
     assert response.test?
     assert_success response
-    assert_nil response.authorization
```
dewyze, line 182: This is no longer nil on any of these calls, it matches with the customer profile id.

test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -263,14 +282,14 @@ def test_successful_get_customer_profile_with_multiple_payment_profiles
...
+    assert_equal @customer_profile_id, response.authorization
     assert customer_payment_profile_id = response.params['customer_payment_profile_id']
     assert customer_payment_profile_id =~ /\d+/, "The customerPaymentProfileId should be numeric. It was #{customer_payment_profile_id}"
 
     assert response = @gateway.get_customer_profile(customer_profile_id: @customer_profile_id)
     assert_equal 2, response.params['profile']['payment_profiles'].size
-    assert_equal 'XXXX4242', response.params['profile']['payment_profiles'][0]['payment']['credit_card']['card_number']
-    assert_equal 'XXXX1234', response.params['profile']['payment_profiles'][1]['payment']['credit_card']['card_number']
+    assert_equal 'XXXX1234', response.params['profile']['payment_profiles'][0]['payment']['credit_card']['card_number']
```
dewyze, line 291: I am not sure if this order is deterministic or not, but every time we ran this, they were reversed.

pi3r (approved):
lgtm

test/remote/gateways/remote_authorize_net_cim_test.rb
JPDuchesne: Seems brittle to depend on that though, any way we can easily identify and find the profile? Seeing as these remote tests are run locally only, it's not great if it ends up breaking in the future, it won't be fun to debug.

JPDuchesne (approved):

test/remote/gateways/remote_authorize_net_cim_test.rb
```diff
@@ -263,14 +282,14 @@ def test_successful_get_customer_profile_with_multiple_payment_profiles
...
-    assert_nil response.authorization
+    assert_equal @customer_profile_id, response.authorization
     assert customer_payment_profile_id = response.params['customer_payment_profile_id']
     assert customer_payment_profile_id =~ /\d+/, "The customerPaymentProfileId should be numeric. It was #{customer_payment_profile_id}"
 
     assert response = @gateway.get_customer_profile(customer_profile_id: @customer_profile_id)
     assert_equal 2, response.params['profile']['payment_profiles'].size
-    assert_equal 'XXXX4242', response.params['profile']['payment_profiles'][0]['payment']['credit_card']['card_number']
-    assert_equal 'XXXX1234', response.params['profile']['payment_profiles'][1]['payment']['credit_card']['card_number']
+    assert(response.params['profile']['payment_profiles'].one? { |payment| payment['payment']['credit_card']['card_number'] == 'XXXX4242' })
```
dewyze, line 291: @JPDuchesne How's this?

JPDuchesne (approved):

PR #3985, opened 2021-05-21
PayTrace: new gateway build

meagabeth:
- Support the following standard AM methods: authorize, capture, purchase, void, refund, store and verify
- Authorize and purchase methods support receiving PAN or vaulted customer ID
- Support additional general methods: redact, settle
- [PayTrace API documentation](https://developers.paytrace.com/support/home)

CE-1601

Local:
4748 tests, 73563 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
16 tests, 49 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
21 tests, 56 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/pay_trace.rb
+++ b/lib/active_merchant/billing/gateways/pay_trace.rb
@@ -1,0 +1,265 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PayTraceGateway < Gateway
+      self.test_url = 'https://api.paytrace.com'
+      self.live_url = 'https://api.paytrace.com'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
+      self.homepage_url = 'https://paytrace.com/'
+      self.display_name = 'PayTrace'
+
+      # Response codes based on API Response Codes found here: https://developers.paytrace.com/support/home#14000041297
+      STANDARD_ERROR_CODE_MAPPING = {
+        '1'   => STANDARD_ERROR_CODE[:error_occurred],
+        '102' => STANDARD_ERROR_CODE[:declined],
+        '103' => STANDARD_ERROR_CODE[:auto_voided],
+        '107' => STANDARD_ERROR_CODE[:unsuccessful_refund],
+        '108' => STANDARD_ERROR_CODE[:test_refund],
+        '110' => STANDARD_ERROR_CODE[:unsuccessful_void],
+        '113' => STANDARD_ERROR_CODE[:unsuccessful_capture]
+      }
+
+      ENDPOINTS = {
+        customer_id_sale: 'transactions/sale/by_customer',
+        keyed_sale: 'transactions/sale/keyed',
+        customer_id_auth: 'transactions/authorization/by_customer',
+        keyed_auth: 'transactions/authorization/keyed',
+        capture: 'transactions/authorization/capture',
+        transaction_refund: 'transactions/refund/for_transaction',
+        transaction_void: 'transactions/void',
+        store: 'customer/create',
+        redact: 'customer/delete'
+      }
+
+      def initialize(options = {})
+        requires!(options, :username, :password, :integrator_id)
+        super
+        acquire_access_token
+      end
+
+      def purchase(money, payment_or_customerid, options = {})
+        post = {}
+        add_invoice(post, money, options)
+        if payment_or_customerid.class == String
+          post[:customer_id] = payment_or_customerid
+
+          response = commit(ENDPOINTS[:customer_id_sale], post)
+          check_token_response(response, ENDPOINTS[:customer_id_sale], post, options)
+        else
+          add_payment(post, payment_or_customerid)
+          add_address(post, payment_or_customerid, options)
+          add_customer_data(post, options)
+
+          response = commit(ENDPOINTS[:keyed_sale], post)
+          check_token_response(response, ENDPOINTS[:keyed_sale], post, options)
+        end
+      end
+
+      def authorize(money, payment_or_customerid, options = {})
+        post = {}
+        add_invoice(post, money, options)
+        if payment_or_customerid.class == String
+          post[:customer_id] = payment_or_customerid
+
+          response = commit(ENDPOINTS[:customer_id_auth], post)
+          check_token_response(response, ENDPOINTS[:customer_id_auth], post, options)
+        else
+          add_payment(post, payment_or_customerid)
+          add_address(post, payment_or_customerid, options)
+          add_customer_data(post, options)
+
+          response = commit(ENDPOINTS[:keyed_auth], post)
+          check_token_response(response, ENDPOINTS[:keyed_auth], post, options)
+        end
+      end
+
+      def capture(authorization, options = {})
+        post = {}
+        post[:amount] = amount(options[:amount]) if options[:amount]
+        post[:transaction_id] = authorization
+        response = commit(ENDPOINTS[:capture], post)
+        check_token_response(response, ENDPOINTS[:capture], post, options)
+      end
+
+      def refund(authorization, options = {})
+        # currently only support full and partial refunds of settled transactions via a transaction ID
+        post = {}
+        post[:amount] = amount(options[:amount]) if options[:amount]
+        post[:transaction_id] = authorization
+        response = commit(ENDPOINTS[:transaction_refund], post)
+        check_token_response(response, ENDPOINTS[:transaction_refund], post, options)
+      end
+
+      def void(authorization, options = {})
+        post = {}
+        post[:transaction_id] = authorization
+
+        response = commit(ENDPOINTS[:transaction_void], post)
+        check_token_response(response, ENDPOINTS[:transaction_void], post, options)
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      # The customer_IDs that come from storing cards can be used for auth and purchase transaction types
+      def store(credit_card, options = {})
+        post = {}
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -0,0 +1,259 @@
...
+        acquire_access_token
+        commit(endpoint, body)
+      end
+
+      def acquire_access_token
+        post = {}
+        post[:grant_type] = 'password'
+        post[:username] = @options[:username]
+        post[:password] = @options[:password]
+        data = post.collect { |key, value| "#{key}=#{CGI.escape(value.to_s)}" }.join('&')
```
therufs, line 148: 🤩

lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -0,0 +1,259 @@
...
+      end
+
+      def purchase(money, payment_or_customerid, options = {})
+        post = {}
+        add_invoice(post, money, options)
+        if payment_or_customerid.class == String
+          post[:customer_id] = payment_or_customerid
+
+          response = commit('transactions/sale/by_customer', post)
+          check_token_response(response, 'transactions/sale/by_customer', post, options)
```
therufs, line 38: I don't necessarily know that we should, but I'm noodling on whether it would clarify things for future generations to factor these endpoint strings out into a hash or something, just to be able to see them all in one place? e.g., 

```
ENDPOINTS = {
  customer_sale: 'transactions/sale/by_customer', 
  keyed_sale: 'transactions/sale/keyed', 
  ... 
}
```

and then calling, e.g., `check_token_response(response, ENDPOINTS[:customer_sale], post, options)`

lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -0,0 +1,259 @@
...
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Bearer )[a-zA-Z0-9:_]+), '\1[FILTERED]').
+          gsub(%r(("credit_card\\?":{\\?"number\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("cvv\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("username\\?":\\?")\w+@+\w+.+\w+), '\1[FILTERED]').
+          gsub(%r(("password\\?":\\?")\w+), '\1[FILTERED]').
+          gsub(%r(("integrator_id\\?":\\?")\w+), '\1[FILTERED]')
+      end
+
```
therufs, line 135: I think we could move this into `private`; it would be neat if we could figure out a way to test `acquire_access_token` without it having to be public 🤔

lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -0,0 +1,259 @@
...
+          test: test?,
+          error_code: success ? nil : error_code_from(response)
+        )
+      rescue JSON::ParserError
+        unparsable_response(raw_response)
+      end
+
+      def unparsable_response(raw_response)
+        message = 'Unparsable response received from PayTrace. Please contact PayTrace if you continue to receive this message.'
+        message += " (The raw response returned by the API was #{raw_response.inspect})"
```
therufs, line 219: 😍

lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -0,0 +1,259 @@
...
+      end
+
+      def redact(customer_id)
+        post = {}
+        post[:customer_id] = customer_id
+        response = commit('customer/delete', post)
+        check_token_response(response, 'customer/delete', post, options)
+      end
+
+      def settle
```
therufs, line 116: Since (I think??) this isn't part of the standard AM implementation (but is definitely handy for testing), I'm wondering if we should monkeypatch it into the gateway from the test file rather than in here? 🤔

test/remote/gateways/remote_pay_trace_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    response = @gateway.purchase(29, @amex, @options)
+    assert_failure response
+    assert_equal false, response.success?
+  end
+
+  def test_successful_authorize_and_capture
+    auth = @gateway.authorize(300, @credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(auth.authorization, @options.merge(amount: 3.00))
```
therufs, line 62: woaaah why is this a decimal when the rest are not?

test/remote/gateways/remote_pay_trace_test.rb
```diff
@@ -0,0 +1,201 @@
...
+
+  def test_successful_refund
+    purchase = @gateway.purchase(200, @credit_card, @options)
+    assert_success purchase
+    authorization = purchase.authorization
+
+    settle = @gateway.settle()
+    assert_success settle
+
+    refund = @gateway.refund(authorization, @options.merge(amount: 1.00))
```
therufs, line 105: I am not sure if it matters, but it looks like this is also a partial refund?

test/remote/gateways/remote_pay_trace_test.rb
```diff
@@ -0,0 +1,201 @@
...
+    assert_success redact
+    assert_equal true, redact.success?
+  end
+
+  def test_duplicate_customer_creation
+    create = @gateway.store(@discover, @options)
+    customer_id = create.params['customer_id']
+    response = @gateway.store(@discover, @options.merge(customer_id: customer_id))
+    assert_failure response
+    assert_equal false, response.success?
```
therufs, line 177: I think this may be superfluous in conjunction with `assert_failure response`; see https://github.com/activemerchant/active_merchant/blob/a9c588f90c4c1d59f7b0f5e1dc40b9ca12cb7fbf/test/test_helper.rb#L78

test/remote/gateways/remote_pay_trace_test.rb
```diff
@@ -0,0 +1,201 @@
...
+  end
+
+  # Not including a test_failed_verify since the only way to force a failure on this gateway is with a specific dollar amount. Since verify is auth and void combined, having separate tests for auth and void should suffice.
```
therufs, line 180: i agree! (also it would be neat to break this comment up into a few shorter lines for the benefit of those of us who insist on using ancient editors)

meagabeth: Will do! 😅

test/remote/gateways/remote_pay_trace_test.rb
meagabeth: I think that you are definitely correct. Will remove!

test/remote/gateways/remote_pay_trace_test.rb
meagabeth: I'll adjust for clarity and proper test naming 😄

test/remote/gateways/remote_pay_trace_test.rb
meagabeth: Great question! Because it is a parameter being given in the options hash instead of as a 'standalone'. Or at least that is what I have discovered to be the case.

lib/active_merchant/billing/gateways/pay_trace.rb
meagabeth: Moved to `private`! Will chat you up tomorrow about the latter part of this comment

lib/active_merchant/billing/gateways/pay_trace.rb
meagabeth: Love this idea! Will build it out before I submit for re-review

lib/active_merchant/billing/gateways/pay_trace.rb
meagabeth: We did this together! 🎉

lib/active_merchant/billing/gateways/pay_trace.rb
meagabeth: Also together!

lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -0,0 +1,272 @@
...
+      end
+
+      def redact(customer_id)
+        post = {}
+        post[:customer_id] = customer_id
+        response = commit(ENDPOINTS[:redact], post)
+        check_token_response(response, ENDPOINTS[:redact], post, options)
+      end
+
+      # def settle
```
therufs, line 129: are we ready to clean this up? :D

meagabeth: ooooops

test/remote/gateways/remote_pay_trace_test.rb
therufs: I think we maybe should be passing this option through the [`amount` method](https://github.com/activemerchant/active_merchant/blob/a9c588f90c4c1d59f7b0f5e1dc40b9ca12cb7fbf/lib/active_merchant/billing/gateway.rb#L250) so that it can obey the `money_format`, the way we're already doing in `add_invoice`?

test/unit/gateways/pay_trace_test.rb
```diff
@@ -0,0 +1,287 @@
...
+  def test_successful_verify
+    @gateway.expects(:ssl_post).times(2).returns(successful_authorize_response).then.returns(successful_void_response)
+
+    response = @gateway.verify(@credit_card, @options)
+    assert_success response
+    assert_equal true, response.success?
+  end
+
+  def test_successful_verify_with_failed_void
+    @gateway.expects(:ssl_post).times(2).returns(successful_authorize_response).then.returns(failed_void_response)
```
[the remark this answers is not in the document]
therufs: by which i mean "i have never really understood how `expects` with multiple posts worked :P

therufs (approved):
🎉 💃🏻 🤩 :shipit: 🎊

PR #3986, opened 2021-05-24
CASHNet: Update Gateway Adapter

dsmcclain:
**Note for running CASHNet remote tests:**

If you run the remote tests as is, you will see a failure on `test_successful_purchase_and_refund` and `test_successful_refund_with_options`. This is because CASHNet's test api requires one business day to pass between the purchase transactions and the attempted refund. One way around this is to send a test purchase transaction, copy the transaction number in the `tx` field of the response, _wait one day_, and then replace the `purchase.authorization` variable with your saved transaction number in the test's refund call.

I have saved up a few of these transactions numbers so you can message me whenever you intend to run the remote tests.

CE-798

Local:
4719 tests, 73458 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
6 tests, 28 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/cashnet.rb
+++ b/lib/active_merchant/billing/gateways/cashnet.rb
@@ -8,7 +8,7 @@ module ActiveMerchant #:nodoc:
 
       self.supported_countries = ['US']
       self.supported_cardtypes = %i[visa master american_express discover diners_club jcb]
-      self.homepage_url        = 'http://www.higherone.com/'
+      self.homepage_url        = 'https://transactcampus.com'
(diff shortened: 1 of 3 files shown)


therufs (approved):
Hey @dsmcclain -- 

LGTM overall. I haven't been able to get `test_successful_refund_with_options` to pass, but I'm okay with putting that in the bucket of `just activemerchant things`.  

🚢

PR #3987, opened 2021-05-25
Support braintree 3.0.1

Smittttty:
We want to be able to support Braintree's 3.0.1 gem with active merchant, as this give us access to some new parameters we need. 

This required two changes:

1. Update lib/active_merchant/billing/gateways/braintree_blue.rb to support braintree versions of 3.0.x
2. `android_pay_card` has been renamed to `google_pay_card`, so `ActiveMerchant::Billing::BraintreeBlueGateway#add_payment_method` has been modified to be aware of that.

I have ran the Braintree remote test suite, and nothing is breaking with this change. This shouldn't introduce any breaking changes.

```
4731 tests, 73512 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* Braintree: Bump required braintree gem version to 3.0.1
 * Stripe PI: ensure `setup_future_sage` and `off_session` work when using SetupIntents.
 * Orbital: Update commit to accept retry_logic in params [jessiagee] #3890
 * Orbital: Update remote 3DS tests [jessiagee] #3892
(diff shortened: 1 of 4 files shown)


InPermutation:
jinx! #3988

pi3r:
@InPermutation We should probably do yours instead since it's a higher version bump?

Smittttty:
> @InPermutation We should probably do yours instead since it's a higher version bump?

This is fine with me as long as its fine with you folks.

InPermutation:
@pi3r @Smittttty I guess it depends on how urgent the upgrade is for you. We aren't ready to merge our change yet.

Smittttty:
> I guess it depends on how urgent the upgrade is for you. We aren't ready to merge our change yet.

We need to start testing this in production soon, so we can ship this PR I guess. 

@pi3r

pi3r (approved):

PR #3993, opened 2021-06-01
Remove Braintree Version Restriction

naashton:
Lift restriction on the Braintree gem version to allow for versions
older than 3.* to work with AM. A part of this is to dynamically
determine the key name for pay card (`google_pay_card` for versions > 3
and `android_pay_card` for older versions).

CE-1579

Unit: 83 tests, 190 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 84 tests, 444 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.619% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -7,7 +7,7 @@ rescue LoadError
   raise 'Could not load the braintree gem.  Use `gem install braintree` to install it.'
 end
 
-raise "Need braintree gem >= 3.0.0. Run `gem install braintree --version '~>3.0.0'` to get the correct version." unless Braintree::Version::Major == 3 && Braintree::Version::Minor == 0
+raise 'Need braintree gem >= 2.0.0.' unless Braintree::Version::Major >= 2 && Braintree::Version::Minor >= 0
 
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -7,7 +7,7 @@
   raise 'Could not load the braintree gem.  Use `gem install braintree` to install it.'
 end
 
-raise "Need braintree gem >= 3.0.0. Run `gem install braintree --version '~>3.0.0'` to get the correct version." unless Braintree::Version::Major == 3 && Braintree::Version::Minor == 0
+raise "Need braintree gem >= 2.0.0. Run `gem install braintree --version '~>3.0.0'` to get the correct version." unless Braintree::Version::Major >= 2 && Braintree::Version::Minor >= 0
```
jessiagee, line 10: Oh wait @naashton can you update this comment to show '~> 2.0.0'?

jessiagee (changes requested):

lib/active_merchant/billing/gateways/braintree_blue.rb
naashton: I actually think that 2nd sentence should just be removed. Thanks for pointing it out!

jessiagee (approved):

lib/active_merchant/billing/gateways/braintree_blue.rb
jessiagee: 👍🏻

PR #3998, opened 2021-06-09
vPOS: credit + refund

therufs:
CE-1549

Unit:
7 tests, 22 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
11 tests, 36 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/vpos.rb
```diff
@@ -146,15 +181,21 @@ def success_from(response)
       end
 
       def message_from(response)
-        response.dig('confirmation', 'extended_response_description') ||
-          response.dig('confirmation', 'response_description') ||
-          response.dig('confirmation', 'response_details') ||
-          response.dig('messages', 0, 'key')
+        %w(confirmation refund).each do |m|
```
meagabeth, line 184: 🤩  Fancy!

lib/active_merchant/billing/gateways/vpos.rb
```diff
@@ -146,15 +181,21 @@ def success_from(response)
...
+        end
+        [response.dig('messages', 0, 'key'), response.dig('messages', 0, 'dsc')].join(':')
       end
 
       def authorization_from(response)
-        authorization_number = response.dig('confirmation', 'authorization_number')
-        shop_process_id = response.dig('confirmation', 'shop_process_id')
+        response_body = response.dig('confirmation') || response.dig('refund')
+        return unless response_body
+        authorization_number = response_body.dig('authorization_number') || response_body.dig('authorization_code')
```
meagabeth, line 197: General curiosity, what is the difference between `authorization_number` and `authorization_code`? And is there ever a possibility where both could be returned in the response body?

lib/active_merchant/billing/gateways/vpos.rb
```diff
@@ -146,15 +181,21 @@ def success_from(response)
...
 
       def message_from(response)
-        response.dig('confirmation', 'extended_response_description') ||
-          response.dig('confirmation', 'response_description') ||
-          response.dig('confirmation', 'response_details') ||
-          response.dig('messages', 0, 'key')
+        %w(confirmation refund).each do |m|
+          message =
+            response.dig(m, 'extended_response_description') ||
+            response.dig(m, 'response_description') ||
```
meagabeth, line 187: Does the `||` at the end of the above 2 lines act the same way as when they are used for things all on the same line? I like the break up here, nice and easy to read!

meagabeth (changes requested):
Added a few general questions, but they are not for changes. Rubocop is not pleased and reported 2 offenses:
Offenses:

lib/active_merchant/billing/gateways/vpos.rb:66:48: C: Style/CommentedKeyword: Do not place comments on the same line as the def keyword.
      def credit(money, payment, options = {}) # Not permitted for foreign cards.
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/active_merchant/billing/gateways/vpos.rb:196:9: C: Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.
        return unless response_body
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^

699 files inspected, 2 offenses detected

lib/active_merchant/billing/gateways/vpos.rb
therufs: It seems like we get `authorization_number`s with `purchase`s (in a `confirmation` response) and `authorization_code`s with `credit`s (in a `refund` response). I am reasonably sure we'll never get both, since as far as I can tell we never get a response with both a `confirmation` and a `refund`.

lib/active_merchant/billing/gateways/vpos.rb
therufs: It does! (In fact, I think Rubocop might fuss at you for putting multiple `||` on the same line?)

meagabeth (approved):
Thanks for answering my questions! Tests pass, go forth! 🚀

PR #4004, opened 2021-06-14
Confiable: support card type

therufs:
Unit tests:
4775 tests, 73688 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

CE-1727

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -25,6 +25,8 @@ module ActiveMerchant #:nodoc:
     # * UnionPay
     # * Alia
     # * Olimpica
+    # * Creditel
+    # * Confiable
     #
     # For testing purposes, use the 'bogus' credit card brand. This skips the vast majority of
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/credit_card.rb
```diff
@@ -103,6 +105,7 @@ def number=(value)
       # * +'alia'+
       # * +'olimpica'+
       # * +'creditel'+
+      # * +'confiable'+
```
dsmcclain, line 108: Just as a side-comment not related to your PR, I think it is weird that we have two separate lists of supported card types in this one file.

dsmcclain (approved):
Looks good, tests pass, good to go. 😄

lib/active_merchant/billing/credit_card.rb
therufs: I agree that it is weird. It is so weird that when I added Creditel I missed one of the lists :P

PR #4005, opened 2021-06-14
PayULatam: adjust phone number mapping

dsmcclain:
Corrects the mapping of phone numbers from `shipping_address` object, so that if a payment method has a phone number saved to it that will be mapped first (`billing_address[:phone]`), otherwise if the transaction contains `shipping_address[:phone_number]` that will be mapped, otherwise we will map an empty string.

CE-1665

2 unrelated failing tests (also failing on master):
`test_successful_purchase_colombia` (timeout response from colombia sandbox)
`test_well_formed_refund_fails_as_expected` (failing for at least a while, e.g.: https://github.com/activemerchant/active_merchant/pull/3657)

Rubocop:
702 files inspected, no offenses detected

Unit:
4760 tests, 73628 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
35 tests, 83 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.2857% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * NMI: Support cardholder_auth field for 3DS2 [cdmackeyfree] #4002
 * Confiable: Support cardtype [therufs] #4004
 * Maestro: Add BIN [therufs] #4003
+* PayULatam: Ensure phone number is pulled from shipping_address correctly [dsmcclain] #4005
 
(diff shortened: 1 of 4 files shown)


jessiagee (approved):
Looks good and I had the same failures/passes!

PR #4007, opened 2021-06-16
Adyen: select correct reference for `cancelOrRefund`

dsmcclain:
Adyen returns references to transactions in compound and non-compound formats, that can look a few different ways:
* `"#883623883719135C#"` (non-compound)
* `"883623883719135C"` (non-compound)
* `"883623883719135C#863623883721426D#"` (compound)

Previously we employed some complex logic to parse this string and select the correct reference. This caused a bug when `cancelOrRefund` was called with a compound reference, because it would select the second reference (`863623883721426D` in the example above) instead of the first (`883623883719135C`).

This PR simplifies the parsing of adyen references and adds some tests demonstrating the `cancelOrRefund` call selects the correct reference across a variety of situations.

CE-1543

Rubocop
705 files inspected, no offenses detected

Unit
4776 tests, 73694 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote
101 tests, 390 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -75,7 +75,7 @@ module ActiveMerchant #:nodoc:
       def refund(money, authorization, options = {})
         post = init_post(options)
         add_invoice_for_modification(post, money, options)
-        add_original_reference(post, authorization, options)
+        add_reference(post, authorization, options)
         add_splits(post, options)
         add_network_transaction_reference(post, options)
         commit('refund', post, options)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -395,20 +395,11 @@ def capture_options(options)
       end
 
       def add_reference(post, authorization, options = {})
-        _, psp_reference, = authorization.split('#')
-        post[:originalReference] = single_reference(authorization) || psp_reference
+        original_reference = authorization.split('#').reject(&:empty?).first
```
naashton, line 398: Nice

naashton (approved):
Good job simplifying this!

PR #4014, opened 2021-06-24
Orbital: Correct Success Logic for Refund

tatsianaclifton:
It is not sufficient to look at ProcStatus 0=Success for setting refunds as successful. Additional response parameter that needs to be checked is ApprovalStatus that can have the following values:  0=Declined 1=Approved 2=Message/System Error. If it was not approved RespCode will indicate the reason.

ECS-1881

Unit:
125 tests, 726 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
70 tests, 323 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* Orbital: Correct success logic for refund [tatsianaclifton] #4014
 * usaepay: Added pin gateway setting [DustinHaefele] #4026
 * MercadoPago: Added external_reference, more payer object options, and metadata field [DustinHaefele] #4020
 * Element: Add duplicate_override_flag [almalee24] #4012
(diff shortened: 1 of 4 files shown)


curiousepic:
Since eCheck support was touched recently, I think we should resolve those failures in one way or another before merging.

curiousepic:
One way to do this would be to pass information to the `success?` method so it can distinguish eCheck refunds, and omit the check for RespCode.

tatsianaclifton:
> Since eCheck support was touched recently, I think we should resolve those failures in one way or another before merging.

Making change that should fix it

Carrigan (approved):
Looks good to me :+1: Remote tests, local tests, and linting all work for me.

PR #4016, opened 2021-06-25
SimplePay gateways implementation. 

simahero:
Dear ActiveMerchant Team!

The gateway I would like to get accepted is quite complicated and I would like to get some reviews on how to imporve it to fit to your scheme.

In the contribution documentation, it is mentioned, that the option hash should be as it is, but I am afraid it is impossible to do it in SimplePay's case. There is just so much data, that would not fit in the scheme.

I have also wrote a detailed documentation for this specific gateway, which is aviable here: https://github.com/simahero/active_merchant/wiki

Reading the documentation, you might see, how it gets tricky when browser information, tokens, secrets, multiple amounts are required to make a transaction happen.

I would really appreciate your time and help.

Best regard!


simahero:
I made a branch where I have refacted a lot of stuff to get closer to active merchants "style".

Should I make a pull request for that branch?
https://github.com/simahero/active_merchant/tree/refactor

Refactored:

- no camelCase values in the options
- amount, credit_card parametes as it should be (not in the options)
- standard options hash keys, where it is possible (for example: :order_id)

PR #4018, opened 2021-06-25
Worldpay: Add support for Network Tokenization purchases

Klaiv:
EVS-874

Unit:
86 tests, 533 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote (2 pre-existing failures):
68 tests, 292 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.0588% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -747,7 +747,7 @@ module ActiveMerchant #:nodoc:
 
       def payment_details_from(payment_method)
         payment_details = {}
-        if payment_method.is_a?(NetworkTokenizationCreditCard)
+        if payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source == :network_token
           payment_details[:payment_type] = :network_token
         elsif payment_method.respond_to?(:number)
           payment_details[:payment_type] = :credit
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -390,7 +390,7 @@ module ActiveMerchant #:nodoc:
             end
           end
         elsif options[:payment_type] == :network_token
-          add_network_tokenization(xml, payment_method)
+          add_network_tokenization_card(xml, payment_method)
         else
           xml.paymentDetails credit_fund_transfer_attribute(options) do
             if options[:payment_type] == :token
@@ -417,7 +417,7 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def add_network_tokenization(xml, payment_method)
+      def add_network_tokenization_card(xml, payment_method)
         xml.paymentDetails do
           xml.tag! 'EMVCO_TOKEN-SSL', 'type' => 'NETWORKTOKEN' do
             xml.tokenNumber payment_method.number
@@ -428,6 +428,7 @@ module ActiveMerchant #:nodoc:
               )
             end
             xml.cryptogram payment_method.payment_cryptogram
+            xml.eciIndicator payment_method.eci
           end
         end
       end
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -130,6 +130,10 @@ module ActiveMerchant #:nodoc:
         true
       end
 
+      def supports_network_tokenization?
+        true
+      end
+
       def scrub(transcript)
         transcript.
           gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -428,7 +428,7 @@ module ActiveMerchant #:nodoc:
               )
             end
             xml.cryptogram payment_method.payment_cryptogram
-            xml.eciIndicator payment_method.eci
+            xml.eciIndicator format(payment_method.eci, :two_digits)
           end
         end
       end
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -389,6 +389,8 @@ module ActiveMerchant #:nodoc:
               add_amount(xml, amount, options)
             end
           end
+        elsif options[:payment_type] == :network_token
+          add_network_tokenization(xml, payment_method)
         else
           xml.paymentDetails credit_fund_transfer_attribute(options) do
             if options[:payment_type] == :token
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -730,7 +747,9 @@ def token_details_from_authorization(authorization)
 
       def payment_details_from(payment_method)
         payment_details = {}
-        if payment_method.respond_to?(:number)
+        if payment_method.is_a?(NetworkTokenizationCreditCard)
```
shasum, line 750: Like in here, it's typical to check on `NetworkTokenizationCreditCard` and it's source field to determine the type of payment method. So, something like,

```ruby
payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source == :network_token
```

shasum (commented):
@Klaiv Good start to NT on Worldpay!

Added some comments below. Also, curious if you had a chance to inspect the XML payload to confirm the new block is getting added to the request. You should be able to enable verbose output with something like,

```bash
$ DEBUG_ACTIVE_MERCHANT=true ruby -Itest test/remote/gateways/remote_worldpay_test.rb -n test_successful_purchase_with_network_token
```

lib/active_merchant/billing/gateways/worldpay.rb
Klaiv: @shasum thank you for your comments, I fixed this part in the latest commit

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -389,6 +389,8 @@ def add_payment_method(xml, amount, payment_method, options)
               add_amount(xml, amount, options)
             end
           end
+        elsif options[:payment_type] == :network_token
+          add_network_tokenization(xml, payment_method)
```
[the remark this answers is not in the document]
Klaiv: @shasum From what I can tell, this seems to be the way that worldpay payment method were being checked and I tried to stick to the same pattern.

And yes, I inspected the XML payload:
`<paymentService version="1.4" merchantCode="SPREEDLY">
  <submit>
    <order orderCode="cfd10a06f158bd97135e732240a91261">
      <description>Purchase</description>
      <amount value="100" currencyCode="GBP" exponent="2"/>
      <paymentDetails>
        <EMVCO_TOKEN-SSL type="NETWORKTOKEN">
          <tokenNumber>4895370015293175</tokenNumber>
          <expiryDate>
            <date month="09" year="2022"/>
          </expiryDate>
          <cryptogram>EHuWW9PiBkWvqE5juRwDzAUFBAk=</cryptogram>
        </EMVCO_TOKEN-SSL>
      </paymentDetails>
      <shopper>
        <shopperEmailAddress>wow@example.com</shopperEmailAddress>
        <browser>
          <acceptHeader/>
          <userAgentHeader/>
        </browser>
      </shopper>
    </order>
  </submit>
</paymentService>`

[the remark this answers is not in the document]
Klaiv: @shasum Also, the `options` here for the payment method is the ones set in the method `payment_details_from` so not reliant on the default `options`

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -415,6 +417,22 @@ def add_payment_method(xml, amount, payment_method, options)
...
+          xml.tag! 'EMVCO_TOKEN-SSL', 'type' => 'NETWORKTOKEN' do
+            xml.tokenNumber payment_method.number
+            xml.expiryDate do
+              xml.date(
+                'month' => format(payment_method.month, :two_digits),
+                'year' => format(payment_method.year, :four_digits)
+              )
+            end
+            xml.cryptogram payment_method.payment_cryptogram
+            xml.eciIndicator payment_method.eci
```
shasum, line 431: Will this result in an empty tag In case `eci` is not set in the payment method. If so, I'd double check if Worldpay is okay with an empty tag or prefers a default value or no tag when eci is not available.

shasum (approved):
LGTM 🎉

lib/active_merchant/billing/gateways/worldpay.rb
Klaiv: I tested against the sandbox with an empty tag and there were no issues

shasum: From https://developer.worldpay.com/docs/wpg/reference/decryptwallets

>Zero padded to 2 digit format (if a value of 7 is received, then add a value of 07 to the XML)

@Klaiv  Might have to format the value like is done for month and year.

Klaiv: @shasum Thank you for catching that, this is resolved now in the latest commit

jimilpatel24:
Looks pretty good to me @Klaiv. Great job. A minor question and a nit-pick below.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -415,6 +417,22 @@ def add_payment_method(xml, amount, payment_method, options)
...
+              xml.date(
+                'month' => format(payment_method.month, :two_digits),
+                'year' => format(payment_method.year, :four_digits)
+              )
+            end
+            xml.cryptogram payment_method.payment_cryptogram
+            xml.eciIndicator format(payment_method.eci, :two_digits)
+          end
+        end
+      end
```
jimilpatel24, line 434: Does it make sense to rename this method to something like `add_payment_network_token` to align with the implementation of other gateways (https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L748)?

jimilpatel24: Also, I see a number of gateways supporting the `supports_network_tokenization` method including [CyberSource](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L248) and others. The base [gateway.rb](https://github.com/activemerchant/active_merchant/blob/f54d1b6b32b9514d71b29bcc8fb0ad40aa7b9ff8/lib/active_merchant/billing/gateway.rb#L198) sets it to `false` by default. Should we add the same for worldpay? I'm not exactly sure how/where that method is used though.

cc: @shasum

Klaiv: Thanks @jimilpatel24 The methods naming seems to be varied across the gateways eg [FirstData](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/firstdata_e4.rb) uses `add_network_tokenization_credit_card` while other gateways just include the token in  `add_payment_method` eg [NMI](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/nmi.rb). Wanted to name the method `add_network_tokenization_card` because when/if we enable apple_pay/google_pay for WorldPay, it will follow the same structure.
I am fine with changing it to `add_payment_network_token`, let me know your thoughts.

As for `supports_network_tokenization`, it doesn't seem to be used anywhere but it doesn't hurt to include it so I pushed in the latest commit.

jimilpatel24: Sounds good. I will leave the method renaming decision upto you. I am fine either way.

shasum: Standardisation may not offer a lot of benefit for private methods. A reasonably expressive name should be fine.

`supports_network_tokenization?` is a good callout 👍

jimilpatel24 (approved):
I'm happy with the PR. Maybe also add a corresponding test for the new method you added? [Example](https://github.com/activemerchant/active_merchant/blob/f54d1b6b32b9514d71b29bcc8fb0ad40aa7b9ff8/test/unit/gateways/cyber_source_test.rb#L1153)

As a final step, should we also ask customer engineering to review?

Klaiv:
@jimilpatel24 Thanks, I added the unit test and customer engineering is tagged for review

therufs (approved):
tests pass && lgtm! 🏁

PR #4019, opened 2021-06-28
Mercado Pago: Added more fields purchase requests.

DustinHaefele:
Added more options to the payer object if they are present, added
Metadata object if present. Checking for exact external_reference field
as a second option when setting

CE-843

Unit:
39 tests, 181 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: The failed tests here are also failing for me on the master branch.
35 tests, 100 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.4286% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/mercado_pago.rb
+++ b/lib/active_merchant/billing/gateways/mercado_pago.rb
@@ -129,6 +129,7 @@ module ActiveMerchant #:nodoc:
 
       def add_additional_data(post, options)
         post[:sponsor_id] = options[:sponsor_id]
+        post[:metadata] = options[:metadata] if options[:metadata]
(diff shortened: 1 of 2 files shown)


DustinHaefele:
Couple thoughts for the reviewer. 

- I put the explicit external reference field as the second option in the || statement to be sure that any current behavior stayed the same but I could go both ways on that. 
- We were creating a payer object already so I merged the explicitly set payer object into that to perserve old functionality as well as add any fields to that that are explicitly passed in. 
- The ticket requests aditional nested fields for the additional_info field but any field on additional info is already being passed on so I didn't see anything that needed updating on that.

PR #4022, opened 2021-06-29
Ce 1646 barclays e pdq extra plus add gsf

yyapuncich:
`add_eci` is already a method used in this gateway. 
- wrote a test for checking a custom eci option works
- there are remote tests failing, but they are failing on master as well:
  -  `test_unsuccessful_refund`
  - `test_successful_refund`
  - `test_invalid_login`


therufs (approved):
lgtm && tests behave as described! 🚢

PR #4026, opened 2021-06-30
usaEPay gateway: Added pin as a gateway setting

DustinHaefele:
The gateway uses the pin and other data to create a  SHA hash to
authenticate transactions for accounts where pin is enabled. So now if
pin is set when creating the gatway it will pass it into the hash if
not it will be ignored like it has been.

CE-1743

unit: 50 tests, 303 assertions, 0 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications 100% passed

remote: 32 tests, 114 assertions, 1 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications 96.875% passed <= the failed test is
the scrubbed test, and it foils because the response from the gateway
includes UMrefNum field which happens to have in the log number the
same 3 digit verification_value as the test card.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* usaepay: Added pin gateway setting [DustinHaefele] #4026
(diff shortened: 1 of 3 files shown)


therufs (approved):
noice :shipit:

PR #4030, opened 2021-07-07
PayTrace: Support $0 authorize in verify method

meagabeth:
Remove support for `include_capture_amount` field, it is not needed for partial captures to succeed

CE-1777

Local:
4795 tests, 73802 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
100% passed
Remote:
27 tests, 69 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/pay_trace.rb
+++ b/lib/active_merchant/billing/gateways/pay_trace.rb
@@ -90,7 +90,7 @@ module ActiveMerchant #:nodoc:
       def refund(money, authorization, options = {})
         # currently only support full and partial refunds of settled transactions via a transaction ID
         post = {}
-        add_amount(post, money, options) if money.class == Integer
+        add_amount(post, money, options)
         post[:transaction_id] = authorization
         response = commit(ENDPOINTS[:transaction_refund], post)
         check_token_response(response, ENDPOINTS[:transaction_refund], post, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -179,7 +176,7 @@ def build_purchase_request(money, payment_or_customer_id, options)
       def build_capture_request(money, authorization, options)
         post = {}
         post[:transaction_id] = authorization
-        add_amount(post, money, options) if options[:include_capture_amount] == true
+        add_amount(post, money, options) if money.class == Integer
```
jessiagee, line 179: Why do we want to specifically check the .class here?

jessiagee (changes requested):

lib/active_merchant/billing/gateways/pay_trace.rb
meagabeth: When I was writing tests, if an empty string was passed through for `money` then it would give an error saying it needed to be a positive integer so I added this in to bypass the `add_amount` method in those cases. Based on how `core` handles capture and refund requests, AM will never receive an empty string so I have removed it and updated tests! 🎉

jessiagee (approved):
Looks good and tests pass!

Loaded suite test/unit/gateways/pay_trace_test

100% passed

Loaded suite test/remote/gateways/remote_pay_trace_test

27 tests, 69 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

bundle exec rake test:local

4795 tests, 73802 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Running RuboCop...

705 files inspected, no offenses detected

PR #4037, opened 2021-07-13
Worldpay: Synchronous Response with purchase transaction

naashton:
Worldpay is moving to a new synchronous response workflow and in some
scenarios will lead to different success cases, particularly for the
`purchase` transaction, where a `MultiResponse`, asynchronous
transaction workflow is invoked.

For some MID configurations, a transaction can be executed with a
`skip_capture` option, which allows for the `authorize` to be
executed and subsequently will be implicitly captured, rendering a
`CAPTURED` value for the `lastEvent` tag.

This change is for broadening the scope of our success criteria within
the context of an `authorize` transaction to allow for `AUTHORISED` or
`CAPTURED`.

CE-1786

Unit: 89 tests, 542 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 71 tests, 304 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.1831% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,6 +31,7 @@
 * Update bank routing account validation check [jessiagee] #4029
 * Kushki: Add 'contactDetails' fields [mbreenlyles] #4033
 * Adyen: Truncating order_id and remote test [yyapuncich] #4036
+* Worldpay: Synchronous Response with purchase transaction [naashton] #4037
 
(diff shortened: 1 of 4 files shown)


meagabeth (approved):
Looks good! 2 failures for remotes, but they have been historically failing [and on master] so good to go 🚀

PR #4041, opened 2021-07-15
Orbital: Strip null characters from responses

britth:
Sometimes, the response we receive from Orbital will have some content
that can't be parsed (specifically, we've seen unexpected null chars -
\u0000). Instead of blowing up when this happens, this change subs the
null chars in the body with a placeholder string. As this is the only
character we've seen cause issues at this point, only null is stripped
in this PR.

Remote:
70 tests, 323 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
128 tests, 747 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -37,6 +37,7 @@
 * Authorize.net: Remove cardholderAuthentication for non-3DS transactions [BritneyS] #4045
 * BlueSnap: Handle 429 errors [britth] #4044
 * Orbital: Update unit test files [meagabeth] #4046
+* Orbital: Strip null characters from responses [britth] #4041
 
 == Version 1.121 (June 8th, 2021)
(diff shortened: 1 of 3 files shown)


curiousepic:
Made a comment in the ticket

britth:
@curiousepic good call out! Updated this PR to just strip null characters from the response body, as that's the only character I've seen cause issues at this point. Since responses are non-standard, could be revisited and updated if other problem characters are observed in the future. What do you think?

curiousepic (approved):
Nice! I like it.

britth:
thanks @curiousepic !

PR #4043, opened 2021-07-15
CyberSource: Allow string content for Ignore AVS/CVV flags

curiousepic:
Previously if a string such as 'false' is sent in for ignore_avs or
ignore_cvv, those fields would be sent with the value 'true'. Now we
convert these field values to strings and check them against 'true' to
prevent false positives.

Unit:
107 tests, 528 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote (7 unrelated failures that match master):
101 tests, 512 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.0693% passed


meagabeth (approved):
Solid solution! 🚀

PR #4044, opened 2021-07-15
BlueSnap: Handle 429 errors

britth:
BlueSnap can throw 429 errors (rate limiting) at times, which are not
currently handled by the AM adapter. We do however, handle 401s and
403s. This PR adds 429s to the types of errors specifically handled.
When they occur, we will send back the response body without trying to
parse any xml.

Info on 429 error from bluesnap docs: https://developers.bluesnap.com/v8976-JSON/docs/error-handling-overview#section-http-503-and-429-errors

Unit:
40 tests, 233 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: PENDING

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -386,7 +386,7 @@ module ActiveMerchant
 
       def parse(response)
         return bad_authentication_response if response.code.to_i == 401
-        return forbidden_response(response.body) if response.code.to_i == 403
+        return generic_error_response(response.body) if [403, 429].include?(response.code.to_i)
 
         parsed = {}
         doc = Nokogiri::XML(response.body)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -564,7 +564,7 @@ def bad_authentication_response
         { 'description' => 'Unable to authenticate.  Please check your credentials.' }
       end
 
-      def forbidden_response(body)
+      def generic_error_response(body)
```
britth, line 567: not sure about this name - I was going to add another method specifically for rate limiting errors, but I wanted to return the same thing. Figured I'd use this method for both types

britth:
if anyone can run remote tests, that would be amazing - I don't seem to have access (guessing my IP isn't whitelisted)

curiousepic (approved):
👍  Looks good, I'll run remote tests.

curiousepic:
✅ 
48 tests, 164 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

britth:
@curiousepic thanks for the review and running tests!

PR #4048, opened 2021-07-16
Braintree: Fix NoMethodError for failed card verification

molbrown:
Responses from failed verify transactions using the verification endpoint
instead of the MultiResponse authorize + void have a different response
structure than successes and other transaction responses. Adds guards to
fix NoMethodError when contents are not present.

Remote:
87 tests, 458 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
85 tests, 192 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-1933

### DIFF
--- a/test/remote/gateways/remote_braintree_blue_test.rb
+++ b/test/remote/gateways/remote_braintree_blue_test.rb
@@ -199,6 +199,14 @@ class RemoteBraintreeBlueTest < Test::Unit::TestCase
     assert_equal 'P', response.avs_result['code']
   end
 
+  def test_failed_credit_card_verification
+    credit_card = credit_card('378282246310005', verification_value: '544')
+
+    assert response = @gateway.verify(credit_card, @options.merge({ allow_card_verification: true }))
+    assert_failure response
+    assert_match 'CVV must be 4 digits for American Express and 3 digits for other card types. (81707)', response.message
+  end
+
   def test_successful_verify_with_device_data
     # Requires Advanced Fraud Tools to be enabled
     assert response = @gateway.verify(@credit_card, @options.merge({ device_data: 'device data for verify' }))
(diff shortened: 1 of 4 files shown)


Carrigan (commented):
Code here looks great :+1:

Only suggestion might be to also add a unit test here with a known problematic response since gateways can change the remote test results. What do you think?

test/remote/gateways/remote_braintree_blue_test.rb
```diff
@@ -199,6 +199,14 @@ def test_successful_credit_card_verification
     assert_equal 'P', response.avs_result['code']
   end
 
+  def test_failed_credit_card_verification
+    credit_card = credit_card('378282246310005', verification_value: '544')
+
+    assert response = @gateway.verify(credit_card, @options.merge({ allow_card_verification: true }))
+    assert_failure response
+    assert_match "CVV must be 4 digits for American Express and 3 digits for other card types. (81707)", response.message
```
molbrown, line 207: @Carrigan I can actually remove this line with the message match, the assert_failure alone has the same change in behavior on master vs this branch. What do you think?

Carrigan: I think the remote test is fine, it might just be good to have an additional test with a crystalized response so that it is more repeatable and obvious what type of error this catches

molbrown: Totally 👍  updated!

Carrigan (changes requested):
Nice! All tests pass for me :+1: Only last thing is there is a linting error:

```
test/remote/gateways/remote_braintree_blue_test.rb:207:18: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://github.com/rubocop-hq/ruby-style-guide#consistent-string-literals)
    assert_match "CVV must be 4 digits for American Express and 3 digits for other card types. (81707)", response.message
```

molbrown:
Good catch, thanks 😁

Carrigan (approved):

PR #4049, opened 2021-07-19
Elavon: Add support for special characters

mbreenlyles:
Properly handle encoding and decoding of special characters and string truncation to avoid invalid XML and max length exceeded errors, according to gateway specifications.

Remote:
36 tests, 164 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.6667% passed
- Failed on test_successful_purchase_with_custom_fields: not setup on account
- Failed on test_successful_purchase_with_level_3_fields: not setup on account
- Failed on test_successful_purchase_with_shipping_address_and_l3: not setup on account

Unit:
45 tests, 237 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ECS-1894

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -201,8 +201,8 @@ module ActiveMerchant #:nodoc:
       private
 
       def add_invoice(xml, options)
-        xml.ssl_invoice_number    truncate((options[:order_id] || options[:invoice]), 25)
-        xml.ssl_description       truncate(options[:description], 255)
+        xml.ssl_invoice_number    url_encode_truncate((options[:order_id] || options[:invoice]), 25)
+        xml.ssl_description       url_encode_truncate(options[:description], 255)
       end
 
       def add_approval_code(xml, authorization)
@@ -219,8 +219,8 @@ module ActiveMerchant #:nodoc:
 
         add_verification_value(xml, creditcard) if creditcard.verification_value?
 
-        xml.ssl_first_name    truncate(creditcard.first_name, 20)
-        xml.ssl_last_name     truncate(creditcard.last_name, 30)
+        xml.ssl_first_name    url_encode_truncate(creditcard.first_name, 20)
+        xml.ssl_last_name     url_encode_truncate(creditcard.last_name, 30)
       end
 
       def add_currency(xml, money, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -1,5 +1,7 @@
 require 'active_merchant/billing/gateways/viaklix'
 require 'nokogiri'
+require 'erb'
+require 'cgi'
```
curiousepic, line 4: I think this isn't necessary here since it's required by `active_merchant.rb`

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -1,5 +1,7 @@
 require 'active_merchant/billing/gateways/viaklix'
 require 'nokogiri'
+require 'erb'
```
curiousepic, line 3: This would be the first time this is required in AM. I'm wondering if it's necessary or if there is another way to perform the url encoding...

test/remote/gateways/remote_elavon_test.rb
```diff
@@ -424,6 +424,21 @@ def test_successful_purchase_with_truncated_data
...
+
+    response = @gateway.purchase(@amount, special_card, @options)
+
+    assert_success response
+    assert response.test?
+    assert_equal 'APPROVAL', response.message
+    assert_equal 'Fear & Loathin', response.params['first_name']
+    assert_equal 'Castañeda', response.params['last_name']
+    assert response.authorization
+  end
```
curiousepic, line 440: Just to confirm - are we getting the gateway error for this test with the current adapter behavior on master?

mbreenlyles: No, this is just testing if the gateway request was encoded/decoded correctly. I think that specific error resulted from an issue with invalid UTF-8 bytes being submitted thru iFrame and then .delete being called on an invalid byte. I will write a test for that!

mbreenlyles: Possibly that error was also related to the iso-8859-1 forced encoding introducing an invalid UTF-8 byte

mbreenlyles:
Updated the case to test what it says it's fixing and confirmed it failed on master and passes on mbreenlyles branch ohhhh yeaaahhh 🤙

mbreenlyles:
Confirmed that CE-1345 and ECS-1744 errors can be duplicated on master and are both fixed by this encoding update.

curiousepic (approved):
This looks really good and tests pass locally. Great work!

PR #4053, opened 2021-07-21
PayArc: Formatting CC month, adding tax_rate, removing default void reason

jessiagee:
* Fixed an issue with CC months not working for '01', '02', etc.
* Added `tax_rate` to the charges allowed list
* Removed the `duplicate` default void reason

Loaded suite test/remote/gateways/remote_pay_arc_test

`test_successful_credit` fails due to MID configuration on PayArc's side

22 tests, 54 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.4545% passed

Loaded suite test/unit/gateways/pay_arc_test

14 tests, 42 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

bundle exec rake test:local

4835 tests, 73897 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Running RuboCop...
Inspecting 707 files

707 files inspected, no offenses detected

PR #4054, opened 2021-07-22
Pay arc bug fix for void.

cdm-83:
HI @jessiagee ,

We have removed the default reason for the void as per request from Punith.

Senthil


jessiagee:
I've pulled this into https://github.com/activemerchant/active_merchant/pull/4053 to consolidate

PR #4060, opened 2021-07-27
NTID field for Stripe Payment Intents

cdmackeyfree:
Adds a way to send in a network_transaction_id outside of the stored credentials hash for Stripe Payment Intents. Stripe PI accepts network_transaction_id and ds_transaction_id via mit field under card.

Remote:

56 tests, 267 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
24 tests, 144 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -31,6 +31,7 @@ module ActiveMerchant #:nodoc:
         setup_future_usage(post, options)
         add_exemption(post, options)
         add_stored_credentials(post, options)
+        add_ntid(post, options)
         add_error_on_requires_action(post, options)
         request_three_d_secure(post, options)
 
@@ -295,6 +296,11 @@ module ActiveMerchant #:nodoc:
         post[:payment_method_options][:card][:moto] = true if options[:moto]
       end
 
+      # Stripe Payment Intents does not pass any parameters for cardholder/merchant initiated
+      # it also does not support installments for any country other than Mexico (reason for this is unknown)
+      # The only thing that Stripe PI requires for stored credentials to work currently is the network_transaction_id
+      # network_transaction_id is created when the card is authenticated using the field `setup_for_future_usage` with the value `off_session` see def setup_future_usage below
+
       def add_stored_credentials(post, options = {})
         return unless options[:stored_credential] && !options[:stored_credential].values.all?(&:nil?)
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -303,9 +309,19 @@ def add_stored_credentials(post, options = {})
         post[:payment_method_options][:card] ||= {}
         post[:payment_method_options][:card][:mit_exemption] = {}
 
-        # Stripe PI accepts network_transaction_id and ds_transaction_id via mit field under card.
-        post[:payment_method_options][:card][:mit_exemption][:network_transaction_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
+        # Stripe PI accepts network_transaction_id and ds_transaction_id via mit field under card. It can be sent in nested under stored credentials OR as its own field (add_ntid handles when it is sent in on its own)
         post[:payment_method_options][:card][:mit_exemption][:ds_transaction_id] = stored_credential[:ds_transaction_id] if stored_credential[:ds_transaction_id]
+        post[:payment_method_options][:card][:mit_exemption][:network_transaction_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
```
jessiagee, line 314: Do we need to have a check not to set this if options[:network_transaction_id] exists?

cdmackeyfree: I believe the `return unless` portion handles that. If the options include [:stored_credentials] this path is taken and the ntid will be set from the network_transaction_id that is inside stored_credentials, but if they do not, it isn't set. Same with the `add_ntid` method below it using `return unless` network_transaction_id is in options.

cdmackeyfree: I guess there could be a chance that someone might send in stored_credentials AND a network_transaction_id in the options level. 🤔

cdmackeyfree: I ran the test with ntid being sent in under options AND under stored credentials and it still sent the request with the options[:network_transaction_id]. So it looks like the stored credentials will not overwrite the field if it is sent in separately. Thoughts?

jessiagee: Sounds good! You can leave the test in there to show what happens in case anyone else has the same question later on, but that was mostly my interest too re: sending in both in the `options` field.

meagabeth:
Main unit and remote tests are passing, but Rubocop is complaining

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -295,6 +296,11 @@ def add_exemption(post, options = {})
         post[:payment_method_options][:card][:moto] = true if options[:moto]
       end
 
+      # Stripe Payment Intents does not pass any parameters for cardholder/merchant initiated
+      # it also does not support installments for any country other than Mexico (reason for this is unknown)
+      # The only thing that Stripe PI requires for stored credentials to work currently is the network_transaction_id
+      # network_transaction_id is created when the card is authenticated using the field `setup_for_future_usage` with the value `off_session` see def setup_future_usage below
```
meagabeth, line 302: I appreciate these comments! 🌟

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -14,27 +14,27 @@ def setup
     @three_ds_off_session_credit_card = credit_card('4000002500003155',
       verification_value: '737',
       month: 10,
-      year: 2022)
+      year: 2028)
```
meagabeth, line 17: Thanks for being proactive about these dates!

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -445,6 +462,26 @@ def test_purchase_works_with_stored_credentials_without_optional_ds_transaction_
     end
   end
 
+  def test_succeeds_with_ntid_in_stored_credentials_and_separately
```
meagabeth, line 465: 📌  For later, but I feel like we should be sure to mention in the docs update what happens if/when `network_transaction_id` is sent in both places.

meagabeth (changes requested):
This all looks great! Just a few Rubocop complaints, but once those are resolved this should be good to go ✨

meagabeth (approved):
Awesome work! :shipit:

PR #4070, opened 2021-07-30
Worldpay: Support 'CAPTURED' response for `authorize` transactions

naashton:
Some merchants that execute `authorize` transactions with Worldpay are configured to
automatically capture those transactions via the payment processor
(dependent on region) and therefore can expect to receive 'CAPTURED' in
the list of success criteria when checking the `last_event` value.

Expanding the `authorize_request` to expect `AUTHORISED` or `CAPTURED`
handles this case for those merchants

CE-1786

Unit: 88 tests, 538 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 71 tests, 304 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.1831% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -54,6 +54,7 @@
 * Spreedly: Support gateway_specific_response_fields in response params [abarrak] #4064
 * Payeezy: Add support for `add_soft_descriptors` [rachelkirk] #4069
 * Stripe Payment Intents: Add support for network_transaction_id field [cdmackeyfree] #4060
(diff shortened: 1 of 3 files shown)


cdmackeyfree (approved):
Tests pass (except the two failing assertions on master from  test_3ds_version_1_parameters_pass_thru(RemoteWorldpayTest)) 🚢

PR #4073, opened 2021-08-02
Ingenico (Global Collect): New idempotence key header

BritneyS:
This change adds a new header capable of passing an idempotence key to the `global_collect` gateway. Note that the only non-passing remote test fails with a `502` bad gateway error response, but this is noted as being an intermittent response from the gateway's side.

Test Summary
Local: 4844 tests, 73931 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Unit: 28 tests, 136 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Remote: 25 tests, 86 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -33,7 +33,7 @@ module ActiveMerchant #:nodoc:
         add_creator_info(post, options)
         add_fraud_fields(post, options)
         add_external_cardholder_authentication_data(post, options)
-        commit(:authorize, post)
+        commit(:authorize, post, options: options)
       end
 
       def capture(money, authorization, options = {})
@@ -41,7 +41,7 @@ module ActiveMerchant #:nodoc:
         add_order(post, money, options, capture: true)
         add_customer_data(post, options)
         add_creator_info(post, options)
-        commit(:capture, post, authorization)
+        commit(:capture, post, authorization: authorization)
       end
 
       def refund(money, authorization, options = {})
@@ -49,13 +49,13 @@ module ActiveMerchant #:nodoc:
         add_amount(post, money, options)
         add_refund_customer_data(post, options)
         add_creator_info(post, options)
-        commit(:refund, post, authorization)
+        commit(:refund, post, authorization: authorization)
       end
 
       def void(authorization, options = {})
         post = nestable_hash
         add_creator_info(post, options)
-        commit(:void, post, authorization)
+        commit(:void, post, authorization: authorization)
       end
 
       def verify(payment, options = {})
@@ -277,9 +277,13 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def commit(action, post, authorization = nil)
+      def idempotency_key_for_signature(options)
+        "x-gcs-idempotence-key:#{options[:idempotency_key]}" if options[:idempotency_key]
+      end
+
+      def commit(action, post, authorization: nil, options: {})
         begin
-          raw_response = ssl_post(url(action, authorization), post.to_json, headers(action, post, authorization))
+          raw_response = ssl_post(url(action, authorization), post.to_json, headers(action, post, authorization, options))
           response = parse(raw_response)
         rescue ResponseError => e
           response = parse(e.response.body) if e.response.code.to_i >= 400
@@ -306,21 +310,26 @@ module ActiveMerchant #:nodoc:
         }
       end
 
-      def headers(action, post, authorization = nil)
-        {
+      def headers(action, post, authorization = nil, options = {})
+        headers = {
           'Content-Type' => content_type,
-          'Authorization' => auth_digest(action, post, authorization),
+          'Authorization' => auth_digest(action, post, authorization, options),
           'Date' => date
         }
+
+        headers['X-GCS-Idempotence-Key'] = options[:idempotency_key] if options[:idempotency_key]
+        headers
       end
 
-      def auth_digest(action, post, authorization = nil)
+      def auth_digest(action, post, authorization = nil, options = {})
         data = <<~REQUEST
           POST
           #{content_type}
           #{date}
+          #{idempotency_key_for_signature(options)}
           #{uri(action, authorization)}
         REQUEST
+        data = data.each_line.reject { |line| line.strip == '' }.join
         digest = OpenSSL::Digest.new('sha256')
         key = @options[:secret_api_key]
         "GCS v1HMAC:#{@options[:api_key_id]}:#{Base64.strict_encode64(OpenSSL::HMAC.digest(digest, key, data))}"
--- a/test/unit/gateways/global_collect_test.rb
+++ b/test/unit/gateways/global_collect_test.rb
@@ -368,6 +368,16 @@ class GlobalCollectTest < Test::Unit::TestCase
     assert_equal @gateway.scrub(response), scrubbed_invalid_json_plus
   end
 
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -307,11 +307,14 @@ def json_error(raw_response)
...
 
       def headers(action, post, authorization = nil)
-        {
+        headers = {
           'Content-Type' => content_type,
           'Authorization' => auth_digest(action, post, authorization),
           'Date' => date
         }
+
+        headers['x-gcs-idempotence-key'] = options[:idempotency_key] if options[:idempotency_key]
```
BritneyS, line 316: This header key is lowercase due to the [example headers in Ingenico's documentation](https://epayments-api.developer-ingenico.com/s2sapi/v1/en_US/java/authentication.html?paymentPlatform=GLOBALCOLLECT), but I'm open to feedback about this 📝

dsmcclain: Thank you for linking the documentation.

My understanding is that HTTP headers are case-insensitive. So theoretically it shouldn't matter either way. 

That being said, I couldn't find any lowercase examples in the documentation, am I missing it? The only examples I can find of this header are from [this page](https://epayments-api.developer-ingenico.com/s2sapi/v1/en_US/java/idempotent-requests.html?paymentPlatform=GLOBALCOLLECT), and they show it being capitalized.

I agree with you that it would be preferable to follow whatever convention is displayed in the gateway documentation. If the docs show both uppercase and lowercase, then we might as well leave it as is. But if it's only showing uppercase then I would request you change it to be uppercase.

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -307,11 +307,14 @@ def json_error(raw_response)
       end
 
       def headers(action, post, authorization = nil)
```
dsmcclain, line 309: `options` is not being passed to this method. So `options[:idempotency_key]` is always going to be `nil` here.

Part of what's confusing about this is that you would expect `options[:idempotency_key]` on line 316 to throw an error if `options` doesn't exist. But what's happening is that `options` has _also_ been declared as an instance variable with read-only access in the Gateway class [here](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateway.rb#L145), and since every gateway adapter inherits from that class, there is an `options` variable that floats around in our gateway adapters and causes problems in situations like this.

If you want to see this in action, I recommend using a `binding.pry` breakpoint both in this method and in `def authorize`, then running your unit test and seeing what `options` returns in each scope.

This issue with `options` as an invisible instance variable is a perennial source of pain that we are kind of just stuck with, unfortunately.

Anyways, this method needs access to the local `options` variable so that the `idempotency_key` can be added, if present.

test/unit/gateways/global_collect_test.rb
```diff
@@ -368,6 +368,16 @@ def test_scrub_invalid_response
     assert_equal @gateway.scrub(response), scrubbed_invalid_json_plus
   end
 
+  def test_authorize_with_optional_idempotency_key_header
+    response = stub_comms do
+      @gateway.authorize(@accepted_amount, @credit_card, @options.merge(idempotency_key: 'test123'))
+    end.check_request do |_endpoint, _data, headers|
+      headers && headers['x-gcs-idempotence-key'] == 'test123'
```
dsmcclain, line 375: This is an awesome test that is exactly what we want, but this is the line that is letting you down.

When you run this test `headers && headers['x-gcs-idempotence-key'] == 'test123'` returns `false`. But there is no assertion so the test doesn't care and just moves on, ultimately passing.

I think this line could probably something like `assert_equal headers['x-gcs-idempotence-key'], 'test123'`

dsmcclain (changes requested):
Almost there, @BritneyS !

There is a tricky little gotcha that is causing problems, which I explain in my comments. Let me know if you have any questions about it.

lib/active_merchant/billing/gateways/global_collect.rb
BritneyS: @dsmcclain Thank you for the thorough review. I've pushed up changes! They ended up being significant due to some 'aha' moments I had when making corrections.

lib/active_merchant/billing/gateways/global_collect.rb
dsmcclain: You new revisions took me deeper into Ingenico's documentation on building their authorization signature, and you are right that the header needs to be lowercase, at least in the `auth_digest` method because they state that you must lowercase each header name when creating the signature.

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -277,9 +277,13 @@ def uri(action, authorization)
         end
       end
 
-      def commit(action, post, authorization = nil)
+      def idempotency_key(options)
```
dsmcclain, line 280: Love this method. Do you think it would be better named `idempotency_key_for_signature` to distinguish it from the actual header that gets added on line 320? Just a suggestion, I defer to your judgement on it.

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -277,9 +277,13 @@ def uri(action, authorization)
...
+        "x-gcs-idempotence-key:#{options[:idempotency_key]}" if options[:idempotency_key]
+      end
+
+      def commit(action, post, authorization = nil, options = {})
```
dsmcclain, line 284: You did a nice job managing the addition of another argument with a default value to this method.

An alternative way to do this would be to use ruby's Keyword Arguments.

In this approach, the method could be defined like
```
def commit(action, post, authorization: nil, options: {})
```
where `authorization:` and `options:` uses the keyword syntax. Keyword Arguments are not positionally dependent, so this would allow you to set a default for both arguments AND ALSO let you call the method and pass `options` without needing to pass a value for `authorization`. So in `def authorize` you could call it like this:
```
commit(:authorize, post, options: options)
```
while in other places like `def capture` you can call it like this
```
commit(:capture, post, authorization: authorization)
```

This saves you from having to pass this annoying `nil` around. The downside is that you cannot pass values for the keyword arguments without using the keyword (i.e. you cannot call it like `commit(:capture, post, authorization)`), so you would have to change it everywhere you call it.

Maybe this would be a confusing pattern for some people. I don't see it used much in Active Merchant (although it is already being used in this file, with `def add_order`). Another judgement call that comes down to what you think is more readable and understandable.

My apologies if you were already familiar with keyword arguments and had already considered this approach.

dsmcclain (approved):
Wow, nice work @BritneyS . The way this gateway creates an authorization signature is complicated and nit-picky. Well done going deep into the documentation to get it built correctly.

I added two suggestions that are both ultimately judgement calls. The PR is fine as is or you can make the changes I suggest if you think they are improvements. I will approve this, and if you make the changes you are welcome to re-tag me for review again.

If you do merge the PR, don't forget the changelog entry!

lib/active_merchant/billing/gateways/global_collect.rb
BritneyS: @dsmcclain I agreed with this and everything else so I pushed up the changes. Thanks again!

dsmcclain (approved):
Great work on this one, @BritneyS .

All tests pass, good to ship 👍

PR #4074, opened 2021-08-02
PayTrace: Adjust handling of `line_items` subfields

meagabeth:
Several of the fields within the `line_items` field were not being formatted properly due to them being nested. Previously, the `amount` method was being called on *_amount fields, which was giving an error related to them needing to be integers. The fields that have boolean values were being transformed into strings and throwing errors at the gateway level. These changes remedy both of those issues.

Local:
4846 tests, 73941 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
100% passed
Remote:
27 tests, 69 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/pay_trace.rb
+++ b/lib/active_merchant/billing/gateways/pay_trace.rb
@@ -200,6 +200,17 @@ module ActiveMerchant #:nodoc:
         payment_or_customer_id.class == String
       end
 
+      def string_literal_to_boolean(value)
+        return value unless value.class == String
+
+        if value.casecmp('true').zero?
+          true
+        elsif value.casecmp('false').zero?
+          false
+        else return nil
+        end
+      end
+
       def add_customer_data(post, options)
         return unless options[:email]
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -200,6 +200,16 @@ def customer_id?(payment_or_customer_id)
         payment_or_customer_id.class == String
       end
 
+      def to_boolean(value)
```
dsmcclain, line 203:
```
to_boolean(true)
=> nil

to_boolean(nil)
=>nil
```

I don't know if these are scenarios that we need to worry about or not. We probably don't have to worry about `nil` because you are only calling `to_boolean` if the field is present. But could the field ever come in as a boolean already? In that case `to_boolean` will return `nil` instead of the value we expect.

These cases might not matter for this gateway. But I do think the name of the method is misleading. Maybe something like `def string_literal_to_boolean`? Or `def convert_true_or_false_string`? Not very charismatic but at least it would force anyone who might use the method to look closely at what it does. I just worry about calling it `to_boolean` in case somebody else ends up using it in a way that's not intended.

If it turns out you do want to handle more cases than the two string literals, you could do something like this:

```
def to_boolean(value)
  value.to_s.downcase == 'true'
end

to_boolean(true)
=> true

to_boolean(nil)
=> false
```

...but then any random value in there, like 'test', will return `false` and maybe that's not what you want.

lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -232,16 +242,16 @@ def add_payment(post, payment)
       def add_level_3_data(post, options)
         post[:invoice_id] = options[:invoice_id] if options[:invoice_id]
         post[:customer_reference_id] = options[:customer_reference_id] if options[:customer_reference_id]
-        post[:tax_amount] = amount(options[:tax_amount]) if options[:tax_amount]
-        post[:national_tax_amount] = amount(options[:national_tax_amount]) if options[:national_tax_amount]
+        post[:tax_amount] = options[:tax_amount].to_i if options[:tax_amount]
```
dsmcclain, line 245: I have a similar worry here, that weird values passed in to these fields will have unintended results.
```
'test'.to_i
=> 0

''.to_i
=> 0
```
So I just want to ask if this is something we should guard against? Maybe it doesn't even matter because maybe a zero in these fields would have no impact.

Please excuse me if this is a scenario you have considered.

dsmcclain (changes requested):
I have one request for change on the naming of the `to_boolean` method and one question about the use of `to_i`.

lib/active_merchant/billing/gateways/pay_trace.rb
meagabeth: Thanks for pointing this out, that's a great thing to consider. In this case, the value of `0` is fine. It would be accepted by the gateway and wouldn't have an impact on the transaction.

dsmcclain (approved):
Thanks for making those changes, @meagabeth . Nice job adding a test to cover the new behavior.

All tests passing, approved for takeoff ✈️

PR #4076, opened 2021-08-03
Worldpay: Correct Expiration Year Format

tatsianaclifton:
This pull request fixes the issue when expiration 2 digits year is converted to 4 digits by adding 00to the beginning. For example, '22' becomes '0022'. Now if year is 2 digits '20' is added to the beginning. For example, '22' becomes '2022'.

ECS-1973
bundle exec rake test:local
4847 tests, 73948 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
534.01 tests/s, 8147.08 assertions/s
Running RuboCop...
Inspecting 707 files

Unit:
89 tests, 541 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote (2 failed same as on master):
72 tests, 307 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.2222% passed

### DIFF
--- a/lib/active_merchant/billing/credit_card_formatting.rb
+++ b/lib/active_merchant/billing/credit_card_formatting.rb
@@ -15,6 +15,7 @@ module ActiveMerchant #:nodoc:
         case option
         when :two_digits  then sprintf('%.2i', number.to_i)[-2..-1]
         when :four_digits then sprintf('%.4i', number.to_i)[-4..-1]
+        when :four_digits_year then number.to_s.length == 2 ? '20' + number.to_s : format(number, :four_digits)
         else number
         end
       end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -57,6 +57,7 @@
 * Worldpay: Support 'CAPTURED' response for authorize transactions [naashton] #4070
 * Ingenico (Global Collect): New idempotence key header [BritneyS] #4073
 * PayTrace: Adjust handling of line_items subfields [meagabeth] #4074
+* Worldpay: Correct Expiration Year Format [tatsianaclifton] #4076
 
 == Version 1.121 (June 8th, 2021)
 * Braintree: Lift restriction on gem version to allow for backwards compatibility [naashton] #3993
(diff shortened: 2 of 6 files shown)


lib/active_merchant/billing/credit_card_formatting.rb
```diff
@@ -15,6 +15,7 @@ def format(number, option)
         case option
         when :two_digits  then sprintf('%.2i', number.to_i)[-2..-1]
         when :four_digits then sprintf('%.4i', number.to_i)[-4..-1]
+        when :four_digits_year then number.to_s.length == 2 ? '20' + number.to_s : number.to_s
```
Carrigan, line 18: The other two cases protect against a longer number being given; this new case will always yield 4 digits if the input is 2 digits, but for any other digit count, it will equal the input digit count.

I might recommend switching to leverage the existing four digit year code:

```suggestion
        when :four_digits_year then number.to_s.length == 2 ? '20' + number.to_s : format(number, :four_digits)
```

Carrigan (approved):
The change makes sense and looks good! All related remote and unit tests pass locally for me (though there are some failures on Monei in unit tests right now). I left one suggestion below but feel free to take it or leave it.

lib/active_merchant/billing/credit_card_formatting.rb
tatsianaclifton: That is a good suggestion, I will change to use that

tatsianaclifton:
> The change makes sense and looks good! All related remote and unit tests pass locally for me (though there are some failures on Monei in unit tests right now). I left one suggestion below but feel free to take it or leave it.

Thanks for review and suggestions. I made a change based on your suggestions. I am not sure why you saw some failures on Monei tests, all pass for me when running just them as well as running bundle exec rake test:local

Carrigan:
@tatsianaclifton I had not added Monei as a fixture yet - it is all good now :+1:

PR #4081, opened 2021-08-10
Redsys: Route MIT Exemptions to webservice endpoint

curiousepic:
For MIT-exempt transactions, automatically add the DIRECTPAYMENT field
and route them to the webservice endpoint as a trataPeticion for proper
handling.

Also add flag to force routing to the webservice endpoint and renames
some methods and values for clarity and accuracy.

### DIFF
--- a/test/remote/gateways/remote_redsys_sha256_test.rb
+++ b/test/remote/gateways/remote_redsys_sha256_test.rb
@@ -207,7 +207,7 @@ class RemoteRedsysSHA256Test < Test::Unit::TestCase
     assert_equal 'Y', response.params['ds_card_psd2']
     assert_equal 'CardConfiguration', response.message
 
-    # ensure MIT is supported
+    # ensure MIT exemption is recognized
     assert response.params['ds_excep_sca']
     assert_match 'MIT', response.params['ds_excep_sca']
   end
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_redsys_sha256_test.rb
```diff
@@ -198,31 +199,31 @@ def test_successful_auth_purchase_with_stored_credentials
   end
 
   def test_successful_3ds2_purchase_with_mit_exemption
-    options = @options.merge(execute_threed: true, terminal: 12)
-    response = @gateway.purchase(@amount, @threeds2_credit_card, options.merge(sca_exemption: 'MIT', sca_exemption_direct_payment_enabled: true))
+    options = @options.merge(execute_threed: true)
```
esmitperez, line 202: what's behind the removal of terminal 12?

Carrigan (changes requested):
Rubocop has a couple of failures related to the change that we should take care of before merging, but otherwise the changes all look good and make sense to me :+1: All remote tests pass the same as master.

curiousepic:
Thanks, addressed the Rubocop alerts.

Carrigan (approved):
Thanks for fixing the linting errors! LGTM :+1:

esmitperez:
I was able to run remote tests! I forgot to adjust the AM hash 🙈 . 

I don't recall the exact steps to reproduce fix in Postman, but we can pair on it

Carrigan (approved):

PR #4084, opened 2021-08-17
Feature/paypal express order

mzishtiaq:
This PR adds a method with `Order` payment action, more information about payment actions can be found [here](https://www.paypal.com/uk/smarthelp/article/what-are-the-differences-between-the-express-checkout-payment-actions-ts1501). This action allows merchants to partially capture payments after shipping individual items.


mzishtiaq:
Sorry folks. This was meant to be a PR on my fork instead of on the project.

PR #4086, opened 2021-08-20
Enables Recurring Transactions on Elavon

cdmackeyfree:
For recurring transactions, an initial call must be made to the gateway using merchant_initiated_unscheduled, as well as ssl_get_token (to request the token) and ssl_add_token (to store that token with the payment method). Once stored the ssl_token will be sent with subsequent purchases that are marked as "recurring".

Local:
4853 tests, 73995 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
46 tests, 243 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
38 tests, 171 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
89.4737% passed

The four failing remote tests are also failing on the master branch and I am working with Elavon to get them resolved.

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -301,6 +301,7 @@ module ActiveMerchant #:nodoc:
         xml.ssl_entry_mode                      entry_mode(options) if entry_mode(options)
         add_custom_fields(xml, options) if options[:custom_fields]
         add_stored_credential(xml, options) if options[:stored_credential]
+        recurring_transaction(xml, options) if options[:stored_credential] && options.dig(:stored_credential, :reason_type) == 'recurring'
       end
 
       def add_custom_fields(xml, options)
@@ -367,7 +368,16 @@ module ActiveMerchant #:nodoc:
 
       def merchant_initiated_unscheduled(options)
         return options[:merchant_initiated_unscheduled] if options[:merchant_initiated_unscheduled]
-        return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled'
+        return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled' || options.dig(:stored_credential, :reason_type) == 'recurring'
+      end
+
+      def recurring_transaction(xml, options)
+        init_txn = options.dig(:stored_credential, :initial_transaction)
+        if init_txn == true
+          xml.ssl_add_token     'Y'
+        else
+          xml.ssl_add_token     'N'
+        end
       end
 
       def entry_mode(options)
--- a/test/remote/gateways/remote_elavon_test.rb
+++ b/test/remote/gateways/remote_elavon_test.rb
@@ -235,6 +235,42 @@ class RemoteElavonTest < Test::Unit::TestCase
     assert_success capture
   end
 
+  def test_successful_recurring_purchase
+    options = {
+      email: 'human@domain.com',
+      description: 'Test Transaction',
+      billing_address: address,
+      ip: '203.0.113.0',
+      merchant_initiated_unscheduled: 'Y'
+
+    }
+
+    stored_credential_params = {
+      initial_transaction: true,
+      reason_type: 'recurring',
+      initiator: 'merchant',
+      network_transaction_id: nil
+    }
+
+    assert first_purchase = @gateway.purchase(@amount, @credit_card, options.merge({ stored_credential: stored_credential_params }))
+
+    assert_success first_purchase
+    assert_equal 'APPROVAL', first_purchase.message
+
+    second_sc_params = {
+      initial_transaction: false,
+      reason_type: 'recurring',
+      initiator: 'merchant',
+      network_transaction_id: first_purchase.network_transaction_id
+    }
+
+    assert next_purchase = @gateway.purchase(@amount, @credit_card, options.merge({ stored_credential: second_sc_params }))
+    assert next_purchase.authorization
+
+    assert_success next_purchase
+    assert_equal 'APPROVAL', next_purchase.message
+  end
+
   def test_successful_auth_and_capture_with_unscheduled_stored_credential
     stored_credential_params = {
       initial_transaction: true,
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
 == HEAD
 * Paysafe: Add gateway integration [meagabeth] #4085
+* Elavon: Support recurring transactions with stored credentials [cdmackeyfree] #4086
 
 == Version 1.122.0 (August 3rd, 2021)
 * Orbital: Correct success logic for refund [tatsianaclifton] #4014
--- a/test/unit/gateways/elavon_test.rb
+++ b/test/unit/gateways/elavon_test.rb
@@ -32,6 +32,13 @@ class ElavonTest < Test::Unit::TestCase
       billing_address: address,
       description: 'Store Purchase'
     }
+
+    @options2 = {
+      order_id: '2',
+      billing_address: address,
+      description: 'Store Purchase',
+      merchant_initiated_unscheduled: 'Y'
+    }
   end
 
   def test_successful_purchase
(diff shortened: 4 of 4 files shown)


lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -296,11 +296,14 @@ def add_ip(xml, options)
...
         xml.ssl_dynamic_dba                     options[:dba] if options.has_key?(:dba)
         xml.ssl_merchant_initiated_unscheduled  merchant_initiated_unscheduled(options) if merchant_initiated_unscheduled(options)
+        xml.ssl_get_token                       recurring_txn(options) if recurring_txn(options)
+        xml.ssl_add_token                       recurring_txn(options) if recurring_txn(options)
         xml.ssl_customer_code                   options[:customer] if options.has_key?(:customer)
         xml.ssl_customer_number                 options[:customer_number] if options.has_key?(:customer_number)
         xml.ssl_entry_mode                      entry_mode(options) if entry_mode(options)
         add_custom_fields(xml, options) if options[:custom_fields]
         add_stored_credential(xml, options) if options[:stored_credential]
+        recurring_txn(options) if options
```
naashton, line 306: Was this left in accidentally?

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -370,6 +373,15 @@ def merchant_initiated_unscheduled(options)
         return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled'
       end
 
+      def recurring_txn(options)
```
naashton, line 376: We should spell out `transaction` here, just to be consistent with the rest of the code.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -370,6 +373,15 @@ def merchant_initiated_unscheduled(options)
...
+        return options[:merchant_initiated_unscheduled] if options[:merchant_initiated_unscheduled]
+        return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'recurring'
+        return options[:ssl_get_token] if options[:ssl_get_token]
+        return 'Y' if options.dig(:stored_credential, :reason_type) == 'recurring' && options.dig(:store_credential, :initial_transaction) == true
+        return options[:ssl_add_token] if options[:ssl_add_token]
+        return 'Y' if options.dig(:stored_credential, :reason_type) == 'recurring' && options.dig(:store_credential, :initial_transaction) == true
```
naashton, line 382: I think this last `return` here is a duplicate.

I do think we can clean this up a little. It looks like the value we want to return will be 'Y' or we don't want to do anything. We can combine `377-381` into one conditional check, perhaps like this:

```ruby
  if options.dig(:stored_credential, :reason_type) == 'recurring' && (options.dig(:stored_credential, :initial_transaction) == true || options.dig(:stored_credential, :initiator) == 'merchant')
      return 'Y'
  elsif [options.dig(:merchant_initiated_unscheduled), options.dig(:ssl_get_token), options.dig(:ssl_add_token)].include?('Y')
      return 'Y'
  end
```

For the first check, if we know that `reason_type` is `recurring` AND one of the other 2 conditions is "true", then we can return 'Y'. Else, check if one of the other 3 fields (`ssl_add_token`, `ssl_get_token`, or `merchant_initiated_unscheduled`) includes 'Y' (which I assume is what we expect to be passed there).

I know these stored credential fields can get a bit hairy, but I think this is a bit more readable. Feel free to take it and play around with it, I only tested it against your test without trying to throw a wrench into it.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -370,6 +372,15 @@ def merchant_initiated_unscheduled(options)
         return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled'
       end
 
+      def recurring_transaction(options)
+        if options.dig(:stored_credential, :reason_type) == 'recurring' && options.dig(:store_credential, :initial_transaction) == true
+          return options[:ssl_get_token] if options[:ssl_get_token] && options[:ssl_add_token] if options[:ssl_add_token] && options[:merchant_initiated_unscheduled] if options[:merchant_initiated_unscheduled]
```
naashton, line 377: So, my question for these is, would they be any other value other than 'Y'?

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -296,6 +296,8 @@ def add_ip(xml, options)
       def add_auth_purchase_params(xml, options)
         xml.ssl_dynamic_dba                     options[:dba] if options.has_key?(:dba)
         xml.ssl_merchant_initiated_unscheduled  merchant_initiated_unscheduled(options) if merchant_initiated_unscheduled(options)
+        xml.ssl_get_token                       recurring_transaction(options) if recurring_transaction(options)
+        xml.ssl_add_token                       recurring_transaction(options) if recurring_transaction(options)
```
naashton, line 300: I couldn't find `ssl_get_token` in the Zendesk tickets, but I think this is where my hang up is. I was just assuming it would be the same as the `ssl_add_token`.

lib/active_merchant/billing/gateways/elavon.rb
naashton: And do we still need to check for this field, `options.dig(:stored_credential, :initiator) == 'merchant'`?

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -370,6 +371,15 @@ def merchant_initiated_unscheduled(options)
         return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled'
       end
 
+      def recurring_transaction(options)
+        if options.dig(:stored_credential, :reason_type) == 'recurring' && options.dig(:store_credential, :initial_transaction) == true
+          return options[:ssl_add_token] if options[:ssl_add_token] && options[:merchant_initiated_unscheduled] if options[:merchant_initiated_unscheduled]
```
naashton, line 376: This is throwing a linter error for me. I also don't think you want this nested here. Perhaps individually add a return above the condition for each of these fields.

```ruby
return options[:ssl_add_token] if options[:ssl_add_token]
return options[:merchant_initiated_unscheduled] if options[:merchant_initiated_unscheduled]
if 
  ...
end
```

test/remote/gateways/remote_elavon_test.rb
```diff
@@ -226,6 +226,7 @@ def test_successful_auth_and_capture_with_recurring_stored_credential
       reason_type: 'recurring',
       initiator: 'merchant',
       network_transaction_id: auth.network_transaction_id
+
```
naashton, line 229: Didn't notice this empty line in here before.

lib/active_merchant/billing/gateways/elavon.rb
cdmackeyfree: For posterity since we chatted outside of the ticket, the ssl_get_token was requested specifically in the Jira ticket, but the ssl_add_token performs that function as well as store so it was removed.

lib/active_merchant/billing/gateways/elavon.rb
cdmackeyfree: Weirdly, rubocop didn't catch this in rake test:local for me, but did when I ran it separately. ¯\_(ツ)_/¯

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -367,7 +368,12 @@ def add_stored_credential(xml, options)
 
       def merchant_initiated_unscheduled(options)
         return options[:merchant_initiated_unscheduled] if options[:merchant_initiated_unscheduled]
-        return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled'
+        return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled' || options.dig(:stored_credential, :reason_type) == 'recurring'
```
naashton, line 371: This looks right, but it seems unrelated to the `ssl_add_token` field. Just want to make sure this should be changed.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -367,7 +368,12 @@ def add_stored_credential(xml, options)
...
+      end
+
+      def recurring_transaction(options)
+        return options[:ssl_add_token] if options.dig(:stored_credential, :reason_type) == 'recurring' && options.dig(:store_credential, :initial_transaction) == true
```
naashton, line 375: I think this will produce an error if `options[:ssl_add_token]` isn't set.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -367,7 +368,12 @@ def add_stored_credential(xml, options)
...
+        return 'Y'
```
naashton, line 376: So, if the condition above isn't met, this method will always return 'Y'. Is that what we want?

lib/active_merchant/billing/gateways/elavon.rb
cdmackeyfree: Since recurring transactions always need the merchant_intiated_unscheduled field but only the initial recurring transaction needs ssl_add_token, upon further reflection it made more sense for that check to live here.

lib/active_merchant/billing/gateways/elavon.rb
naashton: Never mind, I see now that you're returning `false` if that condition is met. So, no error.

cdmackeyfree: I'm not seeing any failures from calls without it being set from Elavon currently. Other tests are sending without this parameter being set and it isn't causing an error.

naashton: Ok, maybe this won't lead to an error, but if `options[:ssl_add_token]` isn't set, it will just return nil/empty. The gateway seems to handle it, but it might be the case if the user doesn't pass this value, the transaction might not behave how they want/expect.

lib/active_merchant/billing/gateways/elavon.rb
cdmackeyfree: There must be something amiss in the code because as it is a 'Y' is sent no matter what as long as it is a recurring purchase.  This doesn't present an issue because if it is sent on a subsequent call it just registers it as an update without changing the token. But I am tinkering with it now.

lib/active_merchant/billing/gateways/elavon.rb
cdmackeyfree: The field will not send it unless they are sending it in as a recurring, anything else will not include the field at all. With my latest commit it will either send a 'Y' or 'N' depending on if it is an initial transaction or not, both are accepted values for the gateway. In my prior code, it would only send add_ssl_token = 'Y' no matter what kind of recurring transaction it was, it did not send for non- recurring at all.

lib/active_merchant/billing/gateways/elavon.rb
cdmackeyfree: Conditional added plus a new test and some clean up!

naashton (approved):
Thanks for enduring this review process! 🚢

PR #4088, opened 2021-08-23
Credorax: Determine ISK decimal by datetime

curiousepic:
Credorax is changing the decimal value for the ISK currency on August
31 2021 8:00 UTC. This determines how the amount should be calculated
by checking against this deadline.

Their sandbox already considers ISK zero-decimal, so it is difficult to
test a behavior that is dependent on whether you're running transactions
against the sandbox or a live endpoint, so I have emulated the change
on the sandbox by setting a date in the past, in the same way as the
future deadline is set for Live.

Remote:
45 tests, 164 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
77 tests, 368 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -20,8 +20,9 @@ module ActiveMerchant #:nodoc:
       self.live_url = 'https://assigned-subdomain.credorax.net/crax_gate/service/gateway'
 
       self.supported_countries = %w(AD AT BE BG HR CY CZ DK EE FR DE GI GR GG HU IS IE IM IT JE LV LI LT LU MT MC NO PL PT RO SM SK ES SE CH GB)
+
       self.default_currency = 'EUR'
-      self.currencies_without_fractions = %w(BIF CLP DJF GNF JPY KMF KRW PYG RWF VND VUV XAF XOF XPF)
+      self.currencies_without_fractions = %w(BIF CLP DJF GNF ISK JPY KMF KRW PYG RWF VND VUV XAF XOF XPF)
       self.currencies_with_three_decimal_places = %w(BHD IQD JOD KWD LYD OMR TND)
 
       self.money_format = :cents
@@ -120,6 +121,10 @@ module ActiveMerchant #:nodoc:
         'R3' => 'Revocation of all Authorisations Order',
         '1A' => 'Strong Customer Authentication required'
       }
+
+      # Can be removed when no longer used after August 31 2021 8:00 UTC
+      LIVE_ISK_CHANGE_TIME = 1630396800 # August 31 2021 8:00 UTC
+      TEST_ISK_CHANGE_TIME = 1629619200 # August 22 2021 8:00 UTC
 
       def initialize(options = {})
         requires!(options, :merchant_id, :cipher_key)
@@ -253,10 +258,28 @@ module ActiveMerchant #:nodoc:
 
       private
 
+      # Can be removed after 2021 August 31 8:01 UTC
+      def current_time(options)
+        options[:current_time_test_value] || Time.now.utc
+      end
+
+      # Can be removed after 2021 August 31 8:01 UTC
+      def isk_change_check(money, options)
+        isk_change_time = test? ? TEST_ISK_CHANGE_TIME : LIVE_ISK_CHANGE_TIME
+        amount = amount(money)
+
+        if current_time(options).to_i >= isk_change_time
+          return sprintf('%.0f', amount.to_f / 100)
+        else
+          return amount
+        end
+      end
+
       def add_invoice(post, money, options)
         currency = options[:currency] || currency(money)
 
-        post[:a4] = localized_amount(money, currency)
+        # ISK change check can be removed after 2021 August 31 8:01 UTC
+        post[:a4] = currency == 'ISK' ? isk_change_check(money, options) : localized_amount(money, currency)
         post[:a1] = generate_unique_id
         post[:a5] = currency
         post[:h9] = options[:order_id]
--- a/test/unit/gateways/credorax_test.rb
+++ b/test/unit/gateways/credorax_test.rb
@@ -980,6 +980,41 @@ class CredoraxTest < Test::Unit::TestCase
     end.respond_with(successful_authorize_response)
   end
 
+  # Can be removed after August 31 8:00 UTC
+  def test_isk_passed_as_nonfractional_on_test
+    stub_comms do
+      @gateway.authorize(200, @credit_card, @options.merge(currency: 'ISK'))
+    end.check_request do |_endpoint, data, _headers|
+      assert_match(/a4=2&a1=/, data)
+    end.respond_with(successful_authorize_response)
+  end
+
+  # Can be removed after August 31 8:00 UTC
+  def test_isk_passed_as_two_decimal_on_live_before_august_31
+    ActiveMerchant::Billing::Base.mode = :production
+
+    stub_comms do
+      @gateway.authorize(200, @credit_card, @options.merge(currency: 'ISK', current_time_test_value: 1630396799)) # August 31 7:59:59 UTC
+    end.check_request do |_endpoint, data, _headers|
+      assert_match(/a4=200&a1=/, data)
+    end.respond_with(successful_authorize_response)
+  ensure
+    ActiveMerchant::Billing::Base.mode = :test
+  end
+
+  # Can be removed after August 31 8:00 UTC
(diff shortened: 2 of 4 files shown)


curiousepic:
I would appreciate scrutiny around the methods used for the date-times and their comparison.

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -253,10 +254,27 @@ def add_3ds_2_optional_fields(post, options)
 
       private
 
+      def isk_change_check(money)
+        if test?
```
Carrigan, line 258: These timestamps look good and explicitly set it to UTC 👍

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -253,10 +254,27 @@ def add_3ds_2_optional_fields(post, options)
...
+          isk_change_time = Time.new(2021, 8, 22, 8, 0, 0, '+00:00')
```
esmitperez, line 259: Is there a "cost" of instantiating the same static date for every method call? If so, could these 2 values just be frozen and used instead?

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -253,10 +254,27 @@ def add_3ds_2_optional_fields(post, options)
...
+        else
+          isk_change_time = Time.new(2021, 8, 31, 8, 0, 0, '+00:00')
+        end
+
+        amount = amount(money)
+
+        if Time.now.utc >= isk_change_time
+          return sprintf('%.0f', amount.to_f / 100)
```
Carrigan, line 267: This formatting looks good for amounts in `cents`, but for non-cents users of the library, this does not retain all of the logic in `localized_amount`:

https://github.com/activemerchant/active_merchant/blob/b353ff30848206540dac348420a661ca08ff377c/lib/active_merchant/billing/gateway.rb#L284-L288

I'd recommend either:
- Adding support for non-cents users of the library here, or
- It would be a bit of a refactor, but we can avoid rewriting a new formatter and leverage existing code better by overwriting the `non_fractional_currency?` method for this gateway. It currently reads:

https://github.com/activemerchant/active_merchant/blob/b353ff30848206540dac348420a661ca08ff377c/lib/active_merchant/billing/gateway.rb#L270-L272

It could simplify this change a lot to add `|| (currency is ISK AND time > isk_change_time)` to this function rather than creating a new amount formatter.

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -253,10 +254,27 @@ def add_3ds_2_optional_fields(post, options)
...
+          return amount
+        end
+      end
+
       def add_invoice(post, money, options)
         currency = options[:currency] || currency(money)
 
-        post[:a4] = localized_amount(money, currency)
+        # ISK check can be removed after 2021 August 31 8:01 UTC
+        post[:a4] = currency == 'ISK' ? isk_change_check(money) : localized_amount(money, currency)
```
esmitperez, line 277: `localized_amount` seems to simply return the `amount` when `non_fractional_currency?`, which the addition of `ISK` to `currencies_without_fractions` seems to trigger.

Could the following work?

```ruby
money = isk_change_check(money) if currency == 'ISK'
post[:a4] = localized_amount(money, currency)
```

esmitperez (commented):
LGTM, but have a couple comments/suggestions

test/unit/gateways/credorax_test.rb
```diff
@@ -980,6 +980,16 @@ def test_nonfractional_currency_handling
     end.respond_with(successful_authorize_response)
   end
 
+  # Utilizes similar conditional as the Live deadline of August 31
```
Carrigan, line 983: Would it be possible to add a test that simulates a timestamp after the August 31st deadline to confirm that the new logic switches over? I don't believe that minitest has any methods for doing this easily like rspec does, so it might mean that we need to use dependency injection. Instead of using `Time.now.utc` directly in code, we could make it its own function that can have a value injected:

```
attr_accessor :current_time_test_value

def current_time
  @current_time_test_value || Time.now.utc
end
```

Carrigan (commented):
Nice! I think the only blocker here is not supporting non-cents users of the library with the `isk_change_check` function. I think a test that simulates a timestamp after August 31st, 8AM UTC would be really valuable though, and there is an opportunity to simplify the logic change here as well.

lib/active_merchant/billing/gateways/credorax.rb
curiousepic: Possibly - I think that's a good idea.

lib/active_merchant/billing/gateways/credorax.rb
curiousepic: I'm not sure I understand your concern, but the idea here was that ISK amounts won't get to `localized_amount` until the check is removed, when the updated list will be current. I think your suggestion could also work though, and separates the check even more cleanly. I'll try it out.

test/unit/gateways/credorax_test.rb
curiousepic: Nice - I had an inkling something like this was possible, but didn't quite coalesce. I'll implement this.

lib/active_merchant/billing/gateways/credorax.rb
esmitperez: just wanted to approximate an implementation where we keep the original code, and then just change what needs to change, but both work and once the cut-off date is reached, it won't matter :)

test/unit/gateways/credorax_test.rb
curiousepic: @Carrigan On consideration, I'm not sure this would actually test anything, since we cannot test the Live case, and the sandbox has already implemented the change. However, simulating a _past_ date for the sandbox would actually be helpful. Does that make sense, or am I missing something?

lib/active_merchant/billing/gateways/credorax.rb
curiousepic: Do you think Esmit's suggestion for freezing them (to the integer form) would be preferred?

lib/active_merchant/billing/gateways/credorax.rb
curiousepic: Actually I like Brian's simpler suggestion for overriding `non_fractional_currency?` and will probably go with that instead.

lib/active_merchant/billing/gateways/credorax.rb
Carrigan: Ahh, I didn't realize that the gateway itself determines whether it is set to `cents`, I thought it was the caller. In that case the logic is good as is - your call on if you want to switch implementation or not.

test/unit/gateways/credorax_test.rb
Carrigan: My thinking was that for a unit test, we could set `current_time_test_value` to `August 31, 8:01AM UTC` and then check `assert_match(/a4=200&a1=/, data)`. For the sandbox, I am not sure how valuable it is to test a past date since there will never be a case where we make transactions before 8/22/21 again. Does that make sense?

curiousepic: @Carrigan Hm, let me rephrase, sorry if I'm still misunderstanding. The problem is that we cannot simulate the Live switchover date, because `test?` is always true for AM unit tests (it would be nice if we could change this, but I don't know how if it's possible). So we can only test the test/sandbox case. Since the sandbox already switched, the current test is already testing the "after the switch" case.

But adding a test that sets the current time prior to Aug 22 would have parity to the real-current-date for the Live case.

curiousepic: Actually I found a way to change the mode from the test. That's probably the best way to go here, along with your suggestion for overwriting the current time.

lib/active_merchant/billing/gateways/credorax.rb
curiousepic: Overwriting `non_fractional_currency?` proved complex in combination with allowing a test time value, which I think is more beneficial, so I'm going to leave it similar to this initial pattern.

curiousepic:
Thanks for the swift and thorough feedback @esmitperez @Carrigan - I've implemented most of the suggestions if you could take another look.

Carrigan (approved):
LGTM :+1: I left one suggestion to keep the test suite rolling until we can deploy a permanent change after August 31st, but there are no blockers. Thanks for adding the tests that show the time based logic switch working! 🎉

PR #4089, opened 2021-08-24
Updates Elavon recurring transactions

cdmackeyfree:
As the initial_transaction parameter is triggered in the response but not generally sent to active merchant, updated to be sent in as its own field, which will only be sent if included in the request.

Unit:
47 tests, 245 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
38 tests, 177 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.3684% passed


lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -371,13 +371,9 @@ def merchant_initiated_unscheduled(options)
...
       end
 
-      def recurring_transaction(xml, options)
-        init_txn = options.dig(:stored_credential, :initial_transaction)
-        if init_txn == true
-          xml.ssl_add_token     'Y'
-        else
-          xml.ssl_add_token     'N'
-        end
+      def add_recurring_token(options)
```
therufs, line 374: It looks like we don't actually call this method as part of `add_auth_purchase_params` -- is this intentional?

cdmackeyfree: Ah no, that was just missed when I was cleaning up (didn't get added before pushing oops) corrected version up now.,

PR #4090, opened 2021-08-24
Adyen: Upgrade API Version

almalee24:
Upgrade the Payment API Version to v64 and the Recurring
API Version to v49.

103 tests, 399 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

77 tests, 402 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@
 * Maestro: support BINs without Luhn check [therufs] #4097
 * Maestro: support BINs [therufs] #4098
 * Redsys: Route MIT Exemptions to webservice endpoint [curiousepic] #4081
+* Adyen: Update Classic Integration API to v64 and Recurring API to v49 [almalee24] #4090
 
 == Version 1.122.0 (August 3rd, 2021)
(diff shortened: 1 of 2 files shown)


almalee24:
According to Adyen
"Regarding the two versions mentioned, in fact the main breaking change between the two is the new endpoint. The way it is built the endpoint are compatible backwards, in other words, the endpoints should be compatible with previous versions however we always advise to test them in the test enviroment."

curiousepic:
@almalee24 Sounds good- but I'm confused about the "new endpoint", does that not mean we need to change the urls?

almalee24:
> @almalee24 Sounds good- but I'm confused about the "new endpoint", does that not mean we need to change the urls? Or are they just referring to the version number in the endpoint url?

Right I should've clarified that. Basically, v64 has an endpoint, `/getAuthenticationResult`, that v40 didn't have.

curiousepic (approved):
I see 👍

PR #4103, opened 2021-09-03
Pull UnionPay's 62* BIN ranges out of Discover's

leahriffell:
China UnionPay cards beginning with 62 are able to run on Discover rails in the US and Canada. Because of this, these cards were previously lumped together with Discover. There is no change to how these transactions will be processed, but we would like to break them out in order to identify UnionPay cards correctly. 

Discover's [BIN list](https://www.discoverglobalnetwork.com/content/dam/discover/en_us/dgn/pdfs/IPP-VAR-Enabler-Compliance.pdf) was used in order to understand specific BIN ranges belonging to UnionPay.

____

**Unit tests:**
4897 tests, 74184 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -7,7 +7,7 @@ module ActiveMerchant #:nodoc:
         'master'             => ->(num) { num&.size == 16 && in_bin_range?(num.slice(0, 6), MASTERCARD_RANGES) },
         'elo'                => ->(num) { num&.size == 16 && in_bin_range?(num.slice(0, 6), ELO_RANGES) },
         'alelo'              => ->(num) { num&.size == 16 && in_bin_range?(num.slice(0, 6), ALELO_RANGES) },
-        'discover'           => ->(num) { num =~ /^(6011|65\d{2}|64[4-9]\d)\d{12,15}|(62\d{14,17})$/ },
+        'discover'           => ->(num) { num =~ /^(6011|65\d{2}|64[4-9]\d)\d{12,15}$/ },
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -196,11 +196,10 @@ module CreditCardMethods
         589562..589562
       ]
 
-      # In addition to the BIN ranges listed here that all begin with 81, UnionPay cards
-      # include many ranges that start with 62.
-      # Prior to adding UnionPay, cards that start with 62 were all classified as Discover.
-      # Because UnionPay cards are able to run on Discover rails, this was kept the same.
+      # UnionPay cards beginning with 62 are able to run on Discover rails in the US and Canada
```
bdewater, line 199: I would remove this comment (and also adjust the tests below). The doc linked doesn't support the claim it's only the 62* range - it also mentions the 81* one :) leaving the doc itself as a source is fine.

We don't make this sort of comment for JCB either - which as discussed on Slack has a more complicated acceptance story (Discover rails in the US, Amex in Canada, Australia, New Zealand, native in Japan). The "card brand in country using other network"-concept is something that is not a defined thing in Active Merchant at the moment. That could be a useful addition but for now I would recommend to just call a UnionPay card a UnionPay card.

Nitpicking aside - code looks good!

c-bacon:
Great work! 💯

c-bacon (approved):

lib/active_merchant/billing/credit_card_methods.rb
leahriffell: @bdewater that makes sense to me! I've removed this comment and consolidated the tests.

PR #4105, opened 2021-09-07
Adyen: Add ACH Support

almalee24:
Adding ACH/Bank Accounts to Adyen.

Remote tests:
114 tests, 433 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit tests:
83 tests, 434 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -82,12 +82,13 @@ module ActiveMerchant #:nodoc:
       end
 
       def credit(money, payment, options = {})
+        action = 'refundWithData'
         post = init_post(options)
         add_invoice(post, money, options)
-        add_payment(post, payment, options)
+        add_payment(post, payment, options, action)
         add_shopper_reference(post, options)
         add_network_transaction_reference(post, options)
-        commit('refundWithData', post, options)
+        commit(action, post, options)
       end
 
       def void(authorization, options = {})
@@ -159,9 +160,12 @@ module ActiveMerchant #:nodoc:
       def scrub(transcript)
         transcript.
           gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
-          gsub(%r(("number\\?":\\?")[^"]*)i, '\1[FILTERED]').
-          gsub(%r(("cvc\\?":\\?")[^"]*)i, '\1[FILTERED]').
-          gsub(%r(("cavv\\?":\\?")[^"]*)i, '\1[FILTERED]')
+          gsub(%r(("number\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("cvc\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("cavv\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("bankLocationId\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("iban\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("bankAccountNumber\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
       end
 
       private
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -157,7 +157,9 @@ def scrub(transcript)
           gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
           gsub(%r(("number\\?":\\?")[^"]*)i, '\1[FILTERED]').
           gsub(%r(("cvc\\?":\\?")[^"]*)i, '\1[FILTERED]').
-          gsub(%r(("cavv\\?":\\?")[^"]*)i, '\1[FILTERED]')
+          gsub(%r(("cavv\\?":\\?")[^"]*)i, '\1[FILTERED]').
```
esmitperez, line 160: I know you followed the existing pattern, but this is a good opportunity to tighten up these RegExp's even more.  

Currently these do NOT get filtered :/ (notice the space between the colons).

'"cvc" : "addafd", "bankLocationId" : "aaa"'

A potential way to prevent this is to surround the `:` with `\s*` -> `gsub(%r(("cvc\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')` etc.

esmitperez (changes requested):
Have an enhancement request for RE's.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -372,12 +374,26 @@ def add_payment(post, payment, options)
           _, _, recurring_detail_reference = payment.split('#')
           post[:selectedRecurringDetailReference] = recurring_detail_reference
           options[:recurring_contract_type] ||= 'RECURRING'
+        elsif payment.is_a?(Check)
+          add_bank_account(post, payment, options[:billing_address][:country])
```
curiousepic, line 378: This is a very minor thing, but usually we'll pass the entire options hash into `add_` methods like this.

curiousepic (approved):
This looks really good, great work! I like the parity with `add_card` by checking `requires!`. Made one small suggestion, and I see Esmit's as well, but otherwise this is good to go.

almalee24:
@curiousepic I just want to confirm that the credit method in Adyen is only for credit cards?

curiousepic:
@almalee24 The credit method is not a normal refund, but a "general credit", aka "standalone refund", where there is no original transaction associated. So there is no payment method involved. Does that answer your question?

curiousepic:
@almalee24 I don't know. It's possible that a credit could be sent to a bank account. You'd need to check in the gatway docs. FYI Spreedly refers to these as "general credit", and I don't think there is any restriction on payment method type.

esmitperez (approved):
👍

almalee24:
> @almalee24 I don't know. It's possible that a credit could be sent to a bank account. You'd need to check in the gateway docs. FYI Spreedly refers to these as "general credit" (as distinct from refund), and I don't think there is any restriction on payment method type.

Alright, I get an error when testing credit with bank account. I'm going to wait to merge this until I clear this with Adyen on why this is failing. Can't find anything in their docs.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -367,17 +372,32 @@ def add_invoice_for_modification(post, money, options)
...
       end
 
+      def add_bank_account(post, bank_account, options, action)
+        bank = {
+          bankAccountNumber: bank_account.account_number,
+          ownerName: bank_account.name,
+          countryCode: options[:billing_address][:country]
+        }
+
+        action == 'refundWithData' ? bank[:iban] = bank_account.routing_number : bank[:bankLocationId] = bank_account.routing_number
```
curiousepic, line 395: weird, but ok!

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -663,6 +710,16 @@ def test_failed_refund
 
   def test_successful_credit
     response = @gateway.credit(@amount, @credit_card, @options)
+
+    assert_success response
+    assert_equal 'Received', response.message
+  end
+
+  def test_successful_credit_with_bank_account
```
curiousepic, line 718: If we want to test that the correct field name is used, you can utilize `stub_comms` and `.check_request`,  ala the `test_installments_sent` test. This will help catch any regressions for that special logic.

curiousepic:
Looking good, great catch on the support for general credit. Looks like you don't have a unit test for that special behavior though, unless I'm missing something. Good use case for the `stub_comms` and `.check_request` pattern, to help catch any regressions for that special logic.

almalee24:
> Looking good, great catch on the support for general credit. Looks like you don't have a unit test for that special behavior though, unless I'm missing something. Good use case for the `stub_comms` and `.check_request` pattern, to help catch any regressions for that special logic.

I do, its called `test_successful_credit_with_bank_account`

curiousepic:
@almalee24 Strange, I'm not seeing it in the unit tests.

almalee24:
@curiousepic It's only on the remote test, completely forgot to add it to the unit test.

PR #4112, opened 2021-09-15
Paysafe: Adjust profile data

meagabeth:
Adjust logic for profile data so the phone field is being successfully passed in the request whether it is included in the billing address hash within the options hash or if it an independent field included in the options hash

Rubocop:
713 files inspected, no offenses detected
Local:
4906 tests, 74228 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
29 tests, 70 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
13 tests, 45 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paysafe.rb
+++ b/lib/active_merchant/billing/gateways/paysafe.rb
@@ -119,9 +119,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_billing_address(post, options)
-        return unless options[:billing_address] || options[:address]
+        return unless address = options[:billing_address] || options[:address]
 
-        address = options[:billing_address] || options[:address]
         post[:billingDetails] = {}
         post[:billingDetails][:street] = address[:address1]
         post[:billingDetails][:city] = address[:city]
@@ -134,9 +133,8 @@ module ActiveMerchant #:nodoc:
       # The add_address_for_vaulting method is applicable to the store method, as the APIs address
       # object is formatted differently from the standard transaction billing address
       def add_address_for_vaulting(post, options)
-        return unless options[:billing_address || options[:address]]
+        return unless address = options[:billing_address] || options[:address]
 
-        address = options[:billing_address] || options[:address]
         post[:billingAddress] = {}
         post[:billingAddress][:street] = address[:address1]
         post[:billingAddress][:city] = address[:city]
@@ -147,8 +145,6 @@ module ActiveMerchant #:nodoc:
 
       # This data is specific to creating a profile at the gateway's vault level
       def add_profile_data(post, payment, options)
-        address = options[:billing_address] || options[:address]
-
         post[:firstName] = payment.first_name
         post[:lastName] = payment.last_name
         post[:dateOfBirth] = {}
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -147,17 +145,20 @@ def add_address_for_vaulting(post, options)
 
       # This data is specific to creating a profile at the gateway's vault level
       def add_profile_data(post, payment, options)
-        address = options[:billing_address] || options[:address]
-
         post[:firstName] = payment.first_name
         post[:lastName] = payment.last_name
         post[:dateOfBirth] = {}
         post[:dateOfBirth][:year] = options[:date_of_birth][:year]
```
dsmcclain, line 151: I know this isn't exactly germane to your `phone` changes, but I'm wondering if we need to protect against scenarios where `options` doesn't have a `date_of_birth` object here?

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -147,17 +145,20 @@ def add_address_for_vaulting(post, options)
...
         post[:dateOfBirth][:day] = options[:date_of_birth][:day]
         post[:email] = options[:email] if options[:email]
-        post[:phone] = (address[:phone] || options[:phone]) if address[:phone] || options[:phone]
         post[:ip] = options[:ip] if options[:ip]
+
+        if options[:phone]
+          post[:phone] = options[:phone]
+        elsif address = options[:billing_address] || options[:address]
+          post[:phone] = address[:phone] if address[:phone]
+        end
```
dsmcclain, line 161: Just throwing this out there if you like the style 😃 

in `def add_profile_data`:
```
phone = phone(options)
post[:phone] = phone if phone
```
using the helper:
```
def phone(options)
  options[:phone] || options.dig(:billing_address, :phone) || options.dig(:address, :phone)
end
```

test/unit/gateways/paysafe_test.rb
```diff
@@ -121,6 +131,16 @@ def test_successful_verify
     assert_equal '493936', response.params['authCode']
   end
 
+  def test_successful_store
+    @gateway.expects(:ssl_request).returns(successful_store_response)
+
+    response = @gateway.store(@credit_card, @more_options)
+    assert_success response
+
+    assert_equal '111-222-3456', response.params['phone']
```
dsmcclain, line 140: This test isn't really covering your code, since you are providing the `successful_store_response` when the request is made. If you remove `phone` from `@more_options` it still passes. I'm not sure if that was the intention behind this test, so my apologies if I am misreading it.

You could do a `check_request`-style test to verify that a `phone` being passed in at the `options` or `options[:billing_address]` level makes it into the post.

dsmcclain (approved):
I made one comment about something I noticed that isn't exactly related to this PR.

I also made a suggestion on adding a test that checks that `phone` gets passed in to the request correctly, and a separate suggestion for styling changes. Both are optional in my mind, since the changes you've made seem to be working just fine. I'll leave it up to your judgement. You are welcome to tag me for re-review if you do make more changes, otherwise it is :approved: 👍

PR #4116, opened 2021-09-21
dlocal: Add device_id and ip to payer object and add additional_data

aenand:
Added the fields device_id and ip to the payer object. dlocal's device_id is event_uuid and dlocal's ip is onboarding_ip_address. The additional_data field is a blob, anything can be sent in under that.

unit: 23 tests, 93 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

remote: 28 tests, 73 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_d_local_test.rb
+++ b/test/remote/gateways/remote_d_local_test.rb
@@ -72,6 +72,7 @@ class RemoteDLocalTest < Test::Unit::TestCase
     options = @options.merge(
       order_id: '1',
       ip: '127.0.0.1',
+      device_id: '123',
       email: 'joe@example.com',
       birth_date: '03-01-1970',
       document2: '87648987569',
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_d_local_test.rb
```diff
@@ -84,6 +85,15 @@ def test_successful_purchase_with_more_options
     assert_match 'The payment was paid', response.message
   end
 
+  def test_successful_purchase_with_additional_data
+    options = @options.merge(
+      additional_data: { item: 'socks' }
```
meagabeth, line 90: After reviewing the DLocal docs, it seems that the `additional_risk_data` object then contains other objects rather than independent fields like `item` so this test example may be a bit misleading. I'd suggest we change it to match the gateway's expectations a bit more literally. I edited it locally to the following and it still passed so that's good!
` options = @options.merge(
      additional_data: {
        submerchant: {
          name: 'socks'
        }
      }
    )`

test/unit/gateways/d_local_test.rb
```diff
@@ -44,6 +44,16 @@ def test_purchase_with_installments
     end.respond_with(successful_purchase_response)
   end
 
+  def test_successful_purchase_with_additional_data
+    additional_data = { 'item' => 'socks' }
```
meagabeth, line 48: same as above comment for the remote test

meagabeth (changes requested):
Great work overall! Had a suggestion regarding 'realness' in our test examples :)

meagabeth (approved):
Great work! You're good to go! 🎉

PR #4122, opened 2021-09-24
Add UnionPay test card to BIN range

leahriffell:
Revises [Pull UnionPay's 62* BIN ranges out of Discover's #4103](https://github.com/activemerchant/active_merchant/pull/4103)

I ran into issues while testing payment processing using [Stripe's test UnionPay card](https://stripe.com/docs/testing) (6200000000000005) as it wasn't previously included in ActiveMerchant's UnionPay's BIN range. This PR simply adds it to the existing BIN range.

I also added some additional tests for Vr and Carnet to ensure that their 627* range cards are getting classified correctly (Elo also has 627* cards but there was already an existing test). 

------------
**Unit tests**:
4926 tests, 74323 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/credit_card_methods.rb
+++ b/lib/active_merchant/billing/credit_card_methods.rb
@@ -199,7 +199,7 @@ module ActiveMerchant #:nodoc:
 
       # https://www.discoverglobalnetwork.com/content/dam/discover/en_us/dgn/pdfs/IPP-VAR-Enabler-Compliance.pdf
       UNIONPAY_RANGES = [
-        62212600..62379699, 62400000..62699999, 62820000..62889999,
+        620000..620000, 62212600..62379699, 62400000..62699999, 62820000..62889999,
         81000000..81099999, 81100000..81319999, 81320000..81519999, 81520000..81639999, 81640000..81719999
       ]
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,7 @@
 * Stripe and Stripe PI: Add Radar Session Option [tatsianaclifton] #4119
 * PayArc: Fix billing address nil and phone_number issues [dsmcclain] #4114
 * Routex: Update BIN numbers [rachelkirk] #4123
+* UnionPay: Add Stripe's UnionPay test card to UnionPay BIN range #4122
 
 == Version 1.123.0 (September 10th, 2021)
 * Paysafe: Add gateway integration [meagabeth] #4085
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -199,7 +199,7 @@ module CreditCardMethods
 
       # https://www.discoverglobalnetwork.com/content/dam/discover/en_us/dgn/pdfs/IPP-VAR-Enabler-Compliance.pdf
       UNIONPAY_RANGES = [
-        62212600..62379699, 62400000..62699999, 62820000..62889999,
+        620000, 62212600..62379699, 62400000..62699999, 62820000..62889999,
         81000000..81099999, 81100000..81319999, 81320000..81519999, 81520000..81639999, 81640000..81719999
       ]
```
leahriffell, line 205: Worked with @c-bacon to come up with an alternative to adding the test card in the BIN range here (modifying `brand?`) but after chatting through the options, we determined it wasn't worth it as CUP is the only test card causing issues right now.

Here's what we worked on and decided not to add in:


```
      # https://help.shopify.com/en/manual/payments/shopify-payments/testing-shopify-payments, https://stripe.com/docs/testing
      KNOWN_TEST_CARDS = {
        'american_express' => [378282246310005, 371449635398431],
        'diners_club' => [3056930009020004, 30569309025904, 36227206271667],
        'discover' => [6011111111111117, 6011000990139424],
        'jcb' => [3530111333300000, 3566002020360505],
        'master' => [5555555555554444, 2223003122003222, 5200828282828210, 5105105105105100],
        'unionpay' => [6200000000000005],
        'visa' => [4242424242424242, 4000056655665556]
      }


        def known_test_card_brand(number)
          found = KNOWN_TEST_CARDS.find {|brand, bins| bins.include?(number)}
          found ? found[0] : nil
        end


        def brand?(number)
          return 'bogus' if valid_test_mode_card_number?(number)

          return known_test_card_brand(number) if known_test_card_brand(number)

          CARD_COMPANY_DETECTORS.each do |company, func|
            return company.dup if func.call(number)
          end

          return nil
        end

```

bdewater: This is the right call, we should avoid implementing Stripe-specific test card numbers in this file as these vary by provider. See for example the test numbers for [Adyen](https://docs.adyen.com/development-resources/test-cards/test-card-numbers) and [Braintree](https://developer.paypal.com/braintree/docs/reference/general/testing/ruby#credit-card-numbers) 😄

PR #4129, opened 2021-09-28
Moka: Remove additional transaction data from subsequent calls

naashton:
Moka returns an error when `OrderTrxCode` is given on subsequent calls;
`capture`, `void`, and `refund`. Removing the method call from those
transactions should resolve the issue.

CE-1990

Unit: 16 tests, 73 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 22 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


cdmackeyfree (approved):
Good quick solution

PR #4134, opened 2021-09-29
Moka: Pass Currency on Capture transactions

dsmcclain:
Moka's default currency is 'TL'. We discovered that if you create an
authorization for a different currency (say 'USD'), you still must pass
that currency on the capture, otherwise Moka will default to 'TL' and
the response will be a `CurrencyDoesntMatch` error.

CE-1999

Rubocop:
716 files inspected, no offenses detected

Local:
4928 tests, 74333 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
Loaded suite test/remote/gateways/remote_moka_test
24 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


cdmackeyfree (approved):
Tests pass and it looks good!

PR #4137, opened 2021-10-04
SafeCharge: Account for non-fractional currencies

dsmcclain:
Safe Charge accepts at least one currency (Chilean Peso) that does not
include fractional units. If a request includes a decimal point in the
amount, the gateway will return an error. This PR leverages the `localized_amount`
method to truncate the amount to whole units.

CE-1909

Rubocop:
716 files inspected, no offenses detected

Local:
4928 tests, 74333 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
29 tests, 77 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.5517% passed
You should expect this remote test to fail for reasons not related to this PR:
`test_successful_regular_purchase_through_3ds_flow_with_invalid_pa_res`

### DIFF
--- a/lib/active_merchant/billing/gateways/safe_charge.rb
+++ b/lib/active_merchant/billing/gateways/safe_charge.rb
@@ -126,9 +126,11 @@ module ActiveMerchant #:nodoc:
       private
 
       def add_transaction_data(trans_type, post, money, options)
+        currency = options[:currency] || currency(money)
+
         post[:sg_TransType] = trans_type
-        post[:sg_Currency] = (options[:currency] || currency(money))
-        post[:sg_Amount] = amount(money)
+        post[:sg_Currency] = currency
+        post[:sg_Amount] = localized_amount(money, currency)
         post[:sg_ClientLoginID] = @options[:client_login_id]
         post[:sg_ClientPassword] = @options[:client_password]
         post[:sg_ResponseFormat] = '4'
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_safe_charge_test.rb
```diff
@@ -65,6 +65,15 @@ def test_successful_purchase
     assert_equal 'Success', response.message
   end
 
+  def test_successful_purchase_with_non_fractional_currency
+    options = @options.merge(currency: 'CLP')
+    response = @gateway.purchase(127999, @credit_card, options)
+
+    assert_success response
+    assert_equal 'Success', response.message
+    assert_equal '1279', response.params['requestedamount']
```
meagabeth, line 74: My only thought on this is why did it round down when it is 1 cent away from being `1280`, but that has nothing to do with the code that you wrote 😆

lib/active_merchant/billing/gateways/safe_charge.rb
```diff
@@ -126,9 +126,11 @@ def scrub(transcript)
       private
 
       def add_transaction_data(trans_type, post, money, options)
+        currency = options[:currency] || currency(money)
```
meagabeth, line 129: Nice!

meagabeth (approved):
Looks great - go forth! :shipit:

test/remote/gateways/remote_safe_charge_test.rb
dsmcclain: Yes, good question. I provide a description of this in the ticket for the task.

Basically, the `localized_amount` method also looks at the gateway adapter's `money_format`, and depending on whether that is set to `:cents` or not determines how it handles the extra digits.

PR #4153, opened 2021-10-11
Nuvei (SafeCharge): Add support for `email` field in capture

rachelkirk:
[CE-1992](https://spreedly.atlassian.net/browse/CE-1992)

Note: You will need to be connected to the VPN to run remote tests for this gateway.
Unit:
25 tests, 136 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
29 tests, 77 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.5517% passed
test_successful_regular_purchase_through_3ds_flow_with_invalid_pa_res(RemoteSafeChargeTest) - This test was failing before my changes.

Local:
4932 tests, 74352 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_safe_charge_test.rb
+++ b/test/remote/gateways/remote_safe_charge_test.rb
@@ -200,7 +200,7 @@ class RemoteSafeChargeTest < Test::Unit::TestCase
     auth = @gateway.authorize(@amount, @credit_card, extra)
     assert_success auth
 
-    assert capture = @gateway.capture(@amount, auth.authorization)
+    assert capture = @gateway.capture(@amount, auth.authorization, extra)
     assert_success capture
     assert_equal 'Success', capture.message
   end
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_safe_charge_test.rb
```diff
@@ -200,7 +200,7 @@ def test_successful_authorize_and_capture_with_more_options
     auth = @gateway.authorize(@amount, @credit_card, extra)
     assert_success auth
 
-    assert capture = @gateway.capture(@amount, auth.authorization)
+    assert capture = @gateway.capture(@amount, auth.authorization, extra)
     assert_success capture
     assert_equal 'Success', capture.message
```
rachelkirk, line 205: There wasn't a need to write a new remote test, adding this to the existing test should cover this field being passed. For some reason, none of the fields in this `extra` object are being mapped in the response. I paired with Mo on this and we were both stumped by this. We tried to write an integration test in core, but the test file there needs to be updated. Jimil submitted a PR for that.

naashton (approved):
🚢

PR #4166, opened 2021-10-21
Adding Priority gateway

jessiagee:
* Submits external contributor's code to Active Merchant

Loaded suite test/unit/gateways/priority_test

100% passed

100% passed

bundle exec rake test:local

100% passed

Running RuboCop...

719 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/priority.rb
+++ b/lib/active_merchant/billing/gateways/priority.rb
@@ -1,0 +1,381 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PriorityGateway < Gateway
+      # Sandbox and Production
+      self.test_url = 'https://sandbox.api.mxmerchant.com/checkout/v3/payment'
+      self.live_url = 'https://api.mxmerchant.com/checkout/v3/payment'
+
+      class_attribute :test_url_verify, :live_url_verify, :test_auth, :live_auth, :test_env_verify, :live_env_verify, :test_url_batch, :live_url_batch, :test_url_jwt, :live_url_jwt, :merchant
+
+      # Sandbox and Production - verify card
+      self.test_url_verify = 'https://sandbox-api2.mxmerchant.com/merchant/v1/bin'
+      self.live_url_verify = 'https://api2.mxmerchant.com/merchant/v1/bin'
+
+      # Sandbox and Production - check batch status
+      self.test_url_batch = 'https://sandbox.api.mxmerchant.com/checkout/v3/batch'
+      self.live_url_batch = 'https://api.mxmerchant.com/checkout/v3/batch'
+
+      # Sandbox and Production - generate jwt for verify card url
+      self.test_url_jwt = 'https://sandbox-api2.mxmerchant.com/security/v1/application/merchantId'
+      self.live_url_jwt = 'https://api2.mxmerchant.com/security/v1/application/merchantId'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
+      self.homepage_url = 'https://mxmerchant.com/'
+      self.display_name = 'Priority'
+
+      def initialize(options = {})
+        requires!(options, :merchant_id, :key, :secret)
+        super
+      end
+
+      def basic_auth
+        Base64.strict_encode64("#{@options[:key]}:#{@options[:secret]}")
+      end
+
+      def request_headers
+        {
+          'Content-Type' => 'application/json',
+          'Authorization' => "Basic #{basic_auth}"
+        }
+      end
+
+      def request_verify_headers(jwt)
+        {
+          'Authorization' => "Bearer #{jwt}"
+        }
+      end
+
+      def purchase(amount, credit_card, options = {})
+        params = {}
+        params['amount'] = localized_amount(amount.to_f, options[:currency])
+        params['authOnly'] = false
+
+        add_credit_card(params, credit_card, 'purchase', options)
+        add_type_merchant_purchase(params, @options[:merchant_id], true, options)
+        commit('purchase', params: params, jwt: options)
+      end
+
+      def authorize(amount, credit_card, options = {})
+        params = {}
+        params['amount'] = localized_amount(amount.to_f, options[:currency])
+        params['authOnly'] = true
+
+        add_credit_card(params, credit_card, 'purchase', options)
+        add_type_merchant_purchase(params, @options[:merchant_id], false, options)
+        commit('purchase', params: params, jwt: options)
+      end
+
+      def refund(amount, credit_card, options)
+        params = {}
+        # refund amounts must be negative
+        params['amount'] = ('-' + localized_amount(amount.to_f, options[:currency])).to_f
+
+        add_bank(params, options[:auth_code])
+        add_credit_card(params, credit_card, 'refund', options) unless options[:auth_code]
+        add_type_merchant_refund(params, options)
+        commit('refund', params: params, jwt: options)
+      end
+
+      def capture(amount, authorization, options = {})
+        params = {}
+        params['amount'] = localized_amount(amount.to_f, options[:currency])
+        params['authCode'] = options[:authCode]
+        params['merchantId'] = @options[:merchant_id]
+        params['paymentToken'] = authorization
+        params['shouldGetCreditCardLevel'] = true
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/priority.rb
```diff
@@ -0,0 +1,341 @@
...
+
+      def capture(amount, authorization, options = {})
+        params = {}
+        params['amount'] = amount
+        params['authCode'] = options[:authCode]
+        params['merchantId'] = @options[:merchant_id]
+        params['paymentToken'] = authorization
+        params['shouldGetCreditCardLevel'] = true
+        params['source'] = options['source']
+        params['tenderType'] = 'Card'
```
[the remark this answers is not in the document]
jessiagee: Fixed it so that we can pass that in as a GSF

lib/active_merchant/billing/gateways/priority.rb
```diff
@@ -0,0 +1,341 @@
...
+      rescue JSON::ParserError
+        message = 'Invalid JSON response received from Priority Gateway. Please contact Priority Gateway if you continue to receive this message.'
+        message += " (The raw response returned by the API was #{body.inspect})"
+        {
+          'message' => message
+        }
+      end
+
+      def success_from(response)
+        return true if response['paymentToken'] || response['status'] == 'Approved' || response['code'] == '204' || response['tenderType'] == 'Card' || response == ''
```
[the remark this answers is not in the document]
jessiagee: There were! I fixed this method after noticing that in my testing.

test/remote/gateways/remote_priority_test.rb
```diff
@@ -0,0 +1,352 @@
...
+
+  # Void tests
+  # Batch status is by default is set to Open wehn Sale transaction is created
+  def test_successful_void_batch_open
+    response = @gateway.purchase(@amount_purchase, @credit_card_purchase_success, @option_spr)
+    assert_success response
+
+    # check is this transaction associated batch is "Closed".
+    batch_check = @gateway.get_payment_status(response.params['batchId'], @option_spr)
+    # if batch Open then fail test. Batch must be closed to perform a Refund
```
dsmcclain, line 245: I have two questions here:
1. Are we just at the mercy of the sandbox environment regarding whether or not the batch is "Open" or "Closed"?
2. If the batch status is "Closed" does the `void` call still succeed?

test/remote/gateways/remote_priority_test.rb
```diff
@@ -0,0 +1,352 @@
...
+    @option_spr.update(authCode: '12345')
+    capture = @gateway.capture(@amount_authorize, 'bogus', @option_spr)
+    assert_failure capture
+
+    assert_equal 'Original Transaction Not Found', capture.params['authMessage']
+    assert_equal 'Declined', capture.params['status']
+  end
+
+  # Void tests
+  # Batch status is by default is set to Open wehn Sale transaction is created
```
dsmcclain, line 238: wehn 😄

dsmcclain (commented):
This is looking good. All tests are passing, I just have two questions about the `void` and `refund` tests.

lib/active_merchant/billing/gateways/priority.rb
```diff
@@ -0,0 +1,347 @@
...
+          card_details['cardPresent'] = credit_card['cardPresent']
+          card_details['cardType'] = credit_card['cardType']
+          card_details['entryMode'] = credit_card['entryMode']
+          card_details['expiryMonth'] = credit_card['expiryMonth']
+          card_details['expiryYear'] = credit_card['expiryYear']
+          card_details['hasContract'] = credit_card['hasContract']
+          card_details['isCorp'] = credit_card['isCorp']
+          card_details['isDebit'] = credit_card['isDebit']
+          card_details['last4'] = credit_card['last4']
+          card_details['token'] = credit_card['token']
```
naashton, line 163: Looks like `expiry*`, `entryMode`, `last4` are common to each action, so we can probably handle them in 1 place. Also, the `refund` portion is pulling values from `credit_card` as though it's a hash whereas in `purchase` it is pulling them as though they are an attribute of a payment object. I think the way they are handled in the `purchase` action is the correct way, but we should conform them to one or the other.

lib/active_merchant/billing/gateways/priority.rb
```diff
@@ -0,0 +1,347 @@
...
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
+          gsub(%r((number\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r((cvv\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
+      end
+
+      def add_bank(params, auth_code)
+        params['authCode'] = auth_code
+        params['authMessage'] = ''
```
naashton, line 128: Let's remove values we are defaulting to empty string or nil, so we can reduce some of the clutter in here. Unless it's explicitly stated in our SOW, I would rather have this boiled down to what is asked for this implementation.

test/remote/gateways/remote_priority_test.rb
```diff
@@ -0,0 +1,352 @@
...
+    @credit_card_purchase_fail_invalid_number = credit_card('4111', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
+
+    # purchase params fail missing card number month
+    @credit_card_purchase_fail_missing_month = credit_card('4111111111111111', month: '', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
+
+    # purchase params fail missing card verification number
+    @credit_card_purchase_fail_missing_verification = credit_card('4111111111111111', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '')
+
+    # authorize params success
+    @amount_authorize = 7.99
```
naashton, line 39: I think this is the same as the `@amount_purchase` in that it should be passed as an `int`.

test/remote/gateways/remote_priority_test.rb
```diff
@@ -0,0 +1,352 @@
...
+    # authorize params success end
+
+    # verify params
+    @iid = '10000001617842'
+    @card_number_verify = '4111111111111111'
+    # verify params end
+
+    # Refund params
+    @amount_refund = -4.32
```
naashton, line 48: Same on ln39

naashton (changes requested):
Had a couple of comments, but it's in a much better state than it was before.

test/remote/gateways/remote_priority_test.rb
jessiagee: @dsmcclain yeah as far as I can tell the only way to get the payment status for a batch is to call this method and pull it from them.

2. It looks like from https://developer.mxmerchant.com/docs/making-a-full-refund-or-void-a-transaction that if a batch is closed and you void an associated transaction in it, it'll open up a refund for that transaction. I did test it and looks like the status is now `pending` for the batch so I'm going to assume that is the intended flow and put in a test to reflect that.

test/remote/gateways/remote_priority_test.rb
jessiagee: Removed

test/remote/gateways/remote_priority_test.rb
jessiagee: Removed

dsmcclain (approved):
Nice work, @jessiagee. All tests are passing, and I don't have anything else to add at this point.

Approved!

naashton:
Everything looks good, I just had 1 test fail however

```
  => 190:   def test_failed_verify
     191:     # Generate jwt token from key and secret. Pass generated jwt to verify function. The verify function requries a jwt for header authorization.
     192:     jwt_response = @gateway.create_jwt(@option_spr)
     193:     @gateway.verify('123456', jwt_response.params['jwtToken'])
test/remote/gateways/remote_priority_test.rb:190:in `test_failed_verify'
test/remote/gateways/remote_priority_test.rb:199:in `rescue in test_failed_verify'
Failure: test_failed_verify(RemotePriorityTest):
  </No\ bank\ information\ found\ for\ bin\ number/> was expected to be =~
  <"error">.
```

Seems to just be a different message, but I'll go ahead and approve this and let you deal with this as you see fit.

naashton (approved):

PR #4178, opened 2021-11-01
Stripe Payment Intents: Add setup_purchase

aenand:
ECS-2093

In order to add suppport for Stripe alternative payment methods (APMs), the StripePI gateway must support a `setup_purchase` method. This method takes an `amount`, `currency`, and array of `payment_method_types` then submits a post to the `payment_intents` endpoint. It should return a transaction in the `requires_payment_method` status as the user completes the transaction flow for whichever payment method they selected.

Since Stripe APM's are offsite purchases that could be completed at any point, transactions that flow through the `setup_purchase` will have a response of type `StripePaymentIntentsResponse`. This is done to ensure that systems can identify this type of transaction as different from other StripePI transactions, similar to how Paypal Express works.

Test Summary
Local:
4956 tests, 74514 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
31 tests, 171 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
65 tests, 306 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.3846% passed
**There are 3 tests failing on master**

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
 == HEAD
 * Wompi: support gateway [therufs] #4173
+* Stripe Payment Intents: Add setup_purchase [aenand] #4178
 
 == Version 1.124.0 (October 28th, 2021)
 * Worldpay: Add Support for Submerchant Data on Worldpay [almalee24] #4147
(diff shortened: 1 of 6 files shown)


esmitperez (approved):
Very clear and easy to follow.

aenand:
> Very clear and easy to follow.

Shout out to @Carrigan for the help and making it so clean!

Carrigan (approved):
LGTM, especially the remote tests 👍 

My only suggestion would be to rename the `class_name` option to `response_class` so that it is more obvious that the class injection happens specifically for the response type.

StaceyBroers1972:
Same thing here

PR #4183, opened 2021-11-05
Stripe Payment Intents: refactor response

aenand:
In PR #4178 support for a `setup_purchase` transaction was added to the Stripe PI gateway. As part of this, a new `Response class` was added so systems can differentiate that this is an offsite transaction. Creating a new `Response` class is not the best way to implement this and therefore refactoring to remove it.

Test Summary
Local:
4977 tests, 74599 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
31 tests, 171 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
65 tests, 305 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.3846% passed
**There are 3 tests failing on master**

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,7 +7,8 @@
 * Worldpay: Adding support for google pay and apple pay [cristian] #4180
 * Worldpay: Adding scrubbing for network token transactions [cristian] #4181
 * SafeCharge: Add sg_NotUseCVV field [ajawadmirza] #4177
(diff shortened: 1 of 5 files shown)


Carrigan (approved):
Nice! Glad we were able to remove this :+1:

PR #4186, opened 2021-11-08
Wompi: cast error messages to JSON

therufs:
Remote: 9 tests, 21 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 9 tests, 38 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop: 723 files inspected, no offenses detected

CE-1979

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 * SafeCharge: Add sg_NotUseCVV field [ajawadmirza] #4177
 * PayULatam: Correctly map maestro and condensa card types [dsmcclain] #4182
 * StripePaymentIntents: Refactor response for setup_purchase [aenand] #4183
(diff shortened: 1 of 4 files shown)


dsmcclain (approved):
Tests pass. I agree with your reasoning that we should be returning JSON strings in these response fields instead of a ruby Hash.

I'm sort of surprised that this isn't something we've had to do with other gateways as well.

therufs:
ty @dsmcclain! I suspect we haven't had to do it with other gateways mainly because 
- Wompi returns unusually pithy errors, and I wanted to capture all the info 
- I didn't think through how best to do this, whereas other contributors probably did 😂

PR #4188, opened 2021-11-09
IPG: Add support for params &  gateway specific fields

ajawadmirza:
Accept 4 digit payment year (as per the standard credit card options)
Return single order id string in authorization
Add submerchant gateway specific fields
Required changes in test cases

CE-2038

Unit:
4979 tests, 74611 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
14 tests, 41 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/ipg.rb
+++ b/lib/active_merchant/billing/gateways/ipg.rb
@@ -48,7 +48,7 @@ module ActiveMerchant #:nodoc:
 
       def void(authorization, options = {})
         xml = Builder::XmlMarkup.new(indent: 2)
-        add_transaction_details(xml, authorization)
+        add_transaction_details(xml, options.merge!({ order_id: authorization }))
 
         commit('void', xml)
       end
@@ -86,8 +86,9 @@ module ActiveMerchant #:nodoc:
       def build_purchase_and_authorize_request(money, payment, options)
         xml = Builder::XmlMarkup.new(indent: 2)
         add_payment(xml, payment, options)
+        add_sub_merchant(xml, options[:submerchant]) if options[:submerchant]
         add_three_d_secure(xml, options[:three_d_secure]) if options[:three_d_secure]
-        add_stored_credentials(xml, options[:stored_credential]) if options[:stored_credential]
+        add_stored_credentials(xml, options) if options[:stored_credential] || options[:recurring_type]
         add_amount(xml, money, options)
         add_transaction_details(xml, options)
         add_billing(xml, options[:billing]) if options[:billing]
@@ -98,7 +99,7 @@ module ActiveMerchant #:nodoc:
       def build_capture_and_refund_request(money, authorization, options)
         xml = Builder::XmlMarkup.new(indent: 2)
         add_amount(xml, money, options)
-        add_transaction_details(xml, authorization, true)
+        add_transaction_details(xml, options.merge!({ order_id: authorization }), true)
         xml
       end
 
@@ -132,7 +133,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_credentials(xml, params)
-        recurring_type = params[:initial_transaction] ? 'FIRST' : 'REPEAT'
+        recurring_type = params[:stored_credential][:initial_transaction] ? 'FIRST' : 'REPEAT' if params[:stored_credential]
+        recurring_type = params[:recurring_type] if params[:recurring_type]
         xml.tag!('v1:recurringType', recurring_type)
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/ipg.rb
```diff
@@ -160,7 +162,7 @@ def add_payment(xml, payment, options = {}, credit_envelope = 'v1')
           xml.tag!("#{credit_envelope}:CreditCardData") do
             xml.tag!('v1:CardNumber', payment.number) if payment.number
             xml.tag!('v1:ExpMonth', payment.month) if payment.month
-            xml.tag!('v1:ExpYear', payment.year) if payment.year
+            xml.tag!('v1:ExpYear', payment.year % 100) if payment.year && (payment.year.digits.count == 4)
```
dsmcclain, line 165: I'm not sure if this is what we want. It only adds year if it is passed in as 4 digits.

There is a method in `lib/active_merchant/billing/credit_card_formatting.rb` called `format` that can be called like this:

```
xml.tag!('v1:ExpYear', format(payment.year, :two_digits)) if payment.year
```

...that way we don't have to care about how many digits the year on the card is. This will send a 2-character string every time.

lib/active_merchant/billing/gateways/ipg.rb
```diff
@@ -179,6 +181,34 @@ def add_payment(xml, payment, options = {}, credit_envelope = 'v1')
         end
       end
 
+      def add_sub_merchant(xml, submerchant)
+        xml.tag!('v1:SubMerchant') do
+          xml.tag!('v1:Mcc', submerchant[:mcc]) if submerchant[:mcc]
+          xml.tag!('v1:LegalName', submerchant[:legal_name]) if submerchant[:legal_name]
+          add_address(xml, submerchant[:address]) if submerchant[:address]
+          add_document(xml, submerchant[:document]) if submerchant[:document]
```
dsmcclain, line 189: Nice work getting all of these xml fields added in an easy-to-read way! 👍

dsmcclain (changes requested):
These changes look good for the most part. I have a few requests:
* I think we should change the logic on expiration year to use the `format` method (see comment below).
* It would be nice to have a unit test for the `SubMerchant` object using `check_request` to show that all of the fields end up in the right place in the request (tedious, I know).
* It would be nice to add a unit test for the `recurring_type` field, if only just to document it as an alternative to passing the normalized `stored_credential` fields.

I am still working on setting up my credentials so that I can run the remote tests. I will review the PR again once I get the remote tests running, but this should give you a little bit to work on in the meantime.

lib/active_merchant/billing/gateways/ipg.rb
```diff
@@ -132,7 +133,8 @@ def build_soap_request(action, body)
       end
 
       def add_stored_credentials(xml, params)
-        recurring_type = params[:initial_transaction] ? 'FIRST' : 'REPEAT'
+        recurring_type = params[:stored_credential][:initial_transaction] ? 'FIRST' : 'REPEAT' if params[:stored_credential]
+        recurring_type = params[:recurring_type] if params[:recurring_type]
```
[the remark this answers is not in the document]
ajawadmirza: Unit tests are added for both :)

lib/active_merchant/billing/gateways/ipg.rb
ajawadmirza: ah yes. thats what I wanted to do. Thanks for informing, this change is updated in latest commit.

lib/active_merchant/billing/gateways/ipg.rb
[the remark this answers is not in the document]
ajawadmirza: Yes we wanted to add `recurring_type` as an alternative to send through gateway specific field.

lib/active_merchant/billing/gateways/ipg.rb
ajawadmirza: Thanks 😃

test/unit/gateways/ipg_test.rb
```diff
@@ -56,6 +56,56 @@ def test_successful_purchase_with_stored_credentials
     assert_success response
   end
 
+  def test_successful_purchase_with_submerchant
```
dsmcclain, line 59: Thank you for adding these tests! 😅

dsmcclain (approved):
Okay, I have the remote tests working and everything is passing!

This change is good to go. I will approve and then merge 👍

dsmcclain:
Closing this PR as the changes were merged [here](https://github.com/activemerchant/active_merchant/commit/e8e5f18d3109b042b3c872feaa2b66d89a83789c).

PR #4190, opened 2021-11-09
GlobalCollect: Support for Lodging Data

naashton:
Expand support to add a comprehensive set of information for lodgingData
with GlobalCollect.

CE-2053

Unit:

Remote: 26 tests, 62 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.1538% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -104,6 +104,7 @@ module ActiveMerchant #:nodoc:
           'invoiceNumber' => options[:invoice]
         }
         add_airline_data(post, options) if options[:airline_data]
+        add_lodging_data(post, options)
         add_number_of_installments(post, options) if options[:number_of_installments]
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -135,6 +136,61 @@ def add_airline_data(post, options)
         post['order']['additionalInput']['airlineData'] = airline_data
       end
 
+      def add_lodging_data(post, options)
+        return unless lodging_options = options[:lodging_data]
```
dsmcclain, line 140: Just as an aside comment, I personally prefer this pattern where we have a guard clause inside the method over making the method call conditional as in
```add_lodging_data(post, options) if options[:lodging_data]```

File: lib/active_merchant/billing/gateways/global_collect.rb [hunk 2]
```diff
@@ -135,6 +136,61 @@ def add_airline_data(post, options)
...
+        lodging_data['numberOfAdults'] = lodging_options[:number_of_adults] if lodging_options[:number_of_adults]
+        lodging_data['numberOfNights'] = lodging_options[:number_of_nights] if lodging_options[:number_of_nights]
+        lodging_data['numberOfRooms'] = lodging_options[:number_of_rooms] if lodging_options[:number_of_rooms]
+        lodging_data['programCode'] = lodging_options[:program_code] if lodging_options[:program_code]
+        lodging_data['propertyCustomerServicePhoneNumber'] = lodging_options[:property_customer_service_phone_number] if lodging_options[:property_customer_service_phone_number]
+        lodging_data['propertyPhoneNumber'] = lodging_options[:property_phone_number] if lodging_options[:property_phone_number]
+        lodging_data['renterName'] = lodging_options[:renter_name] if lodging_options[:renter_name]
+        lodging_data['rooms'] = add_rooms(lodging_options)
+
+        post['order']['additionalInput']['lodgingData'] = lodging_data
```
dsmcclain, line 161: Wow, who knew there were so many lodging options!

dsmcclain (approved):
Thank you for the thorough testing of these fields.

The single failing remote test I get is `test_successful_purchase_with_installments`, also failing on master.

Everything else is passing so you are good to 🚢 . Don't forget the changelog entry!

lib/active_merchant/billing/gateways/global_collect.rb
naashton: Same. I actually refactored the `add_airline_data` to do the same. It just feels a bit cleaner and reduces the nesting as we fill in the body of the request.

lib/active_merchant/billing/gateways/global_collect.rb
naashton: This PR is a lesson on those hidden lodging fee's.

PR #4192, opened 2021-11-12
Wompi: option for installments

therufs:
Unit:
9 tests, 38 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
10 tests, 24 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:


dsmcclain (approved):
Tests pass.

If somebody puts something weird in `installments` like a non-numeric string, `to_i` will force it to 0 and testing that case produces an `'INPUT VALIDATION ERROR'` from the gateway. Which, is probably what we want? I'm not sure that case will arise in practice anyways.

Approved!

PR #4193, opened 2021-11-15
Stripe Payment Intents: Add support for fulfillment_date and event_type

dsmcclain:
`fulfillment_date` is a root field while `event_type` belongs in the `metadata` object.

Any reviewer should run the remote tests for both the Stripe PI and regular Stripe gateways, as this PR makes changes to both files.

CE-2101

Local:
4980 tests, 74615 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:

Loaded suite test/remote/gateways/remote_stripe_payment_intents_test
66 tests, 311 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Loaded suite test/remote/gateways/remote_stripe_test
74 tests, 338 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -534,7 +534,6 @@ module ActiveMerchant #:nodoc:
         post[:metadata].merge!(options[:metadata]) if options[:metadata]
         post[:metadata][:email] = options[:email] if options[:email]
         post[:metadata][:order_id] = options[:order_id] if options[:order_id]
-        post.delete(:metadata) if post[:metadata].empty?
       end
 
       def add_emv_metadata(post, creditcard)
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -35,6 +35,7 @@ module ActiveMerchant #:nodoc:
         add_ntid(post, options)
         add_claim_without_transaction_id(post, options)
         add_error_on_requires_action(post, options)
+        add_fulfillment_date(post, options)
         request_three_d_secure(post, options)
 
         CREATE_INTENT_ATTRIBUTES.each do |attribute|
@@ -56,6 +57,7 @@ module ActiveMerchant #:nodoc:
         CONFIRM_INTENT_ATTRIBUTES.each do |attribute|
           add_whitelisted_attribute(post, options, attribute)
         end
+
         commit(:post, "payment_intents/#{intent_id}/confirm", post, options)
       end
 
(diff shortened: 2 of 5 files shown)


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -534,7 +534,6 @@ def add_metadata(post, options = {})
         post[:metadata].merge!(options[:metadata]) if options[:metadata]
         post[:metadata][:email] = options[:email] if options[:email]
         post[:metadata][:order_id] = options[:order_id] if options[:order_id]
-        post.delete(:metadata) if post[:metadata].empty?
```
dsmcclain, line 537: This line was removed because:
* It is superfluous. There appears to be no problem adding an empty `metadata` object to the request.
* The Stripe PI gateway uses this method, and this line was precluding us from extending the method easily.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -254,6 +258,16 @@ def add_customer(post, options)
         post[:customer] = customer if customer.start_with?('cus_')
       end
 
+      def add_fulfillment_date(post, options)
+        post[:fulfillment_date] = options[:fulfillment_date] if options[:fulfillment_date]
```
meagabeth, line 262: When core passes in fields for the options hash, they are made into strings. Does this field need to remain an integer when it gets passed to the gateway? If so, we may need to add a `.to_i` to it.

meagabeth (commented):
There is a possibility that there is a change needed, but maybe not? There's more of an explanation made in the comments 😄

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
dsmcclain: Thank you, this is a good point.

The field supposedly accepts a timestamp. Here are my results from testing out various values:
* `1636756194` (Integer) - passes
* `'1636756194'` (String) - passes
* `'12'` (String, not a timestamp) - passes
* `0` (Integer) - fails with error `"Invalid timestamp: must be an integer"`
* `'whatever I want'` - fails with error `"Invalid integer: whatever I want"`

So it looks like the gateway is accepting any integer or any string that resolves to an integer, which seems to indicate that casting the value with `to_i` won't make a difference.

BUT, I'm going to do it anyways, just because we already know it's supposed to be a timestamp so I would rather make it an integer before we send it versus relying on the gateway's API to accept timestamps that are both string and integer, which is functionality we don't have control over and could change at any time.

meagabeth: Very thorough testing, much appreciated. And I think your logic for making the change is a good one. Thanks!

meagabeth (approved):
Tests pass, changes made for `.to_i` are complete. Looks great! 🚀

PR #4195, opened 2021-11-16
Priority: Fix unit test cases

ajawadmirza:
Fixed unit test cases implementation for priority to pass without
fixtures (tests were being failed without fixtures credentials before)

CE-2151

Unit:
4995 tests, 74768 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
100% passed

Rubocop:

### DIFF
--- a/test/unit/gateways/priority_test.rb
+++ b/test/unit/gateways/priority_test.rb
@@ -6,8 +6,7 @@ class PriorityTest < Test::Unit::TestCase
     # run command below to run tests in debug (byebug)
     # byebug -Itest test/unit/gateways/priority_test.rb
 
-    @gateway = PriorityGateway.new(fixtures(:priority))
-    @merchant = '1000003310'
+    @gateway = PriorityGateway.new(key: 'sandbox_key', secret: 'secret', merchant_id: 'merchant_id')
 
     # purchase params success
     @amount_purchase = 4
@@ -41,7 +40,7 @@ class PriorityTest < Test::Unit::TestCase
       customer_code: 'PTHER000IKZK',
       invoice: 'R000IKZK',
       is_duplicate: false,
-      merchant_id: @merchant,
+      merchant_id: @gateway.options[:merchant_id],
       payment_token: 'P6NyKC5UfmZjgAlF3ZEd3YSaJG9qKT6E',
       card_type: 'Visa',
       entry_mode: 'Keyed',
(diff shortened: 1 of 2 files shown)


meagabeth:
After looking at other unit test files, the `priority_test.rb` doesn't follow our standard when defining the `@gateway` variable. Unit tests shouldn't reference `:fixtures` at all so we need to make that change on [line 9.](https://github.com/activemerchant/active_merchant/blob/f11bf19eac71ca143c2b569a28582a7d2521b020/test/unit/gateways/priority_test.rb#L9)
It should reflect something like:
`@gateway = PriorityGateway.new(key: 'sandbox_key', secret: 'secret', merchant_id: 'merchant_id')`

This change may also impact the changes you've made within this file so we'll want to double check if additional changes are needed.

test/unit/gateways/priority_test.rb
```diff
@@ -6,7 +6,7 @@ def setup
     # run command below to run tests in debug (byebug)
     # byebug -Itest test/unit/gateways/priority_test.rb
 
-    @gateway = PriorityGateway.new(fixtures(:priority))
+    @gateway = PriorityGateway.new(key: 'sandbox_key', secret: 'secret', merchant_id: 'merchant_id')
     @merchant = '1000003310'
```
meagabeth, line 10: Again, this is unrelated to changes you've made, but something I noticed and would like us to resolve. I think this is a real merchant_id and should probably be removed from the file. Any place that references this variable throughout the test file (I see it on lines  44 and 340) should instead reference `@gateway.options[:merchant_id]` to use the placeholder credentials that are being used for the unit test.

test/unit/gateways/priority_test.rb
```diff
@@ -170,14 +170,11 @@ def test_failed_capture
   end
 
   def test_failed_void
-    response = stub_comms do
-      @gateway.void(123456, @option_spr)
-    end.respond_with(failed_void_response)
+    @gateway.expects(:ssl_request).returns(failed_void_response)
```
meagabeth, line 173: This looks much cleaner, thank you!

meagabeth (changes requested):
Overall your changes look great! I have requested a change that is unrelated to the work that you did, but will help clean up any exposed information that doesn't need to be in this unit test file. Hopefully it should be a quick edit and then we can get this merged into production. Thanks for your hard work!

test/unit/gateways/priority_test.rb
ajawadmirza: Done please check 👍

test/unit/gateways/priority_test.rb
ajawadmirza: Thanks 😄

meagabeth (approved):
Great work! Approved!

PR #4198, opened 2021-11-18
EBANX: Added `processing_type` Gateway Specific Field

kledoux-spreedly:
The headers that are generated for the `commit` method are now generated via a helper method. This method will look for a gateway specific field `processing_type`; if the field is present with a value of `local`, an additional header will be added to the gateway request: `x-ebanx-api-processing-type`.

CE-2113

Rubocop:

Local:
4995 tests, 74785 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
Loaded suite test/remote/gateways/remote_ebanx_test
25 tests, 69 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -214,6 +214,7 @@ module ActiveMerchant #:nodoc:
         post[:metadata] = options[:metadata] if options[:metadata]
         post[:metadata] = {} if post[:metadata].nil?
         post[:metadata][:merchant_payment_code] = options[:order_id] if options[:order_id]
+        post[:processing_type] = options[:processing_type] if options[:processing_type]
       end
 
       def parse(body)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -214,6 +214,7 @@ def add_additional_data(post, options)
         post[:metadata] = options[:metadata] if options[:metadata]
         post[:metadata] = {} if post[:metadata].nil?
         post[:metadata][:merchant_payment_code] = options[:order_id] if options[:order_id]
+        post[:processing_type] = options[:processing_type] if options[:processing_type]
```
meagabeth, line 217: Out of curiosity, is this field one that they want to receive at the gateway level or is it more to flag that we need to send a specific header?

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -236,6 +238,19 @@ def commit(action, parameters)
         )
       end
 
+      def headers(params)
```
meagabeth, line 241: Very nice way to handle this -- well done!

meagabeth (approved):
Tests pass, code looks excellent! I appreciate the breaking out of new methods for executing this task. It made the code very easy to read. Be sure to add the changelog entry as mentioned in the process docs for merging into Active Merchant ✅ Way to go!

lib/active_merchant/billing/gateways/ebanx.rb
kledoux-spreedly: This field is used to flag sending a specific header to the gateway.

PR #4209, opened 2021-11-29
Stripe update api version

jherreraa:
Summary:
This PR changes the API version to the latest version available (2020-08-27), also includes the update of the use of method YAML.safe_load to avoid a warning on test helper

Test Execution:
test/remote/gateways/remote_stripe_test.rb
Finished in 122.237642 seconds.

74 tests, 338 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
98.6486% passed

test/remote/gateways/remote_stripe_payment_intents_test.rb
Finished in 164.5804 seconds.

66 tests, 313 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -25,7 +25,7 @@ module ActiveMerchant #:nodoc:
 
       DEFAULT_API_VERSION = '2015-04-07'
 
-      self.supported_countries = %w(AT AU BE BG BR CA CH CY CZ DE DK EE ES FI FR GB GR HK IE IT JP LT LU LV MT MX NL NO NZ PL PT RO SE SG SI SK US)
+      self.supported_countries = %w(AE AT AU BE BG BR CA CH CY CZ DE DK EE ES FI FR GB GR HK HU IE IN IT JP LT LU LV MT MX MY NL NO NZ PL PT RO SE SG SI SK US)
       self.default_currency = 'USD'
       self.money_format = :cents
       self.supported_cardtypes = %i[visa master american_express discover jcb diners_club maestro unionpay]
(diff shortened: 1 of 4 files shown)


curiousepic:
@jherreraa  Please remove the unrelated commits and squash/rebase the others.

curiousepic:
@jherreraa Hmm, I'm getting 5 failures in the unit tests for the base `stripe` gateway. Are you getting those locally?

jherreraa:
> @jherreraa Hmm, I'm getting 5 failures in the unit tests for the base `stripe` gateway; they seem to be around network tokenized cards. Are you getting those locally?

@curiousepic after double check on local tests we got the errors also so we correct them and push the solution... please check again

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -223,9 +223,10 @@ def store(payment, options = {})
 
             post[:default_card] = r.params['id'] if options[:set_default] && r.success? && !r.params['id'].blank?
 
-            r.process { update_customer(options[:customer], post) } if post.count > 0
+            r.process { update_customer(options[:customer], post.merge(expand: [:sources])) } if post.count > 0
```
curiousepic, line 226: @jherreraa Nice, this fixes the errors. Though, is there a reason `expand: [:sources]` is being set within the `process` call here, instead of being added on its own line, similar to below? If not, I'd suggest doing the same here for consistency and readability.

jherreraa: @curiousepic, We tried the solution the way you mentioned, nevertheless that solution causes new failures (multiple invokes for methods not allowed in unit tests), so that is the reason to solve the issue in the way you see now.

curiousepic: Ah, of course! Sorry for running in circles. I don't really see a different solution, so I'm ok with this.

curiousepic (approved):

PR #4217, opened 2021-12-02
Update Canadian institution number handling

therufs:
Canadian routing numbers are typically handled either in print/MICR format or
electronic format (see https://en.wikipedia.org/wiki/Routing_number_(Canada)#Format).
Update check validation to handle checks with a valid Canadian institution number, presented in the standard electronic format.

Unit tests:
5002 tests, 74820 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:


naashton (approved):
excellent

PR #4219, opened 2021-12-03
Mercado Pago: Update verify for custom amount

ajawadmirza:
Updated `verify` method to allow custom amount in options to support card
validation rules.

CE-2136

Remote:

36 tests, 102 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.6667% passed

Unit:
5004 tests, 74828 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/test/unit/gateways/mercado_pago_test.rb
+++ b/test/unit/gateways/mercado_pago_test.rb
@@ -252,6 +252,17 @@ class MercadoPagoTest < Test::Unit::TestCase
     assert response.test?
   end
 
+  def test_successful_verify_with_custom_amount
+    response = stub_comms do
+      @gateway.verify(@credit_card, @options.merge({ amount: 200 }))
+    end.check_request do |endpoint, data, _headers|
+      params = JSON.parse(data)
+      assert_equal 2.0, params['transaction_amount'] if endpoint.include? 'payment'
+    end.respond_with(successful_authorize_response)
+
+    assert_success response
+  end
+
   def test_successful_verify_with_failed_void
     @gateway.expects(:ssl_request).at_most(3).returns(failed_void_response)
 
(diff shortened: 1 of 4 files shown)


ajawadmirza:
3 remote tests are already failing from previous implementation and sometimes the gateway responses with `500 internal server error` in some responses.

image

therufs:
Hey @ajawadmirza this looks good! I think for posterity it would be worth it to add into this test a check that the expected amount is being passed, rather than just that the request succeeds. Looking at the transcripts, that could mean checking either the outgoing request or the response for the value.

ajawadmirza:
> Hey @ajawadmirza this looks good! I think for posterity it would be worth it to add into this test a check that the expected amount is being passed, rather than just that the request succeeds. Looking at the transcripts, that could mean checking either the outgoing request or the response for the value.

Thanks @therufs, I've added one unit test case that will now check the expected amount is being passed in the request, You can review now 😄

test/unit/gateways/mercado_pago_test.rb
```diff
@@ -252,6 +252,17 @@ def test_successful_verify
     assert response.test?
   end
 
+  def test_successful_verify_with_custom_amount
+    response = stub_comms do
+      @gateway.verify(@credit_card, @options.merge({ amount: 200}))
```
therufs, line 257: Rubocop requests a space after the closing bracket like so: `({ amount: 200 }))`

therufs (approved):
lgtm && tests pass except for remote tests `test_successful_purchase_with_processing_mode_gateway` and `test_partial_capture`, both failing on master also.

PR #4223, opened 2021-12-07
add skip_response option on request check fro commit stubs

Heavyblade:
### Summary
---
Currently is mandatory to add a "respond_with" to a CommStub instance in order to be able to check the content of a request, this change makes that optional.

The main point of this change si that doesn't make sense to test the validity of a response that is not affected by the input, no matter how bad the request is is not going to affect the response, it is basically an 'assert true'.

### Test Local
---
5004 tests, 74825 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
294.40 tests/s, 4402.22 assertions/s

### Test RuboCop
---

### DIFF
--- a/test/comm_stub.rb
+++ b/test/comm_stub.rb
@@ -8,17 +8,29 @@ module CommStub
       @check = nil
     end
 
-    def check_request(&block)
-      @check = block
-      self
+    def check_request(skip_response: false, &block)
+      if skip_response
+        @complete = true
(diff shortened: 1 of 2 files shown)


Carrigan (approved):
Cool idea! This definitely makes sense to have for tests where we just want to test the contents and not the response. LGTM and tests all pass locally 👍

PR #4240, opened 2021-12-15
Priority: Cleaning up refund method

jessiagee:
* Refund was currently using blind credit instead of refund
* Removed unused params updates

bundle exec rake test:local

5009 tests, 74781 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Running RuboCop...

Loaded suite test/unit/gateways/priority_test

11 tests, 59 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

100% passed

### DIFF
--- a/test/remote/gateways/remote_priority_test.rb
+++ b/test/remote/gateways/remote_priority_test.rb
@@ -22,10 +22,7 @@ class RemotePriorityTest < Test::Unit::TestCase
     @faulty_credit_card = credit_card('12345', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
 
     @option_spr = {
-      billing_address: address(),
-      avs_street: '666',
-      avs_zip: '55044',
-      tender_type: 'Card'
+      billing_address: address()
     }
 
     # purchase params fail inavalid card number
@@ -241,7 +238,7 @@ class RemotePriorityTest < Test::Unit::TestCase
       @gateway.close_batch(response.params['batchId'], @option_spr)
       refund_params = @option_spr.merge(response.params).deep_transform_keys { |key| key.to_s.underscore }.transform_keys(&:to_sym)
 
-      refund = @gateway.refund(response.params['amount'].to_f * 100, @credit_card, refund_params)
+      refund = @gateway.refund(response.params['amount'].to_f * 100, response.authorization.to_s, refund_params)
       assert_success refund
       assert refund.params['status'] == 'Approved'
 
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_priority_test.rb
```diff
@@ -241,7 +238,7 @@ def test_successful_refund_and_batch_closed
       @gateway.close_batch(response.params['batchId'], @option_spr)
       refund_params = @option_spr.merge(response.params).deep_transform_keys { |key| key.to_s.underscore }.transform_keys(&:to_sym)
 
-      refund = @gateway.refund(response.params['amount'].to_f * 100, @credit_card, refund_params)
+      refund = @gateway.refund(response.params['amount'].to_f * 100, response.authorization.to_s, refund_params)
```
dsmcclain, line 241: I think this change is fine for the context of this PR. I was able to verify that it is working.

However, I think it's kind of bad news that this test file contains a few tests with `if/else` statements. It means that when I run the tests I am not hitting all of the assertions. If I run the tests and they follow Path A, who's to say that Path B works?

Case in point: I had to put a breakpoint in this test just to verify that I was actually hitting this code when I run the test. Someone could make a bunch of changes in this `if` block and keep running the test and keep getting a green result without realizing that the test isn't even hitting their code.

I think it would be worthwhile to generate a separate task to rethink some of these tests. I don't think we want tests with multiple paths to green, and I wouldn't want people to replicate this pattern for other gateways.

dsmcclain (approved):
All tests passing, changes make sense to me!

I left a note about a potential issue in the remote test file. I think it extends beyond what you are trying to accomplish in this PR.

test/remote/gateways/remote_priority_test.rb
dsmcclain: Circling back around cause I just realized something: I think maybe the original intention of these `else` blocks was to make an assertion that is guaranteed to fail? The comments next to these tests (`# if batch Open then fail test`) seems to indicate this intention, although the comments are inaccurate (if the batch is open we go ahead and close it and proceed as planned).

Making an `assert` statement that is bound to fail is not a good way to communicate this intention. The presence of an `if/else` further confuses things, as it suggests there are multiple acceptable paths the test could follow.

RSpec offers a `flunk` method for just this purpose, and if we wrap it in a guard clause I think the intention becomes very clear:
```
flunk('the batch status prevents a refund from being tested') unless batch_check.params['status'] == 'Open'
```

jessiagee: Thanks @dsmcclain ! I've created a Jira ticket (CE-2253) to track these updates 👍🏻

PR #4241, opened 2021-12-16
IPG: Add store error response

ajawadmirza:
Added `tpv_error_code` and `tpv_error_msg` response fields when store
transaction fails for IPG implementation.

Unit:
5011 tests, 74854 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -58,6 +58,7 @@
 * Paymentez: Update card mappings [ajawadmirza] #4237
 * Priority: Update parsing for error messages [jessiagee] #4245
 * GlobalCollect: Support for Airline Data [naashton] #4187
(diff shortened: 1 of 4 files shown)


therufs:
Hey @ajawadmirza, this looks like a good start. Would you be able to write a remote test that generates the same response as appears in `failed_store_response`?

ajawadmirza:
> Hey @ajawadmirza, this looks like a good start. Would you be able to write a remote test that generates the same response as appears in `failed_store_response`?

Hi @therufs , we couldn't generate the same response from this gateway anyway even with passing different & invalid `hosted_data_id`. Only way we discussed (with @jessiagee ) is to generate via unit test case.

therufs (approved):
lgtm; all tests & cops pass!

PR #4243, opened 2021-12-16
Adds Multiple Item Codes and Amounts to Cashnet Gateway

peteroas:


test/remote/gateways/remote_cashnet_test.rb
```diff
@@ -22,6 +22,15 @@ def test_successful_purchase
     assert_equal 'Success', purchase.message
   end
 
+  def test_successful_purchase_with_multiple_items
+    options = @options.merge({itemcode: "FEE", itemcode2: "LOBSTER", itemcode3: "CODES", amount2: 1234, amount3: 4321})
```
dsmcclain, line 26: Instead of allowing for an infinite number of separate `itemcode` fields, I think what we want is to allow for an `item_codes` field that holds an object like:
```
"item_codes": {
  "item_code": "something",
  "amount": 100,
  "item_code_2": "something else",
  "amount_2": 50
}
```

Then, in the gateway adapter, `add_invoice` could look for `options[:item_codes]` and unpack it if it is present, mapping each field from snake_case to what the CASHNet API expects. If there is no `item_codes`, the method could do essentially what it does now with `post[:itemcode] = (options[:item_code] || @options[:default_item_code])`.

(Also, it's a real shame that CASHNet's API calls the first fields `itemcode` and `amount` instead of `itemcode1` and `amount1`)

dsmcclain (changes requested):
Two changes:
1. I think we want to change the logic to accommodate a single `item_codes` field that contains an object with all the various item codes, versus having an unlimited number of individual fields.
2. When you are ready to resubmit, make sure to run `bundle exec rake test:local`, which includes `rubocop`. This draft has 16 rubocop errors and you will need to clean that up before merging.

test/remote/gateways/remote_cashnet_test.rb
peteroas: @dsmcclain this seems like a good approach! My only concern is that even in his configuration, how can we NOT allow an infinite number of `itemcode` fields? Should we arbitrarily choose a maximum number of `itemcode` fields that we will map over?

I'm thinking that using your method, the logic would look like this:

```
    if options[:itemcodes].present?
        ###iterate through this map of itemcodes and amounts and add them to our post
    else
        post[:itemcode] = (options[:item_code] || @options[:default_item_code])
end
```

How does this solve the "infinite number of separate itemcode fields" problem?

peteroas: (also, I agree with you. not including that `itemcode1` and `amount1` makes iterating over those fields wayyyy clunkier)

dsmcclain: @peteroas Yes, your pseudocode is precisely what I am imagining. Except the `options` fields would be snake_case like `item_codes`.

> How does this solve the "infinite number of separate itemcode fields" problem?

Good question. Doing it this way provides a few advantages:
* It makes the request more organized. Having a container field like `item_codes` makes it easier for humans to read and write the requests. You are right that we could still add an unlimited number of item codes, but as the number of itemcodes grows the size of the request at the root level only grows by one.
* It makes it easier to code against the requests, because we can sequester the `item_codes` object and perform whatever parsing and logic we want with it without having to handle the entire request body along with it.
* As an example of the above, Spreedly Core defines every field that is passed along to the Active Merchant gateway adapter. If we put each itemcode and amount field at the root level, we would have to enumerate every field or else create some sort of wildcard to handle it. This way, we only have to specify one field: `item_codes`.

lib/active_merchant/billing/gateways/cashnet.rb
```diff
@@ -110,9 +110,34 @@ def add_creditcard(post, creditcard)
...
+
+            if index_array == []
+              post[:itemcode] = value
+            else
+              number = index_array[1]
+              post_key = 'itemcode' + number
+
+              post[post_key] = value
+            end
+          end
```
peteroas, line 130: @dsmcclain I've refactored as we discussed yesterday. However, the move from snake case `item_code#{number}` to `itemcode#{number}` feels extremely fragile and clunky. any advice for how to improve this within ruby?

dsmcclain: Yeah, I feel ya. I do think it is still the right idea to preserve the snake case approach, because part of the promise of Active Merchant is to provide a more-or-less standardized API for merchants that abstracts away the idiosyncrasies of each gateway (in this case, wacky query params).

Ruby provides a `transform_keys` method ([documentation here](https://ruby-doc.org/core-2.6/Hash.html#transform_keys-method)), which I think you could combine with `gsub` to achieve this with a lot less work:
```
if options[:item_codes].present?
  codes_and_amounts = options[:item_codes].transform_keys { |key| key.to_s.gsub('_', '') } # so long, snake case!
  codes_and_amounts.each do |key, value|
    post[key] = value if key.start_with?('itemcode', 'amount') # ignore anything that isn't an item code or amount
  end
else
  ...do other stuff
end
```

One potentially useful side effect of this approach is that if a merchant _did_ end up passing keys like `'itemcode2'` and `'amount2'` it would still go through.

 If this approach still feels fragile to you, it could be nice to add a test case or two showing that passing in an `item_codes` hash with weird keys just causes them to get dropped instead of blowing everything up.

test/unit/gateways/cashnet_test.rb
```diff
@@ -22,6 +22,21 @@ def test_successful_purchase
...
   end
 
+  def test_successful_purchase_with_multiple_items
+    @gateway.expects(:ssl_post).returns(successful_purchase_response)
+    options = { item_codes: { item_code: 'APPFEE', item_code2: 'LOBSTER', item_code3: 'CODES', amount2: 1234, amount3: 4321 } }
+
+    response = stub_comms(@gateway, :ssl_request) do
+      @gateway.purchase(@amount, @credit_card, options)
+    end.check_request do |_method, _endpoint, data, _headers|
+      assert_match('itemcode=FEE&itemcode2=LOBSTER&itemcode3=CODES&amount=1.00&amount2=12.34&amount3=43.21', data)
```
dsmcclain, line 32: NICE

dsmcclain (changes requested):
Great work continuing to wrestle with this @peteroas! I think you are very close. I made a suggestion that will hopefully make the logic you are adding feel less fragile and easier to maintain for future developers (aka us).

lib/active_merchant/billing/gateways/cashnet.rb
peteroas: @dsmcclain amazing, of course, why split it out when all we need to do is remove the underscore!!! Classic "Why didn't I think of that" moment 😂 

Yeah, was thinking some test cases for this would be useful once I got the transform finalized. Will implement now!

test/unit/gateways/cashnet_test.rb
peteroas: @dsmcclain ya know, it's funny you commented here, but just now I was messing around with this test, and noticed that this `assert_match` was passing even when it shouldn't. Here's what I have:

```
end.check_request do |_method, _endpoint, data, _headers|
      assert_match('THIS SHOULD NOT MATCH', data)
```

and it's STILL WORKING?!

Can't figure out why this is the case... any ideas?

dsmcclain: Nice catch @peteroas!

I think the reason is line 26: `@gateway.expects(:ssl_post).returns(successful_purchase_response)`.

I think because of the way stub_comms is written, stubbing `ssl_post` like this means that `respond_with` on line 33 never gets called, which means that the block passed in `check_request` never actually gets evaluated, which means that we never hit your assertion on line 32. (Pretty sure that's the sequence of events. `comm_stub.rb` is not easy to read.)

In any case, you should be able to actually test the request if you remove line 26.

test/unit/gateways/cashnet_test.rb
```diff
@@ -22,6 +22,35 @@ def test_successful_purchase
...
+  end
+
+  def test_successful_purchase_with_filtered_itemcodes
+    options = { item_codes: { item_code: 'APPFEE', item_code2: 'LOBSTER', item_code3: 'CODES', bad_key: 'BADVALUE', amount: 100, amount2: 1234, amount3: 4321 } }
+
+    response = stub_comms(@gateway, :ssl_request) do
+      @gateway.purchase(@amount, @credit_card, options)
+    end.check_request do |_method, _endpoint, data, _headers|
+      assert_match('itemcode=APPFEE&itemcode2=LOBSTER&itemcode3=CODES&amount=1.00&amount2=12.34&amount3=43.21', data)
+      refute_match('badkey=BADVALUE', data)
```
dsmcclain, line 46: NICE

dsmcclain (approved):
Nice work @peteroas. All unit tests and rubocop passing 🏁 

I get two failures on the remote tests:
`test_successful_refund_with_options`
`test_successful_purchase_and_refund`

...but they also fail on master and are unrelated to your changes. Looks like a settlement issue which is somewhat common for gateway sandboxes and refund requests.

Please squash down to one commit and add a CHANGELOG entry before merging.

PR #4244, opened 2021-12-17
IPG: verify method bug fixes for core

ajawadmirza:
Send default `currency` for verify method and allow two digit string in
`ExpMonth` in IPG request.

CE-2249
CE-2250

Unit:
5010 tests, 74849 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ipg.rb
+++ b/lib/active_merchant/billing/gateways/ipg.rb
@@ -63,6 +63,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def verify(credit_card, options = {})
+        options[:currency] = self.default_currency unless options[:currency] && !options[:currency].empty?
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/ipg.rb
```diff
@@ -63,6 +63,7 @@ def store(credit_card, options = {})
       end
 
       def verify(credit_card, options = {})
+        options[:currency] = self.default_currency
```
therufs, line 66: Hey @ajawadmirza -- does this value always have to be set to the default_currency, or could we allow users to pick their currency and only use the default if they don't specify another?

therufs (approved):
lgtm; thanks for adding tests! 🎉

PR #4248, opened 2021-12-29
StripePI: Enable Apple Pay and Google Pay payment methods

gasb150:
Summary:
In order to add  Apple Pay and Google Pay support to Stripe Payment Intents
this commit enables the option to store those payment methods for use in the Payment Intent creation.

Local Tests:
35 tests, 183 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote Tests:
Finished in 168.186212 seconds.
68 tests, 317 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -20,13 +20,19 @@ module ActiveMerchant #:nodoc:
           store_apple_or_google_pay_token = 'Direct Apple Pay and Google Pay transactions are not supported. Those payment methods must be stored before use.'
           return Response.new(false, store_apple_or_google_pay_token)
         end
+
         post = {}
         add_amount(post, money, options, true)
         add_capture_method(post, options)
         add_confirmation_method(post, options)
         add_customer(post, options)
-        result = add_payment_method_token(post, payment_method, options)
-        return result if result.is_a?(ActiveMerchant::Billing::Response)
+        if payment_method.is_a?(String) && payment_method.include?('tok_')
+          add_payment_method_card_data_tok(post, payment_method, options)
+        else
+          result = add_payment_method_token(post, payment_method, options)
+          return result if result.is_a?(ActiveMerchant::Billing::Response)
+
+        end
 
         add_external_three_d_secure_auth_data(post, options)
         add_metadata(post, options)
@@ -83,6 +89,16 @@ module ActiveMerchant #:nodoc:
         post_data[:card][:cvc] = payment_method.verification_value if payment_method.verification_value
         add_billing_address(post_data, options)
         add_name_only(post_data, payment_method) if post_data[:billing_details].nil?
+        post_data
+      end
+
+      def add_payment_method_card_data_tok(post_data, payment_method, options = {})
+        post_data[:payment_method_types] = {}
+        post_data[:payment_method_types][''] = 'card'
+        post_data[:payment_method_data] = {}
+        post_data[:payment_method_data][:type] = 'card'
+        post_data[:payment_method_data][:card] = {}
+        post_data[:payment_method_data][:card][:token] = payment_method
         post_data
       end
 
@@ -189,7 +205,10 @@ module ActiveMerchant #:nodoc:
 
         # If customer option is provided, create a payment method and attach to customer id
         # Otherwise, create a customer, then attach
-        if payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
+        if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard)
+          result = tokenize_apple_google_token(payment_method, options)
+          return result
+        elsif payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
           result = add_payment_method_token(params, payment_method, options)
           return result if result.is_a?(ActiveMerchant::Billing::Response)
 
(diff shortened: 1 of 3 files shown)


test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -26,12 +26,15 @@ def setup
 
     @apple_pay = apple_pay_payment_token
     @google_pay = network_tokenization_credit_card(
-      '4777777777777778',
-      payment_cryptogram: 'BwAQCFVQdwEAABNZI1B3EGLyGC8=',
-      verification_value: '987',
-      source: :google_pay,
+      '4242424242424242',
+      payment_cryptogram: 'reddelicious',
+      source: :android_pay,
```
curiousepic, line 31: Why was this changed? Is it more accurate test data?
Are we sure there are no differences between the source being `android_pay` and `google_pay`?

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -64,6 +86,14 @@ def test_successful_purchase
...
   end
 
+  def test_successful_google_pay_store
+    options = {
+      currency: 'GBP'
+    }
+    assert store = @gateway.store(@google_pay, options)
+    assert store.params['id'].start_with?('cus_')
+  end
+
```
curiousepic, line 96: Let's add one for Apple Pay if we can.

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -35,8 +35,30 @@ def setup
...
+      payment_cryptogram: 'reddelicious',
+      source: 'apple_pay',
+      brand: 'visa',
+      eci: '05',
+      month: '09',
+      year: '2030',
+      first_name: 'Longbob',
+      last_name: 'Longsen'
+    )
+    @credit_card_token = 'card_1KBQpkAWOtgoysogjIqSmzES'
```
curiousepic, line 61: Let's not include this if it is not used

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -189,7 +189,12 @@ def store(payment_method, options = {})
 
         # If customer option is provided, create a payment method and attach to customer id
         # Otherwise, create a customer, then attach
-        if payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
+        card_source_pay = payment_method.source.to_s if defined?(payment_method.source)
+        card_brand_pay = card_brand(payment_method) unless payment_method.is_a?(String) || payment_method.nil?
+        if NO_WALLET_SUPPORT.include?(card_source_pay) || NO_WALLET_SUPPORT.include?(card_brand_pay)
+          result = super(payment_method, options)
```
curiousepic, line 195: I'm a little confused about why we need to call to super and add logic to the base Stripe adapter here. Is it just a matter of not duplicating code?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -205,6 +205,11 @@ def store(payment, options = {})
           return bank_token_response unless bank_token_response.success?
 
           params = { source: bank_token_response.params['token']['id'] }
+        elsif payment.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard)
+          tokenize_apple_google = tokenize_apple_google(payment, options)
+          return tokenize_apple_google unless tokenize_apple_google.success?
+
+          params = { source: tokenize_apple_google.params['token']['id'] }
```
curiousepic, line 212: Would these changes also enable support for storing AP/GP via the base Stripe gateway?

curiousepic:
This looks pretty good - I left a few questions, but to summarize them: It's a little weird that supporting this on PI requires changes to the Base gateway, which already kind of supports AP/GP (but maybe not storing them)? So I'm wondering if we can also support storing them on the Base gateway as part of this update...

lib/active_merchant/billing/gateways/stripe.rb
gasb150: yes

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -262,6 +267,32 @@ def tokenize_apple_pay_token(apple_pay_payment_token, options = {})
         end
       end
 
+      def tokenize_apple_google(payment, options = {})
+        if payment.inspect.include?('google_pay')
```
curiousepic, line 271: I would check `payment.source == :google_pay` directly here, rather than `.include?`

There is also some possibility the `source` will be something other than these two (`android_pay` or `network_token`), which we should handle/guard against with a guard clause or a third case.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -189,7 +189,12 @@ def store(payment_method, options = {})
 
         # If customer option is provided, create a payment method and attach to customer id
         # Otherwise, create a customer, then attach
-        if payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
+        card_source_pay = payment_method.source.to_s if defined?(payment_method.source)
+        card_brand_pay = card_brand(payment_method) unless payment_method.is_a?(String) || payment_method.nil?
```
curiousepic, line 193: I don't think this final `payment_method.nil?` check is necessary at this point...

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -189,7 +189,12 @@ def store(payment_method, options = {})
...
+        if NO_WALLET_SUPPORT.include?(card_source_pay) || NO_WALLET_SUPPORT.include?(card_brand_pay)
+          result = super(payment_method, options)
+          return result if result.is_a?(ActiveMerchant::Billing::Response)
```
curiousepic, line 196: What are the other potential outcomes here other than an AM Response? And what happens if it's not?

gasb150:
We have decided to modify the current implementation a little bit. Therefore we will close this pull request to create a new one. Thanks so much.

PR #4252, opened 2022-01-04
Update Google and Apple Pay implementation

jherreraa:
StripePI: Enable Apple Pay and Google Pay payment methods

    Summary:
    In order to add  Apple Pay and Google Pay support to Stripe Payment Intents
    this commit enables the option to store those payment methods for use in the Payment Intent creation.

    Local Tests:
    Finished in 0.069145 seconds.

    37 tests, 193 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed

    Remote Tests:
    Finished in 191.73121 seconds.
    80 tests, 344 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed

    RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -11,43 +11,44 @@ module ActiveMerchant #:nodoc:
       CONFIRM_INTENT_ATTRIBUTES = %i[receipt_email return_url save_payment_method setup_future_usage off_session]
       UPDATE_INTENT_ATTRIBUTES = %i[description statement_descriptor_suffix statement_descriptor receipt_email setup_future_usage]
       DEFAULT_API_VERSION = '2020-08-27'
-      NO_WALLET_SUPPORT = %w(apple_pay google_pay android_pay)
 
       def create_intent(money, payment_method, options = {})
-        card_source_pay = payment_method.source.to_s if defined?(payment_method.source)
-        card_brand_pay = card_brand(payment_method) unless payment_method.is_a?(String) || payment_method.nil?
-        if NO_WALLET_SUPPORT.include?(card_source_pay) || NO_WALLET_SUPPORT.include?(card_brand_pay)
-          store_apple_or_google_pay_token = 'Direct Apple Pay and Google Pay transactions are not supported. Those payment methods must be stored before use.'
-          return Response.new(false, store_apple_or_google_pay_token)
+        MultiResponse.run do |r|
+          if payment_method.is_a?(NetworkTokenizationCreditCard)
+            r.process { tokenize_apple_google(payment_method, options) }
+            payment_method = (r.params['token']['id']) if r.success?
+          end
+          r.process do
+            post = {}
+            add_amount(post, money, options, true)
+            add_capture_method(post, options)
+            add_confirmation_method(post, options)
+            add_customer(post, options)
+
+            result = add_payment_method_token(post, payment_method, options)
+            return result if result.is_a?(ActiveMerchant::Billing::Response)
+
+            add_external_three_d_secure_auth_data(post, options)
+            add_metadata(post, options)
+            add_return_url(post, options)
+            add_connected_account(post, options)
+            add_radar_data(post, options)
+            add_shipping_address(post, options)
+            setup_future_usage(post, options)
+            add_exemption(post, options)
+            add_stored_credentials(post, options)
+            add_ntid(post, options)
+            add_claim_without_transaction_id(post, options)
+            add_error_on_requires_action(post, options)
+            add_fulfillment_date(post, options)
+            request_three_d_secure(post, options)
+
+            CREATE_INTENT_ATTRIBUTES.each do |attribute|
+              add_whitelisted_attribute(post, options, attribute)
+            end
+            commit(:post, 'payment_intents', post, options)
+          end
         end
-        post = {}
-        add_amount(post, money, options, true)
-        add_capture_method(post, options)
-        add_confirmation_method(post, options)
-        add_customer(post, options)
-        result = add_payment_method_token(post, payment_method, options)
-        return result if result.is_a?(ActiveMerchant::Billing::Response)
-
-        add_external_three_d_secure_auth_data(post, options)
-        add_metadata(post, options)
-        add_return_url(post, options)
-        add_connected_account(post, options)
-        add_radar_data(post, options)
-        add_shipping_address(post, options)
-        setup_future_usage(post, options)
-        add_exemption(post, options)
-        add_stored_credentials(post, options)
-        add_ntid(post, options)
-        add_claim_without_transaction_id(post, options)
-        add_error_on_requires_action(post, options)
-        add_fulfillment_date(post, options)
-        request_three_d_secure(post, options)
-
-        CREATE_INTENT_ATTRIBUTES.each do |attribute|
-          add_whitelisted_attribute(post, options, attribute)
-        end
-
-        commit(:post, 'payment_intents', post, options)
       end
 
       def show_intent(intent_id, options)
@@ -84,6 +85,13 @@ module ActiveMerchant #:nodoc:
         add_billing_address(post_data, options)
         add_name_only(post_data, payment_method) if post_data[:billing_details].nil?
         post_data
+      end
+
+      def add_payment_method_card_data_token(post_data, payment_method)
+        post_data.merge!({
+          payment_method_types: ['card'],
+          payment_method_data: { type: 'card', card: { token: payment_method } }
+        })
       end
 
       def update_intent(money, intent_id, payment_method, options = {})
@@ -186,7 +194,6 @@ module ActiveMerchant #:nodoc:
       def store(payment_method, options = {})
         params = {}
         post = {}
-
         # If customer option is provided, create a payment method and attach to customer id
         # Otherwise, create a customer, then attach
         if payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
@@ -285,7 +292,13 @@ module ActiveMerchant #:nodoc:
       def add_payment_method_token(post, payment_method, options)
         case payment_method
         when StripePaymentToken
-          post[:payment_method] = payment_method.payment_data['id']
+          post[:payment_method_data] = {
+            type: 'card',
+            card: {
+              token: payment_method.payment_data['id'] || payment_method.payment_data
+            }
+          }
+          post[:payment_method] = payment_method.payment_data['id'] || payment_method.payment_data
         when String
           extract_token_from_string_and_maybe_add_customer_id(post, payment_method)
         when ActiveMerchant::Billing::CreditCard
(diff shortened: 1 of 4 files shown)


test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -63,6 +78,28 @@ def test_successful_create_and_capture_intent
     assert_equal 'Payment complete.', capture.params.dig('charges', 'data')[0].dig('outcome', 'seller_message')
   end
 
+  def test_store_does_not_pass_validation_to_attach_by_default2
```
curiousepic, line 81:
```suggestion
  def test_store_google_pay_does_not_pass_validation_to_attach_by_default
```

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -64,6 +100,160 @@ def test_successful_purchase
...
+  end
+
+  def test_successful_authorize_with_google_pay
+    options = {
+      currency: 'GBP'
+    }
+    store = @gateway.store(@google_pay, options)
+    assert store.success?
+    stripe_token = store.params['token']['id']
+    auth = @gateway.authorize(@amount, stripe_token, options)
```
curiousepic, line 190: It may not matter for these specific tests, but when testing using a stored payment method, we should be passing the `store` response's `.authorization` rather than a specific param.
So this should work:
```suggestion
    stripe_token = store.params['token']['id']
    auth = @gateway.authorize(@amount, stripe_token, options)
```

curiousepic: For Stripe, this is relevant if the PM has an attached `customer`: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L693-L703

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -186,10 +198,13 @@ def refund(money, intent_id, options = {})
...
         post = {}
-
         # If customer option is provided, create a payment method and attach to customer id
         # Otherwise, create a customer, then attach
-        if payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
+        if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard)
+          tokenize_apple_google = tokenize_apple_google(payment_method, options)
+          return tokenize_apple_google
+
+        elsif payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
```
curiousepic, line 207: We should move the comment from line 201 to 206, since it applies to this line above.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -86,6 +88,16 @@ def add_payment_method_data(payment_method, options = {})
         post_data
       end
 
+      def add_payment_method_card_data_tok(post_data, payment_method)
+        post_data[:payment_method_types] = {}
+        post_data[:payment_method_types][''] = 'card'
```
Heavyblade, line 93: ❓ Suggestion:
Just leaving a note here for @curiousepic, is very weird that the post_data method translates a 

`payment_method_types: { '': 'card' }` into `payment_method_types[]=card`

maybe that needs a refactor or a change to make that:  

`payment_method_types:  ['card']` translates into: `payment_method_types[]=card`

 which makes more sense

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -299,7 +320,34 @@ def extract_token_from_string_and_maybe_add_customer_id(post, payment_method)
...
+        if payment_method.include?('tok_')
+          add_payment_method_card_data_tok(post, payment_method)
+        else
+          post[:payment_method] = payment_method
+        end
+      end
+
+      def tokenize_apple_google(payment, options = {})
+        tokenization_method = options[:tokenization_method] || 'android_pay'
+        tokenization_method = 'apple_pay' if payment.source.to_sym == :apple_pay
```
Heavyblade, line 332: 🛑   Warning:

We need a better logic here, seems to be the case that tokenization_method param can be determined on the options or on the payment object but is not clear the priority order, so I have some questions:

1) what is the logic behind that couple of possibilities ? my understanding is the `payment.source` should be only truth. 
2) What happens if options[:tokenization_method] maybe has the value "google_pay" ?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -299,7 +320,34 @@ def extract_token_from_string_and_maybe_add_customer_id(post, payment_method)
...
+            eci: payment.eci,
+            cryptogram: payment.payment_cryptogram
+          }
+        }
+        token_response = api_request(:post, 'tokens', post, {})
+        success = token_response['error'].nil?
+
+        if success && token_response['id']
+          Response.new(success, nil, token: token_response)
+        else
```
Heavyblade, line 348: ❓  Question:
 Is possible to have a success response that doesn't have an 'id' key ?, in that case we are going to return an error with out a message

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -35,7 +35,43 @@ def setup
       verification_value: '737',
       month: 10,
       year: 2028)
-    @apple_pay = apple_pay_payment_token
+
+    @google_pay = network_tokenization_credit_card(
+      '4242424242424242',
+      payment_cryptogram: 'reddelicious',
```
Heavyblade, line 41: 🛑  Warning:

Even when this works, I think is a better idea to use a random Base64 string, is not clear at all that just out of luck 'reddelicious' is a valid Base64 encoded string.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -299,7 +320,34 @@ def extract_token_from_string_and_maybe_add_customer_id(post, payment_method)
...
+        tokenization_method = options[:tokenization_method] || 'android_pay'
+        tokenization_method = 'apple_pay' if payment.source.to_sym == :apple_pay
+        post = {
+          card: {
+            number: payment.number,
+            exp_month: payment.month,
+            exp_year: payment.year,
+            tokenization_method: tokenization_method,
+            eci: payment.eci,
+            cryptogram: payment.payment_cryptogram
```
Heavyblade, line 340: ❓ Question:

Are we assuming that the payment_cryptogram is already being Base64 Encoded?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -186,10 +198,13 @@ def refund(money, intent_id, options = {})
       def store(payment_method, options = {})
         params = {}
         post = {}
-
         # If customer option is provided, create a payment method and attach to customer id
         # Otherwise, create a customer, then attach
-        if payment_method.is_a?(StripePaymentToken) || payment_method.is_a?(ActiveMerchant::Billing::CreditCard)
+        if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard)
```
Heavyblade, line 203: 🛑  Warning:

Seems that we have a big issue here, tokenization and storing payment methods are different things on Stripe, tokenization is just a temporary storage (One time usage) for a particular payment method and vaulting a payment method on stripe terms means save payment method into a customer object to be use later.

So here we are adding code to just tokenized a payment method to be used only one time (just for the particular payment intent), we should move this code into the authorize/purchase flow where it belongs.

Heavyblade (commented):
@gasb150 / @jherreraa / @javierpedrozaing some comments for your consideration

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -186,10 +198,13 @@ def refund(money, intent_id, options = {})
...
+          tokenize_apple_google = tokenize_apple_google(payment_method, options)
```
Heavyblade, line 204: 🛑  Warning:

naming a variable with the same name that a method can cause some issues with code execution or IDE tooling.

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -64,13 +88,80 @@ def test_successful_purchase
...
+  def test_successful_purchase_with_apple_pay
+    options = {
+      currency: 'GBP'
+    }
+
+    purchase = @gateway.purchase(@amount, @apple_pay, options)
+    assert purchase.success?
+  end
+
+  def test_successful_authorize_and_void_google_pay
```
Heavyblade, line 140: 💬  Comment:

I'm not sure if is worth checking this kind of post transaction test, I mean, no matter if the initial payment was performed with Google pay or Apple pay, the transaction was charged to the underlying real credit card, to so void and refund should work in the same as if the payment was made directly with a Credit Card.

Heavyblade (commented):
Nice refactor @gasb150, leaved some comments for your consideration

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -11,43 +11,44 @@ class StripePaymentIntentsGateway < StripeGateway
...
-        add_stored_credentials(post, options)
-        add_ntid(post, options)
-        add_claim_without_transaction_id(post, options)
-        add_error_on_requires_action(post, options)
-        add_fulfillment_date(post, options)
-        request_three_d_secure(post, options)
-
-        CREATE_INTENT_ATTRIBUTES.each do |attribute|
-          add_whitelisted_attribute(post, options, attribute)
+        MultiResponse.run do |r|
```
curiousepic, line 16: I'm trying to figure out if changing from a single `Response` to a `MultiResponse` for existing calls could have side effects for library users.  We could instead check for `NetworkTokenizationCreditCard` before determining which type of response we create, but that means two different types could be returned from the same call.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -86,6 +87,13 @@ def add_payment_method_data(payment_method, options = {})
         post_data
       end
 
+      def add_payment_method_card_data_tok(post_data, payment_method)
```
curiousepic, line 90:
```suggestion
      def add_payment_method_card_data_token(post_data, payment_method)
```
We'd prefer to avoid abbreviations in method names unless they are super long. Keeps things readable and explicit.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
curiousepic: Running our system tests using this branch didn't detect any issues; I think `MultiResponse` is smart enough to handle this change gracefully.

test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -64,13 +88,58 @@ def test_successful_purchase
...
+  def test_successful_purchase_with_apple_pay
+    options = {
+      currency: 'GBP'
+    }
+
+    purchase = @gateway.purchase(@amount, @apple_pay, options)
+    assert_match('apple_pay', purchase.responses.first.params.dig('token', 'card', 'tokenization_method'))
+    assert purchase.success?
+    assert_match('apple_pay', purchase.params.dig('charges', 'data')[0]['payment_method_details']['card']['wallet']['type'])
   end
```
curiousepic, line 143: Could  we add a test case for the tokenization step failing?

curiousepic:
@gasb150 Could we add a test case for the tokenization step failing?

curiousepic:
@gasb150 Looks good! Thanks for the test addition. Could you please squash the two commits before I merge?

curiousepic:
Closed with 20447eb4160d7fe04ac10fec9b13156f153f22a4

PR #4254, opened 2022-01-05
IPG: Add redact

ajawadmirza:
Added `unstore` method in IPG implementation to delete vaulted payment
method.

CE-2283

Remote:
17 tests, 53 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
5021 tests, 74840 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/ipg.rb
+++ b/lib/active_merchant/billing/gateways/ipg.rb
@@ -16,6 +16,8 @@ module ActiveMerchant #:nodoc:
         'ARS' => '032'
       }
 
+      ACTION_REQUEST_ITEMS = %w(vault unstore)
+
       def initialize(options = {})
         requires!(options, :store_id, :user_id, :password, :pem, :pem_password)
         @credentials = options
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/ipg.rb
```diff
@@ -62,6 +64,13 @@ def store(credit_card, options = {})
         commit('vault', xml)
       end
 
+      def redact(hosted_data_id)
```
jessiagee, line 67: Rename this to `def unstore`

jessiagee (changes requested):

lib/active_merchant/billing/gateways/ipg.rb
ajawadmirza: Thanks, updated !

jessiagee (approved):
Looks good and all tests pass!

PR #4257, opened 2022-01-06
Cashnet: Transforms amounts into integers

peteroas:
This covers an edge-case where the list of `amounts` within the `item_codes` option is coverted into a string

### DIFF
--- a/lib/active_merchant/billing/gateways/cashnet.rb
+++ b/lib/active_merchant/billing/gateways/cashnet.rb
@@ -117,11 +117,11 @@ module ActiveMerchant #:nodoc:
           codes_and_amounts = options[:item_codes].transform_keys { |key| key.to_s.delete('_') }
           codes_and_amounts.each do |key, value|
             post[key] = value if key.start_with?('itemcode')
-            post[key] = amount(value) if key.start_with?('amount')
+            post[key] = amount(value.to_i) if key.start_with?('amount')
           end
         else
           post[:itemcode] = (options[:item_code] || @options[:default_item_code])
-          post[:amount] = amount(money)
+          post[:amount] = amount(money.to_i)
         end
       end
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -68,6 +68,7 @@
 * Kushki: Add support for `metadata` [rachelkirk] #4253
 * IPG: Add `redact` operation [ajawadmirza] #4254
 * Wompi: Update sandbox and production endpoints [rachelkirk] #4255
+* Cashnet: convert amounts to integers for proper gateway handling [peteroas] #2207
 
 == Version 1.124.0 (October 28th, 2021)
 * Worldpay: Add Support for Submerchant Data on Worldpay [almalee24] #4147


lib/active_merchant/billing/gateways/cashnet.rb
```diff
@@ -117,11 +117,11 @@ def add_invoice(post, money, options)
           codes_and_amounts = options[:item_codes].transform_keys { |key| key.to_s.delete('_') }
           codes_and_amounts.each do |key, value|
             post[key] = value if key.start_with?('itemcode')
-            post[key] = amount(value) if key.start_with?('amount')
+            post[key] = amount(value.to_i) if key.start_with?('amount')
```
meagabeth, line 120: The amount method seems to be then transforming this back into a string 🤦🏻  This was discovered by adding a puts statement to confirm the class. I think instead that the fix would be `amount(money).to_i`

lib/active_merchant/billing/gateways/cashnet.rb
```diff
@@ -117,11 +117,11 @@ def add_invoice(post, money, options)
...
           end
         else
           post[:itemcode] = (options[:item_code] || @options[:default_item_code])
-          post[:amount] = amount(money)
+          post[:amount] = amount(money.to_i)
```
meagabeth, line 124: Same as above: The amount method seems to be then transforming this back into a string 🤦🏻  This was discovered by adding a puts statement to confirm the class. I think instead that the fix would be `amount(money).to_i`

meagabeth:
The requested changes above will then result in a few failing remote tests due to a different error message being sent so those remote tests will need to be edited to reflect the new error message coming back.

meagabeth (changes requested):
A couple small changes and then edits to tests -- as discussed in Slack, we'll pair on this and knock it out together.

lib/active_merchant/billing/gateways/cashnet.rb
peteroas: so the Cashnet gateway actually WANTS this in a string, with cents included such as this: `1001` converts to `'10.01'`. The reason we made this conversion is because when we were passing a hash with nested `amount` and `item_code`s in it, the amounts were getting converted to strings, which WOULD be fine if we also didn't need to transform their amounts into what Cashnet expects.

We can discuss in pairing, but basically slapping the `to_i` method onto the money value is ensuring that IF money is passed as a string (which it shouldn't be), than we convert it back into an integer before running the `amount` method on it.

meagabeth: Ah, thank you for the explanation! This is great. There are 2 failing remote tests then, but they are also happening on the `master` branch so you're good to go!

meagabeth (approved):
Great work, thanks for the explanation for this use case! 2 failing remote tests are also failing on `master` so they are unrelated. Good to go!

lib/active_merchant/billing/gateways/cashnet.rb
peteroas: yeah, those darn remote tests :/ we noticed those two, but decided to kick that can down the road 😂 thanks for your feedback!!

PR #4260, opened 2022-01-07
Add metadata to network_tokenization_credit_card

mark-sim:
Some payment gateways require attributes that cannot currently
be instantiated with network_tokenization_credit_card.

For example - some gateways require `deviceManufacturerIdentifer` and
`paymentDataType` to be passed in for apple pay which can now be stored
in the metadata.

--------------
How `metadata` can be used - I did a quick 📖 and while `eci`, `payment_cryptogram`, `transaction_id` are common across wallets (hence we have these attributes in the model), there are some [apple-pay specific fields](https://developer.apple.com/library/archive/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.html) such as `deviceManufacturerIdentifer` and `paymentDataType` and [google-pay specific fields](https://developers.google.com/pay/api/android/guides/resources/payment-data-cryptography) such as `authMethod`, `paymentMethod`. 

With `metadata`, you don't have to dump all these provider specific attributes and you have access to these data if a payment gateway requires them.

### DIFF
--- a/lib/active_merchant/billing/network_tokenization_credit_card.rb
+++ b/lib/active_merchant/billing/network_tokenization_credit_card.rb
@@ -14,7 +14,7 @@ module ActiveMerchant #:nodoc:
       self.require_verification_value = false
       self.require_name = false
 
-      attr_accessor :payment_cryptogram, :eci, :transaction_id
+      attr_accessor :payment_cryptogram, :eci, :transaction_id, :metadata
       attr_writer :source
 
       SOURCES = %i(apple_pay android_pay google_pay network_token)
--- a/test/unit/network_tokenization_credit_card_test.rb
+++ b/test/unit/network_tokenization_credit_card_test.rb
@@ -5,7 +5,8 @@ class NetworkTokenizationCreditCardTest < Test::Unit::TestCase
     @tokenized_card = ActiveMerchant::Billing::NetworkTokenizationCreditCard.new({
       number: '4242424242424242', brand: 'visa',
       month: default_expiration_date.month, year: default_expiration_date.year,
-      payment_cryptogram: 'EHuWW9PiBkWvqE5juRwDzAUFBAk=', eci: '05'
+      payment_cryptogram: 'EHuWW9PiBkWvqE5juRwDzAUFBAk=', eci: '05',
+      metadata: { device_manufacturer_id: '1324' }
     })
     @tokenized_apple_pay_card = ActiveMerchant::Billing::NetworkTokenizationCreditCard.new({
       source: :apple_pay


mark-sim:
Based on the conversation : 
>  Mark Sim  3 hours ago
:wave: everyone, not sure if this is the right channel to ask since it's not very active but redirect me to the right one if it isn't!
I was reading through apple pay token reference and it is documented that the decrypted data contains paymentDataType and deviceManufacturerIdentifier . However, looking at the ActiveMerchant::Billing::NetworkTokenizationCreditCard,  there's no way we can instantiate this object with these two attributes.
Before opening a PR to add these attributes, is there a recommended way to instantiate this object with these attributes?

> David Perry  2 hours ago
I don't think Spreedly does anything with those fields (yet), so I don't think so. However I did just realize that ApplePayPaymentToken also exists. I'm not sure about your use case but that might be a more suitable place for them? (edited) 

> David Perry  2 hours ago
It doesn't look widely used though, so may be stale (edited) 

> Jacob Krall  2 hours ago
Yeah, agree with @dperry – Spreedly is storing both of these fields in our data store but not using them anywhere we can see. (edited) 

> Mark Sim  2 hours ago
Thanks! I'll take a look and get back :eyes:

> Mark Sim  2 hours ago
Hmm, ApplePayPaymentToken seems to store everything in the metadata .
For our use case, NetworkTokenizationCreditCard is still the preferred approach. I know you guys are not using those fields yet but would it be ok for me to add a simple PR to add those two attr_accessors?
If these seem too specific for apple pay, we could add a metadata in this model instead which'd be useful. (edited) 

> Jacob Krall  1 hour ago
Yeah, either a metadata hash, or a subclass. What’s your use case for including it in AM? Is there a gateway that needs these fields?

InPermutation:
Hi @mark-sim, thanks for checking in with us at `org:spreedly`!

Here are my thoughts from over the weekend:

Instead of using a `Hash`, I would prefer `ApplePayNetworkTokenizationCreditCard` and `GooglePayNetworkTokenizationCreditCard` subclasses with attribute accessors for mobile-pay-brand-specific metadata, so we get some type safety. (e.g. `NoMethodError` instead of `nil` if you typo an attr name, better IDE autocomplete hints, and compile-time safety if we decide to add [Sorbet](https://sorbet.org/) to ActiveMerchant)

Also, can you please update the tests to verify this new approach? Specifically `test/unit/network_tokenization_credit_card_test.rb` should assert the behavior of the new subclasses and their accessors, but it would also be good to update things like `test/unit/gateways/litle_test.rb` and `test/remote/gateways/remote_litle_test.rb`, to take advantage of the existing separate test arms for (Apple|Google) Pay.

mark-sim:
Closed this in favor of https://github.com/activemerchant/active_merchant/pull/4265

PR #4264, opened 2022-01-10
Decidir Plus: Add Gateway Adapter

naashton:
This is a PR to add the gateway adapter for Decidir Plus

Things to note in this PR:
* No auth/capture or void/cancel transactions
* Payment Method's must be `stored` and the returned third party token
  will be used in combination with other card data in subsequent
`purchase` and `refund` transactions
* The reference token is comprised of the `stored` payment method's `id`
  (returned from Decidir+) and the `bin` of the payment method,
separated by a pipe (`|`)
* Authorization is handled by an `apikey` passed into the headers.
  Either the `public_key` for the tokenization endpoint or `private_key`
for the rest
* Some fields that are stubbed in with defaults ARE REQUIRED in order
  for the transaction to be successful (i.e., `purchase` will fail
without `sub_payments` array)

CE-2145

Unit: 6 tests, 20 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 8 tests, 27 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir_plus.rb
+++ b/lib/active_merchant/billing/gateways/decidir_plus.rb
@@ -1,0 +1,147 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class DecidirPlusGateway < Gateway
+      self.test_url = 'https://developers.decidir.com/api/v2'
+      self.live_url = 'https://live.decidir.com/api/v2'
+
+      self.supported_countries = ['ARG']
+      self.default_currency = 'ARS'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
+      self.homepage_url = 'http://decidir.com.ar/home'
+      self.display_name = 'Decidir Plus'
+
+      def initialize(options = {})
+        requires!(options, :public_key, :private_key)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        post = {}
+        post[:site_transaction_id] = options[:site_transaction_id] || SecureRandom.hex
+        post[:token] = options[:payment_id].split('|')[0]
+        post[:payment_method_id] = 1
+        post[:bin] = payment.number.to_s[0..5]
+        post[:amount] = money
+        post[:currency] = options[:currency] || self.default_currency
+        post[:installments] = options[:installments] || 1
+        post[:payment_type] = options[:payment_type] || 'single'
+        post[:sub_payments] = []
+
+        commit(:post, 'payments', post)
+      end
+
+      def refund(money, authorization, options = {})
+        post = {}
+        post[:amount] = money
+
+        commit(:post, "payments/#{add_reference(authorization)}/refunds")
+      end
+
+      def store(payment, options = {})
+        post = {}
+        add_payment(post, payment, options)
+
+        commit(:post, 'tokens', post)
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Apikey: )\w+), '\1[FILTERED]').
+          gsub(%r(("card_number\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("security_code\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
+      end
+
+      private
+
+      def add_reference(authorization)
+        return unless authorization
+
+        authorization.split('|')[0]
+      end
+
+      def add_payment(post, payment, options = {})
+        post[:card_number] = payment.number
+        post[:card_expiration_month] = payment.month.to_s.rjust(2, '0')
+        post[:card_expiration_year] = payment.year.to_s[-2..-1]
+        post[:security_code] = payment.verification_value.to_s
+        post[:card_holder_name] = payment.name
+        post[:card_holder_identification] = {}
+        post[:card_holder_identification][:type] = options[:dni]
+        post[:card_holder_identification][:number] = options[:card_holder_identification_number]
+      end
+
+      def parse(body)
+        JSON.parse(body)
+      end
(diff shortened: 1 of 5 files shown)


jessiagee:
@ajawadmirza For some reason I can't add you as a reviewer but can you give this a lookover? 🙇🏻‍♀️

lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -0,0 +1,151 @@
...
+      self.display_name = 'Decidir Plus'
+
+      def initialize(options = {})
+        requires!(options, :public_key, :private_key)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        post = {}
+        post[:site_transaction_id] = options[:site_transaction_id]
```
meagabeth, line 21: Regarding the site_transaction_id, since it is a required field, have we considered creating one via SecureRandom.hex if one is not provided from the merchant?

lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -0,0 +1,151 @@
...
+      private
+
+      def add_reference(authorization)
+        return unless authorization
+
+        authorization.split('|')[0]
+      end
+
+      def add_invoice(post, money, options)
+        post[:amount] = amount(money)
```
meagabeth, line 68: Bringing this up since it’s happened a few times lately, does the gateway expect the amount to always be an integer? If so, it may be worthwhile to throw in a `.to_i` now and avoid the errors that might pop up in core integration tests later on.

meagabeth (changes requested):
Awesome start to this gateway, especially considering the interesting mechanics of it all. I've added a couple comments for consideration. I also got 5 rubocop errors when running local tests that will need to be resolved. Thanks!

lib/active_merchant/billing/gateways/decidir_plus.rb
naashton: Good catch, I think I left this out thinking we might pass in a GSF and handle the randomly generated value in Core, but the behavior makes it clear here what the gateway expects.

lib/active_merchant/billing/gateways/decidir_plus.rb
naashton: I'm actually not calling this method, so I just removed it. The `amount` method converts the integer representation of the amount to a 2 decimal string representation of the amount, but Decidir+ is just looking for that integer representation.

test/remote/gateways/remote_decidir_plus_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(payment_id: response.authorization))
+    assert_success response
+    assert_equal 'approved', response.message
+  end
+
+  def test_failed_purchase
+    assert response = @gateway.store(@credit_card)
+
+    @options[:site_transaction_id] = SecureRandom.hex
```
meagabeth, line 28: This can be removed since the gateway does this as part of the purchase method now

test/remote/gateways/remote_decidir_plus_test.rb
```diff
@@ -0,0 +1,90 @@
...
+    response = @gateway.purchase(@amount, @declined_card, @options.merge(payment_id: response.authorization))
+    assert_failure response
+    assert_equal 'invalid_param: bin', response.message
+  end
+
+  def test_successful_refund
+    response = @gateway.store(@credit_card)
+
+    @options[:site_transaction_id] = SecureRandom.hex
```
meagabeth, line 37: Same as above

test/remote/gateways/remote_decidir_plus_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    assert refund = @gateway.refund(@amount, purchase.authorization)
+    assert_success refund
+    assert_equal 'approved', refund.message
+  end
+
+  def test_partial_refund
+    assert response = @gateway.store(@credit_card)
+
+    @options[:site_transaction_id] = SecureRandom.hex
```
meagabeth, line 50: Same here

test/unit/gateways/decidir_plus_test.rb
```diff
@@ -0,0 +1,155 @@
...
+  include CommStub
+
+  def setup
+    @gateway = DecidirPlusGateway.new(public_key: 'public_key', private_key: 'private_key')
+    @credit_card = credit_card
+    @amount = 100
+
+    @options = {
+      payment_id: '2bf7bffb-1257-4b45-8d42-42d090409b8a|448459',
+      site_transaction_id: SecureRandom.hex,
```
meagabeth, line 13: This can be removed since the gateway does this as part of the purchase method now

lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -0,0 +1,146 @@
...
+      self.display_name = 'Decidir Plus'
+
+      def initialize(options = {})
+        requires!(options, :public_key, :private_key)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        post = {}
+        post[:site_transaction_id] = options[:site_transaction_id] || SecureRandom.hex
```
meagabeth, line 21: Sweet!

meagabeth (changes requested):
Thanks for making those adjustments! Since you added the built in to create the site id if it's not provided, it can be removed from various test file lines so I added comments to those places where I saw it. I'm seeing 5 rubocop complaints, too.

meagabeth (approved):
BOOM 💥 Looks great and tests are all passing! Go forth! :shipit:

PR #4265, opened 2022-01-10
Add `ApplePayNetworkTokenizationCreditCard` and `GooglePayNetworkTokenizationCreditCard`

mark-sim:
As mentioned in this [comment](https://github.com/activemerchant/active_merchant/pull/4260#issuecomment-1009013750), we'll be adding two new subclasses `ApplePayNetworkTokenizationCreditCard` and `GooglePayNetworkTokenizationCreditCard`
> Hi @mark-sim, thanks for checking in with us at `org:spreedly`!
> 
> Here are my thoughts from over the weekend:
> 
> Instead of using a `Hash`, I would prefer `ApplePayNetworkTokenizationCreditCard` and `GooglePayNetworkTokenizationCreditCard` subclasses with attribute accessors for mobile-pay-brand-specific metadata, so we get some type safety. (e.g. `NoMethodError` instead of `nil` if you typo an attr name, better IDE autocomplete hints, and compile-time safety if we decide to add [Sorbet](https://sorbet.org/) to ActiveMerchant)
> 
> Also, can you please update the tests to verify this new approach? Specifically `test/unit/network_tokenization_credit_card_test.rb` should assert the behavior of the new subclasses and their accessors, but it would also be good to update things like `test/unit/gateways/litle_test.rb` and `test/remote/gateways/remote_litle_test.rb`, to take advantage of the existing separate test arms for (Apple|Google) Pay.

These subclasses act exactly like `NetworkTokenizationCreditCard` with the ability to instantiate it with brand-specific metadata. i.e:
- `payment_data_type` and `device_manufacturer_id` for [apple pay](https://developer.apple.com/library/archive/documentation/PassKit/Reference/PaymentTokenJSON/PaymentTokenJSON.htm)
- `auth_method` and `payment_method` for [google pay](https://developers.google.com/pay/api/android/guides/resources/payment-data-cryptography#encrypted-message)

Doesn't do too much for now but we've already had a case where a payment gateway required some of these attributes.
Furthermore, if these wallets decide to add more brand-specific metadata in the future, these classes are the place to add them.


mark-sim:
Hi, can I get some 👀 on this PR?

This is a blocker for one of our projects.

InPermutation:
Hi Mark, sorry for the delay. These changes look good to me so far. I'm adding @shasum as a reviewer.

shasum (commented):
Late to the party here. @mark-sim  Thanks for looking to enhance Active Merchant's wallet types. Although this is a cleaner pattern, I'm leaning towards the earlier `metadata` approach for reasons.
1. A bunch of gateways rely on the `source` attribute of `NetworkTokenizationCreditCard` to identify the wallet payload, which is the existing pattern.
2. This change introduces a new pattern based on inheritance.
3. If not backported, we would have to deal with differing abstractions for the same wallet across gateways.
4. Could end up being confusing to future devs.
5. Also, systems that use Active Merchant gem would have to handle both types. Ex: Gateway-A uses NTCC with source=apple_pay, while Gateway-B uses APNTCC.

Sticking to one format across existing and future gateways appears a lesser burden. Open to other's thoughts.

mark-sim:
Hi @shasum thanks for the feedback.

I don't have a strong opinion on either approaches but yeah I understand your concern with this one.
I couldn't re-open the existing PR, so opened a [new one](https://github.com/activemerchant/active_merchant/pull/4270) instead.
Would appreciate it if you can take a look.

InPermutation:
Closed in favor of #4270

PR #4266, opened 2022-01-10
CyberSource: calculate_tax is not being called

jessiagee:
* The calculate_tax method is not currently being called which is causing line_items to fail


jessiagee:
@ajawadmirza for some reason I can't add you to this PR but here's a fix for the CyberSource issue. The `calculate_tax` method isn't currently called from anywhere so it wasn't setting the `line_items` information. I *think* it only needs to be called from authorize/purchase but please double check that for me.

Also, I'm making this a draft PR because updating this causes the unit/remote tests to fail so I'm going to leave that for you to fix. Thank you!

PR #4267, opened 2022-01-11
Paymentez: Add verify order functionality

ajawadmirza:
Added verify order functionality to approve pending transaction to
perform otp transaction flow in paymentez gateway. Tests to be performed with `ecaudor` fixtures.

CE-1920

Unit:
5022 tests, 74850 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
33 tests, 82 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
87.8788% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -34,6 +34,8 @@ module ActiveMerchant #:nodoc:
         28 => :card_declined
       }.freeze
 
+      SUCCESS_STATUS = ['success', 'pending', 1, 0]
+
       CARD_MAPPING = {
         'visa' => 'vi',
         'master' => 'mc',
@@ -77,13 +79,21 @@ module ActiveMerchant #:nodoc:
         commit_transaction('authorize', post)
       end
 
-      def capture(money, authorization, _options = {})
+      def capture(money, authorization, options = {})
         post = {
           transaction: { id: authorization }
         }
-        post[:order] = { amount: amount(money).to_f } if money
+        verify_flow = options[:type] && options[:value]
 
-        commit_transaction('capture', post)
+        if verify_flow
+          add_customer_data(post, options)
+          add_verify_value(post, options)
+        elsif money
+          post[:order] = { amount: amount(money).to_f }
+        end
+
+        action = verify_flow ? 'verify' : 'capture'
+        commit_transaction(action, post)
       end
 
       def refund(money, authorization, options = {})
@@ -139,10 +149,10 @@ module ActiveMerchant #:nodoc:
       private
 
       def add_customer_data(post, options)
-        requires!(options, :user_id, :email)
+        requires!(options, :user_id)
         post[:user] ||= {}
         post[:user][:id] = options[:user_id]
-        post[:user][:email] = options[:email]
+        post[:user][:email] = options[:email] if options[:email]
         post[:user][:ip_address] = options[:ip] if options[:ip]
         post[:user][:fiscal_number] = options[:fiscal_number] if options[:fiscal_number]
         if phone = options[:phone] || options.dig(:billing_address, :phone)
@@ -177,6 +187,11 @@ module ActiveMerchant #:nodoc:
           post[:card][:cvc] = payment.verification_value
           post[:card][:type] = CARD_MAPPING[payment.brand]
         end
+      end
+
+      def add_verify_value(post, options)
+        post[:type] = options[:type] if options[:type]
+        post[:value] = options[:value] if options[:value]
       end
 
       def add_extra_params(post, options)
@@ -262,7 +277,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(response)
-        !response.include?('error') && (response['status'] || response['transaction']['status']) == 'success'
+        return false if response.include?('error')
+
+        SUCCESS_STATUS.include?(response['status'] || response['transaction']['status'])
       end
 
       def card_success_from(response)
@@ -278,7 +295,7 @@ module ActiveMerchant #:nodoc:
         if !success_from(response) && response['error']
           response['error'] && response['error']['type']
         else
-          response['transaction'] && response['transaction']['message']
+          (response['transaction'] && response['transaction']['message']) || (response['message'])
         end
       end
 
--- a/test/remote/gateways/remote_paymentez_test.rb
+++ b/test/remote/gateways/remote_paymentez_test.rb
@@ -3,9 +3,11 @@ require 'test_helper'
 class RemotePaymentezTest < Test::Unit::TestCase
   def setup
     @gateway = PaymentezGateway.new(fixtures(:paymentez))
+    @ecuador_gateway = PaymentezGateway.new(fixtures(:paymentez_ecuador))
 
     @amount = 100
     @credit_card = credit_card('4111111111111111', verification_value: '666')
+    @otp_card = credit_card('36417002140808', verification_value: '666')
     @elo_credit_card = credit_card('6362970000457013',
       month: 10,
       year: 2022,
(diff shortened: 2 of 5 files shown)


test/remote/gateways/remote_paymentez_test.rb
```diff
@@ -255,6 +256,16 @@ def test_failed_capture
     assert_equal 'The modification of the amount is not supported by carrier', response.message
   end
 
+  def test_successful_capture_with_otp
+    @options[:vat] = 0.1
+    response = @gateway.purchase(@amount, @otp_card, @options)
+    assert_equal 'pending', response.params['transaction']['status']
```
dsmcclain, line 262: Thank you for the helpful note about the Ecuador fixtures. This test passes when I use the Ecuador fixtures, but a lot of other tests fail.

Therefore I think it would be a good idea to define a second set of fixtures in this test file (maybe called `paymentez_ecuador`). Then we could create an `ecuador_gateway` in this file and use that for this test. We could use our standard Brazil fixtures for the rest of the tests.

See [here](https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_mercado_pago_test.rb#L7) for an example of what I mean.

dsmcclain (changes requested):
@ajawadmirza nice work on a complex transaction type.

I think there are a few changes we need to consider (see comments below).

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -77,13 +77,20 @@ def authorize(money, payment, options = {})
...
 
-        commit_transaction('capture', post)
+        if money
+          post[:order] = { amount: amount(money).to_f }
+        else
+          add_customer_data(post, options)
+          add_verify_value(post, options)
+        end
+
+        action = money ? 'capture' : 'verify'
```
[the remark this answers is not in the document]
ajawadmirza: Thanks, changes are added along with nil amount test case.

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -77,13 +77,21 @@ def authorize(money, payment, options = {})
...
         post = {
           transaction: { id: authorization }
         }
-        post[:order] = { amount: amount(money).to_f } if money
+        verify_flow = options[:type] && options[:value]
+
+        if verify_flow
+          add_customer_data(post, options)
+          add_verify_value(post, options)
+        elsif money
```
ajawadmirza, line 89: `elsif` is added here instead of `else` so that the capture can work in case of **nil** amount as well.

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -77,13 +77,21 @@ def authorize(money, payment, options = {})
         commit_transaction('authorize', post)
       end
 
-      def capture(money, authorization, _options = {})
+      def capture(money, authorization, options = {})
         post = {
           transaction: { id: authorization }
         }
-        post[:order] = { amount: amount(money).to_f } if money
+        verify_flow = options[:type] && options[:value]
```
dsmcclain, line 84: I think this is good, because it conforms to the fields required by the Paymentez API.

One thing I wonder about is whether a combined field, like `options[:otp_value]`, that sets `post[:type] = 'BY_OTP'` and `post[:value] = options[:otp_value]` would be a benefit for the customer in this use case? It could make things more simple for them to handle on their end, because they only have to worry about passing the OTP as `otp_value` and then we do the rest.

☝️ I thought for a long time about this and I decided that the way you have it right now makes sense. I am just trying to think about things from a customer's perspective and what is easiest for us to maintain in the long run.

dsmcclain (changes requested):
Good work on this @ajawadmirza. This one is complicated and I think you are almost there.

I would like you to check whether or not we want active merchant to treat a `status: pending` transaction as a failure. I think this will be important down the line as we figure out how to handle these complex transactions within Spreedly.

I also added a comment about an alternative to the `value` and `type` fields, just in case you find the discussion interesting. There are no changes needed to that part of the code at this time.

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -262,7 +275,7 @@ def headers
       end
 
       def success_from(response)
-        !response.include?('error') && (response['status'] || response['transaction']['status']) == 'success'
+        !response.include?('error') && ((response['status'] || response['transaction']['status']) == 'success' || response['status'] == 1)
```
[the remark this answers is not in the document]
ajawadmirza: Good catch, we didn't have this `pending` transaction before so let me add the logic to treat them as success.

lib/active_merchant/billing/gateways/paymentez.rb
ajawadmirza: Yes, I had thought of this approach but the reason I am not in the favor of this approach is that the paymentez API has more `type` & `value` pairs other than just **BY_OTP**. so if we will have this fixed otp_value in the options then we will need to manually set and fix other [4 options for these pairs](https://paymentez.github.io/api-doc/#payment-methods-cards-verify) as well and also in future if new types are added in the API we will manually need to update this in our implementation as well, so what are your suggestions on this?

dsmcclain: Yes, that is a very good point. I think the way you have it works just fine.

But, if we wanted to preserve the ability for a merchant to pass in a different `type`, we could combine both approaches, like this:

```
verify_flow = verify_flow?(options)

# later in the class
def verify_flow?(options)
  options[:otp_value] || (options[:type] && options[:value])
end

# even later in the class
def add verify_value(post, options)
  type, value = if options[:otp_value]
    ['BY_OTP', options[:otp_value]]
  else
    [options[:type], options[:value]]
  end

  post[:type] = type if type
  post[:value] = value if value
end
```
This way, we could add the convenience `otp_value` for our primary use-case, but if somebody in the future wanted to use this verify flow for a different purpose (say for example passing in a `type` of `'BY_AMOUNT'`), then the adapter already supports that.

I'm not sure if this is better or not. It is nice to have `otp_value` to make it easier to use for the customer, but it also introduces multiple options which could be confusing, and it makes the code a little more complex. So it's a judgement call. What do you think?

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -262,7 +277,7 @@ def headers
       end
 
       def success_from(response)
-        !response.include?('error') && (response['status'] || response['transaction']['status']) == 'success'
+        !response.include?('error') && (SUCCESS_STATUS.include?(response['status'] || response['transaction']['status']) || response['status'] == 1)
```
dsmcclain, line 280: I like the way you've used a constant to define the statuses we are looking for 👍 

It seems like we also want to include `response['status'] == 0` just to be safe. I know the test transaction gives us `response['status']: pending` but the API documentation suggests that the value could also be `0`.

You could use the constant you have created to do this, and simplify the `success_from` method at the same time:

```
SUCCESS_STATUSES = ['success', 'pending', 1, 0]

def success_from(response)
  return false if response.include?('error')
 
  SUCCESS_STATUSES.include?(response['status'] || response['transaction']['status'])
end
```

dsmcclain (changes requested):
Thank you for your continued effort on this one!

I have one more change requested on the `success_from` method.

I also added a comment about the possibility of an `otp_value` field. I will leave that one up to your judgement, I'm happy with it either way.

lib/active_merchant/billing/gateways/paymentez.rb
ajawadmirza: In my opinion, this may add extra complexity to the code and more importantly this may make our request bit inconsistent with the original parameters being passed in the API and may create some confusion for the customers as it also creates some separation in logic for otp and other type value pairs without some documentation on API. If you suggest to go with this approach then I am also good to go on this as well 😄

lib/active_merchant/billing/gateways/paymentez.rb
ajawadmirza: thanks, updated 👍

lib/active_merchant/billing/gateways/paymentez.rb
dsmcclain: Ok, I like your reasoning. Let's leave it the way you have it now 👍

dsmcclain (approved):
Excellent work @ajawadmirza!

I get the same 4 failing remote tests as you. Otherwise everything passes. I will merge later today.

PR #4274, opened 2022-01-14
Decidir Plus: Sub Payment Fields

naashton:
Add `sub_payment` fields to `purchase` transactions. This field needs to
be set in the post body, regardless of whether or not it is populated
with data.

Fix country code for this gateway.

CE-2145

Unit: 7 tests, 25 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 9 tests, 30 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir_plus.rb
+++ b/lib/active_merchant/billing/gateways/decidir_plus.rb
@@ -4,7 +4,7 @@ module ActiveMerchant #:nodoc:
       self.test_url = 'https://developers.decidir.com/api/v2'
       self.live_url = 'https://live.decidir.com/api/v2'
 
-      self.supported_countries = ['ARG']
+      self.supported_countries = ['AR']
       self.default_currency = 'ARS'
       self.supported_cardtypes = %i[visa master american_express discover]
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -82,7 +82,23 @@ def add_purchase_data(post, money, payment, options = {})
         post[:currency] = options[:currency] || self.default_currency
         post[:installments] = options[:installments] || 1
         post[:payment_type] = options[:payment_type] || 'single'
+        add_sub_payments(post, options)
+      end
+
+      def add_sub_payments(post, options)
+        # sub_payments field is required for purchase transactions, even if empty
```
naashton, line 89: Let me know if this comment is unnecessary. I just want people to be aware that this field must be set for transactions against the gateway to be successful.

jessiagee: @naashton I'm always down for more comments especially for required fields. I say leave it.

jessiagee (approved):
Looks good to me and all tests pass!!

PR #4275, opened 2022-01-14
GWI-68 Allow $0 auth Orbital Gateway

javierpedrozaing:
Orbital Gateway: Add support to $0 verify

DESCRIPTION
Added verify with $0 amount into Orbital Gateway
Discover card type does not  support $0.00 auth instead use $1.00

JIRA TICKET NUMBER
GWI-71

UNIT TEST
Finished in 0.661291 seconds.

135 tests, 779 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

129.10 tests/s, 750.16 assertions/s

REMOTE TEST
Finished in 211.728643 seconds.

80 tests, 355 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

0.38 tests/s, 1.68 assertions/s

RUBOCOP
726 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -220,8 +220,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def verify(creditcard, options = {})
+        amount = allow_0_auth?(creditcard) ? 0 : 100
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, creditcard, options) }
+          r.process { authorize(amount, creditcard, options) }
           r.process(:ignore_result) { void(r.authorization) }
         end
       end
@@ -274,6 +275,11 @@ module ActiveMerchant #:nodoc:
 
         order = build_void_request_xml(authorization, options)
         commit(order, :void, options[:retry_logic], options[:trace_number])
+      end
+
+      def allow_0_auth?(credit_card)
+        # Discover does not support a $0.00 authorization instead use $1.00
+        %w(visa master american_express diners_club jcb).include?(credit_card.brand)
       end
 
       # ==== Customer Profiles
--- a/test/remote/gateways/remote_orbital_test.rb
+++ b/test/remote/gateways/remote_orbital_test.rb
@@ -908,7 +908,28 @@ class RemoteOrbitalGatewayTest < Test::Unit::TestCase
   def test_successful_verify
     response = @gateway.verify(@credit_card, @options)
     assert_success response
+    assert_equal 'No reason to decline', response.message
+  end
+
+  def test_successful_different_cards
+    @credit_card.brand = 'master'
+    response = @gateway.verify(@credit_card, @options)
+    assert_success response
+    assert_equal 'No reason to decline', response.message
+  end
+
+  def test_successful_verify_with_discover_brand
+    @credit_card.brand = 'discover'
+    response = @gateway.verify(@credit_card, @options)
+    assert_success response
     assert_equal 'Approved', response.message
+  end
+
+  def test_unsuccessful_verify_with_invalid_discover_card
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -276,6 +277,11 @@ def void(authorization, options = {}, deprecated = {})
         commit(order, :void, options[:retry_logic], options[:trace_number])
       end
 
+      def allow_0_auth?(credit_card, options = {})
```
Heavyblade, line 280: ❓  Question:

Why passing the options hash if you are not using it inside the method ?

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -220,8 +220,9 @@ def authorize(money, payment_source, options = {})
       end
 
       def verify(creditcard, options = {})
+        amount = allow_0_auth?(creditcard, options) ? 0 : 1
```
Heavyblade, line 223: ❓ Question:

Maybe I'm wrong but as far as I remember values are passed as cents, that's why the initial value was 100, not 1, perhaps the values for the ternary operation should be 0 and 100 ?

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -906,11 +906,33 @@ def test_void_transactions
...
 
   def test_successful_verify
+    response = @gateway.verify(@credit_card, @options)
+    assert_success response
+    assert_equal 'No reason to decline', response.message
+  end
+
+  def test_successful_different_cards
+    cards = %w(visa master american_express diners_club jcb)
+    @credit_card.brand = cards.sample
```
Heavyblade, line 916: 💬  Comment:

By using sample here you are adding randomness to the test, maybe is an edge case but let's say that jcb is failing for some reason, in that case, some tests execution are going to fail and others not.

test/unit/gateways/orbital_test.rb
```diff
@@ -1492,6 +1492,56 @@ def test_successful_verify
     assert_equal 'Approved', response.message
   end
 
+  def test_valid_amount_with_different_cards
+    cards = %w(visa master american_express diners_club jcb)
+    @credit_card.brand = cards.sample
```
Heavyblade, line 1497: 💬  Comment:

Same comment about randomnes here

test/unit/gateways/orbital_test.rb
```diff
@@ -1492,6 +1492,56 @@ def test_successful_verify
...
+    assert_equal 'Approved', response.message
+  end
+
+  def test_valid_amount_with_discover_brand
+    @credit_card.brand = 'discover'
+    stub_comms do
+      @gateway.verify(@credit_card, @options)
+    end.check_request(skip_response: true) do |_endpoint, data, _headers|
+      assert_match %r{<Amount>1<\/Amount>}, data
+    end
```
Heavyblade, line 1522: so nice to use this without the unneeded response 🤜 🤛

Heavyblade (commented):
Great job @javierpedrozaing, I left some comments and questions.

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: Thx, you're righ, fixed

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: Thx, you're righ, fixed

Heavyblade (approved):
LGTM

curiousepic:
Looks good; I agree with @Heavyblade's concerns and appreciate their resolution! Will merge soon.

curiousepic (approved):

PR #4284, opened 2022-01-20
Decidir Plus: Add supported methods

ajawadmirza:
Added `authorize`, `capture`, `void`, `verify` methods for decidir plus
along with its test cases.

CE-2159

Remote:
15 tests, 51 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
5059 tests, 75068 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir_plus.rb
+++ b/lib/active_merchant/billing/gateways/decidir_plus.rb
@@ -18,19 +18,42 @@ module ActiveMerchant #:nodoc:
 
       def purchase(money, payment, options = {})
         post = {}
-
-        add_payment(post, payment, options)
-        add_purchase_data(post, money, payment, options)
-        add_fraud_detection(post, options)
+        build_purchase_authorize_request(post, money, payment, options)
 
         commit(:post, 'payments', post)
+      end
+
+      def authorize(money, payment, options = {})
+        post = {}
+        build_purchase_authorize_request(post, money, payment, options)
+
+        commit(:post, 'payments', post)
+      end
+
+      def capture(money, authorization, options = {})
+        post = {}
+        post[:amount] = money
+
+        commit(:put, "payments/#{add_reference(authorization)}", post)
       end
 
       def refund(money, authorization, options = {})
         post = {}
         post[:amount] = money
 
+        commit(:post, "payments/#{add_reference(authorization)}/refunds", post)
+      end
+
+      def void(authorization, options = {})
         commit(:post, "payments/#{add_reference(authorization)}/refunds")
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { store(credit_card, options) }
+          r.process { authorize(100, r.authorization, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
       end
 
       def store(payment, options = {})
@@ -53,6 +76,12 @@ module ActiveMerchant #:nodoc:
 
       private
 
+      def build_purchase_authorize_request(post, money, payment, options)
+        add_payment(post, payment, options)
+        add_purchase_data(post, money, payment, options)
+        add_fraud_detection(post, options)
+      end
+
       def add_reference(authorization)
         return unless authorization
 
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -66,8 +95,8 @@ def add_payment(post, payment, options = {})
           post[:bin] = bin
         else
           post[:card_number] = payment.number
-          post[:card_expiration_month] = payment.month.to_s.rjust(2, '0')
-          post[:card_expiration_year] = payment.year.to_s[-2..-1]
+          post[:card_expiration_month] = format(payment.month, :two_digits)
+          post[:card_expiration_year] = format(payment.year, :two_digits)
```
naashton, line 99: 👍  I didn't know about this helper method

naashton:
All of this looks good, but the remote tests are blowing up over some missing parameters in the request body.

```
       {"error_type"=>"invalid_request_error",
        "validation_errors"=>[{"code"=>"param_required", "param"=>"card_data"}]},
```

jessiagee (changes requested):

ajawadmirza:
> All of this looks good, but the remote tests are blowing up over some missing parameters in the request body.
> 
> ```
>        {"error_type"=>"invalid_request_error",
>         "validation_errors"=>[{"code"=>"param_required", "param"=>"card_data"}]},
> ```



> All of this looks good, but the remote tests are blowing up over some missing parameters in the request body.
> 
> ```
>        {"error_type"=>"invalid_request_error",
>         "validation_errors"=>[{"code"=>"param_required", "param"=>"card_data"}]},
> ```

@naashton I have shared the updated fixtures, let me know if it still doesn't work?

test/remote/gateways/remote_decidir_plus_test.rb
```diff
@@ -3,7 +3,8 @@
 
 class RemoteDecidirPlusTest < Test::Unit::TestCase
   def setup
-    @gateway = DecidirPlusGateway.new(fixtures(:decidir_plus))
+    @gateway_purchase = DecidirPlusGateway.new(fixtures(:decidir_plus_purchase))
```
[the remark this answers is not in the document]
ajawadmirza: updated, please check :)

lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -106,10 +135,17 @@ def add_fraud_detection(post, options)
...
           hsh[:channel] = fraud_detection[:channel] if fraud_detection[:channel]
           hsh[:dispatch_method] = fraud_detection[:dispatch_method] if fraud_detection[:dispatch_method]
-          hsh[:csmdds] = fraud_detection[:csmdds] if fraud_detection[:csmdds]
+          csmdds_arr = []
+          fraud_detection[:csmdds].each do |csmdds|
+            csmdds_hsh = {}
+            csmdds_hsh[:code] = csmdds[:code].to_i
+            csmdds_hsh[:description] = csmdds[:description]
+            csmdds_arr.append(csmdds_hsh)
+          end
```
naashton, line 147: This code will result in an exception if `:csmdds` isn't set in the `options[:fraud_detection]` hash. I think we should break this code out into it's own method and you can call it with a conditional, `if fraud_detection[:csmdds]` or return early from the method.

naashton (changes requested):
On the `csmdds` note, it might be worth adding a test case for when that field is not set, regardless of whether the request passes or fails.

jessiagee (approved):
Adding my 👍🏻 now just waiting for you to address @naashton 's points then should be g2g!

lib/active_merchant/billing/gateways/decidir_plus.rb
ajawadmirza: Thanks, change is updated with unit test case as well.

naashton:
I'm seeing 7 test failures, mostly with this error response:

```
        {"type"=>"processing_error",
         "reason"=>{"id"=>-1, "description"=>"", "additional_description"=>""}}}
```

Are you guys not seeing these?

PR #4285, opened 2022-01-20
Braintree adding ACH support for store transaction

Heavyblade:
### Description

This PR modifies the store method to add support for bank accounts in order to enable ACH transactions on purchases, for that is needed extra work to simulate the tokenization flow that Braintree enables for the front-end libraries to create token-nonces for payment methods.

This nonce is a requirement to store the bank account and set a flag to require verification via network checking.

### Test Execution

Remote:

> Finished in 261.78288 seconds.
> 97 tests, 525 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
> 100% passed

Unit:

> Finished in 14.747631 seconds.
> 5043 tests, 74980 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
> 100% passed

### Rubocop
Inspecting 730 files
730 files inspected, no offenses detected


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -46,6 +47,8 @@ class BraintreeBlueGateway < Gateway
         cannot_refund_if_unsettled: 91506
       }
 
+      DIRECT_BANK_ERROR = 'Direct bank account transactions are not supported. Bank accounts must be stored and verified before use.'.freeze
```
curiousepic, line 50:
```suggestion
      DIRECT_BANK_ERROR = 'Direct bank account transactions are not supported. Bank accounts must be successfully stored before use.'.freeze
```
Slightly more useful message, as verificatino could be taken to mean running a verify action, when the verification is built into the store action.

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -827,6 +827,86 @@ def add_payment_method(parameters, credit_card_or_vault_id, options)
...
         end
       end
+
+      def bank_account_errors(payment_method, options)
+        if payment_method.validate.present?
+          payment_method.validate
+        elsif options[:billing_address].blank?
+          'billing_address is required parameter to store and verify Bank accounts.'
+        elsif options[:ach_mandate].blank?
+          'ach_mandate is a required parameter to process bank acccount transactions see (https://developer.paypal.com/braintree/docs/guides/ach/client-side#show-required-authorization-language)'
```
curiousepic, line 837: Nice, this is great!

test/remote/gateways/remote_braintree_blue_test.rb
```diff
@@ -15,6 +16,21 @@ def setup
       billing_address: address(country_name: 'Canada'),
       description: 'Store Purchase'
     }
+
+    ach_mandate = 'By clicking ["Checkout"], I authorize Braintree, a service of PayPal, ' \
```
curiousepic, line 20: Very small thing but let's normalize this to what it would actually be sent as
```suggestion
    ach_mandate = 'By clicking "Checkout", I authorize Braintree, a service of PayPal, ' \
```

curiousepic:
Quick note - let's add a note about the mandate to the commit message.

lib/active_merchant/billing/gateways/braintree/braintree_common.rb
```diff
@@ -18,6 +18,11 @@ def scrub(transcript)
     transcript.
       gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
       gsub(%r((&?ccnumber=)\d*(&?)), '\1[FILTERED]\2').
-      gsub(%r((&?cvv=)\d*(&?)), '\1[FILTERED]\2')
+      gsub(%r((&?cvv=)\d*(&?)), '\1[FILTERED]\2').
+      gsub(%r((<account-number>)\d+(</account-number>)), '\1[FILTERED]\2').
+      gsub(%r((<payment-method-nonce>)[^<]+(</payment-method-nonce>)), '\1[FILTERED]\2').
+      gsub(%r((<payment-method-token>)[^<]+(</payment-method-token>)), '\1[FILTERED]\2').
+      gsub(%r((<value>)[^<]{100,}(</value>)), '\1[FILTERED]\2').
```
curiousepic, line 25: ` value` is a very generic term. Perhaps this should key off of `<client-token>` instead?

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -148,21 +153,16 @@ def verify(creditcard, options = {})
...
-            end
+      def store(payment_method, options = {})
+        return Response.new(false, bank_account_errors(payment_method, options)) if payment_method.is_a?(Check) && bank_account_errors(payment_method, options).present?
+        return add_customer_with_payment_method(payment_method, options) if options[:customer].blank?
+
+        MultiResponse.new.tap do |r|
+          r.process { check_customer_exists(options[:customer]) }
+          r.process do
+            process_by = r.params['exists'] ? :add_payment_method_to_customer : :add_customer_with_payment_method
+            send process_by, payment_method, options
```
curiousepic, line 164: Hmm - I'm also not sure we want to nest logic within a multiresponse.
I also see that `MultiResponse` blocks are usually initiated with `.run` rather than `.new`. Do you have any thoughts on an alternative?

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -827,6 +827,86 @@ def add_payment_method(parameters, credit_card_or_vault_id, options)
...
+          'billing_address is required parameter to store and verify Bank accounts.'
+        elsif options[:ach_mandate].blank?
+          'ach_mandate is a required parameter to process bank acccount transactions see (https://developer.paypal.com/braintree/docs/guides/ach/client-side#show-required-authorization-language)'
+        end
+      end
+
+      def add_customer_with_payment_method(payment_method, options)
+        if payment_method.is_a? Check
+          create_customer_from_payment_method(payment_method, options)
+          add_bank_account_to_customer(payment_method, options)
```
curiousepic, line 844: I can't quite tell, but it seems like this would create two `Response`s outside of the context of a `MultiResponse`. I think this _might_ mean that the first response wouldn't be captured in the final transcript? Which ideally would be there for debugging purposes.

curiousepic:
Nice work on the Nonce module, I think it's a good solution. A few concerns about some of the adapter logic, but we're close here.

lib/active_merchant/billing/gateways/braintree_blue.rb
Heavyblade: @curiousepic on the .run vs .new.tap there is no difference .run is basically an alias for .new.tap, so I just change it for consistency.

For the usage of Multiresponse, I made a refactor to make the code clear and include the customer creation request in the multiresponse.

lib/active_merchant/billing/gateways/braintree_blue.rb
Heavyblade: Already covered on the refactor 👍

Heavyblade:
@curiousepic refactor and corrections done, ready for a final round.

curiousepic:
@Heavyblade Is this dependent on the gem upgrade?

Heavyblade:
@curiousepic no, it doesn't depend on that.

test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -373,7 +373,7 @@ def test_purchase_with_google_pay
   def test_purchase_with_shipping_options
     options = {
       currency: 'GBP',
-      customer: @customer,
+      customer: 'abc123',
```
curiousepic, line 376: Was this included intentionally? seems unrelated

Heavyblade: @curiousepic there was a failing unit test on Stripe Payment Intents and the reason was that undeclared @customer instance variable, given that was an easy fix I just include that on this PR

curiousepic (approved):
Awesome, I think this refactor is more robust and reasonable, thanks! Will merge soon.

PR #4289, opened 2022-01-21
Decidir Plus: Fraud Detection Fields

naashton:
Add support for fraud detection with Decidir Plus.

This PR includes a method to add `fraud_detection` fields to the post
body of a `purchase` transaction. The `csmdds` subfield of
`fraud_detection` is expected to be given as an array of hashes with a
`code` and `description` value.

CE-2337

Unit: 8 tests, 30 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 10 tests, 33 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir_plus.rb
+++ b/lib/active_merchant/billing/gateways/decidir_plus.rb
@@ -21,6 +21,7 @@ module ActiveMerchant #:nodoc:
 
         add_payment(post, payment, options)
         add_purchase_data(post, money, payment, options)
+        add_fraud_detection(post, options)
 
         commit(:post, 'payments', post)
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -21,6 +21,7 @@ def purchase(money, payment, options = {})
 
         add_payment(post, payment, options)
         add_purchase_data(post, money, payment, options)
+        add_fraud_detection(post, options)
```
jessiagee, line 24: This needs to be done for auth/purchase I believe

jessiagee (changes requested):

lib/active_merchant/billing/gateways/decidir_plus.rb
jessiagee: Oh wait this is Phase 1 so only purchase ignore meeee

naashton: Yea, I am just implementing the method to add `fraud_detection`. This should allow Jawad to simply call the method when he implements the `authorize` transaction method for the gateway.

jessiagee (approved):
All tests pass and looks good to me for phase 1!

lib/active_merchant/billing/gateways/decidir_plus.rb
rachelkirk: love to see this lively discussion and teamwork in action 🤝

PR #4291, opened 2022-01-24
Priority

mikebyrneprth:
Add multiple product lines, tax rates and discount rates.


jessiagee:
Hi @mikebyrneprth ,

Your fork is pretty far behind the upstream master, so you're going to need to either A) create a new fork with latest code and submit a PR to that or B) cherry-pick `4b9afc18c` onto a fresh branch made from a freshly-pulled copy of upstream master. I would suggest A) as it would be substantially faster for you.

Once you've done that, please re-submit just the code from `4b9afc18c` and we'll review it for you.

Thanks!
Jessi

PR #4298, opened 2022-01-27
Litle: fixed refund with custom billing descriptor

softbrada:
Currently, when doing a refund with a  custom billing descriptor the response I get this error

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<litleOnlineResponse xmlns="http://www.litle.com/schema" version="9.14" response="1" message="Error validating xml data against the schema: cvc-complex-type.2.4.a: Invalid content was found starting with element 'customBilling'. One of '{&quot;http://www.litle.com/schema&quot;:transaction, &quot;http://www.litle.com/schema&quot;:recurringTransaction}' is expected." />
```

I checked all other versions of schema, higher versions also, and the problem is that the `customBilling` element is currently a sibling with the `credit` or `echeckCredit`  element , but according to the schema it should be inside the `credit` or `echeckCredit` 

The other methods `add_echeck_purchase_params` and `add_credit_params` already handle the custom billing descriptors correctly

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -78,12 +78,12 @@ module ActiveMerchant #:nodoc:
       def refund(money, payment, options = {})
         request = build_xml_request do |doc|
           add_authentication(doc)
-          add_descriptor(doc, options)
           doc.send(refund_type(payment), transaction_attributes(options)) do
             if payment.is_a?(String)
               transaction_id, = split_authorization(payment)
               doc.litleTxnId(transaction_id)
               doc.amount(money) if money
+              add_descriptor(doc, options)
             elsif check?(payment)
               add_echeck_purchase_params(doc, money, payment, options)
             else


softbrada:
please re-open , this is still valid

softbrada:
@almalee24 @DeeMeyers  can someone help with this?

almalee24:
We will review this

Buitragox (approved):
Looks good!

Buitragox (changes requested):
I was discussing this change with @almalee24 and there are a couple of things to update here. Sorry for the previous approval. 

It would be good to add a couple of tests, at least a unit and remote test.
Maybe something similar to this for a unit test:
```ruby
  def test_passing_custom_billing_on_refund
    options = {
      descriptor_name: 'Name',
      descriptor_phone: 'Phone'
    }
    
    response = stub_comms do
      @gateway.refund(@amount, '100000000000000006;sale;100', options)
      assert_match(%r(<credit.*<customBilling>.*<phone>Phone<.*<descriptor>Name<.*</customBilling>.*</credit>)m, data)
    end.respond_with(successful_refund_response)

    assert_success response
  end
```

And this for a remote test:
```ruby
  def test_successful_refund_with_custom_billing
    options = @options.merge(descriptor_name: 'test name', descriptor_phone: '1234567890')

    assert purchase = @gateway.purchase(100, @credit_card1, options)

    assert refund = @gateway.refund(444, purchase.authorization, options)
    assert_success refund
    assert_equal 'Approved', refund.message
  end
```

Another thing, [the docs of customBilling](http://support.worldpay.com/support/CNP-API/content/custombilling.htm) mention "Although the schema includes the customBilling element as a child of echeckCredit and echeckSale, Worldpay does not support its use in these instances." so we could consider removing the call to `add_descriptor` when doing these types of transactions. 

This call could also be removed for [echeckVerification](http://support.worldpay.com/support/CNP-API/content/echeckveri.htm?Highlight=echeckVerification) and [capture](http://support.worldpay.com/support/CNP-API/content/capture.htm?Highlight=capture) since customBilling does not appear to be a child element of these transactions.

PR #4299, opened 2022-01-27
CyberSource: Add national tax indicator field

ajawadmirza:
Added `national_tax_indicator` in purchase and authorize calls for cyber
source implementation.

CE-2338

Unit:
5050 tests, 75009 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
103 tests, 520 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.2039% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -291,6 +291,7 @@ module ActiveMerchant #:nodoc:
         xml = Builder::XmlMarkup.new indent: 2
         add_customer_id(xml, options)
         add_payment_method_or_subscription(xml, money, creditcard_or_reference, options)
+        add_other_tax(xml, options)
         add_threeds_2_ucaf_data(xml, creditcard_or_reference, options)
         add_decision_manager_fields(xml, options)
         add_mdd_fields(xml, options)
@@ -349,6 +350,7 @@ module ActiveMerchant #:nodoc:
         xml = Builder::XmlMarkup.new indent: 2
         add_customer_id(xml, options)
         add_payment_method_or_subscription(xml, money, payment_method_or_reference, options)
+        add_other_tax(xml, options)
         add_threeds_2_ucaf_data(xml, payment_method_or_reference, options)
         add_decision_manager_fields(xml, options)
         add_mdd_fields(xml, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -898,6 +898,7 @@ def add_payment_method_or_subscription(xml, money, payment_method_or_reference,
           add_purchase_data(xml, money, true, options)
           add_installments(xml, options)
           add_creditcard(xml, payment_method_or_reference)
+          add_other_tax(xml, options)
```
dsmcclain, line 901: Adding the method call here makes it so that these fields are only passed on credit card transactions.

What about when the transaction is a check or a reference (third party token)? I don't see anything in the conversation about this change that indicates the tax fields should be excluded from those transactions.

Unless I am missing something, it seems to me like `add_other_tax` should be called from the `build_auth_request` and `build_purchase_request` methods.

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -248,7 +247,7 @@ def test_successful_asynchronous_adjust
     assert_successful_response(authorize)
     assert adjust = @gateway_latam.adjust(@amount * 2, authorize.authorization, @options)
     assert_success adjust
-    assert capture = @gateway_latam.capture(@amount, authorize.authorization, @options)
+    assert capture = @gateway_latam.capture(@amount, authorize.authorization, @options.merge({ national_tax_indicator: 1 }))
```
dsmcclain, line 250: Is there a reason why this value is removed from the `@options` hash only to be merged back into the hash in all of these tests?

dsmcclain (changes requested):

test/remote/gateways/remote_cyber_source_test.rb
ajawadmirza: To keep the test cases same, this value is being passed in the options as previously we were sending `national_tax_indicator ` in **@options** and the otherTax values were being set for capture only. In this implementation, this indicator key is being sent in for auth and purchase calls as well so this key was removed from global @options and are being passed for capture by merge.

lib/active_merchant/billing/gateways/cyber_source.rb
ajawadmirza: Right, this change is updated so that the `otherTax` is passed in for both request separately.

dsmcclain (approved):
@ajawadmirza thank you for changing where the method is called.

I get the same 7 test failures as you on the remote tests. They also fail on master.

Approved. I will merge this in a moment.

PR #4300, opened 2022-01-27
Stripe PI: confirm_intent should have add_payment_method_types

aenand:
Stripe supports passing payment_method_types as an optional parameter for the `confirm_intent` endpoint.
Adding this support allows for users to change the payment method on a transaction
and attempt to complete it in one request.

ECS-2193

Test Summary
Local:
5048 tests, 75004 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
37 tests, 191 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
72 tests, 329 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_stripe_payment_intents_test.rb
+++ b/test/remote/gateways/remote_stripe_payment_intents_test.rb
@@ -869,6 +869,22 @@ class RemoteStripeIntentsTest < Test::Unit::TestCase
     assert_equal 'requires_confirmation', update_response.params['status']
   end
 
+  def test_create_a_payment_intent_and_confirm_with_different_payment_method
+    options = {
+      currency: 'USD',
+      payment_method_types: %w[afterpay_clearpay],
+      metadata: { key_1: 'value_1', key_2: 'value_2' }
+    }
+    assert create_response = @gateway.setup_purchase(@amount, options)
+    assert_equal 'requires_payment_method', create_response.params['status']
+    intent_id = create_response.params['id']
+    assert_equal 2000, create_response.params['amount']
+    assert_equal 'afterpay_clearpay', create_response.params['payment_method_types'][0]
+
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -743,8 +743,9 @@ def test_create_a_payment_intent_and_confirm
     assert get_response = @gateway.show_intent(intent_id, options)
     assert_equal 'requires_confirmation', get_response.params['status']
 
-    assert confirm_response = @gateway.confirm_intent(intent_id, nil, return_url: 'https://example.com/return-to-me')
+    assert confirm_response = @gateway.confirm_intent(intent_id, nil, options.merge(return_url: 'https://example.com/return-to-me', payment_method_types: 'card'))
```
tatsianaclifton, line 746: Since `payment_method_type` is optional what do you think about having a separate test instead of changing existing one?

aenand: Absolutely! I'll add an extra remote test and have it check that the `payment_method` on the intent changes.

aenand:
@tatsianaclifton I added a separate test to show how to change the payment_method on an existing intent with this new change.

test/remote/gateways/remote_stripe_payment_intents_test.rb
tatsianaclifton: New test is great! Maybe remove `payment_method_types` from this test that way we will know that it still works if `payment_method_types` are not provided.

aenand: Done @tatsianaclifton !

tatsianaclifton (approved):
Thanks, looks great!

PR #4303, opened 2022-01-31
Paysafe: Fix commit method for unstore operation

ajawadmirza:
Fixed commit method for `unstore` operation in paysafe implementation to
parse the response correctly for empty strings.

CE-2285

Unit:
5052 tests, 75017 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
31 tests, 94 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


dsmcclain (approved):
Nice work @ajawadmirza.

I have verified with Paysafe's documentation[ here](https://developer.paysafe.com/en/vault/api/#/reference/0/profiles/delete-a-profile) that a successful `profile:delete` request will return a 200 response with no body, so I think your change is safe.

I will add a changelog entry and merge this PR.

Edit: all tests are passing 👍

PR #4305, opened 2022-01-31
CE-2343 Bluesnap Idempotency

drkjc:
Bluesnap: Add support for optional Idempotency Key field on purchase, authorize, capture, refund, and void. Idempotency Key is passed as a header.

Unit tests:
41 tests, 237 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests:
49 tests, 166 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -78,7 +78,7 @@ module ActiveMerchant
       def purchase(money, payment_method, options = {})
         payment_method_details = PaymentMethodDetails.new(payment_method)
 
-        commit(:purchase, :post, payment_method_details) do |doc|
+        commit(:purchase, options, :post, payment_method_details) do |doc|
           if payment_method_details.alt_transaction?
             add_alt_transaction_purchase(doc, money, payment_method_details, options)
           else
@@ -88,13 +88,13 @@ module ActiveMerchant
       end
 
       def authorize(money, payment_method, options = {})
-        commit(:authorize) do |doc|
+        commit(:authorize, options) do |doc|
           add_auth_purchase(doc, money, payment_method, options)
         end
       end
 
       def capture(money, authorization, options = {})
-        commit(:capture, :put) do |doc|
+        commit(:capture, options, :put) do |doc|
           add_authorization(doc, authorization)
           add_order(doc, options)
           add_amount(doc, money, options) if options[:include_capture_amount] == true
@@ -102,7 +102,7 @@ module ActiveMerchant
       end
 
       def refund(money, authorization, options = {})
-        commit(:refund, :put) do |doc|
+        commit(:refund, options, :put) do |doc|
           add_authorization(doc, authorization)
           add_amount(doc, money, options)
           add_order(doc, options)
@@ -110,7 +110,7 @@ module ActiveMerchant
       end
 
       def void(authorization, options = {})
-        commit(:void, :put) do |doc|
+        commit(:void, options, :put) do |doc|
           add_authorization(doc, authorization)
           add_order(doc, options)
         end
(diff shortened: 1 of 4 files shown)


test/unit/gateways/blue_snap_test.rb
```diff
@@ -543,6 +543,17 @@ def test_localizes_currencies
     end
   end
 
+  def test_optional_idempotency_key_header
+    response = stub_comms(@gateway, :raw_ssl_request) do
+      @gateway.purchase(@amount, @credit_card, @options.merge({ idempotency_key: 'test123' }))
+    end.check_request do |headers|
+      headers && headers['Idempotency-Key'] == 'test123'
```
dsmcclain, line 550: We could replace `'test123'` with any value and this test would still pass.

This is because we aren't asserting anything here. It needs to be `assert_equal 'test123', headers['Idempotency-Key']`.

Once you fix that you will see that the test fails because `check_request` hasn't been implemented correctly here. In this instance the headers end up being the fourth value returned by `check_request`, so the syntax should look like:
```
end.check_request do |_method, _url, _data, headers|
```
kind of like you see [here](https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/blue_snap_test.rb#L312).

In my experience, it is not easy to see how `check_request` works under the hood. It yields values depending on how the gateway adapter formulates the request, so whenever I need to use it to extract a value from a request I usually just declare all four values in the block and add a binding inside the block as well to look at what it has extracted.

test/unit/gateways/blue_snap_test.rb
```diff
@@ -543,6 +543,17 @@ def test_localizes_currencies
...
+    end.respond_with(successful_authorize_response)
+
+    assert_success response
+    assert_equal '1012082893', response.authorization
```
dsmcclain, line 554: Nit: These two lines are superfluous, so you might as well remove them.

On line 551 you have stubbed the response, and here you are making assertions against the very response you provided.

I know this pattern has been copied and pasted throughout active merchant but I think the test is much nicer if you remove these lines because then it better communicates what you are actually trying to test (which is that the request itself contains the header).

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -433,15 +433,15 @@ def parse_element(parsed, node)
         end
       end
 
-      def api_request(action, request, verb, payment_method_details)
-        ssl_request(verb, url(action, payment_method_details), request, headers)
+      def api_request(action, request, verb, payment_method_details, options)
```
dsmcclain, line 436: It's kind of weird to me that we have this `api_request` method wrapping the `ssl_request` method. We could just calculate the `url` and the `header` inside the `commit` method and then call `ssl_request` directly [the way we do in most other gateway adapters](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/adyen.rb#L557).

This is just an idle thought on my part, not a request for changes. It has nothing to do with the changes you are making in this PR.

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -1,4 +1,5 @@
 require 'nokogiri'
+require 'pry'
```
dsmcclain, line 2: If I had a nickel for every time I've done this.

Speaking from experience, you might find it easier to do this inline when you add breakpoints, e.g. (`require 'pry'; binding.pry`), that way rubocop will prevent you from trying to check this `require` line into the code.

dsmcclain (changes requested):
Good work so far @drkjc! Just a few changes requested.

test/unit/gateways/blue_snap_test.rb
drkjc: That makes sense. I knew something was a little off here. And thanks for the tip.

test/unit/gateways/blue_snap_test.rb
```diff
@@ -543,6 +543,14 @@ def test_localizes_currencies
     end
   end
 
+  def test_optional_idempotency_key_header
```
dsmcclain, line 546: Nice, lookin good! 👍

dsmcclain (approved):
All tests passing. Good work @drkjc!

Make sure you add a changelog entry before you merge!

PR #4307, opened 2022-02-01
BlueSnap: Update refund

ajawadmirza:
Updated `refund` method to include transaction id in its url for blue
snap implementation

CE-2351

Unit:
5052 tests, 75025 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
48 tests, 163 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -102,10 +102,11 @@ module ActiveMerchant
       end
 
       def refund(money, authorization, options = {})
-        commit(:refund, options, :put) do |doc|
-          add_authorization(doc, authorization)
-          add_amount(doc, money, options)
-          add_order(doc, options)
+        options[:endpoint] = options[:merchant_transaction_id] ? "/refund/merchant/#{options[:merchant_transaction_id]}" : "/refund/#{authorization}"
+        commit(:refund, options, :post) do |doc|
+          add_amount(doc, money, options) if money
+          %i[reason cancel_subscription tax_amount].each { |field| send_when_present(doc, field, options) }
+          add_metadata(doc, options)
         end
       end
 
@@ -234,6 +235,7 @@ module ActiveMerchant
               doc.send('meta-key', truncate(entry[:meta_key], 40))
               doc.send('meta-value', truncate(entry[:meta_value], 500))
               doc.send('meta-description', truncate(entry[:meta_description], 40))
+              doc.send('is-visible', truncate(entry[:meta_is_visible], 5))
             end
           end
         end
@@ -386,7 +388,7 @@ module ActiveMerchant
 
       def parse(response)
         return bad_authentication_response if response.code.to_i == 401
-        return generic_error_response(response.body) if [403, 429].include?(response.code.to_i)
+        return generic_error_response(response.body) if [403, 405, 429].include?(response.code.to_i)
 
         parsed = {}
         doc = Nokogiri::XML(response.body)
@@ -433,7 +435,7 @@ module ActiveMerchant
       end
 
       def api_request(action, request, verb, payment_method_details, options)
-        ssl_request(verb, url(action, payment_method_details), request, headers(options))
+        ssl_request(verb, url(action, options, payment_method_details), request, headers(options))
       rescue ResponseError => e
         e.response
       end
@@ -456,9 +458,10 @@ module ActiveMerchant
         )
       end
 
-      def url(action = nil, payment_method_details = PaymentMethodDetails.new())
+      def url(action = nil, options = {}, payment_method_details = PaymentMethodDetails.new())
         base = test? ? test_url : live_url
         resource = action == :store ? 'vaulted-shoppers' : payment_method_details.resource_url
+        resource += options[:endpoint] if action == :refund
         "#{base}/#{resource}"
       end
 
@@ -531,7 +534,10 @@ module ActiveMerchant
       end
 
       def root_element(action, payment_method_details)
-        action == :store ? 'vaulted-shopper' : payment_method_details.root_element
+        return 'refund' if action == :refund
+        return 'vaulted-shopper' if action == :store
+
+        payment_method_details.root_element
       end
 
       def headers(options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -102,10 +102,13 @@ def capture(money, authorization, options = {})
...
-          add_authorization(doc, authorization)
-          add_amount(doc, money, options)
-          add_order(doc, options)
+        options[:endpoint] = options[:merchant_transaction_id] ? "/refund/merchant/#{options[:merchant_transaction_id]}" : "/refund/#{authorization}"
+        commit(:refund, options, :post) do |doc|
+          send_when_present(doc, :reason, options)
+          send_when_present(doc, :cancel_subscription, options)
+          add_amount(doc, money, options) if money
+          send_when_present(doc, :tax_amount, options)
+          add_transaction_meta_data(doc, options)
```
dsmcclain, line 111: Could be nice to simplify this into:
```
commit(:refund, options, :post) do |doc|
  add_amount(doc, money, options) if money
  %i[reason cancel_subscription tax_amount].each { |field| send_when_present(doc, field, options) }
  add_transaction_meta_data(doc, options)
end
```

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -239,6 +242,21 @@ def add_metadata(doc, options)
         end
       end
 
+      def add_transaction_meta_data(doc, options = {})
+        transaction_meta_data = options[:transaction_meta_data] || []
+        return if transaction_meta_data.empty?
```
dsmcclain, line 247: Could be nice to simplify this into
`return unless (transaction_meta_data = options[:transaction_meta_data])`

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -239,6 +242,21 @@ def add_metadata(doc, options)
...
+
+        doc.send('transaction-meta-data') do
+          options[:transaction_meta_data].each do |meta|
+            doc.send('meta-data') do
+              doc.send('meta-key', meta[:key])
+              doc.send('meta-value', meta[:value])
+              doc.send('meta-description', meta[:description])
```
dsmcclain, line 254: Based on the documentation I'm seeing [here](https://developers.bluesnap.com/v8976-XML/docs/meta-data), there is an `is-visible` boolean that can also be added to the `meta-data` object.

I'm wondering why you chose to include `transaction-meta-data` in the refund request at all? The work ticket does not mention it, and it appears from the gateway documentation that it is entirely optional. I also appears that it can be added to any transaction type, not just refund, so I'm wondering if we do decide to add it, shouldn't we add it to every transaction?

lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -531,6 +550,8 @@ def root_attributes
       end
 
       def root_element(action, payment_method_details)
+        return 'refund' if action == :refund
```
dsmcclain, line 553: I like how you've done this 👍 

I think the method could be even better if we take your idea further:
```
def root_element(action, payment_method_details)
  return 'refund' if action == :refund
  return 'vaulted-shopper' if action == :store

  payment_method_details.root_element
end
```

test/unit/gateways/blue_snap_test.rb
```diff
@@ -376,13 +376,58 @@ def test_failed_capture
...
+      assert_match(/<meta-key>refundedItems<\/meta-key>/, data)
+      assert_match(/<meta-value>KTD<\/meta-value>/, data)
+      assert_match(/<meta-description>Metadata 2<\/meta-description>/, data)
+      transaction_meta_data = doc.root.elements['transaction-meta-data'].elements.to_a
+      transaction_meta_data.each_with_index do |item, index|
+        assert_match item.elements['meta-key'].text, options[:transaction_meta_data][index][:key]
+        assert_match item.elements['meta-value'].text, options[:transaction_meta_data][index][:value]
+        assert_match item.elements['meta-description'].text, options[:transaction_meta_data][index][:description]
+      end
+    end.respond_with(successful_refund_response)
```
dsmcclain, line 416: Thank you for taking to the time to demonstrate that all of these fields end up in the request!

test/unit/gateways/blue_snap_test.rb
```diff
@@ -376,13 +376,58 @@ def test_failed_capture
...
     assert_success response
     assert_equal '1012082907', response.authorization
   end
 
+  def test_successful_refund_with_merchant_id
+    merchant_transaction_id = '12678'
+    response = stub_comms(@gateway, :ssl_request) do
+      @gateway.refund(@amount, '', @options.merge({ merchant_transaction_id: merchant_transaction_id }))
+    end.check_request do |_action, endpoint, _data, _headers|
+      assert_includes endpoint, "/refund/merchant/#{merchant_transaction_id}"
```
dsmcclain, line 426: 👍

dsmcclain (changes requested):
Great work @ajawadmirza! This looks like a big improvement.

I added a few suggestions for simplifying some of the code.

I also have a question about whether the `transaction-meta-data` is necessary, and if so whether we should support it for all transactions.

lib/active_merchant/billing/gateways/blue_snap.rb
ajawadmirza: Thanks, that looks much cleaner. Never thought of this 😄

lib/active_merchant/billing/gateways/blue_snap.rb
ajawadmirza: Looks better, changes are updated.

lib/active_merchant/billing/gateways/blue_snap.rb
ajawadmirza: meta data fields are already being sent in every transaction using method [add_meta_data](https://github.com/activemerchant/active_merchant/blob/ff776e257d3c605ec76c7f9f2620b7cc9a8e5098/lib/active_merchant/billing/gateways/blue_snap.rb#L216) in the code. So this function is reused for refund as well (previously added method is removed)

lib/active_merchant/billing/gateways/blue_snap.rb
ajawadmirza: `add_transaction_meta_data ` is removed and `add_metadata` is reused.

lib/active_merchant/billing/gateways/blue_snap.rb
dsmcclain: Nice, I didn't notice `add_meta_data` was already available. Good change 👍

dsmcclain (approved):
Looks great, thanks for the small changes.

All tests are passing. I will merge.

PR #4309, opened 2022-02-04
Orbital: refactor gateway adapter and indicate support for network tokenization

dsmcclain:
This PR contains six commits. The first five commits are a step-by-step refactoring and reorganizing of the Orbital gateway adapter. This was done in order to make it possible to audit the gateway adapter's support of network tokenization.

The sixth commit adds the `supports_network_tokenization?` method to the gateway adapter.

The results of the audit are that the Orbital gateway adapter currently supports network tokenization as evidenced by the inclusion of the following fields in a transaction with a network token:
`AuthenticationECIInd`
`DPANInd` (set to the value 'Y')
`DigitalTokenCryptogram`

As per the Orbital docs, inclusion of `DigitalTokenCryptogram` obviates the requirement for brand-specific crytogram references in the request (such as `cavv`, `aav`, or `aevv`).

CE-2332

Rubocop:

Unit:
5058 tests, 75061 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
Loaded suite test/remote/gateways/remote_orbital_test
80 tests, 355 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -203,40 +203,25 @@ module ActiveMerchant #:nodoc:
         # ECP for Orbital requires $0 prenotes so ensure
         # if we are doing a force capture with a check, that
         # we do a purchase here
-        if options[:force_capture] && payment_source.is_a?(Check) &&
-           (options[:action_code].include?('W8') || options[:action_code].include?('W9') || options[:action_code].include?('ND'))
-          return purchase(money, payment_source, options)
-        end
+        return purchase(money, payment_source, options) if force_capture_with_echeck?(payment_source, options)
 
-        order = build_new_order_xml(AUTH_ONLY, money, payment_source, options) do |xml|
-          add_payment_source(xml, payment_source, options)
-          add_address(xml, payment_source, options)
-          if @options[:customer_profiles]
-            add_customer_data(xml, payment_source, options)
-            add_managed_billing(xml, options)
-          end
-        end
+        order = build_new_auth_purchase_order(AUTH_ONLY, money, payment_source, options)
+
         commit(order, :authorize, options[:retry_logic], options[:trace_number])
       end
 
-      def verify(creditcard, options = {})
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_orbital_test.rb
```diff
@@ -349,113 +349,6 @@ def test_successful_purchase_with_echeck_on_next_day
...
-        address1: '4 Northeastern Blvd',
-        address2: '',
-        city: 'Salem',
-        state: 'NH',
-        zip: '03105',
-        country: 'US'
-      }
-    }
-  ].each do |fixture|
-    define_method("test_successful_#{fixture[:card][:brand]}_authorization_with_3ds") do
```
dsmcclain, line 417: This is a clever technique, but it is not appropriate to use it in a test file like this, because dynamically creating test cases like this makes it impossible to run these tests individually.

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -840,6 +840,17 @@ def ip_authentication?
...
+          add_payment_source(xml, payment_source, options)
+          add_address(xml, payment_source, options)
+          if @options[:customer_profiles]
+            add_customer_data(xml, payment_source, options)
+            add_managed_billing(xml, options)
+          end
+        end
+      end
+
       def build_new_order_xml(action, money, payment_source, parameters = {})
```
dsmcclain, line 939: It would be possible to further simplify this critical method, except that the Orbital API strictly enforces the _order_ of fields in the request. Therefore the various `add_something` method calls must remain in their seemingly random sequence.

naashton: Usually XML follows a strict format that should adhere to some pre-determined XSD which most API's validate against. That said, I don't think this practice is applied to all API's that use XML.

naashton (approved):
Ok, I think I went over the important parts here and the refactoring. All looks kosher, no glaring issues that stand out. 🚢

PR #4313, opened 2022-02-08
CyberSource: Add option to target 0 verify

gasb150:
Summary:
In order to be able to use $0 authorizations this PR add
the gateway specific field "zero_dollar_auth" as an option, so
I can verify my customers Credit Cards, depending on the acquiring bank,
without making any charges to them.

Local Tests:
115 tests, 562 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote Tests:
105 tests, 526 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.3333% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -158,9 +158,10 @@ module ActiveMerchant #:nodoc:
       end
 
       def verify(payment, options = {})
+        amount = eligible_for_zero_auth?(payment, options) ? 0 : 100
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, payment, options) }
-          r.process(:ignore_result) { void(r.authorization, options) }
+          r.process { authorize(amount, payment, options) }
+          r.process(:ignore_result) { void(r.authorization, options) } unless amount == 0
         end
       end
 
@@ -1078,6 +1079,10 @@ module ActiveMerchant #:nodoc:
         else
           response[:message]
         end
+      end
+
+      def eligible_for_zero_auth?(payment_method, options = {})
+        payment_method.is_a?(CreditCard) && options[:zero_amount_auth]
       end
     end
   end
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -10,6 +10,11 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
 
     @credit_card = credit_card('4111111111111111', verification_value: '987')
     @declined_card = credit_card('801111111111111')
+    @master_credit_card = credit_card('5555555555554444',
+      verification_value: '321',
+      month: '12',
+      year: (Time.now.year + 2).to_s,
+      brand: :master)
     @pinless_debit_card = credit_card('4002269999999999')
     @elo_credit_card = credit_card('5067310000000010',
       verification_value: '321',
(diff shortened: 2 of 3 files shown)


test/unit/gateways/cyber_source_test.rb
```diff
@@ -737,6 +737,25 @@ def test_successful_verify
     assert_success response
   end
 
+  def test_successful_verify_0_amount
```
Heavyblade, line 740: 💬  Comment:

Couple of things here:

1) You are mocking the :ssl_request instead of the :ssl_post, that's why you have the request body on the headers variable.
2) you don't need to assert the response when you are checking something in the request.

You want something like this:

```Ruby
  def test_successful_verify_0_amount
    @options[:zero_dollar_auth] = true
    stub_comms(@gateway, :ssl_post) do
      @gateway.verify(@credit_card, @options)
    end.check_request(skip_response: true) do |_endpoint, data|
      assert_match %r(<grandTotalAmount>0.00<\/grandTotalAmount>), data
    end
  end

```

Heavyblade (commented):
@gasb150 laved some comments for your consideration

javierpedrozaing (commented):
Great work @gasb150 , just I did a couple of comments.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -158,8 +158,9 @@ def adjust(money, authorization, options = {})
       end
 
       def verify(payment, options = {})
+        amount = (eligible_for_0_auth?(payment, options) ? 0 : 100)
```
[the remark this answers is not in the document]
gasb150: Ok, removing it.

test/unit/gateways/cyber_source_test.rb
gasb150: Oh, I see, I'll use :ssl_post instead of :ssl_request.

test/unit/gateways/cyber_source_test.rb
```diff
@@ -737,6 +737,25 @@ def test_successful_verify
...
+    assert_success response
+  end
+
+  def test_successful_verify_1_amount
+    response = stub_comms(@gateway, :ssl_request) do
+      @gateway.verify(@credit_card, @options)
+    end.check_request do |_endpoint, _data, headers|
+      assert_match %r(<grandTotalAmount>1.00<\/grandTotalAmount>), headers
+    end.respond_with(successful_authorization_response)
+    assert_success response
```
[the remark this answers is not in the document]
gasb150: For this test, my priority is to check the request for a specific field.

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -1009,6 +1014,29 @@ def test_verify_credentials
...
+
+  def test_successful_verify_0_amount_master
+    @options[:zero_dollar_auth] = true
+    response = @gateway.verify(@master_credit_card, @options)
+    assert_success response
+    assert_match '0.00', response.params['amount']
+    assert_equal 'Successful transaction', response.message
+  end
+
+  def test_successful_verify_1_amount
```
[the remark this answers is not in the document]
gasb150: OK, Fixed.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -1079,6 +1080,10 @@ def message_from(response)
           response[:message]
         end
       end
+
+      def eligible_for_0_auth?(payment_method, options = {})
```
curiousepic, line 1084: I do see another gateway uses this method name, but I think we want to keep digits out of method names 
```suggestion
      def eligible_for_zero_auth?(payment_method, options = {})
```

curiousepic:
Just a couple naming suggestions, but this looks good!

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -1079,6 +1080,10 @@ def message_from(response)
           response[:message]
         end
       end
+
+      def eligible_for_zero_auth?(payment_method, options = {})
+        payment_method.is_a?(CreditCard) && options[:zero_amount_auth]
```
curiousepic, line 1085: Apologies for the switcheroo, but I just remembered an issue we've had recently with a similar field name being confusing for users, since this flag is specifically used for `verify` rather than `authorize`, and the suggestion was to keep the name as specific as possible.
```suggestion
        payment_method.is_a?(CreditCard) && options[:zero_amount_verify]
```

curiousepic: But now worries - I can make this change as part of the merge.

PR #4324, opened 2022-02-14
Global Collect Success Criteria V2

peteroas:


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -439,8 +438,33 @@ def content_type
...
+            response.dig('cardPaymentMethodSpecificOutput', 'voidResponseId')
+
+          mobile_payment_void_response_id =
+            response.dig('mobilePaymentMethodSpecificOutput', 'voidResponseId')
+
+          if %w(00 0 8 11).include?(card_payment_void_response_id) || %w(00 0 8 11).include?(mobile_payment_void_response_id)
+            true
+          else
+            response.dig('payment', 'status') == 'CANCELLED'
+          end
```
curiousepic, line 461: A few suggestions for simplification here - maybe a little less readable, but more reasonable without the if/else. The last ine is basically the conditional now.

```suggestion
        when :void
          void_response_id = response.dig('cardPaymentMethodSpecificOutput', 'voidResponseId') || response.dig('mobilePaymentMethodSpecificOutput', 'voidResponseId')
          %w(00 0 8 11).include?(void_response_id) || response.dig('payment', 'status') == 'CANCELLED'
```

peteroas:
> What did you change from the previous branch?

@aenand yeah sorry, that's hard to detect. The change is subtle, it's in the `void` action in `success_from/1` method. There was an if conditional that was not actually doing what we meant for it to do.

peteroas:
Because of the revert of the previous PR, it's hard to show the difference between the old PR and the new one.

peteroas:
> I'm getting some errors when running tests. Can you rerun tests and amend your commit message to include the results ([see this template](https://github.com/activemerchant/active_merchant/wiki/Submitting-Your-Gateway#pr-submission-template)) if they all pass?

@BritneyS will do!

BritneyS (approved):
LGTM!

PR #4325, opened 2022-02-15
Decidir Plus: Improve response messaging

naashton:
Improve the quality of the `message` we return in the response from the
gateway. These changes will account for a `status` of `rejected` and
check for an error message returned in the `status_details` from
Decidir Plus and will also account for `fraud_detection` responses

CE-2387

Unit: 13 tests, 53 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 16 tests, 56 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


meagabeth (approved):
This looks like a solid solution - well done! Tests pass 🚀

PR #4331, opened 2022-02-22
LITLE:  Translate or unify :google_pay and android_pay

javierpedrozaing:
Description
-------------------------
we need unify the source  as :andorid_pay if the network token
 is google_pay or android_pay

JIRA ticket number
-------------------------
GWI-96

Unit test
-------------------------
Finished in 0.137911 seconds.

54 tests, 230 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

391.56 tests/s, 1667.74 assertions/s

Remote test
-------------------------
Finished in 35.502618 seconds.

50 tests, 209 assertions, 13 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
74% passed

1.41 tests/s, 5.89 assertions/s

Rubocop
-------------------------
729 files inspected, no offenses detected

Some remote test was fixed in this PR => [https://github.com/activemerchant/active_merchant/pull/4340](https://github.com/activemerchant/active_merchant/pull/4340)

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -386,7 +386,7 @@ module ActiveMerchant #:nodoc:
           doc.orderSource(order_source)
         elsif payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source == :apple_pay
           doc.orderSource('applepay')
-        elsif payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source == :android_pay
+        elsif payment_method.is_a?(NetworkTokenizationCreditCard) && [:google_pay, :android_pay].include?(payment_method.source)
           doc.orderSource('androidpay')
         elsif payment_method.respond_to?(:track_data) && payment_method.track_data.present?
           doc.orderSource('retail')
@@ -394,6 +394,7 @@ module ActiveMerchant #:nodoc:
           doc.orderSource('ecommerce')
         end
       end
+
 
       def order_source(options = {})
         return options[:order_source] unless options[:stored_credential]
--- a/test/remote/gateways/remote_litle_test.rb
+++ b/test/remote/gateways/remote_litle_test.rb
@@ -65,6 +65,17 @@ class RemoteLitleTest < Test::Unit::TestCase
         payment_cryptogram: 'BwABBJQ1AgAAAAAgJDUCAAAAAAA='
       }
     )
+
+    @decrypted_google_pay = ActiveMerchant::Billing::NetworkTokenizationCreditCard.new(
+      {
+        source: :google_pay,
+        month: '01',
+        year: '2021',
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -395,6 +395,13 @@ def add_order_source(doc, payment_method, options)
         end
       end
 
+      def google_token_type(payment_method)
```
Heavyblade, line 398: 💬  Comment:

mmm maybe could be just a matter of taste but why have a method for something that could be easily been addressed by a `[:google_pay, :android_pay].incluide?(payment_method.source)` ?

test/unit/gateways/litle_test.rb
```diff
@@ -258,6 +268,14 @@ def test_add_android_pay_order_source
     end.respond_with(successful_purchase_response)
   end
 
+  def test_add_google_pay_order_source
+    stub_comms do
+      @gateway.purchase(@amount, @decrypted_google_pay)
+    end.check_request do |_endpoint, data, _headers|
```
Heavyblade, line 274: 💬  Comment:

Maybe is good place to have a `check_request(skip_response: true) `and avoid the uneeded response check

test/remote/gateways/remote_litle_test.rb
```diff
@@ -218,6 +229,12 @@ def test_successful_purchase_with_android_pay
     assert_equal 'Approved', response.message
   end
 
+  def test_successful_purchase_with_google_pay
```
Heavyblade, line 232: 💬  Comment:

I not inherently bad, but I think that may this test is unnecessary, the change that this PR introduces doesn't change the current behavior against the API, it just deals with a particular issue with CORE.

Heavyblade (commented):
@javierpedrozaing great job leave some comments for your consideration.

lib/active_merchant/billing/gateways/litle.rb
curiousepic: I agree with this, let's tighten it up into an `include?`

test/remote/gateways/remote_litle_test.rb
curiousepic: I think this is ok, there is actually functionality added (accepting `google_pay` as input), and should be tested.

lib/active_merchant/billing/gateways/litle.rb
javierpedrozaing: ok, Thanks, I will do the change. :rocket:

javierpedrozaing:
Thanks @curiousepic and @Heavyblade I did some changes according your comments, Thx.

curiousepic (approved):
👍 Looks good - I'm going to do a little cleanup of the commit message, but will merge tomorrow.

PR #4340, opened 2022-02-28
Litle: Update schema version to 12.3 and test correction 

jherreraa:
Summary:
----------
This PR updates the used schema to the latest version available according to the docs on https://vantiv.github.io/latest-versions/

Test Execution:
-----------
Finished in 38.911867 seconds.
50 tests, 227 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92% passed
Several failing tests corrected according to docs in the following pages:
https://vantiv.github.io/sandbox/#howItWorks
https://developer.worldpay.com/assets/pdf/us-ecom-pdf/Worldpay_eComm_cnpAPI_Reference_Guide_APIV12.12_V2.16.pdf

RuboCop:
---------


gasb150:
Hi @jherreraa, nice job, but I'm having problems with some unit tests, could you check it?

test/remote/gateways/remote_litle_test.rb
```diff
@@ -498,21 +503,20 @@ def test_void_with_echeck
...
-
-    assert void = @gateway.void(auth.authorization)
+    assert void = @gateway.void(auth.authorization, @options)
     assert_success void
     assert_equal 'Approved', void.message
   end
 
   def test_unsuccessful_void
-    assert void = @gateway.void('123456789012345360;authorization;100')
+    assert void = @gateway.void('123456789012345360;authorization;1000', @options)
```
javierpedrozaing, line 519: :speech_balloon:
Maybe you can change this for:
```suggestion
    authorization = '123456789012345360;authorization;1000'
    assert void = @gateway.void(authorization, @options)
```

or

create first an authorize and get the authorization
```suggestion
    assert authorize = @gateway.authorize(10010, @credit_card1, @options)@options)
    assert void = @gateway.void(authorize.authorization, @options)
```

javierpedrozaing (commented):
Good job, Just a couple of comments.

test/remote/gateways/remote_litle_test.rb
```diff
@@ -132,9 +132,15 @@ def test_successful_authorization_with_echeck
...
   end
 
-  def test_avs_and_cvv_result
+  def test_avs_result
+    @credit_card1.number = 4100401234567000
     assert response = @gateway.authorize(10010, @credit_card1, @options)
     assert_equal 'X', response.avs_result['code']
+  end
+
+  def test_cvv_result
```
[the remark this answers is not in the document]
Heavyblade: 💬  Comment:

In the same line of being clear on the test case, what is the meaning of that "M" ?, what are we checking ?

test/remote/gateways/remote_litle_test.rb
```diff
@@ -207,13 +213,13 @@ def test_successful_purchase_with_3ds_fields
...
   def test_successful_purchase_with_apple_pay
-    assert response = @gateway.purchase(10010, @decrypted_apple_pay)
+    assert response = @gateway.purchase(1000, @decrypted_apple_pay, @options)
     assert_success response
     assert_equal 'Approved', response.message
   end
 
   def test_successful_purchase_with_android_pay
-    assert response = @gateway.purchase(10000, @decrypted_android_pay)
+    assert response = @gateway.purchase(10000, @decrypted_android_pay, @options)
```
Heavyblade, line 222: ❓  Question:
Why the value on the test above is 1000 and this one is 10000, is a "magic" value ?

Heavyblade (commented):
@jherreraa pretty much OK, leaved a couple of questions

test/remote/gateways/remote_litle_test.rb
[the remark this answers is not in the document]
jherreraa: > ❓ just personal curiosity question, Why separate this function, if we can use the same response?

It have been separated because of the sandbox behavior, it doesn't response with both replies simultaneously
Screen Shot 2022-03-07 at 4 05 04 PM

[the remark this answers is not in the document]
jherreraa: > 💬 Comment:
> 
> In the same line of being clear on the test case, what is the meaning of that "M" ?, what are we checking ?

That code is response for a "Match" in the card validation according to docs in:
http://support.worldpay.com/support/CNP-API/content/cardvalrespcodes.htm?Highlight=card%20validation

test/remote/gateways/remote_litle_test.rb
jherreraa: Just an update in the value because of the multiple trying in the process of correcting tests, in this case is not a magic value

test/remote/gateways/remote_litle_test.rb
jherreraa: sandbox behavior doesn't allow to use a previous response as a parameter to later request

jherreraa:
> 💬 An small comment @jherreraa: I'm having problems with some unit tests, could you check it?

corrected

test/remote/gateways/remote_litle_test.rb
javierpedrozaing: ohhh, just I realized that doing that would be a successful void. :grin:, then  the test is failing because it should be `assert_success` instead `assert_failure` and besides there is another test called `test_void_unsuccessful ` in the line 597

javierpedrozaing (changes requested):
Hi @jherreraa looks good but I leave some comments to check.

javierpedrozaing (commented):
LGTM

gasb150:
💪🏽 ✅  LGTM

javierpedrozaing (commented):
LGTM 🚀

curiousepic:
I see that this clears up a bunch of failing remote tests from master. However it looks like running the remote tests locally, I get one failure that doesn't occur on master: `test_successful_purchase_with_google_pay` - I also have 5 failures on this branch, which is 1 more than the output you pasted in the commit message, so I assume it's this one. 

Here's the content of the error message for that failure: `"Error validating xml data against the schema: cvc-complex-type.4: Attribute 'id' must appear on element 'sale'.",`

jherreraa:
> @jherreraa See above - also the commit title is missing?

Corrected

jherreraa:
> I see that this clears up a bunch of failing remote tests from master. However it looks like running the remote tests locally, I get one failure that doesn't occur on master: `test_successful_purchase_with_google_pay` - I also have 5 failures on this branch, which is 1 more than the output you pasted in the commit message, so I assume it's this one. It's a test that was added recently with #4331
> 
> Here's the content of the error message for that failure: `"Error validating xml data against the schema: cvc-complex-type.4: Attribute 'id' must appear on element 'sale'.",`

remote test corrected

curiousepic (approved):
Thanks, merging!

Heavyblade (commented):
grate job @gasb150 looks as a pretty straight forward fix, the only thing to improve a little bit is the description on the commit and PR, could please first tell what was causing the bug and after that what is going to change this PR to fix that bug, so the person that reads the commit and doesn't have any context can understand.

PR #4342, opened 2022-03-01
Airwallex: Support gateway

therufs:
Unit: 16 tests, 57 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 16 tests, 43 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop: 731 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/airwallex.rb
+++ b/lib/active_merchant/billing/gateways/airwallex.rb
@@ -1,0 +1,249 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class AirwallexGateway < Gateway
+      self.test_url = 'https://api-demo.airwallex.com/api/v1'
+      self.live_url = 'https://pci-api.airwallex.com/api/v1'
+
+      # per https://www.airwallex.com/docs/online-payments__overview, cards are accepted in all EU countries
+      self.supported_countries = %w[AT AU BE BG CY CZ DE DK EE GR ES FI FR GB HK HR HU IE IT LT LU LV MT NL PL PT RO SE SG SI SK]
+      self.default_currency = 'AUD'
+      self.supported_cardtypes = %i[visa master]
+
+      self.homepage_url = 'https://airwallex.com/'
+      self.display_name = 'Airwallex'
+
+      ENDPOINTS = {
+        login: '/authentication/login',
+        setup: '/pa/payment_intents/create',
+        sale: '/pa/payment_intents/%{id}/confirm',
+        capture: '/pa/payment_intents/%{id}/capture',
+        refund: '/pa/refunds/create',
+        void: '/pa/payment_intents/%{id}/cancel'
+      }
+
+      def initialize(options = {})
+        requires!(options, :client_id, :client_api_key)
+        @client_id = options[:client_id]
+        @client_api_key = options[:client_api_key]
+        super
+        setup_ids(options) unless options[:request_id] && options[:merchant_order_id]
+        @access_token = setup_access_token
+      end
+
+      def purchase(money, card, options = {})
+        requires!(options, :return_url)
+        payment_intent_id = create_payment_intent(money, @options)
+        post = {
+          'request_id' => update_request_id(@options, 'purchase'),
+          'return_url' => options[:return_url]
+        }
+        add_card(post, card, options)
+        post['payment_method_options'] = { 'card' => { 'auto_capture' => false } } if authorization_only?(options)
+
+        commit(:sale, post, payment_intent_id)
+      end
+
+      def authorize(money, payment, options = {})
+        # authorize is just a purchase w/o an auto capture
+        purchase(money, payment, options.merge({ auto_capture: false }))
+      end
+
+      def capture(money, authorization, options = {})
+        raise ArgumentError, 'An authorization value must be provided.' if authorization.blank?
+
+        update_request_id(@options, 'capture')
+        post = {
+          'request_id' => @options[:request_id],
+          'amount' => amount(money)
+        }
+        commit(:capture, post, authorization)
+      end
+
+      def refund(money, authorization, options = {})
+        raise ArgumentError, 'An authorization value must be provided.' if authorization.blank?
+
+        post = {}
+        post[:amount] = amount(money)
+        post[:payment_intent_id] = authorization
+        post[:request_id] = update_request_id(@options, 'refund')
+        post[:merchant_order_id] = @options[:merchant_order_id]
+
+        commit(:refund, post)
+      end
+
+      def void(authorization, options = {})
+        raise ArgumentError, 'An authorization value must be provided.' if authorization.blank?
+
+        update_request_id(@options, 'void')
+        post = {}
+        post[:request_id] = @options[:request_id]
+        commit(:void, post, authorization)
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(/(\\\"number\\\":\\\")\d+/, '\1[REDACTED]').
+          gsub(/(\\\"cvc\\\":\\\")\d+/, '\1[REDACTED]')
+      end
+
+      private
+
+      def setup_ids(options)
+        request_id, merchant_order_id = generate_ids
+        options[:request_id] = options[:request_id] || request_id
+        options[:merchant_order_id] = options[:merchant_order_id] || merchant_order_id
+      end
+
+      def generate_ids
+        timestamp = (Time.now.to_f.round(2) * 100).to_i.to_s
+        [timestamp.to_s, "mid_#{timestamp}"]
+      end
+
+      def setup_access_token
+        token_headers = {
+          'Content-Type' => 'application/json',
+          'x-client-id' => @client_id,
+          'x-api-key' => @client_api_key
+        }
+        response = ssl_post(build_request_url(:login), nil, token_headers)
+        JSON.parse(response)['token']
+      end
+
+      def build_request_url(action, id = nil)
+        base_url = (test? ? test_url : live_url)
+        base_url + ENDPOINTS[action].to_s % { id: id }
+      end
+
+      def create_payment_intent(money, options = {})
+        post = {}
+        add_invoice(post, money, options)
+        post[:request_id] = options[:request_id]
+        post[:merchant_order_id] = options[:merchant_order_id]
+
+        response = commit(:setup, post)
+        response.params['id']
+      end
+
+      def add_billing(post, card, options = {})
+        return unless card_has_billing_info(card, options)
+
+        billing = post['payment_method']['card']['billing'] || {}
+        billing['email'] = options[:email] if options[:email]
+        billing['phone'] = options[:phone] if options[:phone]
+        billing['first_name'] = card.first_name
+        billing['last_name'] = card.last_name
+        billing['address'] = add_address(card, options) if card_has_address_info(card, options)
+
+        post['payment_method']['card']['billing'] = billing
+      end
+
+      def card_has_billing_info(card, options)
+        # These fields are required if billing data is sent.
+        card.first_name && card.last_name
+      end
+
+      def card_has_address_info(card, options)
+        # These fields are required if address data is sent.
+        options[:address1] && options[:country]
+      end
+
+      def add_address(card, options = {})
+        address = {}
+        address[:country_code] = options[:country]
+        address[:street] = options[:address1]
+        address[:city] = options[:city] if options[:city] # required per doc, not in practice
+        address[:postcode] = options[:zip] if options[:zip]
+        address[:state] = options[:state] if options[:state]
+        address
+      end
+
+      def add_invoice(post, money, options)
+        post[:amount] = amount(money)
+        post[:currency] = (options[:currency] || currency(money))
+      end
+
+      def add_card(post, card, options = {})
+        post['payment_method'] = {
+          'type' => 'card',
+          'card' => {
+            'expiry_month' => format(card.month, :two_digits),
+            'expiry_year' => card.year.to_s,
+            'number' => card.number.to_s,
+            'name' => card.name,
+            'cvc' => card.verification_value
+          }
+        }
+        add_billing(post, card, options)
+      end
+
+      def authorization_only?(options = {})
+        options.include?(:auto_capture) && options[:auto_capture] == false
+      end
+
+      def update_request_id(options, action)
+        options[:request_id] += "_#{action}"
+      end
+
+      def parse(body)
+        JSON.parse(body)
+      end
+
+      def commit(action, post, id = nil)
+        url = build_request_url(action, id)
+        post_headers = { 'Authorization' => "Bearer #{@access_token}", 'Content-Type' => 'application/json' }
+        response = parse(ssl_post(url, post_data(post), post_headers))
+
+        Response.new(
+          success_from(response),
+          message_from(response),
+          response,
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+      def void(authorization, options = {})
+        raise ArgumentError, 'An authorization value must be provided.' if authorization.blank?
+
+        update_request_id(@options, 'void')
+        post = {}
+        post[:request_id] = @options[:request_id]
+        commit(:void, post, authorization)
+      end
+
+      # def credit(money, credit_card, options = {})
```
therufs, line 82: We believe credit isn't supported, but awaiting confirmation.

test/unit/gateways/airwallex_test.rb
```diff
@@ -0,0 +1,200 @@
+require 'test_helper'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class AirwallexGateway
+      def setup_access_token
```
therufs, line 6: Shadowing the method here so we don't have to make remote calls to do unit tests.

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+          }
+        }
+        add_billing(post, card, options)
+      end
+
+      def add_payment_method_options(post, card, options = {})
+        return unless options.include?(:auto_capture) && options[:auto_capture] == false
+
+        post['payment_method_options'] = { 'card' => { 'auto_capture' => false } }
+      end
```
dsmcclain, line 200: The logic here is a little complex because you have to recognize that this is being called both from a `purchase` and an `authorize` transaction. I wonder if removing the guard clause and instead adding a conditional to the caller would make the relationship a little clearer, like this:

```
def purchase(money, card, options = {})
  payment_intent = create_payment_intent(money, @options)
  post = {
    'request_id' => update_request_id(@options, 'purchase'),
    'return_url' => options[:return_url]
  }
  add_card(post, card, options)
  add_payment_method_options(post, card, options) if authorization?(options)

  commit(:sale, post, payment_intent)
end

def authorization?(options)
  options.include?(:auto_capture) && options[:auto_capture] == false
end
```

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+          'return_url' => options[:return_url]
+        }
+        add_card(post, card, options)
+        add_payment_method_options(post, card, options)
+
+        commit(:sale, post, payment_intent)
+      end
+
+      def authorize(money, payment, options = {})
+        # authorize is just a purchase w/o an auto capture
```
dsmcclain, line 46: Thank you for these helpful comments.

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+        purchase(money, payment, options.merge({ auto_capture: false }))
+      end
+
+      def capture(money, authorization, options = {})
+        raise ArgumentError, 'An authorization value must be provided.' if authorization.blank?
```
dsmcclain, line 51: I like this pattern but from a Spreedly perspective I'm not sure it ends up having the intended effect. I will expand on this in Spreedly-specific channels.

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+        post[:request_id] = @options[:request_id]
+        commit(:void, post, authorization)
+      end
+
+      # def credit(money, credit_card, options = {})
+      # end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
```
dsmcclain, line 87: Did anyone reach out to Airwallex to confirm that $0 auths are not supported?

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+      def supports_scrubbing?
+        true
+      end
+
+      def remove_invalid_utf_8_byte_sequences(transcript)
+        transcript.encode(invalid: :replace, replace: '')
+      end
+
+      def scrub(transcript)
+        # clean_transcript = remove_invalid_utf_8_byte_sequences(transcript)
```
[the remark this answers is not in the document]
dsmcclain: It sounds like this is not an issue any more so I would remove it if it isn't strictly necessary. Same with the `remove_invalid_utf_8_byte_sequences` method above.

test/remote/gateways/remote_airwallex_test.rb
```diff
@@ -0,0 +1,140 @@
+require 'test_helper'
+
+class RemoteAirwallexTest < Test::Unit::TestCase
+  def setup
+    @gateway = AirwallexGateway.new(fixtures(:airwallex))
+
+    # https://www.airwallex.com/docs/online-payments__test-card-numbers
```
dsmcclain, line 7: Thanks I think we should do this every time.

test/remote/gateways/remote_airwallex_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    @amount = 100
+    @declined_amount = 8014
+    @credit_card = credit_card('4000100011112224')
+    @declined_card = credit_card('2223 0000 1018 1375')
+    @options = { return_url: 'https://example.com' }
```
dsmcclain, line 12: It seems like this field is necessary so I'm wondering if we should enforce its presence in the gateway adapter somehow? Or is there a plan to make sure that Spreedly's integration with Airwallex is always providing a default value here?

test/remote/gateways/remote_airwallex_test.rb
```diff
@@ -0,0 +1,140 @@
...
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount - 1, auth.authorization, @options)
+    assert_success capture
+  end
+
+  def test_failed_capture
+    response = @gateway.capture(@declined_amount, '12345', @options)
+    assert_failure response
+    assert_match(/The requested endpoint does not exist/, response.message)
```
dsmcclain, line 69: 🤔  This seems like a weird message to return

test/unit/gateways/airwallex_test.rb
```diff
@@ -0,0 +1,200 @@
...
+    assert_raise ArgumentError do
+      @gateway.void('', @options)
+    end
+  end
+
+  def test_failed_void; end
+
+  def test_successful_verify; end
+
+  def test_failed_verify; end
```
dsmcclain, line 132: Are these tests forthcoming?

dsmcclain (changes requested):
This is looking great. There is some complex logic in this gateway, well done making it all look pretty simple in code.

All tests are passing.

In addition to my comments below, I'm wondering if we should consider adding some tests covering the behavior of the `login` and `setup` api calls? Maybe just something showing that `setup_access_token` returns a token, as well as a failure case (if possible)?

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+        void: '/pa/payment_intents/%{id}/cancel'
+      }
+
+      def initialize(options = {})
+        requires!(options, :client_id, :client_api_key)
+        @client_id = options[:client_id]
+        @client_api_key = options[:client_api_key]
+        super
+        setup_ids(options) unless options[:request_id] && options[:merchant_order_id]
+        setup_access_token
```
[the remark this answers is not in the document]
therufs: @dsmcclain That's correct. I thought about doing something like returning the access token for future reuse, but that seems fraught with danger in terms of what happens if the token gets intercepted somehow.

lib/active_merchant/billing/gateways/airwallex.rb
therufs: @dsmcclain I don't remember if we did, but I do think I tried it and it did not work. It might be a thing that's settable at the account level, but I think it makes sense for that to be a thing we save for future development.

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,257 @@
...
+        billing['phone'] = options[:phone] if options[:phone]
+        billing['first_name'] = card.first_name
+        billing['last_name'] = card.last_name
+        billing['address'] = add_address(card, options) if card_has_address_info(card, options)
+
+        post['payment_method']['card']['billing'] = billing
+      end
+
+      def card_has_billing_info(card, options)
+        card.first_name && card.last_name
```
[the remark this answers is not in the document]
therufs: @dsmcclain That's correct; Airwallex rejects billing info that does not have both a first and last name.

test/remote/gateways/remote_airwallex_test.rb
therufs: This field is, in fact, necessary; I felt a little weird about `requires!`ing it in setup, since it's not really a credential, but maybe I should get over it. Airwallex does return a useful error that this integration then passes through, but maybe it's better for the complaint to come from the expected level (i.e., the AM API)?

test/unit/gateways/airwallex_test.rb
therufs: WOW well they certainly should be

lib/active_merchant/billing/gateways/airwallex.rb
therufs: Glad to help! Do you think we should leave them for posterity, or should I clean them up?

lib/active_merchant/billing/gateways/airwallex.rb
therufs: :brilliant:

test/remote/gateways/remote_airwallex_test.rb
therufs: 💡 Turns out I can `requires!` it in `purchase`, which is the only place it's actually needed!

lib/active_merchant/billing/gateways/airwallex.rb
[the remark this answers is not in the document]
dsmcclain: Yeah I agree.

Do the `access_token`s expire? Are they intended as a session-level token? If not, I could see maybe generating a token as part of an initial account setup and then saving it in a fixtures file along with other credentials, and only making a call to `setup_access_token` if there is no token present or if it has already expired.

With an endpoint named `login` my guess is that these tokens probably don't live long, though.

lib/active_merchant/billing/gateways/airwallex.rb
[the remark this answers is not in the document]
dsmcclain: Cool. I see you added a helpful comment to this method, thank you 👍

lib/active_merchant/billing/gateways/airwallex.rb
dsmcclain: I think we should leave them. They are succinct and serve to identify some of the gateway's quirks, which might cause a reader to raise an eyebrow otherwise.

lib/active_merchant/billing/gateways/airwallex.rb
dsmcclain: Ok, saving for future development seems totally reasonable to me. It will for sure be something customers ask for down the line.

test/remote/gateways/remote_airwallex_test.rb
dsmcclain: oh right, that makes more sense. Because the ID is in the url the endpoint literally doesn't exist.

Seems like a minor wart in the gateway API considering refund has a nicely-tailored message.

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -31,15 +31,16 @@ def initialize(options = {})
       end
 
       def purchase(money, card, options = {})
-        payment_intent = create_payment_intent(money, @options)
+        requires!(options, :return_url)
```
dsmcclain, line 34: I think this is the right idea. See my notes in Spreedly channels for ongoing discussion on this pattern.

test/unit/gateways/airwallex_test.rb
```diff
@@ -46,25 +45,37 @@ def test_failed_purchase
     assert_equal 'The card issuer declined this transaction. Please refer to the original response code.', response.message
   end
 
+  def test_purchase_without_return_url_raises_error
```
dsmcclain, line 48: 👍

dsmcclain (changes requested):
I like the updates, seems like things are progressing nicely.

Still marking as `request changes` cause it seems like you are working with the gateway regarding new failures from the sandbox. Also I can see there are still a few unit tests to be added for failed void & successful/failed verify.

lib/active_merchant/billing/gateways/airwallex.rb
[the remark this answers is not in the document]
therufs: They expire every 20 or 30 minutes (depending on whether you believe the doc or the specified expiration -- also possible these are different in dev and prod for some reason?)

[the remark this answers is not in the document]
therufs: They expire either 20 or 30 minutes after issue, depending on whether you trust the returned expiration timestamp or the documentation (also possible they are different in prod and sandbox?)

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,249 @@
...
+      self.supported_cardtypes = %i[visa master]
+
+      self.homepage_url = 'https://airwallex.com/'
+      self.display_name = 'Airwallex'
+
+      ENDPOINTS = {
+        login: '/authentication/login',
+        setup: '/pa/payment_intents/create',
+        sale: '/pa/payment_intents/%{id}/confirm',
+        capture: '/pa/payment_intents/%{id}/capture',
```
naashton, line 19: Nice, I know about string interpolation in Python but I didn't realize this was supported in Ruby

lib/active_merchant/billing/gateways/airwallex.rb
naashton: Is this comment in regards to the use of `requires!`?

lib/active_merchant/billing/gateways/airwallex.rb
naashton: This seems redundant to me. Ruby will already result in an `ArgumentError` if you try to call a method without passing all of the required (non-default) parameters.

lib/active_merchant/billing/gateways/airwallex.rb
therufs: I am 90% sure that it is :D

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -0,0 +1,249 @@
...
+      def success_from(response)
+        %w(REQUIRES_PAYMENT_METHOD SUCCEEDED RECEIVED REQUIRES_CAPTURE CANCELLED).include?(response['status'])
+      end
+
+      def message_from(response)
+        response.dig('latest_payment_attempt', 'status') || response['status'] || response['message']
+      end
+
+      def authorization_from(response)
+        response.dig('latest_payment_attempt', 'payment_intent_id')
```
naashton, line 241: Just bouncing around here, but if this fails to grab the `payment_intent_id` from the response, then the default value will just be `nil` in this case, and so subsequent transactions that rely on an `authorization` shouldn't yield a `blank` value.

naashton (approved):
Took a couple of pass throughs and the tests pass. Code looks great, just the small nitpick about whether or not we want to raise an exception if the authorization is blank (seems to me like something that won't happen unless Airwallex is giving us blank `payment_intent_id` or for some reason we fail to capture that value).

PR #4349, opened 2022-03-03
Priority: Update `add_purchases_data` method

drkjc:
The `add_purchases_data` method tried to call `empty?` on the options
hash. When the `:purchases` key was unavailable `empty?` was called on a
`nil` value which returned a no method error. This fix avoids the no
method error by using an `unless` statement to check that the options
key exists

CE-2443

Unit:

11 tests, 60 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:

20 tests, 57 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


meagabeth (approved):
This looks solid, tests pass! Reminder to add a changelog entry when you go through the process to rebase and push. 🚀

PR #4350, opened 2022-03-03
DecidirPlus: Handle `payment_method_id` by `card_brand`

naashton:
Determine `payment_method_id` based on `card_brand` passed in the
options hash for auth and purchase transactions

CE-2424

Unit: 13 tests, 53 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote0: 17 tests, 58 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir_plus.rb
+++ b/lib/active_merchant/billing/gateways/decidir_plus.rb
@@ -6,7 +6,7 @@ module ActiveMerchant #:nodoc:
 
       self.supported_countries = ['AR']
       self.default_currency = 'ARS'
-      self.supported_cardtypes = %i[visa master american_express discover]
+      self.supported_cardtypes = %i[visa master american_express discover diners_club naranja cabal]
 
       self.homepage_url = 'http://decidir.com.ar/home'
       self.display_name = 'Decidir Plus'
@@ -120,7 +120,7 @@ module ActiveMerchant #:nodoc:
 
       def add_purchase_data(post, money, payment, options = {})
         post[:site_transaction_id] = options[:site_transaction_id] || SecureRandom.hex
-        post[:payment_method_id] = 1
+        post[:payment_method_id] = add_payment_method_id(options)
         post[:amount] = money
         post[:currency] = options[:currency] || self.default_currency
         post[:installments] = options[:installments] || 1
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -144,6 +144,35 @@ def add_sub_payments(post, options)
         end
       end
 
+      def add_payment_method_id(options)
```
dsmcclain, line 147: Is it worth adding a unit test to cover the behavior of this method?

I'm inclined to say no, but I just wanted to bring it up in case you like the idea.

test/remote/gateways/remote_decidir_plus_test.rb
```diff
@@ -186,6 +189,16 @@ def test_successful_purchase_with_fraud_detection
...
   end
 
+  def test_successful_purchase_with_card_brand
+    options = @options.merge(card_brand: 'cabal')
+
+    assert response = @gateway_purchase.store(@cabal)
+    payment_reference = response.authorization
+
+    response = @gateway_purchase.purchase(@amount, payment_reference, options)
+    assert_success response
```
dsmcclain, line 199: Nit: do you think it would be worth including an assertion here showing that `response.params['payment_method_id']` is the expected value?

dsmcclain (approved):
All tests are passing.

The logic in this case statement is straightforward enough. I have no idea how this `payment_method_id` connects to the underlying issue with `bin` described in the ticket, but I trust that you have done your research and that this solves the problem.

Made a few nitpicky comments about testing, take them or leave them. Approving but happy to re-review if you want to make adjustments.

Don't forget the ChAnGeLoG! 😄

lib/active_merchant/billing/gateways/decidir_plus.rb
naashton: I thought about adding more test, but the urgency to get this in is rising, so I'm just working towards expediency.

test/remote/gateways/remote_decidir_plus_test.rb
naashton: Good suggestion, added.

PR #4352, opened 2022-03-07
CE- 2403 Priority: Add GSF `replay_id`

drkjc:
`replay_id` can be thought of like an `idempotency_key`. When `replay_id` is sent the expected behavior is that if the gateway can find a duplicate` replay_id` attached to previous transaction they return that original response. This only works with `purchase` and `authorize`. `voids` and `refund` should be sent w/ out `replay_id`. 

CE-2403

Local:
5070 tests, 75112 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
13 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
24 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_priority_test.rb
+++ b/test/remote/gateways/remote_priority_test.rb
@@ -13,6 +13,7 @@ class RemotePriorityTest < Test::Unit::TestCase
 
     # Run specific remote test
     # ruby -Itest test/remote/gateways/remote_priority_test.rb -n test_fail_refund_already_refunded_purchase_response
+
     @gateway = PriorityGateway.new(fixtures(:priority))
 
     # purchase params success
@@ -20,6 +21,7 @@ class RemotePriorityTest < Test::Unit::TestCase
     @credit_card = credit_card('4111111111111111', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
     @invalid_credit_card = credit_card('123456', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
     @faulty_credit_card = credit_card('12345', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
+    @replay_id = rand(100...1000)
 
     @option_spr = {
       billing_address: address(),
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_priority_test.rb
```diff
@@ -13,13 +13,18 @@ def setup
 
     # Run specific remote test
     # ruby -Itest test/remote/gateways/remote_priority_test.rb -n test_fail_refund_already_refunded_purchase_response
+
+    # If you are attempting to implement `replayId` on `refund` be aware of this caveat:
+    # When attempting duplicate refunds with the same `replayId` gsf, Priority returns this error: "Replay payment account #number does not match original." This is because the payment token is not associated with card information that is `stored` at Priority. When a card is vaulted at Priority and a duplicate refund is attempted the error returned will be "Payment already refunded". Customers should be made aware that using the GSF with a stored vs unstored card will return different results, though the final outcome of both should be a failed duplicate refund.
```
naashton, line 18: This feels like a long winded comment that people will probably gloss over and is only specific to the `replay_id` field. I don't think we necessarily need to explain _why_ we are doing things in test, but we could maybe link to the documentation for the specific test case for this field.

naashton: I think a better way to handle this is to use tests to show how this behavior works, which it looks like you are already doing with your test on ln306 and ln318.

naashton (approved):
All looks good!

test/remote/gateways/remote_priority_test.rb
drkjc: @naashton I see your point. The docs aren't specific about `refund` and `void`. I only learned this through my conversation with their support. I had originally written tests for `refund` and `void` but since they won't be usable with this GSF I didn't know that it was necessary. The response when the GSF is added is more confusing than leaving it out. Do you think I should add the `refund`/`void` tests back in or remove the comment?

naashton: I think it makes sense to add the failed test for refund, regardless of whether the response message is different for a `stored` vs `unstored` card. It'll still convey the same sentiment around the behavior.

PR #4075, opened 2021-08-03
Payflow: add THREEDSVERSION and DSTRANSACTIONID when present

bbraschi:
See https://developer.paypal.com/docs/payflow/3d-secure-mpi/

- Rubocop:
  ```
  707 files inspected, no offenses detected
  ```    

- Local tests:
    ```
    4848 tests, 73971 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed
    ```
    
- Remote tests:

   ```
   bundle exec rake test:remote TEST=test/remote/gateways/remote_payflow_test.rb
   ```
   | branch |  results |
   | -- | -- |
   | payflow_3ds2_update |  `38 tests, 163 assertions, 10 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 73.6842% passed` |
   | master | `38 tests, 163 assertions, 10 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 73.6842% passed` <br> the same |
   
   
   In the most recent Payflow PR, [here](https://github.com/activemerchant/active_merchant/pull/4066#issue-698834138), only 8 tests failed.

PR #3362, opened 2019-09-20
[3DS] Update status to accept the correct field in payflow integration

nebdil:
Due to missing information, we've been getting 3DS errors from payflow_uk:
`Field format error: 12000-Transaction is not compliant due to missing or invalid 3D-secure authentication values`
This PR adds the missing fields following [this documentation](https://developer.paypal.com/docs/classic/payflow/3d-secure-mpi/#how-to-send-3-d-secure-authentication-data)

PR #4356, opened 2022-03-11
Payflow Pro: Add 3DS Support

esmitperez:
This commit introduces MPI-compatible way to pass 3DS authentication
parameters, as documented in [Payflow's 3-D Secure with 3rd-Party Merchant Plug-ins](https://developer.paypal.com/api/nvp-soap/payflow/3d-secure-mpi/).

Prior to this change, we would pass the fields via the `BuyerAuthResult`
inside the `Card` node. Unfortunately, there's no explicit mention of
these, and as far as I can see, this was inferred via a series of
commits, and even then, some of the links referenced in them, are broken
:(, these are the commits:
- #3362
- #4075
- #4113
- #4168

Current Reference documentation, as of March 2022:
- [How to send 3-D Secure authentication data](https://developer.paypal.com/api/nvp-soap/payflow/3d-secure-mpi/#link-payflowfields)

"Legacy" documentation for 3DS data:
- [BuyerAuthStatusEnum, page 109](https://www.paypalobjects.com/webstatic/en_US/developer/docs/pdf/pp_payflowpro_xmlpay_guide.pdf)

ECS-2350

Note: the 10 failing tests are the same ones that fail in `master`.

Unit:
57 tests, 361 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
39 tests, 169 assertions, 10 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
74.359% passed


lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -202,6 +206,29 @@ def build_credit_card_request(action, money, credit_card, options)
...
+        ds_transaction_id = three_d_secure_options[:ds_transaction_id] if version_2_or_newer?(three_d_secure_options)
+
+        xml.tag! 'ExtData', 'Name' => 'AUTHENTICATION_ID', 'Value' => authentication_id unless authentication_id.blank?
+        xml.tag! 'ExtData', 'Name' => 'AUTHENTICATION_STATUS', 'Value' => authentication_status unless authentication_status.blank?
+
+        xml.tag! 'ExtData', 'Name' => 'CAVV', 'Value' => cavv unless cavv.blank?
+        xml.tag! 'ExtData', 'Name' => 'ECI', 'Value' => eci unless eci.blank?
+        xml.tag! 'ExtData', 'Name' => 'XID', 'Value' => xid unless xid.blank?
+        xml.tag! 'ExtData', 'Name' => 'THREEDSVERSON', 'Value' => version unless version.blank?
+        xml.tag! 'ExtData', 'Name' => 'DSTRANSACTIONID', 'Value' => ds_transaction_id unless ds_transaction_id.blank?
```
aenand, line 229: Quick look at the docs shows that we send certain fields depending on "cardholder enrollment in a 3-D Secure program", should/can we add logic here for that or are we expecting that to be accounted for earlier in the process?

aenand (changes requested):
Left a couple of comments. Also there are several rubocop issues. You can autofix them with `be rubocop -a`

lib/active_merchant/billing/gateways/payflow.rb
esmitperez: yeah, I felt setting those values would depend on the client application, but will look into incorporating that logic here too.

aenand: If it doesn't make sense to place here then let's not force it. I just wanted to double check!

lib/active_merchant/billing/gateways/payflow.rb
```diff
@@ -192,6 +192,10 @@ def build_credit_card_request(action, money, credit_card, options)
               xml.tag! 'TotalAmt', amount(money), 'Currency' => options[:currency] || currency(money)
             end
 
+            if [:authorization, :purchase].include? action
```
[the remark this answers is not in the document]
esmitperez: It doesn't make sense to run 3DS on non auth_purchases, the current implementation doesn't enforce that

aenand (approved):
LGTM 🚀

PR #4359, opened 2022-03-15
[PROPOSAL] Adding and configure VCR to be used on remote tests

Heavyblade:
Description
---
This PR Adds the VCR gem in order to record and stub the remote tests.

**Why?**
We need to be able to run the entire test suite each time we need so we are able to:

- Ensure that changes don't have side effects on the requests.
- Be able to make changes/improvements to base classes like CreditCard, Gateway, Check and be sure that there aren't side effects.
- Have more confidence to upgrade support higher Ruby versions and gem dependencies.
- Explore different HTTP clients to implement retry calls
- etc

**How?**

Trying to not be too intrusive a small module was introduced (VCR module), this module changes hooks into the test life cycle and stores a cassette for each test.

**Note:** As an implementation example, this PR adds the VCRmodule for the Braintree token remote tests.

Test Suite
---

**Unit Tests:**
Finished in 31.581706 seconds.
5112 tests, 75300 assertions, 0 failures, 2 errors,
0 pendings, 0 omissions, 0 notifications.
100% passed

**Remote Tests:**
bundle exec ruby -Itest
test/remote/gateways/remote_braintree_token_nonce_test.rb

Finished in 0.343357 seconds.
5 tests, 10 assertions, 0 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications
100 passed

**Rubocop**
737 files inspected, no offenses detected

### DIFF
--- a/activemerchant.gemspec
+++ b/activemerchant.gemspec
@@ -31,6 +31,9 @@ Gem::Specification.new do |s|
   s.add_development_dependency('pry')
   s.add_development_dependency('pry-byebug')
   s.add_development_dependency('rake')
+  s.add_development_dependency('rexml', '3.1.9')
   s.add_development_dependency('test-unit', '~> 3')
   s.add_development_dependency('thor')
+  s.add_development_dependency('vcr')
+  s.add_development_dependency('webmock')
 end
--- a/lib/active_merchant/billing/gateways/braintree/braintree_common.rb
+++ b/lib/active_merchant/billing/gateways/braintree/braintree_common.rb
@@ -23,6 +23,7 @@ module BraintreeCommon
       gsub(%r((<payment-method-nonce>)[^<]+(</payment-method-nonce>)), '\1[FILTERED]\2').
       gsub(%r((<payment-method-token>)[^<]+(</payment-method-token>)), '\1[FILTERED]\2').
       gsub(%r((<value>)[^<]{100,}(</value>)), '\1[FILTERED]\2').
-      gsub(%r((<token>)[^<]+(</token>)), '\1[FILTERED]\2')
+      gsub(%r((<token>)[^<]+(</token>)), '\1[FILTERED]\2').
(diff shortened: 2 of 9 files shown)


javierpedrozaing (commented):
@Heavyblade Nice job, just I leave a question. Thx

aenand:
Thank you for looking to add this I think it will add a lot to the testing stability of the gem! I have some some questions on this:

- Some keys used in the remote tests are sensitive to a particular company, does VCR record these values or make them accessible in anyway?
- Do we need to add the VCR module to each gateway's remote files to record them? Will this add many new files to the repository?
- Does the gem need to be specified as needed for the AM gem to be installed or is there a way to specify it as a dev dependency?

aenand:
Also how do we run the remote tests vs the VCR recorded remote test?

fatcatt316:
Just curious... we were talking about VCR on our team, and I searched around and found this PR. What are y'all's thoughts on adding VCR? It seems like it could be helpful for remote tests.

Heavyblade:
Hey @fatcatt316, yes the initial intention was to add VCR to have a way add real request to the test suite to add more coverage, this PR description list the pros and this PR implements a "framework" to progressible implement VCR gateway by gateway without the need to modify test by test and also for safety uses the gateway `scrub` method to clean the cassette.

let me know if you want to discuss in depth about this, I would love to revive this work.

fatcatt316:
@Heavyblade Yeah, me and @BritneyS were just talking about this yesterday, and wondered why Spreedly wasn't using VCR for remote gateway tests. What stopped y'all from merging this in, and do you think it'd be worth reviving it now?

Heavyblade:
Yes I still think that VCR could add a lot of value to increase the confidence and test coverage, Honestly I just closed this PR because it seemed at that time that there wasn't interest in that and as contractor I don't have the leverage to push more about this.

PR #4360, opened 2022-03-17
Litle: Add support for Level 2 and 3 enhanced data

jherreraa:
Litle: Add support for Level 2 and 3 enhanced data
------

Summary:
This PR includes the enhanced data to enable level 2 and 3 data support in this gateway

Test Execution:
-------
Finished in 37.183788 seconds.
54 tests, 221 assertions, 13 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
75.9259% passed

Failing tests are not related to this requirement

RuboCop:
736 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -39,6 +39,96 @@ module ActiveMerchant #:nodoc:
         check?(payment_method) ? commit(:echeckSales, request, money) : commit(:sale, request, money)
       end
 
+      def add_level_two_data(doc, payment_method, options = {})
+        level_2_data = options[:level_2_data]
+        if level_2_data
+          doc.enhancedData do
+            case payment_method.brand
+            when 'visa'
+              doc.salesTax(level_2_data[:sales_tax]) if level_2_data[:sales_tax]
+            when 'master'
+              doc.customerReference(level_2_data[:customer_code]) if level_2_data[:customer_code]
+              doc.salesTax(level_2_data[:total_tax_amount]) if level_2_data[:total_tax_amount]
+              doc.detailTax do
+                doc.taxIncludedInTotal(level_2_data[:tax_included_in_total]) if level_2_data[:tax_included_in_total]
+                doc.taxAmount(level_2_data[:tax_amount]) if level_2_data[:tax_amount]
+                doc.cardAcceptorTaxId(level_2_data[:card_acceptor_tax_id]) if level_2_data[:card_acceptor_tax_id]
+              end
+            end
+          end
+        end
+      end
+
+      def add_level_three_data(doc, payment_method, options = {})
+        level_3_data = options[:level_3_data]
+        if level_3_data
+          doc.enhancedData do
+            case payment_method.brand
+            when 'visa'
+              add_level_three_information_tags_visa(doc, payment_method, level_3_data)
+            when 'master'
+              add_level_three_information_tags_master(doc, payment_method, level_3_data)
+            end
+          end
+        end
+      end
+
+      def add_level_three_information_tags_visa(doc, payment_method, level_3_data)
+        doc.discountAmount(level_3_data[:discount_amount]) if level_3_data[:discount_amount]
+        doc.shippingAmount(level_3_data[:shipping_amount]) if level_3_data[:shipping_amount]
+        doc.dutyAmount(level_3_data[:duty_amount]) if level_3_data[:duty_amount]
+        doc.detailTax do
+          doc.taxIncludedInTotal(level_3_data[:tax_included_in_total]) if level_3_data[:tax_included_in_total]
+          doc.taxAmount(level_3_data[:tax_amount]) if level_3_data[:tax_amount]
+          doc.taxRate(level_3_data[:tax_rate]) if level_3_data[:tax_rate]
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -39,6 +39,88 @@ def purchase(money, payment_method, options = {})
...
+              doc.detailTax do
+                doc.taxIncludedInTotal(level_3_data[:tax_included_in_total]) if level_3_data[:tax_included_in_total]
+                doc.taxAmount(level_3_data[:tax_amount]) if level_3_data[:tax_amount]
+                doc.taxRate(level_3_data[:tax_rate]) if level_3_data[:tax_rate]
+                doc.taxTypeIdentifier(level_3_data[:tax_type_identifier]) if level_3_data[:tax_type_identifier]
+                doc.cardAcceptorTaxId(level_3_data[:card_acceptor_tax_id]) if level_3_data[:card_acceptor_tax_id]
+              end
+              add_line_item_information_for_level_3_visa(doc, payment_method, level_3_data)
+            when 'master'
+              doc.customerReference :customerReference, level_3_data[:customer_code] if level_3_data[:customer_code]
```
javierpedrozaing, line 80: ❓ 
Why the customerReferencie is repeated here in level_3_data? I thing it is for level II only.

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -39,6 +39,88 @@ def purchase(money, payment_method, options = {})
         check?(payment_method) ? commit(:echeckSales, request, money) : commit(:sale, request, money)
       end
 
+      def add_level_2_data(doc, payment_method, options = {})
```
javierpedrozaing, line 42: 💬 
you can change the name to add_level_two_data and and_level_three_data instead of use numbers in the method name.

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -39,6 +39,88 @@ def purchase(money, payment_method, options = {})
...
+                doc.taxIncludedInTotal(level_2_data[:tax_included_in_total]) if level_2_data[:tax_included_in_total]
+                doc.taxAmount(level_2_data[:tax_amount]) if level_2_data[:tax_amount]
+                doc.cardAcceptorTaxId(level_2_data[:card_acceptor_tax_id]) if level_2_data[:card_acceptor_tax_id]
+              end
+            end
+          end
+        end
+      end
+
+      def add_level_3_data(doc, payment_method, options = {})
```
javierpedrozaing, line 62: 💬 
in order to improve the method size, you could create a method to get the visa card tags and other one for mastercard tags.

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -39,6 +39,88 @@ def purchase(money, payment_method, options = {})
...
+          level_3_data[:line_items].each do |line_item|
+            doc.itemSequenceNumber(line_item[:item_sequence_number]) if line_item[:item_sequence_number]
+            doc.commodityCode(line_item[:commodity_code]) if line_item[:commodity_code]
+            doc.itemDescription(line_item[:item_description]) if line_item[:item_description]
+            doc.productCode(line_item[:product_code]) if line_item[:product_code]
+            doc.quantity(line_item[:quantity]) if line_item[:quantity]
+            doc.unitOfMeasure(line_item[:unit_of_measure]) if line_item[:unit_of_measure]
+            doc.taxAmount(line_item[:tax_amount]) if line_item[:tax_amount]
+            doc.itemDiscountAmount(line_item[:discount_per_line_item]) if line_item[:discount_per_line_item]
+            doc.unitCost(line_item[:unit_cost]) if line_item[:unit_cost]
```
javierpedrozaing, line 112: 💬 
According the documentation you would need add a validation to avoid 0 values in unit Cost tag. and check the same for item Discount Amount.

javierpedrozaing (changes requested):
Good job 💪 , Just a couple of comments to keep in mind.

lib/active_merchant/billing/gateways/litle.rb
jherreraa: Screen Shot 2022-03-23 at 3 12 28 PM
As you can see in the attached image, for mastercard the customerReference field is a field used for both level II and level III data

lib/active_merchant/billing/gateways/litle.rb
jherreraa: thanks for the suggestion, updated!

lib/active_merchant/billing/gateways/litle.rb
jherreraa: thanks for the suggestion, updated!

lib/active_merchant/billing/gateways/litle.rb
jherreraa: Corrected, please review.

lib/active_merchant/billing/gateways/litle.rb
javierpedrozaing: Ohh, I can see, ok

javierpedrozaing (commented):
Good job, LGTM

curiousepic (approved):
Looks great - will merge this in, but we should be sure to cover it in certification if there are required test cases for these fields.

PR #4364, opened 2022-03-22
DecidirPlus: Update `error_code` method

ajawadmirza:
Updated `error_code_from` method to contain reason id in the error code
response.

CE-2490

Unit:
5113 tests, 75341 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:
736 files inspected, no offenses detected

Remote:
23 tests, 77 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


naashton (changes requested):
This all looks good, but I'm wondering if we can't get a little bit more test coverage to handle the other conditions we might see in a response from D+.

I'm not sure which case the remote test ends up covering, but it looks like the unit test covers the first condition. It looks like there is at least one existing `response['error_type']` in the unit test in the `failed_refund_response` method.

naashton (approved):
🚢

PR #4372, opened 2022-03-25
Rapyd: Add gateway support

meagabeth:
- To note: the gateway doesn’t support void transactions for card payment methods, they must be run as refunds
- The gateway does not accept a refund unless the original transaction has been completed/closed/captured. Due to this, the verify method has to run a purchase and then a refund as opposed to the standard authorize and void

CE-2394

Local:
5127 tests, 75381 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Unit:
15 tests, 44 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
17 tests, 47 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+      def void(authorization, options = {})
+        post = {}
+        commit(:delete, "payments/#{authorization}", post)
+      end
+
+      # Gateway returns an error if trying to run a $0 auth as invalid payment amount
+      # Gateway does not support void on a card transaction and refunds can only be done on completed transactions (such as a purchase). Authorize transactions are considered 'active' and not 'complete'  until they are captured.
+      def verify(credit_card, options = {})
+        MultiResponse.run do |r|
+          r.process { purchase(100, credit_card, options) }
```
dsmcclain, line 71: I know on a few other gateways that support payments in many currencies we've run into issues where `100` is below the minimum amount in that currency. I'm wondering if we should test this out with a few different currencies to see how the gateway responds? I think the Mexican Peso and the Japanese Yen are good ones to test this way.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+      def add_payment(post, payment, options)
+        type = options[:type]
+
+        post[:payment_method] = {}
+        post[:payment_method][:fields] = {}
+        post[:payment_method][:type] = type
+
+        if type.include?('card')
+          add_creditcard(post, payment)
+        elsif type.include?('bank')
```
dsmcclain, line 122: Seems like a logical place to use your lovely `payment_is_ach?` method. Also seems like a corresponding `payment_is_card?` method could be useful on lines 26 and 120.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+        post[:amount] = amount(money).to_f.to_s
+        post[:currency] = (options[:currency] || currency(money))
+      end
+
+      def add_payment(post, payment, options)
+        type = options[:type]
+
+        post[:payment_method] = {}
+        post[:payment_method][:fields] = {}
+        post[:payment_method][:type] = type
```
dsmcclain, line 118: Stylistic nitpick: I know these lines are common to both `add_credit_card` and `add_ach` but I'm wondering if they more appropriately belong in each of those methods, so that `add_payment` simply switches on `payment_is_card?` or `payment_is_ach?`.

If you don't like that idea, it might be nice instead to include a guard clause in `add_credit_card` and `add_ach`, like:
```
return unless (pm_fields = post[:payment_method][:fields])

pm_fields[:number] = payment.number
etc...
```

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+
+      def generate_hmac(rel_path, salt, timestamp, payload)
+        signature = "#{rel_path}#{salt}#{timestamp}#{@options[:access_key]}#{@options[:secret_key]}#{payload}"
+        hash = Base64.urlsafe_encode64(OpenSSL::HMAC.hexdigest('sha256', @options[:secret_key], signature))
+        hash
+      end
+
+      def avs_result(response)
+        case resp = response['data']
+        when !resp.nil?
```
dsmcclain, line 192: I will admit being confused by `!resp.nil?`

Could this method be rewritten like:
```
def avs_result(response)
  return nil unless (code = response.dig('data', 'payment_method_data', 'acs_check'))

  AVSResult.new(code: code)
end
```
...or maybe I am misunderstanding what is happening here?

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+          AVSResult.new(code: response['data']['payment_method_data']['acs_check'])
+        when nil
+          nil
+        end
+      end
+
+      def cvv_result(response)
```
dsmcclain, line 199: same comment as above

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+        case resp = response['data']
+        when !resp.nil?
+          CVVResult.new(response['data']['payment_method_data']['cvv_check'])
+        when nil
+          nil
+        end
+      end
+
+      def success_from(response)
+        response['status']['status'] == 'SUCCESS' && response['data']['status'] != 'ERR'
```
dsmcclain, line 209: Wondering if we should protect against unexpected data structures by using `response.dig('status', 'status')`, etc.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+
+      def message_from(response)
+        case response['status']['status']
```
dsmcclain, line 213: same comment as above

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+        when 'ERROR'
+          response['status']['message'] == '' ? response['status']['error_code'] : response['status']['message']
+        else
+          response['status']['status']
+        end
+      end
+
+      def authorization_from(response)
+        response['data'] ? response['data']['id'] : response['status']['operation_id']
```
dsmcclain, line 222: same comment as above

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+
+      def error_code_from(response)
+        response['status']['error_code'] unless success_from(response)
```
dsmcclain, line 226: same comment as above

test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -0,0 +1,156 @@
...
+    @declined_card = credit_card('4111111111111105')
+    @check = check
+    @options = {
+      type: 'us_visa_card',
+      currency: 'USD'
+    }
+    @ach_options = {
+      type: 'us_ach_bank',
+      currency: 'USD',
+      proof_of_authorization: true,
```
dsmcclain, line 18: Is this field always gonna be `true`?

test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -0,0 +1,156 @@
...
+  def test_successful_purchase
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal 'SUCCESS', response.message
+  end
+
+  def test_successful_purchase_with_address
+    options = {
+      address: {
+        name: 'Henry Winkler',
```
dsmcclain, line 32: 😄

dsmcclain (changes requested):
Nice work @meagabeth.

I have a few comments and suggestions. Many are small stylistic suggestions to either make the code easier to maintain or else protect us against unexpected responses.

All tests currently passing.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+        requires!(options, :secret_key, :access_key)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        post = {}
+        add_invoice(post, money, options)
+        add_payment(post, payment, options)
+        add_address(post, payment, options)
+        post[:capture] = true if options[:type].include?('card')
```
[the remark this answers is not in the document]
meagabeth: Great questions! Yes, this will need to be user-supplied and it will be up to the user to know what value to provide based on the customer's country (of which this gateway supports MANY). We will be including details on this field in the public documentation, but here's a [link to the gateway's docs ](https://docs.rapyd.net/build-with-rapyd/reference/payment-method-type-object#list-payment-methods-by-country)related to this.

lib/active_merchant/billing/gateways/rapyd.rb
meagabeth: Good call out. I've added remote tests for peso and yen and both pass without a change in the amount 🎉

lib/active_merchant/billing/gateways/rapyd.rb
meagabeth: Done and done! 🧠

lib/active_merchant/billing/gateways/rapyd.rb
meagabeth: This is much cleaner than my solution, thank you! Revised!

test/remote/gateways/remote_rapyd_test.rb
meagabeth: If the merchant is following protocol and getting authorization from their customer prior to processing an ACH then yes, it'll be true. I tested it with `false` and it'll still pass at the gateway level. This is only a required field for ACH payments.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,239 @@
...
+
+        if payment_is_ach?(options)
+          MultiResponse.run do |r|
+            r.process { commit(:post, 'payments', post) }
+            post = {}
+            post[:token] = r.authorization
+            post[:param2] = r.params['data']['original_amount'].to_s
```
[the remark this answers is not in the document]
meagabeth: Good point! Revised to utilize dig 🧠

lib/active_merchant/billing/gateways/rapyd.rb
meagabeth: I think I revised this to be what you suggested, but let me know if I didn't get it quite right 😅

dsmcclain: Nailed it 💪

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,243 @@
...
+        commit(:post, 'refunds', post)
+      end
+
+      def void(authorization, options = {})
+        post = {}
+        commit(:delete, "payments/#{authorization}", post)
+      end
+
+      # Gateway returns an error if trying to run a $0 auth as invalid payment amount
+      # Gateway does not support void on a card transaction and refunds can only be done on completed transactions (such as a purchase). Authorize transactions are considered 'active' and not 'complete'  until they are captured.
```
dsmcclain, line 71: Thank you for the nice comment here. The flow is atypical and I think the explanation is warranted. Your logic makes sense to me. 👍 

Do you know if the merchant(s) requesting integration with this gateway are aware of this quirk of our implementation? Namely, that a `verify` call will in fact charge a card a set amount and then void that charge? It might be worth passing this info along to them because I think running a `purchase` here is not typical and not something a merchant would expect from `verify`.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,243 @@
...
+      end
+
+      private
+
+      def payment_is_ach?(options)
+        return true if options[:type].include?('bank')
+      end
+
+      def payment_is_card?(options)
+        return true if options[:type].include?('card')
```
dsmcclain, line 97: I love these little methods. I have two questions about the logic here.

1) Are these the only two options? Or could somebody pass a valid `type` that does not include `'bank'` or `'card'`?

2) Is there a world where something like `mx_nationalbank_card` is a valid `type`?

test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -0,0 +1,176 @@
...
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(options))
+    assert_success response
+    assert_equal 'SUCCESS', response.message
+  end
+
+  def test_successful_purchase_using_ach
+    response = @gateway.purchase(100000, @check, @ach_options)
+    assert_success response
+    assert_equal 'SUCCESS', response.message
+    assert_equal 'CLO', response.params['data']['status']
```
dsmcclain, line 46: Nitpicky question: Is there any benefit to asserting the value of this `status` response field?

test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -0,0 +1,176 @@
...
+  end
+
+  def test_successful_verify_with_peso
+    options = {
+      type: 'mx_visa_card',
+      currency: 'MXN'
+    }
+    response = @gateway.verify(@credit_card, options)
+    assert_success response
+    assert_equal 'SUCCESS', response.message
```
dsmcclain, line 138: nice 🇲🇽

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -0,0 +1,243 @@
+require 'pry'
+require 'pry-byebug'
```
dsmcclain, line 2: Oops! Please remove these.

dsmcclain (approved):
Lookin clean @meagabeth!

One request to remove the `require 'pry'` lines.

I also asked a few more questions, but you are free to ponder them without me (or maybe you already have).

I am approving, and I trust you to make the `'pry'` changes on your own. If there is anything larger that needs to change feel free to tag me for another review.

DoNt fOrGeT ChAnGeLoG!!

test/remote/gateways/remote_rapyd_test.rb
meagabeth: Great question and yes! ACH transactions have to go through a multi-step (aka multi-api-call) process to be completed. The `CLO` here means it is CLOSED at the gateway level. If the multi part process had not been successful, it would have a status of `ACT` for ACTIVE. We want to make sure it closes out for purchase, so that's why that status value is being checked.

PR #4373, opened 2022-03-25
Moneris: Add 3DS MPI Fields Support

esmitperez:
The field names, test values and logic parsing follow the guides
available in Moneri's [About 3D Secure
2.0](https://developer.moneris.com/livedemo/3ds2/reference/guide/php)
  guides.

**Caveats**

The available magic test data do not make it possible to trigger
reliable fail-scenarios for MPI-driven authentications.

ECS-1882

---

Test Summary

Note: failing tests are same ones as in `master` branch

Local: 5125 tests, 75361 assertions, 0 failures, 17 errors, 0 pendings, 0 omissions, 0 notifications
99.6683% passed

Unit: 52 tests, 279 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 44 tests, 223 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -73,15 +74,16 @@ def purchase(money, creditcard_or_datakey, options = {})
...
-        action = if post[:cavv]
+        action = if post[:cavv] || options[:three_d_secure]
                    'cavv_purchase'
                  elsif post[:data_key].blank?
                    'purchase'
                  else
                    'res_purchase_cc'
                  end
-        commit(action, post)
+        commit(action, post, options[:three_d_secure].present?)
```
curiousepic, line 86: I'd have a mild preference to pass the entire options hash into the commit method, and have the response dig into the 3ds hash, just to have all options available by default for later features if necessary. What do you think?

lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -314,7 +332,15 @@ def authorization_from(response = {})
       end
 
       # Tests for a successful response from Moneris' servers
-      def successful?(response)
+      def successful?(action, response, threed_ds_transaction = false)
+        # See 9.4 CAVV Result Codes in https://developer.moneris.com/livedemo/3ds2/reference/guide/php
+        cavv_accepted = if threed_ds_transaction
+                          response[:cavv_result_code] && response[:cavv_result_code] == "2"
+                        else
+                          true
```
aenand, line 340: Why is the default `true`?

aenand (changes requested):
Logic looks good! I had a question about one part. I did find some rubocop issues that need to be fixed. You can see them by running `bundle exec rake test:local`

curiousepic:
Refactor LGTM - I'll let Amit approve.

lib/active_merchant/billing/gateways/moneris.rb
esmitperez: thanks for the review, Amit! it's true so that for NON 3DS transactions we accept whatever cavv condition there is (`nil` or not even present). For 3DS transactions we want to verify that it's `2`.

Will fix rubocop!

aenand (approved):
LGTM :shipit: don't forget a CHANGELOG entry!

PR #4377, opened 2022-04-02
Airwallex: QA round 2

therufs:
Unit:
22 tests, 82 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
100% passed

Rubocop:


lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -134,44 +134,52 @@ def build_request_url(action, id = nil)
       def create_payment_intent(money, options = {})
         post = {}
         add_invoice(post, money, options)
+        add_order(post, options)
         post[:request_id] = options[:request_id]
         post[:merchant_order_id] = options[:merchant_order_id]
 
         response = commit(:setup, post)
+        raise ArgumentError.new(response.message) unless response.success?
```
dsmcclain, line 142: This solution makes sense, and I see that you have added logic on Spreedly's end of the integration to handle this. I am curious whether you considered wrapping `create_payment_intent` in some logic to preempt the `:sale` call and return a `failure` response if this initial `:setup` call fails?

Basically I'm just wondering if there is a reason that raising an error is preferable to returning a `failure` response in cases where `create_payment_intent` fails?

dsmcclain (approved):
Looks good. I asked a question simply for my own education/posterity.

All tests pass and I will merge this later today.

PR #4379, opened 2022-04-05
Airwallex: Add `descriptor` field

dsmcclain:
CE-2530

Rubocop:

Unit:
5145 tests, 75476 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

Remote:
Loaded suite test/remote/gateways/remote_airwallex_test

### DIFF
--- a/lib/active_merchant/billing/gateways/airwallex.rb
+++ b/lib/active_merchant/billing/gateways/airwallex.rb
@@ -31,20 +31,21 @@ module ActiveMerchant #:nodoc:
 
       def purchase(money, card, options = {})
         requires!(options, :return_url)
-        setup_ids(options)
+
         payment_intent_id = create_payment_intent(money, options)
         post = {
-          'request_id' => update_request_id(options, 'purchase'),
+          'request_id' => request_id(options),
+          'merchant_order_id' => merchant_order_id(options),
           'return_url' => options[:return_url]
         }
         add_card(post, card, options)
+        add_descriptor(post, options)
         post['payment_method_options'] = { 'card' => { 'auto_capture' => false } } if authorization_only?(options)
 
         commit(:sale, post, payment_intent_id)
       end
 
       def authorize(money, payment, options = {})
-        setup_ids(options)
         # authorize is just a purchase w/o an auto capture
         purchase(money, payment, options.merge({ auto_capture: false }))
       end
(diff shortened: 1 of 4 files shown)


test/unit/gateways/airwallex_test.rb
```diff
@@ -179,15 +179,24 @@ def test_purchase_passes_request_id
     stub_comms do
       @gateway.purchase(@amount, @credit_card, @options.merge(request_id: request_id))
     end.check_request do |_endpoint, data, _headers|
-      assert_match(/request_/, data)
+      assert_match(request_id, data)
```
dsmcclain, line 182: This change is unrelated to the `descriptor` field. Because the request contains a field named `request_id`, this assertion was not really testing the presence of the `request_{timestamp}` value.

test/unit/gateways/airwallex_test.rb
```diff
@@ -179,15 +179,24 @@ def test_purchase_passes_request_id
...
     end.respond_with(successful_purchase_response)
   end
 
   def test_purchase_passes_currency_code
     stub_comms do
       @gateway.purchase(@amount, @credit_card, @options.merge(currency: 'USD'))
     end.check_request do |_endpoint, data, _headers|
-      assert_match(/request_/, data)
+      # only look for currency code on the create_payment_intent request
+      assert_match(/USD/, data) unless data.include?('_purchase')
```
dsmcclain, line 191: This change is unrelated to the `descriptor` field. The previous assertion was not testing the currency code.

test/unit/gateways/airwallex_test.rb
rachelkirk: Nice, thanks for the description 😉

rachelkirk (approved):
Tests pass, nice fixes in the unit test. ✨ 🚢 ✨

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -139,8 +136,8 @@ def create_payment_intent(money, options = {})
         post = {}
         add_invoice(post, money, options)
         add_order(post, options)
-        post[:request_id] = options[:request_id]
-        post[:merchant_order_id] = options[:merchant_order_id]
+        post[:request_id] = "#{request_id(options)}_setup"
+        post[:merchant_order_id] = "#{merchant_order_id(options)}_setup"
```
dsmcclain, line 140: Airwallex will fail a transaction if it uses a `request_id` or `merchant_order_id` that has been used in a previous transaction. Because their typical flow involves two transactions (the `create_payment_intent` followed by some sort of `purchase`), we have chosen to append this suffix to the ids on the `create_payment_intent` call in order to make them different than the subsequent `purchase` call.

rachelkirk: This is very readable code. Thanks for adding this explanation, I always learn something when reviewing your PR's.

rachelkirk (approved):
Tests pass, this looks great! 🚀

PR #4383, opened 2022-04-06
Priority: Refactor gateway integration, add additional fields to request

dsmcclain:
CE-2491

This PR makes significant adjustments to the Priority gateway integration.

I will describe some of the changes below, but I also want to note two areas that warrant additional investigation:
1. the`verify` action and its dependency `create_jwt`. These methods are using the gateway's BIN lookup service, which is not a standard approach to verifying a card.
2. the `get_payment_status` and `close_batch` actions. It's not clear why these actions were implemented, as they are currently being used exclusively to manipulate transactions within the test cases. A merchant going through Spreedly would have no way to take advantage of these actions.

I will open some new tickets to investigate these two areas.

 The changes made in this PR include:
* The  `pos_data` object has been extracted to its own method and is now added line-by-line to the request
* The various `shipping_data` fields have their own method now.
* The `verify` method, as well as the auxiliary `get_payment_status`, `close_batch`, and `create_jwt` methods, have been simplified to conform to Active Merchant standards.
* The `authorization_from` method has been updated to conform to Active Merchant standards. Instead of returning a ruby hash it now returns a string of format `"<request_id>|<payment_token>"`
* Handling of gateway responses has been refactored
* Substantial changes were made to the test files. Instead of including every option in every test case, we are now including the minimum options required to test the behavior in question.
* There were multiple remote tests which rescued a runtime error, meaning that the transaction being tested would cause an uncaught error in production. The behavior under test has been corrected and tests that rescue errors have been reworked or removed.
* Test coverage has been added for fields whose presence were not previously tested.

In addition to the above changes, the following fields were removed from requests to the gateway, because they are either not mentioned in the gateway documentation, not necessary for the request, or else described as gateway response fields (meaning that they do not belong in the request):
* `authCode`
* `cardAccount.code`
* `cardAccount.taxRate`
* `cardAccount.entryMode`
* `cardAccount.cardId`
* `cardAccount.hasContract`
* `cardAccount.isCorp`
* `cardAccount.isDebit`
* `cardAccount.token`
* `cardPresent`
* `cardPresentType`
* `isTicket`
* `mxAdvantageEnabled`
* `sourceZip`
* `purchases.dateCreated`
* `purchases.iId`
* `purchases.transactionIId`
* `purchases.transactionId`

The following fields were removed from the gateway adapter because they were actually dead code, meaning they were never even added to requests in the first place (from the `add_risk_data` method):
* `cvvResponseCode`
* `cvvResponse`
* `cvvMatch`
* `avsResponse`
* `avsAddressMatch`
* `avsZipMatch`

The following fields were removed from requests to the gateway because they were not correctly implemented according to the gateway's documentation and they represent features that are currently not supported by Spreedly's integration with the gateway:
* `shouldVaultCard`

---

Rubocop:

Unit:
5147 tests, 75501 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
27 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_priority_test.rb
+++ b/test/remote/gateways/remote_priority_test.rb
@@ -2,323 +2,268 @@ require 'test_helper'
 
 class RemotePriorityTest < Test::Unit::TestCase
   def setup
-    # Consumer API Key: Generated in MX Merchant for specific test merchant
-    # Consumer API Secret:= Generated in MX Merchant for specific test merchant
-
-    # run command below to run tests in debug (byebug)
-    # byebug -Itest test/unit/gateways/card_stream_test.rb
-    #
-    # bundle exec rake test:remote TEST=test/remote/gateways/remote_priority_test.rb
-    # ruby -Itest test/unit/gateways/priority_test.rb -n test_successful_void
-
-    # Run specific remote test
-    # ruby -Itest test/remote/gateways/remote_priority_test.rb -n test_fail_refund_already_refunded_purchase_response
-
     @gateway = PriorityGateway.new(fixtures(:priority))
 
-    # purchase params success
-    @amount_purchase = 2
-    @credit_card = credit_card('4111111111111111', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
-    @invalid_credit_card = credit_card('123456', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
-    @faulty_credit_card = credit_card('12345', month: '01', year: '2029', first_name: 'Marcus', last_name: 'Rashford', verification_value: '999')
-    @replay_id = rand(100...1000)
+    @amount = 2
+    @credit_card = credit_card
+    @invalid_credit_card = credit_card('123456')
+    @replay_id = rand(100...99999999)
+    @options = { billing_address: address }
 
-    @option_spr = {
-      billing_address: address(),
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_priority_test.rb
```diff
@@ -2,323 +2,268 @@
...
     assert_equal 'CVV is required based on merchant fraud settings', response.message
-    assert_equal 'Declined', response.params['status']
   end
 
-  # Authorize tests
-  def test_successful_authorize
-    response = @gateway.authorize(@amount_purchase, @credit_card, @option_spr)
-    assert_success response
-    assert_equal 'Approved', response.params['status']
+  def test_successful_capture
```
therufs, line 101: This is a minor nitpick, but I think it's conventional to name tests that call multiple methods after both methods, e.g. `test_successful_authorize_and_capture`

therufs (approved):
Good & thorough work; tests pass! 🎉 

I had a suggestion (that might be multiple suggestions) for some cosmetic changes to test names. Otherwise, 🚢

test/remote/gateways/remote_priority_test.rb
dsmcclain: Thank you for this minor but helpful nitpick 😄 . I will change it.

PR #4395, opened 2022-04-12
Airwallex: Add 3DS Global Support

drkjc:
Add support for Spreedly's 3DS Global Solution. One notable distinction
is that Airwallex only supports three digit 3DS versions. Versions
returned from the MPI are formatted to meet that requirement.

Ex: 2.0 => 2.0.0

CE-2301

Unit:
31 tests, 130 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
25 tests, 56 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -71,6 +71,7 @@
 * Conekta: Fix remote test [javierpedrozaing] #4386
 * NMI: Update post URL [jherreraa] #4380
 * Multiple Gateways: Resolve when/case bug [naashton] #4399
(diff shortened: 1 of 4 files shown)


therufs (approved):
lgtm! 

Currently failing tests, for posterity, are the three remote `mit` tests; since those aren't related to this work, I say 🚢 

EDITED TO ADD don't forget a changelog entry!

PR #4397, opened 2022-04-13
Priority: add `settled` and `voided` to list of successful response statuses

dsmcclain:
Turns out, when you use `replay_id` to recall a previous transaction that has already settled or has already been voided, the response.status is `Settled` or `Voided`, respectively. These statuses were not included in the `success_from` method and we were not testing for these scenarios, so the result would be to incorrectly mark the transaction as `failed`.

Rubocop:

Unit:
5160 tests, 75567 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
27 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


naashton (approved):
Cool beans 👍 

🚢

PR #4405, opened 2022-04-25
Simetrik: Fix integer and float types

rachelkirk:
CE-2407

This PR does some refactoring and updates some value types that were being passed incorrectly. The scrub method has also been updated.

Remote tests:
12 tests, 24 assertions, 0 failures, 8 errors, 0 pendings, 0 omissions, 0 notifications
33.3333% passed

Unit tests:
12 tests, 82 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/simetrik.rb
+++ b/lib/active_merchant/billing/gateways/simetrik.rb
@@ -52,7 +52,6 @@ module ActiveMerchant #:nodoc:
         add_forward_route(post, options)
         add_forward_payload(post, money, payment, options)
         add_stored_credential(post, options)
-
         commit('authorize', post, { token_acquirer: options[:token_acquirer] })
       end
 
@@ -70,7 +69,7 @@ module ActiveMerchant #:nodoc:
             acquire_extra_options: options[:acquire_extra_options] || {}
           }
         }
-        post[:forward_payload][:amount][:vat] = options[:vat] if options[:vat]
+        post[:forward_payload][:amount][:vat] = options[:vat].to_f if options[:vat]
 
         add_forward_route(post, options)
         commit('capture', post, { token_acquirer: options[:token_acquirer] })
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_simetrik_test.rb
```diff
@@ -53,12 +53,10 @@ def setup
         id: rand(100000000000..999999999999).to_s,
         datetime_local_transaction: Time.new.strftime('%Y-%m-%dT%H:%M:%S.%L%:z'),
         description: 'apopsicle',
-        installments: 1,
-        amount: {
-          currency: 'USD',
-          vat: 19
-        }
```
dsmcclain, line 60: big improvement considering we don't want customers to have to pass a rich object for `amount` 👍

lib/active_merchant/billing/gateways/simetrik.rb
```diff
@@ -127,8 +126,8 @@ def supports_scrubbing?
 
       def scrub(transcript)
         transcript.
-          gsub(%r((\"number\\\":\\\")\d+), '\1[FILTERED]').
-          gsub(%r((\"security_code\\\":\\\")\d+), '\1[FILTERED]').
+          gsub(%r(("number\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("security_code\\?":\\?")\d+), '\1[FILTERED]').
           gsub(%r((\"exp_month\\\":\\\")\d+), '\1[FILTERED]').
           gsub(%r((\"exp_year\\\":\\\")\d+), '\1[FILTERED]').
           gsub(%r((\"holder_first_name\\\":\\\")"\w+"), '\1[FILTERED]').
```
dsmcclain, line 133: So, one issue I'm having with this Very Important Method is that it claims to scrub `exp_month`, `exp_year`, `holder_first_name`, and `holder_last_name` from the transcript. However, when I examine the scrubbed transcript from the `def test_scrub` unit test, I can see that only the `number`, and `security_code` values have been scrubbed.

I think we need to:
* determine what are the actual sensitive values we should scrub. I think the exp dates and names are not typically scrubbed values, so maybe we can just remove them?
* improve the `def test_scrub` test so that it checks every value that we claim to be scrubbing

dsmcclain (changes requested):
I am requesting a change relating to the scrubbing (see comment below).

The rest of the changes look good and make sense. I am currently getting all tests passing.

lib/active_merchant/billing/gateways/simetrik.rb
rachelkirk: I took out exp_month, exp_year, holder_first_name, holder_last_name and added Bearer and client_secret. Test has been updated as well.

dsmcclain (approved):
Nice work puzzling through this @rachelkirk. All tests are passing for me.

It would be nice if you update your PR description to indicate that the remote tests are currently passing on this branch.

Also don't forget a changelog entry.

Rubocop:

Unit:
5173 tests, 75674 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
100% passed

PR #4409, opened 2022-04-29
Monei: add cardholderName field

dsmcclain:
Monei's sandbox does not care if this field gets passed or not, but its omission will cause failures in production.

This field can be found[ here in Monei's docs ](https://docs.monei.com/api/#operation/payments_create
)if you expand the `paymentMethod` object and the child `card` object.

CE-2562

Rubocop:

Unit:
5174 tests, 75676 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
25 tests, 63 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


naashton (approved):
🚢

PR #4411, opened 2022-05-02
BlueSnap: parse refund-transaction-id correctly

dsmcclain:
Depending on which endpoint is being used, the gateway's response on a refund will include the transaction id in either the `transaction-id` or the `refund-transaction-id` field of the parsed response.

This change updates the `authorize_from` method to account for the presence of `refund-transaction-id` on select refund transactions.

Currently, there is one remote test that is failing on master: `test_successful_purchase_with_level3_data`. When I inspect the response for this test, I do not see any of [the response fields that BlueSnap's documentation lists for level3data](https://developers.bluesnap.com/v8976-JSON/docs/level3data). Instead I see ` "transaction-processed-with-l3d-supported-acquirer"=>"false"`. My best guess here is that this indicates a setting that needs to be changed on our sandbox account. I do not have the ability to do this through BlueSnap's developer portal, so I am just going to leave this test as failing because attempting to contact the gateway to get this changed is outside the scope of this current task.

CE-2522

Rubocop:

Unit:
5174 tests, 75678 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
Loaded suite test/remote/gateways/remote_blue_snap_test
50 tests, 169 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
98% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -512,7 +512,9 @@ module ActiveMerchant
       end
 
       def authorization_from(action, parsed_response, payment_method_details)
-        action == :store ? vaulted_shopper_id(parsed_response, payment_method_details) : parsed_response['transaction-id']
+        return vaulted_shopper_id(parsed_response, payment_method_details) if action == :store
+
+        parsed_response['refund-transaction-id'] || parsed_response['transaction-id']
       end
 
       def vaulted_shopper_id(parsed_response, payment_method_details)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -512,7 +512,9 @@ def fraud_codes_from(response)
       end
 
       def authorization_from(action, parsed_response, payment_method_details)
-        action == :store ? vaulted_shopper_id(parsed_response, payment_method_details) : parsed_response['transaction-id']
+        return vaulted_shopper_id(parsed_response, payment_method_details) if action == :store
```
drkjc, line 515: Nice refactor

test/unit/gateways/blue_snap_test.rb
```diff
@@ -1093,6 +1094,32 @@ def failed_refund_response
     MockResponse.failed(body, 400)
   end
 
+  def successful_refund_without_merchant_transaction_id_response
+    MockResponse.succeeded <<-XML
```
drkjc, line 1098: Today I learned about MockResponse.

drkjc (approved):
🚢 Approved! Looks great. Nice work.

PR #4413, opened 2022-05-03
Vantiv: Read saleResponse duplicate attribute

mashton:
The LitleGateway parse method reads elements and their text values, but does not read any of the elements attributes. Here we read one very specific attribute (probably not worth generalizing just yet). Here we're interested in a Litle saleResponse that may or may not have the attribute `duplicate='true'`. All Litle saleResponses will include duplicate: true/false.

Note on testing: Remote test for this feature is missing. It looks like the `testvantivcnp` api does not return the duplicate attribute the same way the production api does. Additionally, something seems to have changed in the `testvantivcnp` api such that existing remote tests fail. PR coming soon with fixes for existing tests.

CE-2506

Local:
5175 tests, 75684 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
See note.

Linting:

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -532,8 +532,10 @@ module ActiveMerchant #:nodoc:
 
       def parse(kind, xml)
         parsed = {}
-
         doc = Nokogiri::XML(xml).remove_namespaces!
+
+        parsed['duplicate'] = doc.at_xpath('//saleResponse').try(:[], 'duplicate') == 'true' if kind == :sale
+
         doc.xpath("//litleOnlineResponse/#{kind}Response/*").each do |node|
           if node.elements.empty?
             parsed[node.name.to_sym] = node.text
--- a/test/unit/gateways/litle_test.rb
+++ b/test/unit/gateways/litle_test.rb
@@ -112,6 +112,22 @@ class LitleTest < Test::Unit::TestCase
     assert response.test?
   end
 
+  def test_sale_response_duplicate_attribute
+    dup_response = stub_comms do
+      @gateway.purchase(@amount, @credit_card)
+    end.respond_with(duplicate_purchase_response)
+
+    assert_success dup_response
+    assert_true dup_response.params['duplicate']
+
+    non_dup_response = stub_comms do
+      @gateway.purchase(@amount, @credit_card)
+    end.respond_with(successful_purchase_response)
+
+    assert_success non_dup_response
+    assert_false non_dup_response.params['duplicate']
+  end
+
   def test_failed_purchase
     response = stub_comms do
       @gateway.purchase(@amount, @credit_card)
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -532,8 +532,10 @@ def format_exp_date(month, year)
 
       def parse(kind, xml)
         parsed = {}
-
         doc = Nokogiri::XML(xml).remove_namespaces!
+
+        parsed['duplicate'] = doc.at_xpath('//saleResponse')['duplicate'] == 'true' if kind == :sale
```
dsmcclain, line 537: One worry I have here is that if `doc.at_xpath('//saleResponse')` returns `nil` then calling `['duplicate']` on it will cause ruby to say `NoMethodError: undefined method '[]' for nil:NilClass` and then we are returning a sad 500 to the customer.

It looks like we make assumptions like this elsewhere in the `parse` method, and it could very well be that `saleResponse` is ALWAYS there when `kind == :sale`, but I would still feel better if we protected against a `nil` here. Something like:

```
parsed['duplicate'] = doc.at_xpath('//saleResponse').try(:[], 'duplicate') == 'true' if kind == :sale
```

test/unit/gateways/litle_test.rb
```diff
@@ -727,6 +743,25 @@ def successful_purchase_response
     )
   end
 
+  def duplicate_purchase_response
```
dsmcclain, line 746: nicely done

dsmcclain (changes requested):
One request for change that is probably unnecessary but more of a `best practices` sort of thing.

dsmcclain (changes requested):
Now I am getting a failure on `test_unsuccessful_xml_schema_validation` in the litle unit tests. Looks like we still need the `if kind == :sale` check.

mashton:
I sure wish there was a better way to check for xml validation errors than `parse.empty?`, which is what caught us up on this last one. It seems like a bug waiting to happen. But it looks like a Litle/Vantiv overhaul is overdue anyway, since their API version has just about doubled since this was put together.

dsmcclain:
> I sure wish there was a better way to check for xml validation errors than `parse.empty?`, which is what caught us up on this last one. It seems like a bug waiting to happen. But it looks like a Litle/Vantiv overhaul is overdue anyway, since their API version has just about doubled since this was put together.

Agree 100%. The `parse` method in this gateway is unnecessarily brittle.

dsmcclain (approved):
When I run `bundle exec rake test:local` I get a linting error now.

Once that is fixed, rebase with master and squash down to one commit. Then add an entry to the changelog (at the bottom of the `== HEAD` section), and at that point you are good to merge.

I am approving so that you don't have to tag me for re-review on a linting fix, but don't hesitate to reach out if you have questions about the merge process 🙂 _Please don't forget to fix the linting error before merging_

_For posterity: I am getting 13 failing remote tests on this branch, same as master_

PR #4414, opened 2022-05-09
BlueSnap: Add stored credential

ajawadmirza:
Added `stored_credential` support by adding `transaction-initiator` and
`original-network-transaction-id` in the request parameters for blue
snap gateway implementation.

CE-2536

Rubocop:

Unit:
5180 tests, 75721 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
52 tests, 177 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
98.0769% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/blue_snap.rb
+++ b/lib/active_merchant/billing/gateways/blue_snap.rb
@@ -68,6 +68,8 @@ module ActiveMerchant
         'business_savings' => 'CORPORATE_SAVINGS'
       }
 
+      SHOPPER_INITIATOR = %w(CUSTOMER CARDHOLDER)
+
       STATE_CODE_COUNTRIES = %w(US CA)
 
       def initialize(options = {})
@@ -179,6 +181,7 @@ module ActiveMerchant
         doc.send('store-card', options[:store_card] || false)
         add_amount(doc, money, options)
         add_fraud_info(doc, payment_method, options)
+        add_stored_credentials(doc, options)
 
         if payment_method.is_a?(String)
           doc.send('vaulted-shopper-id', payment_method)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/blue_snap.rb
```diff
@@ -190,6 +193,19 @@ def add_auth_purchase(doc, money, payment_method, options)
         end
       end
 
+      def add_stored_credentials(doc, options)
+        return unless stored_credential = options[:stored_credential]
+
+        initiator = stored_credential[:initiator]&.upcase
+        initiator = 'SHOPPER' if SHOPPER_INITIATOR.include?(initiator)
+        doc.send('transaction-initiator', initiator) if stored_credential[:initiator]
```
rachelkirk, line 201: Hey @ajawadmirza I see that you have tests for MIT transactions with the initiator set to `merchant`, but the only initiator option indicated here is `SHOPPER` for CIT transactions. Is calling out both in the code unnecessary? Just wondering for my own knowledge. Your tests are passing and everything else looks good so I'll approve and merge in the morning after reading your response.

rachelkirk (approved):
Looks good, just had one question in comments.

lib/active_merchant/billing/gateways/blue_snap.rb
ajawadmirza: Hi @rachelkirk ,

From the `stored_credential` object we receive three type of initiator, i.e. merchant, customer, cardholder. On the gateway field mapping merchant is passed as it is passed in the stored_credential object but for customer and cardholder it passes with field value `SHOPPER` thats why we pass in merchant as it is and for customer and cardholder we convert them to the valid required value.

PR #4415, opened 2022-05-09
PS - Shift4 - Active Merchant Build

ali-hassan:
PS - Shift4 - Active Merchant Build

Unit Test:
5173 tests, 75672 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
64.86 tests/s, 948.72 assertions/s
Running RuboCop...
Inspecting 742 files
Remote Test:
13 tests, 29 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4.rb
+++ b/lib/active_merchant/billing/gateways/shift4.rb
@@ -1,0 +1,300 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class Shift4Gateway < Gateway
+      self.test_url = 'https://utgapi.shift4test.com/api/rest/v1/'
+      self.live_url = 'https://utg.shift4api.net/api/rest/v1/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
+      self.homepage_url = 'https://shift4.com'
+      self.display_name = 'Shift4'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        'incorrect_number' => STANDARD_ERROR_CODE[:incorrect_number],
+        'invalid_number' => STANDARD_ERROR_CODE[:invalid_number],
+        'invalid_expiry_month' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        'invalid_expiry_year' => STANDARD_ERROR_CODE[:invalid_expiry_date],
+        'invalid_cvc' => STANDARD_ERROR_CODE[:invalid_cvc],
+        'expired_card' => STANDARD_ERROR_CODE[:expired_card],
+        'insufficient_funds' => STANDARD_ERROR_CODE[:card_declined],
+        'incorrect_cvc' => STANDARD_ERROR_CODE[:incorrect_cvc],
+        'incorrect_zip' => STANDARD_ERROR_CODE[:incorrect_zip],
+        'card_declined' => STANDARD_ERROR_CODE[:card_declined],
+        'processing_error' => STANDARD_ERROR_CODE[:processing_error],
+        'lost_or_stolen' => STANDARD_ERROR_CODE[:card_declined],
+        'suspected_fraud' => STANDARD_ERROR_CODE[:card_declined],
+        'expired_token' => STANDARD_ERROR_CODE[:card_declined]
+      }
+
+      def initialize(options = {})
+        requires!(options, :client_guid, :auth_token)
+        @client_guid = options[:client_guid]
+        @auth_token = options[:auth_token]
+        super
+        @access_token = setup_access_token
+      end
+
+      def purchase(money, authorization, options = {})
+        post = {}
+        add_datetime(post, options)
+        add_invoice(post, money, options)
+        add_clerk(post, options)
+        add_transaction(post, options)
+        add_card(post, authorization, options)
+
+        commit('sale', post, options)
+      end
+
+      def authorize(money, card, options = {})
+        post = {}
+        add_datetime(post, options)
+        add_invoice(post, money, options)
+        add_clerk(post, options)
+        add_transaction(post, options)
+        add_card(post, card, options)
+        add_customer(post, options)
+
+        commit('authorization', post, options)
+      end
+
+      def capture(money, authorization, options = {})
+        post = {}
+        add_datetime(post, options)
+        add_invoice(post, money, options)
+        add_clerk(post, options)
+        add_transaction(post, options)
+        add_card(post, get_card_token(authorization), options)
+
+        commit('capture', post, options)
+      end
+
+      def refund(money, authorization, options = {})
+        post = {}
+        add_datetime(post, options)
+        add_invoice(post, money, options)
+        add_clerk(post, options)
+        add_transaction(post, options)
+        add_customer(post, options)
+        add_card(post, get_card_token(authorization), options)
+
+        commit('refund', post, options)
+      end
+
+      def void(authorization, options = {})
+        options[:invoice] = get_invoice(authorization)
+        commit('invoice', {}, options)
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r(("Number\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("expirationDate\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("FirstName\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("LastName\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
+      end
+
+      private
+
+      def add_credentials(post, options)
+        post[:dateTime] = current_date_time
+        post[:credential] = {}
+        post[:credential][:clientGuid] = @client_guid
+        post[:credential][:authToken] = @auth_token
+      end
+
+      def setup_access_token
+        post = {}
+        add_credentials(post, options)
+
+        response = commit('accesstoken', post, request_headers(options))
+        response.params['result'][0]['credential']['accessToken']
+      end
+
+      def add_clerk(post, options)
+        post[:clerk] = {}
+        post[:clerk][:numericId] = options[:clerk_id]
+      end
+
+      def add_invoice(post, money, options)
+        post[:amount] = {}
+        post[:amount][:total] = money.to_f
+        post[:amount][:tax] = options[:tax].to_f || 0.0
+      end
+
+      def add_datetime(post, options)
+        post[:dateTime] = options[:date_time] || current_date_time
+      end
+
+      def add_transaction(post, options)
+        add_purchase_card(post, options)
+        add_override_bus_date(post, options)
+        add_amex(post, options)
+        add_auto(post, options)
+        add_card_on_file(post, options)
+      end
+
+      def add_card(post, credit_card, options)
+        post[:card] = {}
+        if credit_card.is_a?(CreditCard)
+          post[:card][:expirationDate] = "#{format(credit_card.month, :two_digits)}#{format(credit_card.year, :two_digits)}"
+          post[:card][:number] = credit_card.number
+          post[:card][:entryMode] = options[:entry_mode]
+          post[:card][:present] = options[:present]
+          post[:card][:securityCode] = {}
+          post[:card][:securityCode][:indicator] = 1
+          post[:card][:securityCode][:value] = credit_card.verification_value
+        else
+          post[:card] = {} if post[:card].nil?
+          post[:card][:token] = {}
+          post[:card][:token][:value] = credit_card
+        end
+      end
+
+      def add_customer(post, options)
+        if address = options[:billing_address]
+          post[:customer] = {}
+          post[:customer][:addressLine1] = address[:address1] if address[:address1]
+          name = address[:name].split(' ')
+          post[:customer][:firstName] = name[0]
+          post[:customer][:lastName] = name[1]
+          post[:customer][:postalCode] = address[:postal_code]
+        end
+      end
+
+      def add_purchase_card(post, options)
+        post[:transaction] = {}
+        post[:transaction][:invoice] = options[:invoice]
+        post[:transaction][:notes] = options[:notes] if options[:notes].present?
+      end
+
+      def add_override_bus_date(post, options)
+        post[:overrideBusDate] = options[:override_bus_date] if options[:override_bus_date].present?
+      end
+
+      def add_amex(post, options)
+        if options[:amex].present?
+          post[:amex] = {}
+          post[:amex][:propertyCode] = options[:property_code]
+        end
+      end
+
+      def add_auto(post, options)
+        if options[:auto].present?
+          post[:auto] = {}
+          post[:auto][:estimatedDays] = options[:estimated_days]
+        end
+      end
+
+      def add_card_on_file(post, options)
+        if options[:stored_credential].present?
+          post[:cardOnFile] = {}
+          post[:cardOnFile][:indicator] = options[:stored_credential][:indicator] if options[:stored_credential][:indicator]
+          if usage_indicator = options[:stored_credential][:usage_indicator] == 'initial' ? '01' : '02'
+            post[:cardOnFile][:usageIndicator] = usage_indicator
+          end
+          if scheduled_indicator = options[:stored_credential][:scheduled_indicator] == 'scheduled' ? '' : '02'
+            post[:cardOnFile][:scheduledIndicator] = scheduled_indicator
+          end
+          post[:cardOnFile][:transactionId] = options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id]
+        end
+      end
+
+      def commit(action, parameters, option)
+        url_postfix = action == 'accesstoken' ? 'credentials' : 'transactions'
+        url = (test? ? "#{test_url}#{url_postfix}/#{action}" : "#{live_url}#{url_postfix}/#{action}")
+        if action == 'invoice'
+          response = parse(ssl_request(:delete, url, parameters.to_json, request_headers(option)))
+        else
+          response = parse(ssl_post(url, parameters.to_json, request_headers(option)))
+        end
+
+        Response.new(
+          success_from(response),
+          message_from(response),
+          response,
+          authorization: authorization_from(response, action),
+          test: test?,
+          error_code: error_code_from(response)
+        )
+      end
+
+      def handle_response(response)
+        case response.code.to_i
+        when 200...300, 400, 401
+          response.body
+        else
+          raise ResponseError.new(response)
+        end
+      end
+
+      def parse(body)
+        return {} if body == ''
+
+        JSON.parse(body)
+      end
+
+      def message_from(response)
+        success_from(response) ? 'Transaction successful' : error(response)['longText']
+      end
+
+      def error_code_from(response)
+        return unless success_from(response)
+
+        STANDARD_ERROR_CODE_MAPPING[response['primaryCode']]
+      end
+
+      def authorization_from(response, action)
+        return unless success_from(response)
+
+        "#{response.dig('result', 0, 'card', 'token', 'value')}|#{response.dig('result', 0, 'transaction', 'invoice')}"
+      end
+
+      def get_card_token(authorization)
+        authorization.is_a?(CreditCard) ? authorization : authorization.split('|')[0]
+      end
+
+      def get_invoice(authorization)
+        authorization.is_a?(CreditCard) ? authorization : authorization.split('|')[1]
+      end
+
+      def request_headers(options)
+        headers = {
+          'Content-Type' => 'application/x-www-form-urlencoded'
+        }
+        headers['CompanyName'] = options[:company_name]
+        headers['AccessToken'] = @access_token
+        headers['Invoice'] = options[:invoice] if options[:invoice].present?
+        headers
+      end
+
+      def success_from(response)
+        error(response).nil?
+      end
+
+      def error(response)
+        response_result(response)['error']
+      end
+
+      def response_result(response)
+        response['result'][0]
+      end
+
+      def current_date_time
+        DateTime.now.strftime('%Y-%m-%dT%H:%M:%S.%N+%H:%M')
+      end
+    end
+  end
+end
--- a/test/remote/gateways/remote_shift4_test.rb
+++ b/test/remote/gateways/remote_shift4_test.rb
@@ -1,0 +1,166 @@
+require 'test_helper'
+
+class RemoteShift4Test < Test::Unit::TestCase
+  def setup
+    @gateway = Shift4Gateway.new(fixtures(:shift4))
+
+    @amount = 5
+    @invoice = '3333333309'
+    @credit_card = credit_card('4000100011112224')
+    @declined_card = credit_card('4000300011112220')
+    @declined_card.number = '400030001111220'
+    @options = {
+      order_id: '1',
+      company_name: 'Spreedly',
+      billing_address: address,
+      description: 'Store Purchase',
+      "tax": 1,
+      "tip": '',
+      "total": 5.00,
+      clerk_id: 24,
+      invoice: rand(4).to_s,
+      security_code: {
+        indicator: '1',
+        value: '4444'
+      }
+    }
+    @refund_options = @options.except(:invoice).merge({
+      invoice: @invoice,
+      notes: 'Transaction notes are added here',
+      billing_address: address,
+      stored_credential: {
+        indicator: '01',
+        usage_indicator: '01',
+        scheduled_indicator: '01',
+        transaction_id: 'yxx'
+      },
+      token: {
+        serial_number: '',
+        value: '1111g66gw3ryke06'
+      }
+    })
+    @capture_params = @options.except(:invoice).merge({
+      transaction: {
+        invoice: @invoice
+      },
+      card: {}
+    })
+    @purchase_params = {
+      company_name: 'Spreedly',
+      tax: 0,
+      total: 219,
+      clerk_id: 16,
+      transaction: {
+        invoice: '4666309473',
+        notes: 'Test payment'
+      },
+      card: {
+        expiration_date: '0825',
+        present: 'N',
+        indicator: '1',
+        value: '3333'
+      }
+    }
+    @token = '8042677003331111'
+    @void_options = { company_name: 'Spreedly', invoice: @invoice }
+  end
+
+  def test_successful_authorize
+    response = @gateway.authorize(@amount, @credit_card, @options)
+
+    assert_success response
+    assert_equal response.message, 'Transaction successful'
+    assert_equal @options[:total], response_result(response)['amount']['total']
+  end
+
+  def test_successful_capture
+    authorize_res = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success authorize_res
+    response = @gateway.capture(@amount, authorize_res.authorization, @capture_params)
+
+    assert_success response
+    assert_equal response.message, 'Transaction successful'
+    assert_equal @options[:total], response_result(response)['amount']['total']
+    assert response_result(response)['transaction']['invoice'].present?
+  end
+
+  def test_successful_purchase
+    response = @gateway.purchase(@amount, @credit_card, @purchase_params)
+    assert_success response
+    assert_equal @options[:total], response_result(response)['amount']['total']
+  end
+
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.authorize(@amount, @credit_card, @options)
+    end
+    transcript = @gateway.scrub(transcript)
+    assert_scrubbed(@credit_card.number, transcript)
+    assert_scrubbed("0#{@credit_card.month}#{@credit_card.year.to_s[2..4]}", transcript)
+    assert_scrubbed(@options[:billing_address][:name], transcript)
+  end
+
+  def test_failed_purchase
+    auth_token = '8038483489222221'
+    @purchase_params[:card][:expiration_date] = '085'
+    response = @gateway.purchase(@amount, auth_token, @purchase_params)
+
+    assert_failure response
+    assert response_result(response)['error']['longText'].include?('No GTV PAN returned from host')
+  end
+
+  def test_failed_authorize
+    response = @gateway.authorize(@amount, @declined_card, @options)
+
+    assert_failure response
+    assert response_result(response)['error']['longText'].include?('Card  for Merchant Id')
+  end
+
+  def test_failed_capture
+    auth_token = '8038483489222221'
+    @capture_params[:card][:expiration_date] = '085'
+    response = @gateway.capture(@amount, auth_token, @capture_params)
+
+    assert_failure response
+    assert response_result(response)['error']['longText'].include?('No GTV PAN returned from host')
+  end
+
+  def test_failed_refund
+    response = @gateway.refund(nil, @credit_card, @refund_options.except(:access_token))
+
+    assert_failure response
+    assert response_result(response)['error']['longText'].include?('Secondary amounts cannot exceed the total amount')
+  end
+
+  def test_successful_refund
+    res = @gateway.purchase(@amount, @credit_card, @purchase_params)
+    assert_success res
+    response = @gateway.refund(@amount, res.authorization, @refund_options)
+
+    assert_success response
+    assert_equal @refund_options[:total], response_result(response)['amount']['total']
+    assert_equal @refund_options[:invoice], response_result(response)['transaction']['invoice']
+  end
+
+  def test_successful_void
+    authorize_res = @gateway.authorize(@amount, @credit_card, @options)
+    assert response = @gateway.void(authorize_res.authorization, @void_options)
+
+    assert_success response
+    assert_equal @refund_options[:total], response_result(response)['amount']['total']
+    assert_equal @void_options[:invoice], response_result(response)['transaction']['invoice']
+  end
+
+  def test_failed_void
+    response = @gateway.void('', @void_options.except(:invoice))
+    assert_failure response
+    assert response_result(response)['error']['longText'].include?('Invoice Not Found')
+    assert_equal response_result(response)['error']['primaryCode'], 9815
+  end
+
+  private
+
+  def response_result(response)
+    response.params['result'][0]
+  end
+end
--- a/test/unit/gateways/shift4_test.rb
+++ b/test/unit/gateways/shift4_test.rb
@@ -1,0 +1,718 @@
+require 'test_helper'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class Shift4Gateway
+      def setup_access_token
+        '12345678'
+      end
+    end
+  end
+end
+
+class Shift4Test < Test::Unit::TestCase
+  include CommStub
+  def setup
+    @gateway = Shift4Gateway.new(client_guid: '123456', auth_token: 'abcder123')
+    @credit_card = credit_card
+    @amount = 100
+
+    @options = {
+      order_id: '1',
+      company_name: 'Spreedly',
+      billing_address: address,
+      description: 'Store Purchase',
+      tax: 0,
+      total: 5,
+      clerk_id: 24,
+      invoice: '3333333309',
+      purchase_card: {
+        customer_reference: '457',
+        destination_postal_code: '89123',
+        product_descriptors: %w[
+          Potential
+          Wrong
+        ]
+      },
+      present: 'N'
+    }
+    @refund_options = {
+      date_time: '2022-05-02T09:18:26.000-07:00',
+      billing_address: address,
+      total: 5,
+      clerk_id: 24,
+      invoice: rand(4),
+      purchase_card: {
+        customer_reference: '457',
+        destination_postal_code: '89123',
+        product_descriptors: 'Test'
+      },
+      notes: 'Transaction notes are added here',
+      present: 'N'
+    }
+    @invalid_auth_token = '1111g66gw3ryke00'
+    @invoid_options = { company_name: 'Spreedly', invoice: '111' }
+  end
+
+  def test_successful_capture
+    auth_token = '1111g66gw3ryke06'
+    response = stub_comms do
+      @gateway.capture(@amount, auth_token, @options)
+    end.respond_with(successful_capture_response)
+
+    assert response.success?
+    assert_equal response.message, 'Transaction successful'
+    assert_equal @options[:total], response_result(response)['amount']['total']
+    assert_equal response_result(response)['card']['token']['value'].present?, true
+  end
+
+  def test_successful_authorize
+    response = stub_comms do
+      @gateway.authorize(@amount, @credit_card, @options)
+    end.respond_with(successful_authorize_response)
+
+    assert response.success?
+    assert_equal response.message, 'Transaction successful'
+    assert_equal @options[:total], response_result(response)['amount']['total']
+    assert_equal response_result(response)['card']['token']['value'].present?, true
+  end
+
+  def test_successful_purchase
+    auth_token = '1111g66gw3ryke06'
+    response = stub_comms do
+      @gateway.purchase(@amount, auth_token, @options)
+    end.respond_with(successful_purchase_response)
+
+    assert response.success?
+    assert_equal response.message, 'Transaction successful'
+    assert_equal @options[:total], response_result(response)['amount']['total']
+    assert_equal response_result(response)['card']['token']['value'].present?, true
+  end
+
+  def test_successful_refund
+    auth_token = '1111g66gw3ryke06'
+    response = stub_comms do
+      @gateway.refund(@amount, auth_token, @refund_options.merge!(invoice: '4666309473'))
+    end.respond_with(successful_refund_response)
+
+    assert response.success?
+    assert_equal response.message, 'Transaction successful'
+  end
+
+  def test_successful_void
+    @gateway.expects(:ssl_request).returns(successful_void_response)
+    response = @gateway.void('123')
+
+    assert response.success?
+    assert_equal response.message, 'Transaction successful'
+  end
+
+  def test_failed_purchase
+    @invalid_auth_token = '1111g66gw3ryke00'
+    @gateway.expects(:ssl_request).returns(failed_purchase_response)
+
+    response = @gateway.purchase(@amount, @invalid_auth_token, @options)
+    assert_failure response
+    assert_nil response.authorization
+  end
+
+  def test_failed_authorize
+    @gateway.expects(:ssl_request).returns(failed_authorize_response)
+
+    response = @gateway.authorize(@amount, @credit_card, @options)
+    assert_failure response
+    assert_nil response.authorization
+    assert response.test?
+  end
+
+  def test_failed_capture
+    @invalid_auth_token = '1111g66gw3ryke00'
+    @gateway.expects(:ssl_request).returns(failed_capture_response)
+
+    response = @gateway.capture(@amount, @invalid_auth_token, @options)
+    assert_failure response
+    assert_nil response.authorization
+    assert response.test?
+  end
+
+  def test_failed_refund
+    @invalid_auth_token = '1111g66gw3ryke00'
+    @gateway.expects(:ssl_request).returns(failed_refund_response)
+
+    response = @gateway.refund(@amount, @invalid_auth_token, @options)
+    assert_failure response
+    assert_nil response.authorization
+    assert response.test?
+  end
+
+  def test_failed_void
+    @gateway.expects(:ssl_request).returns(failed_void_response)
+
+    response = @gateway.void('', @invoid_options)
+    assert_failure response
+    assert_nil response.authorization
+    assert response.test?
+  end
+
+  def test_support_scrub
+    assert @gateway.supports_scrubbing?
+  end
+
+  def test_scrub
+    assert @gateway.supports_scrubbing?
+    assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
+  end
+
+  private
+
+  def response_result(response)
+    response.params['result'][0]
+  end
+
+  def pre_scrubbed
+    <<-PRE_SCRUBBED
+      opening connection to utgapi.shift4test.com:443...
+      opened
+      starting SSL for utgapi.shift4test.com:443...
+      SSL established, protocol: TLSv1.2, cipher: ECDHE-RSA-AES128-GCM-SHA256
+      <- "POST /api/rest/v1/transactions/authorization HTTP/1.1\r\nContent-Type: application/x-www-form-urlencoded\r\nCompanyname: Spreedly\r\nAccesstoken: 4902FAD2-E88F-4A8D-98C2-EED2A73DBBE2\r\nInvoice: 1\r\nConnection: close\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nAccept: */*\r\nUser-Agent: Ruby\r\nHost: utgapi.shift4test.com\r\nContent-Length: 498\r\n\r\n"
+      <- "{\"dateTime\":\"2022-06-09T14:03:36.413505000+14:03\",\"amount\":{\"total\":5.0,\"tax\":1.0},\"clerk\":{\"numericId\":24},\"transaction\":{\"invoice\":\"1\",\"purchaseCard\":{\"customerReference\":\"457\",\"destinationPostalCode\":\"89123\",\"productDescriptors\":[\"Potential\",\"Wrong\"]}},\"card\":{\"expirationDate\":\"0923\",\"number\":\"4000100011112224\",\"entryMode\":null,\"present\":null,\"securityCode\":{\"indicator\":\"1\",\"value\":\"4444\"}},\"customer\":{\"addressLine1\":\"89 Main Street\",\"firstName\":\"XYZ\",\"lastName\":\"RON\",\"postalCode\":\"89000\"}}"
+      -> "HTTP/1.1 200 OK\r\n"
+      -> "Connection: close\r\n"
+      -> "Content-Type: text/json; charset=ISO-8859-1\r\n"
+      -> "Content-Length: 1074\r\n"
+      -> "Date: Thu, 09 Jun 2022 09:03:40 GMT\r\n"
+      -> "Pragma: no-cache\r\n"
+      -> "X-Frame-Options: deny\r\n"
+      -> "Content-Security-Policy: default-src 'none';base-uri 'none';frame-ancestors 'none';object-src 'none';sandbox;\r\n"
+      -> "Strict-Transport-Security: max-age=31536000; includeSubDomains\r\n"
+      -> "X-XSS-Protection: 1; mode=block\r\n"
+      -> "X-Content-Type-Options: nosniff\r\n"
+      -> "X-Permitted-Cross-Domain-Policies: none\r\n"
+      -> "Referrer-Policy: no-referrer\r\n"
+      -> "X-Powered-By: Electricity\r\n"
+      -> "Expires: 0\r\n"
+      -> "Cache-Control: private, no-cache, no-store, max-age=0, no-transform\r\n"
+      -> "Server: DatasnapHTTPService/2011\r\n"
+      -> "\r\n"
+      reading 1074 bytes...
+      -> ""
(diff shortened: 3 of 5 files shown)


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,245 @@
...
+      end
+
+      private
+
+      def add_clerk(post, money, options)
+        post[:clerk] = {}
+        post[:clerk][:numericId] = options[:clerk][:numeric_id]
+      end
+
+      def add_customer_data(post, options); end
```
therufs, line 97: It looks like these empty methods aren't used anywhere, so we should remove them

ali-hassan: @therufs  Done (y)

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+      self.display_name = 'Shift4'
+
+      STANDARD_ERROR_CODE_MAPPING = {}
+
+      def initialize(options = {})
+        requires!(options, :client_guid, :auth_token)
+        super
+      end
+
+      def access_token(options = {})
```
therufs, line 21: Is this an access token that only ever needs to be generated once per client? If so, I think it would make sense to handle it as a credential. 

If it needs to be refreshed, that should take place as part of the `commit` method, rather than requiring users to call this method separately and then pass the result in to other transaction calls.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+        post = {}
+        add_credentials(post, options)
+        commit('accesstoken', post, request_headers(options))
+      end
+
+      def purchase(money, options = {})
```
therufs, line 27: Method definitions for transactions need to have a payment_method in the signature. See examples here: https://github.com/activemerchant/active_merchant/wiki/Contributing#generating-the-gateway-skeleton

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      private
+
+      def add_clerk(post, money, options)
+        post[:clerk] = {}
+        post[:clerk][:numericId] = options[:clerk][:numeric_id]
```
therufs, line 94: We should avoid nesting options when there's not a compelling reason to do so; this could be something like `options[:clerk_id]`

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+
+      def add_datetime(post, options)
+        post[:dateTime] = options[:dateTime] || DateTime.now.strftime('%Y-%m-%dT%H:%M:%S.%N+%H:%M')
+      end
+
+      def add_transaction(post, options)
+        post[:transaction] = {}
+        post[:transaction][:invoice] = options[:transaction][:invoice]
+        post[:transaction][:notes] = options[:transaction][:notes] if options[:transaction][:notes].present?
+        if options[:transaction][:purchase_card].present?
```
therufs, line 122: I think in the case of `purchase_card` it's fine to have the child options, but we should do away with the `transaction` nesting.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+          post[:transaction][:purchaseCard] = {}
+          post[:transaction][:purchaseCard][:customerReference] = options[:transaction][:purchase_card][:customer_reference]
+          post[:transaction][:purchaseCard][:destinationPostalCode] = options[:transaction][:purchase_card][:destination_postal_code]
+          post[:transaction][:purchaseCard][:productDescriptors] = options[:transaction][:purchase_card][:product_descriptors]
+        end
+        post[:overrideBusDate] = options[:override_bus_date] if options[:override_bus_date].present?
+        if post[:amex].present?
```
therufs, line 129: This `if` block and the five that follow it look like they check the `post` hash for options that never get set -- I'm not sure what's going on here?

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+        if options[:p2pe].present?
+          post[:p2pe] = {}
+          post[:p2pe][:format] = options[:p2pe][:format]
+          post[:p2pe][:ksn] = options[:p2pe][:ksn]
+        end
+      end
+
+      def add_card(post, options)
+        post[:card] = {}
+        if options[:card][:token].present?
```
therufs, line 184: This method should be refactored to use the payment_method object being passed in to the transaction method.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+            post[:card][:securityCode][:indicator] = options[:card][:security_code][:indicator]
+            post[:card][:securityCode][:value] = options[:card][:security_code][:value]
+          end
+        end
+      end
+
+      def add_customer(post, options)
+        if options[:customer].present?
+          post[:customer] = {}
+          post[:customer][:addressLine1] = options[:customer][:address_line1]
```
therufs, line 203: We should use the existing address hash for this: https://github.com/activemerchant/active_merchant/wiki/Contributing#address-hash

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+          post[:customer][:lastName] = options[:customer][:last_name]
+          post[:customer][:postalCode] = options[:customer][:postal_code]
+        end
+      end
+
+      def commit(action, parameters, headers)
+        url_postfix = action == 'accesstoken' ? 'credentials' : 'transactions'
+        url = (test? ? "#{test_url}#{url_postfix}/#{action}" : "#{live_url}#{url_postfix}/#{action}")
+        url = URI(url)
+        https = Net::HTTP.new(url.host, url.port)
```
therufs, line 214: We should be able to use the existing `ssl_post` and `ssl_request` methods here: https://github.com/activemerchant/active_merchant/blob/a9c588f90c4c1d59f7b0f5e1dc40b9ca12cb7fbf/lib/active_merchant/posts_data.rb#L39

lib/active_merchant/billing/gateways/shift4.rb
therufs: Further information on payment methods: https://github.com/activemerchant/active_merchant/wiki/Patterns-&-Standards#payment_method

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,241 @@
...
+        https.use_ssl = true
+        if action == 'invoice'
+          request = Net::HTTP::Delete.new(url, headers)
+        else
+          request = Net::HTTP::Post.new(url, headers)
+          request.body = JSON.dump(parameters)
+        end
+        response = https.request(request)
+        JSON.parse response.read_body
```
therufs, line 223: This method should return a fully populated instance of the Response class: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/response.rb. 

The implementation should look something like this: https://github.com/activemerchant/active_merchant/blob/master/generators/gateway/templates/gateway.rb#L89, with the success_from, message_from, and authorization_from methods extracting the relevant information from the raw response.

therufs (changes requested):
Hey @ali-hassan -- 

I commented on some things to look at in the gateway implementation. Right now, the tests are all passing, but they're behaving in some unexpected ways; the unit tests are making remote gateway calls, and the remote tests aren't returning any output when run in debug mode. I think both of these are due to the `commit` method not having been implemented with `ssl_request` or any of its fellows. 

Thanks for your work so far on this (and your patience!)

lib/active_merchant/billing/gateways/shift4.rb
ali-hassan: @therufs  Done.

lib/active_merchant/billing/gateways/shift4.rb
ali-hassan: Done

lib/active_merchant/billing/gateways/shift4.rb
ali-hassan: @therufs Can we discuss in call?

lib/active_merchant/billing/gateways/shift4.rb
ali-hassan: Done

lib/active_merchant/billing/gateways/shift4.rb
ali-hassan: Done

lib/active_merchant/billing/gateways/shift4.rb
ali-hassan: Done

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,277 @@
...
+          post[:customer][:addressLine1] = address[:address_line1] if address[:address_line1]
+          post[:customer][:firstName] = address[:first_name]
+          post[:customer][:lastName] = address[:last_name]
+          post[:customer][:postalCode] = address[:postal_code]
+        end
+      end
+
+      def commit(action, parameters, headers)
+        url_postfix = action == 'accesstoken' ? 'credentials' : 'transactions'
+        url = (test? ? "#{test_url}#{url_postfix}/#{action}" : "#{live_url}#{url_postfix}/#{action}")
```
ajawadmirza, line 230: I have updated your commit method as per the standards. please rectify requests for failing test cases (test_failed_authorize, test_failed_capture etc.)

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,277 @@
...
+        success_from(response) ? 'Transaction successful' : response['longText']
+      end
+
+      def error_code_from(response)
+        return unless success_from(response)
+
+        STANDARD_ERROR_CODE_MAPPING[response['primaryCode']]
+      end
+
+      def authorization_from(response, action)
```
ajawadmirza, line 258: update this method so that it fetches authorization token from authorize/capture to pass them on capture,void/refund respectively.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,277 @@
...
+        headers = {
+          'Content-Type' => 'application/x-www-form-urlencoded'
+        }
+        headers['CompanyName'] = options[:company_name]
+        headers['AccessToken'] = options[:access_token] if options[:access_token].present?
+        headers['Invoice'] = options[:invoice] if options[:invoice].present?
+        headers
+      end
+
+      def success_from(response)
```
ajawadmirza, line 272: updated this method as per the standards

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,277 @@
...
+        )
+      end
+
+      def parse(body)
+        return {} if body == ''
+
+        JSON.parse(body)
+      end
+
+      def message_from(response)
```
ajawadmirza, line 248: updated this method as per the standards

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,277 @@
...
+        success_from(response) ? 'Transaction successful' : response['longText']
+      end
+
+      def error_code_from(response)
```
ajawadmirza, line 252: updated this method as per the standards

lib/active_merchant/posts_data.rb
```diff
@@ -70,7 +70,7 @@ def raw_ssl_request(method, endpoint, data, headers = {})
 
       connection.proxy_address = proxy_address
       connection.proxy_port    = proxy_port
-
```
ajawadmirza, line 73: No change is required in this file

ajawadmirza (changes requested):
Hi @ali-hassan ,

I have added some changes in your code and have requested few changes that should be incorporated as per the standards. Some rubocops are failing as well so you can update them and we will get it reviewed then. Thanks 👍

ali-hassan:
@therufs Can you please review it again as all the issues you raised been fixed please check and merge if all good :)

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,300 @@
...
+
+      def add_clerk(post, options)
+        post[:clerk] = {}
+        post[:clerk][:numericId] = options[:numeric_id]
+      end
+
+      def add_invoice(post, money, options)
+        post[:amount] = {}
+        post[:amount][:total] = money
+        post[:amount][:tax] = options[:amount][:tax] || 0.0
```
naashton, line 115: Does this value need to be a `float` or an `int`? I see in unit test, we have a test value of `0` for this, but the remote test has a decimal value. 

If it's supposed to be a `float` we should use `int` for our test values and use `to_f` to type cast the value as we put it in the post body.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,300 @@
...
+
+      def add_credentials(post, options)
+        post[:dateTime] = options[:date_time] || DateTime.now.strftime('%Y-%m-%dT%H:%M:%S.%N+%H:%M')
+        post[:credential] = {}
+        post[:credential][:clientGuid] = options[:credential][:client_guid]
+        post[:credential][:authToken] = options[:credential][:auth_token]
+      end
+
+      def add_datetime(post, options)
+        post[:dateTime] = options[:dateTime] || DateTime.now.strftime('%Y-%m-%dT%H:%M:%S.%N+%H:%M')
```
naashton, line 130: I think `options[:dateTime]` should be `options[:date_time]`.

Also, would it be useful to put `DateTime.now.strftime('%Y-%m-%dT%H:%M:%S.%N+%H:%M')` into a helper function? If the gateway always expects this format for any `date` type value.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,300 @@
...
+        if options[:hotel].present?
+          post[:hotel] = {}
+          post[:hotel][:estimatedDays] = options[:hotel][:estimated_days]
+        end
+        if options[:pin].present?
+          post[:pin] = {}
+          post[:pin][:block] = options[:pin][:block]
+          post[:pin][:ksn] = options[:pin][:ksn]
+        end
+      end
```
naashton, line 168: This method looks a bit large and it's not clear that all of the values are related? I would expect shared values added to the `post` body to be sub options for the primary hash field (in this case, `transaction`)

Can we put each of the conditionals (`overrideBusDate`, `amex`, `auto`, `card_on_file`, `hotel`, `pin`) into their own helper methods and then we can call those methods from here?

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,300 @@
...
+          post[:card][:number] = credit_card.number
+          post[:card][:entryMode] = options[:entry_mode]
+          post[:card][:present] = options[:present]
+          post[:card][:securityCode] = {}
+        else
+          post[:card] = {} if post[:card].nil?
+          post[:card][:token] = {}
+          post[:card][:token][:value] = credit_card
+        end
+        if options[:card].present?
```
naashton, line 210: I don't think we should expect to pass any card information via the `options` hash.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,300 @@
...
+        if options[:p2pe].present?
+          post[:p2pe] = {}
+          post[:p2pe][:format] = options[:p2pe][:format]
+          post[:p2pe][:ksn] = options[:p2pe][:ksn]
+        end
+      end
+
+      def add_card(post, credit_card, options)
+        post[:card] = {}
+        if credit_card.respond_to?(:number)
```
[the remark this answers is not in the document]
naashton: Also worth noting, the `Check` class has a `number` attribute, so this code could try to pass a check as a cc (which would then fail when trying to access the `month` and `year` values).

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,321 @@
...
+
+  def test_failed_authorize
+    response = @gateway.authorize(@amount, @credit_card, @options.except(:access_token))
+
+    assert error?(response)
+    assert merchant_error(response)
+  end
+
+  def test_failed_capture
+    auth_token = '8038483489222221'
```
drkjc, line 197: Since both `auth_tokens` are used twice I think they should be put into a variable.

Also, minor suggestion you can take or leave but I think `card_token` might be a better name for the variable.

Do you happen to know if the token is then usable directly with the Shift4 API? Are we vaulting the PMs at Shift4? I'm wondering if this way of passing a token instead of PAN data is an example of using a card that's been `stored`.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,321 @@
...
+
+  def successful_transaction?(response)
+    token(response).present?
+  end
+
+  def token(response)
+    response_result(response)['card']['token']['value']
+  end
+
+  def success?(response)
```
drkjc, line 262: `test_helper.rb` provides an `assert_success` and `assert_failure` method so I think this method and `def error?` can be removed and substituted with those helpers. For example: https://github.com/activemerchant/active_merchant/blob/89922760f0a9586c946ededc10ad1364ed52b87d/test/remote/gateways/remote_airwallex_test.rb#L19

lib/active_merchant/billing/gateways/shift4.rb
drkjc: @ajawadmirza This method doesn't accurately check for errors in the response. Shift4 returns something like this with `error` inside an array.
```
@params=
  {"result"=>
    [{"error"=>
       {"longText"=>"No GTV PAN returned from host TOKEN01CE ENGINE16CE",
        "primaryCode"=>9101,
        "secondaryCode"=>6,
        "shortText"=>"SYSTEM ERROR"},
      "server"=>{"name"=>"UTGAPI09CE"}}]},
 @success=false,
 @test=true>
```

When the above is returned `response['error'].nil?` is always true and we add `'Transaction successful'` to the message even though it failed. Instead it should be `response['result'][0]['error'].nil?`. I don't think there's a standard way to write this method but happy to hear suggestions.

lib/active_merchant/billing/gateways/shift4.rb
drkjc: @ajawadmirza The same for this method. `response['longText']` will return nil since the response looks like this,
```
@params=
  {"result"=>
    [{"error"=>
       {"longText"=>"No GTV PAN returned from host TOKEN01CE ENGINE16CE",
        "primaryCode"=>9101,
        "secondaryCode"=>6,
        "shortText"=>"SYSTEM ERROR"},
      "server"=>{"name"=>"UTGAPI09CE"}}]},
 @success=false,
 @test=true>
```

drkjc (changes requested):
Hey @ali-hassan . Nice work on this so far. I haven't reviewed `shift4_test.rb` file yet, but have requested changes with the other files.

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: Yes this should definitely be `response['result'.first['error'].nil?`. I only converted the same logic that was all in commit method to these standard methods `success_from(response)`. 

@ali-hassan  please make the said changes along with test cases.

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: yes that should be `response['result'].first['error']['longText']`. 
I didn't update the response logic but only converted them to standard methods.
@ali-hassan  will do the required changes, thanks.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,321 @@
...
+      "card": {
+        "expiration_date": '0825',
+        "present": 'N',
+        "indicator": '1',
+        "value": '3333'
+      },
+      access_token: '4902FAD2-E88F-4A8D-98C2-EED2A73DBBE2'
+    }
+
+    @access_token =
```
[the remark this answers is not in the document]
ali-hassan: Resolved @drkjc

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,321 @@
...
+  end
+
+  def test_successful_capture
+    authorize_res = @gateway.authorize(@amount, @credit_card, @options)
+    auth_token = authorize_res.authorization
+
+    response = @gateway.capture(@amount, auth_token, @capture_params)
+    assert success?(response)
+
+    assert response.success?
```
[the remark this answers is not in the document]
ali-hassan: authorize_res is being used for pre capture step and test is for capture anyhow added assert for it. @drkjc

test/remote/gateways/remote_shift4_test.rb
ali-hassan: @drkjc Both have different values you can verify them as well.

Its actually card auth token and its being used as authorisation token that why named it auth token. @drkjc

test/remote/gateways/remote_shift4_test.rb
ali-hassan: @drkjc I have tested them before creating mine, The existing methods and shift4 error response have different response structure I even tried to use but unable to use existing one.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,321 @@
...
+        "value": '3333'
+      },
+      access_token: '4902FAD2-E88F-4A8D-98C2-EED2A73DBBE2'
+    }
+
+    @access_token =
+    @invoice = '3333333309'
+    @token = '8042677003331111'
+    @void_options = { company_name: 'Spreedly', invoice: @invoice, access_token: '4902FAD2-E88F-4A8D-98C2-EED2A73DBBE2' }
+  end
```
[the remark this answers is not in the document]
ali-hassan: Changed, Also some fields are duplicated as there values are changed. @drkjc

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,300 @@
...
+        post[:overrideBusDate] = options[:override_bus_date] if options[:override_bus_date].present?
+        if options[:amex].present?
+          post[:amex] = {}
+          post[:amex][:propertyCode] = options[:amex][:property_code]
+        end
+        if options[:auto].present?
+          post[:auto] = {}
+          post[:auto][:estimatedDays] = options[:auto][:estimated_days]
+        end
+        if options[:card_on_file].present?
```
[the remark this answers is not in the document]
ali-hassan: No its not as it has followings fields which will come from options.

          post[:cardOnFile][:indicator] = options[:card_on_file][:indicator]
          post[:cardOnFile][:usageIndicator] = options[:card_on_file][:usage_indicator]
          post[:cardOnFile][:scheduledIndicator] = options[:card_on_file][:scheduled_indicator]
          post[:cardOnFile][:indicator] = options[:card_on_file][:transaction_id]

lib/active_merchant/billing/gateways/shift4.rb
ali-hassan: @naashton Card object has other info too that is
          post[:card][:entryMode] = options[:card][:entry_mode]
          post[:card][:present] = options[:card][:present]
          post[:card][:securityCode] = {} unless post[:card][:securityCode]
          post[:card][:securityCode][:indicator] = options[:card][:indicator] if options[:card][:indicator]
          post[:card][:securityCode][:value] = options[:card][:value] if options[:card][:value]
          post[:card][:securityCode] = {}

lib/active_merchant/billing/gateways/shift4.rb
drkjc: Gotcha, thanks @ajawadmirza

test/remote/gateways/remote_shift4_test.rb
drkjc: @ali-hassan Gotcha, that makes sense. Thanks. I'll resolve this one.

test/remote/gateways/remote_shift4_test.rb
drkjc: @ali-hassan The existing methods aren't working because of the changes [requested here](https://github.com/activemerchant/active_merchant/pull/4415/files#r890646169), and [here](https://github.com/activemerchant/active_merchant/pull/4415/files#r890646674). 

I've tested it out and once those are fixed the existing methods should work.

drkjc (changes requested):

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,340 @@
...
+      end
+
+      def add_auto(post, options)
+        if options[:auto].present?
+          post[:auto] = {}
+          post[:auto][:estimatedDays] = options[:auto][:estimated_days]
+        end
+      end
+
+      def add_card_on_file(post, options)
```
drkjc, line 240: This still needs to be updated to work with [our Stored Credentials framework](https://spreedly.atlassian.net/wiki/spaces/PS/pages/1358856480/Shift4+Technical+Requirements#Spreedly-Stored-Credentials-Integration).

drkjc (changes requested):
@ali-hassan The unit test file needs to be updated in the same way as the remote test file, e.g. string keys need to be updated. I also noticed there are not `failed` tests in that file. Is there a particular reason for that?

[The `scrub` method](https://spreedly.atlassian.net/wiki/spaces/EN/pages/1231683955/Adding+New+Gateway+Integration+Support+to+Active+Merchant#Standard-Menu-Actions) needs to be added to the gateway adapter and tested as well.

The `card_on_file` method needs to conform to [our Standard Stored Credentials actions](https://spreedly.atlassian.net/wiki/spaces/EN/pages/1231683955/Adding+New+Gateway+Integration+Support+to+Active+Merchant#Standard-Menu-Actions).

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+        add_invoice(post, money, options)
+        add_clerk(post, options)
+        add_transaction(post, options)
+        add_device(post, options)
+        add_p2pe(post, options)
+        add_emv(post, options)
+        add_customer(post, options)
+        add_card(post, authorization, options)
+
+        commit('refund', post, request_headers(options))
```
naashton, line 90: Small nit, but usually we pass `options` to the `commit` method and then call the `headers` method within the `commit` function. This is useful in cases where `options` is needed in that method.

lib/active_merchant/billing/gateways/shift4.rb
naashton: I think it's fine to expect `entry_mode`, `present`, and maybe `indicator` (not quite sure what this is) to be passed in the `options` hash, but for security reasons, we should grab the `security_code` and `value` (I'm assuming this correlates to `credit_card.number`) from our `CreditCard` object.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+
+      def add_p2pe(post, options)
+        if options[:p2pe].present?
+          post[:p2pe] = {}
+          post[:p2pe][:format] = options[:p2pe][:format]
+          post[:p2pe][:ksn] = options[:p2pe][:ksn]
+        end
+      end
+
+      def add_card(post, credit_card, options)
```
naashton, line 182: I think this method could use a slight refactor. The first conditional block is fine, though I would use the suggested check below (`is_a?(CreditCard)`) to add the necessary information to the post body, and then for additional fields, you can simply append them to the hash with simple one liners outside of the conditional block.

For lines `187-188` and `196-197`, you're adding the same option to `post` but expecting the values to be in different places in `options`. I think `187-188` is the way we should handle it, unless we find that we are handling a lot of extra `card` metadata that becomes cumbersome to manage on an individual field basis.

Similarly, ln `203-207` are doing what `198-200` are doing. As I mentioned in my other comment however, we should grab `security_value` from the `CreditCard` object as well as other `CreditCard` attributes (expiration `year`, `month`, etc.).

Since this is a lot, here's some pseudocode of what I'm thinking a refactor might look like:

```ruby
if CreditCard
  post[card_num] = CreditCard.number
  post[exp_mon] = CreditCard.month
  post[exp_year] = CreditCard.year
  post[present] = options[present]
else
  post[token] = token

# additional information not in the CreditCard object
return
```

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+          post[:card][:token][:value] = credit_card
+        end
+        if options[:card].present?
+          post[:card][:expirationDate] = options[:card][:expiration_date] if options[:card][:expiration_date]
+          post[:card][:entryMode] = options[:card][:entry_mode]
+          post[:card][:present] = options[:card][:present]
+          post[:card][:securityCode] = {} unless post[:card][:securityCode]
+          post[:card][:securityCode][:indicator] = options[:card][:indicator] if options[:card][:indicator]
+          post[:card][:securityCode][:value] = options[:card][:value] if options[:card][:value]
+          post[:card][:securityCode] = {}
```
naashton, line 201: Looks like this will empty the hash after appending data to it, so if this were to remain after refactor, you'll want to remove this line as well.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+        end
+      end
+
+      def add_customer(post, options)
+        if address = options[:customer]
+          post[:customer] = {}
+          post[:customer][:addressLine1] = address[:address_line1] if address[:address_line1]
+          post[:customer][:firstName] = address[:first_name]
+          post[:customer][:lastName] = address[:last_name]
+          post[:customer][:postalCode] = address[:postal_code]
```
naashton, line 216: `address` data will consistently be presented in the `billing_address` sub-hash of the `options` hash. I'll leave another comment in the remote test file to suggest a way to build it to fit the common pattern in AM.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+      def add_card_on_file(post, options)
+        if options[:stored_credential].present?
+          post[:cardOnFile] = {}
+          post[:cardOnFile][:indicator] = options[:stored_credential][:indicator] if options[:stored_credential][:indicator]
+          if usage_indicator = options[:stored_credential][:usage_indicator] == '01' ? 'Initial' : 'Subsequent'
+            post[:cardOnFile][:usageIndicator] = usage_indicator
+          end
+          if scheduled_indicator = options[:stored_credential][:scheduled_indicator] == '01' ? 'Scheduled' : 'Unscheduled'
+            post[:cardOnFile][:scheduledIndicator] = scheduled_indicator
+          end
```
naashton, line 259: It looks like your checks are backwards here. You'll want to check if the `options[:stored_credential][:schedule_indicator] == 'scheduled' ...` and then assign the `01` (which looks to be the Shift4 value) to the variable that is added to the `post` body.

naashton (changes requested):
Thanks for your patience in this review process @ali-hassan. I know we've provided a lot of review, but it'll make for the other half of this integration much easier. Let me know if you have any questions or concerns about feedback I've provided.

ali-hassan:
@naashton Updated PR as per your suggestion please check and merge.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+        if options[:device].present?
+          post[:device] = {}
+          post[:device][:manufacturer] = options[:device][:manufacturer]
+          post[:device][:model] = options[:device][:model]
+          post[:device][:promptPostalCode] = options[:device][:prompt_postal_code]
+          post[:device][:promptSecurityCode] = options[:device][:prompt_security_code]
+          post[:device][:promptStreetNumber] = options[:device][:prompt_street_number]
+          post[:device][:terminalId] = options[:device][:terminal_id]
+        end
+      end
```
naashton, line 165: It looks like we don't need this for **Card Number Unencrypted** transactions?

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+            post[:card][:present] = options[:card][:present]
+            post[:card][:securityCode] = {} unless post[:card][:securityCode]
+            post[:card][:securityCode][:indicator] = options[:card][:indicator] if options[:card][:indicator]
+            post[:card][:securityCode][:value] = options[:card][:value] if options[:card][:value]
+          end
+          if options[:security_code].present?
+            post[:card][:securityCode] = {}
+            post[:card][:securityCode][:indicator] = options[:security_code][:indicator]
+            post[:card][:securityCode][:value] = options[:security_code][:value]
+          end
```
naashton, line 205: We don't need this for the token. It looks like passing the token value is sufficient.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+        end
+      end
+
+      def add_card(post, credit_card, options)
+        post[:card] = {}
+        if credit_card.is_a?(CreditCard)
+          post[:card][:expirationDate] = "0#{credit_card.month}#{credit_card.year.to_s[2..4]}"
+          post[:card][:number] = credit_card.number
+          post[:card][:entryMode] = options[:entry_mode]
+          post[:card][:present] = options[:present]
```
naashton, line 188: This is where you should add the `security_code` data. You should be able to pull the `value` from the `credit_card.verification_value`. For the `indicator`, I think we can default that to `1`. We also don't need `entry_mode` or `present` unless those were specifically requested.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+      end
+
+      def add_customer(post, options)
+        if address = options[:billing_address]
+          post[:customer] = {}
+          post[:customer][:addressLine1] = address[:address1] if address[:address1]
+          name = address[:name].split(' ')
+          post[:customer][:firstName] = name[0]
+          post[:customer][:lastName] = name[1]
+          post[:customer][:postalCode] = address[:postal_code]
```
naashton, line 216: I think this is the only value required here and it looks to be only required on `authorize`. We can remove the rest unless it was specifically requested.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+        end
+      end
+
+      def add_purchase_card(post, options)
```
naashton, line 220: I see that this is added in the `add_transaction` method, which is called in our transaction methods, but I see that `capture` doesn't support this hash? Do you know if that will cause errors or if you have tests that cover this scenario?

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,354 @@
...
+        end
+      end
+
+      def add_pin(post, options)
+        if options[:pin].present?
+          post[:pin] = {}
+          post[:pin][:block] = options[:pin][:block]
+          post[:pin][:ksn] = options[:pin][:ksn]
+        end
+      end
```
naashton, line 277: We can remove these unless specifically requested.

naashton (changes requested):
Looking good. I think this needs some more revision before it's ready for Core integration. Reach out if you have any questions.

ali-hassan:
@naashton I have removed access_token method and add accessToken in fixtures as it needed only once.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,285 @@
...
+        'card_declined' => STANDARD_ERROR_CODE[:card_declined],
+        'processing_error' => STANDARD_ERROR_CODE[:processing_error],
+        'lost_or_stolen' => STANDARD_ERROR_CODE[:card_declined],
+        'suspected_fraud' => STANDARD_ERROR_CODE[:card_declined],
+        'expired_token' => STANDARD_ERROR_CODE[:card_declined]
+      }
+
+      def initialize(options = {})
+        requires!(options, :access_token)
+        super
```
ajawadmirza, line 33: assign this options to instance something like this: 
`@options = options`
so that they can be used directly in headers without needing them to pass for each operation.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,285 @@
...
+      def authorization_from(response, action)
+        response.dig('result', 0, 'card', 'token', 'value')
+      end
+
+      def request_headers(options)
+        headers = {
+          'Content-Type' => 'application/x-www-form-urlencoded'
+        }
+        headers['CompanyName'] = options[:company_name]
+        headers['AccessToken'] = options[:access_token]
```
ajawadmirza, line 263: https://github.com/activemerchant/active_merchant/pull/4415/files#r897876783

After updating you can get access toke from @options[:access_token]

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,249 @@
+require 'test_helper'
+
+class RemoteShift4Test < Test::Unit::TestCase
+  def setup
+    @gateway = Shift4Gateway.new(fixtures(:shift4))
+    @access_token = @gateway.options[:access_token]
```
ajawadmirza, line 6: No need to get them as we will already be passing it through fixtures, after updating the code changes as requested above: https://github.com/activemerchant/active_merchant/pull/4415/files#r897877558

test/unit/gateways/shift4_test.rb
```diff
@@ -0,0 +1,630 @@
+require 'test_helper'
+
+class Shift4Test < Test::Unit::TestCase
+  include CommStub
+  def setup
+    @gateway = Shift4Gateway.new(fixtures(:shift4))
+    @credit_card = credit_card
+    @amount = 100
+    @access_token = '4902FAD2-E88F-4A8D-98C2-EED2A73DBBE2'
```
ajawadmirza, line 9: same https://github.com/activemerchant/active_merchant/pull/4415/files#r897878401

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,300 @@
...
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
+      private
+
+      def add_clerk(post, options)
+        post[:clerk] = {}
+        post[:clerk][:numericId] = options[:numeric_id]
```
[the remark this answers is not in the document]
ajawadmirza: I think renaming this `numeric_id` to `clerk_id` will be better as it will make more sense by this name.

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: resolved.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,285 @@
...
+          post[:device][:promptSecurityCode] = options[:device][:prompt_security_code]
+          post[:device][:promptStreetNumber] = options[:device][:prompt_street_number]
+          post[:device][:terminalId] = options[:device][:terminal_id]
+        end
+      end
+
+      def add_card(post, credit_card, options)
+        post[:card] = {}
+        if credit_card.is_a?(CreditCard)
+          post[:card][:expirationDate] = "0#{credit_card.month}#{credit_card.year.to_s[2..4]}"
```
ajawadmirza, line 147: we can update this logic to:
`"#{format(credit_card.month, :two_digits)}#{format(credit_card.year, :two_digits)}"`

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,285 @@
...
+        if options[:stored_credential].present?
+          post[:cardOnFile] = {}
+          post[:cardOnFile][:indicator] = options[:stored_credential][:indicator] if options[:stored_credential][:indicator]
+          if usage_indicator = options[:stored_credential][:usage_indicator] == 'initial' ? '01' : '02'
+            post[:cardOnFile][:usageIndicator] = usage_indicator
+          end
+          if scheduled_indicator = options[:stored_credential][:scheduled_indicator] == 'scheduled' ? '' : '02'
+            post[:cardOnFile][:scheduledIndicator] = scheduled_indicator
+          end
+          post[:cardOnFile][:transactionId] = options[:stored_credential][:transaction_id] if options[:stored_credential][:transaction_id]
```
ajawadmirza, line 206: This should be `options[:stored_credential][:network_transaction_id]`

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,273 @@
...
+        'incorrect_zip' => STANDARD_ERROR_CODE[:incorrect_zip],
+        'card_declined' => STANDARD_ERROR_CODE[:card_declined],
+        'processing_error' => STANDARD_ERROR_CODE[:processing_error],
+        'lost_or_stolen' => STANDARD_ERROR_CODE[:card_declined],
+        'suspected_fraud' => STANDARD_ERROR_CODE[:card_declined],
+        'expired_token' => STANDARD_ERROR_CODE[:card_declined]
+      }
+
+      def initialize(options = {})
+        requires!(options, :access_token)
```
naashton, line 32: It looks like this has changed, but it looks like we need `client_guid` and `auth_token` in order to retrieve this `access_token`?

In the docs for [Access Token Exchange](https://myportal.shift4.com/index.cfm?action=development.shift4api#section/Development-Quick-Start/Access-Token-Exchange), we should require those 2 values here and execute the request to exchange those for an access token that we then store with the gateway in Core.

Take a look at the `initialize` method in [Airwallex](https://github.com/activemerchant/active_merchant/blob/023f3e8685e763c8758292790e2962dbaffc0c1b/lib/active_merchant/billing/gateways/airwallex.rb#L35). In the constructor, we make a call to the required endpoint to generate this value, that we in turn store on our end whenever a customer creates a gateway.

The request is defined in the [`setup_access_token` method](https://github.com/activemerchant/active_merchant/blob/023f3e8685e763c8758292790e2962dbaffc0c1b/lib/active_merchant/billing/gateways/airwallex.rb#L128-L141) which I think you can probably copy over and be able to get it to work here.

It also notes in the documentation that the `auth_token` is discarded after the call to retrieve the `access_token`, so we may need to consider how to handle that scenario so that users are re-creating Shift4 gateways in our system with a bunch of `auth_token`s.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,273 @@
...
+      end
+
+      def add_override_bus_date(post, options)
+        post[:overrideBusDate] = options[:override_bus_date] if options[:override_bus_date].present?
+      end
+
+      def add_amex(post, options)
+        if options[:amex].present?
+          post[:amex] = {}
+          post[:amex][:propertyCode] = options[:amex][:property_code]
```
naashton, line 173: I think we should just pass this value like so `options[:property_code]` and not nest it in an `amex` hash (unless there is a good reason to do so).

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,273 @@
...
+        end
+      end
+
+      def add_auto(post, options)
+        if options[:auto].present?
+          post[:auto] = {}
+          post[:auto][:estimatedDays] = options[:auto][:estimated_days]
```
naashton, line 180: Same as `property_code` above, pass it as `options[:estimated_days]` unless we have a good reason to do so.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,273 @@
...
+      end
+
+      def error_code_from(response)
+        return unless success_from(response)
+
+        STANDARD_ERROR_CODE_MAPPING[response['primaryCode']]
+      end
+
+      def authorization_from(response, action)
+        response.dig('result', 0, 'card', 'token', 'value')
```
naashton, line 243: Excellent!

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+      "tip": '',
+      "total": 5.00,
+      numeric_id: 24,
+      invoice: rand(4).to_s,
+      security_code: {
+        indicator: '1',
+        value: '4444'
+      }
+    }
+    @refund_options = @options.except(:invoice).merge({
```
naashton, line 27: Nice, I did not know about the `except` method in Ruby.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+        present: 'N',
+        indicator: '1',
+        value: '3333'
+      }
+    }
+    @token = '8042677003331111'
+    @void_options = { company_name: 'Spreedly', invoice: @invoice }
+  end
+
+  def test_successful_authorizef
```
naashton, line 70: Typo for `_authorizef`

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+    }
+    @refund_options = @options.except(:invoice).merge({
+      invoice: @invoice,
+      notes: 'Transaction notes are added here',
+      billing_address: address,
+      stored_credential: {
+        indicator: '01',
+        usage_indicator: '01',
+        scheduled_indicator: '01',
+        transaction_id: 'yxx'
```
naashton, line 35: Take a look at [remote_cyber_source_test.rb](https://github.com/activemerchant/active_merchant/blob/023f3e8685e763c8758292790e2962dbaffc0c1b/test/remote/gateways/remote_cyber_source_test.rb#L1009-L1052) for examples as to what `stored_credential` values we pass in the `options` hash.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+      token: {
+        serial_number: '',
+        value: '1111g66gw3ryke06'
+      }
+    })
+    @capture_params = @options.except(:invoice).merge({
+      transaction: {
+        invoice: @invoice
+      },
+      card: {}
```
naashton, line 46: I don't think you need this.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+      },
+      transaction: {
+        invoice: '4666309473',
+        notes: 'Test payment'
+      },
+      card: {
+        expiration_date: '0825',
+        present: 'N',
+        indicator: '1',
+        value: '3333'
```
naashton, line 63: You can remove this.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+    token(response).present?
+  end
+
+  def token(response)
+    response_result(response)['card']['token']['value']
+  end
+
+  def success?(response)
+    response_result(response)['error'].nil?
+  end
```
naashton, line 194: You can remove this.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+    credentials(response)['accessToken'].present?
+  end
+
+  def credentials(response)
+    response_result(response)['credential']
+  end
+
+  def successful_transaction?(response)
+    token(response).present?
+  end
```
naashton, line 186: You can remove this.

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,247 @@
...
+
+  def void_error_msg(response)
+    error_text(response).include?('Invoice Not Found')
+  end
+
+  def error_text(response)
+    error(response)['longText']
+  end
+
+  def error(response)
```
naashton, line 236: You should remove these and just retrieve the necessary values in the same line you make the asserts. What you end up with here is a bunch of nested method calls that can be difficult to triage if we run into errors and it also looks cluttered.

test/unit/gateways/shift4_test.rb
```diff
@@ -0,0 +1,588 @@
...
+    response_result(response).present?
+  end
+
+  def token?(response)
+    response_result(response)['card']['token']['value'].present?
+  end
+
+  def success_message(response)
+    response.message
+  end
```
naashton, line 191: Same recommendation as the remote test. Remove this and retrieve the values in-line with the assertions.

naashton (changes requested):
Overall looking really good. I think we still need to sort out our constructor and the values we need for authenticating with the gateway before transacting. 

I also left various comments about minor things that need to be changed and a few remnant variables and methods that can be removed just to clean up the test files.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,290 @@
...
+        'suspected_fraud' => STANDARD_ERROR_CODE[:card_declined],
+        'expired_token' => STANDARD_ERROR_CODE[:card_declined]
+      }
+
+      def initialize(options = {})
+        requires!(options, :client_guid, :auth_token)
+        @client_guid = options[:client_guid]
+        @auth_token = options[:auth_token]
+        super
+        @access_token = setup_access_token
```
naashton, line 36: Excellent!

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -0,0 +1,290 @@
...
+        post = {}
+        add_credentials(post, options)
+
+        response = commit('accesstoken', post, request_headers(options))
+        response.params['result'][0]['credential']['accessToken']
+      end
+
+      def add_clerk(post, options)
+        post[:clerk] = {}
+        post[:clerk][:numericId] = options[:numeric_id]
```
naashton, line 128: I agree with @ajawadmirza earlier comment about changing `options[:numeric_id]` too `options[:clerk_id]`. I think it'll make more sense when we expose this field to end uses.

naashton (changes requested):
This is pretty much ready to go. There are still a few things, like the `_authorizef` typo, and I also think we should change that one field to `clerk_id` instead of `numeric_id` like @ajawadmirza suggested. I also caught 1 lint issue with rubocop, but that can be cleaned up with the `-a` option pretty easily.

I also hit one remote test failure in `test_successful_purchase`, but it looks like it might be a Shift4 issue and not one as a result of our integration:

```
Failure: test_successful_purchase(RemoteShift4Test):
  Response expected to succeed: <#<ActiveMerchant::Billing::Response:0x00007fd9ba050eb0
   @authorization=nil,
   @avs_result=
    {"code"=>nil, "message"=>nil, "street_match"=>nil, "postal_match"=>nil},
   @cvv_result={"code"=>nil, "message"=>nil},
   @emv_authorization=nil,
   @error_code=nil,
   @fraud_review=nil,
   @message="No GTV PAN returned from host TOKEN02CE ENGINE19CE",
   @network_transaction_id=nil,
   @params=
    {"result"=>
      [{"error"=>
         {"longText"=>"No GTV PAN returned from host TOKEN02CE ENGINE19CE",
          "primaryCode"=>9101,
          "secondaryCode"=>6,
          "shortText"=>"SYSTEM ERROR"},
        "server"=>{"name"=>"UTGAPI11CE"}}]},
   @success=false,
   @test=true>>
```

After those changes are made, I'll approve and merge this up. Thanks again for the hard work @ali-hassan.

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: @naashton  with this change we will be sending access_token request even if its unit test case & will need correct credentials to pass all the unit test cases. Is it correct?

ajawadmirza: I have updated the implementation to pass in an extra key for unit tests so that the access_token call can be avoided in **initialize** method and in the **_shift4_test.rb file's setup method_**. Please suggest if this approach will be good?

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,177 @@
...
+  end
+
+  def test_failed_refund
+    response = @gateway.refund(nil, @credit_card, @refund_options.except(:access_token))
+
+    assert_failure response
+    assert response_result(response)['error']['longText'].include?('Secondary amounts cannot exceed the total amount')
+  end
+
+  def ff
```
ajawadmirza, line 138: test should be correctly named

test/remote/gateways/remote_shift4_test.rb
```diff
@@ -0,0 +1,177 @@
...
+    res = @gateway.purchase(@amount, @token, @purchase_params)
+    transaction = res.params['result'][0]['transaction']
+    card = res.params['result'][0]['card']
+    invoice = transaction['invoice']
+    auth_token = card['token']['value']
```
ajawadmirza, line 143: it should be able to get auth from `res.authorization`

ajawadmirza (changes requested):
I have found some other changes that should be made before merging as it may cause integration issues.

naashton (approved):
I think there might be a few things to cleanup, but I'll create some tickets to handle that. This looks like it's in working order and ready to integrate!

PR #4416, opened 2022-05-09
Braintree Blue: Return client token in AM response

BritneyS:
For ECS-2435

Test Summary
Local: 5176 tests, 75655 assertions, 0 failures, 18 errors, 0 pendings, 0 omissions, 0 notifications
99.6522% passed
Unit: 90 tests, 200 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote: 96 tests, 524 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -22,6 +22,7 @@
 * DLocal: Mark support for additional countries [gasb150] #4427
 * Authorize.net: Enable zero auth for allowed cards [jherreraa] #4430
 * Rapyd: Additional Fields [naashton] #4434
+* Braintree: Return generated client token [BritneyS] #4416
 
 == Version 1.126.0 (April 15th, 2022)
 * Moneris: Add 3DS MPI field support [esmitperez] #4373
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -75,6 +75,12 @@ module ActiveMerchant #:nodoc:
         @braintree_gateway = Braintree::Gateway.new(@configuration)
       end
 
+      def setup_purchase
+        commit do
+          Response.new(true, 'Client token created', { client_token: @braintree_gateway.client_token.generate })
+        end
+      end
+
       def authorize(money, credit_card_or_vault_id, options = {})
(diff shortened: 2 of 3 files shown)


tatsianaclifton:
Is there a reason why you didn't add unit tests? Also I see that not all local tests passed for you. I haven't worked with AM for a while, but as I remember all local tests should pass. Do you see the same failures on master?

molbrown (approved):
I get passing tests, and I see this follows the pattern already used in ACH support.

BritneyS:
> > > Is there a reason why you didn't add unit tests? Also I see that not all local tests passed for you. I haven't worked with AM for a while, but as I remember all local tests should pass. Do you see the same failures on master?
> > 
> > 
> > Yep, those same tests fail on `master`, and they have to do with the `Monei` gateway (so, an unrelated gateway).
> > I didn't add unit tests in this case since this work only exposes what is returned by Braintree within Active Merchant via a local method. This method is already tested by default here, when it is used to generate the payment nonce: https://github.com/activemerchant/active_merchant/blob/master/test/remote/gateways/remote_braintree_token_nonce_test.rb#L29
> > What do you think @tatsianaclifton ?
> 
> You maybe don't have credentials for Monei gateway in your fixtures files. But, I checked and tests are passing for me. So no issues here. As for unit tests, I didn't meant unit tests for functionality that already exist, but for that was added. The `setup_purchase` method is generating a new Response that consumer of this method is relying on. So if that Response object changes consumer will break. But, I understand that too many tests are not also good also, so I leave up to you if you want to add any unit tests. It looks good to me, just need to add CHANGELOG and :shipit:

@tatsianaclifton That makes sense. I gave it a second look, and I'm still leaning towards keeping things as-is. Generating the client token seems very isolated to the Braintree gem in AM, and differs from modules like `Braintree::TransactionGateway` that has a chance of failing due to what is passed in a `:sale` transaction. The client token generating method lacks those dependencies, so any failure for the Braintree gem to deliver that client token is covered by the remote test in this commit.

I will definitely keep your points in mind for future work on our side 👍🏾

PR #4417, opened 2022-05-10
Credorax: Update url gateway and credit cards

javierpedrozaing:
Unit test
-------------------------
Finished in 0.300459 seconds.

78 tests, 374 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

259.60 tests/s, 1244.76 assertions/s

Remote test
-------------------------
Finished in 264.711517 seconds.

46 tests, 157 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
86.9565% passed

0.17 tests/s, 0.59 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -4,7 +4,7 @@ module ActiveMerchant #:nodoc:
       class_attribute :test_url, :live_na_url, :live_eu_url
 
       self.display_name = 'Credorax Gateway'
-      self.homepage_url = 'https://www.credorax.com/'
+      self.homepage_url = 'https://www.finaro.com/'
 
       # NOTE: the IP address you run the remote tests from will need to be
       # whitelisted by Credorax; contact support@credorax.com as necessary to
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -3,8 +3,8 @@ module Billing #:nodoc:
     class CredoraxGateway < Gateway
       class_attribute :test_url, :live_na_url, :live_eu_url
 
-      self.display_name = 'Credorax Gateway'
-      self.homepage_url = 'https://www.credorax.com/'
+      self.display_name = 'Finaro Gateway'
```
curiousepic, line 6: I don't see any direct references to a "Finaro Gateway" in their site or docs, so for now let's retain this older name in AM - we actually use a different field in Core to display the name in  Docs anyway, so we can change it there.

```suggestion
      self.display_name = 'Credorax Gateway'
```

curiousepic (changes requested):

PR #4424, opened 2022-05-12
Wompi: Redirect `refund` to `void`

drkjc:
Refunds issued before 9PM Colombia time will fail. To fix this Wompi is developing a `refund_or_void` endpoint, which will determine the time and run the proper transaction. To avoid failures in the meantime, all refunds will now be run as `void` calls. This is temporary.

I've created a new unit test to cover the `refund to void` flow and have
commented out the the other refund tests, which are superfluous at the
moment. I didn't delete them because we are unsure what the new endpoint will look like, and the tests may still be relevant in the future.

CE-2635

Unit:
11 tests, 49 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
12 tests, 31 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


naashton:
Just assuming that `refund` and `void` behave similarly to other gateways (`refund` only working on transactions that have been `captured` and `void` on transactions that have not) we still want to execute a `refund` when the specified transaction method is called.

I think we can try to tackle this weird issue in a couple of different ways:
* Implement some try/catch/retry logic where we attempt a `refund` and if the response indicates that a transaction has not yet been `captured` we can attempt a `void`
* Grab the timestamp from the reference transaction and compare the date/time to a timestamp generated at runtime and then decide which transaction to run based on those conditions

This recommendation is based on my own assumptions about the gateway since the documentation I was looking for, about their transaction methods, is a bit elusive.

rachelkirk (approved):
Approved! (For posterity: Asked question in Slack about if voids were allowed on settled as well as unsettled transactions and @drkjc indicated that they were)

PR #4425, opened 2022-05-12
Simetrik: Fixes VAT not as cents

simetrik-frank:
VAT quantity was not in cents, resulting in an inconsistent user experience. Ej: An amount of 1000 would be $10 but a VAT of 19 would be $19.

This PR fixes this issue treating VAT as a number in cents.

Unit tests: 
12 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests:
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/simetrik.rb
+++ b/lib/active_merchant/billing/gateways/simetrik.rb
@@ -69,7 +69,7 @@ module ActiveMerchant #:nodoc:
             acquire_extra_options: options[:acquire_extra_options] || {}
           }
         }
-        post[:forward_payload][:amount][:vat] = options[:vat].to_f if options[:vat]
+        post[:forward_payload][:amount][:vat] = amount(options[:vat]).to_f if options[:vat]
 
         add_forward_route(post, options)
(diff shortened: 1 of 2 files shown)


rachelkirk (changes requested):
Thanks for making the changes for audience. I see that when running the remote tests for the VAT changes, they are passing. But when running our Core integration tests, there are some exceptions being raised. Please separate the VAT changes into another PR and I will approve this one for the audience changes. Submit the VAT PR and I'll post some notes about what I find that will work with our Core tests.

rachelkirk (changes requested):
I have suggested a different change that will accomplish the same result, `vat` being entered as cents. I did not run the unit or remote tests in Active Merchant yet but ran our Core tests and it's passing on our end.

simetrik-frank:
Hi @rachelkirk, following your recommendation, I've pushed a new commit which changes the way that the VAT conversion is done.

rachelkirk (approved):
Looks good! I'll get this merged. In the future, please squash your commits before finalizing your PR.

PR #4428, opened 2022-05-17
CyberSource add bank account payment method support

Heavyblade:
Summary:
---
This PR adds some test cases to ensure the current ACH implementation for CyberSource and modifies the credit method so it is able to support 'follow on' credits on transactions created with ACH debits.

**Note:** The 3 failing remote tests correspond to authentication failures because of the lack of a latam-related account.

Test Execution:
---
Remote
Finished in 89.497321 seconds.
110 tests, 555 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.2727% passed

Unit
Finished in 23.96186 seconds.
5183 tests, 75749 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -347,10 +347,11 @@ module ActiveMerchant #:nodoc:
         add_mdd_fields(xml, options)
         add_sales_slip_number(xml, options)
         add_airline_data(xml, options)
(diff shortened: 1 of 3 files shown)


javierpedrozaing (commented):
Good job, LGTM

curiousepic (approved):
This looks good. The pattern of setting `option`s to pass info around internally is somewhat unusual for AM (as opposed to passing arguments all the way down the chain), but this is probably cleaner for this use case, and it's reasonable enough. 👍

PR #4429, opened 2022-05-18
CardConnect:  Add support for 3ds V2

javierpedrozaing:
Description
-------------------------

In order to support 3dsv2 properly this PR does an update
to the recommended files used for this transactions

Unit test
-------------------------
Finished in 0.938125 seconds.

27 tests, 109 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

28.78 tests/s, 116.19 assertions/s

Remote test
-------------------------
Finished in 29.117144 seconds.

25 tests, 58 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
92% passed

0.86 tests/s, 1.99 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/card_connect.rb
+++ b/lib/active_merchant/billing/gateways/card_connect.rb
@@ -87,7 +87,7 @@ module ActiveMerchant #:nodoc:
           add_currency(post, money, options)
           add_address(post, options)
           add_customer_data(post, options)
-          add_3DS(post, options)
+          add_three_ds(post, options)
           add_additional_data(post, options)
           post[:capture] = 'Y'
           commit('auth', post)
@@ -102,7 +102,7 @@ module ActiveMerchant #:nodoc:
         add_payment(post, payment)
         add_address(post, options)
         add_customer_data(post, options)
-        add_3DS(post, options)
+        add_three_ds(post, options)
         add_additional_data(post, options)
         commit('auth', post)
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/card_connect.rb
```diff
@@ -241,10 +241,12 @@ def add_additional_data(post, options)
         post[:userfields] = options[:user_fields] if options[:user_fields]
       end
 
-      def add_3DS(post, options)
-        post[:secureflag] = options[:secure_flag] if options[:secure_flag]
-        post[:securevalue] = options[:secure_value] if options[:secure_value]
-        post[:securexid] = options[:secure_xid] if options[:secure_xid]
+      def add_three_ds(post, options)
```
Heavyblade, line 244: 🛑   Important question:

I see here that you are removing 3DS v1 fields, does this change means that card connect is dropping support for 3DS v1 and just supports now 3DS v2 ?

Checking remote test 3DS v1 seems still working ...

Heavyblade (commented):
No style or code comments, but a big question on the  PR consecuences

lib/active_merchant/billing/gateways/card_connect.rb
javierpedrozaing: I thing they had not support for [3ds1]( https://docs.spreedly.com/payment-gateways/card-connect/ ) , instead it seems, this was an old implementation for 3ds2 .

javierpedrozaing: Hi @curiousepic looking into the Documentation [here](https://support.cardpointe.com/compliance/3-d-secure-2-0#if-i-already-use-3-d-secure-1-0-do-i-need-to-update-to-2-0 ), it seems they don't have support for 3ds v1 since 2020. 
Looks like they had an deprecated implementation in AM independently to Spreedly core, I mean in core they didn't have enable this feature. 
what do you thing?.

curiousepic: I see, this implementation looks good according to those docs 👍

curiousepic (approved):

PR #4430, opened 2022-05-18
Authorize.net: Enable zero auth for allowed cards

jherreraa:
Summary:
This PR modifies the verify operation to perform zero auth when is permitted

Test Execution:
Unit test
Finished in 0.271949 seconds.
108 tests, 640 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote test
Finished in 89.63354 seconds.
75 tests, 272 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

RuboCop:


test/remote/gateways/remote_authorize_net_test.rb
```diff
@@ -384,10 +384,20 @@ def test_successful_authorization_with_moto_retail_type
...
 
   def test_successful_verify
+    card_option = {
+      brand: 'jcb'
+    }
+    other_credit_card = credit_card('3088000000000017', card_option)
+    response = @gateway.verify(other_credit_card, @options)
+    assert_success response
+    assert_equal 'This transaction has been approved', response.message
+    assert_equal response.responses.count, 2
```
Heavyblade, line 394: 💬  Comment:

Pleae remember to fix that test that we saw is using response. responses.last, adding unneeded complexity to that assertion.

test/remote/gateways/remote_authorize_net_test.rb
```diff
@@ -384,10 +384,20 @@ def test_successful_authorization_with_moto_retail_type
   end
 
   def test_successful_verify
+    card_option = {
+      brand: 'jcb'
+    }
+    other_credit_card = credit_card('3088000000000017', card_option)
```
Heavyblade, line 390: 💬  Comment:

why not just add that option to the method call, is pretty small:

```Ruby
other_credit_card = credit_card('3088000000000017', brand: 'jcb')
```

Heavyblade (approved):
Looks pretty good, leave a couple comments with minor suggestions.

test/remote/gateways/remote_authorize_net_test.rb
jherreraa: The test have been modified, I change the assert to check the number of responses in the scenario of one dollar auth, thinking about both operations, auth and void, the responses.last is not used anymore.

test/remote/gateways/remote_authorize_net_test.rb
```diff
@@ -384,10 +384,17 @@ def test_successful_authorization_with_moto_retail_type
...
 
   def test_successful_verify
+    other_credit_card = credit_card('3088000000000017', brand: 'jcb')
+    response = @gateway.verify(other_credit_card, @options)
+    assert_success response
+    assert_equal 'This transaction has been approved', response.message
+    assert_equal response.responses.count, 2
+  end
+
+  def test_successful_verify_with_0_auth
```
curiousepic, line 394: I think the names are switched? I'll fix this in the merge.

curiousepic (approved):
👍

curiousepic:
Missed a blank line after the git title again, BTW.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -176,13 +176,33 @@ def credit(amount, payment, options = {})
...
+
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
-          r.process(:ignore_result) { void(r.authorization, options) }
+          r.process { authorize(amount, payment_method, options) }
+          r.process(:ignore_result) { void(r.authorization, options) } unless amount == 0
         end
       end
 
+      def define_amount_for_verify(payment_method, options)
```
javierpedrozaing, line 189: 💬 
maybe you can refactor a little bit this function.
```suggestion
      def define_amount_for_verify(payment_method, options)
        return 100 if payment_method.is_a?(CreditCard) && payment_method.brand == 'jcb'
     
        if options[:verify_amount].present? && payment_method.is_a?(CreditCard)
          requires!(options, :billing_address)
          requires!(options[:billing_address], :address1, :zip)
          return options[:verify_amount]
        end
        raise ArgumentError, 'verify_amount value must be an integer and greater than zero' if verify_amount_value_verification(options)        
      end
```

javierpedrozaing (commented):
well done, LGTM, only I leave a comment

Heavyblade (commented):
@jherreraa we are pretty close, leaved you some comments

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -176,13 +176,40 @@ def credit(amount, payment, options = {})
...
```
curiousepic, line 189: Small thing, but `define` is redundant in ruby method names
```suggestion
      def amount_for_verify(payment_method, options)
```

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -176,13 +176,40 @@ def credit(amount, payment, options = {})
         end
       end
 
-      def verify(credit_card, options = {})
+      def verify(payment_method, options = {})
+        amount = define_amount_for_verify(payment_method, options)
+        return amount if amount.class == Response
```
curiousepic, line 181: FYI this can be 
```suggestion
        return amount if amount.is_a?(Response)
```

curiousepic (changes requested):
Left a primary suggestion for refactoring and a couple other small things, but this is getting there.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -176,13 +176,40 @@ def credit(amount, payment, options = {})
...
 
+      def define_amount_for_verify(payment_method, options)
+        if options[:verify_amount].present?
+          if options[:verify_amount].class != Integer || options[:verify_amount] < 0
+            response = Response.new(false, 'verify_amount value must be an integer')
+          elsif options[:verify_amount] > 0
+            response = options[:verify_amount]
+          elsif options[:verify_amount] == 0
+            response = 0
+            response = Response.new(false, 'zip and address must be different than blank value') unless validate_billing_address_values?(options)
```
curiousepic, line 197: Have we been able to confirm that these are the fields that are actually required for a $0 auth?

test/remote/gateways/remote_authorize_net_test.rb
jherreraa: code has been refactored, so please check again

lib/active_merchant/billing/gateways/authorize_net.rb
jherreraa: code has been refactored, so please check again

lib/active_merchant/billing/gateways/authorize_net.rb
jherreraa: Thanks a lot! Solved!

lib/active_merchant/billing/gateways/authorize_net.rb
jherreraa: solved! please check again!

lib/active_merchant/billing/gateways/authorize_net.rb
jherreraa: After several test with the combinations we have found these couple as the minimum requirement to 0 auth

lib/active_merchant/billing/gateways/authorize_net.rb
curiousepic: @jherreraa Not sure if you need to push again, but this doesn't seem to have been changed - ping me again when the PR is up to date

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -176,13 +176,40 @@ def credit(amount, payment, options = {})
...
-          r.process(:ignore_result) { void(r.authorization, options) }
+          r.process { authorize(amount, payment_method, options) }
+          r.process(:ignore_result) { void(r.authorization, options) } unless amount == 0
         end
       end
 
+      def define_amount_for_verify(payment_method, options)
+        if options[:verify_amount].present?
+          if options[:verify_amount].class != Integer || options[:verify_amount] < 0
+            response = Response.new(false, 'verify_amount value must be an integer')
```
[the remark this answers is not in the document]
jherreraa: I have checked this and need to comment we tried this because otherwise we will not be able to support the default flow, additionally, about the response, this was the way for us to raise the error in a proper way to have the possibility to test it in the remote tests.
If you have any other suggestion or recommendation about this I will be happy to check it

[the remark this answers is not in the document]
curiousepic: I see, thank you for the explanation.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -176,13 +176,40 @@ def credit(amount, payment, options = {})
...
 
+      def amount_for_verify(payment_method, options)
+        if options[:verify_amount].present?
+          if options[:verify_amount].class != Integer || options[:verify_amount] < 0
+            response = Response.new(false, 'verify_amount value must be an integer')
+          elsif options[:verify_amount] > 0
+            response = options[:verify_amount]
+          elsif options[:verify_amount] == 0
+            response = 0
+            response = Response.new(false, 'zip and address must be different than blank value') unless validate_billing_address_values?(options)
```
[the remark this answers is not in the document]
jherreraa: Corrected, message updated. Thanks a lot!

curiousepic (approved):
Thanks! Merging.

PR #4432, opened 2022-05-23
Plexo: Add new gateway

ajawadmirza:
Added plexo gateway implementation along with remote and unit test
cases.

CE-2536

Unit:
5201 tests, 75845 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
15 tests, 43 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/plexo.rb
+++ b/lib/active_merchant/billing/gateways/plexo.rb
@@ -1,0 +1,260 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class PlexoGateway < Gateway
+      self.test_url = 'https://api.testing.plexo.com.uy/v1/payments'
+      self.live_url = 'https://api.plexo.com.uy/v1/payments'
+
+      self.supported_countries = ['UY']
+      self.default_currency = 'UYU'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
+      self.homepage_url = 'https://www.plexo.com.uy'
+      self.display_name = 'Plexo'
+
+      APPENDED_URLS = %w(captures refunds cancellations)
+
+      def initialize(options = {})
+        requires!(options, :client_id, :api_key)
+        @credentials = options
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        response = MultiResponse.run do |r|
+          r.process { authorize(money, payment, options) }
+          r.process { capture(money, r.authorization, options) }
+        end
+        response.responses.last
+      end
+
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_plexo_test.rb
```diff
@@ -0,0 +1,151 @@
...
+  def test_failed_void
+    response = @gateway.void('123', @cancel_options)
+    assert_failure response
+    assert_equal 'An internal error occurred. Contact support.', response.message
+  end
+
+  def test_successful_verify
+    response = @gateway.verify(@credit_card, @options.merge(@cancel_options))
+    assert_success response
+    assert_match %r{You have been mocked.}, response.message
```
therufs, line 124: Is there a reason to switch from `assert_equal` to `assert_match` here? I think either would be fine, but am wondering why there are some of each.

lib/active_merchant/billing/gateways/plexo.rb
```diff
@@ -0,0 +1,260 @@
...
+      def initialize(options = {})
+        requires!(options, :client_id, :api_key)
+        @credentials = options
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        response = MultiResponse.run do |r|
+          r.process { authorize(money, payment, options) }
+          r.process { capture(money, r.authorization, options) }
```
therufs, line 25: What happens here if the `authorize` fails?

therufs (commented):
Hey @ajawadmirza -- I had a couple of questions, but this looks good!

test/remote/gateways/remote_plexo_test.rb
ajawadmirza: You're right, I have changed it to `assert_equal` to keep it consistent with other tests as there is no any specific reason to use that.

lib/active_merchant/billing/gateways/plexo.rb
ajawadmirza: It should be able to respond with error as expected. Here is the test to cover that case:

https://github.com/activemerchant/active_merchant/pull/4432/files#diff-b724dc322c2cc0818829951d8a93bb9fddbffd55fd3a717be266ecffe9d000a6R44

therufs: Great, thanks! I went looking for the test and just didn't find it.

therufs (approved):
tests pass && lgtm!

PR #4437, opened 2022-05-24
VisaNet Peru: purchase_number bug fix

rachelkirk:
CE-2624

This updates the existing time stamp logic and filters it to be an even more specific slice of time, and appends two randomly generated numbers  to the end of the string. The existing unit test (def test_nonconsecutive_purchase_numbers
 on line 42) seemed sufficient and is passing. 16  of 18 remote tests are failing and have been for at least 6 months.

Unit:
16 tests, 96 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
18 tests, 27 assertions, 16 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
11.1111% passed

Local:
5184 tests, 75754 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,6 +26,7 @@
 * CyberSource: Add bank account payment method support [heavyblade] #4428
 * Rapyd: Zero Dollar Auth [naashton] #4435
 * Rapyd: Scrub ACH [naashton] #4436
+* VisaNet Peru: Update `purchase_number` [rachelkirk] #4437
 
 == Version 1.126.0 (April 15th, 2022)
 * Moneris: Add 3DS MPI field support [esmitperez] #4373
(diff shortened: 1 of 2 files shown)


therufs (approved):
This seems good to me!

For posterity [by which I mean because I spent a lot of time thinking about it and need to validate my life choices], I did want to make one observation: sometimes, `rand(99)` generates a single digit number, so we end up with a purchase number that is only 11 digits instead of 12. 

I'm mentioning this mostly because it doesn't seem outside of the realm of possibility that someone will assume (or has already assumed) that the purchase numbers should be roughly sequential; in the case of a single digit random number, if the stamp is interpreted as an integer, the sequencing will be broken because the number will be about 6 million less than expected.

If we wanted to mitigate this, we could pad the random number with [sprintf](https://apidock.com/ruby/Kernel/sprintf); I'm not at all sure it's necessary, though. 

Also, because I'm a pedant, I feel obligated to point out that dropping the first digit means we may start having collisions again in about 52 years. :P

PR #4438, opened 2022-05-25
Add Ascension Island and Tristan da Cunha

devanandersen:
Similar to https://github.com/Shopify/active_utils/pull/123

Adding Ascension Island and Tristan da Cunha to the countries list under the ISO3166-1 special case of keeping them separate from Saint Helena. Pulled alpha3 from https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3, and alpha2 from https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2.

No numeric codes for AC and TA are listed at this time, example source https://en.wikipedia.org/wiki/ISO_3166-1_numeric, so left those blank for now.

### DIFF
--- a/lib/active_merchant/country.rb
+++ b/lib/active_merchant/country.rb
@@ -67,6 +67,7 @@ module ActiveMerchant #:nodoc:
     COUNTRIES = [
       { alpha2: 'AF', name: 'Afghanistan', alpha3: 'AFG', numeric: '004' },
       { alpha2: 'AL', name: 'Albania', alpha3: 'ALB', numeric: '008' },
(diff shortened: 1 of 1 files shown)


rochlefebvre (approved):
Looks ok to me.  These are the first countries to lack a numeric code format.  If someone finds either region then attempts to call `code(:numeric)`, they will get a `nil`, which is a first.

devanandersen:
@dominiquefl @rochlefebvre thank you!

It seems I do not have write access to this PR. I was wondering if either of you would be able to merge? 🙏🏼

PR #4439, opened 2022-05-25
Rapyd: Support `store` and `unstore`

naashton:
Add support for `store` and `unstore` transactions. `store` at a
minimum, requires the payment method and the customer name to create a
customer object at the Rapyd gateway.

CE-2602

Unit: 19 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 26 tests, 74 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -196,6 +196,31 @@ def test_failed_verify
     assert_equal 'Do Not Honor', response.message
   end
 
+  def test_successful_store_and_unstore
+    store = @gateway.store(@credit_card, @options)
+    assert_success store
+    assert customer_id = store.params.dig('data', 'id')
+
```
therufs, line 203: How would you feel about expanding this test to check that a purchase call with the stored card is successful, and to further expect a failed purchase attempt after the card is unstored?

therufs (commented):
This looks good; just had an idea about testing that I'd like to hear your thoughts on!

test/remote/gateways/remote_rapyd_test.rb
naashton: That sounds like a good idea. I'll add that test.

naashton: I'm trying to figure out the behavior here, but it looks like the payment method needs to go through 3DS authorization in order to be used. Seems odd since I'm using [what I thought was] a non-3DS payment method (`us_visa_card`).

If we want to test this, it'll have to be done manually. Drop these 2 lines below the `store` call and then use the `redirect_url` to go to the browser and "authorize" the card.

```ruby
store.params.dig('data', 'payment_methods', 'data', 0, 'redirect_url')
sleep(20) # adjust sleep as needed
```

test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -146,7 +146,7 @@ def test_partial_refund
   def test_failed_refund
     response = @gateway.refund(@amount, '')
     assert_failure response
-    assert_equal 'The request tried to retrieve a payment, but the payment was not found. The request was rejected. Corrective action: Use a valid payment ID.', response.message
+    assert_equal 'The request attempted an operation that requires a payment ID, but the payment was not found. The request was rejected. Corrective action: Use the ID of a valid payment.', response.message
```
naashton, line 149: The difference in response is because of the change I made to the conditional in `add_payment`, where it is attempting to send a `null` payment instead of `''` (empty string).

test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -196,6 +196,36 @@ def test_failed_verify
...
 
+  def test_successful_store_and_unstore
+    store = @gateway.store(@credit_card, @options)
+    assert_success store
+    assert customer_id = store.params.dig('data', 'id')
+    assert store.params.dig('data', 'default_payment_method')
+
+    # payment_token = store.params.dig('data', 'payment_methods', 'data', 0, 'id')
+    # purchase = @gateway.purchase(@amount, payment_token, {})
+    # assert_success purchase
```
naashton, line 207: LMK if I should add a clarifying comment on this or remove it completely.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -58,26 +58,37 @@ def authorize(money, payment, options = {})
...
       end
 
+      def store(payment, options = {})
+        post = {}
+        add_payment(post, payment, options)
+        add_customer_object(post, payment, options)
+        commit(:post, 'customers', post)
+      end
+
+      def unstore(customer)
```
dsmcclain, line 88: Is this `customer` value being passed in by the merchant as a gateway-specific field?

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -188,6 +199,10 @@ def add_payment_fields(post, options)
         post[:payment][:statement_descriptor] = options[:statement_descriptor] if options[:statement_descriptor]
       end
 
+      def add_customer_object(post, payment, options)
```
dsmcclain, line 202: Nit One: seems like `add_customer_string` and not `add_customer_object`. Or simply `add_name`.

Nit Two: `options` is not used in this method.

test/remote/gateways/remote_rapyd_test.rb
dsmcclain: Thanks for this explanation 👍

test/remote/gateways/remote_rapyd_test.rb
dsmcclain: Yes, I think one or the other. I would say it was a good idea initially, but considering that it doesn't work as part of the automated test my vote would be to remove it completely.

If you disagree, then yes I think at the least a comment on why this is here and how to use it would be appropriate.

dsmcclain (approved):
I have one question about how the `customer` data is supplied for the `unstore` call.

Also made a few smaller suggestions.

Other than that, it looks like you have done a good job making the adapter accommodate the rapyd pm tokens.

Tests pass, so I am approving. If you make changes and wish to have another review, don't hesitate to tag me for re-review 😄

test/remote/gateways/remote_rapyd_test.rb
naashton: Sounds good, I'd rather not have commented out code added to the main branch (that I'm pretty sure will just get forgotten 😞 )

lib/active_merchant/billing/gateways/rapyd.rb
naashton: Thanks for this call out. `customer` should just be the reference to the `store` transaction (which in Rapyd's case, should be the customer token). I modified this call to pass the `customer` to `add_reference` in order to properly extract that value.

lib/active_merchant/billing/gateways/rapyd.rb
naashton: I named this method to match the [Rapyd Customer Object](https://docs.rapyd.net/build-with-rapyd/reference/customer-object). Even though we are only utilizing it for the name, for now, it felt appropriate to "plan ahead" in case we want to pass more fields.

That should explain why I had an `options` argument here, but I removed it since it is not being used.

PR #4441, opened 2022-06-01
Plexo: Update implementation for `plexo`

ajawadmirza:
Added `meta_data` fields along with tests and updated response to
reorder amount field to support implementation in core.

CE-2567

Unit:
5206 tests, 75864 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
16 tests, 47 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/plexo.rb
+++ b/lib/active_merchant/billing/gateways/plexo.rb
@@ -37,6 +37,7 @@ module ActiveMerchant #:nodoc:
 
         add_payment_method(post, payment, options)
         add_items(post, options[:items])
+        add_meta_data(post, options[:meta_data])
         add_amount(money, post, options[:amount_details])
         add_browser_details(post, options[:browser_details])
         add_capture_type(post, options)
@@ -131,6 +132,13 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def add_meta_data(post, meta_data)
+        return unless meta_data
+
+        meta_data.transform_keys! { |key| key.to_s.camelize.to_sym }
+        post[:Metadata] = meta_data
+      end
+
       def add_amount(money, post, amount_options)
         post[:Amount] = {}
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/plexo.rb
```diff
@@ -208,10 +219,21 @@ def get_authorization_from_url(url)
...
+        response['currency'] = amount_obj['currency'] if amount_obj['currency']
+        response['amount_details'] = amount_obj['details'] if amount_obj['details']
+        response
+      end
+
       def commit(action, parameters)
         base_url = (test? ? test_url : live_url)
         url = build_url(action, base_url)
         response = parse(ssl_post(url, parameters.to_json, header(parameters)))
+        response = reorder_amount_fields(response) if action == 'authonly'
```
dsmcclain, line 236: Why are we doing this reordering, but only on an `authonly` transaction?

If this is necessary, I think it is worth including something in the tests showing that `authonly` transactions return responses with the amount fields in the expected format.

ajawadmirza: Discussed and I have added unit test to demonstrate the use case.

lib/active_merchant/billing/gateways/plexo.rb
```diff
@@ -131,6 +132,16 @@ def add_items(post, items)
         end
       end
 
+      def add_meta_data(post, meta_data)
+        return unless meta_data
+
+        post[:Metadata] = {}
+        meta_data.keys.each do |meta_key|
```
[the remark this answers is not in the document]
ajawadmirza: Thanks, I have changed it with the `transform_keys` method.

test/unit/gateways/plexo_test.rb
```diff
@@ -85,6 +89,31 @@ def test_successful_authorize_with_items
...
+
+  def test_successful_reordering_of_amount_in_authorize
+    @gateway.expects(:ssl_post).returns(successful_authorize_response)
+
+    original_response = JSON.parse(successful_authorize_response)
+    response = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal response.params['amount'], original_response['amount']['total']
+    assert_equal response.params['currency'], original_response['amount']['currency']
+    assert_equal response.params['amount_details'], original_response['amount']['details']
```
dsmcclain, line 114: fantastic 👍

dsmcclain (approved):
Thank you for adding that test @ajawadmirza 

All tests passing. I will merge this today.

PR #4447, opened 2022-06-07
Priority

priorityspreedly:
Remove "expirydate" field from credit card params

### DIFF
--- a/lib/active_merchant/billing/gateways/priority.rb
+++ b/lib/active_merchant/billing/gateways/priority.rb
@@ -72,6 +72,17 @@ module ActiveMerchant #:nodoc:
         commit('purchase', params: params)
       end
 
+      def credit(amount, credit_card, options = {})
+        params = {}
+        params['authOnly'] = false
+        params['isSettleFunds'] = true
+
+        add_merchant_id(params)
+        add_amount(params, amount, options)
+        add_credit_params(params, credit_card, options)
+        commit('credit', params: params)
+      end
+
       def refund(amount, authorization, options = {})
         params = {}
         add_merchant_id(params)
@@ -147,6 +158,12 @@ module ActiveMerchant #:nodoc:
         add_additional_data(params, options)
       end
 
+      def add_credit_params(params, credit_card, options)
+        add_replay_id(params, options)
+        add_credit_card(params, credit_card, 'purchase', options)
+        add_additional_data(params, options)
+      end
+
       def add_replay_id(params, options)
         params['replayId'] = options[:replay_id] if options[:replay_id]
       end
@@ -156,7 +173,6 @@ module ActiveMerchant #:nodoc:
         card_details = {}
         card_details['expiryMonth'] = format(credit_card.month, :two_digits).to_s
         card_details['expiryYear'] = format(credit_card.year, :two_digits).to_s
-        card_details['expiryDate'] = exp_date(credit_card)
         card_details['cardType'] = credit_card.brand
         card_details['last4'] = credit_card.last_digits
         card_details['cvv'] = credit_card.verification_value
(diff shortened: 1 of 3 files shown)


rachelkirk (changes requested):
Hey Mike, I'm going to look at this in the morning but at first glance can you squash your commits and then force push back to your branch? Active Merchant conventions like to keep commit history clean. Thanks!

priorityspreedly:
> Hey Mike, I'm going to look at this in the morning but at first glance, can you squash your commits, rebase, and then force push back to your branch? Active Merchant conventions like to keep commit history clean. Thanks!

Can you review now, we did a rebase with upstream/master?

test/remote/gateways/remote_priority_test.rb
```diff
@@ -5,6 +5,7 @@ def setup
     @gateway = PriorityGateway.new(fixtures(:priority))
 
     @amount = 2
+    @credit_amount = -2000
```
rachelkirk, line 8: Should the @credit_amount be negative? the refund tests all have positive integers as the @amount and are passing. If you are crediting a negative number that may have an unintended consequence.

test/remote/gateways/remote_priority_test.rb
```diff
@@ -147,6 +162,22 @@ def test_successful_purchase_with_additional_options
...
 
+  def test_successful_credit
+    options = @options.merge(@additional_creditoptions)
+    response = @gateway.credit(@credit_amount, @credit_card, options)
+    assert_success response
+    assert_equal 'Approved or completed successfully', response.message
+  end
+
+  def test_failed_credit_missing_card_month
+    card_without_month = credit_card('4242424242424242', month: '')
```
rachelkirk, line 173: Is this the only circumstance that would result in a failed credit? For example, I tried editing the test after pulling down your branch and passed in a nil for the amount and a valid card with expiration date and it allowed the transaction to be successful. That might be something you should validate on your end.

rachelkirk (changes requested):
There are some changes that are needed. Also, something weird is happening with the git sha /history. When I pulled down your branch and looked at it in my IDE it's attributing your work to one of our Spreedly developers @curiousepic, and the work will show up as his contribution upon merging to active merchant. Something must have happened when you rebased, this needs to be fixed. In the past I think Mike has had to create a new fork to fix this kind of problem.

test/remote/gateways/remote_priority_test.rb
priorityspreedly: Yes it should be negative that is how our system works.

lib/active_merchant/billing/gateways/priority.rb
```diff
@@ -147,17 +158,21 @@ def add_auth_purchase_params(params, credit_card, options)
...
         params['replayId'] = options[:replay_id] if options[:replay_id]
       end
 
       def add_credit_card(params, credit_card, action, options)
         return unless credit_card&.is_a?(CreditCard)
-
         card_details = {}
         card_details['expiryMonth'] = format(credit_card.month, :two_digits).to_s
         card_details['expiryYear'] = format(credit_card.year, :two_digits).to_s
-        card_details['expiryDate'] = exp_date(credit_card)
```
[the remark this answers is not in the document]
priorityspreedly: Expire date is not required, because we have expiry month and year. Having it may cause an issue in the Priority endpoint.

priorityspreedly:
> There are some changes that are needed. Also, something weird is happening with the git sha /history. When I pulled down your branch and looked at it in my IDE it's attributing your work to one of our Spreedly developers @curiousepic, and the work will show up as his contribution upon merging to active merchant. Something must have happened when you rebased, this needs to be fixed. In the past I think Mike has had to create a new fork to fix this kind of problem.

So I created a new PR using a new branch for fixing all the issue, was the easiest way I think to solve the rebase issues. Also in the new PR I made changes related to the comments in this PR, here is the link for the new PR https://github.com/activemerchant/active_merchant/pull/4517

PR #4450, opened 2022-06-08
VPOS: Reuse encryption key

therufs:
Unit: 
7 tests, 22 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 
11 tests, 36 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
90.9091% passed
Failing test: `test_failed_credit` (seems to have an updated error)

Rubocop: 

CER-9

### DIFF
--- a/lib/active_merchant/billing/gateways/vpos.rb
+++ b/lib/active_merchant/billing/gateways/vpos.rb
@@ -27,6 +27,7 @@ module ActiveMerchant #:nodoc:
         requires!(options, :private_key, :public_key)
         @private_key = options[:private_key]
         @public_key = options[:public_key]
+        @encryption_key = OpenSSL::PKey::RSA.new(options[:encryption_key]) if options[:encryption_key]
         @shop_process_id = options[:shop_process_id] || SecureRandom.random_number(10**15)
         super
       end
@@ -114,14 +115,14 @@ module ActiveMerchant #:nodoc:
         transcript.encode('UTF-8', 'binary', undef: :replace, replace: '')
       end
 
-      private
-
       # Required to encrypt PAN data.
       def one_time_public_key
         token = generate_token('get_encription_public_key', @public_key)
         response = commit(:pci_encryption_key, token: token)
-        OpenSSL::PKey::RSA.new(response.params['encryption_key'])
+        response.params['encryption_key']
       end
+
+      private
 
       def generate_token(*elements)
         Digest::MD5.hexdigest(@private_key + elements.join)
@@ -138,7 +139,9 @@ module ActiveMerchant #:nodoc:
 
         payload = { card_number: card_number, 'cvv': cvv }.to_json
 
-        post[:card_encrypted_data] = JWE.encrypt(payload, one_time_public_key)
+        encryption_key = @encryption_key || OpenSSL::PKey::RSA.new(one_time_public_key)
+
+        post[:card_encrypted_data] = JWE.encrypt(payload, encryption_key)
         post[:card_month_expiration] = format(payment.month, :two_digits)
         post[:card_year_expiration] = format(payment.year, :two_digits)
       end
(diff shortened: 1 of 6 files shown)


therufs:
Hey @activemerchant/spreedly-customer-eng, 

This is ready for re-review. The tests are now ... somewhat complicated, in ways I will describe here: 

I wanted to test both the path where we generate a fresh key per transaction (existing path) and the path where the gateway is set up with an encryption key (new path). 

However, every time we generate a new key, any existing keys are invalidated. This means one's fixtures need to be updated every time the existing path is tested. 

The (admittedly awkward) workaround I have come up with for this is to break the remote tests into two files. 
The existing file, `remote_vpos_test`, expects an `encryption_key` to be provided in the fixtures. 
The new file, `remote_vpos_without_key_test`, will drop the `encryption_key` if one is provided, then will generate a new key per transaction. 

Since this invalidates any key that might be in the fixtures, the test `test_regenerate_encryption_key` can be run independently after the full `remote_vpos_without_key_test` suite is run. This will print a fresh encryption key that can be put into one's fixtures so that `remote_vpos_test` can be run.

lib/active_merchant/billing/gateways/vpos.rb
```diff
@@ -138,7 +139,9 @@ def add_card_data(post, payment)
 
         payload = { card_number: card_number, 'cvv': cvv }.to_json
 
-        post[:card_encrypted_data] = JWE.encrypt(payload, one_time_public_key)
+        encryption_key = @encryption_key || OpenSSL::PKey::RSA.new(one_time_public_key)
```
rachelkirk, line 142: Is there a scenario where the `one_time_public_key` would still be used? I'm assuming this is a failsafe. Is this in place for the time when the switch is being made? Asking for my own knowledge. You mentioned that this would be a breaking change for them so I'm just curious about why the `one_time` is still included here.

test/remote/gateways/remote_vpos_without_key_test.rb
```diff
@@ -0,0 +1,125 @@
...
+    assert_no_match(/card_encrypted_data\\":\\"[^\[FILTERED\]]/, transcript)
+  end
+
+  def test_regenerate_encryption_key
+    puts 'Regenerating encryption key.'
+    puts 'Before running the standard vpos remote test suite, run this test individually:'
+    puts '$ ruby -Ilib:test test/remote/gateways/remote_vpos_without_key_test.rb -n test_regenerate_encryption_key'
+    puts 'Then copy this key into your fixtures file.'
+    p @gateway.one_time_public_key
+  end
```
rachelkirk, line 124: This is so cool! I have not come across anything like this before in a remote test.

test/remote/gateways/remote_vpos_without_key_test.rb
```diff
@@ -0,0 +1,125 @@
+require 'test_helper'
+
+class RemoteVposWithoutKeyTest < Test::Unit::TestCase
```
rachelkirk, line 3: What will be the implications of having two remote test files after this change is made? Will one of them no longer be needed? Genuine curiousity.

rachelkirk (approved):
I have learned a lot by reviewing this and feel like I need to learn more about certificates. You've answered some questions already and have a few more, approving with plans to pick your brain more about the particulars.

test/remote/gateways/remote_vpos_without_key_test.rb
rachelkirk: Please ignore this comment, I totally missed your block of notes above. I only saw what was in the original commit message.

lib/active_merchant/billing/gateways/vpos.rb
therufs: Yes. In summary, this is my attempt to make the breaking change a bit less breaking. :D

test/remote/gateways/remote_vpos_without_key_test.rb
therufs: "cool" is very kind of you, thanks 😂

PR #4451, opened 2022-06-10
Simetrik: Changes order.description and order.id to top-level fields

simetrik-frank:
Changes order.description and order.id to top-level fields.

Remote tests: 
100% passed

Unit tests: 
12 tests, 83 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


simetrik-frank:
Hi @rachelkirk , would appreciate if you could review this PR.

rachelkirk (changes requested):
This looks good and tests are passing, but we need to request two more changes. I apologize that these weren't caught and part of the original request: `billing_address` and `shipping_address` need to be exposed as top level fields as well. That should be it for changes unless something comes up during our QA process with our project manager. Thank you!

simetrik-frank:
Hi @rachelkirk, `billing_address` and `shipping_address` are already top level fields, I have pushed a commit which includes additional tests for these cases.

Please see `test_success_purchase_with_billing_address`, `test_success_purchase_with_shipping_address` unit tests.

rachelkirk (approved):
Thanks for making the changes, don't worry about the address fields. I was looking at a request for docs changes and got things mixed up, my mistake.

PR #4460, opened 2022-06-15
BraintreeBlue: Added Transaction Id and Customer Name

SivakeshPachipulusu:
- Added transaction id to the response parameters when transaction result is success
- Added first_name and last_name to the customer address hash if billing_address has the data

Unit:
5209 tests, 75883 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:


curiousepic:

PR #4461, opened 2022-06-16
Vanco: Purchase with session id from request

BritneyS:
Test Summary
Local: 5215 tests, 75910 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 12 tests, 60 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 15 tests, 34 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_vanco_test.rb
+++ b/test/remote/gateways/remote_vanco_test.rb
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class RemoteVancoTest < Test::Unit::TestCase
+  SECONDS_PER_DAY = 3600 * 24
+
   def setup
     @gateway = VancoGateway.new(fixtures(:vanco))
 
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_vanco_test.rb
```diff
@@ -34,6 +36,37 @@ def test_successful_purchase_with_ip_address
...
   end
 
+  def test_successful_purchase_with_existing_session_id
+    previous_login_response = @gateway.send(:login)
+
+    response = @gateway.purchase(
+      @amount,
+      @credit_card,
+      @options.merge(
+        session_id: { id: previous_login_response.params['response_sessionid'], created_at: Time.now }
```
molbrown, line 46: Where does the created_at value come from? Is it equivalent to the `auth_requesttime` returned in the login request?

BritneyS: I was expecting `created_at` be passed into the request by the client.

Btw, I'm also open to checking the validity of the session id on our side, if you have thoughts on that.

molbrown: Right, I guess I was thinking we would return all the required data for the `session_id` hash in the previous transaction response, to simplify the client passing them into an existing_session_id request, but I wanted to see if this assumption matched your thinking.

molbrown (approved):
👍 This makes sense to me. I think including the `BUFFER_TIME_IN_SECS` buffer to cause a refresh login after just under 24 hours is a good idea.

I get passing tests.

PR #4463, opened 2022-06-17
Worldpay: Fix nil AVS code responses

javierpedrozaing:
Description
-------------------------
Extend the AVS code Map AND modify the response for the MultiResponse
Auth + Caputre when we are performing a purchase in order to return
the avs_code properly

[GWI-257](https://spreedly.atlassian.net/browse/GWI-257)

Unit test
-------------------------
Finished in 0.662202 seconds.

103 tests, 608 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

155.54 tests/s, 918.15 assertions/s

Remote test
-------------------------
Finished in 71.391896 seconds.

99 tests, 412 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop
-------------------------
743 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -45,6 +45,7 @@ module ActiveMerchant #:nodoc:
 
       AVS_CODE_MAP = {
         'A' => 'M', # Match
+        'APPROVED' => 'M', # Match
         'B' => 'P', # Postcode matches, address not verified
         'C' => 'Z', # Postcode matches, address does not match
         'D' => 'B', # Address matched; postcode not checked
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -71,7 +72,13 @@ def initialize(options = {})
...
           r.process { authorize(money, payment_method, options) }
-          r.process { capture(money, r.authorization, options.merge(authorization_validated: true)) } unless options[:skip_capture]
+          unless options[:skip_capture]
+            r.process do
+              response = capture(money, r.authorization, options.merge(authorization_validated: true))
+              response.params['avs_result'] = r.avs_result
+              response
+            end
+          end
         end
```
mbreenlyles, line 82: This looks great and unit tests are passing (I haven't been able to get the remote tests yet due to a slow connection 😅 ) I think a test to confirm the `avs_result` from the authorize is included on the response after the capture could be good to add too :)

curiousepic (changes requested):
As mentioned in the ticket (GWI-257), I think the original problem may not be caused by what this proposes to fix. I'd prefer we are able to create a test that reproduces the problem before the fix, if possible.

PR #4468, opened 2022-06-23
Plexo: Fix `success_from` definition

ajawadmirza:
Gateway response has changed to include `authorize` inplace of `approved` in status of some transactions that is causing bug in the `success_from` method.

SER-157

Unit:
5228 tests, 75971 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
17 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -50,6 +50,7 @@
 * Credorax: Update processor response messages [jcreiff] #4466
 * Shift4: add `customer_reference`, `destination_postal_code`, `product_descriptors` fields and core refactoring [ajawadmirza] #4469
 * Paypal Express: Add checkout status to response object [mbreenlyles] #4467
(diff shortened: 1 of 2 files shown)


drkjc (commented):
@ajawadmirza Is there a way to test which status comes back from Plexo? What tests started failing?

ajawadmirza:
> @ajawadmirza Is there a way to test which status comes back from Plexo? What tests started failing?

All successful tests started failing because of status message change. The status has changed from `approved` to `authorized`. Here is the result when we remove **authorize** from _APPROVED_STATUS_

image

drkjc (approved):
@ajawadmirza Thanks! Merged up.

PR #4470, opened 2022-06-23
Shift4: Scrub Security Code

naashton:
Scrub the `value` from the `securityCode` hash

Unit: 13 tests, 65 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 14 tests, 33 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


dsmcclain (approved):
FYI your PR name says `Rapyd` even though we are in `Shift4` land.

Good work on this. The only (debatable) improvement I could come up with is:
```
sub(%r(("securityCode":{"[\w]+":"[\d]+|[\w]+","value":")[\d]*)i, '\1[FILTERED]')
```
...where the additional `|[\w]+` could accommodate word characters in addition to digits in the value of the `indicator` field. Not sure if that is ever a scenario we would encounter considering we are hard-coding that value to `1` but it could make the implementation slightly more robust. 

All tests passing! Free to ship 🚢

ajawadmirza:
> FYI your PR name says `Rapyd` even though we are in `Shift4` land.
> 
> Good work on this. The only (debatable) improvement I could come up with is:
> 
> ```
> sub(%r(("securityCode":{"[\w]+":"[\d]+|[\w]+","value":")[\d]*)i, '\1[FILTERED]')
> ```
> 
> ...where the additional `|[\w]+` could accommodate word characters in addition to digits in the value of the `indicator` field. Not sure if that is ever a scenario we would encounter considering we are hard-coding that value to `1` but it could make the implementation slightly more robust.
> 
> All tests passing! Free to ship 🚢

@dsmcclain  replacing the `\d` with `\w` should work. Here is the line of code:
`gsub(%r(("securityCode":{"[\w]+":"[\w]+","value":")[\d]*)i, '\1[FILTERED]')`

naashton:
I think that's a solid improvement. I'll change the `\d` to `\w` since that covers `[A-Za-z0-9]`.

naashton:
> FYI your PR name says `Rapyd` even though we are in `Shift4` land.

They all start to look the same after a while.

PR #4474, opened 2022-06-24
Shift4: $0 auth

naashton:
Passing `100` integer value ends up literally being $100, since Shift4
takes the `amount` as a float.

SER-160

Unit: 18 tests, 80 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 27 tests, 79 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -56,6 +56,7 @@
 * Rapyd: Un-nest the payment urls [naashton] #4472
 * Paypal Express: Correct naming mistake for accessor [mbreenlyles] #4473
 * GlobalCollect: Enable Google Pay and Apple Pay [gasb150] #4388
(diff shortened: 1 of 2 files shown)


dsmcclain (approved):
All tests pass.

I'm wondering if we still need to `void` the authorization now that it is a zero amount? When I search through the gateway adapters in AM it looks like we are inconsistent with this. Most of the zero amount verifys don't follow with a `void`, but some do.

It would be nice to have a judgement about this question, but on the other hand I don't think the `void` is hurting anything so if you'd rather just leave it as is and move on that's fine with me.

Approved.

PR #4475, opened 2022-06-27
Shift4: Add `store` method & bug fixes

ajawadmirza:
Added `store` implementation along with test cases and fixed amount to pass value in cents. Also added `present` field in card.

SER-158
SER-164
SER-165

Unit:
5233 tests, 76009 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
16 tests, 35 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -159,10 +175,12 @@ def add_card(post, credit_card, options)
           post[:card][:securityCode] = {}
           post[:card][:securityCode][:indicator] = 1
           post[:card][:securityCode][:value] = credit_card.verification_value
+          post[:card][:present] = 'Y'
```
dsmcclain, line 178: Two things:
1. We should default to `'N'` because an ecommerce transaction, even when it involves a credit card, is considered a "card not present" transaction.
2. From the gateway docs: `This should be set appropriately in the initial authorization/sale request. In subsequent requests, this field should be left blank or should not be sent.`

Therefore, I think we should remove this line as well as the line `post[:card][:present] = options[:present]` out of this method and into a method called `add_card_present`, which should only be called on authorize or purchase transactions.

Something like:
```
def add_card_present(post, options)
  post[:card] = {} unless post[:card].present?
 
  post[:card][:present] = options[:present] || 'N'
end
```

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -159,10 +175,12 @@ def add_card(post, credit_card, options)
...
         else
           post[:card] = {} if post[:card].nil?
           post[:card][:token] = {}
           post[:card][:token][:value] = credit_card
+          post[:card][:present] = 'N'
         end
       end
```
dsmcclain, line 185: I think it would be nice to rename the local variable in this method from `credit_card` to `payment` or `payment_method` because the method is for both credit cards and third party tokens.

dsmcclain (changes requested):
Looks good for the most part, but I think we need to make a few changes on the `card.present` logic.

All tests currently passing.

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: Updated as well.

naashton (approved):
🚢

PR #4478, opened 2022-06-30
Rapyd: Customer Object

naashton:
Add additional fields to the `customer_object` for `store` transactions.

The `address_object` and `payment_object` are subfields of the
`customer_object`, and include `phone_number` and `name`, but these
values are also included directly in the `customer_object` hash which
gives the impression that these are duplicated (which they are).

[Customer Object](https://docs.rapyd.net/build-with-rapyd/reference/customer-object#create-customer)
[Address Object](https://docs.rapyd.net/build-with-rapyd/reference/address-object)

SER-135

Unit: 19 tests, 85 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 27 tests, 79 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -61,6 +61,7 @@
 * Shift4: add `store` method, `present` field in card, and to pass amount in cents [ajawadmirza] #4475
 * Shift4: add `3ds2` implementation [ajawadmirza] #4476
 * Shift4: update `success_from` definition to consider response code [ajawadmirza] #4477
+* Rapyd: Customer Object[naashton] #4478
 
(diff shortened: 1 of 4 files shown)


dsmcclain (approved):
All tests passing. Sorta weird that we are sending these fields in multiple places in the request but c'est la vie.

Approved 👍

naashton:
> Sorta weird that we are sending these fields in multiple places in the request but c'est la vie.

I concur 🤷

PR #4479, opened 2022-06-30
Shift4: Verify Endpoint Fix

naashton:
Send verify transactions to the correct endpoint, with the required card
data.

https://myportal.shift4.com/index.cfm?action=development.shift4api#tag/Cards

SER-170

Unit: 16 tests, 85 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 17 tests, 36 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -61,7 +61,8 @@
 * Shift4: add `store` method, `present` field in card, and to pass amount in cents [ajawadmirza] #4475
 * Shift4: add `3ds2` implementation [ajawadmirza] #4476
 * Shift4: update `success_from` definition to consider response code [ajawadmirza] #4477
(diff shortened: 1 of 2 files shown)


dsmcclain (approved):
All tests passing. Nice to see a dedicated endpoint for this action.

There doesn't appear to be a unit test for `verify` on this gateway, which seems not ideal. However there is a remote test which is passing so I consider that sufficient.

Approved!

naashton:
> There doesn't appear to be a unit test for `verify` on this gateway, which seems not ideal. However there is a remote test which is passing so I consider that sufficient.

I'll create a ticket to add this.

PR #4480, opened 2022-06-30
Airwallex: change timestamps to uuids and remove `_setup` suffix from merchant_order_id

dsmcclain:
Airwallex has requested that we default to a uuid instead of a timestamp for these fields.

They have also requested that we remove the `_setup` suffix from the `merchant_order_id` field, as that field does not need to be unique per transaction and will be used by customers during reconciliation.

The `_setup` suffix remains for the `request_id` field, because that field must be unique for every request, so the suffix is useful for differentiating two separate but associated requests (e.g. a `create_payment_intent` followed by a subsequent `sale`).

SER-65

Unit:
5243 tests, 76069 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
Loaded suite test/remote/gateways/remote_airwallex_test
26 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/airwallex.rb
+++ b/lib/active_merchant/billing/gateways/airwallex.rb
@@ -114,15 +114,15 @@ module ActiveMerchant #:nodoc:
       private
 
       def request_id(options)
-        options[:request_id] || generate_timestamp
+        options[:request_id] || generate_uuid
       end
 
       def merchant_order_id(options)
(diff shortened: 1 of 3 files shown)


naashton (approved):
Looks good! 🚢

PR #4482, opened 2022-07-05
Checkoutv2: Add new fields

ajawadmirza:
Added `processing_channel_id`, `authorization_type`, `capture_type`, `exemption`, and `challenge_indicator` fields along with unit and remote tests.

SER-11

Remote:
46 tests, 117 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
5248 tests, 76088 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:


dsmcclain (changes requested):
This looks good for the most part, but there aren't any tests for the `incremental_authorize` functionality.

This is definitely something we want to test, considering that it involves a different endpoint entirely.

All tests currently passing.

ajawadmirza:
> This looks good for the most part, but there aren't any tests for the `incremental_authorize` functionality.
> 
> This is definitely something we want to test, considering that it involves a different endpoint entirely.
> 
> All tests currently passing.

Thanks @dsmcclain , _Incremental Authorization_ transactions are responding with error so I will add them once we get it going 👍

dsmcclain (changes requested):
@ajawadmirza Thank you for the extra info about `incremental authorization`. It's fine if we need to add tests later.

I do think there is a potential issue with this implementation though.

From Spreedly's perspective, the `options` hash will enter the gateway adapter with string values, so if somebody made a request to Spreedly with `incremental_authorization: false` then the value of `options[:incremental_authorization]` would be the string `'false'`.

This would mean that these lines:
```
 return commit(:authorize, post) unless options[:incremental_authorization]
 return commit(:incremental_authorize, post, payment_method) if options[:incremental_authorization]
```
would end up performing an incremental authorize if the value is `'false'`.

I think a safer way to do this would be to do something like:
```
options[:incremental_authorization].to_s.downcase == 'true' ? commit(:incremental_authorize) : commit(:authorize)
```
or, if you prefer:
```
return commit(:incremental_authorize) if options[:incremental_authorization].to_s.downcase == 'true'
commit(:authorize)
```

ajawadmirza:
> @ajawadmirza Thank you for the extra info about `incremental authorization`. It's fine if we need to add tests later.
> 
> I do think there is a potential issue with this implementation though.
> 
> From Spreedly's perspective, the `options` hash will enter the gateway adapter with string values, so if somebody made a request to Spreedly with `incremental_authorization: false` then the value of `options[:incremental_authorization]` would be the string `'false'`.
> 
> This would mean that these lines:
> 
> ```
>  return commit(:authorize, post) unless options[:incremental_authorization]
>  return commit(:incremental_authorize, post, payment_method) if options[:incremental_authorization]
> ```
> 
> would end up performing an incremental authorize if the value is `'false'`.
> 
> I think a safer way to do this would be to do something like:
> 
> ```
> options[:incremental_authorization].to_s.downcase == 'true' ? commit(:incremental_authorize) : commit(:authorize)
> ```
> 
> or, if you prefer:
> 
> ```
> return commit(:incremental_authorize) if options[:incremental_authorization].to_s.downcase == 'true'
> commit(:authorize)
> ```

Thanks @dsmcclain , I have updated this change and also added unit test in support of incremental_authorization. Let me know if any other change is suggested 👍

dsmcclain (approved):
Thank you for making those changes @ajawadmirza! Looks good and all tests passing.

I will rebase and merge shortly.

PR #4484, opened 2022-07-06
Stripe: Add shipping address

jcreiff:
Update the `add_charge_details` method to include an `add_shipping_address`
method, similar to the [Stripe Payment Intents integration](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe_payment_intents.rb#L36). The [Charges API](https://stripe.com/docs/api/charges/create#create_charge-shipping) accepts these fields as well.

CER-35
Remote
Finished in 138.947824 seconds.
75 tests, 342 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit
Finished in 0.098302 seconds.
143 tests, 759 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Local
Finished in 21.406045 seconds.-
5245 tests, 76084 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -389,6 +389,7 @@ module ActiveMerchant #:nodoc:
         end
 
         add_metadata(post, options)
+        add_shipping_address(post, options)
         add_application_fee(post, options)
         add_exchange_rate(post, options)
         add_destination(post, options)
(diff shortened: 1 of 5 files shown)


test/remote/gateways/remote_stripe_test.rb
```diff
@@ -622,15 +622,15 @@ def test_invalid_login
   # These "track data present" tests fail with invalid expiration dates. The
   # test track data probably needs to be updated.
   def test_card_present_purchase
-    @credit_card.track_data = '%B378282246310005^LONGSON/LONGBOB^2205101130504392?'
+    @credit_card.track_data = '%B378282246310005^LONGSON/LONGBOB^2705101130504392?'
```
jcreiff, line 625: These tests were failing with an "invalid card expiry date" error, so I updated the year from `22` to `27`

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -555,6 +556,22 @@ def add_emv_metadata(post, creditcard)
         post[:metadata][:card_read_method] = creditcard.read_method if creditcard.respond_to?(:read_method)
       end
 
+      def add_shipping_address(post, options = {})
+        return unless shipping = options[:shipping_address]
+
+        post[:shipping] = {}
+
+        post[:shipping][:address] = {}
+        post[:shipping][:name] = shipping[:name] || options[:billing_address][:name]
```
jcreiff, line 565: Stripe requires the `[:shipping][:name]` field to be present, so I included this fallback to the billing address name in case the request to Spreedly includes a shipping address override without a `name`

rachelkirk (approved):
This looks great! I'm approving, but giving you the constructive feedback that you asked for. I think your tests are thorough enough, but if you wanted to be extra thorough you could always add a remote test transaction and add the `shipping_address` via `@options.merge`. This would be a good way to test your theory about the shipping name || billing name.

PR #4487, opened 2022-07-07
Rapyd: Add `stored_credential` support

ajawadmirza:
Added `network_reference_id` to pass in from stored credential framework.

SER-126

Unit:
5245 tests, 76079 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
28 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -149,6 +149,13 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def add_stored_credential(post, options)
+        return unless stored_credential = options[:stored_credential]
+
+        post[:payment_method][:fields][:network_reference_id] = stored_credential[:original_network_transaction_id] if stored_credential[:original_network_transaction_id]
+        post[:initiation_type] = stored_credential[:reason_type] if stored_credential[:reason_type]
+      end
+
       def add_creditcard(post, payment, options)
         post[:payment_method] = {}
         post[:payment_method][:fields] = {}
@@ -158,8 +165,10 @@ module ActiveMerchant #:nodoc:
         pm_fields[:number] = payment.number
         pm_fields[:expiration_month] = payment.month.to_s
         pm_fields[:expiration_year] = payment.year.to_s
-        pm_fields[:cvv] = payment.verification_value.to_s
         pm_fields[:name] = "#{payment.first_name} #{payment.last_name}"
+        pm_fields[:cvv] = payment.verification_value.to_s
+
+        add_stored_credential(post, options)
       end
 
       def add_ach(post, payment, options)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -73,6 +73,7 @@
 * CheckoutV2: Add support for transactions through OAuth [ajawadmirza] #4483
 * Vanco: Update unit test to remove remote call to gateway [ajawadmirza] #4497
 * Shift4: remove support for 3ds2 [ajawadmirza] #4503
+* Rapyd: Add support for stored credential [ajawadmirza] #4487
 
 == Version 1.126.0 (April 15th, 2022)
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -158,8 +160,14 @@ def add_creditcard(post, payment, options)
         pm_fields[:number] = payment.number
         pm_fields[:expiration_month] = payment.month.to_s
         pm_fields[:expiration_year] = payment.year.to_s
-        pm_fields[:cvv] = payment.verification_value.to_s
         pm_fields[:name] = "#{payment.first_name} #{payment.last_name}"
+
+        if stored_credential
+          pm_fields[:recurrence_type] = stored_credential[:reason_type] if stored_credential[:reason_type]
```
drkjc, line 166: I can't find `recurrence_type` in the Rapyd docs. Can you provide a link? [I do see an `initiation_type` field but it lives outside the `payment_method` object](https://docs.rapyd.net/build-with-rapyd/reference/payment-object#payment-method-data-object). Are either of these fields required? I don't really follow the example provided in the Jira description which shows `"name": "recurrence_type",`.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -158,8 +160,14 @@ def add_creditcard(post, payment, options)
...
+          pm_fields[:network_reference_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
```
drkjc, line 167: We should be populating `network_reference_id` with `original_network_transaction_id` instead of `network_transaction_id` since we update the latter on each subsequent transaction.

> Note: For all subsequent payments in the subscription, you must use the network reference ID from the first payment. The response in each subsequent payment contains a different network reference ID, which is not for use in requests.

drkjc (changes requested):
@ajawadmirza I understand wanting to include stored credential logic in the `add_creditcard` method since it correlates to the `cvv`, but it's my opinion that you should move it into its own method to stay in line with convention. For example:

```
def add_stored_credential(pm_fields, options)
        return unless stored_credential = options[:stored_credential]

        pm_fields[:recurrence_type] = stored_credential[:reason_type] if stored_credential[:reason_type]
        pm_fields[:network_reference_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
      end
```

And then inside of `add_creditcard`
```
pm_fields[:cvv] = payment.verification_value.to_s unless add_stored_credential(pm_fields, options)
```

What do you think? I agree with you that if since the `stored_credential` tests are passing that once approved I'll go ahead merge while we wait to hear back from Rapyd about our configuration.

lib/active_merchant/billing/gateways/rapyd.rb
ajawadmirza: I used the field from jira description but I think this might have been mistyped as it should be `initiation_type` as per the documentation, so I have updated it to the correct format. Thanks

lib/active_merchant/billing/gateways/rapyd.rb
ajawadmirza: change is updated 👍

drkjc (approved):
Looks great! Thanks for making those changes. The tests related to Stored Credentials are passing when using our SC specific credentials so I'm going to merge. 

15 tests fail regardless of credentials with this error:

> "The request attempted an operation that requires a payment method, but the payment method type specified is not available for this merchant or does not exist at all. Corrective action: Use a payment method type that this merchant is authorized to use."

28 tests, 52 assertions, 15 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
46.4286% passed

PR #4492, opened 2022-07-11
CheckoutV2: Add `3ds.status` field to send status of 3DS flow of all 3DS transactions

BritneyS:
Test Summary
Local: 5245 tests, 76077 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 41 tests, 238 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 44 tests, 113 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


mbreenlyles (approved):
Looks great and the tests pass! Thanks for adding this! ❤️

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -88,7 +89,8 @@ def setup
         version: '2.0.0',
         eci: '06',
         cavv: 'AgAAAAAAAIR8CQrXcIhbQAAAAAA',
-        ds_transaction_id: 'MDAwMDAwMDAwMDAwMDAwMzIyNzY='
+        ds_transaction_id: 'MDAwMDAwMDAwMDAwMDAwMzIyNzY=',
+        authentication_response_status: 'Y'
```
dsmcclain, line 93: Out of pure morbid curiosity I tried these remote tests with `authentication_response_status: 'mwahahahaha'` and everything still passed. Not sure that really tells us much except that this value many not mean anything to Checkout's sandbox.

dsmcclain (approved):
This looks right to me. Thank you for the thorough paper-trail in the ticket, that helped me connect the dots during my review.

Worth noting: I'm not sure what the possible values of this field are (seems like at the very least `'Y'`, `'A'`, and `'C'` are possible), but I attempted some remote tests with weird values and everything still passed. This might not indicate anything other than a limitation in Checkout's sandbox, but it seems like we won't really know the efficacy of this change until the customer is using it in production, unfortunately.

All tests passing.

test/remote/gateways/remote_checkout_v2_test.rb
BritneyS: Yep, this is more of a negative check on that field. Checkout doesn't provide us MADA cards to test against, otherwise we could write more tests in which adding this field would be more applicable.

PR #4493, opened 2022-07-12
Worldpay: update logic for AM generated response message

rachelkirk:
CER-18

This PR updates the logic in the `required_status_message` and `message_from` methods. It should prevent the AM generated response message from showing up in auth or purchase (where it does not belong) and also adds `issuer_response_description` as an option for the response message.  I was not able to replicate the AM generated failure response message with remote tests, and the IssuerResponseCode (and it’s attribute description) are not returned in the sandbox.

These changes have exposed some flawed logic for 3 of the Remote 3DS Auth tests. I've left comments on them with further explanation for why I've commented some lines out.

A unit test has been included to show that in the absence of a useful response from the gateway, the issuer/bank response can be mapped to `response.message`.

Unit Tests:
104 tests, 613 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote Tests:
99 tests, 405 assertions, 0 failures, 1 errors, 0 pendings, 0 omissions, 0 notifications
98.9899% passed
Note: This error is also occurring on master and seems to be unrelated to my changes.

Local Tests:
5253 tests, 76107 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_worldpay_test.rb
+++ b/test/remote/gateways/remote_worldpay_test.rb
@@ -440,7 +440,7 @@ class RemoteWorldpayTest < Test::Unit::TestCase
       }
     )
     assert first_message = @gateway.authorize(@amount, @threeDS_card, options)
-    assert_equal "A transaction status of 'AUTHORISED' or 'CAPTURED' is required.", first_message.message
+    # assert_equal "A transaction status of 'AUTHORISED' or 'CAPTURED' is required.", first_message.message
     assert first_message.test?
     refute first_message.authorization.blank?
     refute first_message.params['issuer_url'].blank?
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -440,7 +440,7 @@ def test_successful_authorize_with_3ds
       }
     )
     assert first_message = @gateway.authorize(@amount, @threeDS_card, options)
-    assert_equal "A transaction status of 'AUTHORISED' or 'CAPTURED' is required.", first_message.message
+    # assert_equal "A transaction status of 'AUTHORISED' or 'CAPTURED' is required.", first_message.message
```
rachelkirk, line 443: After some debugging, it seems these tests were only passing through the presence of the AM created response message, which in itself is indicating that the transaction wasn't successful. For all three of these tests `first_message.success? => false`. I have commented out these lines in the tests so that they will not begin to "fail" because of my changes, but the underlying problem whatever it may be, is not related to the current task for this ticket.

therufs (approved):
LGTM and tests behave as described!

PR #4495, opened 2022-07-14
Sage Pay: allow sequential reference purchases

therufs:
This modifies repeat purchase handling so that the most recent reference
transaction, rather than the original, may be used to make another
reference purchase.

Remote:
37 tests, 107 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
38 tests, 128 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

CER-17


rachelkirk (approved):
Tests pass, LGTM! Thank you for the more in depth explanation ✨

PR #4497, opened 2022-07-15
Vanco: Fix Unit test to remove remote call

ajawadmirza:
Fixed unit test case `test_successful_purchase_with_existing_session_id` to use mock response and avoid remote call on the gateway.

Unit:
100% passed

Rubocop:


dsmcclain (changes requested):
Good catch on this @ajawadmirza.

Unfortunately, when I run the unit tests on this branch I get a failure for `test_successful_purchase_with_existing_session_id` at line 70: `assert_success response`

I think it might be because of this line: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/vanco.rb#L29

It looks like you will need to dynamically update the `Auth.RequestTime` field of your stubbed login response to be less than one day old.

Also, even though it might not be necessary for a passing test, I think the `login_request_response` stub should probably be a successful response. Currently it shows

```
<Errors>
  <Error>
    <ErrorCode>378</ErrorCode>
    <ErrorDescription>Invalid Login Key</ErrorDescription>
  </Error>
</Errors>
```

ajawadmirza:
Thanks @dsmcclain I think there was an issue with time conversion so I just updated the response along with passing the current time so its always less than 24 hours in the response.

dsmcclain (approved):
Excellent, nicely done.

All tests passing, I will merge.

PR #4502, opened 2022-07-19
Orbital: Remove `DPANInd` for RC transactions

ajawadmirza:
Update condition not to pass `DPANInd` when using RC Transactions.

SER-168

Unit:
100% passed

Rubocop:

Remote:
119 tests, 497 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -86,6 +86,7 @@
 * Shift4: Scrub `securityCode` fix [naashton] #4524
 * Credorax: Update `OpCode` for credit transactions [dsmcclain] #4279
 * CheckoutV2: Add `credit` method [ajawadmirza] #4490
+* Orbital: Remove `DPANInd` field for RC transactions [ajawadmirza] #4502
 
 == Version 1.126.0 (April 15th, 2022)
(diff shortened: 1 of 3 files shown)


drkjc (commented):
@ajawadmirza Everything looks good, but before I merge can you explain how this is related to the Jira for me? Just so I understand what's going on. Can we write a remote test for this change, or is that happening through Chase Orbital? 

Also, can you add a CHANGELOG entry? Thanks

ajawadmirza:
> @ajawadmirza Everything looks good, but before I merge can you explain how this is related to the Jira for me? Just so I understand what's going on. Can we write a remote test for this change, or is that happening through Chase Orbital?
> 
> Also, can you add a CHANGELOG entry? Thanks

when running a test script for GP/AP on current implementation of orbital (shared on jira), you will be getting an error to remove `DPANInd` for RC transaction. So this change is added to make sure DPANInd is not sent when performing RC transaction as per the test cases.

drkjc (approved):
Thanks for the clarification. All test pass. Merging now.

PR #4507, opened 2022-07-20
MerchantE: Update `store` and add `verify` functionality.

ajawadmirza:
Added verify method and updated store to get permanent tokens along with unit and remote test cases.

SER-185
SER-183

Remote:
19 tests, 81 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
5266 tests, 76147 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/merchant_e_solutions.rb
+++ b/lib/active_merchant/billing/gateways/merchant_e_solutions.rb
@@ -18,6 +18,8 @@ module ActiveMerchant #:nodoc:
       # The name of the gateway
       self.display_name = 'Merchant e-Solutions'
 
+      SUCCESS_RESPONSE_CODES = %w(000 085)
+
       def initialize(options = {})
         requires!(options, :login, :password)
         super
@@ -55,10 +57,10 @@ module ActiveMerchant #:nodoc:
       end
 
       def store(creditcard, options = {})
-        post = {}
-        post[:client_reference_number] = options[:customer] if options.has_key?(:customer)
-        add_creditcard(post, creditcard, options)
-        commit('T', nil, post)
+        MultiResponse.run do |r|
+          r.process { temporary_store(creditcard, options) }
+          r.process { verify(r.authorization, { store_card: 'y' }) }
+        end
       end
 
       def unstore(card_id, options = {})
@@ -94,6 +96,13 @@ module ActiveMerchant #:nodoc:
         commit('V', nil, options.merge(post))
       end
 
+      def verify(credit_card, options = {})
+        post = {}
+        post[:store_card] = options[:store_card] if options[:store_card]
+        add_payment_source(post, credit_card, options)
+        commit('A', 0, post)
+      end
+
       def supports_scrubbing?
         true
       end
@@ -106,6 +115,13 @@ module ActiveMerchant #:nodoc:
       end
 
       private
+
+      def temporary_store(creditcard, options = {})
+        post = {}
+        post[:client_reference_number] = options[:customer] if options.has_key?(:customer)
+        add_creditcard(post, creditcard, options)
+        commit('T', nil, post)
+      end
 
       def add_address(post, options)
         if address = options[:billing_address] || options[:address]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/merchant_e_solutions.rb
```diff
@@ -55,10 +57,11 @@ def capture(money, transaction_id, options = {})
       end
 
       def store(creditcard, options = {})
-        post = {}
-        post[:client_reference_number] = options[:customer] if options.has_key?(:customer)
-        add_creditcard(post, creditcard, options)
-        commit('T', nil, post)
+        response = MultiResponse.run do |r|
+          r.process { temporary_store(creditcard, options) }
+          r.process { verify(creditcard, { store_card: 'y' }) }
```
dsmcclain, line 62: I think the idea here is that `temporary_store` will create a temporary token at the gateway, which will be the `transaction_id` in the gateway's response.

THEN we need to grab the temporary token from the store response and pass it in the `verify` call, instead of `creditcard`. The result is that the gateway will accept the temporary token and return a permanent token in `card_id` field.

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
```diff
@@ -55,10 +57,11 @@ def capture(money, transaction_id, options = {})
...
+        end
+        response.responses.last
```
dsmcclain, line 64: I am pretty sure that MultiResponse uses the last response by default (because `use_first_response` [defaults to false](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/response.rb#L49)).

So if you take away the assignment to the `response` variable then `MultiResponse.run` should just return the last response in the process.

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
```diff
@@ -94,6 +97,14 @@ def void(transaction_id, options = {})
         commit('V', nil, options.merge(post))
       end
 
+      def verify(credit_card, options = {})
+        post = {}
+        money = options[:verify_amount].to_i || 0
```
dsmcclain, line 102: [Gateway documentation ](https://anypoint.mulesoft.com/exchange/portals/merchant-e-solutions-3/062a2db4-e8c3-4086-929c-e8e59f8a6ff7/mes-gateway-sapi_v1/minor/0.0/console/documentation/%23145/)says the `transaction_amount` value on verification should always be 0, so I don't think we want to give the option of passing a `verify_amount` here.

test/remote/gateways/remote_merchant_e_solutions_test.rb
```diff
@@ -200,6 +201,12 @@ def test_successful_purchase_with_3dsecure_params
     assert_equal 'This transaction has been approved', response.message
   end
 
+  def test_successful_verify
```
dsmcclain, line 204: Why no tests for `store`?

dsmcclain (changes requested):
This one is kind of tricky. I think there is a missing piece here, which is that we need to retrieve the `transaction_id` from the `temporary_store` call and pass it as the payment method in the `verify` call.

I also think we are missing some tests for `store`.

test/remote/gateways/remote_merchant_e_solutions_test.rb
ajawadmirza: store is already being tested with complete flow from store to unstore operations here: https://github.com/activemerchant/active_merchant/pull/4507/files#diff-050b74195cc97ae48f83123bfd24c42b3f5354155574bda9841bebb728c76f8fL71

should I write any individual test for store as well?

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
ajawadmirza: Thanks, I updated the method to pass `r.authorization`.

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
ajawadmirza: you're right, it responds with the last response so removing this unnecessary line.

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
ajawadmirza: yes that makes more sense. removing the `verify_amount` from options.

dsmcclain (approved):
Good work @ajawadmirza, I think this is ready. Thank you for making some changes.

All tests are passing so I will merge this.

PR #4509, opened 2022-07-21
Shift4: Revamp changes and add fields

ajawadmirza:
Add some required fields and revamp changes to rectify format along with testing for shift4 implementation.

SER-189

Unit:
5267 tests, 76150 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
17 tests, 36 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4.rb
+++ b/lib/active_merchant/billing/gateways/shift4.rb
@@ -148,7 +148,7 @@ module ActiveMerchant #:nodoc:
 
       def add_clerk(post, options)
         post[:clerk] = {}
-        post[:clerk][:numericId] = options[:clerk_id]
+        post[:clerk][:numericId] = options[:clerk_id] || '1'
       end
 
       def add_invoice(post, money, options)
@@ -158,7 +158,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_datetime(post, options)
-        post[:dateTime] = options[:date_time] || current_date_time
+        post[:dateTime] = options[:date_time] || current_date_time(options)
       end
 
       def add_transaction(post, options)
@@ -176,7 +176,6 @@ module ActiveMerchant #:nodoc:
         if payment_method.is_a?(CreditCard)
           post[:card][:expirationDate] = "#{format(payment_method.month, :two_digits)}#{format(payment_method.year, :two_digits)}"
           post[:card][:number] = payment_method.number
-          post[:card][:entryMode] = options[:entry_mode]
           post[:card][:securityCode] = {}
           post[:card][:securityCode][:indicator] = 1
           post[:card][:securityCode][:value] = payment_method.verification_value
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -301,8 +302,9 @@ def error(response)
         response['result'].first['error']
       end
 
-      def current_date_time
-        DateTime.now.strftime('%Y-%m-%dT%H:%M:%S.%N+%H:%M')
+      def current_date_time(options = {})
+        time_zone = options[:merchant_time_zone] || 'Pacific Time (US & Canada)'
+        Time.now.in_time_zone(time_zone).strftime('%Y-%m-%dT%H:%M:%S.%3N+%H:%M')
```
dsmcclain, line 307: Good work on the `strftime` format, but is the `options[:merchant_time_zone]` something that the client has requested?

I think we should avoid exposing this option unless it is something that is specifically being asked for. The Rails method `in_time_zone` is just one of many ways to specify a time zone, and if we were going to implement this I think we would want to provide a list of acceptable values to pass, because if merchants start passing values like `'PDT'` or `'-06:00'` then we will get an error here.

I think it would be best to just let the customer pass their own timestamp in `options[:date_time]` and if they don't do that, they get whatever DateTime.now we provide.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -285,9 +284,11 @@ def request_headers(options)
         headers = {
           'Content-Type' => 'application/x-www-form-urlencoded'
         }
-        headers['CompanyName'] = options[:company_name]
         headers['AccessToken'] = @access_token
         headers['Invoice'] = options[:invoice] if options[:invoice].present?
+        headers['InterfaceVersion'] = options[:interface_version] if options[:interface_version]
+        headers['InterfaceName'] = options[:interface_name] if options[:interface_name]
+        headers['CompanyName'] = options[:company_name] if options[:company_name]
```
dsmcclain, line 291: The conversation we had about the `AccessToken` header applies here too.

Net/HTTP will convert all of these headers to sentence-case (e.g. `Companyname`, `Interfacename`, `Interfaceversion`).

You don't need to change anything with these headers right now. We can just add them this way and wait to hear back from Shift4 about whether they will be able to loosen their requirements about the case of the headers.

dsmcclain (changes requested):
I would like to remove the `merchant_time_zone` option unless we absolutely need it (see comment below).

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: I think it is being asked to provide time in the merchant zone (not from where request is being made) and we have some documented time zones that could be used:
https://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html

dsmcclain: Okay. Let's make sure we document this well and include the rails link you provided in our documentation.

I am debating about rescuing `ArgumentError` here in cases where the merchant passes an invalid value, but the downside of this would be that we end up silently substituting a default value that is probably not the merchant's actual time zone. So I suppose we can just leave it how you have it and hopefully document it well enough that it doesn't cause any issues.

ajawadmirza: Sounds good! I will update the documentation as well 👍

dsmcclain (approved):
All tests pass. We will move forward with `merchant_time_zone` as well as the PascalCase headers that ruby will eventually change case.

I will merge 👍

PR #4513, opened 2022-07-26
Add machine cookie to subsequent calls during 3DS challenge

mbreenlyles:
This change ensures that the machine cookie is extracted from the response correctly and returned in the initial 3DS setup request will be included in the headers on all the subsequent requests in the Worldpay challenge flow.

There are 3 calls total for a purchase using Worldpay's 3ds:
1. The preauth/setup request: sent to WP as an `authorize`, this one won't need a cookie as it's the first request. We'll receive a response that requests 3DS authentication, and contains a cookie. Then there is a redirect to the issuer's challenge form. 
2. The "3ds order": another redirect will happen after the user submits the form, and we get the authentication results from issuer. Then we'll send Worldpay the "3ds order" containing the authentication results in the `paResponse` field. This one must have the cookie in its header. The response will contain another cookie.
3. The capture call: If the "3ds order" call gets a response of `AUTHORISED`, then we'll complete the purchase with a capture call, and because this capture follows a response that contained another cookie from the 3ds flow, we should send that cookie in the header of the capture call as well. 

**Cookie saved in instance variable** 
Both the "3ds order" and capture calls that happen after a user completes a challenge on their purchase are made "by Core", but within the AM backend. In the "3ds order" call, following the redirect the cookie had been saved in the params from the setup response, and is already sent back successfully in the options hash. For purchases, the capture call had not been sending the cookie back, but the capture takes place right after the "3ds order" response, and it will have access to the instance variable cookie that is saved from that response. There will be a follow-up Core PR that adds testing and sends the cookie through the options hash, but this AM change will still ensure the cookie gets added to headers on the capture call by using the @cookie instance var which will contain the cookie saved from the `3ds order` response.

`authorize` (preauth, no cookie) --> *challenge submit, redirect* --> `3ds order` (pass auth results to wp, cookie already saved) --> `capture` (cookie has now been added)

**Cookie is saved as all characters before the first semi-colon**
Per Worldpay's docs, the cookie we must send back should not be the full string given in the Set-Cookie header, it should be all text preceding the first semi-colon.
From their docs:

> For example, if we send a cookie as:
> 
> Set-Cookie: machine=0aa20016;Secure;path=/
> 
> You must include the following HTTP header in your second order message: Cookie: machine=0aa20016

Local:
100% passed

Unit:
103 tests, 608 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
1 tests, 4 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_worldpay_test.rb
+++ b/test/remote/gateways/remote_worldpay_test.rb
@@ -24,6 +24,7 @@ class RemoteWorldpayTest < Test::Unit::TestCase
     @declined_card = credit_card('4111111111111111', first_name: nil, last_name: 'REFUSED')
     @threeDS_card = credit_card('4111111111111111', first_name: nil, last_name: 'doot')
     @threeDS2_card = credit_card('4111111111111111', first_name: nil, last_name: '3DS_V2_FRICTIONLESS_IDENTIFIED')
+    @threeDS2_challenge_card = credit_card('4000000000001091', first_name: nil, last_name: 'challenge-me-plz')
     @threeDS_card_external_MPI = credit_card('4444333322221111', first_name: 'AA', last_name: 'BD')
     @nt_credit_card = network_tokenization_credit_card('4895370015293175',
       brand: 'visa',
@@ -456,6 +457,27 @@ class RemoteWorldpayTest < Test::Unit::TestCase
     assert_equal 'SUCCESS', response.message
   end
 
+  def test_successful_authorize_with_3ds2_challenge
+    session_id = generate_unique_id
+    options = @options.merge(
+      {
+        execute_threed: true,
+        accept_header: 'text/html',
+        user_agent: 'Mozilla/5.0',
+        session_id: session_id,
+        ip: '127.0.0.1'
+      }
+    )
+    assert response = @gateway.authorize(@amount, @threeDS2_challenge_card, options)
+    assert_equal "A transaction status of 'AUTHORISED' or 'CAPTURED' is required.", response.message
+    assert response.test?
+    refute response.authorization.blank?
+    refute response.params['issuer_url'].blank?
+    refute response.params['pa_request'].blank?
+    refute response.params['cookie'].blank?
+    refute response.params['session_id'].blank?
+  end
+
   def test_successful_auth_and_capture_with_normalized_stored_credential
     stored_credential_params = {
       initial_transaction: true,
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -832,7 +854,7 @@ def test_successful_fast_fund_credit_with_token_on_cft_gateway
     assert_success store
 
     options = @options.merge({ fast_fund_credit: true })
-    assert credit = @gateway.credit(@amount, store.authorization, options)
+    assert credit = @cftgateway.credit(@amount, store.authorization, options)
```
aenand, line 857: Why did the gateway change here?

test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -456,6 +457,27 @@ def test_marking_3ds_purchase_as_moto
     assert_equal 'SUCCESS', response.message
   end
 
+  def test_successful_authorize_with_3ds2_challenge
```
aenand, line 460: Does this test case cover the scenario of multiple requests?

test/remote/gateways/remote_worldpay_test.rb
mbreenlyles: Hmm that got in there as I was trying to investigate why all the tests were failing on the cft gateway. That test will not pass unless one or both gateways used to make the store/credit call are set to the CFT gateway, and it seems to make sense because that test appears to be trying to test the fast fund credit action on the CFT gateway using a stored PM's authorization token. After some reading, the

test/remote/gateways/remote_worldpay_test.rb
mbreenlyles: Good question, I was planning to add a more complete test for that to the Core side since it's easier to simulate the multiple requests. I'm not sure how to simulate those in AM because some of that is handled on the Core side, and I couldn't find any examples of a complete challenge flow in AM tests 😅

aenand: Gotcha, I'm a little lost on the use of `@cookie`. I didn't realize things persisted in AM. Does using `@cookie` mean that all subsequent calls to AM for a given transactions get access to it?

BritneyS: I have the same question ^^

mbreenlyles: No, you're right, there will need to be a corresponding Core change that saves each cookie. After some more testing/docs peering, I'll need to change this to reset the cookie each time a new cookie is sent back and parse out part of the string before saving. This is more complicated than I thought at first 😓  Will push some updates later today!

mbreenlyles: updated the top comment with some helpful context!

aenand (approved):
Thank you for the explanation! I'm still a little confused by this but I think I understand the path forward now. LGTM 🚀

PR #4515, opened 2022-07-27
Braintree Blue: Partial capture

aenand:
Braintree blue can support partial captures for certain
types of transactions. This commit adds logic to handle partial capture
logic if the `partial_capture` field is passed down.

Test Summary
Local:
5269 tests, 76156 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Unit:
92 tests, 204 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
98 tests, 530 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -77,6 +77,7 @@
 * Rapyd: Add support for stored credential [ajawadmirza] #4487
 * MerchantE: Update `store` and add `verify` method [ajawadmirza] #4507
 * Shift4: Add default `numericId`, add `InterfaceVersion`, `InterfaceName`, and `CompanyName` header fields, change date time format and allow merchant time zone [ajawadmirza] #4509
+* BraintreeBlue: Add support for partial capture [aenand] #4515
 
 == Version 1.126.0 (April 15th, 2022)
 * Moneris: Add 3DS MPI field support [esmitperez] #4373
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -88,9 +88,16 @@ module ActiveMerchant #:nodoc:
       end
 
       def capture(money, authorization, options = {})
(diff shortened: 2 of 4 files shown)


tatsianaclifton (changes requested):
It looks good, but I think we should add remote tests for partial capture.

aenand:
@tatsianaclifton I didn't add a remote test for partial capture because the Braintree docs mentioned it is only available for Paypal and Venmo transactions and I don't think those auths are added in AM yet. Are they in AM?

https://developer.paypal.com/braintree/docs/reference/request/transaction/submit-for-partial-settlement

tatsianaclifton:
> @tatsianaclifton I didn't add a remote test for partial capture because the Braintree docs mentioned it is only available for Paypal and Venmo transactions and I don't think those auths are added in AM yet. Are they in AM?
> 
> https://developer.paypal.com/braintree/docs/reference/request/transaction/submit-for-partial-settlement

PayPal and Venmo should work for `purchase` and since `purchase` just calls `authorization` I would assume nothing needs to be modified for `authorization` to accept PayPal and Venmo. However, I don't see any remote tests added to that, not sure what the reason. Maybe Braintree sandbox doesn't allow testing for that.

aenand:
@tatsianaclifton good to know! I'll look into that later and see if it works

molbrown:
I think all you would need is a Venmo/ Paypal nonce? Those are available here for testing: https://developer.paypal.com/braintree/docs/reference/general/testing#nonces-representing-alternative-payment-methods

aenand:
@tatsianaclifton @molbrown thanks for those resources! I was able to add a successful partial capture remote test.

tatsianaclifton (approved):
LGTM

PR #4517, opened 2022-07-28
Priority 

priorityspreedly:
Add general credit

### DIFF
--- a/lib/active_merchant/billing/gateways/priority.rb
+++ b/lib/active_merchant/billing/gateways/priority.rb
@@ -72,6 +72,18 @@ module ActiveMerchant #:nodoc:
         commit('purchase', params: params)
       end
 
+      def credit(amount, credit_card, options = {})
+        params = {}
+        params['authOnly'] = false
+        params['isSettleFunds'] = true
+        amount = -amount
+
+        add_merchant_id(params)
+        add_amount(params, amount, options)
+        add_credit_params(params, credit_card, options)
+        commit('credit', params: params)
+      end
+
       def refund(amount, authorization, options = {})
         params = {}
         add_merchant_id(params)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/priority.rb
```diff
@@ -72,6 +72,17 @@ def authorize(amount, credit_card, options = {})
         commit('purchase', params: params)
       end
 
+      def credit(amount, credit_card, options = {})
```
rachelkirk, line 75: Thank you for making a new branch/PR. I should have asked for this sooner, but can you link to the gateway docs where it is referencing `credit`? I tried looking [here](https://developer.mxmerchant.com/docs) but could not find documentation for this transaction type.  I ask because of the question about a negative amount for credit. I've done some investigating and can't find that to be true for other gateways. Thank you!

rachelkirk (changes requested):

test/remote/gateways/remote_priority_test.rb
```diff
@@ -5,6 +5,7 @@ def setup
     @gateway = PriorityGateway.new(fixtures(:priority))
 
     @amount = 2
+    @credit_amount = -2000
```
rachelkirk, line 8: Thank you for sending the documentation on the negative amount. All amounts sent through to the Active Merchant gateway from us must be sent as a positive amount even for refunds, credits, etc. You will need to add logic to handle that in the gateway file if the gateway expects a negative amount before we can approve.

rachelkirk (changes requested):
Thanks for sending the link to your documentation. Please see the note about amounts, you'll need to add handling for that before we can approve. Thank you!

test/remote/gateways/remote_priority_test.rb
priorityspreedly: made the change so the gateway handles the negative amount

rachelkirk (approved):
Thanks for making these changes, I'm approving and they should be live in production in the next day. (We have a multi deploy process and try to batch other deploys together)

PR #4518, opened 2022-07-29
CheckoutV2: Update logic to send payment id via options

ajawadmirza:
Updated checkoutv2 to take payment id from options parameters instead of getting it as a payment method.

SER-201

Remote:
72 tests, 186 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
5269 tests, 76156 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -40,7 +40,7 @@ module ActiveMerchant #:nodoc:
         post[:capture] = false
         build_auth_or_purchase(post, amount, payment_method, options)
 
-        options[:incremental_authorization].to_s.casecmp('true').zero? ? commit(:incremental_authorize, post, payment_method) : commit(:authorize, post)
+        options[:incremental_authorization] ? commit(:incremental_authorize, post, options[:incremental_authorization]) : commit(:authorize, post)
       end
 
       def capture(amount, authorization, options = {})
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -40,7 +40,7 @@ def authorize(amount, payment_method, options = {})
         post[:capture] = false
         build_auth_or_purchase(post, amount, payment_method, options)
 
-        options[:incremental_authorization].to_s.casecmp('true').zero? ? commit(:incremental_authorize, post, payment_method) : commit(:authorize, post)
+        options[:incremental_authorization].to_s.casecmp('true').zero? ? commit(:incremental_authorize, post, options[:payment_id]) : commit(:authorize, post)
```
dsmcclain, line 43: Do we need to use two fields to signify an incremental authorization, or can't we just have a single field called `incremental_authorization`, or maybe `incremental_authorization_id`, and then the presence or absence of a value will indicate whether the customer wishes to perform an incremental authorization:
```
options[:incremental_authorization] ? commit(:incremental_authorize, post, options[:incremental_authorization] : commit(:authorize, post)
```

dsmcclain (changes requested):
@ajawadmirza Thank  you for making this change! I think we can still make it easier for the customer by consolidating the behavior into a single field, instead of two.

lib/active_merchant/billing/gateways/checkout_v2.rb
ajawadmirza: Thanks @dsmcclain , this one looks cleaner. I have updated the change 👍

naashton (approved):
LGTM!

PR #4520, opened 2022-08-02
Priority

priorityspreedly:
Change the payload to only include cvv and zip info if not null, if it was sent with null information this caused an issue in our side.


rachelkirk (changes requested):
When remote tests are run, `test_failed_purchase_missing_card_verification_number` fails. Please adjust gateway code or test to ensure it is passing. After making the change, please merge these changes into PR 4517 so we can deploy them together. Thank you!

priorityspreedly:
> When remote tests are run, `test_failed_purchase_missing_card_verification_number` fails. Please adjust gateway code or test to ensure it is passing. After making the change, please merge these changes into PR 4517 so we can deploy them together. Thank you!

Made the merge of the changes to PR 4517, also the issue with `test_failed_purchase_missing_card_verification_number ` test failing was a merchant setting not requiring the verification number, we turned on now, can you test it again. It is working now in our side.

PR #4521, opened 2022-08-02
Ingenico: add support for `paymentProductId`

rachelkirk:
CER-59

Adds `payment_product_id` as an option that can be passed as an override to the
existing `BRAND MAP`.

Unit:
43 tests, 219 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
40 tests, 104 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.5% passed
The failing test is also failing on master

Local:
5271 tests, 76163 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -80,6 +80,7 @@
 * BraintreeBlue: Add support for partial capture [aenand] #4515
 * Rapyd: Change key name to `network_transaction_id` [ajawadmirza] #4514
 * CyberSource: Handle unsupported Network Token brands [heavyblade] #4500
+* Ingenico(Global Collect): Add support for `payment_product_id` [rachelkirk] #4521
 
 == Version 1.126.0 (April 15th, 2022)
 * Moneris: Add 3DS MPI field support [esmitperez] #4373
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -255,8 +255,9 @@ module ActiveMerchant #:nodoc:
         month = format(payment.month, :two_digits)
         expirydate = "#{month}#{year}"
         pre_authorization = options[:pre_authorization] ? 'PRE_AUTHORIZATION' : 'FINAL_AUTHORIZATION'
+        product_id = options[:payment_product_id] || BRAND_MAP[payment.brand]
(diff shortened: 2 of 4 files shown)


jcreiff (changes requested):
I think this looks good! But I do have one thought about the remote test case(s) - 

The test you've written validates an outcome that would have already been true before the change, because a `cabal` card already has a `paymentProductId` of 135.

I think this test would show the new functionality more clearly if the value for `payment_product_id` was the one the customer wants to pass in for debit (114).

Maybe we should also add a test that validates that we still send 135 for Cabal (for example) when no `payment_product_id` is specified, to further validate the logic you added?

Unrelated side note: I was momentarily concerned about what the outcome might be if no `payment_product_id` was specified and it also happened to be a new card type that has not yet been added to the `BRAND_MAP` hash, but it looks like if that's the case, the outcome would be the same as it would have been before your change:

`nil || nil` will resolve to `nil`, which is what we would get anyway if it's an unrecognized `payment.brand`

(Your change actually makes it way easier for customers to solve the issue they would run into with this "new card type" use case, because they won't have to wait on us to update the mapping hash, they can just send the new value themselves!)

rachelkirk:
> I think this test would show the new functionality more clearly if the value for `payment_product_id` was the one the customer wants to pass in for debit (114).
>  I tried going this route, but both the pre-prod and sandbox credentials don't allow this unfortunately. Most likely because we don't have a designated debit test card which the value they are wanting to add is signifying. 
> Maybe we should also add a test that validates that we still send 135 for Cabal (for example) when no `payment_product_id` is specified, to further validate the logic you added?
> I did leave the `payment_product_id` out of options originally, but then added it for visibility. It still shows up in the response and the test passes. I'll push that change up. I chose the cabal card since it has a different value of `135` compared to what the default seems to be, which is `1` for visa. The unit test shows that it is being formatted correctly and checks the request to show that whatever is being sent through is accurate. 

> (Your change actually makes it way easier for customers to solve the issue they would run into with this "new card type" use case, because they won't have to wait on us to update the mapping hash, they can just send the new value themselves!)
> Woohoo!

PR #4522, opened 2022-08-03
Adyen: add network_transaction_id to store call

jcreiff:
This makes it possible to pass the `networkTxReference` field with `store`

CER-91

Unit:
95 tests, 482 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
124 tests, 443 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.5806% passed

One failing test is related to something that might be an outdated assumption (Adyen doesn’t allow recurring transactions with Cabal). The other two are 3DS2-related. All are failing on master.

Local:
5272 tests, 76166 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -993,6 +993,14 @@ class RemoteAdyenTest < Test::Unit::TestCase
     assert_equal 'Success', response.message
   end
 
+  def test_successful_tokenize_only_store_with_ntid
+    assert response = @gateway.store(@credit_card, @options.merge({ tokenize_only: true, network_transaction_id: '858435661128555' }))
+
+    assert_success response
+    assert !response.authorization.split('#')[2].nil?
+    assert_equal 'Success', response.message
+  end
+
   def test_successful_store_with_elo_card
     assert response = @gateway.store(@elo_credit_card, @options)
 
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -993,6 +993,14 @@ def test_successful_tokenize_only_store
     assert_equal 'Success', response.message
   end
 
+  def test_successful_tokenize_only_store_with_ntid
```
jcreiff, line 996: I'm not sure if there are additional assertions that would make sense to add here...from what I could tell, it doesn't seem like Adyen's response really changes with the addition of this field to the request 🤔

rachelkirk: This looks good, I wonder if you were to add the ntid to a different transaction type does it change those responses vs when it's not there? I don't know that that's relevant in terms of THIS test, but would maybe explain why it's not changing in the store response. Not asking you to go back and investigate that, just came to mind.

rachelkirk (approved):
Tests pass (except the three remote test failures that are also failing on master) and looks good!

PR #4524, opened 2022-08-05
Shift4: Scrub `securityCode` fix

naashton:
Scrub method was not properly scrubbing `securityCode` from the
transcript. This PR improves the regex to find and scrub that value from
the transcript.

SER-224

Unit: 17 tests, 90 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 16 tests, 36 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


drkjc (approved):
Nice! 🚢

PR #4530, opened 2022-08-09
Worldpay: Update  `required_status_message` and `message_from` methods

rachelkirk:
CER-18

Updating the `required_status_message` method to only generate the AM response if using one of the actions called out in the method. The transaction types are all multi-response and use a combination of actions.

Remote Tests:
100 tests, 392 assertions, 4 failures, 1 errors, 0 pendings, 0 omissions, 0 notifications
95% passed
*4 failing tests are 3DS tests that hinge on that message being auto-generated. The message being auto-generated means that the transaction itself failed. When prying into the responses, for these tests `@success = false`. The 1 error is also happening on master, may be related to recent Worldpay updates.

Unit Tests:
104 tests, 613 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Local:
5274 tests, 76180 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -98,6 +98,7 @@
 * Rapyd: Pass Customer ID and fix `add_token` method [naashton] #4538
 * Shift4: If no timezone is sent on transactions, the code uses the hours and minutes as a timezone offset [ali-hassan] #4536
 * Priority: Add support for general credit and updating cvv and zip [priorityspreedly] #4517
+* Worldpay: Update actions for generated message in `required_status_message` method [rachelkirk] #4530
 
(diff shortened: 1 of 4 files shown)


jcreiff:
I think I will probably defer to Marketplace Services team for final approval once this PR is not a draft anymore, but should the test expectations be changed for those 4 failing tests, since we've changed the logic of how the response message is created?

therufs (approved):
We have one remote test failing, `test_successful_purchase_with_level_two_and_three_fields`, that is also failing on master. 
 
 lgtm!

PR #4532, opened 2022-08-11
Shift4: Add `expirationDate` to refund transactions

ajawadmirza:
Add `expirationDate` to refund via options along with unit test.

SER-222

Unit:
5275 tests, 76191 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -91,6 +91,7 @@
 * Shift4: Add card `present` field, use previous transaction authorization for capture, and hardcode header values [ajawadmirza] #4528
 * Orbital: Remove `DPANInd` field for RC transactions [ajawadmirza] #4502
 * EBANX: Add Spreedly tag to payment body [flaaviaa] #4527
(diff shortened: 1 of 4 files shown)


drkjc (changes requested):
@ajawadmirza Everything looks good. I see we're going to handle getting the expiration date in Core. Do you know if this is going to be a required field for `refunds`? If so, can you add it as an option in the remote refund test?

ajawadmirza:
> @ajawadmirza Everything looks good. I see we're going to handle getting the expiration date in Core. Do you know if this is going to be a required field for `refunds`? If so, can you add it as an option in the remote refund test?

This field is not a required field as transactions also get succeeded without this but as per the request this should be added in the refund transactions. I have added a remote test for the refund as well.

drkjc (approved):
Thanks for making those changes. Good to merge.

PR #4534, opened 2022-08-15
Airwallex: Add `skip_3ds` field

ajawadmirza:
Added `skip_3ds` field support for airwallex along with unit and remote test cases.

SEr-229

Unit:
5276 tests, 76200 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
27 tests, 63 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/airwallex.rb
+++ b/lib/active_merchant/billing/gateways/airwallex.rb
@@ -154,6 +154,7 @@ module ActiveMerchant #:nodoc:
         post[:merchant_order_id] = merchant_order_id(options)
         add_referrer_data(post)
         add_descriptor(post, options)
+        post['payment_method_options'] = { 'card' => { 'risk_control' => { 'three_ds_action' => 'SKIP_3DS' } } } if options[:skip_3ds]
 
         response = commit(:setup, post)
         raise ArgumentError.new(response.message) unless response.success?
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -154,6 +154,7 @@ def create_payment_intent(money, options = {})
         post[:merchant_order_id] = merchant_order_id(options)
         add_referrer_data(post)
         add_descriptor(post, options)
+        post['payment_method_options'] = { 'card' => { 'risk_control' => { 'three_ds_action' => 'SKIP_3DS' } } } if options[:skip_3ds] && (options[:skip_3ds].to_s.casecmp('true') == 0)
```
naashton, line 157: I think we can simplify the conditional here: `if options[:skip_3ds] && (options[:skip_3ds].to_s.casecmp('true') == 0)`

I say treat it as a `bool` `true` and then we can just implicitly check `if options[:skip_3ds]`

naashton (changes requested):
This will be good to ship after we simplify the conditional.

lib/active_merchant/billing/gateways/airwallex.rb
ajawadmirza: Thanks @naashton, this change is updated 👍

naashton (approved):
👍

PR #4535, opened 2022-08-15
Wompi: allow partial refund amount on void_sync

jcreiff:
Previous changes to this gateway redirected all refund requests to the
void endpoint. This meant that any partial refund attempts were processed
as full refunds, because the void method did not accept an amount.

Wompi has advised that it is now possible to send amount_in_cents to this
void_sync endpoint to allow for partial refunds.

CER-152

LOCAL
5276 tests, 76199 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

UNIT
12 tests, 58 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

REMOTE
13 tests, 34 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/wompi.rb
+++ b/lib/active_merchant/billing/gateways/wompi.rb
@@ -65,11 +65,12 @@ module ActiveMerchant #:nodoc:
         # commit('refund', post, '/refunds_sync')
 
         # All refunds will instead be voided. This is temporary.
-        void(authorization)
+        void(authorization, options, money)
       end
 
-      def void(authorization, options = {})
-        commit('void', {}, "/transactions/#{authorization}/void_sync")
+      def void(authorization, options = {}, money = nil)
+        post = money ? { amount_in_cents: amount(money).to_i } : {}
+        commit('void', post, "/transactions/#{authorization}/void_sync")
       end
 
       def supports_scrubbing?
--- a/test/remote/gateways/remote_wompi_test.rb
+++ b/test/remote/gateways/remote_wompi_test.rb
@@ -74,13 +74,13 @@ class RemoteWompiTest < Test::Unit::TestCase
     assert_success refund
   end
 
-  # def test_partial_refund
-  #   purchase = @gateway.purchase(@amount, @credit_card, @options)
-  #   assert_success purchase
+  def test_partial_refund
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success purchase
 
-  #   assert refund = @gateway.refund(@amount - 1, purchase.authorization)
-  #   assert_success refund
-  # end
+    assert refund = @gateway.refund(@amount - 50000, purchase.authorization)
+    assert_success refund
+  end
 
   # def test_failed_refund
   #   response = @gateway.refund(@amount, '')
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/wompi.rb
```diff
@@ -65,11 +65,12 @@ def refund(money, authorization, options = {})
...
 
         # All refunds will instead be voided. This is temporary.
-        void(authorization)
+        void(authorization, options, money)
       end
 
-      def void(authorization, options = {})
-        commit('void', {}, "/transactions/#{authorization}/void_sync")
+      def void(authorization, options = {}, money = nil)
+        post = money ? { amount_in_cents: amount(money).to_i } : {}
```
jcreiff, line 72: This logic is an attempt to preserve the functionality of an actual request to the void endpoint (rather than a refund redirected to a void)

I did some local testing (including integration tests) and realized that a /void request would break without this default `nil` amount for `money`, because usually no amount argument is expected for a true void

test/unit/gateways/wompi_test.rb
```diff
@@ -106,18 +106,34 @@ def test_successful_refund_to_void
...
+    assert response.test?
+  end
+
+  def test_successful_void
+    response = stub_comms(@gateway) do
+      @gateway.void(@amount, @options)
+    end.check_request do |endpoint, data, _headers|
+      assert_match 'void_sync', endpoint
+      assert_match '{}', data
+    end.respond_with(successful_void_response)
```
jcreiff, line 136: Would appreciate any feedback on these test cases to make sure they are doing what I think they are doing 😁

jcreiff:
@drkjc I tagged you just because I though you might be familiar with this, since you did the initial work to re-direct refund to void. 🙏🏻

test/remote/gateways/remote_wompi_test.rb
```diff
@@ -74,13 +74,13 @@ def test_successful_refund
...
-  #   purchase = @gateway.purchase(@amount, @credit_card, @options)
-  #   assert_success purchase
+  def test_partial_refund
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success purchase
 
-  #   assert refund = @gateway.refund(@amount - 1, purchase.authorization)
-  #   assert_success refund
-  # end
+    assert refund = @gateway.refund(@amount - 50000, purchase.authorization)
```
rachelkirk, line 81: I `pry`ed into this test as well to check the @amount, also not seeing a change. I guess if you wanted to be more explicit with the test, you could add a couple of lines like
`response_data = refund.params['data']`
    `assert_equal response_data.dig('transaction', 'amount_in_cents'), 100000`
But as the gateway is currently behaving that assert_equal will fail.

rachelkirk (approved):
✨ Looks good and tests are passing. I had a couple of comments about the remote tests. When I ran the whole suite some of them failed but when ran individually those failing tests passed. Wompi's sandbox is still unpredictable it seems. I'm approving but I think it's wise to wait to hear back from them about the partial refund amount before you merge.

test/remote/gateways/remote_wompi_test.rb
jcreiff: Thanks @rachelkirk ! I also experienced some intermittent failure on tests that seemed related to connection issues with the sandbox.

I will definitely wait on Wompi's feedback before merging, and if they indicate that we should be seeing the partial amount in their response, I'll dig into that further and also update the test with the `assert_equal` tip you suggested

drkjc (approved):
@jcreiff @rachelkirk Everything looks good! Though, for some reason I can't get the `partial refund` remote test passing. 

I'm consistently getting a bad request error, what errors where you seeing? Seems like a problem on my end so I'm going to approve, just don't forget a CHANGELOG entry.

PR #4538, opened 2022-08-17
Rapyd: Pass Customer ID and fix `add_token` method

naashton:
For `authorize` and `purchase` transactions we can optionally pass the
`customer` object which is similarly used in the transaction endpoint.
The value is prepended `cus_` and is required if the payment method is
omitted

`add_token` was incorrectly sending the whole `authorization` which is a
combination of a transaction or customer id and a payment token (based
on the `authorization_from` method

Remote test for `test_successful_store_and_purchase` includes a
commented out test purchase (commented out) because the payment method
must go through a 3DS authorization flow before it can be utilized in
payment transactions (sandbox only)
[Create Customer Object](https://docs.rapyd.net/build-with-rapyd/reference/customer-object#create-customer)

SER-228

Unit: 20 tests, 91 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -28,6 +28,7 @@ module ActiveMerchant #:nodoc:
         add_ewallet(post, options)
         add_payment_fields(post, options)
         add_payment_urls(post, options)
+        add_customer_id(post, options)
         post[:capture] = true if payment.is_a?(CreditCard)
 
         if payment.is_a?(Check)
@@ -53,6 +54,7 @@ module ActiveMerchant #:nodoc:
         add_ewallet(post, options)
         add_payment_fields(post, options)
         add_payment_urls(post, options)
+        add_customer_id(post, options)
         post[:capture] = false
 
         commit(:post, 'payments', post)
@@ -185,7 +187,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_token(post, payment, options)
-        post[:payment_method] = payment
+        return unless token = payment.split('|')[1]
+
+        post[:payment_method] = token
       end
 
       def add_3ds(post, payment, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -185,7 +187,9 @@ def add_ach(post, payment, options)
       end
 
       def add_token(post, payment, options)
-        post[:payment_method] = payment
+        return unless token = payment.split('|')[1]
+
+        post[:payment_method] = token
```
dsmcclain, line 192: If I'm understanding you correctly, we can't really test this remotely because of limitations with 3ds flow. But it seems like we should still have a unit test showing that a transaction with an authorization of the format `<auth_id>|<payment_method_id>` grabs the correct portion of the string and adds it to the request.

Maybe that seems unnecessary for such straightforward behavior, but it seems like the lack of test coverage is what let this slip through the cracks in the first place.

dsmcclain (changes requested):
Getting a rubocop error that needs to be fixed.

Also made a suggestion for adding a unit test to cover the behavior of `add_token`

test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -204,6 +204,17 @@ def test_failed_verify
     assert_equal 'Do Not Honor', response.message
   end
 
+  def test_successful_store_and_purchase
+    store = @gateway.store(@credit_card, @options)
+    assert_success store
+    assert customer_id = store.params.dig('data', 'id')
```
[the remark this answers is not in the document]
naashton: Doh

lib/active_merchant/billing/gateways/rapyd.rb
naashton: I think this is a reasonable request. I'll add a specific check for this.

test/unit/gateways/rapyd_test.rb
```diff
@@ -7,6 +7,7 @@ def setup
     @gateway = RapydGateway.new(secret_key: 'secret_key', access_key: 'access_key')
     @credit_card = credit_card
     @amount = 100
+    @authorization = 'cus_9e1b5a357b2b7f25f8dd98827fbc4f22|card_cf105df9e77462deb34ffef33c3e3d05'
```
dsmcclain, line 10: nice

dsmcclain (approved):
Looking good, thanks for adding that test! Everything is passing for me.

Don't forget a changelog entry, but otherwise 🚢

PR #4539, opened 2022-08-17
Stripe: Add shipping address

jcreiff:
Update the `add_charge_details` method to include an `add_shipping_address`
method, similar to the Stripe Payment Intents integration.

CER-35
LOCAL
5288 tests, 76234 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

UNIT
146 tests, 765 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

REMOTE

77 tests, 359 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -389,6 +389,7 @@ module ActiveMerchant #:nodoc:
         end
 
         add_metadata(post, options)
+        add_shipping_address(post, payment, options)
         add_application_fee(post, options)
         add_exchange_rate(post, options)
         add_destination(post, options)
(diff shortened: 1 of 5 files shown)


test/remote/gateways/remote_stripe_test.rb
```diff
@@ -622,15 +668,15 @@ def test_invalid_login
   # These "track data present" tests fail with invalid expiration dates. The
   # test track data probably needs to be updated.
   def test_card_present_purchase
-    @credit_card.track_data = '%B378282246310005^LONGSON/LONGBOB^2205101130504392?'
+    @credit_card.track_data = '%B378282246310005^LONGSON/LONGBOB^2705101130504392?'
```
jcreiff, line 671: Updated the expiration year on these track data tests because the tests were failing due to invalid date

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -555,6 +556,22 @@ def add_emv_metadata(post, creditcard)
         post[:metadata][:card_read_method] = creditcard.read_method if creditcard.respond_to?(:read_method)
       end
 
+      def add_shipping_address(post, payment, options = {})
+        return unless shipping = options[:shipping_address]
+
+        post[:shipping] = {}
+
+        post[:shipping][:address] = {}
+        post[:shipping][:name] = shipping[:name] || options[:billing_address].try(:[], :name) || payment.try(:name)
```
jcreiff, line 565: These changes were originally merged in #4484 but were later reverted because an attempt to use `options[:billing_address][:name] was resulting in a NoMethodError when neither `shipping[:name]` nor `[:billing_address]` were present.

Line 565 here is the primary difference between the last PR and this one. It guards against the NoMethodError with `.try()` and provides another fallback, which is to use the `name` stored with the payment method itself. (The `payment.try(:name)` is an attempt to guard against the possibility of a payment method not having a `:name` attribute, such as in the case of a third party token.)

The purpose of including these fallback options is that Stripe will fail the transaction if the shipping fields don't include a name. Until now, attempts to send shipping address to Stripe have been quietly ignored, so if this change did not include the fallback option, it's possible that some merchants might start to see Stripe failure messages for transactions that had previously succeeded. (In fact, I feel pretty confident that the ActiveMerchant errors that necessitated the earlier revert would have instead manifested as Stripe error messages -- either way, the transaction would not have succeeded.)

Part of me wonders if instead of adding this logic around filling in a shipping name, if we should check for `options[:shipping_address] && options[:shipping_address][:name]` in the guard clause on line 560 instead. (Or perhaps some other alternative?)

jcreiff: I modified this (along with the test suite expectations) to return from the `add_shipping_address` method if no shipping name was present

rachelkirk (approved):
This looks good, very thorough! Did a little playing around with pry and things are working as expected. ✨ 🚢  ✨

PR #4542, opened 2022-08-18
Rapyd: Customer Object on `auth` and `purchase`

naashton:
Include the `customer_object` in `authorize` and `purchase`
transactions.

Added a guard to only perform this action if the payment is an instance
of `CreditCard`

These fields are not documented at Rapyd.

SER-228

Unit: 20 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -29,6 +29,7 @@ module ActiveMerchant #:nodoc:
         add_payment_fields(post, options)
         add_payment_urls(post, options)
         add_customer_id(post, options)
+        add_customer_object(post, payment, options)
         post[:capture] = true if payment.is_a?(CreditCard)
 
         if payment.is_a?(Check)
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -225,6 +227,8 @@ def add_payment_urls(post, options)
       end
 
       def add_customer_object(post, payment, options)
+        return unless payment.instance_of?(CreditCard)
+
         post[:name] = "#{payment.first_name} #{payment.last_name}"
         post[:phone_number] = options[:billing_address][:phone].gsub(/\D/, '') if options[:billing_address]
         post[:email] = options[:email] if options[:email]
```
drkjc, line 234: I saw that `email` isn't required, just recommended, but if an `email` field isn't explicitly included in the auth/purchase request, doesn't it need to be pulled from the payment method like the `name`, where as `[:billing_address]` is populated by Core?

drkjc (changes requested):
Left a comment about `email`.  Other than that, looks good. Don't forget a CHANGELOG entry.

lib/active_merchant/billing/gateways/rapyd.rb
naashton: It seems like `options[:email]` is pretty common throughout AM, but there are instances of email being pulled from the `address` hash. I don't see any instance of `email` coming from the payment method, however.

PR #4543, opened 2022-08-18
Adyen: modify handling of countryCode for ACH

jcreiff:
This change allows ACH transactions to fail with an Adyen error message
instead of with a NoMethodError when no `:country` is present inside the
`:billing_address` hash

CER-151

LOCAL
5284 tests, 76224 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

UNIT
95 tests, 482 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

REMOTE
125 tests, 445 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.6% passed
- There are two 3DS tests failing, and another related to Cabal card type. These failures are not connected to this change and are also failing on master


rachelkirk (approved):
Looks good, a nice useful error! I got same test results with 3 remote tests failing on master.

PR #4544, opened 2022-08-23
Vantiv(Litle): Add support for `fraudFilterOverride`

rachelkirk:
CER-51

This PR adds the `fraudFilterOverride` flag, a boolean which when set to true will override all fraud filters for the submitted transaction.

Unit Tests:
56 tests, 240 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote Tests:
56 tests, 226 assertions, 13 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
76.7857% passed
* I ran the remote tests several times and 13 (sometimes 14) tests are failing and have been for over a year. I think this gateway is due for an overhaul but the failures aren’t related to the changes and are also failing on master.

Local Tests:
5288 tests, 76246 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -320,6 +320,7 @@ module ActiveMerchant #:nodoc:
         add_merchant_data(doc, options)
         add_debt_repayment(doc, options)
         add_stored_credential_params(doc, options)
+        add_fraud_filter_override(doc, options)
       end
 
       def add_credit_params(doc, money, payment_method, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -320,6 +320,7 @@ def add_auth_purchase_params(doc, money, payment_method, options)
         add_merchant_data(doc, options)
         add_debt_repayment(doc, options)
         add_stored_credential_params(doc, options)
+        add_fraud_filter_override(doc, options)
```
rachelkirk, line 323: I added it to `auth_purchase_params` but it may need to be added to `echeck_purchase_params` as well.

jcreiff: Hmm...it's a little bit ambiguous from the Vantiv/WP docs, but what I see is that they call `fraudFilterOverride` a child element of either `authorization` or `sale`. Perhaps it depends on whether they consider `echeckSale` to be an entirely different thing than a `sale` or just a specific type of `sale`?

I think I would probably leave it as you have it here.

jcreiff (approved):
Looks good! 

I confirmed that the remote tests are also failing on master. The newly added tests do pass.

PR #4547, opened 2022-08-23
BP Plus - PS Create Card Type

ali-hassan:
BP Plus - PS Create Card Type

[Ticket](https://spreedly.atlassian.net/browse/SER-259)

Finished in 40.794637 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
5288 tests, 76248 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
129.62 tests/s, 1869.07 assertions/s
Running RuboCop...
Inspecting 746 files

### DIFF
--- a/lib/active_merchant/billing/credit_card.rb
+++ b/lib/active_merchant/billing/credit_card.rb
@@ -33,6 +33,7 @@ module ActiveMerchant #:nodoc:
     # * Creditel
     # * Confiable
     # * Mada
+    # * BpPlus
     #
     # For testing purposes, use the 'bogus' credit card brand. This skips the vast majority of
     # validations, allowing you to focus on your core concerns until you're ready to be more concerned
@@ -118,6 +119,7 @@ module ActiveMerchant #:nodoc:
       # * +'creditel'+
       # * +'confiable'+
       # * +'mada'+
+      # * +'bp_plus'+
       #
       # Or, if you wish to test your implementation, +'bogus'+.
       #
--- a/test/unit/credit_card_methods_test.rb
+++ b/test/unit/credit_card_methods_test.rb
@@ -192,6 +192,14 @@ class CreditCardMethodsTest < Test::Unit::TestCase
     assert_equal 'mada', CreditCard.brand?('9682040000000000')
   end
 
+  def test_should_detect_bp_plus_card
+    assert_equal 'bp_plus', CreditCard.brand?('7051 211111111 111')
+    assert_equal 'bp_plus', CreditCard.brand?('7050 111111111 111')
+    assert_equal 'bp_plus', CreditCard.brand?('70505 911111111 111')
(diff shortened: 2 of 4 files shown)


test/unit/credit_card_methods_test.rb
```diff
@@ -192,6 +192,14 @@ def test_should_detect_mada_card
     assert_equal 'mada', CreditCard.brand?('9682040000000000')
   end
 
+  def test_should_detect_bp_plus_card
+    assert_equal 'bp_plus', CreditCard.brand?('7051 211111111 111')
+    assert_equal 'bp_plus', CreditCard.brand?('7050 111111111 111')
```
dsmcclain, line 197: Based on the documentation I'm reading, I think these two test cases are not correct. The documentation shows that the format for cards starting with `705N` is `705N NNNNNNNN NNNNN`, with the final digit being a luhn check number. I am going to ask for more information about this, but at the very least I think the format after dropping the final number would be `7050 11111111 1111` and not `7050 111111111 111`.

lib/active_merchant/billing/credit_card.rb
```diff
@@ -1,3 +1,4 @@
+
```
dsmcclain, line 1: Please remove this empty line, it causes rubocop to fail.

Are you running `bundle exec rake test:local` on your branch before submitting it? I see you have added test output to your PR description, but it looks like you have just pasted output from a previous test run because it shows rubocop passing and it shows remote tests when there are no remote tests to run for this branch.

Please run `bundle exec rake test:local` on your branch and include that output in your PR description.

dsmcclain (changes requested):
Hi @ali-hassan, there is a rubocop error that needs to be fixed on this branch.

I also think that the bin ranges we are adding here might be a bit more complicated, and I'm not sure the regex is correct for every format. I will ask for more information about this and hopefully we can figure it out together. In the meantime, please fix the rubocop error.

lib/active_merchant/billing/credit_card.rb
ali-hassan: Fixed

Finished in 40.794637 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
5288 tests, 76248 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
129.62 tests/s, 1869.07 assertions/s
Running RuboCop...
Inspecting 746 files

dsmcclain:
Closing this as we will need to do additional work to identify and validate these unusual number formats. A new ticket has been created for this work.

PR #4548, opened 2022-08-23
Add authorization code to handle access_token and encription key

Heavyblade:
### Description

This PR adds code to handle Alelo's access_token, encryption key and request body encription


jherreraa (approved):
👍  LGTM

PR #4552, opened 2022-08-25
Paysafe: Add fundingTransaction object

jcreiff:
Beginning in September 2022, Paysafe will require mandatory changes to the handling of Account Funding Transactions via the Visa network

https://developer.paysafe.com/en/cards/api/#/fundingTransaction

CER-96

LOCAL
5288 tests, 76248 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

UNIT
17 tests, 82 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

REMOTE
33 tests, 101 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paysafe.rb
+++ b/lib/active_merchant/billing/gateways/paysafe.rb
@@ -17,14 +17,8 @@ module ActiveMerchant #:nodoc:
 
       def purchase(money, payment, options = {})
         post = {}
-        add_invoice(post, money, options)
-        add_payment(post, payment)
-        add_billing_address(post, options)
-        add_merchant_details(post, options)
+        add_auth_purchase_params(post, money, payment, options)
         add_airline_travel_details(post, options)
-        add_customer_data(post, payment, options) unless payment.is_a?(String)
-        add_three_d_secure(post, payment, options) if options[:three_d_secure]
-        add_stored_credential(post, options) if options[:stored_credential]
         add_split_pay_details(post, options)
         post[:settleWithAuth] = true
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -286,6 +288,15 @@ def add_split_pay_details(post, options)
         post[:splitpay] = split_pay
       end
 
+      def add_funding_transaction(post, options)
+        return unless options[:funding_transaction]
+
+        post[:fundingTransaction] = {}
+        post[:fundingTransaction][:type] = options[:funding_transaction]
+        post[:profile] ||= {}
```
rachelkirk, line 296: > post[:profile] ||= {}

What is the function of the || here?

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -26,6 +26,7 @@ def purchase(money, payment, options = {})
         add_three_d_secure(post, payment, options) if options[:three_d_secure]
         add_stored_credential(post, options) if options[:stored_credential]
         add_split_pay_details(post, options)
+        add_funding_transaction(post, options)
```
rachelkirk, line 29: Don't want to make extra work for you, but this might be a nice opportunity for some refactoring with a `add_auth_purchase_params` method since the list of added methods is getting pretty long in auth and purchase, just a thought.

rachelkirk (changes requested):
Looks good, thanks for including the gateway docs. Just had one qq!

lib/active_merchant/billing/gateways/paysafe.rb
jcreiff: If there's already a `profile` object that exists within `post`, it will just remain as it is, but if there's not one that exists, it will initialize it as an empty hash.

I think it's roughly equivalent to `post[:profile] || post[:profile] = {}`

It may not be super necessary to include, because I think the `profile` should _probably_ already exist by the time we get to this method, but this should act as a safeguard, just in case for some reason it does not yet exist.

lib/active_merchant/billing/gateways/paysafe.rb
jcreiff: Good call! I pushed a change that extracts the shared logic out into `add_auth_purchase_params`. There were a couple of things that only existed in purchase and not authorize, so I left them out of the new method, in case they were purchase-specific.

rachelkirk (approved):
Tests pass! I think the `auth_purchase_params` method should be a private method. Approving with that suggestion and don't forget your changelog. ✨

PR #4555, opened 2022-08-29
Alelo: Add gateway

Heavyblade:
## Summary

Adding purchase calls and updating the response handling so the code is able to retry the acces_token and key in
unauthorized errors.

### Unit Test:
Finished in 37.256649 seconds.
5295 tests, 76270 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### Remote Test:
Finished in 30.736933 seconds.
16 tests, 41 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:
749 files inspected, no offenses detected


curiousepic:
I think keeping `jose` in the test/dev group is best. Was there a reason you think otherwise?

lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -0,0 +1,243 @@
...
+      def message_from(response)
+        response[:messages] || response[:messageUser]
+      end
+
+      def authorization_from(response)
+        [response[:requestId], response[:authorizationCode]].join('#')
+      end
+
+      def url(action)
+        "#{test? ? test_url : live_url}#{action}"
```
curiousepic, line 229: I'm confused by the lack of action specification for the test_url...

Heavyblade: mmm maybe we were blindly following AM template names, the "action" is just the path for the particular end-point needed, maybe "path" or "resource" is more REST oriented ?

curiousepic: Oh - no I think the name is ok. What I mean is, why does the live_url have a concatenated action value, but the test_url does not? Is the sandbox just that much different?

curiousepic: Oh nevermind - I just can't read 🤦

lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -0,0 +1,245 @@
...
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        force_utf8(transcript.encode).
+          gsub(%r((Authorization: Bearer )[\w -]+), '\1[FILTERED]').
+          gsub(%r((client_id=|Client-Id:)[\w -]+), '\1[FILTERED]\2').
+          gsub(%r((client_secret=|Client-Secret:)[\w -]+), '\1[FILTERED]\2')
```
curiousepic, line 46: We need `cardNumber` and `securityCode`, right?

curiousepic: I see these scrubbed values are headers and hence not encrypted, while the boody itself is. Makes sense 👍

lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -0,0 +1,245 @@
...
+          client_secret: @options[:client_secret],
+          scope: '/capture'
+        }
+
+        headers = {
+          'Accept' => 'application/json',
+          'Content-Type' => 'application/x-www-form-urlencoded'
+        }
+
+        parsed = parse(ssl_post(url('captura-oauth-provider/oauth/token'), post_data(params), headers))
```
curiousepic, line 108: is the spelling of `captura` correct?

lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -0,0 +1,245 @@
...
+      def parse(body)
+        JSON.parse(body, symbolize_names: true)
+      end
+
+      def post_data(params)
+        params.map { |k, v| "#{k}=#{CGI.escape(v.to_s)}" }.join('&')
+      end
+
+      def commit(action, body, options, try_again = true)
+        payload, credentials = encrypt_payload(body, options)
```
curiousepic, line 172: I might suggest extracting  the call to`ensure_credentials` to a new line above, before `encrypt_payload` for better readability. What do you think?

curiousepic:
Looking good so far. The entire commit method is pretty convoluted due to the variable multi-step calls necessary, so there are probably other refactors to improve readability, but we can do without for now.

lib/active_merchant/billing/gateways/alelo.rb
Heavyblade: yup, is in Portuguese

lib/active_merchant/billing/gateways/alelo.rb
Heavyblade: yup, make sense

almalee24:
All the test pass for me and the code looks good. Just wondering if there's a way to test this in Postman?

almalee24 (approved):
Looks great!

lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -0,0 +1,245 @@
...
+          externalTraceNumber: options[:external_trace_number]
+        }.compact)
+      end
+
+      def add_geolocation(post, options)
+        return if options[:geo_latitude].blank? || options[:geo_longitude].blank?
+
+        post.merge!(geolocation: {
+          latitude: options[:geo_latitude],
+          longitude: options[:geo_longitude]
```
curiousepic, line 81: Good, I like having fewer sub-hashes for options 👍

curiousepic (approved):
Cool, merging!

curiousepic:
Closed with https://github.com/activemerchant/active_merchant/commit/7a26741254853f6f2897db99242c22ea659d2a29

PR #4556, opened 2022-08-30
Stripe Payment Intents: fix bug with billing address email

jcreiff:
This corrects an issue where email was being dropped from billing_details in the request.

The previous assumption was that email would be a field contained within the billing_address hash, but (in Spreedly's integration, at least), it's actually nested at the top-level of the options hash.

This change should not affect behavior of the add_billing_address method for any other integrations with Active Merchant that do supply an email attribute as part of the billing_address hash.

CER-41

LOCAL
5292 tests, 76267 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

UNIT
39 tests, 207 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

REMOTE
78 tests, 372 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -451,6 +451,8 @@ module ActiveMerchant #:nodoc:
       def add_billing_address(post, options = {})
         return unless billing = options[:billing_address] || options[:address]
 
+        email = billing[:email] || options[:email]
+
         post[:billing_details] = {}
         post[:billing_details][:address] = {}
         post[:billing_details][:address][:city] = billing[:city] if billing[:city]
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -451,6 +451,8 @@ def setup_future_usage(post, options = {})
       def add_billing_address(post, options = {})
         return unless billing = options[:billing_address] || options[:address]
 
+        email = billing[:email] || options[:email]
+
```
rachelkirk, line 455: Nice!

rachelkirk (approved):
Everything looks good to me! ✨ Go forth and merge! ✨

PR #4557, opened 2022-08-31
Shift4: Add customer object to transactions

ajawadmirza:
Added customer object to purchase and store & removed transaction object from store opertaion for shift4 gateway implementation.

SER-277
SER-275
SER-279

Unit:
5294 tests, 76283 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
19 tests, 45 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4.rb
+++ b/lib/active_merchant/billing/gateways/shift4.rb
@@ -54,19 +54,22 @@ module ActiveMerchant #:nodoc:
         add_transaction(post, options)
         add_card(post, payment_method, options)
         add_card_present(post, options)
+        add_customer(post, payment_method, options)
 
         commit('sale', post, options)
       end
 
       def authorize(money, payment_method, options = {})
         post = {}
+
+        payment_method = get_card_token(payment_method) if payment_method.is_a?(String)
         add_datetime(post, options)
         add_invoice(post, money, options)
         add_clerk(post, options)
         add_transaction(post, options)
         add_card(post, payment_method, options)
         add_card_present(post, options)
-        add_customer(post, options)
+        add_customer(post, payment_method, options)
 
         commit('authorization', post, options)
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -197,17 +198,16 @@ def add_card_present(post, options)
...
-          post[:customer] = {}
-          post[:customer][:addressLine1] = address[:address1] if address[:address1]
-          post[:customer][:postalCode] = address[:zip]
-          name = address[:name].split(' ') if address[:name]
-          if name&.is_a?(Array)
-            post[:customer][:firstName] = name[0]
-            post[:customer][:lastName] = name[1]
-          end
-        end
+      def add_customer(post, options, card = nil)
```
naashton, line 201: We should swap the `options` and `card` params to fit the AM method pattern. 

We don't need to set `card` too `nil` since in each of the transactions where this is referenced, a payment method is expected. Line 207 and 208 handle any issue where the card is not a `CreditCard`.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -54,6 +54,7 @@ def purchase(money, payment_method, options = {})
         add_transaction(post, options)
         add_card(post, payment_method, options)
         add_card_present(post, options)
+        add_customer(post, options, payment_method)
```
naashton, line 57: After changing the `add_customer` methods signature, you'll need to swap the `options` and `payment_method` params. Same for line 70 and 119.

naashton (changes requested):
Looks good, just a small change request.

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: Thanks @naashton , I have updated this change 👍

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: Thanks @naashton , I have updated this change 👍

naashton (approved):
Awesome work!

PR #4560, opened 2022-09-01
MerchantE: Fix condition to disallow null values

ajawadmirza:
Updated condition so that nil value is not passed for the `moto_ecommerce_ind` field to make all the integration tests pass.

SER-272

Unit:
5293 tests, 76279 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
20 tests, 84 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


dsmcclain (approved):
Looks good, I will merge 👍

PR #4562, opened 2022-09-02
Rapyd: No force `capture` for ACH

naashton:
Rapyd automatically settles ACH transactions at EOD and force-capturing.

SER-284

Unit: 20 tests, 93 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -20,42 +20,16 @@ module ActiveMerchant #:nodoc:
 
       def purchase(money, payment, options = {})
         post = {}
-        add_invoice(post, money, options)
-        add_payment(post, payment, options)
-        add_3ds(post, payment, options)
-        add_address(post, payment, options)
-        add_metadata(post, options)
-        add_ewallet(post, options)
-        add_payment_fields(post, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -20,42 +20,16 @@ def initialize(options = {})
...
-        add_payment(post, payment, options)
-        add_3ds(post, payment, options)
-        add_address(post, payment, options)
-        add_metadata(post, options)
-        add_ewallet(post, options)
-        add_payment_fields(post, options)
-        add_payment_urls(post, options)
-        add_customer_id(post, options)
+        add_auth_purchase(post, money, payment, options)
         post[:capture] = true if payment.is_a?(CreditCard)
```
dsmcclain, line 24: The only thing I am wondering here is whether folks will be transacting on Rapyd with any other tokenized payment method type that is not a Check and not a CreditCard, and whether we might want `capture: true` in those cases?

dsmcclain (approved):
All tests pass, including remote. I have one question (see comments). Approving so you can merge if it's a non-issue.

Don't forget changelog!

PR #4563, opened 2022-09-02
Orbital: Fix CardSecValInd

molbrown:
The card sec val ind is only valid on BIN 1 merchant Visa and Discover transactions

This issue is not reflected in the sandbox, but was exposed in Developer
Center testing during certification. The intended application of this
field is also documented in the code comment.

ECS-2144

Unit:
141 tests, 803 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
119 tests, 497 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


aenand (approved):
LGTM

PR #4564, opened 2022-09-08
Shift4: Add SC fields from `cardOnFile` field params

ajawadmirza:
Updated AM implementation to pass `cardOnFile` fields from params along with supporting SC framework fields.

SER-283

Remote:
21 tests, 49 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
5308 tests, 76393 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop:


CHANGELOG
```diff
@@ -114,7 +114,8 @@
 * Add BpPlus card type along with custom validation logic [dsmcclain] #4559
 * PayTrace: Support ACH implementation for new endpoints and request body [ajawadmirza] #4545
 * Rapyd: No force capture for ACH [naashton] #4562
-* Shift4: Applied checks on Shift4 Time/Timezone offset [ali-hassan] #45611
+* Shift4: Applied checks on Shift4 Time/Timezone offset [ali-hassan] #4561
```
ajawadmirza, line 117: Fixed previous PR's type.

lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -222,13 +222,14 @@ def add_purchase_card(post, options)
...
-        post[:cardOnFile][:indicator] = options[:card_on_file_indicator] || '01'
-        post[:cardOnFile][:scheduledIndicator] = RECURRING_TYPE_TRANSACTIONS.include?(stored_credential[:reason_type]) ? '01' : '02' if stored_credential[:reason_type]
-        post[:cardOnFile][:transactionId] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
+        post[:cardOnFile][:usageIndicator] = options[:usage_indicator] || (stored_credential[:initial_transaction] ? '01' : '02') if stored_credential[:reason_type] || options[:usage_indicator]
+        post[:cardOnFile][:indicator] = options[:indicator] || '01' if stored_credential[:reason_type] || options[:indicator]
+        post[:cardOnFile][:scheduledIndicator] = options[:scheduled_indicator] || (RECURRING_TYPE_TRANSACTIONS.include?(stored_credential[:reason_type]) ? '01' : '02') if stored_credential[:reason_type] || options[:scheduled_indicator]
+        post[:cardOnFile][:transactionId] = options[:transaction_id] || (stored_credential[:network_transaction_id]) if stored_credential[:network_transaction_id] || options[:transaction_id]
```
naashton, line 231: I think you might be able to reduce your the logic here to this:
`... = options[:usage_indicator] || (stored_credential[:initial_transaction] ? '01' : '02')`.

If no `options[:usage_indicator]` is set, then it should handle the `||` part of the condition. Anything after the `if` will be ignored.

I see what you're trying to do here and I think it's good intuition, but it might make sense to validate the existence of one or the other set of fields (stored_credential vs GSFs) and return early from the method if neither exists. That said, I'm not sure how far we need to go here in preventing anything other than TypeErrors and other logic errors that might bubble up. `options[:usage_indicator]` will just return `nil` if that key isn't found, in which case the gateway might yield some error message we can pass back to the user.

test/unit/gateways/shift4_test.rb
```diff
@@ -127,6 +127,39 @@ def test_successful_purchase_with_stored_credential
     end.respond_with(successful_purchase_response)
   end
 
+  def test_successful_purchase_with_card_on_file_fields
```
naashton, line 130: This is sufficient, I think, but it might be nice to add a test that has both sets of fields from `stored_credentials` and the GSF's and assert that the ones we put in the post body are what we are expecting.

naashton (changes requested):
This looks good, but I think the logic can be reduced and I have a suggestion for some test coverage.

lib/active_merchant/billing/gateways/shift4.rb
ajawadmirza: Yes my idea was to prevent it from passing nil values to the gateway in any case (even when one card on file is passed and other is not) but since all tests do pass with the changes you suggested so we should be good to apply them.

I have applied these changes with the extra test to validate the fields when GSFs and SC fields are combinedly passed 👍

test/unit/gateways/shift4_test.rb
ajawadmirza: Thanks @naashton , the changes are applied.

test/unit/gateways/shift4_test.rb
```diff
@@ -127,6 +127,60 @@ def test_successful_purchase_with_stored_credential
...
+      usage_indicator: '02',
+      indicator: '02',
+      scheduled_indicator: '02'
+    }
+    stored_credential_options = {
+      initial_transaction: true,
+      reason_type: 'recurring'
+    }
+    stub_comms do
+      @gateway.purchase(@amount, @credit_card, @options.merge(card_on_file_fields, { stored_credential: stored_credential_options }))
```
naashton, line 174: I'm getting an error here, `Error: test_card_on_file_fields_and_stored_credential_framework_combined(Shift4Test): ArgumentError: wrong number of arguments (given 2, expected 1)`

naashton (changes requested):
This looks good, just need to fix the broken test and I'll approve.

test/unit/gateways/shift4_test.rb
ajawadmirza: image
They are getting passed for me. I think there is an issue with merge as it doesn't support multiple arguments in older version of ruby.

naashton (approved):
Awesome work!

PR #4565, opened 2022-09-08
Redsys: update Base64 encryption handling for secret key

jcreiff:
Some Redsys customers receive a secret key with url safe characters. This has caused problems with the `encrypt` method, because Base64.strict_decode64 throws an error when it encounters one of these characters. Using urlsafe_decode64 instead should prevent this error from occurring in the future.

CER-75
Local
5307 tests, 76382 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit
33 tests, 104 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote
22 tests, 62 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 77.2727% passed
(The 5 failures also occur on master)


curiousepic (approved):
👍

PR #4566, opened 2022-09-08
Shift4: Timezone Offset

naashton:
Shift4 requires a timestamp be sent with the timezone offset (offset from GMT).

Correct format: `yyyy-mm-ddThh:mm:ss.nnn+hh:mm`

This PR corrects the offset for the timestamp to be either `-hh:mm` or `+hh:mm` depending on the offset relative to GMT (i.e., `-05:00` for 'EST').

SER-273

Unit: 19 tests, 112 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote: 21 tests, 51 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/test/remote/gateways/remote_shift4_test.rb
+++ b/test/remote/gateways/remote_shift4_test.rb
@@ -78,6 +78,7 @@ class RemoteShift4Test < Test::Unit::TestCase
       initial_transaction: true,
       reason_type: 'recurring'
     }
+    @options[:merchant_time_zone] = 'EST'
     first_response = @gateway.purchase(@amount, @credit_card, @options.merge(@extra_options.merge({ stored_credential: stored_credential_options })))
     assert_success first_response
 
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_shift4_test.rb
```diff
@@ -78,6 +78,7 @@ def test_successful_purchase_with_stored_credential
       initial_transaction: true,
       reason_type: 'recurring'
     }
+    @options[:merchant_time_zone] = 'EST'
```
aenand, line 81: Can we add an expect here to verify the timezone is set properly?

naashton: Just added a unit test for this. Thanks for the suggestion!

aenand (approved):
LGTM

PR #4568, opened 2022-09-09
Accept both formats of Canadian routing numbers

molbrown:
Canadian routing numbers are typically handled either in print/MICR format or electronic format (see https://en.wikipedia.org/wiki/Routing_number_(Canada)#Format).

Adds ability to send valid MICR formatted Canadian routing numbers, which may still be expected by some gateways.

Unit tests:
5308 tests, 76382 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop:

ECS-2144

### DIFF
--- a/test/unit/check_test.rb
+++ b/test/unit/check_test.rb
@@ -4,11 +4,34 @@ class CheckTest < Test::Unit::TestCase
   VALID_ABA     = '111000025'
   INVALID_ABA   = '999999999'
   MALFORMED_ABA = 'I like fish'
-  VALID_CBA = '000194611'
+  VALID_ELECTRONIC_CBA = '000194611'
+  VALID_MICR_CBA = '94611001'
   INVALID_NINE_DIGIT_CBA = '012345678'
   INVALID_SEVEN_DIGIT_ROUTING_NUMBER = '0123456'
 
   ACCOUNT_NUMBER = '123456789012'
+
+  CHECK_US = Check.new(
+    name: 'Fred Bloggs',
+    routing_number: VALID_ABA,
+    account_number: ACCOUNT_NUMBER,
+    account_holder_type: 'personal',
+    account_type: 'checking'
+  )
+  CHECK_CAN_E_FORMAT = Check.new(
+    name: 'Tim Horton',
+    routing_number: VALID_ELECTRONIC_CBA,
+    account_number: ACCOUNT_NUMBER,
+    account_holder_type: 'personal',
+    account_type: 'checking'
+  )
+  CHECK_CAN_MICR_FORMAT = Check.new(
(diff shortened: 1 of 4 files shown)


test/unit/check_test.rb
```diff
@@ -4,7 +4,8 @@ class CheckTest < Test::Unit::TestCase
   VALID_ABA     = '111000025'
   INVALID_ABA   = '999999999'
   MALFORMED_ABA = 'I like fish'
```
aenand, line 6: I also like fish

test/unit/check_test.rb
```diff
@@ -82,10 +83,20 @@ def test_account_type
...
+      routing_number: VALID_ELECTRONIC_CBA,
+      account_number: ACCOUNT_NUMBER,
+      account_holder_type: 'personal',
+      account_type: 'checking'
+    )
+  end
+
+  def test_valid_canada_routing_number_micr_format
+    assert_valid Check.new(
+      name: 'Tim Horton',
```
aenand, line 98: Tim Horton's is 🐐

aenand (approved):
LGTM

therufs:
Kind of a philosophical question, but my original thought when doing this implementation was that if the gateway expects a specific format of routing number, that's something the gateway adapter should handle, rather than requiring someone who's integrating against AM to potentially have to create different payment methods for the same account. 

Thoughts?

molbrown:
@therufs It's a good question. I stuck it back in here because I saw this format used to be here. In response to your question, I see these two formats as potentially equally standard, and can't be certain how widely used either is. For example, the format validated for US routing numbers via the checksum is also an MICR format. In that case, it's the only standard format, but I think the Canadian MICR does fit in here as well.

I don't consider the retainment and reuse of payment methods to be concerns of Active Merchant. 🤔  What to you think?

therufs:
@molbrown I see it less as being relevant to payment method retention, more about presenting an interface that's gateway agnostic -- it seems like (?) some gateways want one format and some want the other, and I don't think an AM user should have to keep track of which gateway wants which behavior. (I know there's a whole options hash that belies this, but if it's avoidable, we should probably avoid it.)

I wonder if we could implement a pair of methods on `check.rb` that would allow the number to be returned in either format? (I wish I had thought of this when I was doing the original implementation 😅)

molbrown:
@therufs I like that idea, here's a stab at such an approach, with some mild refactoring (maybe could use more).

therufs (approved):
Aw heck yeah love this 🤩  

I don't see anything that obviously wants refactoring, but if there's particular stuff you have in mind, happy to take another look!

PR #4572, opened 2022-09-13
Shift4: Retrieve `access_token` once

naashton:
Remove the `setup_access_token` method call from the constructor and make the method public so that users may retrieve the `access_token` one time and pass the value to other instances of the gateway class

Due to reasons at Shift4, calls to retrieve the `access_token` must be made once because the `auth_token` value has a one-time use. Any subsequent requests to exchange the `auth_token` for an `access_token` will lead to an invalid `access_token`

Because of the way we instantiate the gateway subclasses, in order to run this access token request on setup, we need to call `super` to initialize the `options` instance variable

SER-221

Unit: 21 tests, 121 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote: 21 tests, 51 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4.rb
+++ b/lib/active_merchant/billing/gateways/shift4.rb
@@ -40,8 +40,8 @@ module ActiveMerchant #:nodoc:
         requires!(options, :client_guid, :auth_token)
         @client_guid = options[:client_guid]
         @auth_token = options[:auth_token]
+        @access_token = options[:access_token]
         super
-        @access_token = setup_access_token
       end
 
       def purchase(money, payment_method, options = {})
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -136,14 +136,6 @@ def scrub(transcript)
...
 
-      private
-
-      def add_credentials(post, options)
-        post[:credential] = {}
-        post[:credential][:clientGuid] = @client_guid
-        post[:credential][:authToken] = @auth_token
-      end
-
       def setup_access_token
```
molbrown, line 139: Is the idea that `setup_access_token` is part of creation of a gateway object, allowing for retaining the access_token on the gateway? Or will this result in an extra API request?

naashton: Yes, this is to enable us to make the exchange for an `access_token` when a gateway object is created and we can store that value to be used for other requests. 

It's going to require an additional API request at the time of creation for that gateway but we will check for the presence of an `access_token` when auth is being setup to avoid making extraneous requests to the gateway.

molbrown (approved):
Nice, this makes sense 👍

PR #4576, opened 2022-09-15
GSFs on MerchantE Part 3 - Active Merchant

ali-hassan:
GSFs on MerchantE Part 3 - Active Merchant

[Ticket](https://spreedly.atlassian.net/browse/SER-310)

5333 tests, 76509 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

116.59 tests/s, 1672.66 assertions/s
Running RuboCop...
Inspecting 749 files

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -124,6 +124,7 @@
 * Shift4: Add `usage_indicator`, `indicator`, `scheduled_indicator`, and `transaction_id` fields [ajawadmirza] #4564
 * Shift4: Retrieve `access_token` once [naashton] #4572
 * Redsys: Update Base64 encryption handling for secret key [jcreiff] #4565
+* MerchantE: add `card_on_file`, 'account_data_source' and `cit_mit_indicator` fields if present [ali-hassan] #4576
 
 == Version 1.126.0 (April 15th, 2022)
 * Moneris: Add 3DS MPI field support [esmitperez] #4373
--- a/lib/active_merchant/billing/gateways/merchant_e_solutions.rb
+++ b/lib/active_merchant/billing/gateways/merchant_e_solutions.rb
@@ -31,6 +31,9 @@ module ActiveMerchant #:nodoc:
         post[:moto_ecommerce_ind] = options[:moto_ecommerce_ind] if options[:moto_ecommerce_ind]
         post[:recurring_pmt_num] = options[:recurring_pmt_num] if options[:recurring_pmt_num]
         post[:recurring_pmt_count] = options[:recurring_pmt_count] if options[:recurring_pmt_count]
+        post[:card_on_file] = options[:card_on_file] if options[:card_on_file]
+        post[:cit_mit_indicator] = options[:cit_mit_indicator] if options[:cit_mit_indicator]
+        post[:account_data_source] = options[:account_data_source] if options[:account_data_source]
         add_invoice(post, options)
         add_payment_source(post, creditcard_or_card_id, options)
         add_address(post, options)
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/merchant_e_solutions.rb
```diff
@@ -31,6 +31,9 @@ def authorize(money, creditcard_or_card_id, options = {})
         post[:moto_ecommerce_ind] = options[:moto_ecommerce_ind] if options[:moto_ecommerce_ind]
         post[:recurring_pmt_num] = options[:recurring_pmt_num] if options[:recurring_pmt_num]
         post[:recurring_pmt_count] = options[:recurring_pmt_count] if options[:recurring_pmt_count]
+        post[:card_on_file] = options[:card_on_file] if options[:card_on_file]
+        post[:cit_mit_indicator] = options[:cit_mit_indicator] if options[:cit_mit_indicator]
+        post[:account_data_source] = options[:account_data_source] if options[:account_data_source]
```
naashton, line 36: This all looks good, but I think this is looking a bit bloated. We should create a common method to append these values to the `post` body and call that method within the `authorize` and `purchase` method transactions.

naashton (changes requested):

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
naashton: Since these all look like some variation of stored credentials, we should go ahead and make a method for `add_stored_credentials` and then move these and the `recurring_*` fields as well.

CHANGELOG
```diff
@@ -127,8 +127,8 @@
 * Openpay: Update url endpoints [ajawadmirza] #4573
 * Shift4: refuse `postalCode` when its null [ajawadmirza] #4574
 * Plexo: Update param key to `refund_type` [ajawadmirza] #4575
+* MerchantE: add `card_on_file`, 'account_data_source' and `cit_mit_indicator` fields if present [ali-hassan] #4576
 
-== Version 1.126.0 (April 15th, 2022)
```
naashton, line 131: This will need to be added back

naashton (changes requested):

lib/active_merchant/billing/gateways/merchant_e_solutions.rb
ali-hassan: Done

CHANGELOG
ali-hassan: Done

test/remote/gateways/remote_merchant_e_solutions_test.rb
```diff
@@ -22,7 +22,10 @@ def setup
         zip: '55555',
         phone: '555-555-5555',
         recurring_pmt_num: 11,
-        recurring_pmt_count: 10
+        recurring_pmt_count: 10,
+        card_on_file: 'Y',
+        cit_mit_indicator: 'C101',
+        account_data_source: 'Y'
```
naashton, line 28: Can we add some tests to check the behavior of adding these fields to the request?

test/unit/gateways/merchant_e_solutions_test.rb
```diff
@@ -19,7 +19,10 @@ def setup
       recurring_pmt_num: 11,
       recurring_pmt_count: 10,
       billing_address: address,
-      description: 'Store Purchase'
+      description: 'Store Purchase',
+      card_on_file: 'Y',
+      cit_mit_indicator: 'C101',
+      account_data_source: 'Y'
```
naashton, line 25: If nothing else, we should add in some tests to assert that these values are being added to the post body.

naashton (changes requested):
We need to add some asserts or additional tests to validate what we are doing here.

test/unit/gateways/merchant_e_solutions_test.rb
ali-hassan: Added @naashton

ali-hassan:
@naashton Please check it as required asserts are being added.

PR #4582, opened 2022-09-16
TransFirstTransactionExpress: Add prefix to transCode xml tag

dsmcclain:
This change makes the gateway compatible with Nokogiri versions 1.2x and greater.

While the PR does not include an change to the Nokogiri version within Active Merchant, it should be tested locally against Nokogiri version 1.13.4 to ensure that downstream apps using Rails 6 can enforce this version of Nokogiri without causing problems for this gateway.

_NOTE: NO REMOTE TESTS HAVE BEEN RUN ON THIS BRANCH_ because attempts to connect to the gateway's sandbox are timing out. It is not clear if this is an account-specific issue or if the gateway sandbox no longer exists. Attempts to consult the gateway itself have revealed that TransactionExpress may be nearing its end-of-life.

ECS-2536

Unit:
5354 tests, 76627 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

### DIFF
--- a/test/unit/gateways/trans_first_transaction_express_test.rb
+++ b/test/unit/gateways/trans_first_transaction_express_test.rb
@@ -257,6 +257,15 @@ class TransFirstTransactionExpressTest < Test::Unit::TestCase
     assert_equal nil, response.message
   end
 
+  def test_transaction_code_xml_tag_added_with_correct_prefix
+    stub_comms do
+      @gateway.purchase(@amount, @credit_card)
+    end.check_request do |_endpoint, data, _headers|
+      doc = Nokogiri::XML(data)
+      assert_not_empty doc.xpath('//v1:tranCode', 'v1' => 'http://postilion/realtime/merchantframework/xsd/v1/')
+    end.respond_with(successful_purchase_response)
+  end
+
   def test_transcript_scrubbing
     assert_equal scrubbed_transcript, @gateway.scrub(transcript)
   end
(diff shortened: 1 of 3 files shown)


test/unit/gateways/trans_first_transaction_express_test.rb
```diff
@@ -257,6 +257,15 @@ def test_empty_response_fails
     assert_equal nil, response.message
   end
 
+  def test_transaction_code_xml_tag_added_with_correct_prefix
+    stub_comms do
+      @gateway.purchase(@amount, @credit_card)
+    end.check_request do |_endpoint, data, _headers|
+      doc = Nokogiri::XML(data)
+      assert_not_empty doc.xpath('//v1:tranCode', 'v1' => 'http://postilion/realtime/merchantframework/xsd/v1/')
```
aenand, line 265: Does this create a sample purchase and check the formatted post request for the xpath?

dsmcclain: Yes.

In the `check_request` block the `data` variable holds the request we send to the gateway. So this test is just showing that there is a non-empty `tranCode` field with a `v1` prefix in the request.

If you run this test with Nokogiri 1.13.4 and my changes in the gateway adapter commented-out it will fail.

aenand:
Have we heard any response from the gateway about their sandbox being down?

dsmcclain:
> Have we heard any response from the gateway about their sandbox being down?

Yes, I have been updating the JIRA ticket with the progress. Their latest response is "we are working on it".

aenand (approved):
LGTM

PR #4584, opened 2022-09-19
dLocal: Add transaction query API(s) request

almalee24:
Get payment status by payment id and find payments by order id.

Unit: 36 tests, 155 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Remote: 33 tests, 89 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.9697% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -69,6 +69,11 @@ module ActiveMerchant #:nodoc:
           gsub(%r((\"cvv\\\":\\\")\d+), '\1[FILTERED]')
       end
 
+      def inquire(parameters, options)
+        action = parameters[:payment_id] ? 'status' : 'orders'
+        commit(action, parameters, options)
+      end
+
       private
 
       def add_auth_purchase_params(post, money, card, action, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -69,6 +69,11 @@ def scrub(transcript)
           gsub(%r((\"cvv\\\":\\\")\d+), '\1[FILTERED]')
       end
 
+      def inquiry_request(parameters, options)
```
molbrown, line 72: I would suggest changing this method name to `inquire`, to follow the pattern that these public methods are named for a verb/action. We will use the same method name across all other gateways we add this into.

The [worldpay code](https://github.com/activemerchant/active_merchant/blob/e50f4cd30acd60c54f9c5783f2a06e668634e246/lib/active_merchant/billing/gateways/worldpay.rb#L178) this is modeled after is a private method that is called by worldpay's action methods, so worldpay will also need the addition of this public method eventually.

molbrown (commented):
Quick work! 👏  One small suggestion, just thinking about names since this will be the new method name across the board.

lib/active_merchant/billing/gateways/d_local.rb
almalee24: Will change that, I thought I wrote `inquire`

molbrown: Just to clarify, I would also remove `_request`, so the method is just called `inquire`. This better matches authorize, purchase, capture, void, etc.

molbrown (approved):
Thanks! I get passing tests, except for an unrelated Naranja remote test.

PR #4588, opened 2022-09-20
MercadoPago: Add transaction inquire request

molbrown:
Get transaction by authorization (id)
or by external_reference (order_id)

Also adjusts inquire method signature in dlocal to match mercado_pago's

ECS-2589

mercado_pago:
Unit:
43 tests, 201 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
remote requests not currently working, all 500s :(

d_local:
Unit:
36 tests, 155 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
33 tests, 89 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.9697% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -58,6 +58,13 @@ module ActiveMerchant #:nodoc:
         authorize(0, credit_card, options.merge(verify: 'true'))
       end
 
+      def inquire(authorization, options = {})
+        post = {}
+        post[:payment_id] = authorization
+        action = authorization ? 'status' : 'orders'
+        commit(action, post, options)
+      end
+
       def supports_scrubbing?
         true
       end
@@ -67,11 +74,6 @@ module ActiveMerchant #:nodoc:
           gsub(%r((X-Trans-Key: )\w+), '\1[FILTERED]').
           gsub(%r((\"number\\\":\\\")\d+), '\1[FILTERED]').
           gsub(%r((\"cvv\\\":\\\")\d+), '\1[FILTERED]')
-      end
-
-      def inquire(parameters, options)
-        action = parameters[:payment_id] ? 'status' : 'orders'
-        commit(action, parameters, options)
       end
 
       private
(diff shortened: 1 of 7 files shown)


lib/active_merchant/billing/gateways/mercado_pago.rb
```diff
@@ -61,6 +61,10 @@ def verify(credit_card, options = {})
         end
       end
 
+      def inquire(authorization, options = {})
```
molbrown, line 64: This introduces a different method signature than what @almalee24 established in 430664e58bf630a13244ca855be71eccfd35444c. My thinking is that this `(authorization, options)` signature is more consistent with other public transaction action methods. Also, gateways across the gem are most likely to support a `GET by authorization id` request, if not both, and are already used to seeing `order_id` in options. Thoughts?

aenand: I agree. Can you update the method signature in the `dlocal` gateway as part of this commit

molbrown:
Sort of interesting to propose review while the endpoint cannot confirm functionality. The docs give example responses, and I did have the `test_successful_inquire_with_id` remote test passing yesterday before it started 500ing today. Doesn't seem like huge risk, as this makes no change to the existing transactional requests.

lib/active_merchant/billing/gateways/mercado_pago.rb
molbrown: Updated!

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -58,6 +58,13 @@ def verify(credit_card, options = {})
         authorize(0, credit_card, options.merge(verify: 'true'))
       end
 
+      def inquire(authorization, options)
```
aenand, line 61: Should this be `options = {}`?

molbrown: 👍

aenand (approved):
This looks good to me. Do we want to wait another day to see if maybe the mercado pago gateway works tomorrow?

molbrown:
I think that's reasonable 👍

PR #4590, opened 2022-09-21
Alelo: Adding homologation changes

Heavyblade:
## Summary:

Adding changes to properly use the encryption key UUID on transactions, update types for amount and expiration year.

Adding a separate file for the remote tests that represents the tests cases used for certification

## Unit Test:

Finished in 39.980662 seconds.
5343 tests, 76567 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

## Remote Test:

test/remote/gateways/remote_alelo_test.rb
Finished in 35.552773 seconds.
17 tests, 43 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

test/remote/gateways/remote_alelo_test_certification.rb Finished in 49.461157 seconds.
14 tests, 38 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

## RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/alelo.rb
+++ b/lib/active_merchant/billing/gateways/alelo.rb
@@ -3,8 +3,11 @@ require 'jose'
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class AleloGateway < Gateway
+      class_attribute :prelive_url
+
       self.test_url = 'https://sandbox-api.alelo.com.br/alelo/sandbox/'
-      self.live_url = 'https://desenvolvedor.alelo.com.br'
+      self.live_url = 'https://api.alelo.com.br/alelo/prd/'
+      self.prelive_url = 'https://api.homologacaoalelo.com.br/alelo/uat/'
 
       self.supported_countries = ['BR']
       self.default_currency = 'BRL'
@@ -22,7 +25,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_order(post, options)
         add_amount(post, money)
-        add_payment(post, payment, options)
+        add_payment(post, payment)
         add_geolocation(post, options)
         add_extra_data(post, options)
 
@@ -56,11 +59,11 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_amount(post, money)
-        post[:amount] = amount(money)
+        post[:amount] = amount(money).to_f
       end
 
       def add_order(post, options)
-        post[:request_id] = options[:order_id]
+        post[:requestId] = options[:order_id]
       end
 
       def add_extra_data(post, options)
@@ -82,12 +85,12 @@ module ActiveMerchant #:nodoc:
         })
       end
 
-      def add_payment(post, payment, options)
+      def add_payment(post, payment)
         post.merge!({
           cardNumber: payment.number,
           cardholderName: payment.name,
           expirationMonth: payment.month,
-          expirationYear: format(payment.year, :two_digits),
+          expirationYear: format(payment.year, :two_digits).to_i,
           securityCode: payment.verification_value
         })
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -4,7 +4,7 @@ module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class AleloGateway < Gateway
       self.test_url = 'https://sandbox-api.alelo.com.br/alelo/sandbox/'
-      self.live_url = 'https://desenvolvedor.alelo.com.br'
+      self.live_url = 'https://api.alelo.com.br/alelo/prd/'
```
Agajanth, line 7: should we include a variable with the homologation URL?  like self.pre_live

lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -82,12 +82,12 @@ def add_geolocation(post, options)
         })
       end
 
-      def add_payment(post, payment, options)
+      def add_payment(post, payment)
         post.merge!({
           cardNumber: payment.number,
           cardholderName: payment.name,
           expirationMonth: payment.month,
```
Agajanth, line 89: shouldn't the month also be an integer?

Agajanth (approved):

lib/active_merchant/billing/gateways/alelo.rb
curiousepic: That would be good - we do this for some other gateways like `litle`

test/unit/gateways/alelo_test.rb
```diff
@@ -70,12 +71,12 @@ def test_successful_purchase_with_provided_credentials
       decrypted = JOSE::JWE.block_decrypt(secret_key, JSON.parse(data)['token']).first
       request = JSON.parse(decrypted, symbolize_names: true)
 
-      assert_equal @options[:order_id], request[:request_id]
-      assert_equal '1.00', request[:amount]
+      assert_equal @options[:order_id], request[:requestId]
+      assert_equal 1.0, request[:amount]
```
curiousepic, line 75: Is this really correct? are we sure only `.to_f` is sufficient for correct formatting? I wouldn't expect `1.0` to be proper.

lib/active_merchant/billing/gateways/alelo.rb
Heavyblade: The Month is indeed an integer

lib/active_merchant/billing/gateways/alelo.rb
Heavyblade: Done

test/unit/gateways/alelo_test.rb
Heavyblade: well, the money method applies the proper format but it returns '1.00' and Alelo needs float number, documentation says 'string' but based on the homologation work and the API example, I'm sending the return of amount method as a float number.

capture___sandbox-desenvolvedor_alelo_com_br_br

Heavyblade: mmm that should not be correct, I tested on a random local test with an amount of 100 that end up being translated as 1.0
tmux

curiousepic: Here is what I'm seeing:
Screen Shot 2022-09-26 at 4 19 57 PM

curiousepic: I do see `1.0` in the request, but that is what I get for the response.

Heavyblade: Ahhh the beauty of Alelo's Sandbox @curiousepic , no matter what you send in the body, using the uuid '5314 ...' you are always going to get that response:


Screen Shot 2022-09-26 at 3 28 15 PM

curiousepic: Aah, it looks like the sandbox always returns that amount 😩

curiousepic (approved):
👍  Thanks for the tweaks

curiousepic:
Closed with 68a7f5b0ae6ededb0a4ec127b3eaaecfb4700722

PR #4591, opened 2022-09-22
[internal] remove paypal commerce platform

mizinsky:
This will revert [f0fe324bf02cb3e68e18d99c244f9681c345a9fe](https://github.com/chargify/active_merchant/commit/f0fe324bf02cb3e68e18d99c244f9681c345a9fe)


peterbysh:
HI,

I know it's weird to respond to an old merge request but we're a happy user of Active Merchant and in the last month we've been contacted by Paypal and they're kindly asking / forcing us to move to Paypal commerce platform. It seems like they want to deprecate older checkouts and they might start forcing it onto other users as well. My question here is why was support for Paypal commerce platform removed?

peterbysh:
Ok thanks for the quick response, not sure if this is the right place to ask but there's no support to Paypal Ecommerce Platform in Active merchant in its current state right? But these code changes to me indicate that there once was.  I'm just trying to figure why this code either never made it to master or has been removed. Reasoning behind this is Paypal kindly asking / forcing me to switch from express checkout to commerce platform.

PR #4593, opened 2022-09-27
Adyen: Map Standard Error Codes

ajawadmirza:
Added new error codes in mapping for `STANDARD_ERROR_CODE` so that for dispatch `known_error_list` can support passing of standard error codes instead of gateway error code numbers. Unit test for each failover errors are added as well.

SER-266

Unit:
5352 tests, 76613 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop:

Remote:
125 tests, 443 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.8% passed


naashton (approved):
This looks good

PR #4602, opened 2022-10-04
Bambora: Adding support for Third party 3DS.

Heavyblade:
## Summary:

This PR adds support for third party 3DS2 authentication data on the Bambora (formerly Beanstream) gateway.

### Remote
Finished in 124.551138 seconds.
45 tests, 198 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.3333% passed

### Unit
Finished in 36.493983 seconds.
5354 tests, 76627 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/beanstream.rb
+++ b/lib/active_merchant/billing/gateways/beanstream.rb
@@ -76,6 +76,7 @@ module ActiveMerchant #:nodoc:
         add_transaction_type(post, :authorization)
         add_customer_ip(post, options)
         add_recurring_payment(post, options)
+        add_three_ds(post, options)
         commit(post)
       end
 
@@ -88,6 +89,7 @@ module ActiveMerchant #:nodoc:
         add_transaction_type(post, purchase_action(source))
         add_customer_ip(post, options)
         add_recurring_payment(post, options)
+        add_three_ds(post, options)
         commit(post)
       end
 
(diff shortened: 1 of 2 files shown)


jherreraa (approved):
LGTM! 👍

lib/active_merchant/billing/gateways/beanstream.rb
```diff
@@ -215,6 +217,22 @@ def scrub(transcript)
       def build_response(*args)
         Response.new(*args)
       end
+
+      def add_three_ds(post, options)
+        return unless three_d_secure = options[:three_d_secure]
+
+        post[:SecureXID] = (three_d_secure[:ds_transaction_id] || three_d_secure[:xid]) if three_d_secure.slice(:tres, :cuatro).values.any?
```
curiousepic, line 224: 🤔 Do these keywords actually work? They don't seem to in irb. In any case, I don't think we should rely on the order of elements in this hash for logic.

Heavyblade: Ohhh my bad on that @curiousepic , it should be 

`three_d_secure.slice(:ds_transaction_id, :xid).values.any?`

going to fix that inmediatly

basically the idea is to use transaction _id is present what I asume is 90%+ of the cases and if no present the fallback to xid.

curiousepic (approved):
👍 Thanks

curiousepic:
@Heavyblade Actually I'm getting a couple of remote tests failures including `test_successful_check_purchase`, are you getting those?

Heavyblade:
> @Heavyblade Actually I'm getting a couple of remote tests failures including `test_successful_check_purchase`, are you getting those?

Yup @curiousepic  those have been failing for some time now (last relevant commit 2019), I already update the remote tests

curiousepic:
Closed with https://github.com/activemerchant/active_merchant/commit/b899647bcd13aafda9a941b5399b5ae1112c9003

PR #4604, opened 2022-10-06
CommerceHub: Add new gateway

ajawadmirza:
Added new gateway `commerce_hub` along with remote and unit tests.

SER-207

Unit:
5360 tests, 76672 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:

Remote:
16 tests, 39 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
87.5% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/commerce_hub.rb
+++ b/lib/active_merchant/billing/gateways/commerce_hub.rb
@@ -1,0 +1,327 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CommerceHubGateway < Gateway
+      self.test_url = 'https://cert.api.fiservapps.com/ch'
+      self.live_url = 'https://prod.api.fiservapps.com/ch'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
+      self.homepage_url = 'https://developer.fiserv.com/product/CommerceHub'
+      self.display_name = 'CommerceHub'
+
+      STANDARD_ERROR_CODE_MAPPING = {}
+
+      SCHEDULED_REASON_TYPES = %w(recurring installment)
+      ENDPOINTS = {
+        'sale' => '/payments/v1/charges',
+        'authonly' => '/payments/v1/charges',
+        'capture' => '/payments/v1/charges',
+        'void' => '/payments/v1/cancels',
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -0,0 +1,327 @@
...
+          response,
+          authorization: authorization_from(action, response),
+          test: test?,
+          error_code: error_code_from(response)
+        )
+      end
+
+      def handle_response(response)
+        case response.code.to_i
+        when 200...300, 400, 401
```
naashton, line 302: I'm getting a bunch of `429` error codes when running the tests. Might be worth adding that to the `when` clause here.

lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -0,0 +1,327 @@
...
+        add_transaction_interaction(post, options)
+        commit('vault', post, options)
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
```
naashton, line 83: This gateway supports $0 auth with the `charges` endpoint. We should implement this instead of the traditional auth/void. Here's the [documentation](https://developer.fiserv.com/product/CommerceHub/docs/?path=docs/Resources/API-Documents/Payments/Charges.md&branch=main).

naashton (commented):
Overall, this looks pretty good. I'm also getting 500's from the gateway, so maybe we'll see what Fiserv says and I'll give it another review.

lib/active_merchant/billing/gateways/commerce_hub.rb
ajawadmirza: Thanks @naashton, the change is updated 👍

lib/active_merchant/billing/gateways/commerce_hub.rb
ajawadmirza: I have updated the `handle_response` to include 429 as well. Can you also share the error you are getting because the tests are passing for me except for 3 tests: `test_successful_purchase_with_apple_pay`, `test_successful_purchase_with_google_pay`, and `test_successful_store_with_purchase`

PR #4605, opened 2022-10-07
D Local: Add support for `original_order_id` field

rachelkirk:
CER-212

Remote:
34 tests, 92 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 97.0588% passed
*Failing test is also failing on master

Unit:
37 tests, 158 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Local:
5355 tests, 76630 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

NOTE: I looked into test cards on the DLocal docs to try to fix the failing test, and all card payments are supposed to pass by default in sandbox mode. The only reason for failure should be if you pass in a code in the description. I'm not sure why this particular card type is causing a failure.

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -87,6 +87,7 @@ module ActiveMerchant #:nodoc:
         add_card(post, card, action, options)
         add_additional_data(post, options)
         post[:order_id] = options[:order_id] || generate_unique_id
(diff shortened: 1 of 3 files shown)


jcreiff (approved):
Pretty straightforward - looks good to me!

I also saw the single remote failure, but it looks unrelated

PR #4608, opened 2022-10-18
DLocal: Add Network Tokens

gasb150:
Summary:
To support NetworkTokens this PR adds the 
needed parameters in the add_cart method 
and the supports_network_tokenization? 
method is set in true.

Unit Test:
Finished in 0.256649 seconds.
40 tests, 172 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Test:
Finished in 43.736933 seconds.
40 tests, 108 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -69,6 +69,10 @@ module ActiveMerchant #:nodoc:
         true
       end
 
+      def supports_network_tokenization?
+        true
+      end
+
       def scrub(transcript)
         transcript.
           gsub(%r((X-Trans-Key: )\w+), '\1[FILTERED]').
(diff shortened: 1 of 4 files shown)


test/unit/gateways/card_stream_test.rb
```diff
@@ -33,7 +33,7 @@ def setup
       three_d_secure: {
         enrolled: 'true',
         authentication_response_status: 'Y',
-        eci: 05,
+        eci: '05',
```
javierpedrozaing, line 36: ❓ 
Why this value change ?

javierpedrozaing (commented):
LGTM just I left a question

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -155,11 +159,22 @@ def parse_house_number(address)
 
       def add_card(post, card, action, options = {})
         post[:card] = {}
+        if card.is_a?(NetworkTokenizationCreditCard)
+          post[:card][:network_token] = card.number
+          post[:card][:cryptogram] = card.payment_cryptogram
+          post[:card][:eci] = card.eci
+          # used case of Network Token: 'CARD_ON_FILE', 'SUBSCRIPTION', 'UNSCHEDULED_CARD_ON_FILE'
+          post[:card][:stored_credential_type] = options[:stored_credential_type] if options[:stored_credential_type]
+          # required for MC debit recurrent in BR 'FIRST'(subsecuence Payments) . 'USED' an inital payment
```
curiousepic, line 168: this comment seems reversed? shouldn't "first" be an initial payment?

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -155,11 +159,22 @@ def parse_house_number(address)
 
       def add_card(post, card, action, options = {})
         post[:card] = {}
+        if card.is_a?(NetworkTokenizationCreditCard)
+          post[:card][:network_token] = card.number
+          post[:card][:cryptogram] = card.payment_cryptogram
+          post[:card][:eci] = card.eci
+          # used case of Network Token: 'CARD_ON_FILE', 'SUBSCRIPTION', 'UNSCHEDULED_CARD_ON_FILE'
+          post[:card][:stored_credential_type] = options[:stored_credential_type] if options[:stored_credential_type]
```
curiousepic, line 167: I think we will want to utilize the existing stored credential hash structure, and map those values to what is expected here.

lib/active_merchant/billing/gateways/d_local.rb
gasb150: yes, sorry, I have to fix it.

lib/active_merchant/billing/gateways/d_local.rb
gasb150: Yes, you are right, I will use the hash structure.

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -155,11 +159,30 @@ def parse_house_number(address)
...
+          post[:card][:network_token] = card.number
+          post[:card][:cryptogram] = card.payment_cryptogram
+          post[:card][:eci] = card.eci
+          # used case of Network Token: 'CARD_ON_FILE', 'SUBSCRIPTION', 'UNSCHEDULED_CARD_ON_FILE'
+          if options.dig(:stored_credential, :reason_type) == 'unscheduled'
+            if options.dig(:stored_credential, :initiator) == 'merchant'
+              post[:card][:stored_credential_type] = 'UNSCHEDULED_CARD_ON_FILE'
+            else
+              post[:card][:stored_credential_type] = 'CARD_ON_FILE'
+            end
```
curiousepic, line 172: I just want to double check this condition is correct, since it specifies "Unscheduled" only for one option here, even though both possible conditions are triggered by the `unscheduled` stored cred flag.

curiousepic (approved):
👍 Thanks

PR #4611, opened 2022-10-19
CheckoutV2: Add support for `merchant_initiated_transaction_id`

rachelkirk:
CER-219

For this field, the customer should provide a previous network transaction id from the CheckoutV2 gateway. It is the customer’s responsibility to store and manage the network transaction ids. When a value is present, this field should override the existing stored credential logic and add the stored credential values to the transaction request.

Local:
5363 tests, 76675 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
50 tests, 278 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
74 tests, 188 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 98.6486% passed
Note: Failing test is also failing on master and unrelated to this change.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -17,6 +17,8 @@
 * Beanstream: Adding Third Party 3DS fields [heavyblade] #4602
 * Borgun: Add support for 3DS preauth [ajawadmirza] #4603
 * Accept both formats of Canadian routing numbers [molbrown] #4568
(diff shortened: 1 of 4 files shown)


jcreiff (approved):
This looks good! Some nice refactoring in there too ⭐

PR #4612, opened 2022-10-21
Stripe: Fix header for test account deletion call

curiousepic:
Previously using Bearer Authorization instead of Basic. Previously, the call did not seem to result in actual deletions, whereas after the switch there were multiple deletions.


aenand (approved):
LGTM

PR #4614, opened 2022-10-25
Alelo: Trigger access token refresh on 404

curiousepic:
In production, expired UUIDs may return as a 404 instead of a 401. This expands the condition for refreshing access tokens to include a 404. No Remote test is added because sending a bad UUID seems to trigger a 500 error instead of a 404.

Remote:
17 tests, 43 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
23 tests, 105 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


dsmcclain (approved):
All tests passing, good to 🚢

Heavyblade (approved):
@curiousepic  looks good 👍

PR #4615, opened 2022-10-26
Redsys: enable NTID generation with zero-value verify

jcreiff:
A temporary fix to enable NTID generation for cards without an NTID before the Nov 1 Visa deadline

CER-234


lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -266,9 +266,13 @@ def refund(money, authorization, options = {})
       end
 
       def verify(creditcard, options = {})
-        MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, creditcard, options) }
-          r.process(:ignore_result) { void(r.authorization, options) }
+        if options[:sca_exemption_behavior_override] = 'endpoint_and_ntid'
```
jcreiff, line 269: Does this value get passed to `options` when included as a gateway specific field?

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -357,14 +361,17 @@ def add_stored_credential_options(data, options)
...
           data[:DS_MERCHANT_COF_INI] = 'N'
           data[:DS_MERCHANT_COF_TXNID] = options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id]
+          # recurring_cof_type = '0'
         end
 
         case options[:stored_credential][:reason_type]
         when 'recurring'
-          data[:DS_MERCHANT_COF_TYPE] = 'R'
+          data[:DS_MERCHANT_COF_TYPE] = "R"
+          # data[:DS_MERCHANT_COF_TYPE] = recurring_cof_type
```
jcreiff, line 374: This conditionally set variable would satisfy the direction given by Banco Sabadell, but stored credentials tests start failing when this logic is included

lib/active_merchant/billing/gateways/redsys.rb
jcreiff: Looks like the answer is yes, if we add the `on: :all` property.
We will also need to do this for `sca_exemption`

test/remote/gateways/remote_redsys_sha256_test.rb
```diff
@@ -212,6 +212,22 @@ def test_successful_3ds2_purchase_with_mit_exemption
     assert_match 'MIT', response.params['ds_excep_sca']
   end
 
+  def test_successful_ntid_generation_with_verify
```
dsmcclain, line 215: perhaps you can add a comment here to explain that the gateway's sandbox does not actually generate the ntid that we expect to see in production.

lib/active_merchant/billing/gateways/redsys.rb
```diff
@@ -91,7 +91,7 @@ class RedsysGateway < Gateway
       # More operations are supported by the gateway itself, but
       # are not supported in this library.
       SUPPORTED_TRANSACTIONS = {
-        purchase:   'A',
+        purchase:   '0',
```
dsmcclain, line 94: Based on our in-person conversation, we will move forward with this change that will affect all purchase transactions on this gateway, because Banco Sabadell has assured us that `A` is not a valid transaction type and we should be using `0` instead.

dsmcclain (approved):
Getting only timeouts from the gateway sandbox, so unfortunately it is hard to verify these changes. However I acknowledge that there are time constraints here and the gateway has assured us that this is the correct path to follow, so we will trust them and monitor the changes once they go live.

I have requested a comment explaining your new test. Once you add that, update the changelog and you are free to merge.

PR #4617, opened 2022-10-27
Adyen: Field support for Level 2 and level 3 information

edgarv09:
## Description

Adding support in Adyen gateway to append level 2 and level 3 information that provides commercial shoppers with additional information about purchases on their card statements. 

### Remote Tests:
```
Finished in 139.349852 seconds.
127 tests, 423 assertions, 14 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
88.9764% passed
```

### Unit Tests:
```
Finished in 0.117629 seconds.
103 tests, 522 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -66,6 +66,8 @@ module ActiveMerchant #:nodoc:
         add_recurring_contract(post, options)
         add_network_transaction_reference(post, options)
         add_application_info(post, options)
+        add_level_2_data(post, options)
+        add_level_3_data(post, options)
         commit('authorise', post, options)
       end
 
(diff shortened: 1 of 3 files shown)


javierpedrozaing (commented):
Great job, would be great if we can include some test for purchase transaction

edgarv09:
> Great job, would be great if we can include some test for purchase transaction

Yeah!! sure, already done.

javierpedrozaing (commented):
LGTM

edgarv09:
> Looks good, just needs some cleanup of the commit title and message. Thanks!

Cleanup of commit message and title already done.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -246,6 +248,45 @@ def add_extra_data(post, payment, options)
...
 
+      def extract_and_transform(mapper, from)
+        mapper.each_with_object({}) do |key_map, hsh|
+          key, item_key = key_map[0], key_map[1]
+          hsh[key] = from[item_key.to_sym]
+        end
+      end
+
+      def add_level_2_data(post, options)
+        mapper = {
```
curiousepic, line 259: I just had a thought - given the complexity of these methods, it would be good to begin them with a guard clause that will exit early if there is no level 2/3 data available.
```suggestion
        return unless options[:level_2_data]

        mapper = {
```

curiousepic: I guess for this one it is slightly more complex due to the tax and cutomer reference fields not being contained in the level_2_data subhash...

edgarv09: Ok, perfect. I understood the point, you are right it would be better to have it. 

note:
early return was added.

PR #4618, opened 2022-10-27
Reach: adding gateway

Heavyblade:
## Summary:

Adding Reach gateway with authorize, purchase and capture calls, also small correction on the CardStream local test to complain with Rubocop warnings.

#### Remote Test:
Finished in 6.627757 seconds.
7 tests, 14 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

#### Unit Tests:
5365 tests, 76686 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

#### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -1,0 +1,162 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class ReachGateway < Gateway
+      # TODO: Things to check
+      # * The list of three digit fractions but only accept 2
+      # * Not sure the list of countries and currencies
+
+      self.test_url = 'https://checkout.rch.how/'
+      self.live_url = 'https://checkout.rch.io/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa diners_club american_express jcb master discover maestro]
+
+      self.homepage_url = 'https://www.withreach.com/'
+      self.display_name = 'Reach'
+      self.currencies_without_fractions = %w(BIF BYR CLF CLP CVE DJF GNF ISK JPY KMF KRW PYG RWF UGX UYI VND VUV XAF XOF XPF IDR MGA MRO)
+
+      API_VERSION = 'v2.22'.freeze
+      STANDARD_ERROR_CODE_MAPPING = {}
+      PAYMENT_METHOD_MAP = {
+        american_express: 'AMEX',
+        cabal: 'CABAL',
+        check: 'ACH',
+        dankort: 'DANKORT',
+        diners_club: 'DINERS',
+        discover: 'DISC',
+        elo: 'ELO',
+        jcb: 'JCB',
+        maestro: 'MAESTRO',
+        master: 'MC',
+        naranja: 'NARANJA',
+        visa: 'VISA'
+      }
+
+      def initialize(options = {})
+        requires!(options, :merchant_id, :secret)
+        super
+      end
+
+      def authorize(money, payment, options = {})
+        request = build_checkout_request(money, payment, options)
+        add_customer_data(request, options, payment)
+
+        post = { request: request, card: add_payment(payment) }
+        commit('checkout', post)
+      end
+
+      def purchase(money, payment, options = {})
+        options[:capture] = true
+        authorize(money, payment, options)
+      end
+
+      def capture(money, authorization, options = {})
+        post = { request: { MerchantId: @options[:merchant_id], OrderId: authorization } }
+        commit('capture', post)
+      end
+
+      private
+
+      def build_checkout_request(amount, payment, options)
+        {
+          MerchantId: @options[:merchant_id],
+          ReferenceId: options[:order_id],
+          ConsumerCurrency: options[:currency] || currency(options[:amount]),
+          Capture: options[:capture] || false,
+          PaymentMethod: PAYMENT_METHOD_MAP[payment.brand.to_sym],
+          Items: [
+            Sku: options[:item_sku] || SecureRandom.alphanumeric,
+            ConsumerPrice: amount,
+            Quantity: (options[:item_quantity] || 1)
+          ],
+          ViaAgent: true # Indicates this is server to server API call
+        }
+      end
+
+      def add_payment(payment)
+        {
+          Name: payment.name,
+          Number: payment.number,
+          Expiry: { Month: payment.month, Year: payment.year },
+          VerificationCode: payment.verification_value
+        }
+      end
+
+      def add_customer_data(request, options, payment)
+        address = options[:billing_address] || options[:address]
+
+        return if address.blank?
+
+        request[:Consumer] = {
+          Name: payment.respond_to?(:name) ? payment.name : address[:name],
+          Email: options[:email],
+          Address: address[:address1],
+          City: address[:city],
+          Country: address[:country]
+        }.compact
+      end
+
+      def sign_body(body)
+        Base64.strict_encode64(OpenSSL::HMAC.digest('sha256', @options[:secret].encode('utf-8'), body.encode('utf-8')))
+      end
+
+      def parse(body)
+        hash_response = URI.decode_www_form(body).to_h.transform_keys!(&:to_sym)
+        hash_response[:response] = JSON.parse(hash_response[:response], symbolize_names: true)
+
+        hash_response
+      end
+
+      def format_and_sign(post)
+        post[:request] = post[:request].to_json
+        post[:card] = post[:card].to_json if post[:card].present?
+        post[:signature] = sign_body(post[:request])
+        post
+      end
+
+      def commit(action, parameters)
+        body = post_data format_and_sign(parameters)
+        raw_response = ssl_post url(action), body
+        response = parse(raw_response)
+
+        Response.new(
+          success_from(response),
+          message_from(response) || '',
+          response,
+          authorization: authorization_from(response[:response]),
+          # avs_result: AVSResult.new(code: response['some_avs_response_key']),
+          # cvv_result: CVVResult.new(response['some_cvv_response_key']),
+          test: test?,
+          error_code: error_code_from(response)
+        )
+      rescue ActiveMerchant::ResponseError => e
+        Response.new(false, (e.response.body.present? ? e.response.body : e.response.msg), {}, test: test?)
+      end
+
+      def success_from(response)
+        response.dig(:response, :Error).blank?
+      end
+
+      def message_from(response)
+        success_from(response) ? '' : response.dig(:response, :Error, :ReasonCode)
+      end
+
+      def authorization_from(response)
+        response[:OrderId]
+      end
+
+      def post_data(params)
+        params.map { |k, v| "#{k}=#{CGI.escape(v.to_s)}" }.join('&')
+      end
+
+      def error_code_from(response)
+        response[:response][:Error][:Code] unless success_from(response)
+      end
+
+      def url(action)
+        "#{test? ? test_url : live_url}#{API_VERSION}/#{action}"
+      end
+    end
+  end
+end
--- a/test/unit/gateways/card_stream_test.rb
+++ b/test/unit/gateways/card_stream_test.rb
@@ -33,7 +33,7 @@ class CardStreamTest < Test::Unit::TestCase
       three_d_secure: {
         enrolled: 'true',
         authentication_response_status: 'Y',
-        eci: 05,
+        eci: '05',
         cavv: 'Y2FyZGluYWxjb21tZXJjZWF1dGg',
         xid: '362DF058-6061-47F1-A504-CACCBDF422B7'
       }
@@ -330,7 +330,7 @@ class CardStreamTest < Test::Unit::TestCase
       assert_match(/threeDSRequired=Y/, data)
       assert_match(/threeDSEnrolled=Y/, data)
       assert_match(/threeDSAuthenticated=Y/, data)
-      assert_match(/threeDSECI=5/, data)
+      assert_match(/threeDSECI=05/, data)
       assert_match(/threeDSCAVV=Y2FyZGluYWxjb21tZXJjZWF1dGg/, data)
       assert_match(/threeDSXID=362DF058-6061-47F1-A504-CACCBDF422B7/, data)
     end
--- a/test/unit/gateways/reach_test.rb
+++ b/test/unit/gateways/reach_test.rb
@@ -1,0 +1,127 @@
+require 'test_helper'
+
+class ReachTest < Test::Unit::TestCase
+  include CommStub
+
+  def setup
+    @gateway = ReachGateway.new(fixtures(:reach))
+    @credit_card = credit_card
+    @amount = 100
+
+    @options = {
+      email: 'johndoe@reach.com',
+      order_id: '123',
+      currency: 'USD',
+      billing_address: {
+        address1: '1670',
+        address2: '1670 NW 82ND AVE',
+        city: 'Miami',
+        state: 'FL',
+        zip: '32191',
+        country: 'US'
+      }
+    }
+  end
+
+  def test_required_merchant_id_and_secret
+    error = assert_raises(ArgumentError) { ReachGateway.new }
+    assert_equal 'Missing required parameter: merchant_id', error.message
+  end
+
+  def test_supported_card_types
+    assert_equal ReachGateway.supported_cardtypes, %i[visa diners_club american_express jcb master discover maestro]
+  end
+
+  def test_should_be_able_format_a_request
+    post = {
+      request: { someId: 'abc123' },
+      card: { number: '12132323', name: 'John doe' }
+    }
+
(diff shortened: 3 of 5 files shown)


lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,174 @@
...
+
+      def scrub(transcript)
+        transcript
+      end
+
+      private
+
+      def build_checkout_request(amount, payment, options)
+        {
+          MerchantId: @options[:merchant_id],
```
javierpedrozaing, line 75: ❓ 
Why `@options` instead `options` param?

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,174 @@
...
+          Items: [
+            Sku: SecureRandom.alphanumeric,
+            ConsumerPrice: amount,
+            Quantity: 1
+          ],
+          ViaAgent: true # Indicates this is server to server API call
+        }
+      end
+
+      def add_payment(request, payment)
```
javierpedrozaing, line 89: 💬 
we can remove `request` param, this is not used in the function

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,174 @@
...
+        post = { request: request, card: add_payment(request, payment) }
+        commit('checkout', post)
+      end
+
+      def purchase(money, payment, options = {})
+        options[:capture] = true
+        authorize(money, payment, options)
+      end
+
+      def capture(money, authorization, options = {})
```
javierpedrozaing, line 56: 💬 
we can remove `options` param, this is not used in the function

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,174 @@
...
+      end
+
+      def success_from(response)
+        response[:response][:Error].blank?
+      end
+
+      def message_from(response)
+        return response[:response][:Error][:ReasonCode] unless success_from(response)
+
+        ''
```
javierpedrozaing, line 154: ❓ 
What is this extra param for?

javierpedrozaing (commented):
Nice work, Just I left some comments and questions

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: you are right 👍

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: Nop, beacase that is the expected interface used by core to call the method.

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: @javierpedrozaing a this point is a place holder, currently the when a transaccion is a success you only get back an OrderId and Signature but not a particular message or code

lib/active_merchant/billing/gateways/reach.rb
javierpedrozaing: Ohh I can see

lib/active_merchant/billing/gateways/reach.rb
javierpedrozaing: thanks

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,174 @@
...
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa diners_club american_express jcb master discover maestro]
+
+      self.homepage_url = 'https://www.withreach.com/'
+      self.display_name = 'Reach'
+      self.currencies_without_fractions = %w(BIF BYR CLF CLP CVE DJF GNF ISK JPY KMF KRW PYG RWF UGX UYI VND VUV XAF XOF XPF IDR MGA MRO)
+
+      API_VERSION = 'v2.21'.freeze
```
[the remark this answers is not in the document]
Heavyblade: @gasb150 is just a good practice to make strings constants a freeze to have the same effect that ruby symbols have.

or maybe we cant start to use "# frozen_string_literal: true" in the AM files.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,174 @@
...
+        commit('checkout', post)
+      end
+
+      def purchase(money, payment, options = {})
+        options[:capture] = true
+        authorize(money, payment, options)
+      end
+
+      def capture(money, authorization, options = {})
+        merchant_id, order_id = *(authorization.split '#')
```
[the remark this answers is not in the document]
Heavyblade: Ahhh you are right, thanks for the catch 👍

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
+require 'base64'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class ReachGateway < Gateway
+      # TODO: Things to check
+      # * The list of three digit fractions but only accept 2
+      # * Not sure the list of countries and currencies
```
curiousepic, line 8: Are these todos still true?

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+        commit('checkout', post)
+      end
+
+      def purchase(money, payment, options = {})
+        options[:capture] = true
+        authorize(money, payment, options)
+      end
+
+      def capture(money, authorization, options = {})
+        merchant_id, order_id = authorization.split '#'
```
curiousepic, line 57: minor thing, but if the `order_id` we receive in a response, which is added to the authorization string on line 160, is a gateway-specific identifier instead of the normal concept of a merchant-defined `order_id` used in the option hash, we should rename it here to something like `gateway_transaction_id` to prevent confusion.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+        {
+          MerchantId: @options[:merchant_id],
+          ReferenceId: options[:order_id],
+          ConsumerCurrency: options[:currency] || currency(options[:amount]),
+          Capture: options[:capture] || false,
+          PaymentMethod: PAYMENT_METHOD_MAP[payment.brand.to_sym],
+          Items: [
+            Sku: SecureRandom.alphanumeric,
+            ConsumerPrice: amount,
+            Quantity: 1
```
curiousepic, line 83: Do we know it is ok to hard-code Sku and Quantity or should these be customizable with `options`?

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+        raw_response = ssl_post url(action), body
+        response = parse(raw_response)
+
+        Response.new(
+          success_from(response),
+          message_from(response),
+          response,
+          authorization: authorization_from(response[:response]),
+          # avs_result: AVSResult.new(code: response['some_avs_response_key']),
+          # cvv_result: CVVResult.new(response['some_cvv_response_key']),
```
curiousepic, line 141: If we know they do not support these, we can remove the comments.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+      end
+
+      def success_from(response)
+        response[:response][:Error].blank?
+      end
+
+      def message_from(response)
+        return response[:response][:Error][:ReasonCode] unless success_from(response)
+
+        ''
```
curiousepic, line 156: I find this a little awkward. Could we instead just `.dig` for the error code?

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+
+      def capture(money, authorization, options = {})
+        merchant_id, order_id = authorization.split '#'
+        post = { request: { MerchantId: merchant_id, OrderId: order_id } }
+
+        commit('capture', post)
+      end
+
+      def supports_scrubbing?
+        true
```
curiousepic, line 64: It doesn't look like it...

test/unit/gateways/card_stream_test.rb
```diff
@@ -33,7 +33,7 @@ def setup
       three_d_secure: {
         enrolled: 'true',
         authentication_response_status: 'Y',
-        eci: 05,
+        eci: '05',
```
curiousepic, line 36: I thought this was previously changed to this for some reason? was Rubocop complaining about it?

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: 👍 Agree with you @curiousepic

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: @curiousepic the sku is a required parameter so my thinking is => Given this is just the base implementation I'm just going to set fixed/random values, if in the future we have Level III information we can use that to pass real items data down to Reach.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
+require 'base64'
```
naashton, line 1: I don't think you need this since it's already included in the `ActiveMerchant` module.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa diners_club american_express jcb master discover maestro]
+
+      self.homepage_url = 'https://www.withreach.com/'
+      self.display_name = 'Reach'
+      self.currencies_without_fractions = %w(BIF BYR CLF CLP CVE DJF GNF ISK JPY KMF KRW PYG RWF UGX UYI VND VUV XAF XOF XPF IDR MGA MRO)
+
+      API_VERSION = 'v2.21'.freeze
```
naashton, line 21: We can update this to `'v2.22'` from Charlotte's comment

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+      def authorization_from(response)
+        "#{@options[:merchant_id]}##{response[:OrderId]}"
+      end
+
+      def post_data(params)
+        params.map { |k, v| "#{k}=#{CGI.escape(v.to_s)}" }.join('&')
+      end
+
+      def error_code_from(response)
+        response[:response][:Error][:Code] unless success_from(response)
```
naashton, line 168: I agree with @curiousepic's suggestion to use `dig` to protect against any exceptions where the nested key's aren't present in the `response`.

test/unit/gateways/reach_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    assert_equal '12132323', JSON.parse(formatted[:card])['number']
+    assert formatted[:signature].present?
+  end
+
+  def test_successfully_build_a_purchase
+    stub_comms do
+      @gateway.authorize(@amount, @credit_card, @options)
+    end.check_request do |_endpoint, data, _headers|
+      request = JSON.parse(URI.decode_www_form(data)[0][1])
+      card = JSON.parse(URI.decode_www_form(data)[1][1])
```
naashton, line 57: If we stub the decoded responses then we can probably avoid decoding them here.

naashton (changes requested):
Just a few changes from me, along with David's suggestions.

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: @curiousepic I wanted to leaved there as reminder that there is a pending question for Reach to enable if possible that feature.

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: 👍

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,176 @@
...
+      end
+
+      def message_from(response)
+        return response[:response][:Error][:ReasonCode] unless success_from(response)
+
+        ''
+      end
+
+      def authorization_from(response)
+        "#{@options[:merchant_id]}##{response[:OrderId]}"
```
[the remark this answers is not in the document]
Heavyblade: You are right @curiousepic  forgot that @options is going to be available any way on capture/void ... going to make the fix

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: mm @curiousepic I left the method there because the work to add the scrubbing is in another story and is almos done, do you think is necessary to remote that line from here and add it on that PR ?

test/unit/gateways/card_stream_test.rb
Heavyblade: yup @curiousepic  I change due a Rubocop complain.

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: yes removing it 👍

test/unit/gateways/reach_test.rb
```diff
@@ -0,0 +1,86 @@
...
+      @gateway.purchase(@amount, @credit_card, @options)
+    end.check_request do |_endpoint, data, _headers|
+      request = JSON.parse(URI.decode_www_form(data)[0][1])
+      assert_equal true, request['Capture']
+    end.respond_with(successful_purchase_response)
+  end
+
+  private
+
+  def successful_purchase_response
```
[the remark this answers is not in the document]
Heavyblade: Mmm I tried @naashton  something like that,  but I  got decode errors because `stub_com` is sutubbing the call to `ssl_post` and the result is going to be passed to the `parse` method, so if I set the stubbed return value to something that is already parsed I'm going to get decode error.

The only way that I can think of to implement this is to stub out the `parse` method but i'm not sure if formatting is worth no using the real code that handles the response.

maybe you know  other option/angle to implement that ?

test/unit/gateways/reach_test.rb
Heavyblade: Same argument as above @naashton

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: yup @curiousepic, need to clarify that on grooming meeting

lib/active_merchant/billing/gateways/reach.rb
curiousepic: Technically that would be proper, if that piece will be merged in a separate commit.

lib/active_merchant/billing/gateways/reach.rb
curiousepic: Interesting, alright. Any opinion here @naashton ?

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: ok I'm going to remove it 👍

lib/active_merchant/billing/gateways/reach.rb
curiousepic: Yeah I think that's a preferable approach.

Heavyblade: @naashton  if we are going to open that door from the begining should we also include Quantity as a posible GSF field ?

Heavyblade: Done @curiousepic  / @naashton change is donde, going to tadd comments to de GSF ticket to add them

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,162 @@
...
+        {
+          MerchantId: @options[:merchant_id],
+          ReferenceId: options[:order_id],
+          ConsumerCurrency: options[:currency] || currency(options[:amount]),
+          Capture: options[:capture] || false,
+          PaymentMethod: PAYMENT_METHOD_MAP[payment.brand.to_sym],
+          Items: [
+            Sku: options[:item_sku] || SecureRandom.alphanumeric,
+            ConsumerPrice: amount,
+            Quantity: (options[:item_quanity] || 1)
```
curiousepic, line 71:
```suggestion
            Quantity: (options[:item_quantity] || 1)
```
Right?
Otherwise looking good

Heavyblade: Ohhhh yes thanks @curiousepic

Heavyblade: Fixed @curiousepic

naashton (approved):
Awesome work!

PR #4619, opened 2022-11-01
IPG Gateway: override store id

aenand:
ECS-2655

The IPG gateway requires a `store_id` to be
used as part of the authentication. This
value is correlates to a given merchant and
is currently saved on the gateway initalize.
By saving it only then, it prevents merchants
with submerchants from using this gateway.

This commit adds a new optional field that
will override the saved `store_id` with one
passed in for that merchant.

Test Summary
Remote:
18 tests, 54 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


naashton (approved):
Tests pass and this looks good!

PR #4620, opened 2022-11-03
Reach: add Scrubing

gasb150:
Summary:
    In order to scrub fields in the transaction transript, this commit
    adds fields to be filtered.
    
    Remote Test:
    Finished in 10.189499 seconds.
    7 tests, 18 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed
    
    Unit Test:
    6 tests, 27 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
    100% passed
    
    Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -56,6 +56,18 @@ module ActiveMerchant #:nodoc:
         commit('capture', post)
       end
 
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r(((MerchantId%22%3A%22)[\w-]+)), '\2[FILTERED]').
+          gsub(%r((signature=)[\w%]+), '\1[FILTERED]\2').
+          gsub(%r((Number%22%3A%22)[\d]+), '\1[FILTERED]\2').
+          gsub(%r((VerificationCode%22%3A)[\d]+), '\1[FILTERED]\2')
+      end
+
       private
 
       def build_checkout_request(amount, payment, options)
--- a/test/remote/gateways/remote_reach_test.rb
+++ b/test/remote/gateways/remote_reach_test.rb
@@ -75,13 +75,14 @@ class RemoteReachTest < Test::Unit::TestCase
   end
 
   def test_transcript_scrubbing
-    # transcript = capture_transcript(@gateway) do
-    #   @gateway.purchase(@amount, @credit_card, @options)
-    # end
-    # transcript = @gateway.scrub(transcript)
-    #
-    # assert_scrubbed(@credit_card.number, transcript)
-    # assert_scrubbed(@credit_card.verification_value, transcript)
-    # assert_scrubbed(@gateway.options[:password], transcript)
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+
+    transcript = @gateway.scrub(transcript)
+    assert_scrubbed(@credit_card.number, transcript)
+    assert_scrubbed(@credit_card.verification_value, transcript)
+    assert_scrubbed(@gateway.options[:merchant_id], transcript)
+    assert_scrubbed(@gateway.options[:secret], transcript)
   end
 end
--- a/test/unit/gateways/reach_test.rb
+++ b/test/unit/gateways/reach_test.rb
@@ -119,9 +119,57 @@ class ReachTest < Test::Unit::TestCase
     refute @gateway.send(:success_from, response)
   end
 
+  def test_scrub
+    assert @gateway.supports_scrubbing?
+
+    assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
+  end
+
   private
 
   def successful_purchase_response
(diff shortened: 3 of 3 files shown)


lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,178 @@
...
+        commit('capture', post)
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r(((MerchantId%22%3A%22)[\w-]+)), '\2[FILTERED]').
```
Heavyblade, line 69: 💬  Comment:

Not sure if MercahtId is senstive field is not ecrypted on the core implementation for example

test/remote/gateways/remote_reach_test.rb
```diff
@@ -0,0 +1,88 @@
...
+    assert_equal 'Not Found', response.message
+  end
+
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+
+    transcript = @gateway.scrub(transcript)
+    assert_scrubbed(@credit_card.number, transcript)
```
Heavyblade, line 83: 💬  Comment:

Really nice approach 💪

test/unit/transcripts/reach_purchase
```diff
@@ -0,0 +1,16 @@
...
+-> "Connection: close\r\n"
+-> "Server: ipCheckoutApi/unreleased ibiHttpServer\r\n"
+-> "Strict-Transport-Security: max-age=60000\r\n"
+-> "Cache-Control: no-cache\r\n"
+-> "Access-Control-Allow-Origin: *\r\n"
+-> "\r\n"
+reading 235 bytes...
+-> "response=%7B%22OrderId%22%3A%22621a0c76-69fb-4c05-854a-e7e731759ad3%22%2C%22UnderReview%22%3Afalse%2C%22Authorized%22%3Atrue%2C%22Completed%22%3Afalse%2C%22Captured%22%3Afalse%7D&signature=23475signature23123%3D"
+read 235 bytes
+Conn close
```
Heavyblade, line 16: 💬  Comment:

Small complain from Rubocop and Github, transcript files require and extra line

test/unit/transcripts/reach_purchase_scrubbed
```diff
@@ -0,0 +1,16 @@
...
+-> "Connection: close\r\n"
+-> "Server: ipCheckoutApi/unreleased ibiHttpServer\r\n"
+-> "Strict-Transport-Security: max-age=60000\r\n"
+-> "Cache-Control: no-cache\r\n"
+-> "Access-Control-Allow-Origin: *\r\n"
+-> "\r\n"
+reading 235 bytes...
+-> "response=%7B%22OrderId%22%3A%22621a0c76-69fb-4c05-854a-e7e731759ad3%22%2C%22UnderReview%22%3Afalse%2C%22Authorized%22%3Atrue%2C%22Completed%22%3Afalse%2C%22Captured%22%3Afalse%7D&signature=[FILTERED]"
+read 235 bytes
+Conn close
```
Heavyblade, line 16: 💬  Comment:

Same here

Heavyblade (approved):
@gasb150 well done is basically done, leave a couple of comments.

test/remote/gateways/remote_reach_test.rb
gasb150: thanks 😊

test/unit/transcripts/reach_purchase
gasb150: yes, I missed it.

curiousepic (approved):
LGTM, @naashton for confirmation & merge

test/unit/transcripts/reach_purchase
naashton, line 16: For the transcript, you can just create a method and add the contents here to the method like so: https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/stripe_test.rb#L1609L1636

Then you can just reference the method instead of using `File.read`.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -56,6 +56,18 @@ def capture(money, authorization, options = {})
...
+          gsub(%r((signature=)[\w%]+), '\1[FILTERED]\2').
+          gsub(%r((Number%22%3A%22)[\d]+), '\1[FILTERED]\2').
+          gsub(%r((VerificationCode%22%3A)[\d]+), '\1[FILTERED]\2')
```
naashton, line 68: All of these look good, but we might need to consider other payment method types and if the transcript for those will have additional data that we need to scrub.

naashton (changes requested):
This looks pretty good. Just one "food for thought" comment and a requested change to put the transcript dump in a method instead of adding 2 new files.

lib/active_merchant/billing/gateways/reach.rb
Heavyblade: @naashton for Reach the initial implementation is only going to consider CC, if new payment methods ara added in the future, the natural thing to do is to update the scrub method.

naashton: Yes, I think you are correct that we need to handle these after we implement alternate payment methods. Just wanted to make sure I called it out, mostly so that I don't forget 😅

test/unit/transcripts/reach_purchase
Heavyblade: well @naashton from the pure Engineering point of view I don't like it, but from the consistency point of view you are totally right, is not perfect, but is an well stablished pattern, maybe we should use the file.read approach just for really big responses.

@gasb150 we should perform that change so we can merge this PR and move on ...

naashton: I do agree, adding large blobs of text that act like file input unnecessarily increase file sizes, but as a dev shop, we tend to favor existing patterns. Maybe this is something we can revisit in an independent issue?

PR #4621, opened 2022-11-04
Reach: Add customs fields

jherreraa:
Reach: Add customs fields

## Summary:

Adding Reach custom fields

Remote tests
10 tests, 23 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

unit tests
6 tests, 28 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -1,0 +1,162 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class ReachGateway < Gateway
+      # TODO: Things to check
+      # * The list of three digit fractions but only accept 2
+      # * Not sure the list of countries and currencies
+
+      self.test_url = 'https://checkout.rch.how/'
+      self.live_url = 'https://checkout.rch.io/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa diners_club american_express jcb master discover maestro]
+
+      self.homepage_url = 'https://www.withreach.com/'
+      self.display_name = 'Reach'
+      self.currencies_without_fractions = %w(BIF BYR CLF CLP CVE DJF GNF ISK JPY KMF KRW PYG RWF UGX UYI VND VUV XAF XOF XPF IDR MGA MRO)
+
+      API_VERSION = 'v2.22'.freeze
+      STANDARD_ERROR_CODE_MAPPING = {}
+      PAYMENT_METHOD_MAP = {
+        american_express: 'AMEX',
+        cabal: 'CABAL',
+        check: 'ACH',
+        dankort: 'DANKORT',
+        diners_club: 'DINERS',
+        discover: 'DISC',
+        elo: 'ELO',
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,200 @@
...
+          Email: options[:email],
+          Address: address[:address1],
+          City: address[:city],
+          Country: address[:country]
+        }.compact
+      end
+
+      def add_custom_fields_data(request, options)
+        if options[:fingerprint].present?
+          request[:DeviceFingerprint] = options[:fingerprint]
```
javierpedrozaing, line 113: 🤔 
Hi @jherreraa I'm wondering if the DeviceFingerprint field should be sent through other kind of transaction as the capture, according to this part of the docs
Screen Shot 2022-11-04 at 3 05 56 PM

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -0,0 +1,200 @@
...
+          request[:ViaAgent] = false
+        end
+        add_shipping_data(request, options) if options[:consumer_taxes].present?
```
javierpedrozaing, line 116: ❓ 
For what reason validate only presence of consumer_taxes field, what happen whether the other shipping data are present as the consumer_price and consumer_duty, that data wouldn’t added ?

test/remote/gateways/remote_reach_test.rb
```diff
@@ -0,0 +1,134 @@
...
+    @options[:consignee_country] = 'US'
+
+    response = @gateway.purchase(@amount, @credit_card, @options)
+
+    assert_success response
+    assert response.params['response'][:Authorized]
+    assert response.params['response'][:OrderId]
+  end
+
+  def test_successful_purchase_with_items
```
javierpedrozaing, line 86: 💬 
would be great if we can add some test for the ConsumerTaxes or ConsumerDuty fields

javierpedrozaing (commented):
Great work, just I left some comments and I'm not sure if we need to update the changelog 🤔

lib/active_merchant/billing/gateways/reach.rb
jherreraa: Hello! 
According to reach API documentation this field is required in the checkout request which is the base to the other requests and capture request does not require the field.

lib/active_merchant/billing/gateways/reach.rb
jherreraa: These three consumer fields are required for the shipping information also the consignee data, nevertheless, this validation was implemented that way because every field are required, so if one of them is present, all of them should be. If any of them is missing, gateway will reply with the data error.
Let me know if you think otherwise and we could talk about a way to write it better

test/remote/gateways/remote_reach_test.rb
jherreraa: New test added for consumer info and consignee, please check.

javierpedrozaing (commented):
LGTM

PR #4623, opened 2022-11-04
CyberSource: Refactor and add support for several fields

rachelkirk:
CER-243

This PR refactors the order of some methods to prevent XML parse errors. It also adds support for the following fields:

installment_gracePeriodDuration
taxManagementIndicator
purchaseTotals_invoiceAmount
purchaseTotals_originalAmount
invoiceHeader_referenceDataCode
invoiceHeader_invoiceNumber
ccCaptureService_gratuityAmount
ccAuthService_mobileRemotePaymentType
otherTax_vatTaxRate

Local:
5371 tests, 76705 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
118 tests, 600 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.6102% passed
There is one additional remote test failure due to the fields not being added at the top level.

Unit:
123 tests, 588 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -62,7 +62,21 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
       ignore_cvv: 'true',
       commerce_indicator: 'internet',
       user_po: 'ABC123',
-      taxable: true
+      taxable: true,
+      sales_slip_number: '456',
+      airline_agent_code: '7Q',
+      tax_management_indicator: 1,
+      installment_grace_period_duration: '1',
+      invoice_amount: '3',
+      original_amount: '4',
+      reference_data_code: 'ABC123',
+      invoice_number: '123',
+      mobile_remote_payment_type: 'A1',
+      vat_tax_rate: '1'
+    }
+
+    @capture_options = {
+      gratuity_amount: '3.50'
     }
 
     @subscription_options = {
@@ -209,7 +223,14 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
   end
 
   def test_successful_authorization_with_installment_data
-    options = @options.merge(installment_total_count: 2, installment_total_amount: 0.50, installment_plan_type: 1, first_installment_date: '300101', installment_annual_interest_rate: 1.09)
+    options = @options.merge(
+      installment_total_count: 2,
+      installment_total_amount: 0.50,
+      installment_plan_type: 1,
+      first_installment_date: '300101',
+      installment_annual_interest_rate: 1.09,
+      installment_grace_period_duration: 1
+    )
     assert response = @gateway.authorize(@amount, @credit_card, options)
     assert_successful_response(response)
     assert !response.authorization.blank?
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -527,7 +554,7 @@ def test_authorize_and_capture
     assert auth = @gateway.authorize(@amount, @credit_card, @options)
     assert_successful_response(auth)
 
-    assert capture = @gateway.capture(@amount, auth.authorization)
+    assert capture = @gateway.capture(@amount, auth.authorization, @capture_options)
     assert_successful_response(capture)
```
rachelkirk, line 558: This is testing for `gratuity_amount`

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -690,15 +717,13 @@ def test_successful_capture_with_mdd_fields
...
-        merchantDescriptor: {
-          name: 'Test Name',
-          address1: '123 Main Dr',
-          locality: 'Durham'
-        }
-      }
+      merchant_descriptor_name: 'Test Name',
+      merchant_descriptor_address1: '123 Main Dr',
+      merchant_descriptor_locality: 'Durham'
     }
```
rachelkirk, line 727: This tests passes as implemented, but that's because the fields are not added as top level options. I was on the fence about whether to deal with this issue now or leave it for a later clean up task, so I'd love to hear your thoughts.

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -62,7 +62,21 @@ def setup
...
+      airline_agent_code: '7Q',
+      tax_management_indicator: 1,
+      installment_grace_period_duration: '1',
+      invoice_amount: '3',
+      original_amount: '4',
+      reference_data_code: 'ABC123',
+      invoice_number: '123',
+      mobile_remote_payment_type: 'A1',
+      vat_tax_rate: '1'
+    }
```
rachelkirk, line 76: I didn't end up writing tests for each individual field, bc it seemed more important to make sure they'd be added to the request in order no matter the transaction type so I added them to the main options hash. I did include a few remote tests or addition where the situation seemed to call for it, like adding the `capture_options` since that field would only be added to a capture request.

test/remote/gateways/remote_cyber_source_test.rb
jcreiff: This test failure just exposes that this field implementation was already broken

test/remote/gateways/remote_cyber_source_test.rb
jcreiff: I think that's reasonable; given the strictness of the XML parser errors that you fought with, if this options hash doesn't cause problems throughout the test suite, we can assume the fields are applied safely to the request

jcreiff (approved):
This looks good to me - your work has uncovered that there is an existing issue with the `add_merchant_description` method, but that can be addressed in a separate task

PR #4627, opened 2022-11-10
Reach: Adding custom fields

jherreraa:
## Summary:

Adding Reach custom fields

Remote tests
12 tests, 27 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

unit tests
11 tests, 40 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed"


lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -97,6 +98,30 @@ def add_customer_data(request, options, payment)
         }.compact
       end
 
+      def add_custom_fields_data(request, options)
+        if options[:fingerprint].present?
```
curiousepic, line 102: I think we should use `device_fingerprint` for the option name as well, as this pattern is used elsewhere.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -97,6 +98,30 @@ def add_customer_data(request, options, payment)
...
+          request[:DeviceFingerprint] = options[:fingerprint]
+          request[:ViaAgent] = false
+        end
+        add_shipping_data(request, options) if options[:consumer_taxes].present?
+        request[:RateOfferId] = options[:rate_offer_id] if options[:rate_offer_id].present?
+        request[:Items] = options[:items] if options[:items].present?
```
curiousepic, line 108: Would this be a hash with sub-items, or would it only ever be a single string?

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -97,6 +98,30 @@ def add_customer_data(request, options, payment)
...
+      end
+
+      def add_shipping_data(request, options)
+        request[:Shipping] = {
+          ConsumerPrice: options[:consumer_price],
+          ConsumerTaxes: options[:consumer_taxes],
+          ConsumerDuty: options[:consumer_duty]
```
curiousepic, line 115: Unless we know there are other specific types of these fields used by this API, `tax` and `duty` are commonly used elsewhere, so I think we can remove the `consumer_` and use these here:
`price`
`tax`
`duty`

curiousepic:
Looks good, but I have a few suggestions. @naashton to confirm them and approve/merge.

lib/active_merchant/billing/gateways/reach.rb
jherreraa: This would be a hash of sub items

lib/active_merchant/billing/gateways/reach.rb
jherreraa: Noted! corrected.

curiousepic (approved):
changes look good, @naashton to confirm & merge.

naashton (approved):
All looks good here! Thanks for the diligent work here, @jherreraa and for your review @curiousepic.

PR #4629, opened 2022-11-11
fixed source.stored added in root by mistake, moved to its right place

marioarranzr:
Changed made in https://github.com/activemerchant/active_merchant/pull/4611 caused the issue I describe bellow:

I think `source.type` was added in a way that is setting the attribute in the root

image
 instead of in 
```
    "source": {
        "type": "card",
         "stored": "true" 
```


rachelkirk (changes requested):
Thanks for opening this PR, I'll approve it but I need you to squash your commits first and then rebase and force push back to your branch. Thanks!

marioarranzr:
> Thanks for opening this PR, I'll approve it but I need you to squash your commits first and then rebase and force push back to your branch. Thanks!

Done @rachelkirk 👍

PR #4632, opened 2022-11-17
Reach - PSP Integration - Verify / Void / Refund

sinourain:
Description
-------------------------
This integration support the following payment operations:

Verify
Void
Refund

Unit test
-------------------------

Finished in 35.278047 seconds.
5370 tests, 76698 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

152.22 tests/s, 2174.10 assertions/s

Remote test
-------------------------

Finished in 11.168364 seconds.

12 tests, 27 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

1.07 tests/s, 2.42 assertions/s

Rubocop
-------------------------

Running RuboCop...
Inspecting 753 files

### DIFF
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -56,6 +56,33 @@ module ActiveMerchant #:nodoc:
         commit('capture', post)
       end
 
+      def refund(amount, authorization, options = {})
+        post = {}
+        request = post[:request] = {}
+        request[:MerchantId] = @options[:merchant_id]
+        request[:OrderId] = authorization
+        request[:ReferenceId] = options[:reference_id]
+        request[:Amount] = amount
+
+        commit('refund', post)
+      end
+
+      def void(authorization, options = {})
+        post = {}
+        request = post[:request] = {}
+        request[:MerchantId] = @options[:merchant_id]
+        request[:OrderId] = authorization
+
+        commit('cancel', post)
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
+      end
+
       private
 
       def build_checkout_request(amount, payment, options)
--- a/test/remote/gateways/remote_reach_test.rb
+++ b/test/remote/gateways/remote_reach_test.rb
@@ -24,6 +24,7 @@ class RemoteReachTest < Test::Unit::TestCase
         country: 'US'
       }
     }
+    @non_valid_authorization = SecureRandom.uuid
   end
 
   def test_successful_authorize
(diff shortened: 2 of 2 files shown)


lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -56,6 +56,40 @@ def capture(money, authorization, options = {})
         commit('capture', post)
       end
 
+      def supports_scrubbing?
```
gasb150, line 59: 💬  Comment:
 This PR should not affect the #supports_scrubbing? and #scrub methods.

gasb150 (commented):
LGTM, I just left a small comment.

test/remote/gateways/remote_reach_test.rb
```diff
@@ -74,6 +75,60 @@ def test_failed_capture
...
+  #
+
+  # def test_successful_refund
+  #   purchase = @gateway.purchase(@amount, @credit_card, @options)
+  #   sleep(1.minute)
+  #   response = @gateway.refund(@amount, purchase.authorization, { reference_id: 'REFUND_TAG' })
+
+  #   assert_success response
+  #   assert response.params['response'][:RefundId].present?
+  # end
```
Heavyblade, line 90: 💬  Comment:

If refund responses are idempotent isn't possible to have a test with fixed orderId value an use that just to run the refund code ans ensure that works properly ?

test/remote/gateways/remote_reach_test.rb
```diff
@@ -74,6 +75,60 @@ def test_failed_capture
...
+
+  def test_failed_refund
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    response = @gateway.refund(@amount, purchase.authorization, { reference_id: 'REFUND_TAG' })
+
+    assert_failure response
+    assert_equal 'OrderStateInvalid', response.params['response'][:Error][:Code]
```
Heavyblade, line 97: ⚠️  Warning:

If you are getting this as a real error code, please check and modify the `error_code_from` method, so you can check the error from response.error_code instead of checking the actual response body

test/remote/gateways/remote_reach_test.rb
```diff
@@ -74,6 +75,60 @@ def test_failed_capture
...
+    assert response.params['response'][:OrderId].present?
+  end
+
+  def test_failed_void
+    response = @gateway.void(@non_valid_authorization, @options)
+    assert_failure response
+
+    assert_equal 'Not Found', response.message
+    assert response.params.blank?
+  end
```
Heavyblade, line 115: 💬  Comment:

Just to be sure that void works properly and that we handling different responses properly consider the cases for:

- Partial void and other partial void to complement the total amount
- Double void to check if the functionality is idempotent or we should handle errors
- Void for a higher amount

test/remote/gateways/remote_reach_test.rb
```diff
@@ -74,6 +75,60 @@ def test_failed_capture
...
+    assert_success response
+    assert response.params['response'][:OrderId].present?
+  end
+
+  def test_failed_verify
+    response = @gateway.verify(@declined_card, @options)
+
+    assert_failure response
+    assert response.params['response'][:OrderId].present?
+    assert_equal 'PaymentAuthorizationFailed', response.params['response'][:Error][:Code]
```
Heavyblade, line 129: 💬  Comment:

Same comment here about the error_code.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -56,6 +56,40 @@ def capture(money, authorization, options = {})
...
+        commit('refund', post)
+      end
+
+      def void(authorization, options = {})
+        post = { request: { MerchantId: @options[:merchant_id], OrderId: authorization } }
+
+        commit('cancel', post)
+      end
+
+      def verify(credit_card, options = {})
```
Heavyblade, line 86: ⚠️ Warning/Suggestion:

Here you have some testable behavior on how you want your reponse, is posible to add a unit test that checks that:
- Response is Multireponse instance
- The authorize response is the default one
- last response is not used

Heavyblade (commented):
@sinourain great work here, leave some comments to improve some areas.

lib/active_merchant/billing/gateways/reach.rb
sinourain: Done!

test/remote/gateways/remote_reach_test.rb
sinourain: Done!

test/remote/gateways/remote_reach_test.rb
sinourain: Done!

test/remote/gateways/remote_reach_test.rb
sinourain: Done! I added the extra test cases

test/remote/gateways/remote_reach_test.rb
sinourain: ok, done!

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -56,6 +56,32 @@ def capture(money, authorization, options = {})
...
 
+      def refund(amount, authorization, options = {})
+        post = {
+          request: {
+            MerchantId: @options[:merchant_id],
+            OrderId: authorization,
+            ReferenceId: options[:reference_id],
+            Amount: amount
+          }
+        }
```
naashton, line 67: Small nit, but we typically follow the pattern of building these hashes, line-by-line like so:

```ruby
request = post[:request] = {}
request[:MerchantId] = @options[:merchant_id]
request[:OrderId] = authorization
...
```

This pattern makes it a bit easier to read (imo) and also helps to break the code into smaller, more manageable methods.

test/remote/gateways/remote_reach_test.rb
```diff
@@ -74,6 +75,88 @@ def test_failed_capture
...
 
+  def test_successful_refund
+    response = @gateway.refund(
+      @amount,
+      '5cd04b6a-7189-4a71-a335-faea4de9e11d',
+      { reference_id: 'REFUND_TAG' }
+    )
+
+    assert_success response
+    assert response.params['response'][:RefundId].present?
```
naashton, line 86: Hmm, not sure why we are getting mixed symbols and strings in the response. Based on the `parse` method, all of the contents of the response hash _should_ be symbolized, but it seems like that is being undone incidentally elsewhere?

naashton (changes requested):
I noticed a couple of issues. One small one, with the `Random` call, but the other is in the response. It seems the call to symbolize is being undone, which is leading to checking mixed field types in the `response` hash.

We should correct that, but since it's not in the scope of this ticket, maybe we should create a bug ticket to remedy that.

lib/active_merchant/billing/gateways/reach.rb
naashton: This might be better served to touch up in a refactor ticket. Not something that needs to be corrected here, but perhaps something we should consider going forward.

test/remote/gateways/remote_reach_test.rb
```diff
@@ -24,6 +24,7 @@ def setup
         country: 'US'
       }
     }
+    @non_valid_authorization = Random.uuid
```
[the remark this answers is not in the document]
sinourain: Done!

lib/active_merchant/billing/gateways/reach.rb
sinourain: Done!

test/remote/gateways/remote_reach_test.rb
sinourain: good catch, I change them to symbols

naashton (approved):
Excellent work!

PR #4634, opened 2022-11-21
CyberSource: Refactor to better adhere to XSD

rachelkirk:
This PR updates a few methods so that fields can be sent independently without causing the method to return if certain fields aren’t present. It also updates the order in which some methods are added to the request re: the XSD file.

Unit Tests:
124 tests, 592 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
119 tests, 605 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 95.7983% passed
*There are 3 tests that have been failing for months and a few others that are failing intermittently associated with bank accounts. I tried to look and see if maybe just the response message had changed but the failures are unpredictable and sometimes pass, while another bank account test fails.

Local:
5399 tests, 76829 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -33,6 +33,7 @@
 * iVeri: add new url [almalee24] #4630
 * Payeezy: Enable Apple Pay support [naashton] #4631
 * Payeezy: Scrub Cryptogram [naashton] #4633
+* CyberSource: Refactor several methods to better adhere to XSD [rachelkirk] #4634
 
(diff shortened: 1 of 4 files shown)


jcreiff (approved):
Looks good! 🚢

PR #4636, opened 2022-11-24
Reach: Add stored credential support

gasb150:
Summary:
In order to be able to store credentials, this commit adds add_stored_credentials method, and get_network_payment_reference to be used in network_transaction_id

Remote Tests:
Finished in 36.906747 seconds.
16 tests, 42 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit Tests:
Finished in 2.793742 seconds.
15 tests, 51 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -42,9 +42,20 @@ module ActiveMerchant #:nodoc:
         request = build_checkout_request(money, payment, options)
         add_custom_fields_data(request, options)
         add_customer_data(request, options, payment)
-
-        post = { request: request, card: add_payment(payment) }
-        commit('checkout', post)
+        add_stored_credentials(request, options)
+        post = { request: request, card: add_payment(payment, options) }
+        if options[:stored_credential]
+          MultiResponse.run(:use_first_response) do |r|
+            r.process { commit('checkout', post) }
+            r.process do
+              r2 = get_network_payment_reference(r.responses[0])
+              r.params[:network_transaction_id] = r2.message
+              r2
+            end
+          end
+        else
+          commit('checkout', post)
+        end
       end
 
       def purchase(money, payment, options = {})
@@ -63,7 +74,7 @@ module ActiveMerchant #:nodoc:
 
       def scrub(transcript)
         transcript.
-          gsub(%r(((MerchantId%22%3A%22)[\w-]+)), '\2[FILTERED]').
+          gsub(%r(((MerchantId)[% \w]+[%]\d{2})[\w -]+), '\1[FILTERED]').
           gsub(%r((signature=)[\w%]+), '\1[FILTERED]\2').
           gsub(%r((Number%22%3A%22)[\d]+), '\1[FILTERED]\2').
           gsub(%r((VerificationCode%22%3A)[\d]+), '\1[FILTERED]\2')
@@ -82,18 +93,18 @@ module ActiveMerchant #:nodoc:
             Sku: options[:item_sku] || SecureRandom.alphanumeric,
             ConsumerPrice: amount,
             Quantity: (options[:item_quantity] || 1)
-          ],
-          ViaAgent: true # Indicates this is server to server API call
+          ]
         }
       end
 
-      def add_payment(payment)
+      def add_payment(payment, options)
+        ntid = options.dig(:stored_credential, :network_transaction_id)
+        cvv_or_previos_reference = (ntid ? { PreviousNetworkPaymentReference: ntid } : { VerificationCode: payment.verification_value })
         {
           Name: payment.name,
           Number: payment.number,
-          Expiry: { Month: payment.month, Year: payment.year },
-          VerificationCode: payment.verification_value
-        }
+          Expiry: { Month: payment.month, Year: payment.year }
+        }.merge!(cvv_or_previos_reference)
       end
 
       def add_customer_data(request, options, payment)
@@ -110,11 +121,41 @@ module ActiveMerchant #:nodoc:
         }.compact
       end
 
+      def add_stored_credentials(request, options)
+        request[:PaymentModel] = payment_model(options)
+        raise ArgumentError, 'Unexpected combination of stored credential fields' if request[:PaymentModel].nil?
+
+        request[:DeviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint] && request[:PaymentModel].match?(/CIT-/)
+      end
+
+      def payment_model(options)
+        stored_credential = options[:stored_credential]
+        return options[:payment_model] if options[:payment_model]
+        return 'CIT-One-Time' unless stored_credential
+
+        payment_model_options = {
+          initial_transaction: {
+            'cardholder' => {
+              'installment' => 'CIT-Setup-Scheduled',
+              'unschedule' => 'CIT-Setup-Unscheduled-MIT',
+              'recurring' => 'CIT-Setup-Unschedule'
+            }
+          },
+          no_initial_transaction: {
+            'cardholder' => {
+              'unschedule' => 'CIT-Subsequent-Unscheduled'
+            },
+            'merchant' => {
+              'recurring' => 'MIT-Subsequent-Scheduled',
+              'unschedule' => 'MIT-Subsequent-Unscheduled'
+            }
+          }
+        }
+        initial = (stored_credential[:initial_transaction] ? :initial_transaction : :no_initial_transaction)
+        payment_model_options[initial].dig(stored_credential[:initiator], stored_credential[:reason_type])
+      end
+
       def add_custom_fields_data(request, options)
-        if options[:device_fingerprint].present?
-          request[:DeviceFingerprint] = options[:device_fingerprint]
-          request[:ViaAgent] = false
-        end
         add_shipping_data(request, options) if options[:taxes].present?
         request[:RateOfferId] = options[:rate_offer_id] if options[:rate_offer_id].present?
         request[:Items] = options[:items] if options[:items].present?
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -110,11 +110,41 @@ def add_customer_data(request, options, payment)
...
+            "stored credential options doesn't match with PaymentModels combination list, you can also pass in the options directly the PaymentModels value"
         end
+
+        if request[:PaymentModel].match?(/CIT-/)
+          request[:DeviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
+        end
+      end
+
+      def add_payment_model(options)
+        return options[:payment_model] if options[:payment_model]
```
Heavyblade, line 126: 💬  Comment:

Is a pretty particualr implementation and works, going to leave you a hash based alternative for your consideration => https://gist.github.com/54e9832c74bce33de4e84a49b8b85e29

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -110,11 +110,41 @@ def add_customer_data(request, options, payment)
...
+          raise ArgumentError,
+            "stored credential options doesn't match with PaymentModels combination list, you can also pass in the options directly the PaymentModels value"
         end
+
+        if request[:PaymentModel].match?(/CIT-/)
+          request[:DeviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
+        end
+      end
+
+      def add_payment_model(options)
```
Heavyblade, line 125: 💬  Comment:

Maybe the name should'nt be "add_payment_model" because you are not receiving the request body object to add more params.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -110,11 +110,41 @@ def add_customer_data(request, options, payment)
         }.compact
       end
 
-      def add_custom_fields_data(request, options)
-        if options[:device_fingerprint].present?
-          request[:DeviceFingerprint] = options[:device_fingerprint]
-          request[:ViaAgent] = false
+      def add_stored_credentials(request, options)
+        request[:PaymentModel] = add_payment_model(options)
+        if request[:PaymentModel].nil?
```
Heavyblade, line 115: 💬  Comment:

A simple guard clause would make this a one liner.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -152,18 +182,32 @@ def format_and_sign(post)
         post
       end
 
+      def get_network_payment_reference(response)
+        parameters = { request: { MerchantId: @options[:merchant_id], OrderId: response[:response][:OrderId] } }
+        body = post_data format_and_sign(parameters)
+
+        # unit tests doesn't work with ssl_post
+        raw_response = ssl_request :post, url('query'), body, {}
```
Heavyblade, line 190: 💬 Comment:

For debugging and tracing this request should be part of a multiresponse object

test/unit/gateways/reach_test.rb
```diff
@@ -130,6 +129,33 @@ def test_sucess_from_on_failure
     refute @gateway.send(:success_from, response)
   end
 
+  def test_stored_credential
+    @options[:stored_credential] = { initiator: 'cardholder', initial_transaction: true, reason_type: 'unschedule' }
```
Heavyblade, line 133: 💬  Comment:

Maybe this is a good scenario to test a combination of values and their expected output so you can have a lot of combinations covered in the test without creating a test for each transactions, something like this https://gist.github.com/54e9832c74bce33de4e84a49b8b85e29

Heavyblade (commented):
@gasb150  almost there, leaved some comments for your consideration.

lib/active_merchant/billing/gateways/reach.rb
gasb150: I could not inject the result of the 2nd request into the 1st one, so I tried this approach.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -152,11 +188,23 @@ def format_and_sign(post)
         post
       end
 
+      def get_network_payment_reference(response)
+        parameters = { request: { MerchantId: @options[:merchant_id], OrderId: response.params['response'][:OrderId] } }
+        body = post_data format_and_sign(parameters)
+
+        raw_response = ssl_request :post, url('query'), body, {}
+        response = parse(raw_response)
+        Response.new(true, response[:response][:Payment][:NetworkPaymentReference], {})
```
naashton, line 197: I think you'll want to use `dig` here to guard against `nil` but it also might be better to retrieve the value outside of building the `Response` object.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -194,6 +242,14 @@ def error_code_from(response)
...
         "#{test? ? test_url : live_url}#{API_VERSION}/#{action}"
       end
+
+      # Helper method to get fingerprint for testing
+      def fingerprint
+        raw_response = ssl_get url("fingerprint?MerchantId=#{@options[:merchant_id]}")
+
+        fingerprint = raw_response.match(/(gip_device_fingerprint=')([\w -]+)/)[2]
+        fingerprint
+      end
```
naashton, line 252: If this method is just for testing, can we move it to the remote test file?

test/remote/gateways/remote_reach_test.rb
```diff
@@ -140,6 +141,48 @@ def test_successful_authorize_and_capture
...
   end
 
+  def test_successful_purchase_with_store_credentials
+    @options[:stored_credential] = { initiator: 'cardholder', initial_transaction: true, reason_type: 'installment' }
+    response = @gateway.purchase(@amount, @credit_card, @options)
+
+    assert_success response
+
+    assert response.params['response'][:Authorized]
+    assert response.params['response'][:OrderId]
```
naashton, line 151: This came up in a previous PR as well. The mixed use of strings and symbols indicates to me that something is incorrect between getting a response from the gateway and building the AM `Response` object. I think it was resolved in a previous PR, so it might resolve itself when that is merged in, but just want to call it out just in case.

naashton (changes requested):
There are some things that should be changed, but also some suggestions for improvement, but overall great work.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -110,11 +117,40 @@ def add_customer_data(request, options, payment)
         }.compact
       end
 
+      def add_stored_credentials(request, options)
+        request[:PaymentModel] = payment_model(options)
+        raise ArgumentError, "stored credential options doesn't match with PaymentModels combination list, you can also pass in the options directly the PaymentModels value" if request[:PaymentModel].nil?
```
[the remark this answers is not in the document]
Heavyblade: mmm maybe I'm wrong @naashton, but the pattern that I'm seeing on AM for param issues that fail the transaction is to raise an `Argument` and CORE is well prepared to deal with that on the Gateway#remote method and response with the proper 400 HTTP Error.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -110,11 +117,40 @@ def add_customer_data(request, options, payment)
...
+
+        request[:DeviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint] && request[:PaymentModel].match?(/CIT-/)
+      end
+
+      def payment_model(options)
+        return options[:payment_model] if options[:payment_model]
```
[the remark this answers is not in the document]
gasb150: Ok, then I will remove the payment_model options

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -42,9 +42,16 @@ def authorize(money, payment, options = {})
...
-        post = { request: request, card: add_payment(payment) }
-        commit('checkout', post)
+        add_stored_credentials(request, options)
+        post = { request: request, card: add_payment(payment, options) }
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { commit('checkout', post) }
+          r.process do
+            r2 = get_network_payment_reference(r.responses[0])
+            r.params[:network_transaction_id] = r2.message
+            r2
```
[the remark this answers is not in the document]
gasb150: OK then, I'll add a conditional

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -82,18 +89,18 @@ def build_checkout_request(amount, payment, options)
...
             Quantity: (options[:item_quantity] || 1)
-          ],
-          ViaAgent: true # Indicates this is server to server API call
+          ]
         }
       end
 
-      def add_payment(payment)
+      def add_payment(payment, options)
+        ntk_transa_id = options.dig(:stored_credential, :network_transaction_id)
```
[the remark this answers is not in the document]
gasb150: You are right.

lib/active_merchant/billing/gateways/reach.rb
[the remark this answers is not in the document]
naashton: I think you are right, Core will handle it gracefully. I think I was confused about what the `raise` was checking for, but now I see how it's being used. I think this is fine as it is.

test/remote/gateways/remote_reach_test.rb
naashton: Yes, I think it should be corrected in a separate issue. This is fine for now.

naashton (approved):
This all looks good! Thanks for working through these changes.

PR #4637, opened 2022-11-24
Openpay: able to set endpoint URL by merchant country flag

edgarv09:
Summary:
Add support to set service URL by setting the merchant country flag, with this capability openpay knows from which region is happening the transaction.

Remote:
25 tests, 83 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96% passed

Unit:
22 tests, 113 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/openpay.rb
+++ b/lib/active_merchant/billing/gateways/openpay.rb
@@ -1,8 +1,15 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class OpenpayGateway < Gateway
-      self.live_url = 'https://api.openpay.mx/v1/'
-      self.test_url = 'https://sandbox-api.openpay.mx/v1/'
+      class_attribute :mx_live_url, :mx_test_url
+      class_attribute :co_live_url, :co_test_url
+
+      self.co_live_url = 'https://api.openpay.co/v1/'
+      self.co_test_url = 'https://sandbox-api.openpay.co/v1/'
+      self.mx_live_url = 'https://api.openpay.mx/v1/'
+      self.mx_test_url = 'https://sandbox-api.openpay.mx/v1/'
+      self.live_url = self.co_live_url
+      self.test_url = self.co_test_url
 
       self.supported_countries = %w(CO MX)
       self.supported_cardtypes = %i[visa master american_express carnet]
@@ -22,6 +29,16 @@ module ActiveMerchant #:nodoc:
         @api_key = options[:key]
         @merchant_id = options[:merchant_id]
         super
+      end
+
+      def gateway_url(options = {})
+        country = options[:merchant_country] || @options[:merchant_country]
+
+        if country == 'MX'
+          test? ? mx_test_url : mx_live_url
+        else
+          test? ? co_test_url : co_live_url
+        end
       end
 
       def purchase(money, creditcard, options = {})
@@ -192,7 +209,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def http_request(method, resource, parameters = {}, options = {})
-        url = (test? ? self.test_url : self.live_url) + @merchant_id + '/' + resource
+        url = gateway_url(options) + @merchant_id + '/' + resource
         raw_response = nil
         begin
           raw_response = ssl_request(method, url, (parameters ? parameters.to_json : nil), headers(options))
--- a/test/remote/gateways/remote_openpay_test.rb
+++ b/test/remote/gateways/remote_openpay_test.rb
@@ -16,6 +16,19 @@ class RemoteOpenpayTest < Test::Unit::TestCase
   end
 
   def test_successful_purchase
+    @options[:email] = '%d@example.org' % Time.now
+    @options[:name] = 'Customer name'
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_nil response.message
+  end
+
+  def test_successful_purchase_with_mexico_url
+    @options[:email] = '%d@example.org' % Time.now
+    @options[:name] = 'Customer name'
+    credentials = fixtures(:openpay).merge(merchant_country: 'MX')
+    @gateway = OpenpayGateway.new(credentials)
+
     assert response = @gateway.purchase(@amount, @credit_card, @options)
     assert_success response
     assert_nil response.message
@@ -31,7 +44,7 @@ class RemoteOpenpayTest < Test::Unit::TestCase
   def test_unsuccessful_purchase
     assert response = @gateway.purchase(@amount, @declined_card, @options)
     assert_failure response
-    assert_equal 'The card was declined', response.message
+    assert_equal 'The card was declined by the bank', response.message
   end
 
   def test_successful_refund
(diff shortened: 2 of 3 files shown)


Heavyblade (commented):
Almost there @edgarv09

test/remote/gateways/remote_openpay_test.rb
```diff
@@ -16,6 +16,19 @@ def setup
...
   def test_successful_purchase
+    @options[:email] = '%d@example.org' % Time.now
+    @options[:name] = 'Customer name'
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_nil response.message
+  end
+
+  def test_successful_purchase_with_mexico_url
+    @options[:email] = '%d@example.org' % Time.now
```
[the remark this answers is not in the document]
edgarv09: Yeah!! I asked David perry for credentials, and he told me to use the credentials already in the fixture file. Taking into account does not matter which endpoint we point to.

lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -1,8 +1,13 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class OpenpayGateway < Gateway
-      self.live_url = 'https://api.openpay.mx/v1/'
-      self.test_url = 'https://sandbox-api.openpay.mx/v1/'
+      class_attribute :mx_live_url
```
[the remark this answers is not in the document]
edgarv09: Am following the structure that the team uses in other cases like [citrus_pay](https://github.com/activemerchant/active_merchant/blob/5bc423f7f1927c6fc2947f2a9ed115cb3eaa3722/lib/active_merchant/billing/gateways/citrus_pay.rb#L6) and here is the logic to select the URL [mastercard_gateway](https://github.com/activemerchant/active_merchant/blob/5bc423f7f1927c6fc2947f2a9ed115cb3eaa3722/lib/active_merchant/billing/gateways/mastercard.rb#L219)


One thing with Openpay is that it allows the URL to be selected upon gateway creation or transaction execution. 
[Slack discusion](https://spreedly.slack.com/archives/CH7B6NWA0/p1663866546455239?thread_ts=1663086078.255359&cid=CH7B6NWA0)

lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -21,9 +28,20 @@ def initialize(options = {})
...
         @api_key = options[:key]
         @merchant_id = options[:merchant_id]
+        assign_urls(options)
         super
       end
 
+      def assign_urls(options = {})
+        self.test_url = options[:merchant_country] == 'MX' ? self.mx_test_url : self.co_test_url
+        self.live_url = options[:merchant_country] == 'MX' ? self.mx_live_url : self.co_live_url
+      end
```
edgarv09, line 38: class attribute generates accesor for class and instance https://apidock.com/rails/Class/class_attribute

```
## Instances may overwrite the class value in the same way:

Base.setting = true
object = Base.new
object.setting          # => true
object.setting = false
object.setting          # => false
Base.setting            # => true
```

test/unit/gateways/openpay_test.rb
```diff
@@ -31,6 +31,32 @@ def test_successful_purchase
...
   end
 
+  def test_successful_purchase_with_mexico_url
+    gateway = OpenpayGateway.new(
+      key: 'key',
+      merchant_id: 'merchant_id',
+      merchant_country: 'MX'
+    )
+
+    gateway.expects(:ssl_request).returns(successful_purchase_response)
```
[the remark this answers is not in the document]
edgarv09:
```ruby
assert_equal gateway.gateway_url, OpenpayGateway.mx_test_url
```

`gateway_url` should be returning the desired endpoint. The above example shows that behavior.

lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -1,8 +1,15 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class OpenpayGateway < Gateway
-      self.live_url = 'https://api.openpay.mx/v1/'
-      self.test_url = 'https://sandbox-api.openpay.mx/v1/'
+      class_attribute :mx_live_url, :mx_test_url
+      class_attribute :co_live_url, :co_test_url
+
+      self.co_live_url = 'https://api.openpay.co/v1/'
```
almalee24, line 7: The code looks good but just out of curiosity why change it to `co_live_url` and `co_test_url`?

edgarv09: the idea to be more verbose about the configuration of the gateway. Currently, openpay supports two domains, one for Colombia and the other for Mexico, being Colombia the default.  that's the reason.

almalee24: Okay, thanks for the explanation!

almalee24 (approved):
Lgtm!

almalee24 (changes requested):
I think it would be good to update `gateway_url` to use `live_url` and `test_url` just to keep in pattern to what we do in AM.

edgarv09:
> I think it would be good to update `gateway_url` to use `live_url` and `test_url` just to keep in pattern to what we do in AM.

Hi! @almalee24 I understood your point but I check other gateways to have an idea of how to handle this behavior following the patterns and I found in so many cases the pattern of having a method to return a URL is present. 

if you do a search using the keyword `def url` you will find 69 matches across 68 files. 

Jetpay behavior is so similar to openpay case
https://github.com/activemerchant/active_merchant/blob/c728c5265ec4d6edf26d6a70bece17be6f9a4de0/lib/active_merchant/billing/gateways/jetpay_v2.rb#296

other examples:

https://github.com/activemerchant/active_merchant/blob/c728c5265ec4d6edf26d6a70bece17be6f9a4de0/lib/active_merchant/billing/gateways/litle.rb#L617

https://github.com/activemerchant/active_merchant/blob/c728c5265ec4d6edf26d6a70bece17be6f9a4de0/lib/active_merchant/billing/gateways/jetpay_v2.rb#L308

https://github.com/activemerchant/active_merchant/blob/c728c5265ec4d6edf26d6a70bece17be6f9a4de0/lib/active_merchant/billing/gateways/merchant_ware_version_four.rb#L258

https://github.com/activemerchant/active_merchant/blob/c728c5265ec4d6edf26d6a70bece17be6f9a4de0/lib/active_merchant/billing/gateways/merchant_ware.rb#L287

lib/active_merchant/billing/gateways/openpay.rb
```diff
@@ -24,6 +31,16 @@ def initialize(options = {})
         super
       end
 
+      def gateway_url(options = {})
+        country = options[:merchant_country] || @options[:merchant_country]
+
+        if country == 'MX'
+          test? ? mx_test_url : mx_live_url
+        else
+          test? ? co_test_url : co_live_url
```
almalee24, line 40: What I was suggesting in my request was to change this line to be `test? ? test_url : live_url`. They're assigned in line 11 & 12 but not used again. The code looks good and this is not a deal breaker so I'm going to approve your PR.

almalee24 (approved):

curiousepic (approved):
Nice 👍

PR #4640, opened 2022-12-02
CommerceHub: Add new gateway

naashton:
Added new gateway `commerce_hub` along with remote and unit tests.

This integration includes support for `stored_credential` fields along with basic gateway transactions, except for `capture`.

`capture` is currently not in a working state due to unknown configuration issues with our test account.

In order to unblock this integration, I have left the transaction method in the gateway class and opted to comment out the test. CommerceHub is working to enable our test account for `capture` transactions, and once completed, we can re-test the transaction type.

We also encounter a number of rate limit issues with our remote tests, which leads to some of the test failures.

SER-207

Unit: 9 tests, 78 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop:

Remote: 15 tests, 30 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 66.6667% passed


lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -0,0 +1,309 @@
...
+      self.homepage_url = 'https://developer.fiserv.com/product/CommerceHub'
+      self.display_name = 'CommerceHub'
+
+      STANDARD_ERROR_CODE_MAPPING = {}
+
+      SCHEDULED_REASON_TYPES = %w(recurring installment)
+      ENDPOINTS = {
+        'sale' => '/payments/v1/charges',
+        'authonly' => '/payments/v1/charges',
+        'capture' => '/payments/v1/charges',
```
aenand, line 20: Since all three of these are the same endpoint, should they be consolidated into one?

test/remote/gateways/remote_commerce_hub_test.rb
```diff
@@ -0,0 +1,144 @@
...
+    assert_equal 'Approved', response.message
+  end
+
+  # def test_successful_authorize_and_capture
+  #   authorize = @gateway.authorize(@amount, @credit_card, @options)
+  #   assert_success authorize
+
+  #   capture = @gateway.capture(@amount, authorize.authorization)
+  #   assert_success capture
+  # end
```
aenand, line 63: Why is this commented out?

test/remote/gateways/remote_commerce_hub_test.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+    transcript = @gateway.scrub(transcript)
+
+    assert_scrubbed(@credit_card.number, transcript)
+    assert_scrubbed(@gateway.options[:api_key], transcript)
+    assert_scrubbed(@gateway.options[:api_secret], transcript)
```
aenand, line 142: Can you add an assertion for CVV scrubbing

test/unit/gateways/commerce_hub_test.rb
```diff
@@ -0,0 +1,461 @@
...
+    end.check_request do |_endpoint, data, _headers|
+      request = JSON.parse(data)
+      assert_equal request['transactionDetails']['captureFlag'], false
+      assert_equal request['transactionDetails']['primaryTransactionType'], 'AUTH_ONLY'
+      assert_equal request['transactionDetails']['accountVerification'], true
+    end.respond_with(successful_authorize_response)
+
+    assert_success response
+  end
+
```
aenand, line 146: Can you add a unit test for scrubbing?

aenand (changes requested):
Looks good! Thanks for following the general guide so well. Just a few comments from looking through it

lib/active_merchant/billing/gateways/commerce_hub.rb
naashton: Good call out!

test/remote/gateways/remote_commerce_hub_test.rb
naashton: Will do

aenand (approved):
LGTM

PR #4645, opened 2022-12-09
Reach: Fixing bug when using union pay card

Heavyblade:
## Summary:

This change aims to fix an bug related with the way to handle not supported card brands.

## Tests:

Finished in 130.664153 seconds.
25 tests, 66 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Finished in 35.408869 seconds.
5415 tests, 76945 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

152.93 tests/s, 2173.04 assertions/s
Inspecting 756 files

### DIFF
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -30,6 +30,7 @@ module ActiveMerchant #:nodoc:
         maestro: 'MAESTRO',
         master: 'MC',
         naranja: 'NARANJA',
+        union_pay: 'UNIONPAY',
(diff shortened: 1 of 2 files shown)


Heavyblade:
@naashton indeed this PR solves that issue, the problem was that Union pay was not mapped and the PaymentMethod param end-up being empty, so I added union pay as an options and an Argument Error in case that something like alelo,sodexo,etc. gets there.

naashton (approved):
Awesome! Thanks for the clarification.

PR #4647, opened 2022-12-14
Element/Vantiv: Add google pay and apple pay support

jherreraa:
Summary:

This PR updates element/vantiv gateway to support apple/google pay transactions

Remote Tests

Finished in 35.250506 seconds.
31 tests, 86 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/element.rb
+++ b/lib/active_merchant/billing/gateways/element.rb
@@ -17,6 +17,11 @@ module ActiveMerchant #:nodoc:
       SERVICE_TEST_URL = 'https://certservices.elementexpress.com/express.asmx'
       SERVICE_LIVE_URL = 'https://services.elementexpress.com/express.asmx'
 
+      NETWORK_TOKEN_TYPE = {
+        apple_pay: '2',
+        google_pay: '1'
+      }
+
       def initialize(options = {})
         requires!(options, :account_id, :account_token, :application_id, :acceptor_id, :application_name, :application_version)
         super
@@ -171,6 +176,8 @@ module ActiveMerchant #:nodoc:
           add_payment_account_id(xml, payment)
         elsif payment.is_a?(Check)
           add_echeck(xml, payment)
+        elsif payment.is_a?(NetworkTokenizationCreditCard)
+          add_network_tokenization_card(xml, payment)
         else
           add_credit_card(xml, payment)
         end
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_element_test.rb
```diff
@@ -12,6 +12,25 @@ def setup
...
+    @apple_pay_network_token = network_tokenization_credit_card('4895370015293175',
+      month: 10,
+      year: Time.new.year + 2,
+      first_name: 'John',
+      last_name: 'Smith',
+      verification_value: '737',
+      payment_cryptogram: 'abc1234567890',
+      eci: '07',
+      transaction_id: 'abc123',
+      source: :apple_pay)
```
Heavyblade, line 33: 💬 Comment:

Is an small silly thing but adding extra data like first_name, last_name and verification_value kind of shows that apple pay tokens somehow are different than google pay tokens.

Heavyblade (commented):
@jherreraa basically done, nice work !, leaved you a couple of comments.

lib/active_merchant/billing/gateways/element.rb
```diff
@@ -246,6 +253,18 @@ def add_echeck(xml, payment)
...
 
+      def add_network_tokenization_card(xml, payment)
+        xml.card do
+          xml.CardNumber payment.number
+          xml.ExpirationMonth format(payment.month, :two_digits)
+          xml.ExpirationYear format(payment.year, :two_digits)
+          xml.CardholderName payment.first_name + ' ' + payment.last_name
+          xml.Cryptogram payment.payment_cryptogram
+          xml.ElectronicCommerceIndicator payment.eci
+          xml.WalletType NETWORK_TOKEN_TYPE.fetch(payment.source)
```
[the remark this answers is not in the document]
jherreraa: I have checked other gateways in order to see if have the suggested behavior, but I haven't seen some kind of error handling there, let me know if I should tune up the approach.
Thanks!

lib/active_merchant/billing/gateways/element.rb
```diff
@@ -200,7 +207,7 @@ def add_transaction(xml, money, options = {})
           xml.TransactionID options[:trans_id] if options[:trans_id]
           xml.TransactionAmount amount(money.to_i) if money
           xml.MarketCode market_code(money, options) if options[:market_code] || money
-          xml.ReferenceNumber options[:order_id] || SecureRandom.hex(20)
+          xml.ReferenceNumber options[:order_id][0, 50] || SecureRandom.hex(20)
```
naashton, line 210: I'm actually not sure what this is trying to achieve? It also looks like it's causing a bunch of failures in the unit tests, which is why the CI checks are failing.

naashton (changes requested):
I think the AP/GP integration looks sound, but it seems something was added that doesn't belong. Remove that and this should be GTG!

lib/active_merchant/billing/gateways/element.rb
jherreraa: Hello! 
The purpose of the modify was adjust the reference number according to element documentation requirements, sorry, missed the validations in other tests related with the transaction, but now is corrected.

naashton (approved):
Awesome work! Could you also include a Changelog entry before merging?

Heavyblade (approved):
🚢  it !

PR #4654, opened 2022-12-22
Adyen: Add support for `skip_mpi_data` flag

rachelkirk:
CER-333

This change will allow for recurring payments with Apple Pay on Adyen by eliminating mpi data hash from request after initial payment.

Unit:
104 tests, 528 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
131 tests, 440 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
90.8397% passed
*Failing tests mostly related to 3DS/Store transactions, also failing on master. Newly added remote test passes.

Local:
5430 tests, 77037 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -435,7 +435,7 @@ module ActiveMerchant #:nodoc:
         elsif payment.is_a?(Check)
           add_bank_account(post, payment, options, action)
         else
-          add_mpi_data_for_network_tokenization_card(post, payment) if payment.is_a?(NetworkTokenizationCreditCard)
+          add_mpi_data_for_network_tokenization_card(post, payment, options) if payment.is_a?(NetworkTokenizationCreditCard)
           add_card(post, payment)
         end
       end
@@ -486,7 +486,9 @@ module ActiveMerchant #:nodoc:
         post[:originalReference] = original_reference
       end
 
-      def add_mpi_data_for_network_tokenization_card(post, payment)
+      def add_mpi_data_for_network_tokenization_card(post, payment, options)
+        return if options[:skip_mpi_data] == 'Y'
+
         post[:mpiData] = {}
         post[:mpiData][:authenticationResponse] = 'Y'
         post[:mpiData][:cavv] = payment.payment_cryptogram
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -1161,7 +1161,7 @@ def test_invalid_expiry_month_for_purchase
     card = credit_card('4242424242424242', month: 16)
     assert response = @gateway.purchase(@amount, card, @options)
     assert_failure response
-    assert_equal 'The provided Expiry Date is not valid.: Expiry month should be between 1 and 12 inclusive', response.message
+    assert_equal 'The provided Expiry Date is not valid.: Expiry month should be between 1 and 12 inclusive: 16', response.message
   end
```
rachelkirk, line 1165: Unrelated to my change but it was an easy fix for a failing test.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -486,10 +486,12 @@ def add_reference(post, authorization, options = {})
...
       end
 
-      def add_mpi_data_for_network_tokenization_card(post, payment)
+      def add_mpi_data_for_network_tokenization_card(post, payment, options)
+        return if options[:skip_mpi_data] == 'Y'
+
         post[:mpiData] = {}
         post[:mpiData][:authenticationResponse] = 'Y'
-        post[:mpiData][:cavv] = payment.payment_cryptogram
+        post[:mpiData][:cavv] = payment.payment_cryptogram unless options[:skip_mpi_data] == 'Y'
```
jcreiff, line 494: I think this `unless` statement is irrelevant, right? If we ever reach it, it should evaluate false. If it's going to evaluate true, we'll never reach it because we'll `return` at line 490

test/remote/gateways/remote_adyen_test.rb
jcreiff: Thanks for updating!

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -1343,6 +1343,39 @@ def test_auth_capture_refund_with_network_txn_id
     assert_success refund
   end
 
+  def test_purchase_with_skip_mpi_data
```
jcreiff, line 1346: Nice work with a complex test scenario!

jcreiff (changes requested):
Just one small suggestion to remove an unneeded change, otherwise I think this looks good to go!

lib/active_merchant/billing/gateways/adyen.rb
rachelkirk: Oh, good catch! I meant to take that out.

jcreiff (approved):
🚢

PR #4656, opened 2022-12-26
Reach: Fixing store credentials and Exception issues

Heavyblade:
## Summary:

This PR solves 5 issues:

1) Remove exception for no allowed store credentials combination and leave the field empty.

2) Remove exception for not supported card brand and sent instead a not supported string.

3) Fixes a Typo on stored credentials code and test related with 'unscheduled' transactions.

4) Uses the order_id (transaction token) to populate the ReferenceId field.

5) Fixes and add the corresponding tests for 1 to 3.

### Remote Test:
Finished in 136.495605 seconds.
27 tests, 70 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### Unit Tests:
Finished in 33.332219 seconds.
5419 tests, 76964 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -114,14 +114,12 @@ module ActiveMerchant #:nodoc:
       private
 
       def build_checkout_request(amount, payment, options)
-        raise ArgumentError.new("Payment method #{payment.brand} is not supported, check https://docs.withreach.com/docs/credit-cards#technical-considerations") if PAYMENT_METHOD_MAP[payment.brand.to_sym].blank?
-
         {
           MerchantId: @options[:merchant_id],
           ReferenceId: options[:order_id],
           ConsumerCurrency: options[:currency] || currency(options[:amount]),
           Capture: options[:capture] || false,
-          PaymentMethod: PAYMENT_METHOD_MAP[payment.brand.to_sym],
+          PaymentMethod: PAYMENT_METHOD_MAP.fetch(payment.brand.to_sym, 'unsupported'),
           Items: [
             Sku: options[:item_sku] || SecureRandom.alphanumeric,
             ConsumerPrice: amount,
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -153,10 +153,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_credentials(request, options)
-        request[:PaymentModel] = payment_model(options)
-        raise ArgumentError, 'Unexpected combination of stored credential fields' if request[:PaymentModel].nil?
-
-        request[:DeviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint] && request[:PaymentModel].match?(/CIT-/)
+        request[:PaymentModel] = payment_model(options) || ''
+        request[:DeviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
       end
 
       def payment_model(options)
--- a/lib/active_merchant/billing/gateways/reach.rb
+++ b/lib/active_merchant/billing/gateways/reach.rb
@@ -166,21 +166,21 @@ module ActiveMerchant #:nodoc:
           initial_transaction: {
             'cardholder' => {
               'installment' => 'CIT-Setup-Scheduled',
-              'unschedule' => 'CIT-Setup-Unscheduled-MIT',
+              'unscheduled' => 'CIT-Setup-Unscheduled-MIT',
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_reach_test.rb
```diff
@@ -202,6 +215,17 @@ def test_successful_refund
...
+  def test_successful_refund
+    response = @gateway.refund(
+      @amount,
+      '5cd04b6a-7189-4a71-a335-faea4de9e11d',
+      { order_id: 'REFUND_TAG' }
+    )
+
+    assert_success response
+    assert response.params['response']['RefundId'].present?
+  end
```
sinourain, line 227: Both test **test_successful_refund_with_reference_id** and **test_successful_refund** are the same, maybe you can remove the second one and preserve the most complete `def` description

sinourain (commented):
LGTM

test/remote/gateways/remote_reach_test.rb
sinourain: oh now I see the difference, it is about the `order_id` and `reference_id`, then NVM

test/remote/gateways/remote_reach_test.rb
```diff
@@ -202,6 +215,17 @@ def test_successful_refund
     assert response.params['response']['RefundId'].present?
   end
 
+  def test_successful_refund
```
sinourain, line 218: maybe: 
`def test_successful_refund_with_oder_id`

Could be more readable

test/remote/gateways/remote_reach_test.rb
```diff
@@ -191,7 +204,7 @@ def test_failed_capture
     assert_equal 'Not Found', response.message
   end
```
gasb150, line 206: 💭 Opinion:
 Could be nice to have a remote test case to see what's happened if the payment_model is nil.

gasb150 (commented):
LGTM, I'm just putting in a small comment.

test/remote/gateways/remote_reach_test.rb
Heavyblade: test

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -111,14 +114,12 @@ def verify(credit_card, options = {})
...
       def build_checkout_request(amount, payment, options)
-        raise ArgumentError.new("Payment method #{payment.brand} is not supported, check https://docs.withreach.com/docs/credit-cards#technical-considerations") if PAYMENT_METHOD_MAP[payment.brand.to_sym].blank?
-
         {
           MerchantId: @options[:merchant_id],
           ReferenceId: options[:order_id],
           ConsumerCurrency: options[:currency] || currency(options[:amount]),
           Capture: options[:capture] || false,
-          PaymentMethod: PAYMENT_METHOD_MAP[payment.brand.to_sym],
+          PaymentMethod: PAYMENT_METHOD_MAP.fetch(payment.brand.to_sym, 'unsupported'),
```
naashton, line 122: We shouldn't set a default value here. If a payment method is given, we can map it to the appropriate string value to give back to the gateway, but if we default to `unsupported`, that will end up returned to the user as an error of `unsupported` payment type of `unsupported` which would be confusing.

An additional note, because I didn't notice this before. It looks like the strings we map to are just `capitalized` versions of the payment method? I think we could just use the Ruby built in for `#string.upcase`.

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -167,21 +166,21 @@ def payment_model(options)
           initial_transaction: {
             'cardholder' => {
               'installment' => 'CIT-Setup-Scheduled',
-              'unschedule' => 'CIT-Setup-Unscheduled-MIT',
+              'unscheduled' => 'CIT-Setup-Unscheduled-MIT',
```
naashton, line 169: Thanks for correcting this!

naashton (changes requested):
Just one change to remove the default value for `PaymentMethod` and I think this is good to go. Thanks for your diligent work!

lib/active_merchant/billing/gateways/reach.rb
naashton: Excellent, I think my concern for 1 was that we would be short circuiting the customer value and then getting odd responses back, but sense they return parsing errors, your solution here makes sense. Thanks for the reminder on #2 that any underscore value wouldn't tolerate `.upcase`ing.

naashton (approved):
GTG! 🚢

PR #4657, opened 2022-12-27
Reach: Update store credentials to use a GSF

javierpedrozaing:
Description
-------------------------

Reach has implemented stored credentials using the Spreedly’s determination and this commit Update this in order to use directly the options from Reach gateway using  GSF (payment_model)

SER-463

Unit test
-------------------------

Finished in 0.199618 seconds.
18 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
90.17 tests/s, 460.88 assertions/s

Remote test
-------------------------

Finished in 146.011021 seconds.
28 tests, 75 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
0.19 tests/s, 0.51 assertions/s

Rubocop
-------------------------

### DIFF
--- a/test/unit/gateways/reach_test.rb
+++ b/test/unit/gateways/reach_test.rb
@@ -163,6 +163,27 @@ class ReachTest < Test::Unit::TestCase
     end.respond_with(successful_purchase_response)
   end
 
+  def test_stored_credential_with_reach_payment_model
+    reach_options = [
+      'CIT-Setup-Scheduled',
+      'CIT-Setup-Unscheduled-MIT',
+      'CIT-Setup-Unscheduled',
+      'CIT-Subsequent-Unscheduled',
+      'MIT-Subsequent-Scheduled',
+      'MIT-Subsequent-Unscheduled'
+    ]
+
+    reach_options.each do |payment_model|
+      @options[:payment_model] = payment_model
+      stub_comms do
+        @gateway.purchase(@amount, @credit_card, @options)
+      end.check_request do |_endpoint, data, _headers|
+        request = JSON.parse(URI.decode_www_form(data)[0][1])
+        assert_equal payment_model, request['PaymentModel']
+      end.respond_with(successful_purchase_response)
+    end
+  end
+
   def test_stored_credential_with_wrong_combination_stored_credential_paramaters
     @options[:stored_credential] = { initiator: 'merchant', initial_transaction: true, reason_type: 'unschedule' }
     e = assert_raise ArgumentError do
--- a/test/remote/gateways/remote_reach_test.rb
+++ b/test/remote/gateways/remote_reach_test.rb
@@ -142,6 +142,36 @@ class RemoteReachTest < Test::Unit::TestCase
     assert_success response
   end
 
+  def test_successful_purchase_store_credentials_with_payment_model_option
+    @options[:payment_model] = 'CIT-Subsequent-Unscheduled'
+    response = @gateway.purchase(@amount, @credit_card, @options)
+
+    assert_success response
+
+    assert response.params['response']['Authorized']
+    assert response.params['response']['OrderId']
+  end
+
+  def test_successful_purchase_store_credentials_with_payment_model_option_mit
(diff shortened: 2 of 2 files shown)


sinourain (approved):
LGTM

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -128,7 +128,7 @@ def build_checkout_request(amount, payment, options)
       end
 
       def add_payment(payment, options)
-        ntid = options.dig(:stored_credential, :network_transaction_id)
+        ntid = options.dig(:stored_credential, :network_transaction_id) if options[:stored_credential]
```
Heavyblade, line 131: There is not need of the `if` check because the `dig` method is being used

lib/active_merchant/billing/gateways/reach.rb
```diff
@@ -45,7 +45,7 @@ def authorize(money, payment, options = {})
         add_customer_data(request, options, payment)
         add_stored_credentials(request, options)
         post = { request: request, card: add_payment(payment, options) }
-        if options[:stored_credential]
+        if options[:stored_credential] || options[:payment_model]
```
Heavyblade, line 48: Can't be an `or` here, `:stored_credential` needs to be present, the point on the GSF is to provide the merchant with the posibility of overwrite the paymentModel, but the paymentReference still needs to be there.

test/unit/gateways/reach_test.rb
```diff
@@ -163,6 +163,28 @@ def test_stored_credential_with_no_store_credential_parameters
...
 
+  def test_stored_credential_with_reach_payment_model
+    reach_options = [
+      { payment_model: 'CIT-Setup-Scheduled' },
+      { payment_model: 'CIT-Setup-Unscheduled-MIT' },
+      { payment_model: 'CIT-Setup-Unscheduled' },
+      { payment_model: 'CIT-Subsequent-Unscheduled' },
+      { payment_model: 'MIT-Subsequent-Scheduled' },
+      { payment_model: 'MIT-Subsequent-Unscheduled' }
+    ]
```
Heavyblade, line 174: Comment:

What is the point on having a hash in something that should be an array ?

something like:

```ruby
  reach_options = [
        'CIT-Setup-Scheduled',
        'CIT-Setup-Unscheduled-MIT',
        'CIT-Setup-Unscheduled',
        'CIT-Subsequent-Unscheduled',
        'MIT-Subsequent-Scheduled',
        'MIT-Subsequent-Unscheduled'
  ]

    reach_options.each do |payment_model|
      @options[:payment_model] = payment_model
```

Heavyblade:
Small comment => Check the PR description, the "Spreedly's determination" is kind confusing

lib/active_merchant/billing/gateways/reach.rb
javierpedrozaing: you're right I will remove it, thanks

test/unit/gateways/reach_test.rb
javierpedrozaing: Oh yes, make sense, I will update it, thanks

lib/active_merchant/billing/gateways/reach.rb
javierpedrozaing: ok I see, I will remote it, thanks

naashton:
It looks like this PR just adds tests to validate the support for the existing feature described in the commit message. Is this correct?

javierpedrozaing:
> It looks like this PR just adds tests to validate the support for the existing feature described in the commit message. Is this correct?

Hi @naashton yes, because previously we were already using the[ `payment_model` as GSF](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/reach.rb#L160)

naashton (approved):
This looks good

curiousepic:

PR #4661, opened 2022-12-29
Credorax: enable google pay apple pay

edgarv09:
Summary:
Enable the Credorax gateway to support payments via google pay and apple pay 

Remote Test:
54 tests, 66 assertions, 0 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications
100% passed

Unit Tests:
5416 tests, 76949 assertions, 0 failures, 0 errors, 0 pendings,
0 omissions, 0 notifications
100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -28,6 +28,11 @@ module ActiveMerchant #:nodoc:
       self.money_format = :cents
       self.supported_cardtypes = %i[visa master maestro american_express jcb discover diners_club]
 
+      NETWORK_TOKENIZATION_CARD_SOURCE = {
+        'apple_pay' => 'applepay',
+        'google_pay' => 'googlepay'
+      }
+
       RESPONSE_MESSAGES = {
         '00' => 'Approved or completed successfully',
         '01' => 'Refer to card issuer',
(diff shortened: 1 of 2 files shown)


javierpedrozaing (commented):
Looks good, would be a good idea to add a couple of unit test in order to check the request.

edgarv09:
> Looks good, would be a good idea to add a couple of unit test in order to check the request.

Could you give me an example of that kind of test?

javierpedrozaing:
> 

Hi @edgarv09 for sure, you can use [this example](https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/worldpay_test.rb#L1359)

curiousepic:
@edgarv09 I'm getting "expired card" messages for the remote tests now, probably due to the new year. I'm not sure where/how expiry dates are getting added though, which means it might be the gateway-provided test card numbers themselves that have expired...

curiousepic:
Actually, even after fixing the exp date, I'm getting a wide variety of other types of errors in the remote tests (13 total) with messages including   "Issuer not available" and  "Transaction has been denied. Invalid parameter c2. Originating component 3DS SERVER."  Probably nothing related with this change though. Let me know if you are also getting those as well.

edgarv09:
> Actually, even after adjusting the exp date, I'm getting a wide variety of other types of errors in the remote tests (13 total) with messages including "Issuer not available" and "Transaction has been denied. Invalid parameter c2. Originating component 3DS SERVER." Probably nothing related with this change though. Let me know if you are also getting those as well.

When I execute remote tests in my local environment am getting those errors but the CI is running correctly.

curiousepic:
Thanks for updating the cards. FYI the CI tests here don't run remote tests, so we have to rely on the local runs. Anyway, I see the same failures in the prior commit/PR and they are unrelated, so not a problem for this change. But do please squash the commits before we merge, thanks!

edgarv09:
> @edgarv09 Thanks for updating the cards. FYI the CI tests here don't run remote tests, so we have to rely on the local runs. Anyway, I see the same failures in the prior commit/PR and they are unrelated, so not a problem for this change. But do please squash the commits before we merge, thanks!

squash done. 💪🏽

PR #4664, opened 2022-12-30
Rapyd: Add customer object to transactions

javierpedrozaing:
Description
-------------------------
Rapyd Gateway send customer info on store transactions, with this commit it will be able to send customer on authorize and purchase transaction, when it use a US “PMT”s include the addresses object into the customer data in order to be able to perform US transactions properly.

GWI-489

Unit test
-------------------------

Finished in 0.123245 seconds.
22 tests, 103 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote test
-------------------------

Finished in 100.082672 seconds.
33 tests, 92 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
0.33 tests/s, 0.92 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -11,6 +11,8 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'https://www.rapyd.net/'
       self.display_name = 'Rapyd Gateway'
 
+      USA_PAYMENT_METHODS = %w[us_debit_discover_card us_debit_mastercard_card us_debit_visa_card us_ach_bank]
+
       STANDARD_ERROR_CODE_MAPPING = {}
 
       def initialize(options = {})
@@ -98,6 +100,7 @@ module ActiveMerchant #:nodoc:
       def add_auth_purchase(post, money, payment, options)
         add_invoice(post, money, options)
         add_payment(post, payment, options)
+        add_customer_object(post, payment, options)
         add_3ds(post, payment, options)
         add_address(post, payment, options)
         add_metadata(post, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -11,6 +11,8 @@ class RapydGateway < Gateway
       self.homepage_url = 'https://www.rapyd.net/'
       self.display_name = 'Rapyd Gateway'
 
+      USA_PAYMENT_METHODS = %w[us_debit_discover_card us_debit_mastercard_card us_debit_visa_card us_ach_bank]
```
edgarv09, line 14: am curious about this change, What is the purpose of this line? I have not seen where it is used.

javierpedrozaing: Ohm you are right I forgot to push my last changes, it is used to validate if is a US PMT to include customer object data, Thank you

edgarv09 (approved):
🤖

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -211,8 +214,15 @@ def add_payment_urls(post, options)
...
-        post[:phone_number] = options[:billing_address][:phone].gsub(/\D/, '') if options[:billing_address]
+        if options[:customer]
+          post[:name] = options.dig(:customer, :name)
+          post[:phone_number] = options.dig(:customer, :phone_number)
+          post[:email] = options.dig(:customer, :email)
+          post[:addresses] = options.dig(:customer, :addresses)
+        end
+
+        post[:name] = "#{payment.first_name} #{payment.last_name}" unless payment.is_a?(String)
+        post[:phone_number] = options[:billing_address][:phone].gsub(/\D/, '') unless options[:billing_address].nil?
```
naashton, line 225: We should be able to derive these fields from the `billing_address` set of fields, within the `options` hash.

naashton (changes requested):

lib/active_merchant/billing/gateways/rapyd.rb
naashton: We should get the fields from the `options[:billing_address]` instead of `options[:customer]`.

javierpedrozaing:
Hi @naashton I did some changes, let me know if these looks ok, thanks

aenand (approved):
LGTM :shipit:

naashton (approved):

PR #4675, opened 2023-01-11
Avoid anonymous eval

casperisfine:
That makes it hard to locate code when profiling etc.

### DIFF
--- a/lib/active_merchant/billing/response.rb
+++ b/lib/active_merchant/billing/response.rb
@@ -100,11 +100,11 @@ module ActiveMerchant #:nodoc:
       end
 
       %w(params message test authorization error_code emv_authorization test? fraud_review?).each do |m|
-        class_eval %(
+        class_eval <<~RUBY, __FILE__, __LINE__ + 1
           def #{m}
             (@responses.empty? ? nil : primary_response.#{m})
           end
-        )
+        RUBY
       end
     end
   end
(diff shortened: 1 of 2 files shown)


casperisfine:
@naashton @DustinHaefele could I get this merged please, It's really a trivial change.

lib/active_merchant/billing/response.rb
```diff
@@ -100,11 +100,11 @@ def cvv_result
       end
 
       %w(params message test authorization error_code emv_authorization test? fraud_review?).each do |m|
-        class_eval %(
+        class_eval <<~RUBY, __FILE__, __LINE__ + 1
```
Heavyblade, line 103: 💬 Comment:

Hey @casperisfine / @naashton if we are going to pay attention to this, what about removing that dirty eval entirely:

```ruby
      %i(params message test authorization error_code emv_authorization test? fraud_review?).each do |m|
        define_method m do |*args, &block|
          @responses.empty? ? nil : primary_response.send(m, *args, &block)
        end
      end

```

casperisfine: It's debatable. `define_method` creates `BMETHOD` which are slower to call and can lead to leaks because they hold on the surrounding context.

All this to say that `define_method` isn't inherently cleaner than `class_eval`. In Rails our preference goes to `eval` over `define_method`.

Heavyblade: Ohhh that's very interesting

casperisfine:
This isn't stale at all, still worth merging...

PR #4676, opened 2023-01-11
GlobalCollect: fix inquire bug

almalee24:


naashton (approved):
This looks like a reasonable change. I assume this is what the method signature should look like when `inquire` is implemented for a gateway.

🚢

almalee24:
> This looks like a reasonable change. I assume this is what the method signature should look like when `inquire` is implemented for a gateway.
> 
> 🚢

Thanks, right because some gateways could use `order_id` for inquire. Currently, this is causing a error in Core because it's expecting to pass two parameters

PR #4677, opened 2023-01-11
CheckoutV2: Add store/unstore

gasb150:
Summary:
    In order to use Third Party Vaulting (TPV), this commit adds store and unstore methods.

Remote test
    -----------------------
    Finished in 240.268913 seconds.
    82 tests, 203 assertions, 2 failures, 0 errors,
    97.561% passed

Unit test
    -----------------------
    Finished in 0.268913 seconds.
    52 tests, 289 assertions, 0 failures, 0 errors,
    100% passed

Rubocop
    -----------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -19,12 +19,14 @@ module ActiveMerchant #:nodoc:
       def initialize(options = {})
         @options = options
         @access_token = nil
-        begin
+
+        if options.has_key?(:secret_key)
           requires!(options, :secret_key)
-        rescue ArgumentError
+        else
           requires!(options, :client_id, :client_secret)
           @access_token = setup_access_token
         end
+
         super
       end
 
@@ -39,7 +41,6 @@ module ActiveMerchant #:nodoc:
         post = {}
         post[:capture] = false
         build_auth_or_purchase(post, amount, payment_method, options)
-
         options[:incremental_authorization] ? commit(:incremental_authorize, post, options[:incremental_authorization]) : commit(:authorize, post)
       end
 
@@ -86,7 +87,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def verify_payment(authorization, option = {})
-        commit(:verify_payment, authorization)
+        commit(:verify_payment, nil, authorization, :get)
       end
 
       def supports_scrubbing?
@@ -99,7 +100,34 @@ module ActiveMerchant #:nodoc:
           gsub(/("number\\":\\")\d+/, '\1[FILTERED]').
           gsub(/("cvv\\":\\")\d+/, '\1[FILTERED]').
           gsub(/("cryptogram\\":\\")\w+/, '\1[FILTERED]').
-          gsub(/(source\\":\{.*\\"token\\":\\")\d+/, '\1[FILTERED]')
+          gsub(/(source\\":\{.*\\"token\\":\\")\d+/, '\1[FILTERED]').
+          gsub(/("token\\":\\")\w+/, '\1[FILTERED]')
+      end
+
+      def store(payment_method, options = {})
+        post = {}
+        MultiResponse.run do |r|
+          if payment_method.is_a?(NetworkTokenizationCreditCard)
+            r.process { verify(payment_method, options) }
+            break r unless r.success?
+
+            r.params['source']['customer'] = r.params['customer']
+            r.process { response(:store, true, r.params['source']) }
+          else
+            r.process { tokenize(payment_method, options) }
+            break r unless r.success?
+
+            token = r.params['token']
+            add_payment_method(post, token, options)
+            post.merge!(post.delete(:source))
+            add_customer_data(post, options)
+            r.process { commit(:store, post) }
+          end
+        end
+      end
+
+      def unstore(id, options = {})
+        commit(:unstore, nil, id, :delete)
       end
 
       private
@@ -142,7 +170,8 @@ module ActiveMerchant #:nodoc:
 
       def add_payment_method(post, payment_method, options, key = :source)
         post[key] = {}
-        if payment_method.is_a?(NetworkTokenizationCreditCard)
+        case payment_method
+        when NetworkTokenizationCreditCard
           token_type = token_type_from(payment_method)
           cryptogram = payment_method.payment_cryptogram
           eci = payment_method.eci || options[:eci]
@@ -153,7 +182,7 @@ module ActiveMerchant #:nodoc:
           post[key][:token_type] = token_type
           post[key][:cryptogram] = cryptogram if cryptogram
           post[key][:eci] = eci if eci
-        elsif payment_method.is_a?(CreditCard)
+        when CreditCard
           post[key][:type] = 'card'
           post[key][:name] = payment_method.name
           post[key][:number] = payment_method.number
@@ -169,7 +198,14 @@ module ActiveMerchant #:nodoc:
             post[key][:last_name] = payment_method.last_name if payment_method.last_name
           end
         end
-        unless payment_method.is_a?(String)
+        if payment_method.is_a?(String)
+          if /tok/.match?(payment_method)
+            post[:type] = 'token'
+            post[:token] = payment_method
+          else
+            add_source(post, options)
+          end
+        elsif payment_method.try(:year)
           post[key][:expiry_year] = format(payment_method.year, :four_digits)
           post[key][:expiry_month] = format(payment_method.month, :two_digits)
         end
@@ -280,11 +316,12 @@ module ActiveMerchant #:nodoc:
         response['access_token']
       end
 
-      def commit(action, post, authorization = nil)
+      def commit(action, post, authorization = nil, method = :post)
         begin
-          raw_response = (action == :verify_payment ? ssl_get("#{base_url}/payments/#{post}", headers) : ssl_post(url(post, action, authorization), post.to_json, headers))
+          raw_response = ssl_request(method, url(action, authorization), post.to_json, headers(action))
           response = parse(raw_response)
           response['id'] = response['_links']['payment']['href'].split('/')[-1] if action == :capture && response.key?('_links')
+          source_id = authorization if action == :unstore
         rescue ResponseError => e
           raise unless e.response.code.to_s =~ /4\d\d/
 
@@ -293,45 +330,62 @@ module ActiveMerchant #:nodoc:
 
         succeeded = success_from(action, response)
 
-        response(action, succeeded, response)
+        response(action, succeeded, response, source_id)
       end
 
-      def response(action, succeeded, response)
+      def response(action, succeeded, response, source_id = nil)
         successful_response = succeeded && action == :purchase || action == :authorize
         avs_result = successful_response ? avs_result(response) : nil
         cvv_result = successful_response ? cvv_result(response) : nil
-
+        authorization = authorization_from(response) unless action == :unstore
+        body = action == :unstore ? { response_code: response.to_s } : response
         Response.new(
           succeeded,
           message_from(succeeded, response),
-          response,
-          authorization: authorization_from(response),
-          error_code: error_code_from(succeeded, response),
+          body,
+          authorization: authorization,
+          error_code: error_code_from(succeeded, body),
           test: test?,
           avs_result: avs_result,
           cvv_result: cvv_result
         )
       end
 
-      def headers
+      def headers(action)
         auth_token = @access_token ? "Bearer #{@access_token}" : @options[:secret_key]
+        auth_token = @options[:public_key] if action == :tokens
         {
           'Authorization' => auth_token,
           'Content-Type' => 'application/json;charset=UTF-8'
         }
       end
 
-      def url(_post, action, authorization)
-        if %i[authorize purchase credit].include?(action)
+      def tokenize(payment_method, options = {})
+        post = {}
+        add_authorization_type(post, options)
+        add_payment_method(post, payment_method, options)
+        add_customer_data(post, options)
+        commit(:tokens, post[:source])
+      end
+
+      def url(action, authorization)
+        case action
+        when :authorize, :purchase, :credit
           "#{base_url}/payments"
-        elsif action == :capture
+        when :unstore, :store
+          "#{base_url}/instruments/#{authorization}"
+        when :capture
           "#{base_url}/payments/#{authorization}/captures"
-        elsif action == :refund
+        when :refund
           "#{base_url}/payments/#{authorization}/refunds"
-        elsif action == :void
+        when :void
           "#{base_url}/payments/#{authorization}/voids"
-        elsif action == :incremental_authorize
+        when :incremental_authorize
           "#{base_url}/payments/#{authorization}/authorizations"
+        when :tokens
+          "#{base_url}/tokens"
+        when :verify_payment
+          "#{base_url}/payments/#{authorization}"
         else
           "#{base_url}/payments/#{authorization}/#{action}"
         end
@@ -363,7 +417,12 @@ module ActiveMerchant #:nodoc:
 
       def success_from(action, response)
         return response['status'] == 'Pending' if action == :credit
+        return true if action == :unstore && response == 204
 
+        store_response = response['token'] || response['id']
+        if store_response
+          return true if (action == :tokens && store_response.match(/tok/)) || (action == :store && store_response.match(/src_/))
+        end
         response['response_summary'] == 'Approved' || response['approved'] == true || !response.key?('response_summary') && response.key?('action_id')
       end
 
@@ -414,6 +473,16 @@ module ActiveMerchant #:nodoc:
           'googlepay'
         when :apple_pay
           'applepay'
+        end
+      end
+
+      def handle_response(response)
+        case response.code.to_i
+        # to get the response code after unstore(delete instrument), because the body is nil
+        when 200...300
+          response.body || response.code
+        else
+          raise ResponseError.new(response)
         end
       end
     end
--- a/test/fixtures.yml
+++ b/test/fixtures.yml
@@ -198,6 +198,10 @@ checkout_v2:
   client_id: CLIENT_ID_FOR_OAUTH_TRANSACTIONS
   client_secret: CLIENT_SECRET_FOR_OAUTH_TRANSACTIONS
 
+checkout_v2_token:
+  secret_key: sk_sbox_xxxxxxxxxxxxxxxxx
+  public_key: pk_sbox_xxxxxxxxxxxxxxxxx
+
 citrus_pay:
   userid: CPF00001
   password: 7c70414732de7e0ba3a04db5f24fcec8
--- a/test/remote/gateways/remote_checkout_v2_test.rb
+++ b/test/remote/gateways/remote_checkout_v2_test.rb
@@ -3,8 +3,10 @@ require 'test_helper'
 class RemoteCheckoutV2Test < Test::Unit::TestCase
   def setup
     gateway_fixtures = fixtures(:checkout_v2)
+    gateway_token_fixtures = fixtures(:checkout_v2_token)
     @gateway = CheckoutV2Gateway.new(secret_key: gateway_fixtures[:secret_key])
     @gateway_oauth = CheckoutV2Gateway.new({ client_id: gateway_fixtures[:client_id], client_secret: gateway_fixtures[:client_secret] })
+    @gateway_token = CheckoutV2Gateway.new(secret_key: gateway_token_fixtures[:secret_key], public_key: gateway_token_fixtures[:public_key])
 
     @amount = 200
     @credit_card = credit_card('4242424242424242', verification_value: '100', month: '6', year: '2025')
@@ -132,6 +134,16 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
     assert_scrubbed(@apple_pay_network_token.payment_cryptogram, transcript)
     assert_scrubbed(@apple_pay_network_token.number, transcript)
     assert_scrubbed(@gateway.options[:secret_key], transcript)
+  end
+
+  def test_store_transcript_scrubbing
+    response = nil
+    transcript = capture_transcript(@gateway) do
+      response = @gateway_token.store(@credit_card, @options)
+    end
+    token = response.responses.first.params['token']
+    transcript = @gateway.scrub(transcript)
+    assert_scrubbed(token, transcript)
   end
 
   def test_successful_purchase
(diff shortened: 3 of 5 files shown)


Heavyblade:
@gasb150 you need to add a new item in test/fixtures.yml with the expected structure of the `:checkout_v2_api`

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -169,7 +193,12 @@ def add_payment_method(post, payment_method, options, key = :source)
             post[key][:last_name] = payment_method.last_name if payment_method.last_name
           end
         end
-        unless payment_method.is_a?(String)
+        if payment_method.is_a?(String)
```
Heavyblade, line 196: 💬  Comment:

Just to follow the previous logi, shouldn't be rasonable here to use:

```ruby
  elsif payment_method.is_a?(String)
```

or maybe is a good time for a refactor and add a case statement here ?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -169,7 +193,12 @@ def add_payment_method(post, payment_method, options, key = :source)
...
+          if /tok/.match?(payment_method)
+            post[key][:type] = 'token'
```
Heavyblade, line 198: 💬 Comment:

Not sure if the proper behavior is to name space this post[:source][:type] when you are going to remove the `source` level on the `store` method, should be simpler just to put `type` and `token` at the root level of the post object ?

❓ Question:

I can uderstand why you need to move `type` and `token`  one level up, but it aslo applies for billing address ?

```ruby
### Before
{:source=>
  {:type=>"token",
   :token=>"tok_nlnv2zsawtbulj577oxrzaa7pe",
   :billing_address=>
    {:address_line1=>"456 My Street",
     :address_line2=>"Apt 1",
     :city=>"Ottawa",
     :state=>"ON",
     :country=>"CA",
     :zip=>"K1C2N6"}},
 :customer=>{:email=>"longbob.longsen@example.com"}}

## After
{:customer=>{:email=>"longbob.longsen@example.com"},
 :type=>"token",
 :token=>"tok_nlnv2zsawtbulj577oxrzaa7pe",
 :billing_address=>
  {:address_line1=>"456 My Street",
   :address_line2=>"Apt 1",
   :city=>"Ottawa",
   :state=>"ON",
   :country=>"CA",
   :zip=>"K1C2N6"}}
```

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -99,7 +101,29 @@ def scrub(transcript)
...
-          gsub(/(source\\":\{.*\\"token\\":\\")\d+/, '\1[FILTERED]')
+          gsub(/(source\\":\{.*\\"token\\":\\")\d+/, '\1[FILTERED]').
+          gsub(/("token\\":\\")\w+/, '\1[FILTERED]')
+      end
+
+      def store(payment_method, options = {})
+        post = {}
+        MultiResponse.run do |r|
+          response = r.process { tokenize(payment_method, options) }
+          if response.success?
```
Heavyblade, line 112: 💬 Comment:

Just to prevent if nesting, what do you think about something like this:

```Ruby
MultiResponse.run do |r|
  response = r.process { tokenize(payment_method, options) }
  break unless response.success?

  token = response.params['token']
  add_payment_method(post, token, options)
  post = post.merge(post.delete(:source))
  r.process { commit(:store, post) }
end
```

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -321,17 +350,45 @@ def headers
...
+
+          response = parse(e.response.body, error: e.response)
+        end
+        response(:tokens, succeeded, response)
+      end
+
       def url(_post, action, authorization)
-        if %i[authorize purchase credit].include?(action)
+        case action
+        when ->(a) { %i[authorize purchase credit].include?(a) }
```
Heavyblade, line 378: 💬 Comment:

Small ruby trick => you achieve the same with the splat operator:

```ruby
      def url(_post, action, authorization)
        case action
        when *[:authorize, :purchase, :credit]
          "#{base_url}/payments"
        when *[:store, :unstore]
          "#{base_url}/instruments/#{authorization}"
        # ...
        # ...
        else
          "#{base_url}/payments/#{authorization}/#{action}"
        end
      end

```

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -416,6 +474,16 @@ def token_type_from(payment_method)
           'applepay'
         end
       end
+
+      def handle_response(response)
```
Heavyblade, line 478: 💬 Comment:

Handle response is not being used anywhere.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -99,7 +101,29 @@ def scrub(transcript)
...
+            add_customer_data(post, options)
+            post = post.merge(post.delete(:source))
+            r.process { commit(:store, post) }
+          end
+        end
+      end
+
+      def unstore(id, options = {})
+        post = {}
+        succeeded = ssl_request(:delete, url(post, :unstore, id), nil, headers) == '204'
```
Heavyblade, line 124: 💬 Comment:
You are creating an empty post object to send it to a method that is not going to use it, maybe is a good time for an small refactor and remove the post param on the `url` method ?, is only used on 4 places.

Heavyblade (changes requested):
@gasb150 done with the review left some comments for you

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -321,17 +350,45 @@ def headers
...
+        add_customer_data(post, options)
+        begin
+          response = parse(ssl_request(:post, url(post, :tokens, nil), post[:source].to_json, public_key_headers))
+          succeeded = success_from(:tokens, response)
+        rescue ResponseError => e
+          raise unless e.response.code.to_s =~ /4\d\d/
+
+          response = parse(e.response.body, error: e.response)
+        end
+        response(:tokens, succeeded, response)
```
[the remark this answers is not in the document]
gasb150: I'm following the pattern, but I can rename it to avoid possible issues.

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: This example doesn't apply because, after the tokenizing process, the add_customer_data only will add the :customer=>{:email=>"longbob.longsen@example.com"}. The reason the add_customer_data requires the existence of post[:source] to add billing_address information. So I will need to use add_payment_method, which also adds the source hash, and add to this source hash all the billing address information through the add_customer_data, and then move the type and token to one level up and with that the billing address, at the end still working in this way.



```ruby
# token request
### before
{:source=>
  {:type=>"card",
   :name=>"Longbob Longsen",
   :number=>"4242424242424242",
   :cvv=>"100",
   :first_name=>"Longbob",
   :last_name=>"Longsen",
   :expiry_year=>"2025",
   :expiry_month=>"06",
   :billing_address=>{:address_line1=>"456 My Street", :address_line2=>"Apt 1", :city=>"Ottawa", :state=>"ON", :country=>"CA", :zip=>"K1C2N6"}},
 :customer=>{:email=>"longbob.longsen@example.com"}}

### after
{:type=>"card",
 :name=>"Longbob Longsen",
 :number=>"4242424242424242",
 :cvv=>"100",
 :first_name=>"Longbob",
 :last_name=>"Longsen",
 :expiry_year=>"2025",
 :expiry_month=>"06",
 :billing_address=>{:address_line1=>"456 My Street", :address_line2=>"Apt 1", :city=>"Ottawa", :state=>"ON", :country=>"CA", :zip=>"K1C2N6"}}
```

```
### store request
{:type=>"token", :token=>"tok_fytopfhgnimuxb753662nzymre", :customer=>{:email=>"longbob.longsen@example.com"}}
``
`
We don't need to add more billing_address values because the stored card has the information, and the request-response contains the customer_id and email associated.

File: lib/active_merchant/billing/gateways/checkout_v2.rb (hunk 1 above)
Comment (gasb150) [reply lib/active_merchant/billing/gateways/checkout_v2.rb:196]: I'll try with case statements.

**Review by jherreraa — COMMENTED** [review]
No more comments besides Cristian's.
LGTM! 👍

File: lib/active_merchant/billing/gateways/checkout_v2.rb [hunk 8]
```diff
@@ -321,17 +350,45 @@ def headers
...
+      def tokenize(payment_method, options = {})
+        post = {}
+        add_authorization_type(post, options)
+        add_payment_method(post, payment_method, options)
+        add_customer_data(post, options)
+        begin
+          response = parse(ssl_request(:post, url(post, :tokens, nil), post[:source].to_json, public_key_headers))
+          succeeded = success_from(:tokens, response)
+        rescue ResponseError => e
+          raise unless e.response.code.to_s =~ /4\d\d/
```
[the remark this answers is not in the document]
gasb150: I'm following the pattern in the gateway itself. But looks clearest in the way you mention.

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: Nice

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: Yes, that's right

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: It's used in the unstore method. I overwrite it to get the response code and create a response properly.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -19,12 +19,14 @@ class CheckoutV2Gateway < Gateway
       def initialize(options = {})
         @options = options
         @access_token = nil
-        begin
+
+        if options.has_key?(:secret_key)
           requires!(options, :secret_key)
-        rescue ArgumentError
+        else
```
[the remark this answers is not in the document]
gasb150: It's easy to read and achieve the same objective.

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -583,6 +595,36 @@ def test_successful_credit
...
+    store = @gateway_api.store(@credit_card, @options)
+    assert_success store
+    assert_equal 'Succeeded', store.message
+    source_id = store.params['id']
+    response = @gateway_api.unstore(source_id, @options)
+    assert_success response
+    assert_match "The instrument #{source_id} has bein deleted/unstored", response.message
+  end
+
+  def test_successful_unstore_after_purchase
```
Heavyblade, line 620: 💬 Comment:

If this is already working like this, we need to raise our hand and ask if is OK, is weird to store thhe payment method by default.

Heavyblade (approved):
@gasb150 looks ok, leave one final comment and something to take into account and ask Spreedly.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -99,7 +101,28 @@ def scrub(transcript)
...
+          add_payment_method(post, token, options)
+          post.merge!(post.delete(:source))
+          add_customer_data(post, options)
+          r.process { commit(:store, post) }
+        end
+      end
+
+      def unstore(id, options = {})
+        succeeded = ssl_request(:delete, url(:unstore, id), nil, headers) == '204'
+        message = "The instrument #{id} has bein deleted/unstored"
```
AMHOL, line 124:
```suggestion
        message = "The instrument #{id} has been deleted/unstored"
```

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -583,6 +595,36 @@ def test_successful_credit
...
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    source_id = purchase.params['source']['id']
+    response = @gateway.unstore(source_id, @options)
+    assert_success response
+    assert_match "The instrument #{source_id} has bein deleted/unstored", response.message
```
AMHOL, line 625:
```suggestion
    assert_match "The instrument #{source_id} has been deleted/unstored", response.message
```

gasb150: thanks

test/fixtures.yml
```diff
@@ -198,6 +198,10 @@ checkout_v2:
   client_id: CLIENT_ID_FOR_OAUTH_TRANSACTIONS
   client_secret: CLIENT_SECRET_FOR_OAUTH_TRANSACTIONS
 
+checkout_v2_api:
+  secret_key: sk_sbox_xxxxxxxxxxxxxxxxx
+  public_key: pk_sbox_xxxxxxxxxxxxxxxxx
```
naashton, line 203: I don't think we need this, the `checkout_v2` entry in the fixtures has the same fields.

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -3,8 +3,10 @@
 class RemoteCheckoutV2Test < Test::Unit::TestCase
   def setup
     gateway_fixtures = fixtures(:checkout_v2)
+    gateway_api_fixtures = fixtures(:checkout_v2_api)
     @gateway = CheckoutV2Gateway.new(secret_key: gateway_fixtures[:secret_key])
     @gateway_oauth = CheckoutV2Gateway.new({ client_id: gateway_fixtures[:client_id], client_secret: gateway_fixtures[:client_secret] })
+    @gateway_api = CheckoutV2Gateway.new(secret_key: gateway_api_fixtures[:secret_key], public_key: gateway_api_fixtures[:public_key])
```
naashton, line 9: Is there a particular reason we need to use a different gateway setup from the existing one?

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -134,6 +136,16 @@ def test_network_transaction_scrubbing
     assert_scrubbed(@gateway.options[:secret_key], transcript)
   end
 
+  def test_store_transcript_scrubbing
+    response = nil
+    transcript = capture_transcript(@gateway) do
+      response = @gateway_api.store(@credit_card, @options)
+    end
+    token = response.responses.first.params['token']
```
naashton, line 144: I'm getting this error when I run the tests:

`NoMethodError: undefined method 'responses' for #<ActiveMerchant::Billing::Response:0x00007f901c164b40>`

naashton (changes requested):
It looks like the remote tests for `store` and `unstore` are failing for me. Do we need a different set of credentials to run these transactions? 

I also had a few suggestions, but overall it looks great.

test/remote/gateways/remote_checkout_v2_test.rb
gasb150: The value in @gateway_api for the secret_key and the public_key are different from the @gateway, the same secret_key used in the @gateway is not going to work in the transactions that @gatway_api works and the @gateway_api secret_key is not going to work in the transactions that @gateway does.

test/fixtures.yml
gasb150: Yes, but the values are different from the checkout_v2 and the chekout_v2_api.

could works add something like this

checkout_v2:
  secret_key: SECRET_KEY_FOR_BASIC_TRANSACTIONS
  client_id: CLIENT_ID_FOR_OAUTH_TRANSACTIONS
  client_secret: CLIENT_SECRET_FOR_OAUTH_TRANSACTIONS
  secret_api_key: sk_sbox_xxxxxxxxxxxxxxxxx
  secret_public_key: pk_sbox_xxxxxxxxxxxxxxxxx

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -99,7 +101,28 @@ def scrub(transcript)
...
+          post.merge!(post.delete(:source))
+          add_customer_data(post, options)
+          r.process { commit(:store, post) }
+        end
+      end
+
+      def unstore(id, options = {})
+        succeeded = ssl_request(:delete, url(:unstore, id), nil, headers) == '204'
+        message = "The instrument #{id} has been deleted/unstored"
+        Response.new(succeeded, message)
```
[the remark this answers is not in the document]
gasb150: yes, that's possible.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -321,17 +350,45 @@ def headers
...
+        add_authorization_type(post, options)
+        add_payment_method(post, payment_method, options)
+        add_customer_data(post, options)
+        begin
+          r = parse(ssl_request(:post, url(:tokens, nil), post[:source].to_json, public_key_headers))
+          succeeded = success_from(:tokens, r)
+        rescue ResponseError => e
+          raise unless (400..499).cover?(e.response.code.to_i)
+
+          r = parse(e.response.body, error: e.response)
```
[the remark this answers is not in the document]
gasb150: Yes, you are right.

test/remote/gateways/remote_checkout_v2_test.rb
gasb150: Could you try again after the changes?

test/fixtures.yml
naashton: Ah, if they are different then what you are doing is fine. Maybe we call it `checkout_v2_token` so it's a bit more descriptive.

The only other concern is, if our users need to have 2 sets of credentials, then we should handle it as one set of credentials and call the ones used for store/unstore something else.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -280,11 +307,12 @@ def setup_access_token
         response['access_token']
       end
 
-      def commit(action, post, authorization = nil)
+      def commit(action, post, authorization = nil, method = :post)
         begin
-          raw_response = (action == :verify_payment ? ssl_get("#{base_url}/payments/#{post}", headers) : ssl_post(url(post, action, authorization), post.to_json, headers))
+          raw_response = (action == :verify_payment ? ssl_get("#{base_url}/payments/#{post}", headers) : ssl_request(method, url(action, authorization), post.to_json, headers))
```
naashton, line 312: Nice work!

naashton, line 312: Sorry for coming back to this. I think the part of this statement, `"#{base_url}/payments/#{post}", headers`, can be added to the `def url` method as an additional switch/case?

Also, since we are providing a `method` can we just use `ssl_request`?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -321,17 +359,45 @@ def headers
...
       end
 
-      def url(_post, action, authorization)
-        if %i[authorize purchase credit].include?(action)
+      def public_key_headers
+        {
+          'Authorization' => @options[:public_key],
+          'Content-Type' => 'application/json;charset=UTF-8'
+        }
+      end
```
naashton, line 367: Seems like this is redundant and we can put a conditional in the `headers` method to use `public_key` if we are doing a `store` transaction.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -321,17 +359,45 @@ def headers
...
+        add_payment_method(post, payment_method, options)
+        add_customer_data(post, options)
+        begin
+          r = parse(ssl_post(url(:tokens, nil), post[:source].to_json, public_key_headers))
+          succeeded = success_from(:tokens, r)
+        rescue ResponseError => e
+          raise unless (400..499).cover?(e.response.code.to_i)
+
+          r = parse(e.response.body, error: e.response)
+        end
```
naashton, line 381: I think we should be able to just call `commit` and use the existing request logic there. And where your conditional for `unstore` handles a slimmed down response, we can likewise, do the same for this transaction type.

naashton (changes requested):
Just a few more suggestions. Thanks for your hard work, @gasb150!

naashton:
I'm also getting `401`'s on the `store` and `unstore` tests, but I think they passed on the last iteration.

naashton (commented):
Everything looks great, but I'm getting this error on the `store` `unstore` tests:
```
    {"error_type"=>"401",
     "message"=>
      "Invalid JSON response received from Checkout.com Unified Payments Gateway. Please contact Checkout.com if you continue to receive this message.",
     "raw_response"=>"",
     "error_codes"=>["Unauthorized"]}
```

Everything looks to be correct, but maybe there is one small detail that I'm missing.

naashton (approved):
Excellent work, @gasb150!

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -99,7 +100,34 @@ def scrub(transcript)
...
+      end
+
+      def store(payment_method, options = {})
+        post = {}
+        MultiResponse.run do |r|
+          if payment_method.is_a?(NetworkTokenizationCreditCard)
+            response = r.process { verify(payment_method, options) }
+            break response unless response.success?
+
+            response.params['source']['customer'] = r.params['customer']
```
Heavyblade, line 114: 💬 Comment:

Two things that are confusing unless you have context and know what you are reading:

1) response shadowing a method called also `response`
2) in line 114 you are using `response` and `r` to just level down `the customer` it has two effects:
  a) the original response is going the have `customer` at the top level and also in the `source` key.
  b) is very misleading using `response` and `r`, at this point, both point to the same object => the primary response of the Multiresponse.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -99,7 +100,34 @@ def scrub(transcript)
...
-          gsub(/(source\\":\{.*\\"token\\":\\")\d+/, '\1[FILTERED]')
+          gsub(/(source\\":\{.*\\"token\\":\\")\d+/, '\1[FILTERED]').
+          gsub(/("token\\":\\")\w+/, '\1[FILTERED]')
+      end
+
+      def store(payment_method, options = {})
+        post = {}
+        MultiResponse.run do |r|
+          if payment_method.is_a?(NetworkTokenizationCreditCard)
+            response = r.process { verify(payment_method, options) }
```
Heavyblade, line 111: 💬 Comment:

Maybe comment with an small explanation of this particular approach ?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -293,45 +330,71 @@ def commit(action, post, authorization = nil)
...
       end
 
-      def response(action, succeeded, response)
+      def response(action, succeeded, response, source_id = nil)
         successful_response = succeeded && action == :purchase || action == :authorize
         avs_result = successful_response ? avs_result(response) : nil
         cvv_result = successful_response ? cvv_result(response) : nil
 
+        if action == :unstore
+          message_text = "The instrument #{source_id} has been deleted/unstored"
```
Heavyblade, line 342: 💬 Comment:

Are you using this particular text for some reason, maybe the gateway just responds with 200 and that's all ?

on side effect for consideration => the merchant is going the get this response and going to think that is the actual message from the gateway, in case of some contact con the gateway support this could lead to misunderstandings.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -293,45 +330,71 @@ def commit(action, post, authorization = nil)
...
+          authorization = nil
```
Heavyblade, line 343: 💬 Comment:

You don't need to set `authorization` to nil, Ruby is going to do that for you.

Heavyblade (commented):
@gasb150 done with the review left some comments, please remove that extra file error.txt

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: 2. The idea is to have the same information displayed in a normal store response, that's why `r.params[:source]` is that I'll return in the response. Because in a store response, you'll see the same information in the `verify.params[:source]` except by the customer that is on a different level.

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: Because this comes from a Verify, I thought it could have two responses, the verify itself, and the "stored" (a formatted verify to look like a store)

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: Thanks

Heavyblade (approved):
PR looks good enough @gasb150, confirm with @curiousepic about the success message on the unstore call an we are good to go

lib/active_merchant/billing/gateways/checkout_v2.rb
naashton: I agree with @Heavyblade here. I actually don't think we need this conditional block. If no value is returned from the gateway, the default value should just be `nil` when we put it in the response. I think the current `message_from` and `authorization_from` helper methods should do that.

naashton (approved):
Excellent work!

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -280,11 +316,12 @@ def setup_access_token
         response['access_token']
       end
 
-      def commit(action, post, authorization = nil)
+      def commit(action, post, authorization = nil, method = :post)
         begin
-          raw_response = (action == :verify_payment ? ssl_get("#{base_url}/payments/#{post}", headers) : ssl_post(url(post, action, authorization), post.to_json, headers))
+          raw_response = ssl_request(method, url(action, authorization), post.to_json, headers(action))
```
aenand, line 321: I see there's a path where `post` can be `nil`. Does passing in `nil.to_json` have any issues?

naashton: I think because we initialize the `post` variable within each of the methods that eventually calls `commit`, the value of `post` should always at least be an empty hash. We could conceivably default the value to `{}`, but I think this is a common pattern in AM.

aenand (approved):
LGTM

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -19,12 +19,14 @@ class CheckoutV2Gateway < Gateway
...
-        begin
+
+        if options.has_key?(:secret_key)
           requires!(options, :secret_key)
-        rescue ArgumentError
+        else
           requires!(options, :client_id, :client_secret)
           @access_token = setup_access_token
         end
+
```
naashton, line 29: I think this might be the problematic change.

What happens with `requires!` is that it raises an `ArgumentError` which was being `rescue`d previously in the case where a user was using the OAuth authentication.

naashton: Sorry, this is not the root cause of the error we encountered

PR #4679, opened 2023-01-16
Enable the Credorax gateway to support payments via network_token

jherreraa:
Summary:
Enable the Credorax gateway to support payments via network_token 

Remote Test:
51 tests, 163 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
76.4706% passed

tests failures not related with the changes

Unit Tests:
82 tests, 393 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -30,7 +30,8 @@ module ActiveMerchant #:nodoc:
 
       NETWORK_TOKENIZATION_CARD_SOURCE = {
         'apple_pay' => 'applepay',
-        'google_pay' => 'googlepay'
+        'google_pay' => 'googlepay',
+        'network_token' => 'vts_mdes_token'
       }
 
       RESPONSE_MESSAGES = {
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -290,6 +291,8 @@ def add_payment_method(post, payment_method)
         post[:b5] = payment_method.verification_value
         post[:b4] = format(payment_method.year, :two_digits)
         post[:b3] = format(payment_method.month, :two_digits)
+        post[:token_eci] = payment_method&.eci if payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source.to_s == 'network_token'
+        post[:token_crypto] = payment_method&.payment_cryptogram if payment_method.is_a?(NetworkTokenizationCreditCard) && payment_method.source.to_s == 'network_token'
```
gasb150, line 295: 💭  Opinion:
I think it could be better if try something like this:
```
if payment_method.is_a?(NetworkTokenizationCreditCard)
  post[:token_eci] = payment_method&.eci if payment_method.source.to_s
  post[:token_crypto] = payment_method&.payment_cryptogram if payment_method.source.to_s == 'network_token'
end

  ```

File: test/remote/gateways/remote_credorax_test.rb [hunk 2]
```diff
@@ -79,6 +85,13 @@ def test_successful_purchase_with_google_pay
     assert_equal 'Succeeded', response.message
   end
 
+  def test_successful_purchase_with_network_token
+    response = @gateway.purchase(@amount, @nt_credit_card, @options)
+    assert_success response
+    assert_equal '1', response.params['H9']
+    assert_equal 'Succeeded', response.message
+  end
+
```
gasb150, line 94: 💭 Suggestion:

I think that, could be good to add some tests without `source: :network_token`

gasb150, line 94: 💬 

Add some unit tests to check the request and evaluate the conditionals.

gasb150 (commented):
LGTM, I left just a couple of comments.

test/remote/gateways/remote_credorax_test.rb
jherreraa: The proposed test would be useless because if payment source is not included in the adapter sources the default value is setted to 'apple_pay'

lib/active_merchant/billing/gateways/credorax.rb
jherreraa: updated with a different approach

test/remote/gateways/remote_credorax_test.rb
jherreraa: a pair of unit tests added to check the requests in both cases, network token and apple/google pay

curiousepic (approved):
👍 Looks good, merging

The git commit message has formatting errors (needs a title and empty line) but I'll fix them.

PR #4682, opened 2023-01-18
[WIP] Add cybersource rest gateway

Heavyblade:


lib/active_merchant/billing/gateways/cybersource_rest.rb
```diff
@@ -0,0 +1,104 @@
...
+          error_code: error_code_from(response)
+        )
+      end
+
+      # This implementation follows the detailed Cybersource guide on how create the
+      # request signature, see: https://developer.cybersource.com/docs/cybs/en-us/payments/developer/all/rest/payments/GenerateHeader/httpSignatureAuthentication.html
+      def get_http_signature(resource, digest, http_method = 'post', gmtdatetime = Time.now.httpdate)
+        string_to_sign = {
+          host: URI.parse(test? ? test_url : live_url).host,
+          date: gmtdatetime,
```
javierpedrozaing, line 58: we can use directly the Time.now.httpdate here instead, what happen if the date format put int the param is wrong ❓

lib/active_merchant/billing/gateways/cybersource_rest.rb
```diff
@@ -0,0 +1,104 @@
...
+      end
+
+      def sign_payload(payload)
+        decoded_key = Base64.decode64(@options[:private_key])
+        Base64.strict_encode64(OpenSSL::HMAC.digest('sha256', decoded_key, payload))
+      end
+
+      def auth_headers(action, post, http_method = 'post')
+        digest = "SHA-256=#{Digest::SHA256.base64digest(post.to_json)}" if post.present?
+        date = Time.now.http_method
```
javierpedrozaing, line 79:
```suggestion
        date = Time.now.httpdate
```

javierpedrozaing (commented):
Awesome work from scratch 💪 , just a couple of comments

lib/active_merchant/billing/gateways/cybersource_rest.rb
Heavyblade: @javierpedrozaing nop, because the date should be the same for the request http header and the one used in the signature, in fact, the most reasonable change is to remove the default value from the method signature.

lib/active_merchant/billing/gateways/cybersource_rest.rb
Heavyblade: Ohhh yes, a typo there

PR #4684, opened 2023-01-20
Payeezy change `method` on capture

naashton:
For Apple Pay transactions, the value for `method` is set to `3DS`, but when executing a `capture`, this value should be changed to `credit_card`.

This differs from other use cases where the value provided on auth transactions should be the same one given for capture.

Unit: 45 tests, 206 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

46 tests, 184 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


aenand (approved):
LGTM

aenand:
@naashton do you mind adding a code comment about why we are overwriting the method value?

naashton:
Will do!

PR #4688, opened 2023-01-26
Checkout: Add support for `allow_upgrade` field

rachelkirk:
CER-338

Whether to process this payment as 3D Secure if the authorization was soft declined due to 3DS authentication required

Remote:
74 tests, 188 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 98.6486% passed
*1 failing test is also failing on master, not due to my change

Unit:
51 tests, 281 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Local:
5441 tests, 77075 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -280,6 +280,7 @@ module ActiveMerchant #:nodoc:
           post[:'3ds'][:attempt_n3d] = options[:attempt_n3d] if options[:attempt_n3d]
           post[:'3ds'][:challenge_indicator] = options[:challenge_indicator] if options[:challenge_indicator]
           post[:'3ds'][:exemption] = options[:exemption] if options[:exemption]
+          post[:'3ds'][:allow_upgrade] = options[:allow_upgrade] if options[:allow_upgrade]
         end
 
         if options[:three_d_secure]
--- a/test/remote/gateways/remote_checkout_v2_test.rb
+++ b/test/remote/gateways/remote_checkout_v2_test.rb
@@ -557,6 +557,15 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
     assert auth.params['_links']['redirect']
   end
 
+  def test_allow_upgrade_with_direct_3ds
+    auth = @gateway.authorize(@amount, @threeds_card, @options.merge(execute_threed: true, allow_upgrade: true))
+
+    # using this assertion to show that the allow_upgrade field is accepted with the direct 3DS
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -241,6 +241,7 @@ def add_3ds(post, options)
           post[:'3ds'][:attempt_n3d] = options[:attempt_n3d] if options[:attempt_n3d]
           post[:'3ds'][:challenge_indicator] = options[:challenge_indicator] if options[:challenge_indicator]
           post[:'3ds'][:exemption] = options[:exemption] if options[:exemption]
+          post[:'3ds'][:allow_upgrade] = options[:allow_upgrade] if options[:allow_upgrade]
```
jcreiff, line 244: Looking at the [Checkout docs for this field](https://api-reference.checkout.com/previous/#operation/requestAPaymentOrPayout!path=0/3ds/allow_upgrade&t=request), it seems that the default value is `true` if nothing is specified in the request.

If that's the case, it seems like what we should be enabling here is the ability to send this field as `false`, but if options[:allow_upgrade] evaluates as `false`, this line won't be added to the request.

rachelkirk: So this field appears to be intended for use with the Gateway Specific/Direct 3DS only. The main thing I notice is on line 277 we have hard coded `post[:'3ds'][:enabled] = true` and in [their docs](https://api-reference.checkout.com/previous/#operation/requestAPaymentOrPayout) you can see the default is false and is set for `Whether to process this payment as a 3D Secure payment`. If it's always hard coded to true I'm not sure what passing in allow_upgrade is going to do. The transaction stays in a pending state so it's not possible to test the direct flow within a single remote test.

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -557,6 +557,15 @@ def test_direct_3ds_authorize
     assert auth.params['_links']['redirect']
   end
```
rachelkirk, line 559: I took `allow_upgrade` out of the 3ds2 options and wrote a dedicated remote test here as it seems it's intended for the gateway direct 3DS only. The field doesn't come back in the response so I can't add an assertion for anything other than the fact that the merchant is enrolled in 3DS, but happy to hear feedback on this test.

rachelkirk:
Determined this was a `won't do`

PR #4689, opened 2023-01-27
Moneris: Fix google pay (update apple pay)

sinourain:
Description
-------------------------
Truncate google pay and apple pay order id to 100 (max length of characters that google pay / apple pay accepts)

Unit test
-------------------------
Finished in 27.168945 seconds.
5439 tests, 77066 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote test
-------------------------
200.19 tests/s, 2836.55 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/moneris.rb
+++ b/lib/active_merchant/billing/gateways/moneris.rb
@@ -9,6 +9,8 @@ module ActiveMerchant #:nodoc:
     # Response Values", available at Moneris' {eSelect Plus Documentation
     # Centre}[https://www3.moneris.com/connect/en/documents/index.html].
     class MonerisGateway < Gateway
+      WALLETS = %w(APP GPP)
+
       self.test_url = 'https://esqa.moneris.com/gateway2/servlet/MpgRequest'
       self.live_url = 'https://www3.moneris.com/gateway2/servlet/MpgRequest'
 
@@ -47,7 +49,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_payment_source(post, creditcard_or_datakey, options)
         post[:amount] = amount(money)
-        post[:order_id] = options[:order_id]
+        post[:order_id] = format_order_id(post[:wallet_indicator], options[:order_id])
         post[:address] = options[:billing_address] || options[:address]
         post[:crypt_type] = options[:crypt_type] || @options[:crypt_type]
         add_external_mpi_fields(post, options)
@@ -71,7 +73,7 @@ module ActiveMerchant #:nodoc:
         post = {}
         add_payment_source(post, creditcard_or_datakey, options)
         post[:amount] = amount(money)
-        post[:order_id] = options[:order_id]
+        post[:order_id] = format_order_id(post[:wallet_indicator], options[:order_id])
         post[:address] = options[:billing_address] || options[:address]
         post[:crypt_type] = options[:crypt_type] || @options[:crypt_type]
         add_external_mpi_fields(post, options)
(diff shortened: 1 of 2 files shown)


test/remote/gateways/remote_moneris_test.rb
```diff
@@ -141,6 +141,20 @@ def test_successful_purchase_with_network_tokenization_google_pay_source
...
+      '4242424242424242',
+      payment_cryptogram: 'BwABB4JRdgAAAAAAiFF2AAAAAAA=',
+      verification_value: nil,
+      source: :google_pay
+    )
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal 'Approved', response.message
+    assert_false response.authorization.blank?
+  end
```
edgarv09, line 156: just a thing of format and background, for me I understood `without_order_id` that order_id is not present in options (that test already validated at line 131), the intent of this test is to check what happens if order_id has nil value, it is just what the description says about it.  It also could be excellent to add a test that validates the value of order_id remains exactly the same during the process due to that we have an order_id generation if the value is nil or not present.

test/remote/gateways/remote_moneris_test.rb
```diff
@@ -141,6 +141,20 @@ def test_successful_purchase_with_network_tokenization_google_pay_source
     assert_false response.authorization.blank?
   end
 
+  def test_successful_purchase_with_network_tokenization_google_pay_source_without_order_id
```
javierpedrozaing, line 144:
```suggestion
  def test_successful_purchase_with_network_tokenization_google_pay_source_with_nil_order_id
```

javierpedrozaing (commented):
LGTM

Heavyblade (commented):
@sinourain done with the review, leaved you a comment.

test/remote/gateways/remote_moneris_test.rb
sinourain: Sure, I made the changes and I added some extra tests. Thanks!

lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -438,6 +438,18 @@ def wallet_indicator(token_source)
         }[token_source]
       end
 
+      def format_order_id(wallet_indicator_code, order_id = nil)
+        # Truncate (max 100 characters) order id for
+        # google pay (a specific wallet / token source)
+        return truncate_google_pay_order_id(order_id) if wallet_indicator_code == 'GPP'
```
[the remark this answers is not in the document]
sinourain: Sure, I added apple pay and also some extra tests. Thanks!

lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -438,6 +438,19 @@ def wallet_indicator(token_source)
         }[token_source]
       end
 
+      def format_order_id(wallet_indicator_code, order_id = nil)
+        # Truncate (max 100 characters) order id for
+        # google pay and apple pay (specific wallets / token sources)
+        wallets = %w(APP GPP)
```
naashton, line 444: Can we take this out of the method and make it a class variable?

lib/active_merchant/billing/gateways/moneris.rb
```diff
@@ -438,6 +438,19 @@ def wallet_indicator(token_source)
...
+        return truncate_google_pay_order_id(order_id) if wallets.include?(wallet_indicator_code)
+
+        order_id
+      end
+
+      def truncate_google_pay_order_id(order_id = nil)
```
naashton, line 450: The method is references `google_pay` specifically, but it seems we need to truncate if  it's AP or GP. Should we change the name of the method to just be `truncate_order_id`?

naashton (changes requested):
Just a few nitpicks, otherwise this looks pretty straight forward.

lib/active_merchant/billing/gateways/moneris.rb
sinourain: good catch I will :)

lib/active_merchant/billing/gateways/moneris.rb
sinourain: Yes

naashton (approved):
Excellent work!

PR #4691, opened 2023-01-30
TrustCommerce: Update `authorization_from` to handle `store` response

jherreraa:
## Summary:

Fix Store/Unstore features

Remote tests
22 tests, 80 assertions, 3 failures, 0 errors, 0 pendings, 
0 omissions, 0 notifications 
86.3636% passed

Failing tests not related with changes

unit tests
20 tests, 64 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/trust_commerce.rb
+++ b/lib/active_merchant/billing/gateways/trust_commerce.rb
@@ -476,9 +476,14 @@ module ActiveMerchant #:nodoc:
       end
 
       def authorization_from(action, data)
-        authorization = data['transid']
-        authorization = "#{authorization}|#{action}" if authorization && VOIDABLE_ACTIONS.include?(action)
-        authorization
+        case action
+        when 'store'
+          data['billingid']
+        when *VOIDABLE_ACTIONS
+          "#{data['transid']}|#{action}"
+        else
+          data['transid']
+        end
       end
 
       def split_authorization(authorization)
--- a/test/remote/gateways/remote_trust_commerce_test.rb
+++ b/test/remote/gateways/remote_trust_commerce_test.rb
@@ -182,11 +182,28 @@ class TrustCommerceTest < Test::Unit::TestCase
     assert_bad_data_response(response)
   end
 
+  def test_successful_unstore
+    assert store = @gateway.store(@credit_card)
+    assert_equal 'approved', store.params['status']
+    assert response = @gateway.unstore(store.params['billingid'])
+    assert_success response
+  end
+
   def test_unstore_failure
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/trust_commerce.rb
```diff
@@ -476,7 +476,7 @@ def message_from(data)
       end
 
       def authorization_from(action, data)
-        authorization = data['transid']
+        action == 'store' ? authorization = data['billingid'] : authorization = data['transid']
```
sinourain, line 479: authorization can be defined only once:
`authorization = action == 'store' ? data['billingid'] : data['transid']`

sinourain (dismissed):
LGTM just a minor change

test/remote/gateways/remote_trust_commerce_test.rb
```diff
@@ -182,6 +182,13 @@ def test_failed_store
     assert_bad_data_response(response)
   end
 
+  def test_successful_unstore
+    assert store = @gateway.store(@credit_card)
+    assert_equal 'approved', store.params['status']
+    assert response = @gateway.unstore(store.params['billingid'])
+    assert_success response
+  end
+
```
Heavyblade, line 191: @jherreraa just to have a better remote tests and to ensure that store works as expected, please add a test that can use an stored payment method to do a purchase/authorize

lib/active_merchant/billing/gateways/trust_commerce.rb
Heavyblade: maybe is a good time for a refactor ?, something like:

```ruby
def authorization_from(action, data)
  case action
  when 'store'
    data['billingid']
  when *VOIDABLE_ACTIONS
    "#{data['transid']}|#{action}"
  else
    data['transid']
  end
end
```

Heavyblade (changes requested):
@jherreraa we are close, leave you a couple of comments.

test/remote/gateways/remote_trust_commerce_test.rb
jherreraa: Test added! Please check it again!

lib/active_merchant/billing/gateways/trust_commerce.rb
jherreraa: Suggestion taken!

jherreraa (commented):
solved! Thanks for the feedback

naashton (approved):
This change looks good! 👍

PR #4693, opened 2023-02-01
Credorax/finaro use eci on network tokens

Sainterman:
[Credorax] Use ECI on purchase, credit and authorization from options object

To enable network tokenization the token_eci field must be present so this PR adds conditions and default values
Currently the add_network_tokenization_card method is expecting the ECI field to come in the payment_method.
This change allows users to use network tokenization with default and custom values for the token_eci field.

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -141,7 +141,7 @@ module ActiveMerchant #:nodoc:
       def purchase(amount, payment_method, options = {})
         post = {}
         add_invoice(post, amount, options)
-        add_payment_method(post, payment_method)
+        add_payment_method(post, payment_method, options)
         add_customer_data(post, options)
         add_email(post, options)
         add_3d_secure(post, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -283,19 +283,22 @@ def add_invoice(post, money, options)
...
       end
 
-      def add_network_tokenization_card(post, payment_method)
+      def add_network_tokenization_card(post, payment_method, options)
         post[:b21] = NETWORK_TOKENIZATION_CARD_SOURCE[payment_method.source.to_s]
-        post[:token_eci] = payment_method&.eci if payment_method.source.to_s == 'network_token'
+        post[:token_eci] = '07'
+        post[:token_eci] = '00' if payment_method.brand.to_s == 'master'
+        post[:token_eci] =  options[:eci] if options[:eci]
+        post[:token_eci] =  options[:eci] || payment_method&.eci if payment_method.source.to_s == 'network_token'
```
Heavyblade, line 301: 💬 Comment:

I get the point of the code, but we need a simpler logic, something like:

```Ruby
 post[:token_eci] =  options[:eci] || payment_method&.eci || payment_method.brand.to_s == 'master' ? '00' : '07'
# Note: you are using the safe operator so don't need to check for payment method source to try to get the eci
```


```

Heavyblade (approved):
@Sainterman looks pretty much OK, leave you an small comment on an improvement for your consideration

curiousepic (approved):
Made a small formatting change for rubocop, but looks good, thanks. Will plan to merge sooon.

PR #4694, opened 2023-02-06
Payeezy: Handle nil and empty values for Apple Pay

naashton:
Payeezy support has indicated that passing empty or nil values in place of the `xid` and `cavv` may result in failed transactions for AMEX based AP tokens. They also informed us that the `eci_indicator` value should default to `5` instead of passing a nil value, following a similar pattern.

This PR ignores empty `payment_cryptogram` and defaults the `eci_indicator` to `5` if that value is `nil`.

Unit: 6 tests, 211 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote: 46 tests, 184 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payeezy.rb
+++ b/lib/active_merchant/billing/gateways/payeezy.rb
@@ -247,13 +247,13 @@ module ActiveMerchant
         nt_card[:card_number] = payment_method.number
         nt_card[:exp_date] = format_exp_date(payment_method.month, payment_method.year)
         nt_card[:cvv] = payment_method.verification_value
-        nt_card[:xid] = payment_method.payment_cryptogram
-        nt_card[:cavv] = payment_method.payment_cryptogram
+        nt_card[:xid] = payment_method.payment_cryptogram unless payment_method.payment_cryptogram.empty?
+        nt_card[:cavv] = payment_method.payment_cryptogram unless payment_method.payment_cryptogram.empty?
         nt_card[:wallet_provider_id] = 'APPLE_PAY'
 
         params['3DS'] = nt_card
         params[:method] = '3DS'
-        params[:eci_indicator] = payment_method.eci
+        params[:eci_indicator] = payment_method.eci.nil? ? '5' : payment_method.eci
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -247,13 +247,13 @@ def add_network_tokenization(params, payment_method, options)
         nt_card[:card_number] = payment_method.number
         nt_card[:exp_date] = format_exp_date(payment_method.month, payment_method.year)
         nt_card[:cvv] = payment_method.verification_value
-        nt_card[:xid] = payment_method.payment_cryptogram
-        nt_card[:cavv] = payment_method.payment_cryptogram
+        nt_card[:xid] = payment_method.payment_cryptogram unless payment_method.payment_cryptogram.empty?
+        nt_card[:cavv] = payment_method.payment_cryptogram unless payment_method.payment_cryptogram.empty?
```
aenand, line 251: Do these need to be set to `nil` or omitted from being sent if there is no `payment_cryptogram`?

chudel: Does this impact/benefit all customers of this gateway type (i.e.: is it an issue every customer faces presently) or just the one  customer (Priceline?)

naashton: We were explicitly told by Payeezy to NOT send `nil` values for these fields as they "could" lead to errors. These fields are `optional` according to Payeezy's documentation.

AFAIK, Priceline is the only user of AP on Payeezy, but this should not negatively impact any other customers of this gateway.

aenand (approved):

lib/active_merchant/billing/gateways/payeezy.rb
aenand: @chudel this would benefit all customers of this gateway type

PR #4696, opened 2023-02-10
DEPRECATED: CybersourceREST - Refund | Credit

sinourain:
IMPORTANT NOTE: I will create a new PR because I made this based in my personal folk repository

Description
-------------------------
This integration support the following payment operations:
- Refund
- Credit

Unit test
-------------------------
Finished in 29.776044 seconds.
5452 tests, 77126 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote test
-------------------------
183.10 tests/s, 2590.20 assertions/s

Rubocop
-------------------------

GWI-471

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -1,0 +1,221 @@
+require 'active_merchant/billing/gateways/cyber_source/cyber_source_common'
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class CyberSourceRestGateway < Gateway
+      include ActiveMerchant::Billing::CyberSourceCommon
+
+      self.test_url = 'https://apitest.cybersource.com'
+      self.live_url = 'https://api.cybersource.com'
+
+      self.supported_countries = ActiveMerchant::Billing::CyberSourceGateway.supported_countries
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover diners_club jcb maestro elo union_pay cartes_bancaires mada]
+
+      self.homepage_url = 'http://www.cybersource.com'
+      self.display_name = 'CyberSourceRest'
+
+      CREDIT_CARD_CODES = {
+        american_express: '003',
+        cartes_bancaires: '036',
+        dankort: '034',
+        diners_club: '005',
+        discover: '004',
+        elo: '054',
+        jcb: '007',
+        maestro: '042',
+        master: '002',
+        unionpay: '062',
+        visa: '001'
+      }
+
+      def initialize(options = {})
+        requires!(options, :merchant_id, :public_key, :private_key)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        authorize(money, payment, options) do |post|
+          post[:processingInformation] = { capture: true }
+        end
+      end
+
+      def authorize(money, payment, options = {})
+        post = build_auth_request(money, payment, options)
+
+        yield post if block_given?
+
+        commit('/pts/v2/payments/', post)
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(/(\\?"number\\?":\\?")\d+/, '\1[FILTERED]').
+          gsub(/(\\?"securityCode\\?":\\?")\d+/, '\1[FILTERED]').
+          gsub(/(signature=")[^"]*/, '\1[FILTERED]').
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -74,6 +90,22 @@ def build_auth_request(amount, payment, options)
         end.compact
       end
 
+      def build_refund_request(amount, options)
+        { clientReferenceInformation: {}, orderInformation: {} }.tap do |post|
```
javierpedrozaing, line 94: I like this approach  👍

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -48,6 +48,22 @@ def authorize(money, payment, options = {})
         commit('/pts/v2/payments/', post)
       end
 
+      def refund(money, options = {})
+        post = build_refund_request(money, options)
+
+        yield post if block_given?
```
javierpedrozaing, line 54: 👍

javierpedrozaing (commented):
Good job @sinourain 

💡  I can see that the` '/pts/v2/'` is repeated in each request, maybe we could remove it for each request and add it in the url function

Every commit call would be similar to
```suggestion
        commit('/credits', post)
```
and in the url function concat `/pts/v2/`
```suggestion
      def url(action)
        "#{(test? ? test_url : live_url)}/pts/v2/#{action}"
      end
```

also if you concider this update, you need to update this line in the `get_http_signature` function

```suggestion
       "(request-target)": "#{http_method} /pts/v2/#{resource}",
```

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -74,6 +90,22 @@ def build_auth_request(amount, payment, options)
...
+          add_amount(post, amount)
+        end.compact
+      end
+
+      def build_credit_request(amount, payment, options)
+        { clientReferenceInformation: {}, paymentInformation: {}, orderInformation: {} }.tap do |post|
+          add_code(post, options)
+          add_credit_card(post, payment)
+          add_amount(post, amount)
+          add_address(post, payment, options[:billing_address], options, :billTo)
```
Heavyblade, line 105: 💬 Comment:

Seems that `add_merchant_description` is missing here ... comparing the actual implementation against the old one

test/remote/gateways/remote_cyber_source_rest_test.rb
```diff
@@ -70,6 +70,27 @@ def test_successful_purchase
...
+
+  def test_successful_credit
+    response = @gateway.credit(@amount, @visa_card, @options)
+
+    assert_success response
+    assert response.test?
+    assert_equal 'PENDING', response.message
+    assert response.params['id'].present?
+    assert_nil response.params['_links']['capture']
+  end
```
Heavyblade, line 92: 💬 Comment:

Maybe is a good idea to test not just the happy path:

- maybe checking how to handle error handling in case of a double refund
- doing partial refunds and values above the initial transaction
- credits on invalid credit cards
- etc.

Heavyblade (commented):
@sinourain left some comments,

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -48,6 +48,22 @@ def authorize(money, payment, options = {})
...
+
+        yield post if block_given?
+
+        commit("/pts/v2/payments/#{options[:order_id]}/refunds", post)
+      end
+
+      def credit(money, payment, options = {})
+        post = build_credit_request(money, payment, options)
+
+        yield post if block_given?
```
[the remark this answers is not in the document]
sinourain: Yes, I was wrong copy / paste this approach and it is only necessary for the purchase/authorization methods

lib/active_merchant/billing/gateways/cyber_source_rest.rb
sinourain: Done!

test/remote/gateways/remote_cyber_source_rest_test.rb
sinourain: Done!

lib/active_merchant/billing/gateways/cyber_source_rest.rb
sinourain: Yes, I was wrong copy / paste this approach and it is only necessary for the purchase/authorization methods

PR #4698, opened 2023-02-14
Tns: update test URL

almalee24:
Update test URL to be only secure.uat.tnspayments.


aenand:
There are some unit tests failing in `tns_test.rb`. Other than that the change looks good

almalee24:
> There are some unit tests failing in `tns_test.rb`. Other than that the change looks good
Just pushed the change

PR #4699, opened 2023-02-16
TrustCommerce Verify feature added

jherreraa:
## Summary

Enable verify feature on TrustCommerce Gateway

Remote tests
21 tests, 74 assertions, 3 failures, 0 errors, 0 pendings, 
0 omissions, 0 notifications 85.7143% passed

Failing tests not related with changes

Unit tests
22 tests, 72 assertions, 0 failures, 0 errors, 0 pendings, 
0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/trust_commerce.rb
+++ b/lib/active_merchant/billing/gateways/trust_commerce.rb
@@ -249,8 +249,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def verify(credit_card, options = {})
-        add_creditcard(options, credit_card)
-        commit('verify', options)
+        parameters = {}
+        add_creditcard(parameters, credit_card)
+        commit('verify', parameters)
       end
 
       # recurring() a TrustCommerce account that is activated for Citadel, TrustCommerce's
--- a/test/remote/gateways/remote_trust_commerce_test.rb
+++ b/test/remote/gateways/remote_trust_commerce_test.rb
@@ -211,6 +211,13 @@ class TrustCommerceTest < Test::Unit::TestCase
     assert_success response
   end
 
+  def test_failed_verify_with_invalid_card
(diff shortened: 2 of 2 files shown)


lib/active_merchant/billing/gateways/trust_commerce.rb
```diff
@@ -250,6 +250,7 @@ def void(authorization, options = {})
 
       def verify(credit_card, options = {})
         add_creditcard(options, credit_card)
+        options[:order_id] = nil
```
naashton, line 253: This change doesn't seem like it'll do anything because nothing ever happens with the `options` hash after `add_creditcard` or in the `commit` method, unless I'm missing something.

naashton (changes requested):
it doesn't seem like this change is doing anything? What is this trying to accomplish?

lib/active_merchant/billing/gateways/trust_commerce.rb
jherreraa: According to the tests, the gateways rejects a field that is added by default in the library (order_id) so it needs to be removed for the verify in order to complete the verify requests successfully. If you need further clarification please let me know.

naashton: Oh I see here. So, it looks like you are using the `options` instead of the `parameters` (what would normally be `post`).

You should create a `parameters` hash and pass that to the `add_creditcard` method, instead of the `options` hash. That hash is going to be passed to the method from the user and AM takes the values from that hash and constructs the payload that we send to the gateway API.

It also doesn't look like `add_creditcard` adds any `order_id` to the payload, so you probably won't need to delete that value.

naashton (commented):
LGTM!

naashton (approved):

PR #4702, opened 2023-02-17
Ebanx: Add support of Elo & Hipercard

almalee24:
Unit:
19 tests, 84 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed  

Remote:
28 tests, 74 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.4286% passed


curiousepic:
Please add some context for the change in the commit message about the strategy here

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -278,7 +270,7 @@ def message_from(response)
 
       def authorization_from(action, parameters, response)
         if action == :store
-          "#{response.try(:[], 'token')}|#{CARD_BRAND[parameters[:payment_type_code].to_sym]}"
+          response.try(:[], 'token')
```
curiousepic, line 273: I think we should probably keep this just in case we need to change this again, otherwise we lose data for any cards stored in the meantime, and the extra info is not causing any problem

curiousepic: Only on successes is fine, that makes sense because authorization strings from failed store txns wouldn't get used on followups.

lib/active_merchant/billing/credit_card_methods.rb
```diff
@@ -241,6 +242,11 @@ module CreditCardMethods
         3528..3589, 3088..3094, 3096..3102, 3112..3120, 3158..3159, 3337..3349
       ]
 
+      HIPERCARD_RANGES = [
+        384100..384100, 384140..384140, 384160..384160, 606282..606282, 637095..637095,
+        637568..637568, 637599..637599, 637609..637609, 637612..637612
+      ]
+
```
aenand, line 249: Do we want to add `HIPERCARD` ranges since they are a subset of mastercard?

almalee24: The subset of mastercard are some of the ones that start with 6. The ones that don't fall within the `MASTERCARD_RANGES`.

aenand: Hmmmm I'm just thinking, does AM perform any check to see if the bin for the card is in a gateway's support card list?

almalee24: Core does based on AM, if the bin isn't found then no type will be assigned which makes the PM invalid

aenand: So do we think there's a scenario where core sends in a bin thinking it's mastercard, gets back hipercard, then it attempts to send to the gateway and it fails because the gateway doesn't support it?

almalee24: That wouldn't happen because of the order of events, MasterCard is on the top `CARD_COMPANY_DETECTORS`.

almalee24: 606282, 637095, 637568, 637599, 637609 are the ones that match the Mastercard bin ranges. And those bin ranges were only added 22 months to Mastercard, https://github.com/activemerchant/active_merchant/commit/db68b54d2942d9b9a26a3d9a65b2728794071a18

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -189,14 +181,14 @@ def add_invoice(post, money, options)
       end
 
       def add_card_or_token(post, payment, options)
-        payment, brand = payment.split('|') if payment.is_a?(String)
-        post[:payment][:payment_type_code] = payment.is_a?(String) ? brand : CARD_BRAND[payment.brand.to_sym]
+        payment = payment.split('|')[0] if payment.is_a?(String)
```
aenand, line 184: What happens to payment and why are we editing it?

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -189,14 +181,14 @@ def add_invoice(post, money, options)
...
+        post[:payment][:payment_type_code] = 'creditcard'
```
aenand, line 185: Do we not need to map to a cardbrand anymore?

PR #4709, opened 2023-02-24
Cyber source rest store unstore

Sainterman:


lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -141,30 +225,34 @@ def parse(body)
...
       end
 
-      def success_from(response)
-        response['status'] == 'AUTHORIZED'
+      def success_from(action, response)
+        case action
+        when /payments/
+          response['status'] == 'AUTHORIZED'
+        else
+          !response['id'].nil?
```
sinourain, line 250: why don't we use `response['id'].present?`

gasb150 (commented):
I left a few comments

curiousepic:

PR #4712, opened 2023-02-24
ChekoutV2:Add store/unstore

gasb150:
Summary:
In order to use Third Party Vaulting (TPV), this commit adds store and unstore methods. For apple pay/google pay the actual credentials could work, but for credit cards, and bank account the credentials required are secret_api_key and public_api_key.

This commits also, refactor the initialize method removing the rescue ArgumentError, and modifing the commit method to use ssl_request.

Remote test
-----------------------
Finished in 164.163913 seconds.
88 tests, 216 assertions, 0 failures, 0 errors,
100% passed

Unit test
-----------------------
Finished in 0.046893 seconds.
53 tests, 297 assertions, 0 failures, 0 errors,
100% passed

Rubocop
-----------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -19,12 +19,14 @@ module ActiveMerchant #:nodoc:
       def initialize(options = {})
         @options = options
         @access_token = nil
-        begin
+
+        if options.has_key?(:secret_key)
           requires!(options, :secret_key)
-        rescue ArgumentError
+        else
           requires!(options, :client_id, :client_secret)
           @access_token = setup_access_token
         end
+
         super
       end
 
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -280,11 +319,12 @@ def setup_access_token
         response['access_token']
       end
 
-      def commit(action, post, authorization = nil)
+      def commit(action, post, authorization = nil, method = :post)
         begin
-          raw_response = (action == :verify_payment ? ssl_get("#{base_url}/payments/#{post}", headers) : ssl_post(url(post, action, authorization), post.to_json, headers))
+          raw_response = ssl_request(method, url(action, authorization), post.nil? ? nil : post.to_json, headers(action))
```
naashton, line 324: I think you might want `post.empty?` here. If `post` is already `nil`, then this would already resolve to `nil`, but if `post` is an empty hash `{}` then `post.nil?` would resolve to false and still pass a value.

gasb150: Instead of using post.empty? (because if we use `nil.empty?` it will return an error), I choose to use both; if it's not nil, then check if it's not empty.

PR #4714, opened 2023-02-27
Remove raise ArgumentError on get requests

naashton:
The case for initiating a GET request raises a frivolous ArgumentError when the code already ignores any attempt to pass in a body to the request.

### DIFF
--- a/lib/active_merchant/connection.rb
+++ b/lib/active_merchant/connection.rb
@@ -85,8 +85,6 @@ module ActiveMerchant
           result =
             case method
             when :get
-              raise ArgumentError, 'GET requests do not support a request body' if body
-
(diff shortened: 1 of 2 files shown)


Heavyblade:
@naashton I was having the same thinking about that 👍 , the only thing that didn't feel right is changing that and something/someone relies on that behavior, it seems unlikely but it is better to not break someone else's code.

aenand:
We got an inbox requests about 2 requests for Checkoutv2 failing on the GET request for redirects in 3DS. Can you check in the core repo the checkoutV2 threeds extension to see what GET request that is?

aenand (approved):
LGTM 🚀

PR #4724, opened 2023-03-02
Orbital: dismiss CardSecValInd restriction

jherreraa:
GWI-567

Remote:
122 tests, 509 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
144 tests, 817 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop:

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -594,7 +594,7 @@ module ActiveMerchant #:nodoc:
         #   Null-fill this attribute OR
         #   Do not submit the attribute at all.
         # - http://download.chasepaymentech.com/docs/orbital/orbital_gateway_xml_specification.pdf
-        xml.tag! :CardSecValInd, '1' if %w(visa discover diners_club).include?(credit_card.brand) && bin == '000001'
+        xml.tag! :CardSecValInd, '1' if %w(visa discover diners_club).include?(credit_card.brand)
         xml.tag! :CardSecVal, credit_card.verification_value
       end
 


sinourain (approved):
LGTM

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -594,7 +594,7 @@ def add_verification_value(xml, credit_card)
         #   Null-fill this attribute OR
         #   Do not submit the attribute at all.
         # - http://download.chasepaymentech.com/docs/orbital/orbital_gateway_xml_specification.pdf
-        xml.tag! :CardSecValInd, '1' if %w(visa discover diners_club).include?(credit_card.brand) && bin == '000001'
+        xml.tag! :CardSecValInd, '1' if %w(visa discover diners_club).include?(credit_card.brand)
```
Heavyblade, line 597: 💬 Comment:

While checking the docs the only thing that seems odd, is that API docs version 9 says that CardSecValInd should be included when transacting with `visa`, `discover`, or `jbc` but here we have dinners_club.

Orbital_Gateway_XML_Interface_Developer_Guide



Should we update this section ?

Heavyblade (approved):
@jherreraa looks pretty much OK, but left a comment for you or maybe @naashton  consideration.

lib/active_merchant/billing/gateways/orbital.rb
naashton: This is a good call out. I think for this change, we'll leave it as is and raise this internally.

naashton (approved):
LGTM!

PR #4725, opened 2023-03-02
Ebanx: Add transaction inquire request

almalee24:
Get transaction by authorization.

Remote:
27 tests, 73 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.2963% passed

Unit:
20 tests, 89 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -27,7 +27,8 @@ module ActiveMerchant #:nodoc:
         capture: 'capture',
         refund: 'refund',
         void: 'cancel',
-        store: 'token'
+        store: 'token',
+        inquire: 'query'
       }
 
       HTTP_METHOD = {
@@ -36,7 +37,8 @@ module ActiveMerchant #:nodoc:
         capture: :get,
         refund: :post,
         void: :get,
-        store: :post
+        store: :post,
+        inquire: :get
       }
 
       VERIFY_AMOUNT_PER_COUNTRY = {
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -126,6 +128,14 @@ def verify(credit_card, options = {})
         end
       end
 
+      def inquire(authorization, options = {})
```
BritneyS, line 131: I think `status` would be a better name for this method, since `inquire` is a term used very specifically upstream.

aenand: Does this method fetch the status of a transaction? If so then I think keeping it `inquire` would be right

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -257,13 +267,15 @@ def add_processing_type_to_commit_headers(commit_headers, processing_type)
       end
 
       def success_from(action, response)
+        payment_status = response.try(:[], 'payment').try(:[], 'status')
```
aenand, line 270: Nice refactor!

lib/active_merchant/billing/gateways/ebanx.rb
almalee24: I don't know, now that I think about it I think inquire is the way to go because it's basically doing the exact same thing as Transaction Sync. We're using the authorization to inquire about the payment status on Ebanx.

almalee24: @aenand yes it does

aenand: Yeah I'd agree that we should use `inquire` for any method that takes in a transaction reference, fetches the status from the gateway, and returns it

aenand (approved):
LGTM :shipit:

PR #4731, opened 2023-03-09
CyberSourceRest: Add store/unstore

gasb150:
Summary:
In order to perform store and unstore transactions, this commit add store method.

Remote Test:
Finished in 27.754991 seconds.
32 tests, 101 assertions, 0 failures, 0 errors, 0 pendings,0 omissions, 0 notifications 100% passed

Unit Tests:
Finished in 0.04074 seconds.
18 tests, 104 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -48,21 +48,19 @@ module ActiveMerchant #:nodoc:
       def authorize(money, payment, options = {}, capture = false)
         post = build_auth_request(money, payment, options)
         post[:processingInformation][:capture] = true if capture
-
-        commit('payments', post, options)
+        commit('payments', post, :post, options)
       end
 
       def capture(money, authorization, options = {})
         payment = authorization.split('|').first
         post = build_reference_request(money, options)
-
-        commit("payments/#{payment}/captures", post, options)
+        commit(endpoint_url('capture', payment), post, :post, options)
       end
 
       def refund(money, authorization, options = {})
         payment = authorization.split('|').first
         post = build_reference_request(money, options)
-        commit("payments/#{payment}/refunds", post, options)
+        commit(endpoint_url('refund', payment), post, :post, options)
       end
 
       def credit(money, payment, options = {})
@@ -73,7 +71,7 @@ module ActiveMerchant #:nodoc:
       def void(authorization, options = {})
         payment, amount = authorization.split('|')
         post = build_void_request(amount)
-        commit("payments/#{payment}/reversals", post)
+        commit(endpoint_url('void', payment), post, :post, options)
       end
 
       def verify(credit_card, options = {})
@@ -100,8 +98,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def unstore(stored_token)
-        customer_id, payment_instrument_id = stored_token.split('|')
-        commit("customers/#{customer_id}/payment-instruments/#{payment_instrument_id}/", nil, :delete)
+        payment = stored_token.split('|').drop(2)
+        commit(endpoint_url('unstore', payment), nil, :delete)
       end
 
       def supports_scrubbing?
@@ -138,7 +136,6 @@ module ActiveMerchant #:nodoc:
           add_business_rules_data(post, payment, options)
           add_partner_solution_id(post)
           add_stored_credentials(post, payment, options)
-          add_order_id(post, options)
         end.compact
       end
 
@@ -147,7 +144,7 @@ module ActiveMerchant #:nodoc:
           add_order_id(post, options)
           add_code(post, options)
           add_mdd_fields(post, options)
-          add_amount(post, amount)
+          add_amount(post, amount, options)
           add_partner_solution_id(post)
         end.compact
       end
@@ -163,6 +160,18 @@ module ActiveMerchant #:nodoc:
         end.compact
       end
 
+      def add_code(post, options)
+        return unless options[:order_id].present?
+
+        post[:clientReferenceInformation][:code] = options[:order_id]
+      end
+
+      def add_customer_id(post, options)
+        return unless options[:customer_id].present?
+
+        post[:paymentInformation][:customer] = { customerId: options[:customer_id] }
+      end
+
       def create_customer_request(payment, options)
         customer = { buyerInformation: {}, clientReferenceInformation: {}, merchantDefinedInformation: [] }.tap do |post|
           post[:buyerInformation][:merchantCustomerId] = options[:merchant_customer_id] if options[:merchant_customer_id]
@@ -173,25 +182,43 @@ module ActiveMerchant #:nodoc:
       end
 
       def create_instrument_identifier_request(payment, options)
-        instrument_identifier = {
-          card: {
-            number: payment.number
+        if payment.is_a?(Check)
+          instrument_identifier = {
+            bankAccount: {
+              number: payment.number,
+              routingNumber: payment.routing_number
+            }
           }
-        }
+        else
+          instrument_identifier = {
+            card: {
+              number: payment.number
+            }
+          }
+        end
         commit('instrumentidentifiers', instrument_identifier)
       end
 
       def create_customer_payment_instrument_request(payment, options, customer_id, instrument_identifier)
         post = {}
-        post[:default] = 'true'
-        post[:card] = {}
-        post[:card][:type] = CREDIT_CARD_CODES[payment.brand.to_sym]
-        post[:card][:expirationMonth] = payment.month.to_s
-        post[:card][:expirationYear] = payment.year.to_s
+        if payment.is_a?(Check)
+          add_bank_account_information(payment, post)
+        else
+          post[:default] = 'true'
+          post[:card] = {}
+          post[:card][:type] = CREDIT_CARD_CODES[payment.brand.to_sym]
+          post[:card][:expirationMonth] = payment.month.to_s
+          post[:card][:expirationYear] = payment.year.to_s
+        end
         add_address(post, payment, options[:billing_address], options, :billTo, nil)
         post[:instrumentIdentifier] = {}
         post[:instrumentIdentifier][:id] = instrument_identifier.params['id']
-        commit("customers/#{customer_id}/payment-instruments", post)
+        commit(endpoint_url('customer-payment', customer_id), post, :post, options)
+      end
+
+      def add_bank_account_information(payment, post)
+        post[:bankAccount] = {}
+        post[:bankAccount][:type] = 'savings'
       end
 
       def add_order_id(post, options)
@@ -240,7 +267,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_payment(post, payment, options)
-        customer_id, payment_instrument_id = payment.split('|')
+        customer_id, payment_instrument_id = payment.split('|').drop(2)
         post[:paymentInstrument] = {
           id: payment_instrument_id
         }
@@ -405,7 +432,6 @@ module ActiveMerchant #:nodoc:
         response = parse(ssl_request(http_method, url(action), post.nil? || post.empty? ? nil : post.to_json, auth_headers(action, post, http_method)))
         succeeded = success_from(action, response, http_method)
         body = action == :delete ? { response_code: response.to_s } : response
-
         Response.new(
           succeeded,
           message_from(body, succeeded, http_method),
@@ -448,11 +474,8 @@ module ActiveMerchant #:nodoc:
         if /payment-instruments/.match(action) && !response.empty?
           customer_id = response['_links']['customer']['href'].split('/').last
           payment_instrument_id = response['id']
-          return [customer_id, payment_instrument_id].join('|')
         end
-        return id if amount.blank?
-
-        [id, amount].join('|')
+        [id, amount, customer_id, payment_instrument_id].join('|')
       end
 
       def error_code_from(response, succeeded)
@@ -540,6 +563,23 @@ module ActiveMerchant #:nodoc:
         return unless application_id
 
         post[:clientReferenceInformation][:partner] = { solutionId: application_id }
+      end
+
+      def endpoint_url(action, param = '')
+        case action
+        when 'capture'
+          "payments/#{param}/captures"
+        when 'refund'
+          "payments/#{param}/refunds"
+        when 'void'
+          "payments/#{param}/reversals"
+        when 'unstore'
+          "customers/#{param[0]}/payment-instruments/#{param[1]}/"
+        when 'customer-payment'
+          "customers/#{param}/payment-instruments"
+        else
+          ''
+        end
       end
     end
   end
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -399,10 +399,14 @@ module ActiveMerchant #:nodoc:
         JSON.parse(body)
       end
 
-      def commit(action, post, http_method = :post)
+      def commit(action, post, http_method = :post, options = {})
+        add_reconciliation_id(post, options)
+        add_sec_code(post, options)
+        add_invoice_number(post, options)
         response = parse(ssl_request(http_method, url(action), post.nil? || post.empty? ? nil : post.to_json, auth_headers(action, post, http_method)))
         succeeded = success_from(action, response, http_method)
         body = action == :delete ? { response_code: response.to_s } : response
+
         Response.new(
           succeeded,
           message_from(body, succeeded, http_method),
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -127,6 +141,51 @@ def build_credit_request(amount, payment, options)
...
+        add_address(post, payment, options[:billing_address], options, :billTo, nil)
+        commit('paymentinstruments', post)
+      end
+
+      def create_customer_payment_instrument(payment, options, customer_id, instrument_identifier)
+        post = {}
+        post[:deafult] = 'true'
+        post[:card] = {}
+        post[:card][:type] = CREDIT_CARD_CODES[payment.brand.to_sym]
+        post[:card][:expirationMonth] = payment.month.to_s
```
Heavyblade, line 181: ❓ Question:

Is not sensitive data, but is a must to send the data even when you already sent it on the payment_instrument call ?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -245,68 +320,103 @@ def add_merchant_description(post, options)
       end
 
       def url(action)
-        "#{(test? ? test_url : live_url)}/pts/v2/#{action}"
+        case action
+        when /customers/
+          "#{(test? ? test_url : live_url)}/tms/v2/#{action}"
+        when 'instrumentidentifiers', 'paymentinstruments'
+          "#{(test? ? test_url : live_url)}/tms/v1/#{action}"
```
Heavyblade, line 327: ❓ Question:

Is this a typo or you actually want to use the `v1` API version here ?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -245,68 +320,103 @@ def add_merchant_description(post, options)
...
         )
       rescue ActiveMerchant::ResponseError => e
         response = e.response.body.present? ? parse(e.response.body) : { 'response' => { 'rmsg' => e.response.msg } }
         Response.new(false, response.dig('response', 'rmsg'), response, test: test?)
       end
 
-      def success_from(response)
-        %w(AUTHORIZED PENDING REVERSED).include?(response['status'])
+      def success_from(action, response, http_method)
+        return response.empty? if http_method == :delete
```
Heavyblade, line 365: ❓ Question:

Are you getting just 200 response without body here ?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -245,68 +320,103 @@ def add_merchant_description(post, options)
...
+          target = "/tms/v1/#{resource}"
+        else
+          target = "/pts/v2/#{resource}"
+        end
+
+        hash_to_sign = {
           host: host,
           date: gmtdatetime,
-          "(request-target)": "#{http_method} /pts/v2/#{resource}",
+          "(request-target)": "#{http_method} #{target}",
```
Heavyblade, line 414: 💬 Comment:

If we need to build this dynamically is better to rely on the actual URL that will be used and extract the path from there, so we prevent some hard-to-debug issues in the future with the signature.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -225,14 +295,19 @@ def add_address(post, payment_method, address, options, address_type)
           postalCode:            address[:zip],
           country:               lookup_country_code(address[:country])&.value,
           email:                 options[:email].presence || 'null@cybersource.com',
-          phoneNumber:           address[:phone]
+          phoneNumber:           address[:phone],
           # merchantTaxID:         ship_to ? options[:merchant_tax_id] : nil,
-          # company:               address[:company],
+          company:               address[:company]
           # companyTaxID:          address[:companyTaxID],
```
jherreraa, line 301: ❓ Is this set of comments required for a further development?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -324,9 +434,9 @@ def sign_payload(payload)
       def auth_headers(action, post, http_method = 'post')
         digest = "SHA-256=#{Digest::SHA256.base64digest(post.to_json)}" if post.present?
         date = Time.now.httpdate
-
+        accept = /payments/.match?(action) || /credits/.match?(action) ? 'application/hal+json;charset=utf-8' : 'application/json;charset=utf-8'
         {
-          'Accept' => 'application/hal+json;charset=utf-8',
+          'Accept' => accept,
```
Heavyblade, line 439: 💬 Comment:

We should revisit this at the end, seem very improbable that an "Accept" header has an impact on the actual request.

Heavyblade (commented):
@gasb150 a lot of hard work from your side here, left some comments for you

jherreraa (commented):
Is mostly ok for me... just a few questions...

lib/active_merchant/billing/gateways/cyber_source_rest.rb
gasb150: Yes, response without a body.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
gasb150: It's weird, but modifying the "accept" affects the sandbox response for these "store" calls.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -77,6 +77,20 @@ def verify(credit_card, options = {})
...
+      def store(payment, options = {})
+        MultiResponse.run do |r|
+          customer_id = options[:customer_id].nil? ? r.process { commit('customers', create_customer(payment, options)) }.params['id'] : options[:customer_id]
+          instrument_identifier = r.process { create_instrument_identifier(payment, options) }
+          r.process { create_payment_instrument(payment, instrument_identifier, options) }
+          r.process { create_customer_payment_instrument(payment, options, customer_id, instrument_identifier) }
+        end
+      end
+
+      def unstore(third_party_token)
```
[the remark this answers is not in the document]
gasb150: No, it is unnecessary; only the payment_instrument associated with the customer will be deleted.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
gasb150: I'm not sure about it; I just removed the company committed because I needed it.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
gasb150: I think I can remove the create_payment_instrument method, considering the response from this method is not used in the subsequent.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -127,6 +141,51 @@ def build_credit_request(amount, payment, options)
         end.compact
       end
 
+      def create_customer(payment, options)
```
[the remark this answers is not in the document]
gasb150: I agree with you.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -127,6 +141,51 @@ def build_credit_request(amount, payment, options)
...
+        { buyerInformation: {}, clientReferenceInformation: {}, merchantDefinedInformation: [] }.tap do |post|
+          post[:buyerInformation][:merchantCustomerId] = options[:customer_id]
```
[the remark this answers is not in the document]
gasb150: I will rename it; the customer_id and merchant_customer_id seem to be different values.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -127,6 +141,51 @@ def build_credit_request(amount, payment, options)
...
+            id: instrument_identifier.params['id']
+          }
+        }
+        add_address(post, payment, options[:billing_address], options, :billTo, nil)
+        commit('paymentinstruments', post)
+      end
+
+      def create_customer_payment_instrument(payment, options, customer_id, instrument_identifier)
+        post = {}
+        post[:deafult] = 'true'
```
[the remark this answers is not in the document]
gasb150: yes, it is 😅

lib/active_merchant/billing/gateways/cyber_source_rest.rb
gasb150: Yes

lib/active_merchant/billing/gateways/cyber_source_rest.rb
gasb150: Maybe use 'Accept' => 'application/hal+json, application/json;charset=utf-8', which accepts both media types.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -77,6 +77,26 @@ def verify(credit_card, options = {})
...
+            customer_id = options[:customer_id]
+          end
+          instrument_identifier = r.process { create_instrument_identifier_request(payment, options) }
+          break instrument_identifier unless instrument_identifier.success?
+
+          r.process { create_customer_payment_instrument_request(payment, options, customer_id, instrument_identifier) }
+        end
+      end
+
+      def unstore(third_party_token)
```
curiousepic, line 95:
```suggestion
      def unstore(stored_token)
```
TPT is specific terminology

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -47,7 +47,7 @@ def purchase(money, payment, options = {})
 
       def authorize(money, payment, options = {}, capture = false)
         post = build_auth_request(money, payment, options)
-        post[:processingInformation][:capture] = true if capture
+        post[:processingInformation][:capture] = true if capture && !options[:third_party_token]
```
curiousepic, line 50: We usually designate this by checking if the payment method argument is a string
```suggestion
        post[:processingInformation][:capture] = true if capture && !payment.is_a?(String)
```

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -127,16 +146,40 @@ def build_credit_request(amount, payment, options)
...
+        post[:card][:expirationMonth] = payment.month.to_s
+        post[:card][:expirationYear] = payment.year.to_s
+        add_address(post, payment, options[:billing_address], options, :billTo, nil)
+        post[:instrumentIdentifier] = {}
+        post[:instrumentIdentifier][:id] = instrument_identifier.params['id']
+        commit("customers/#{customer_id}/payment-instruments", post)
+      end
 
-        post[:paymentInformation][:customer] = { customerId: options[:customer_id] }
+      def add_code(post, options)
```
curiousepic, line 179: Would suggest this, since "add_code" sounds like it could mean anything 😄 
```suggestion
      def add_order_id(post, options)
```

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -47,7 +47,7 @@ def purchase(money, payment, options = {})
 
       def authorize(money, payment, options = {}, capture = false)
         post = build_auth_request(money, payment, options)
-        post[:processingInformation][:capture] = true if capture
+        post[:processingInformation][:capture] = true if capture && !payment.is_a?(String)
```
curiousepic, line 50: hm, it seems like the purchase remote tests are failing due to this not being added - my suggested change may have been incorrect, but I'm not sure why it would be.

curiousepic: oops - it should be `payment.is_a?(String)` 😆 sorry!

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -77,6 +77,26 @@ def verify(credit_card, options = {})
         end
       end
 
+      def store(payment, options = {})
+        MultiResponse.run do |r|
+          if options[:customer_id].nil?
+            customer_id = r.process { commit('customers', create_customer_request(payment, options)) }.params['id']
+            break r unless r.success?
```
curiousepic, line 84: I think `break` is usually used to escape a block without returning a value, but we're returning something here, so it should probably be `return`?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -245,68 +305,95 @@ def add_merchant_description(post, options)
...
-      def authorization_from(response)
+      def authorization_from(response, action)
         id = response['id']
         has_amount = response['orderInformation'] && response['orderInformation']['amountDetails'] && response['orderInformation']['amountDetails']['authorizedAmount']
         amount = response['orderInformation']['amountDetails']['authorizedAmount'].delete('.') if has_amount
-
+        if /payment-instruments/.match(action) && !response.empty?
+          customer_id = response['_links']['customer']['href'].split('/').last
+          payment_instrument_id = response['id']
+          return [customer_id, payment_instrument_id].join('|')
```
curiousepic, line 372: I think we should keep the ordering of elements in an authorization string as consistent as we can. To that end, there should only be one final concatenation, which, if there are any elements missing, are just blank. For example, 

[response_id, amount, customer_id, payment_instrument_id].join('|')

This lets us know for sure which element means what, for any given authorization string.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -77,6 +77,26 @@ def verify(credit_card, options = {})
...
+          instrument_identifier = r.process { create_instrument_identifier_request(payment, options) }
+          return instrument_identifier unless instrument_identifier.success?
+
+          r.process { create_customer_payment_instrument_request(payment, options, customer_id, instrument_identifier) }
+        end
+      end
+
+      def unstore(stored_token)
+        customer_id, payment_instrument_id = stored_token.split('|')
+        commit("customers/#{customer_id}/payment-instruments/#{payment_instrument_id}/", nil, :delete)
```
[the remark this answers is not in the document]
curiousepic: Then the `success_from` can also use a single `case` list as well.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
curiousepic: I think we should also record the `instrument_id` from the second call, if it is available.

test/unit/gateways/cyber_source_rest_test.rb
```diff
@@ -233,6 +223,46 @@ def test_url_building
...
+        assert_equal request['instrumentIdentifier']['id'], '7010000000016241111'
+        assert_includes request, 'billTo'
+      when /customers/
+        assert_equal request['buyerInformation']['email'], 'test@cybs.com'
+        assert_includes request['buyerInformation'], 'merchantCustomerId'
+        assert_includes request, 'clientReferenceInformation'
+        assert_includes request, 'merchantDefinedInformation'
+      end
+    end.respond_with(successful_create_customer, successful_create_instrument_identifiers, successful_store_response)
+  end
```
curiousepic, line 253: A couple tests I think we should try:

- If the instrument creation call fails, does the authroization string still contain a customer_id from the first step? Otherwise we have an "orphan" customer.
- Similar, for instrument identifier, if the last call fails - but we're currently not returning the instrument id in the authorization string, so we should add that if possible.

curiousepic (changes requested):
I think we're using the correct method, but I think we can definitely clean up some of the patterns, and we definitely want to standardize the authorization_from string instead of returning different formats for different actions.

curiousepic:
@gasb150 Could you clean up this branch? a bunch of unrelated commits seemed to get bundled with it

lib/active_merchant/billing/gateways/cyber_source_rest.rb
jherreraa: Hello @curiousepic, this issue was changed, please take a look of the new version (amend on this same PR) and let me know if the change complies with the suggestion.

test/unit/gateways/cyber_source_rest_test.rb
jherreraa: the case with the payment instrument creation failure was added, the final case was tried but not able to inject failure due to previous calls guarantee the success of this statement.

curiousepic:
@gasb150 Sorry for such a hassle with this but a rebase is still showing conflicts - you can see in the GH interface below.

After consulting with someone who knows Git better than me, it looks like you previously merged master to the branch instead of rebasing, which copied a bunch of prior AM commits to new commits with different hashes, and then did some kind of cherry pick to remove them, but then there are still conflicts relevant to the gateway.

Their suggestion to get a clean slate on this branch is:
```
(note the commit hashes of the commits relevant to this change)
git reset --hard HEAD~50
git checkout master; git pull
git checkout -
git rebase master
git cherry-pick <relevant_commit_hash>
(repeat cherry picks as necessary)
```

curiousepic:

PR #4736, opened 2023-03-13
Adyen: add support for shopper_statement field for capture action

yunnydang:
shopper_statement is already a supported gateway field however Adyen is claiming that it works fine for auth, but not for capture. This is to fix/add to the capture action requests so that shopperStatement is nested within addtionalData and set properly.

Local:
5477 tests, 77201 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6531% passed

Unit:
105 tests, 531 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
132 tests, 443 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
90.9091% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -82,6 +82,7 @@
 * Ebanx: Add transaction inquire request [almalee24] #4725
 * Ebanx: Add support for Elo & Hipercard [almalee24] #4702
 * CheckoutV2: Add Idempotency key support [yunnydang] #4728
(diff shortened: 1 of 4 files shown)


rachelkirk (approved):
Tests pass and looks good to me!

PR #4743, opened 2023-03-24
Payeezy: Add `last_name` for `add_network_tokenization`

naashton:
This change updates the `add_network_tokenization` method to include the `last_name` in the `cardholder_name` value when getting the name from a payment method

Unit: 49 tests, 227 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote: 50 tests, 201 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/payeezy.rb
+++ b/lib/active_merchant/billing/gateways/payeezy.rb
@@ -219,7 +219,7 @@ module ActiveMerchant
         tele_check[:check_type] = 'P'
         tele_check[:routing_number] = echeck.routing_number
         tele_check[:account_number] = echeck.account_number
-        tele_check[:accountholder_name] = "#{echeck.first_name} #{echeck.last_name}"
+        tele_check[:accountholder_name] = name_from_payment_method(echeck)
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/gateways/payeezy.rb
```diff
@@ -287,6 +287,12 @@ def name_from_address(options)
         return address[:name] if address[:name]
       end
 
+      def name_from_payment_method(payment_method)
+        return unless payment_method.first_name && payment_method.last_name
```
aenand, line 291: Do payment methods have a `full_name` attribute in AM or is it just in core?

curiousepic: We have `name` which should work instead of needing this method; I'll take this from here.

curiousepic: Ah I see now why this is being done; `.name` will return an empty string, which fouls up the conditional. I think this is fine.

curiousepic (approved):
👍

PR #4747, opened 2023-04-03
Ebanx: Remove default email

aenand:
ECS-2829

Ebanx requires that merchants pass in an email in
order to complete a transaction. Previously, ActiveMerchant was sending in a default email if one was not provided which causes Ebanx's fraud detection to mark these transactions as failed for fraud. This incorrect failure message has led to merchant frustration since they could not quickly know the root of the problem

Test Summary
Remote:
32 tests, 88 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.875% passed


curiousepic (approved):
👍

PR #4748, opened 2023-04-03
Stripe PI: tokenize PM for verify

aenand:
ECS-2867

For the Stripe PI gateway, when a merchant runs a verify they want to get back the resulting card information that Stripe provides. In off_session cases we are not tokenizing the card at Stripe which prevents us from getting back valuable card details.

This commit updates the logic to always get the card details back from Stripe on verify transactions.

This commit also improves the resiliency of Stripe PI remote tests by dynamically creating a customer object before running the remote tests so that they do not error out with a too many payment methods for customer message.

Test Summary
Remote:
80 tests, 380 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -3,7 +3,8 @@
 class RemoteStripeIntentsTest < Test::Unit::TestCase
   def setup
     @gateway = StripePaymentIntentsGateway.new(fixtures(:stripe))
-    @customer = fixtures(:stripe_verified_bank_account)[:customer_id]
+    # @customer = fixtures(:stripe_verified_bank_account)[:customer_id]
```
curiousepic, line 6: Let's remove this if you don't anticipate needing it again

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -56,6 +56,11 @@ def show_intent(intent_id, options)
         commit(:get, "payment_intents/#{intent_id}", nil, options)
       end
 
+      def create_test_customer
```
curiousepic, line 59: nice addition

curiousepic (approved):
👍

PR #4749, opened 2023-04-05
Bogus: Add verify functionality

willemk:
## Summary

Add support for the `verify` functionality to the Bogus Gateway. 

While the [gateway template](https://github.com/activemerchant/active_merchant/blob/master/generators/gateway/templates/gateway.rb#L53) traditionally defines `verify` as an `authorize` + `void`, for the sake of the bogus gateway where responses are faked, this has been simplified to just an `authorize` since the `void` call is traditionally ignored.

### DIFF
--- a/lib/active_merchant/billing/gateways/bogus.rb
+++ b/lib/active_merchant/billing/gateways/bogus.rb
@@ -90,6 +90,10 @@ module ActiveMerchant #:nodoc:
         end
       end
 
+      def verify(credit_card, options = {})
+        authorize(0, credit_card, options)
+      end
+
       def store(paysource, options = {})
         case normalize(paysource)
         when /1$/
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/bogus.rb
```diff
@@ -90,6 +90,10 @@ def void(reference, options = {})
         end
       end
 
+      def verify(credit_card, options = {})
+        authorize(100, credit_card, options)
+      end
```
aenand, line 95: A `verify` transaction should either be a zero amount authorization or a `MultiResponse` with a low amount followed by a `void`

willemk: Updated.

aenand:
This looks good! Please add a changelog entry and squash the commits into one. Thank you!

aenand:
Thank you!

PR #4751, opened 2023-04-06
Borgun: Update TrCurrencyExponent

aenand:
ECS-2861

A merchant using the Borgun gateway reported that when a user was completing the challenge, the gateway was displaying a value 100 times greater than what was requested. This casused the ccardholders to stop the 3DS flow and abandon the cart.

After reaching out to the Borgun gateway they explained that the ISK currency on Borgun is sometimes a 0 decimal currency and sometimes a 2 decimal currency. The explanation given via email is that we must provide the TrCurrencyExponent of 2 when utilizing the 3DS flow but not on the finish sale portion.

Remote:
22 tests, 47 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 90.9091% passed
Note: these 2 tests fail on master


curiousepic (approved):
💡 👍

PR #4756, opened 2023-04-17
Close stale issues/PRs

aenand:
The ActiveMerchant repository often has old issues and PRs that have gone stale. This makes it hard
to keep track of new requests. This commit adds
the `stale` GHA in debug only mode to have a dry run of auto-closing issues and PRs.

Issues will be able to live for 60 days without activity and PRs fro 21 days (excluding draft PRs).

Docs: https://github.com/marketplace/actions/close-stale-issues

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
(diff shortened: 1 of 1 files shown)


rachelkirk (commented):
Inviting myself to leave a comment: 

We still have the GSS (gateway self service) program (hopefully not for much longer) and it's not uncommon for PR's to sit without activity for longer than 21 days depending on how long it takes for them to respond with changes we've asked for. I don't know what the magic number is and I think this is a good thing you're doing but just wanted to throw that out there for some context.

PR #4758, opened 2023-04-18
POC: Ebanx V2

aenand:
This commit creates a mixin for EBANX V2 so that
the gateway can begin to experiment with altering
the gateway's integration. By having a mixin, consumers of AM can continue to extend the base class and have a bit of visibility into the changes that EBANX is trying out.

Once a change has "graduated", it should be moved from the mixin to the base class


aenand:
Not needed

PR #4760, opened 2023-04-24
Improve error handling: OAuth

aenand:
ECS-2845

OAuth has become a standard authentication
mechanism for many gateways in recent years
however AM has not been updated to support error
handling when a merchant passes incorrect details
to the gateway.

In non OAuth flows we would return a failed response. This commit now raises a new exception type indicating that the request failed at the OAuth stage rather than the transaction stage of a request

Remote:
25 tests, 57 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 92% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4.rb
+++ b/lib/active_merchant/billing/gateways/shift4.rb
@@ -153,7 +153,7 @@ module ActiveMerchant #:nodoc:
         add_datetime(post, options)
 
         response = commit('accesstoken', post, request_headers('accesstoken', options))
-        raise ArgumentError, response.params.fetch('result', [{}]).first.dig('error', 'longText') unless response.success?
+        raise OAuthResponseError.new(response, response.params.fetch('result', [{}]).first.dig('error', 'longText')) unless response.success?
 
         response.params['result'].first['credential']['accessToken']
       end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
 == HEAD
 * Adyen: Update selectedBrand mapping for Google Pay [jcreiff] #4763
 * Shift4: Add vendorReference field [jcreiff] #4762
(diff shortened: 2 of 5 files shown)


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -153,7 +153,7 @@ def setup_access_token
         add_datetime(post, options)
 
         response = commit('accesstoken', post, request_headers('accesstoken', options))
-        raise ArgumentError, response.params.fetch('result', [{}]).first.dig('error', 'longText') unless response.success?
+        raise OAuthResponseError.new(response, response.params.fetch('result', [{}]).first.dig('error', 'longText')) unless response.success?
```
BritneyS, line 156: Are there other failures possible? I'm concerned that this change would register all non-success statuses as an OAuth error when that might not apply 🤔 I would have had the same concern with the previous iteration.

aenand: Good question! I see three possible paths in `setup_access_token`. 

1. Success
2. Failure due to invalid credentials
3. Failure due to the gateway being down

🤔 Right now both failures get reported as an OAuthResponseError but maybe I should explore trying to differentiate between the two? Can you think of any other paths that I may be missing?

BritneyS: There could be a malformed request -- that would be caused by us, but hopefully caught in testing before merging. So I think starting out with differentiating between those two would be already accounting for the most probable errors 👍🏾

BritneyS: Oh good, as long as the other "unacceptable" errors are handled elsewhere, I'm fine with what you have here 👍🏾

BritneyS (approved):
Approved! I just confirmed that those same remote tests fail on `master`.

PR #4764, opened 2023-04-28
CyberSource: Handling Canadian bank accounts

Heavyblade:
### Summary:

Add changes to CyberSource to properly format Canadian routing numbers when payment method is bank account.

GWS-48

### Remote Test:
Finished in 114.912426 seconds.
121 tests, 611 assertions, 5 failures, 0 errors,
95.8678% passed

*Notes on erros*:
The test suite was ran with a single account rason for errors:
- 2 errors correspondsto account not enabled for canadian bank transactions.
- 2 errors correspond to outdated 3ds PAR values.
- 1 error related with account not enabled for async.

### Unit Tests:
Finished in 41.573188 seconds.
5501 tests, 77346 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -705,8 +705,8 @@ module ActiveMerchant #:nodoc:
       def add_check(xml, check, options)
         xml.tag! 'check' do
           xml.tag! 'accountNumber', check.account_number
-          xml.tag! 'accountType', check.account_type[0]
-          xml.tag! 'bankTransitNumber', check.routing_number
+          xml.tag! 'accountType', check.account_type == 'checking' ? 'C' : 'S'
+          xml.tag! 'bankTransitNumber', format_routing_number(check.routing_number, options)
           xml.tag! 'secCode', options[:sec_code] if options[:sec_code]
         end
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -1131,6 +1131,10 @@ def message_from(response)
       def eligible_for_zero_auth?(payment_method, options = {})
         payment_method.is_a?(CreditCard) && options[:zero_amount_auth]
       end
+
+      def format_routing_number(routing_number, options)
+        options[:currency] == 'CAD' && routing_number.length > 8 ? routing_number[-8..-1] : routing_number
```
javierpedrozaing, line 1136: Some special reason to use routing_number[-8..-1] instead number.slice!(0) ❓, personally I like more the rubiest way 😅

javierpedrozaing (approved):
Looks good to me, just a minor question, and also looks like you need to re-run the job to pass the checks.

lib/active_merchant/billing/gateways/cyber_source.rb
Heavyblade: well, two reasons:

1) the semantics of [-8, -1] mean "the latest 8 characters", while slice!(0) means removing the first element.
2) just for fun debate => Handling strings with ranges is more idiomatic in Ruby than slicing strings that you will see in the Js world.

curiousepic (approved):
Looks good! Will merge soon.

PR #4768, opened 2023-05-02
Add rexml as a gem dependency

pi3r:
Rexml is no longer included with Ruby 3+, we therefore need to add the dependency explicitely. This fixes the issue encountered in https://github.com/activemerchant/active_merchant/pull/3852.

See https://www.ruby-lang.org/en/news/2020/09/25/ruby-3-0-0-preview1-released/

### DIFF
--- a/test/unit/gateways/payflow_test.rb
+++ b/test/unit/gateways/payflow_test.rb
@@ -1144,7 +1144,7 @@ class PayflowTest < Test::Unit::TestCase
 
   def threeds_xpath_for_extdata(attr_name, tx_type: 'Authorization')
     xpath_prefix = xpath_prefix_for_transaction_type(tx_type)
-    %(string(#{xpath_prefix}/PayData/ExtData[@Name='#{attr_name}']/@Value)")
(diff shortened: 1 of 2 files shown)


test/unit/gateways/payflow_test.rb
```diff
@@ -1144,7 +1144,7 @@ def xpath_prefix_for_transaction_type(tx_type)
 
   def threeds_xpath_for_extdata(attr_name, tx_type: 'Authorization')
     xpath_prefix = xpath_prefix_for_transaction_type(tx_type)
-    %(string(#{xpath_prefix}/PayData/ExtData[@Name='#{attr_name}']/@Value)")
+    %(string(#{xpath_prefix}/PayData/ExtData[@Name='#{attr_name}']/@Value))
```
pi3r, line 1147: @esmitperez I'm assuming the `"` introduced here was a typo?

hvenables (approved):

aenand (approved):
LGTM :shipit:

PR #4770, opened 2023-05-04
Adyen: Update Mastercard error messaging

kylene-spreedly:
Adyen error messaging uses the generic refusalReasonRaw field as part of the response message. Adyen offers a more detailed Mastercard-specific field called merchantAdviceCode that should be present for failed Mastercard transactions.

Adyen error messaging now checks for the merchantAdviceCode first. If it is not present (i.e. this is not a Mastercard transaction or it is a Mastercard transaction and this field is missing for some reason) then the default refusalReasonRaw field is used (like previous functionality).

ECS-2767

Unit:
107 tests, 539 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
132 tests, 443 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 90.9091% passed
*These 12 tests fail on master

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,7 @@
 * CyberSource: Handling Canadian bank accounts [heavyblade] #4764
 * CyberSource Rest: Fixing currency detection [heavyblade] #4777
 * CyberSource: Allow business rules for requests with network tokens [aenand] #4764
+* Adyen: Update Mastercard error messaging [kylene-spreedly] #4770
 
 == Version 1.129.0 (May 3rd, 2023)
 * Adyen: Update selectedBrand mapping for Google Pay [jcreiff] #4763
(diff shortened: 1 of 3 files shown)


aenand (commented):
This looks great! Is there a way we could trigger a `merchantAdviceCode` message in a remote test?

curiousepic (approved):
Looks good, agree it would be nice to have a remote test; you might check Adyen docs to see if there is a combination of test values that will invoke the response code, but not a big deal if not.

kylene-spreedly:
> This looks great! Is there a way we could trigger a `merchantAdviceCode` message in a remote test?

@aenand I tried a bunch of ways to do this (various test cards, triggering auth responses like [this](https://docs.adyen.com/development-resources/testing/result-code-testing/testing-with-card-holder-name)) and I couldn't get a response with a `merchantAdviceCode` unfortunately. If you have any other ideas, let me know!

aenand (approved):
Sometimes we can't engineer a response from a gateway sandbox. Change still looks good!

PR #4772, opened 2023-05-05
Cybersource: Add business rules for NT

aenand:
ECS-2849

A previous commit from 2015 restricted the ability to pass business rules such as `ignoreAVSResult` and `ignoreCVResult` on API requests with NetworkTokenization cards. Merchants are now asking for this to be allowed on requests with payment methods such as NT/AP/GP and the remote tests seem to indicate we can add these fields for these types of payment methods.

Remote:
119 tests, 607 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 97.479% passed


aenand:
Note for reviewers: These fields being passed will be covered by existing tests since those fields are in the `options` hash of the remote tests. Additionally the 3 failing tests are also failing on `master`

PR #4773, opened 2023-05-05
PayuLatam:

yunnydang:
The original method of surfacing error codes was redundant and didn't actually surface a network code that is particularly useful. This PR aims to fix that issue for failing transactions.


jcreiff (approved):
Looks good! Thanks for changing this

PR #4774, opened 2023-05-11
Worlpay: Fix Google Pay

almalee24:
Ensure that we don't send cardHolderName if empty and that Google Pay and Apple Pay fall into the network tokenization code path and not the credit card path.

Remote Tests:
100 tests, 416 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit Tests:
107 tests, 633 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -683,7 +683,8 @@ module ActiveMerchant #:nodoc:
             'year' => format(payment_method.year, :four_digits_year)
           )
         end
-        xml.cardHolderName card_holder_name(payment_method, options)
+        name = card_holder_name(payment_method, options)
+        xml.cardHolderName name if name.present?
         xml.cvc payment_method.verification_value
 
         add_address(xml, (options[:billing_address] || options[:address]), options)
@@ -995,10 +996,10 @@ module ActiveMerchant #:nodoc:
         case payment_method
         when String
           token_type_and_details(payment_method)
-        when NetworkTokenizationCreditCard
-          { payment_type: :network_token }
         else
-          { payment_type: :credit }
+          type = payment_method.respond_to?(:source) ? :network_token : :credit
+
+          { payment_type: type }
         end
       end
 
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -6,6 +6,7 @@ def setup
     @cftgateway = WorldpayGateway.new(fixtures(:world_pay_gateway_cft))
 
     @amount = 100
+    @year = (Time.now.year + 2).to_s[-2..-1].to_i
```
BritneyS, line 9: Nice. It may be controversial, but I'm a fan of negative indices in Ruby. Very straightforwards and readable.

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -995,10 +996,10 @@ def payment_details(payment_method)
         case payment_method
         when String
           token_type_and_details(payment_method)
-        when NetworkTokenizationCreditCard
-          { payment_type: :network_token }
         else
-          { payment_type: :credit }
+          type = payment_method.respond_to?(:source) ? :network_token : :credit
```
BritneyS, line 1000: Can you add an assertion to an existing tests for Google Pay/Apple Pay that confirms that they are now going through the network tokenization path?

almalee24: Yes I can. The weird this is that they were going through the network tokenization path with the current way which makes me think that we are saving Google or Apple Pay as NetworkTokenizationCreditCard but as CreditCard

test/unit/gateways/worldpay_test.rb
```diff
@@ -145,11 +145,36 @@ def setup
     }
   end
 
+  def test_successfully_attaching_payment_type
+    [@credit_card, @nt_credit_card].each do |pm|
+      payment = @gateway.send(:payment_details, pm)[:payment_type]
+      if pm.respond_to?(:source)
```
BritneyS, line 151: I think it's better to split these into separate tests to keep the passing conditions clear. This test may result in a false positive later on if, for example, a payment method contains a `source` field and it shouldn't.

almalee24: It wouldn't be possible for CreditCard to have the `source` field but I can still separate.

BritneyS (approved):
🚀

PR #4775, opened 2023-05-11
BT: Add support for network tokens

aenand:
This PR adds support for network tokens to the Braintree gateway. The gateway class utilizes the Braintree SDK which added support for network tokens in major version 4. This commit updates the logic to add support for tagging NT as per the BT docs but does not require the CI tests to move to the new BT version. 

Question for reviewers: Should we wait for the BT ruby gem to get updated in the repo's CI tests before merging this?

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -835,52 +835,84 @@ module ActiveMerchant #:nodoc:
 
       def add_payment_method(parameters, credit_card_or_vault_id, options)
         if credit_card_or_vault_id.is_a?(String) || credit_card_or_vault_id.is_a?(Integer)
-          if options[:payment_method_token]
-            parameters[:payment_method_token] = credit_card_or_vault_id
-            options.delete(:billing_address)
-          elsif options[:payment_method_nonce]
-            parameters[:payment_method_nonce] = credit_card_or_vault_id
-          else
-            parameters[:customer_id] = credit_card_or_vault_id
-          end
+          add_third_party_token(parameters, credit_card_or_vault_id, options)
         else
           parameters[:customer].merge!(
             first_name: credit_card_or_vault_id.first_name,
             last_name: credit_card_or_vault_id.last_name
           )
           if credit_card_or_vault_id.is_a?(NetworkTokenizationCreditCard)
-            if credit_card_or_vault_id.source == :apple_pay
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -827,54 +827,88 @@ def add_stored_credential_data(parameters, credit_card_or_vault_id, options)
...
+
+      def add_network_tokenization_card(parameters, payment_method)
+        parameters[:credit_card] = {
+          number: payment_method.number,
+          expiration_month: payment_method.month.to_s.rjust(2, '0'),
+          expiration_year: payment_method.year.to_s,
+          cardholder_name: payment_method.name,
+          network_tokenization_attributes: {
+            cryptogram: payment_method.payment_cryptogram,
+            ecommerce_indicator: payment_method.eci
```
BritneyS, line 907: Does Braintree require the the TAVV `eci` or CAVV `eci` here? If it's the TAVV `eci`, then we may need the upstream work I captured to get that value. Also, if that's the case, as long as this value can safely resolve to `nil` and that's valid from BT's perspective, I think this is fine as-is.

aenand: Good question! I'm not sure, I'll check out the docs and let you know

aenand: This seems to be CAVV

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -827,54 +827,88 @@ def add_stored_credential_data(parameters, credit_card_or_vault_id, options)
...
-                expiration_year: credit_card_or_vault_id.year.to_s,
-                google_transaction_id: credit_card_or_vault_id.transaction_id,
-                source_card_type: credit_card_or_vault_id.brand,
-                source_card_last_four: credit_card_or_vault_id.last_digits,
-                eci_indicator: credit_card_or_vault_id.eci
-              }
+            case credit_card_or_vault_id.source
+            when :apple_pay
+              add_apple_pay(parameters, credit_card_or_vault_id)
+            when :android_pay
```
almalee24, line 840: What do you think about removing `android_pay` in this PR?

BritneyS (approved):
Looks good 👍 Appreciate the tokenization decision refactor

PR #4776, opened 2023-05-12
Ogone: Enable 3ds Global for Ogone Gateway

javierpedrozaing:
Description
-------------------------------------------

This commit introduces the support for 3DS Global payments using the Ogone Direct API through GlobalCollect. As Ogone and GlobalCollect share the same underlying payment service provider (PSP), Worldline, we can leverage the new attribute 'ogone_direct' to use the appropriate credentials and endpoint to connect with the Ogone Direct API.

[SER-562](https://spreedly.atlassian.net/browse/SER-562)

UNIT TEST
-------------------------------------------

Finished in 0.253826 seconds.

44 tests, 225 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

173.35 tests/s, 886.43 assertions/s

REMOTE TEST
-------------------------------------------

Finished in 71.318909 seconds.

43 tests, 115 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 97.6744% passed

0.60 tests/s, 1.61 assertions/s

Note:
During testing, a single failure related to installment processing was identified with GlobalCollect. The error message "NO_ACQUIRER_CONFIGURED_FOR_INSTALLMENTS" I think that the issue may be related to GlobalCollect's account configuration, which is outside the scope of this update.

RUBOCOP
-------------------------------------------


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -401,18 +411,21 @@ def parse(body)
       end
 
       def url(action, authorization)
+        return ogone_direct + uri(action, authorization) if @options[:url_override].to_s == 'ogone_direct'
```
sinourain, line 414: `@options[:url_override]` is always populated? 

If the answer is no, I think it will fail when this is `nil`

sinourain (approved):
LGTM!

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -2,13 +2,15 @@ module ActiveMerchant #:nodoc:
...
       class_attribute :preproduction_url
+      class_attribute :ogone_direct
 
       self.display_name = 'GlobalCollect'
       self.homepage_url = 'http://www.globalcollect.com/'
 
       self.test_url = 'https://eu.sandbox.api-ingenico.com'
       self.preproduction_url = 'https://world.preprod.api-ingenico.com'
       self.live_url = 'https://world.api-ingenico.com'
+      self.ogone_direct = 'https://payment.preprod.direct.worldline-solutions.com'
```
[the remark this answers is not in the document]
javierpedrozaing: You're right updated, thanks

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -378,17 +380,25 @@ def add_external_cardholder_authentication_data(post, options)
         return unless threeds_2_options = options[:three_d_secure]
 
         authentication_data = {}
```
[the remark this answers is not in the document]
javierpedrozaing: thanks updated

lib/active_merchant/billing/gateways/global_collect.rb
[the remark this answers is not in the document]
javierpedrozaing: Totally agree, I have added a new function called ogone_direct? Thank you

test/fixtures.yml
```diff
@@ -439,9 +439,14 @@ garanti:
 
 global_collect:
   merchant_id: 2196
-  api_key_id: c91d6752cbbf9cf1
-  secret_api_key: xHjQr5gL9Wcihkqoj4w/UQugdSCNXM2oUQHG5C82jy4=
-
+  api_key_id: b2311c2c832dd238
+  secret_api_key: Av5wKihoVlLN8SnGm6669hBHyG4Y4aS4KwaZUCvEIbY=
+
+global_collect_preprod:
```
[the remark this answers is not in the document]
javierpedrozaing: ok I will ask in slack

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -474,13 +487,13 @@ def auth_digest(method, action, post, authorization = nil, options = {})
...
-        digest = OpenSSL::Digest.new('sha256')
+        digest = OpenSSL::Digest.new('SHA256')
         key = @options[:secret_api_key]
-        "GCS v1HMAC:#{@options[:api_key_id]}:#{Base64.strict_encode64(OpenSSL::HMAC.digest(digest, key, data))}"
+        "GCS v1HMAC:#{@options[:api_key_id]}:#{Base64.strict_encode64(OpenSSL::HMAC.digest(digest, key, data)).strip}"
       end
 
       def date
-        @date ||= Time.now.gmtime.strftime('%a, %d %b %Y %H:%M:%S %Z') # Must be same in digest and HTTP header
+        @date ||= Time.now.gmtime.strftime('%a, %d %b %Y %H:%M:%S GMT')
```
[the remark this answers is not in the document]
javierpedrozaing: yes, works

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -378,17 +380,25 @@ def add_external_cardholder_authentication_data(post, options)
...
         authentication_data[:eci] = threeds_2_options[:eci] if threeds_2_options[:eci]
         authentication_data[:threeDSecureVersion] = threeds_2_options[:version] if threeds_2_options[:version]
         authentication_data[:validationResult] = threeds_2_options[:authentication_response_status] if threeds_2_options[:authentication_response_status]
         authentication_data[:xid] = threeds_2_options[:xid] if threeds_2_options[:xid]
+        authentication_data[:acsTransactionId] = threeds_2_options[:acs_transaction_id] if threeds_2_options[:acs_transaction_id]
+        authentication_data[:flow] = threeds_2_options[:flow] if threeds_2_options[:flow]
 
         post['cardPaymentMethodSpecificInput'] ||= {}
         post['cardPaymentMethodSpecificInput']['threeDSecure'] ||= {}
+        post['cardPaymentMethodSpecificInput']['threeDSecure']['merchantFraudRate'] = threeds_2_options[:merchant_fraudRate] || 99
```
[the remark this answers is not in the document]
javierpedrozaing: yes authenticationFlow, challengeIndicator not are needed for 3ds2, and the others extra attributes works, but these are optional, I will update this, thanks

javierpedrozaing:
Hi @Heavyblade thanks you for your comments. 👍 
The code has been improved with validations based on whether the `Ogone Direct API` is being utilized. Furthermore, the test suite has been modified to incorporate additional tests and updated expectations. 
This update was necessary as the response varies depending on whether `Global Collect`, `preprod`, or `Ogone Direct `is used. 
Finally, the `fixture` file has been modified to remove the `global_collect_preprod` credentials. I think that as the Ogone direct API credentials, we should not include this in the code.

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -248,9 +252,10 @@ def add_creator_info(post, options)
       end
 
       def add_amount(post, money, options = {})
+        currency_ogone = 'EUR' if ogone_direct?
         post['amountOfMoney'] = {
           'amount' => amount(money),
-          'currencyCode' => options[:currency] || currency(money)
+          'currencyCode' => currency_ogone || options[:currency] || currency(money)
```
Heavyblade, line 258: ⚠️  Warning:

The problem with this approach is that you are forcing all Ogone direct users to use EUR in the future and with that limiting the future possibility for direct handling USD or other currencies

Heavyblade (commented):
Done with my Review @javierpedrozaing leaved just 1 comment for your consideration and one general question:

> My understanding was that you should somehow update the response structure from direct to the GloballCollect one, and fix a bunch of remote test failures, wasn't that necessary at all ?

lib/active_merchant/billing/gateways/global_collect.rb
javierpedrozaing: Hi @Heavyblade you're right, I did a small change in order to give the opportunity to get first a currency if was specified. thanks

javierpedrozaing:
> 

Hi @Heavyblade I apologize for the misunderstanding, but from my perspective, it is not necessary because we are dealing with distinct endpoints and gateways, and there is no requirement to modify the response for each one. I think that our focus should primarily be on adapting or adjusting the expectations, if it is necessary.

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -2,13 +2,17 @@ module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class GlobalCollectGateway < Gateway
       class_attribute :preproduction_url
+      class_attribute :ogone_direct_test
+      class_attribute :ogone_direct_live
```
aenand, line 6: Why is this being added to the `global_collect` gateway if it's ogone?

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -377,18 +382,24 @@ def add_fraud_fields(post, options)
...
+          xid: threeds_2_options[:xid],
+          acsTransactionId: threeds_2_options[:acs_transaction_id],
+          flow: threeds_2_options[:flow]
+        }.compact
 
         post['cardPaymentMethodSpecificInput'] ||= {}
         post['cardPaymentMethodSpecificInput']['threeDSecure'] ||= {}
+        post['cardPaymentMethodSpecificInput']['threeDSecure']['merchantFraudRate'] = threeds_2_options[:merchant_fraud_rate] || 99
+        post['cardPaymentMethodSpecificInput']['threeDSecure']['exemptionRequest'] = 'none'
+        post['cardPaymentMethodSpecificInput']['threeDSecure']['secureCorporatePayment'] = threeds_2_options[:secure_corporate_payment] || true
```
aenand, line 402: Why do we set a default for these three? Does Ogone support the MOTO exemption for 3DS2?

lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -114,7 +118,7 @@ def add_order(post, money, options, capture: false)
         post['order']['references']['invoiceData'] = {
           'invoiceNumber' => options[:invoice]
         }
-        add_airline_data(post, options)
+        add_airline_data(post, options) unless ogone_direct?
```
aenand, line 121: Why is `airline_data` only available for ogone_direct?

test/fixtures.yml
```diff
@@ -439,9 +439,9 @@ garanti:
 
 global_collect:
   merchant_id: 2196
-  api_key_id: c91d6752cbbf9cf1
-  secret_api_key: xHjQr5gL9Wcihkqoj4w/UQugdSCNXM2oUQHG5C82jy4=
-
+  api_key_id: b2311c2c832dd238
+  secret_api_key: Av5wKihoVlLN8SnGm6669hBHyG4Y4aS4KwaZUCvEIbY=
```
naashton, line 443: Is this supposed to be for a new fixtures entry for `global_collect_direct`?

lib/active_merchant/billing/gateways/global_collect.rb
javierpedrozaing: It is a little weird,  as far I know  `Ogone` and `GlobalCollect` has the same PSP, `Ogone` has two types of integration Classic/Direct that's why we are using GlobalCollect gateway to connect the Direct API for `Ogone`, because the `GlobalCollect` API seems to have a similar integration for the Direct API

lib/active_merchant/billing/gateways/global_collect.rb
javierpedrozaing: No, it's the opposite, `airline_data` is only for `GlobalCollect` if `ogone_direct` is false we fill this information.

test/fixtures.yml
javierpedrozaing: Not really, these are the current credentials for `GlobalCollect`. For `Ogone` Direct API the credentials are in 1password, but I think that is not a good idea to include it into this commit

test/remote/gateways/remote_global_collect_test.rb
```diff
@@ -6,8 +6,12 @@ def setup
     @gateway_preprod = GlobalCollectGateway.new(fixtures(:global_collect_preprod))
     @gateway_preprod.options[:url_override] = 'preproduction'
 
+    @gateway_direct = GlobalCollectGateway.new(fixtures(:global_collect_direct))
```
naashton, line 9: Since we are adding a new fixtures entry, we should stub `global_collect_direct` in the `fixtures.yml`.

javierpedrozaing: Thanks @naashton I will add the fixtures.

naashton (approved):
LGTM!

PR #4138, opened 2021-10-04
Adyen: Adjust phone number mapping

aenand:
Adds a bit of logic to the way the Adyen gateway passes phone numbers so that if a payment method has a phone number saved to it that will be mapped first (`billing_address[:phone]`), otherwise if the transaction has (`billing_address[phone_number]`) that will be mapped. If neither are available we will map an empty string. Similar to [#4005](https://github.com/activemerchant/active_merchant/pull/4005).

CE-1908

Test Summary:

There were 17 test errors due to not having monei api_key
Local:
4930 tests, 74297 assertions, 0 failures, 17 errors, 0 pendings, 0 omissions, 0 notifications
99.6552% passed

Unit:
88 tests, 450 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
119 tests, 441 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

PR #4778, opened 2023-05-12
Authorize.net: update mapping for billing address phone number

jcreiff:
Adds a bit of logic to the Authorize.net gateway so that phone number can be passed via `billing_address[phone_number]` in addition to `billing_address[phone]` This is similar to #4138

CER-590

LOCAL
5503 tests, 77374 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

UNIT
121 tests, 681 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

REMOTE
84 tests, 301 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


rachelkirk (approved):
Looks good, just to refresh my memory are we doing this in an addition to a broader change in Core?

PR #4782, opened 2023-05-16
Stripe PI: Update paramters for creating of customer

almalee24:
Start sending address, shipping, phone and email when creating a customer.

Remote Tests
84 tests, 396 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit Tests

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -76,22 +76,27 @@ module ActiveMerchant #:nodoc:
 
       def create_payment_method(payment_method, options = {})
         post_data = add_payment_method_data(payment_method, options)
-
         options = format_idempotency_key(options, 'pm')
         commit(:post, 'payment_methods', post_data, options)
       end
 
       def add_payment_method_data(payment_method, options = {})
-        post_data = {}
-        post_data[:type] = 'card'
-        post_data[:card] = {}
-        post_data[:card][:number] = payment_method.number
-        post_data[:card][:exp_month] = payment_method.month
-        post_data[:card][:exp_year] = payment_method.year
-        post_data[:card][:cvc] = payment_method.verification_value if payment_method.verification_value
-        add_billing_address(post_data, options)
-        add_name_only(post_data, payment_method) if post_data[:billing_details].nil?
-        post_data
+        post = {
+          type: 'card',
+          card: {
+            number: payment_method.number,
+            exp_month: payment_method.month,
+            exp_year: payment_method.year
+          }
+        }
+
+        post[:card][:cvc] = payment_method.verification_value if payment_method.verification_value
+        if billing = options[:billing_address] || options[:address]
+          post[:billing_details] = add_address(billing, options)
+        end
+
+        add_name_only(post, payment_method) if post[:billing_details].nil?
+        post
       end
 
       def add_payment_method_card_data_token(post_data, payment_method)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -478,30 +490,34 @@ def setup_future_usage(post, options = {})
...
-        post[:billing_details][:address][:city] = billing[:city] if billing[:city]
-        post[:billing_details][:address][:country] = billing[:country] if billing[:country]
-        post[:billing_details][:address][:line1] = billing[:address1] if billing[:address1]
-        post[:billing_details][:address][:line2] = billing[:address2] if billing[:address2]
-        post[:billing_details][:address][:postal_code] = billing[:zip] if billing[:zip]
-        post[:billing_details][:address][:state] = billing[:state] if billing[:state]
-        post[:billing_details][:email] = email if email
-        post[:billing_details][:name] = billing[:name] if billing[:name]
-        post[:billing_details][:phone] = billing[:phone] if billing[:phone]
+      def add_address(post, address, options)
```
aenand, line 507: Why are we passing in `post` if we don't add to it here?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -231,6 +227,22 @@ def store(payment_method, options = {})
...
+        post[:description] = options[:description] if options[:description]
+        post[:expand] = [:sources]
+
+        if billing = options[:billing_address] || options[:address]
+          post.merge!(add_address(post, billing, options))
+        end
+
+        if shipping = options[:shipping_address]
+          post[:shipping] = add_address({}, shipping, options)
+        end
```
aenand, line 240: Why do we sometimes merge the result of `add_address` and sometimes it not?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: I'll remove sending post

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: For customer billing address needs to be in the [address hash ](https://stripe.com/docs/api/customers/object#customer_object-address) and for shipping in needs to be in the [shipping hash](https://stripe.com/docs/api/customers/object#customer_object-shipping)

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -212,16 +217,7 @@ def store(payment_method, options = {})
...
-            customer_id = options[:customer]
-          else
-            post[:description] = options[:description] if options[:description]
-            post[:email] = options[:email] if options[:email]
-            options = format_idempotency_key(options, 'customer')
-            post[:expand] = [:sources]
-            customer = commit(:post, 'customers', post, options)
-            customer_id = customer.params['id']
-          end
+          customer_id = options[:customer] || customer(post, payment_method, options).params['id']
```
aenand, line 220: There are some fields being mutated with `post` here when creating a customer, why delete them? Am I missing where `email` or the re-formatted `idempotency_key` would now get added to `post`?

aenand (changes requested):

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: `email` get set in `add_address` for `billing_address` and [idempotency_key here](https://github.com/activemerchant/active_merchant/pull/4782/files#diff-19492101b546175aa3ca6a16ec7d0a610d0adae03cbaf130eff54092dec8e0fbR242).  I just moved the logic to [customer method](https://github.com/activemerchant/active_merchant/pull/4782/files#diff-19492101b546175aa3ca6a16ec7d0a610d0adae03cbaf130eff54092dec8e0fbR230) because that's the only place that post get's used.

aenand: `email` can be sent in the top level `options[:email]` not just in the nested `address` hash and the current code of `add_address` is only run if an address is passed in. Would this mean that there is a scenario where someone passes in an `email` at the top level, no address, and the email does not get sent?

almalee24: This is true, I'll update the email

aenand (approved):
LGTM 🚀

PR #4783, opened 2023-05-17
Remove Android Pay

almalee24:
Remove any use of the Android Pay payment method.


CHANGELOG
```diff
@@ -1684,13 +1684,12 @@
 == Version 1.68.0 (June 27, 2017)
 * Authorize.Net: Return failed response if forced refund settlement fails [bizla] #2476
 * Authorize.net: Concatenate address1 and address2 [dtykocki] #2479
-* Braintree Blue: Braintree Blue: Add ECI indicator to Android Pay transactions [davidsantoso] #2474
```
aenand, line 1687: We shouldn't remove lines from the CHANGELOG but instead add a new line that says "remove support for Android Pay" that way the CHANGELOG serves as an accurate record of history

almalee24: Okay that makes more sense lol

PR #4784, opened 2023-05-25
Redsys: Set appropriate request fields for stored credentials with CITs and MITs

BritneyS:
Following pre-determined guidance for CIT/MIT request fields for this gateway. We were getting it mostly right, but almost didn't count, so the `DS_MERCHANT_DIRECTPAYMENT=false` value was added for initial CITs.

Both CITs and MITs should be indicated with the `stored_credential` field `recurring`, so as long as that happens, `DS_MERCHANT_COF_TYPE` should have the value as `R` in both transactions.

An outstanding task is to pass `DS_MERCHANT_IDENTIFIER` as 'REQUIRED' for CITs. I'm currently blocked on the credentials for the Redsys sandbox environment (getting the error `SIS0042` - Signature calculation error). This means that I'm unable to confirm that Redsys indeed returns a `DS_MERCHANT_IDENTIFIER` value in their response, in order to pass in a future MIT.

Test Summary
Local: 5512 tests, 77418 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
       100% passed

Unit: 35 tests, 122 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
      100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys.rb
+++ b/lib/active_merchant/billing/gateways/redsys.rb
@@ -538,6 +538,7 @@ module ActiveMerchant #:nodoc:
             xml.DS_MERCHANT_COF_INI data[:DS_MERCHANT_COF_INI]
             xml.DS_MERCHANT_COF_TYPE data[:DS_MERCHANT_COF_TYPE]
             xml.DS_MERCHANT_COF_TXNID data[:DS_MERCHANT_COF_TXNID] if data[:DS_MERCHANT_COF_TXNID]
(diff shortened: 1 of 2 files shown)


aenand (approved):
LGTM

BritneyS:
I'm going to manually close this PR, but the change actually made it to this commit SHA: https://github.com/activemerchant/active_merchant/commit/9c309870fcb00490c7065e8619043c7f1bbc0136

PR #4785, opened 2023-05-25
DLocal: Add gateway specific field save on card

yunnydang:
Add the save field to card object, documentation found [here](https://docs.dlocal.com/docs/saving-cards#save-a-card-on-a-payment)

Local:
5512 tests, 77364 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6553% passed

Unit:
44 tests, 191 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
39 tests, 105 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.8718% passed


test/remote/gateways/remote_d_local_test.rb
```diff
@@ -50,6 +50,14 @@ def test_successful_purchase
     assert_match 'The payment was paid', response.message
   end
 
+  def test_successful_purchase_with_save_option
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(save: true))
+    assert_success response
+    assert_equal true, response.params['card']['save']
+    assert_not_empty response.params['card']['card_id']
```
jcreiff, line 57: Can you also assert that we're receiving the `[card][type]` in the response, since that's one of the motivations for adding this?

yunnydang: Hey @jcreiff after a little more digging, I see that the card['type'] is not returned in the sandbox tests response even when calling verify instead of purchase. This may be a limitation of the gateway sandbox but if there's something I'm missing, please let me know, thanks!

yunnydang: hey @jcreiff the sandbox is now returning the card type so i added the assertion to the remote test, thanks!

jcreiff (approved):
🚀 Nice work!

PR #4794, opened 2023-05-31
Cybersource: auto void r230

aenand:
ECS-2870

Cybersource transactions that fail with a reasonCode of `230` are in a state where the gateway advises the merchant to decline but has not declined it themselves. Instead the transaction is pending capture which can create a mismatch in reporting.

This commit attempts to auto void transactions that have this response code and a flag that indicates the merchants would like to auto void these kinds of transactions. I wasn't able to add a remote test for this because the CYBS sandbox does not have a way to trigger this response code. 

Remote:
121 tests, 611 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 95.8678% passed

5 test failures on master as well

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -1053,12 +1053,22 @@ module ActiveMerchant #:nodoc:
         message = message_from(response)
         authorization = success || in_fraud_review?(response) ? authorization_from(response, action, amount, options) : nil
 
+        message = auto_void?(authorization_from(response, action, amount, options), response, message, options)
+
         Response.new(success, message, response,
           test: test?,
           authorization: authorization,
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -1053,6 +1053,8 @@ def commit(request, action, amount, options)
         message = message_from(response)
         authorization = success || in_fraud_review?(response) ? authorization_from(response, action, amount, options) : nil
 
+        auto_void?(authorization_from(response, action, amount, options), response, options)
```
almalee24, line 1056: Should we update the message to reflect that we voided the transaction because of the error code?

almalee24 (approved):
Looks great!

PR #4796, opened 2023-06-01
Stripe PI: Update parameters for creation of customer

almalee24:
Start sending address, shipping, phone and email when creating a customer.

Remote Tests
84 tests, 396 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit Tests

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -76,22 +76,27 @@ module ActiveMerchant #:nodoc:
 
       def create_payment_method(payment_method, options = {})
         post_data = add_payment_method_data(payment_method, options)
-
         options = format_idempotency_key(options, 'pm')
         commit(:post, 'payment_methods', post_data, options)
       end
 
       def add_payment_method_data(payment_method, options = {})
-        post_data = {}
-        post_data[:type] = 'card'
-        post_data[:card] = {}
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -231,6 +227,23 @@ def store(payment_method, options = {})
         end
       end
 
+      def customer(post, payment, options)
```
BritneyS, line 230: ~~Maybe call this method `add_customer` to match the style of the rest of the class~~
Nevermind, I see that `customer` is actually an endpoint method

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -231,6 +227,23 @@ def store(payment_method, options = {})
...
+        if billing = options[:billing_address] || options[:address]
+          post.merge!(add_address(billing, options))
+        end
+
+        if shipping = options[:shipping_address]
+          post[:shipping] = add_address(shipping, options).except(:email)
+        end
+
+        options = format_idempotency_key(options, 'customer')
+        commit(:post, 'customers', post, options)
```
BritneyS, line 244: This is interesting that this is a synchronous call to `/customers` within the flow that later calls `/payment_methods/{:payment_method}/attach`. Is there a remote test flow for `store` that confirms both responses are handled well?

PR #4800, opened 2023-06-03
Checkout: Add support for several customer data fields

rachelkirk:
CER-595

Ran into a lot of tricky data collisions and had to do some finagling to make sure existing test cases would pass. I left open two possibilities for passing in the phone number depending on how users would like to pass it in: manually via options or via the phone number that’s attached to credit card billing data on the payment method.

It’s possible to pay with a stored payment method which would be a String. In this case there’s no name data attached so added some guarding against NoMethodErrors that were resulting from trying to call payment_method.name.

Remote Tests:
92 tests, 221 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.7391% passed
*5 remote tests are failing on master but I fixed two of them

Unit Tests:
57 tests, 319 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Local Tests:
5516 tests, 77434 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -139,6 +139,7 @@ module ActiveMerchant #:nodoc:
         add_authorization_type(post, options)
         add_payment_method(post, payment_method, options)
         add_customer_data(post, options)
+        add_extra_customer_data(post, payment_method, options)
         add_shipping_address(post, options)
         add_stored_credential_options(post, options)
         add_transaction_data(post, options)
@@ -237,6 +238,15 @@ module ActiveMerchant #:nodoc:
           post[:source][:billing_address][:country] = address[:country] unless address[:country].blank?
           post[:source][:billing_address][:zip] = address[:zip] unless address[:zip].blank?
         end
+      end
+
+      # created a separate method for these fields because they should not be included
+      # in all transaction types that include methods with source and customer fields
+      def add_extra_customer_data(post, payment_method, options)
+        post[:source][:phone] = {}
+        post[:source][:phone][:number] = options[:phone] || options.dig(:billing_address, :phone) || options.dig(:billing_address, :phone_number)
+        post[:source][:phone][:country_code] = options[:phone_country_code] if options[:phone_country_code]
+        post[:customer][:name] = payment_method.name if payment_method.respond_to?(:name)
       end
 
       def add_shipping_address(post, options)
--- a/test/remote/gateways/remote_checkout_v2_test.rb
+++ b/test/remote/gateways/remote_checkout_v2_test.rb
@@ -100,6 +100,10 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
         authentication_response_status: 'Y'
       }
     )
+    @extra_customer_data = @options.merge(
+      phone_country_code: '1',
+      phone: '9108675309'
+    )
   end
 
   def test_transcript_scrubbing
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -239,6 +240,15 @@ def add_customer_data(post, options)
         end
       end
 
+      # created a separate method for these fields because they should not be included
+      # in all transaction types that include methods with source and customer fields
+      def add_extra_customer_data(post, payment_method, options)
+        post[:source][:phone] = {}
+        post[:source][:phone][:number] = options[:phone] || options.dig(:billing_address, :phone)
+        post[:source][:phone][:country_code] = options[:phone_country_code] if options[:phone_country_code]
+        post[:customer][:name] = payment_method.name if payment_method.is_a?(CreditCard) || payment_method.is_a?(NetworkTokenizationCreditCard)
```
jcreiff, line 249: What would you think about guarding here with `if payment_method.respond_to?(:name)` instead?

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -470,7 +480,7 @@ def test_avs_failed_authorize
   def test_invalid_shipping_address
     response = @gateway.authorize(@amount, @credit_card, shipping_address: address.update(country: 'Canada'))
     assert_failure response
-    assert_equal 'request_invalid: address_country_invalid', response.message
+    assert_equal 'request_invalid: country_address_invalid', response.message
```
jcreiff, line 483: Thanks for fixing these! Do you, uh, want to fix the other 3 as well? If you're getting the same failures I'm getting, I think those are all test expectations that can be updated

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -239,6 +240,15 @@ def add_customer_data(post, options)
         end
       end
 
+      # created a separate method for these fields because they should not be included
+      # in all transaction types that include methods with source and customer fields
+      def add_extra_customer_data(post, payment_method, options)
+        post[:source][:phone] = {}
+        post[:source][:phone][:number] = options[:phone] || options.dig(:billing_address, :phone)
```
jcreiff, line 247: There's part of me that thinks we should also check for `billing_address.phone_number` but in reality, we probably just need to fix the root of that problem, not the symptom 😞

jcreiff (changes requested):
Just a couple of small things!

test/remote/gateways/remote_checkout_v2_test.rb
rachelkirk: Haha sure! I was working on this way too late at night so I was probably just too tired to notice.

lib/active_merchant/billing/gateways/checkout_v2.rb
rachelkirk: I remember you wrangling with this but I think in my tired-ness I was thinking that `billing_address.phone_number` was dependent on the gateway but couldn't remember. Was working down to the wire but would love to talk about this more.

lib/active_merchant/billing/gateways/checkout_v2.rb
rachelkirk: Lovely, thank you for making me aware of a new Ruby method. I will give it a whirl!

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -432,7 +442,7 @@ def test_successful_purchase_with_shipping_address
   end
 
   def test_successful_purchase_without_phone_number
```
jcreiff, line 444: Good idea to include this! Perhaps we should include one that validates nothing unexpected happens if `payment_method.name` contains no data too?

PR #4801, opened 2023-06-05
Stripe & Stripe PI: Validate API Key

almalee24:
Stripe Unit:
145 tests, 765 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Stripe Remote:
77 tests, 362 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Stripe PI Unit:
Stripe PI Remote:
83 tests, 391 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -652,18 +652,19 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def headers(options = {})
-        key = options[:key] || @api_key
-        idempotency_key = options[:idempotency_key]
+      def key(options = {})
+        options[:key] || @api_key
+      end
 
+      def headers(options = {})
         headers = {
-          'Authorization' => 'Basic ' + Base64.strict_encode64(key.to_s + ':').strip,
+          'Authorization' => 'Basic ' + Base64.strict_encode64(key(options).to_s + ':').strip,
           'User-Agent' => "Stripe/v1 ActiveMerchantBindings/#{ActiveMerchant::VERSION}",
           'Stripe-Version' => api_version(options),
           'X-Stripe-Client-User-Agent' => stripe_client_user_agent(options),
           'X-Stripe-Client-User-Metadata' => { ip: options[:ip] }.to_json
         }
-        headers['Idempotency-Key'] = idempotency_key if idempotency_key
+        headers['Idempotency-Key'] = options[:idempotency_key] if options[:idempotency_key]
         headers['Stripe-Account'] = options[:stripe_account] if options[:stripe_account]
         headers
       end
@@ -694,6 +695,10 @@ module ActiveMerchant #:nodoc:
 
       def commit(method, url, parameters = nil, options = {})
         add_expand_parameters(parameters, options) if parameters
+
+        return Response.new(false, 'Invalid API Key provided') if test? && !key(options).start_with?('sk_test')
+        return Response.new(false, 'Invalid API Key provided') if !test? && !key(options).start_with?('sk_live')
+
         response = api_request(method, url, parameters, options)
         response['webhook_id'] = options[:webhook_id] if options[:webhook_id]
         success = success_from(response, options)
--- a/test/unit/gateways/stripe_payment_intents_test.rb
+++ b/test/unit/gateways/stripe_payment_intents_test.rb
@@ -4,7 +4,7 @@ class StripePaymentIntentsTest < Test::Unit::TestCase
   include CommStub
 
   def setup
-    @gateway = StripePaymentIntentsGateway.new(login: 'login')
+    @gateway = StripePaymentIntentsGateway.new(login: 'sk_test_login')
 
     @credit_card = credit_card()
     @threeds_2_card = credit_card('4000000000003220')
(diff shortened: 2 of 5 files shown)


test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -259,6 +259,20 @@ def test_failed_payment_methods_post
     assert_equal 'invalid_request_error', create.params.dig('error', 'type')
   end
 
+  def test_invalid_login_test_transaction
+    gateway = StripePaymentIntentsGateway.new(login: 'sk_live_3422')
```
aenand, line 263: Why would we expect this to fail? Is test `true` by default?

test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -259,6 +259,20 @@ def test_failed_payment_methods_post
...
+    assert response = gateway.purchase(@amount, @credit_card, @options)
+    assert_failure response
+    assert_match 'Invalid API Key provided', response.message
+  end
+
+  def test_invalid_login_live_transaction
+    gateway = StripePaymentIntentsGateway.new(login: 'sk_test_3422', test: false)
```
aenand, line 270: This may be my morning brain but how are we doing a check for `production` gateways? I see `test? && !key(options).include?('sk_test')` and to me that reads that it only validates if `test=true` but here you're saying we validate also when `test=false`?

almalee24: You are right, I forgot that part which I realized after I created the PR lol I'll add the production check

test/unit/gateways/webpay_test.rb
```diff
@@ -4,7 +4,7 @@ class WebpayTest < Test::Unit::TestCase
   include CommStub
 
   def setup
-    @gateway = WebpayGateway.new(login: 'login')
+    @gateway = WebpayGateway.new(login: 'sk_test_login')
```
aenand, line 7: Why make a change to `webpay`?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -652,18 +652,19 @@ def flatten_array(flattened, array, prefix)
...
-        key = options[:key] || @api_key
-        idempotency_key = options[:idempotency_key]
+      def key(options = {})
+        options[:key] || @api_key
+      end
 
+      def headers(options = {})
         headers = {
-          'Authorization' => 'Basic ' + Base64.strict_encode64(key.to_s + ':').strip,
+          'Authorization' => 'Basic ' + Base64.strict_encode64(key(options).to_s + ':').strip,
```
mashton, line 661: will `.to_s` here crash if `@api_key` is `nil`? I see that this isn't a new issue here, so maybe not worth blocking this time sensitive change?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -694,6 +695,10 @@ def api_request(method, endpoint, parameters = nil, options = {})
 
       def commit(method, url, parameters = nil, options = {})
         add_expand_parameters(parameters, options) if parameters
+
+        return Response.new(false, 'Invalid API Key provided') if test? && !key(options).include?('sk_test')
+        return Response.new(false, 'Invalid API Key provided') if !test? && !key(options).include?('sk_live')
```
mashton, line 700: Would it be more accurate and maybe more secure to use `.starts_with?` instead of `.include?` here?

test/unit/gateways/webpay_test.rb
aenand: :themoreyouknow:

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -694,6 +695,10 @@ def api_request(method, endpoint, parameters = nil, options = {})
 
       def commit(method, url, parameters = nil, options = {})
         add_expand_parameters(parameters, options) if parameters
+
+        return Response.new(false, 'Invalid API Key provided') if test? && !key(options).start_with?('sk_test')
+        return Response.new(false, 'Invalid API Key provided') if !test? && !key(options).start_with?('sk_live')
```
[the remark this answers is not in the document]
almalee24: Ah for `response_is_test?` the seems to get it from the response which would correlate with the API key used. But we want to return the error before we make a request so using `test?` is appropriate.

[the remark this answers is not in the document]
britth: gotcha! I guess I'm just wondering why we wouldn't be using `test?` down there too (i.e. was there a reason the boolean wasn't reliable for determining test status), but I might be overthinking it!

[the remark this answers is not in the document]
almalee24: It looks like it's been like that since it the gateway was created, https://github.com/activemerchant/active_merchant/commit/ca39e94032097c53c2848119431da32ff6da3f31#diff-75f5fee715fd4b5bedbae9c805cbeb103100cdb5c1bf8d56c248aaa31c947e1c. Not sure why they decided to do it that way

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -652,18 +652,19 @@ def flatten_array(flattened, array, prefix)
         end
       end
 
-      def headers(options = {})
-        key = options[:key] || @api_key
-        idempotency_key = options[:idempotency_key]
+      def key(options = {})
+        options[:key] || @api_key
```
britth, line 656: just to double check, do we know that the `@fee_refund_api_key` (which gets set as options[:key]) follows the same format `sk_*` as a regular api key?

almalee24: I'm not sure

almalee24: But they key we use for Authentication for all request would be `options[:key] || @api_key`

almalee24: It also looks like support for the refund_api_key was removed at some point. It's missing all the methods it was created with, https://github.com/activemerchant/active_merchant/commit/36406508acaab8a70c9f1d1866c378a9152f6241

britth: i saw [this reference here](https://github.com/activemerchant/active_merchant/blob/4f96d540d676369e2ec2071af3b4f14109e6bd62/lib/active_merchant/billing/gateways/stripe.rb#L169-L174) but not really seeing where we use it on our side of the house so 🤔

almalee24: Yeah I'm not sure either, I think the important key is the one in the headers method. Let me look at the Stripe docs though

almalee24: So we might end up using it because it gets set as key in options which is what we check in the `key?` method. However it looks like it should follow the same `sk...` format if you look at the stripe docs, https://stripe.com/docs/api/fee_refunds/create. However, I don't think we used it cause it's not a field we offer.

britth: cool, thanks for checking that!

britth (approved):
🎉

lib/active_merchant/billing/gateways/stripe.rb
almalee24: I think it's something that we might need to look at removing later.

PR #4807, opened 2023-06-09
Stripe PI: Add new stored credential flag

aenand:
Stripe has a field called `stored_credential_transaction_type` to assist merchants who vault outside of Stripe to recognize card on file transactions at Stripe. This field does require Stripe enabling your account with this field.

The standard stored credential fields map to the various possibilities that Stripe makes available.

Test Summary
Remote:
87 tests, 409 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -419,6 +421,41 @@ def add_stored_credentials(post, options = {})
...
 
+      def add_stored_credential_transaction_type(post, options = {})
+        stored_credential = options[:stored_credential]
+        # Do not add anything unless these are present.
+        return unless stored_credential[:reason_type] && stored_credential[:initiator]
+
+        # Not compatible with off_session parameter.
+        options.delete(:off_session)
+
+        if stored_credential[:initial_transaction]
```
almalee24, line 432: Why not instead stored_credential[:initial_transaction] && stored_credential[:initiator] == 'cardholder'?

aenand: If the first conditional was that then a scenario of `initial_transaction = true and initiator = 'merchant'` would get sent to `else` and marked as `stored_off_session_unscheduled` which is incorrect

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -399,17 +399,19 @@ def add_exemption(post, options = {})
         post[:payment_method_options][:card][:moto] = true if options[:moto]
       end
 
-      # Stripe Payment Intents does not pass any parameters for cardholder/merchant initiated
-      # it also does not support installments for any country other than Mexico (reason for this is unknown)
-      # The only thing that Stripe PI requires for stored credentials to work currently is the network_transaction_id
-      # network_transaction_id is created when the card is authenticated using the field `setup_for_future_usage` with the value `off_session` see def setup_future_usage below
+      # Stripe Payment Intents now supports specifying on a transaction level basis stored credential information.
```
BritneyS, line 402: Does this field also enable installments to be supported in Mexico? Wondering if that part of the note should be removed or not.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -419,6 +421,41 @@ def add_stored_credentials(post, options = {})
...
+        return unless stored_credential[:reason_type] && stored_credential[:initiator]
+
+        # Not compatible with off_session parameter.
+        options.delete(:off_session)
+
+        if stored_credential[:initial_transaction]
+          # Initial transactions must by CIT
+          return unless stored_credential[:initiator] == 'cardholder'
+
+          if stored_credential[:reason_type] == 'unscheduled'
```
BritneyS, line 436: The `if/else` nesting is a bit hard to read. Maybe try handling the nested `if/else` conditional in a separate method?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
aenand: So I'm not really sure what the point of that comment was before, there was never installment logic/flagging nor country specific logic. I don't think it's worth keeping since it's not related to a particular piece of code logic

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -419,6 +421,50 @@ def add_stored_credentials(post, options = {})
...
 
+      def add_stored_credential_transaction_type(post, options = {})
+        stored_credential = options[:stored_credential]
+        # Do not add anything unless these are present.
+        return unless stored_credential[:reason_type] && stored_credential[:initiator]
+
+        # Not compatible with off_session parameter.
+        options.delete(:off_session)
+        if stored_credential[:initial_transaction]
+          # Initial transactions must by CIT
```
BritneyS, line 432: Nice breakdown by scenario 🌟

BritneyS (approved):
🚢

PR #4813, opened 2023-06-16
Element Gateway: Add all lodging fields

yunnydang:
Documentation for this can be found [here](https://developerengine.fisglobal.com/apis/express/express-xml/classes#lodging). I had to add support for a few more fields otherwise the lodging fields cannot be supported (they're extended parameters class). also with the tests files, new credentials were given and I had to clean up the remote tests specs to follow along with the new sandbox configurations. All tests are now passing and updated in this change.

Local:
5625 tests, 78037 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6622% passed

Unit:
27 tests, 149 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 
32 tests, 88 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/element.rb
+++ b/lib/active_merchant/billing/gateways/element.rb
@@ -37,6 +37,7 @@ module ActiveMerchant #:nodoc:
             add_transaction(xml, money, options)
             add_terminal(xml, options)
             add_address(xml, options)
+            add_lodging(xml, options)
           end
         end
 
@@ -51,6 +52,7 @@ module ActiveMerchant #:nodoc:
             add_transaction(xml, money, options)
             add_terminal(xml, options)
             add_address(xml, options)
+            add_lodging(xml, options)
           end
         end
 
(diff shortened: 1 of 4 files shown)


jcreiff (dismissed):
OK there's one small typo that needs revising (`nummber`)

But I also have two thoughts for consideration:

- If we're going to include `add_lodging` on all transaction types, should we at least have a single test for each type? An auth + capture, for example, and/or an auth + void. Or, alternatively, do we think the use case is such that the lodging fields aren't required (or available?) on all transaction types?
- What if instead of a bunch of fields that are `[:options][:lodging_*]`, we constructed a separate `[:options][:lodging]` block? Then you could use a guard clause at the top of the `add_lodging` method to check if `[:options][:lodging]` exists and `return` if not. It might look something like the `add_address` method if you opt for this route

lib/active_merchant/billing/gateways/element.rb
```diff
@@ -135,6 +141,7 @@ def verify(credit_card, options = {})
             add_payment_method(xml, credit_card)
             add_transaction(xml, 0, options)
             add_terminal(xml, options)
+            add_lodging(xml, options)
```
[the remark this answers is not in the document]
yunnydang: Thanks Joe, will take a closer look at this!

rachelkirk (approved):
Looks great, thanks for making those adjustments!

jcreiff (approved):

lib/active_merchant/billing/gateways/element.rb
```diff
@@ -222,16 +224,44 @@ def market_code(money, options)
...
+              end
+            end
+          end
+        end
+      end
+
       def add_terminal(xml, options)
         xml.terminal do
           xml.TerminalID options[:terminal_id] || '01'
+          xml.TerminalType options[:terminal_type] if options[:terminal_type]
```
jcreiff, line 256: Just wondering, since TerminalType and PartialApprovedFlag are the only net new fields here in this block, do they need to have defaults like the others? Do we know what happens if they are not provided in the request?

lib/active_merchant/billing/gateways/element.rb
```diff
@@ -222,16 +224,44 @@ def market_code(money, options)
...
+                xml.LodgingCheckInDate lodging[:check_in_date] if lodging[:check_in_date]
+                xml.LodgingCheckOutDate lodging[:check_out_date] if lodging[:check_out_date]
+                xml.LodgingRoomAmount lodging[:room_amount] if lodging[:room_amount]
+                xml.LodgingRoomTax lodging[:room_tax] if lodging[:room_tax]
+                xml.LodgingNoShowIndicator lodging[:no_show_indicator] if lodging[:no_show_indicator]
+                xml.LodgingDuration lodging[:duration] if lodging[:duration]
+                xml.LodgingCustomerName lodging[:customer_name] if lodging[:customer_name]
+                xml.LodgingClientCode lodging[:client_code] if lodging[:client_code]
+                xml.LodgingExtraChargesDetail lodging[:extra_charges_detail] if lodging[:extra_charges_detail]
+                xml.LodgingExtraChargesAmounts lodging[:extra_charges_amount] if lodging[:extra_charges_amount]
```
jcreiff, line 243: Small detail here, but this is the only field where the snake case name doesn't exactly match the Pascal case name — I would vote for making it `:extra_charges_amounts`

lib/active_merchant/billing/gateways/element.rb
jcreiff: ...in retrospect I realize this was perhaps a silly question, since the fields have not been present in the request before now 😁 

Do we have any indication of _when_ they should be used that we could document somewhere? Do  we have the ability to test adding the partial approved flag?

yunnydang: hey @jcreiff yep! So I had to add a couple of new fields because the lodging fields require them. The sandbox is pretty limited but these specifics would need to be documented.

lib/active_merchant/billing/gateways/element.rb
yunnydang: done!

PR #4815, opened 2023-06-20
Adyen:  Add additional data for airline and lodging

javierpedrozaing:
Description
-------------------------
GWS-67

This commit adds additional data for Adyen in order to be able to send information regarding the airline and lodging. To send this new data, it sends fields additional_data_airline and additional_data_lodging as a GSF.

Unit test
-------------------------
Finished in 0.157969 seconds.
109 tests, 567 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
690.01 tests/s, 3589.31 assertions/s

Remote test
-------------------------
Finished in 202.708099 seconds.
134 tests, 447 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 91.0448% passed
0.66 tests/s, 2.21 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -68,6 +68,8 @@ module ActiveMerchant #:nodoc:
         add_application_info(post, options)
         add_level_2_data(post, options)
         add_level_3_data(post, options)
+        add_data_airline(post, options)
+        add_data_lodging(post, options)
         commit('authorise', post, options)
       end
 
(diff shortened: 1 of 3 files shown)


sinourain (approved):
LGTM no comments

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -291,6 +293,81 @@ def add_level_3_data(post, options)
...
+          passenger_data = {
+            'airline.passenger.date_of_birth': 'date_of_birth',
+            'airline.passenger.first_name': 'first_name',
+            'airline.passenger.last_name': 'last_name',
+            'airline.passenger.telephone_number': 'telephone_number',
+            'airline.passenger.traveller_type': 'traveller_type'
+          }
+          post[:additionalData].merge!(extract_and_transform(passenger_data, options[:additional_data_airline][:passenger]))
+        end
+        post[:additionalData].compact!
```
Heavyblade, line 344: ⚠️ Warning:

`compact!` is a tricky method if all the keys on the has had values it returns nil:

```bash
irb(main):001:0> {uno: 1}.compact!
=> nil
irb(main):002:0> {uno: 1, dos: nil}.compact!
=> {:uno=>1}
```

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -1015,11 +1015,9 @@ def test_successful_store_with_elo_card
     assert_equal 'Authorised', response.message
   end
 
-  # Adyen does not currently support recurring transactions with Cabal cards
   def test_failed_store_with_cabal_card
```
Heavyblade, line 1018: 💬 Comment:

My understanding of this change is that Adyen now supports cabal cc, if that is the case the test title is wrong because you are asserting a success.

test/unit/gateways/adyen_test.rb
```diff
@@ -1332,6 +1332,80 @@ def test_level_3_data
...
+        last_name: 'Doe'
+      }
+    }
+
+    response = stub_comms do
+      @gateway.authorize(@amount, @credit_card, @options.merge(additional_data_airline: airline_data))
+    end.check_request do |_endpoint, data, _headers|
+      parsed = JSON.parse(data)
+      additional_data = parsed['additionalData']
+      assert_equal additional_data['airline.agency_invoice_number'], airline_data[:agency_invoice_number]
```
Heavyblade, line 1359: 💬 Comment:

I know that is a little bit annoying but this test is the only checkpoint that verifies that airline data mapping is correct (gateway is not going to tell you if something is missing), so I think Is a good idea to include all the fields.

lib/active_merchant/billing/gateways/adyen.rb
javierpedrozaing: I think that could be ok since we don’t want to save nil values

javierpedrozaing:
> @javierpedrozaing lot of work there, I leave you some specific comments, and going to leave a question here:
> 
> In this PR we are treating airline data as separate data from level III data which is product-specific data, shouldn't this "live" inside level III ?

Yes, that’s because for [Adyen](https://docs.adyen.com/api-explorer/Checkout/latest/post/payments#request-additionalData-AdditionalDataLevel23), airline and lodging data are separated object of Level 2/3.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -291,6 +293,82 @@ def add_level_3_data(post, options)
...
+            telephone_number
+            traveller_type
+          ].each_with_object({}) { |value, hash| hash["airline.passenger.#{value}"] = value }
+
+          post[:additionalData].merge!(extract_and_transform(passenger_data, options[:additional_data_airline][:passenger]))
+        end
+        post[:additionalData].compact!
+      end
+
+      def add_data_lodging(post, options)
```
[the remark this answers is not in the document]
javierpedrozaing: Thank you, @almalee24 good catch, I have not included these fields because seems to be not part of the `lodging` object, instead they are into a `travelEntertainmentAuthData` but any way I don't think it would hurt to include these fields as well.

almalee24 (approved):
Lgtm!

PR #4817, opened 2023-06-26
CyberSource: update NT methods to allow for recurring AP

rachelkirk:
CER-701

According to CyberSource, the cryptogram should be eliminated from the request and stored credentials should be set to merchant and unscheduled. Also in order for this to work, commerce_indicator must be `internet`.

Remote Tests:
122 tests, 609 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 95.082% passed

Unit Tests:
132 tests, 628 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
* Didn’t add a unit test but I can if needed

Local Tests:
5541 tests, 77545 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -799,23 +799,31 @@ module ActiveMerchant #:nodoc:
 
         brand = card_brand(payment_method).to_sym
 
+        # stored_credential_overrides is not documented on the gateway guide in docs
+        # I think the easiest way to solve for apple pay recurring is to create a new field within the hash
+        # commerce_indicator can be passed in as 'internet' but is assigned that value in an earlier method so will be overwritten by the card brand if there is not something done within this method
+        commerce_indicator = 'internet' if options.dig(:stored_credential_overrides, :type) == 'apple_pay'
+
         case brand
         when :visa
           xml.tag! 'ccAuthService', { 'run' => 'true' } do
-            xml.tag!('cavv', payment_method.payment_cryptogram)
-            xml.tag!('commerceIndicator', ECI_BRAND_MAPPING[brand])
-            xml.tag!('xid', payment_method.payment_cryptogram)
+            # these guard clauses need more consideration but work for the needs of this edge case remote test
+            xml.tag!('cavv', payment_method.payment_cryptogram) unless options[:stored_credential_overrides]
+            xml.commerceIndicator commerce_indicator.nil? ? ECI_BRAND_MAPPING[brand] : commerce_indicator
+            xml.tag!('xid', payment_method.payment_cryptogram) unless options[:stored_credential_overrides]
             xml.tag!('reconciliationID', options[:reconciliation_id]) if options[:reconciliation_id]
           end
         when :master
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -799,23 +799,31 @@ def add_auth_network_tokenization(xml, payment_method, options)
 
         brand = card_brand(payment_method).to_sym
 
+        # stored_credential_overrides is not documented on the gateway guide in docs
+        # I think the easiest way to solve for apple pay recurring is to create a new field within the hash
+        # commerce_indicator can be passed in as 'internet' but is assigned that value in an earlier method so will be overwritten by the card brand if there is not something done within this method
+        commerce_indicator = 'internet' if options.dig(:stored_credential_overrides, :type) == 'apple_pay'
```
aenand, line 805: Are you envisioning the merchant passing in this GSF or adding when we build the params in the gateway class?

aenand:
@rachelkirk do we know if this also works with GP? If so would a solution of checking `initiator: merchant && initial_transaction: false && payment_method.NetworkTokenizationCreditCard?` to determine if we need to engage this new stored creds flow work? The reason I ask is because it seems we are only relying on the `stored_credential_overrides` GSF, which merchants provide, to determine if it's a Apple Pay payment method.

lib/active_merchant/billing/gateways/cyber_source.rb
rachelkirk: That was a thought, but this was more of a means to an end to see if I could get a passing remote test. I'm open to other options.

BritneyS (commented):
At first I wasn't sure if using only the token but not cryptogram was accurate to how a digital wallet was intended to be used. But it looks like the token is meant to be used for recurring payments, or any payment not originating on the device ([source](https://www.engadget.com/2014-10-02-apple-pay-an-in-depth-look-at-whats-behind-the-secure-payment.html#:~:text=A%202012%20white,existing%20business%20processes.)). I just had a question about if we should be localizing when we remove the cryptogram.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -799,23 +799,31 @@ def add_auth_network_tokenization(xml, payment_method, options)
...
+
         case brand
         when :visa
           xml.tag! 'ccAuthService', { 'run' => 'true' } do
-            xml.tag!('cavv', payment_method.payment_cryptogram)
```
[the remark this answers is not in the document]
rachelkirk: Only for recurring Apple Pay transactions. That's why I said the guard clauses should probably be revisited, but this was more of a draft/means to get getting a passing test.

rachelkirk:
> 
I don't think Google Pay is supported on CyberSource, or at least it's not listed in docs. 
The intent behind using the GSF was so that I could quickly get the behavior and honestly the stored credentials methods in CyberSource are not super great. They are confusing and could do with a refactor but I know that some customers are using this current flow. Your suggestion is certainly more elegant and thorough.

PR #4822, opened 2023-07-03
CheckoutV2: Parse AVS and CVV checks

aenand:
The CheckoutV2 gateway had logic in the response parsing method that would limit the scope of parsing to only authorize or purchase. This presents an issue for merchants using `verify-payment` or other methods that may have the AVS and CVV checks in the response.

This commit also updates the AVS and CVV checks to use `dig` to safely try parsing out the values

Test Summary
Remote:
93 tests, 227 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -363,9 +363,6 @@ module ActiveMerchant #:nodoc:
       end
 
       def response(action, succeeded, response, source_id = nil)
-        successful_response = succeeded && action == :purchase || action == :authorize
-        avs_result = successful_response ? avs_result(response) : nil
-        cvv_result = successful_response ? cvv_result(response) : nil
         authorization = authorization_from(response) unless action == :unstore
         body = action == :unstore ? { response_code: response.to_s } : response
         Response.new(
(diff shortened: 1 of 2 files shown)


aenand:
I felt comfortable not adding a test here since we already have tests to check for AVS and CVV results that passed. Is there another scenario people think I should write a test for?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -427,11 +424,11 @@ def base_url
       end
 
       def avs_result(response)
-        response['source'] && response['source']['avs_check'] ? AVSResult.new(code: response['source']['avs_check']) : nil
+        response.respond_to?(:dig) && response.dig('source', 'avs_check') ? AVSResult.new(code: response['source']['avs_check']) : nil
```
aenand, line 427: Checking the `respond_to` since the response is an integer if the action is unstore. Question for reviewers: do you feel more comfortable with this handling or based on action type?

BritneyS: I think this implementation is fine. A comment wouldn't hurt, but that's up to you

BritneyS (approved):
LGTM 👍🏾

PR #4828, opened 2023-07-11
Stripe PI: Gate sending NTID

almalee24:
Don't send NTID in add_stored_credential if
post[:payment_method_options][:card][:stored_credential_transaction_type] = 'setup_on_session' and setup_future_usage=off_session.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -418,63 +418,73 @@ module ActiveMerchant #:nodoc:
       # the existing logic by default. To be able to utilize this field, you must reach out to Stripe.
 
       def add_stored_credentials(post, options = {})
-        return unless options[:stored_credential] && !options[:stored_credential].values.all?(&:nil?)
+        stored_credential = options[:stored_credential]
+        return unless stored_credential && !stored_credential.values.all?(&:nil?)
 
         post[:payment_method_options] ||= {}
         post[:payment_method_options][:card] ||= {}
-        add_stored_credential_transaction_type(post, options) if options[:stored_credential_transaction_type]
 
-        stored_credential = options[:stored_credential]
-        post[:payment_method_options][:card][:mit_exemption] = {}
+        card_options = post[:payment_method_options][:card]
+        card_options[:mit_exemption] = {}
 
         # Stripe PI accepts network_transaction_id and ds_transaction_id via mit field under card.
         # The network_transaction_id can be sent in nested under stored credentials OR as its own field (add_ntid handles when it is sent in on its own)
         # If it is sent is as its own field AND under stored credentials, the value sent under its own field is what will send.
-        post[:payment_method_options][:card][:mit_exemption][:ds_transaction_id] = stored_credential[:ds_transaction_id] if stored_credential[:ds_transaction_id]
-        post[:payment_method_options][:card][:mit_exemption][:network_transaction_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
+        card_options[:mit_exemption][:ds_transaction_id] = stored_credential[:ds_transaction_id] if stored_credential[:ds_transaction_id]
+        unless options[:setup_future_usage] == 'off_session'
+          card_options[:mit_exemption][:network_transaction_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
+        end
+
+        add_stored_credential_transaction_type(post, options)
       end
 
       def add_stored_credential_transaction_type(post, options = {})
+        return unless options[:stored_credential_transaction_type]
+
         stored_credential = options[:stored_credential]
         # Do not add anything unless these are present.
         return unless stored_credential[:reason_type] && stored_credential[:initiator]
 
         # Not compatible with off_session parameter.
         options.delete(:off_session)
-        if stored_credential[:initial_transaction]
-          # Initial transactions must by CIT
-          return unless stored_credential[:initiator] == 'cardholder'
 
-          initial_transaction_stored_credential(post, stored_credential[:reason_type])
-        else
-          # Subsequent transaction
-          subsequent_transaction_stored_credential(post, stored_credential[:initiator], stored_credential[:reason_type])
-        end
+        stored_credential_type = if stored_credential[:initial_transaction]
+                                   return unless stored_credential[:initiator] == 'cardholder'
+
+                                   initial_transaction_stored_credential(post, stored_credential)
+                                 else
+                                   subsequent_transaction_stored_credential(post, stored_credential)
+                                 end
+
+        card_options = post[:payment_method_options][:card]
+        card_options[:stored_credential_transaction_type] = stored_credential_type
+        card_options[:mit_exemption].delete(:network_transaction_id) if stored_credential_type == 'setup_on_session'
       end
 
-      def initial_transaction_stored_credential(post, reason_type)
-        if reason_type == 'unscheduled'
+      def initial_transaction_stored_credential(post, stored_credential)
+        case stored_credential[:reason_type]
+        when 'unscheduled'
           # Charge on-session and store card for future one-off payment use
-          post[:payment_method_options][:card][:stored_credential_transaction_type] = 'setup_off_session_unscheduled'
-        elsif reason_type == 'recurring'
+          'setup_off_session_unscheduled'
+        when 'recurring'
           # Charge on-session and store card for future recurring payment use
-          post[:payment_method_options][:card][:stored_credential_transaction_type] = 'setup_off_session_recurring'
+          'setup_off_session_recurring'
         else
           # Charge on-session and store card for future on-session payment use.
-          post[:payment_method_options][:card][:stored_credential_transaction_type] = 'setup_on_session'
+          'setup_on_session'
         end
       end
 
-      def subsequent_transaction_stored_credential(post, initiator, reason_type)
-        if initiator == 'cardholder'
+      def subsequent_transaction_stored_credential(post, stored_credential)
+        if stored_credential[:initiator] == 'cardholder'
           # Charge on-session customer using previously stored card.
-          post[:payment_method_options][:card][:stored_credential_transaction_type] = 'stored_on_session'
-        elsif reason_type == 'recurring'
+          'stored_on_session'
+        elsif stored_credential[:reason_type] == 'recurring'
           # Charge off-session customer using previously stored card for recurring transaction
-          post[:payment_method_options][:card][:stored_credential_transaction_type] = 'stored_off_session_recurring'
+          'stored_off_session_recurring'
         else
           # Charge off-session customer using previously stored card for one-off transaction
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -418,74 +418,73 @@ def add_exemption(post, options = {})
...
-        post[:payment_method_options][:card][:mit_exemption][:ds_transaction_id] = stored_credential[:ds_transaction_id] if stored_credential[:ds_transaction_id]
-        post[:payment_method_options][:card][:mit_exemption][:network_transaction_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
-      end
+        card_options[:mit_exemption][:ds_transaction_id] = stored_credential[:ds_transaction_id] if stored_credential[:ds_transaction_id]
 
-      def add_stored_credential_transaction_type(post, options = {})
-        stored_credential = options[:stored_credential]
-        # Do not add anything unless these are present.
-        return unless stored_credential[:reason_type] && stored_credential[:initiator]
+        return unless options[:stored_credential_transaction_type] && %i(reason_type initiator).all? { |k| stored_credential.key?(k) }
```
aenand, line 434: We should only engage the `initial_transaction_stored_credential` and `subsequent_transaction_stored_credential` methods in flows where `stored_credential_transaction_type` is passed down as `true`. This commit seems to engage them in all cases. 

Also we should be able to add a NTID when `stored_credential_transaction_type` is `false`.

aenand (changes requested):

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: Okay, thanks for clarifying that wasn't very clear when looking at the ticket. Will change it

aenand: Ah my bad! Basically the context for the change is that we are trying to add a new field for Stripe stored creds but we cannot turn it on for everyone since it requires a configuration on the merchant side. Therefore we should only add this new "stored_credential_transaction_type" field if the merchant passes `stored_credential_transaction_type` as `true`. 

In testing it I found that Stripe fails transactions with an NTID in 2 scenarios:

1. When `stored_credential_transaction_type` is engaged AND the mapping logic leads to us setting `stored_credential_transaction_type == 'setup_on_session'`. 
2. When `stored_credential_transaction_type` is not engaged AND `setup_future_usage= 'off_session'`. 

This 2nd scenario is currently preventing us from being able to pass the standard stored credential fields (initial_transaction, initiator, etc) for CIT. I see you have the guard for the first case but we need the guard for the second case as well

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -418,74 +418,72 @@ def add_exemption(post, options = {})
...
 
-          initial_transaction_stored_credential(post, stored_credential[:reason_type])
-        else
-          # Subsequent transaction
-          subsequent_transaction_stored_credential(post, stored_credential[:initiator], stored_credential[:reason_type])
-        end
+        stored_credential_type = if stored_credential[:initial_transaction] && stored_credential[:initiator] == 'cardholder'
+                                   initial_transaction_stored_credential(post, stored_credential)
+                                 else
+                                   subsequent_transaction_stored_credential(post, stored_credential)
```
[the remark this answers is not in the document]
almalee24: I'm not sure I'm following but do you mean [this line was removed in my changes](https://github.com/activemerchant/active_merchant/pull/4828/files#diff-19492101b546175aa3ca6a16ec7d0a610d0adae03cbaf130eff54092dec8e0fbL440). I was thinking yesterday just to go ahead an remove all the extra code change I did and just got back to how it was before. Of course just adding the NTID changes.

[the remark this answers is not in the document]
aenand: Nope, I was talking about this [line](https://github.com/activemerchant/active_merchant/pull/4828/files#diff-19492101b546175aa3ca6a16ec7d0a610d0adae03cbaf130eff54092dec8e0fbL446)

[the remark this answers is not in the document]
almalee24: Does [this line](https://github.com/activemerchant/active_merchant/pull/4828/files#diff-19492101b546175aa3ca6a16ec7d0a610d0adae03cbaf130eff54092dec8e0fbR440) not do the same thing?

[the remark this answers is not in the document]
aenand: I don't think so because if `initial_transaction ` is true and stored_credential[:initiator] == 'merchant' it will go into the else statement rather than return.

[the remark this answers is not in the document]
almalee24: Oh I see what you're saying, let me fix that. So we only set `:stored_credential_transaction_type` for `stored_credential[:initiator] = 'cardholder'`?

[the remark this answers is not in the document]
aenand: No, if `initial_transaction == true` then `stored_credential[:initiator]` must be set to `'cardholder'` in order to set the `stored_credential_transaction_type`. There are some valid combinations for `stored_credential[:initiator]` when set to `'merchant'`. 

Sorry if I'm explaining this bad, it's just a tangled web of logic. I can write all the scenarios if that's helpful!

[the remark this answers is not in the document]
almalee24: I get it, I realized what you were saying after I read your earlier comment again. It can be confusing.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -418,74 +418,83 @@ def add_exemption(post, options = {})
...
         else
           # Charge off-session customer using previously stored card for one-off transaction
-          post[:payment_method_options][:card][:stored_credential_transaction_type] = 'stored_off_session_unscheduled'
+          'stored_off_session_unscheduled'
         end
       end
 
       def add_ntid(post, options = {})
         return unless options[:network_transaction_id]
+        return if options[:setup_future_usage] == 'off_session'
```
aenand, line 491: This method is used when someone passes the NTID via the GSF override. I should move this check from `add_ntid` to `add_stored_credentials` so that if we send down `setup_future_usage ==  'off_session'` then it's a CIT and don't add any NTID flagging. What do you think?

almalee24: I think we should keep it as is for now. What if a customer just passes NTID but nothing else required by `add_stored_credentials`.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -418,63 +418,71 @@ def add_exemption(post, options = {})
...
+                                   return unless stored_credential[:initiator] == 'cardholder'
+
+                                   initial_transaction_stored_credential(post, stored_credential)
+                                 else
+                                   subsequent_transaction_stored_credential(post, stored_credential)
+                                 end
+
+        card_options = post[:payment_method_options][:card]
+        card_options[:stored_credential_transaction_type] = stored_credential_type
+        card_options[:mit_exemption].delete(:network_transaction_id) if stored_credential_type == 'setup_on_session' || options[:setup_future_usage] == 'off_session'
```
aenand, line 459: Would this cover the scenario where we do not send the NTID if `stored_credential_transaction_type` is not passed in AND an NTID is passed in AND `setup_future_usage == off_session`?

test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -329,6 +329,31 @@ def test_successful_purchase_with_level3_data
...
+            network_transaction_id: network_transaction_id, # TEST env seems happy with any value :/
+            ds_transaction_id: 'null' # this is optional and can be null if not available.
+          }
+        })
+      end.check_request do |_method, _endpoint, data, _headers|
+        assert_no_match(%r{payment_method_options\[card\]\[mit_exemption\]\[network_transaction_id\]=}, data)
+        assert_match(%r{payment_method_options\[card\]\[mit_exemption\]\[ds_transaction_id\]=null}, data)
+      end.respond_with(successful_create_intent_response)
+    end
+  end
```
aenand, line 355: Can we add another test like this except `setup_future_usage = true` and check that the NTID is not sent?

aenand (changes requested):

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: I thought you said that we wanted to send NTID even if `stored_credential_transaciton_type` is not passed, https://github.com/activemerchant/active_merchant/pull/4828#discussion_r1260162091? At least I interpreted at `stored_credential_transaciton_type: false` being the same thing as it not being passed. 

Let me see if I'm understanding correctly we DO NOT send NTID if 
1. `setup_future_usage == off_session`
2.  `stored_credential_transaction_type ==  setup_on_session`
3.  options doesn't have [:stored_credential_transaction_type]

aenand: Ahhh I see the confusion, it's scenarios 1 and 2 in that list.

If `stored_credential_transaction_type` doesn't exist in options then we should not add `stored_credential_transaction_type` to the `post` hash. 

For context, the reason we want this change is so that we can send our standard stored credential fields down every time. Right now we only do it on MITs because there's no guard on scenario 1 in AM. There will be a FF gating `stored_credential_transaction_type` and when that is turned on, we'll not pass `setup_future_usage` and instead rely on our stored credential mapping to pick the correct `stored_credential_transaction_type`

test/unit/gateways/stripe_payment_intents_test.rb
aenand: wait sorry should it be `setup_future_usage = 'off_session'`?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: Okay I think now we're on the same page. We don't add `stored_credential_transaction_type` to post if it doesn't exist in options https://github.com/activemerchant/active_merchant/pull/4828/files#diff-19492101b546175aa3ca6a16ec7d0a610d0adae03cbaf130eff54092dec8e0fbR440. But I'll move the `options[:setup_future_usage] == 'off_session'` from here to the other method and add a test for it. Thanks for all the help!

aenand (approved):
LGTM 🚀 Nice work on this

test/unit/gateways/stripe_payment_intents_test.rb
almalee24: Yup that's what I did, I think read your other comment as `setup_future_usage = 'off_session'`

PR #4831, opened 2023-07-17
Nuvie/SafeCharge: Add unreferenced refund field

yunnydang:
Note: 
I didn't include a unit test because this fields purpose is to exclude sending the sg_TransactionID and not meant to return anything. 

Local:
5545 tests, 77509 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6573% passed

Unit:
25 tests, 144 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
32 tests, 89 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/safe_charge.rb
+++ b/lib/active_merchant/billing/gateways/safe_charge.rb
@@ -73,10 +73,10 @@ module ActiveMerchant #:nodoc:
         add_transaction_data('Credit', post, money, options.merge!({ currency: original_currency }))
         post[:sg_CreditType] = 2
         post[:sg_AuthCode] = auth
-        post[:sg_TransactionID] = transaction_id
         post[:sg_CCToken] = token
         post[:sg_ExpMonth] = exp_month
         post[:sg_ExpYear] = exp_year
+        post[:sg_TransactionID] = transaction_id unless options[:unreferenced_refund]
 
         commit(post)
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/safe_charge.rb
```diff
@@ -73,12 +73,16 @@ def refund(money, authorization, options = {})
         add_transaction_data('Credit', post, money, options.merge!({ currency: original_currency }))
         post[:sg_CreditType] = 2
         post[:sg_AuthCode] = auth
-        post[:sg_TransactionID] = transaction_id
         post[:sg_CCToken] = token
         post[:sg_ExpMonth] = exp_month
         post[:sg_ExpYear] = exp_year
 
-        commit(post)
+        if options[:unreferenced_refund] == true
```
jcreiff, line 80: Instead of duplicating the `commit(post)` statement with this logic, could you instead add an `unless` to the end of line 76?

`post[:sg_TransactionID] = transaction_id unless options[:unreferenced_refund]`

I think this would make the change a little bit more DRY

jcreiff (changes requested):
One small suggestion for refactoring -- I also think you could add a unit test here; the test expectation would be _refuting_ that the `sg_TransactionID` is present in the request when `unreferenced_refund` is present in the `options`

lib/active_merchant/billing/gateways/safe_charge.rb
yunnydang: yep I totally agree! I was completely brain fogged and was already fixing it when the linter failed lol, thanks!

yunnydang:
> _refuting_ that the `sg_TransactionID` is present in the request when `unreferenced_refund` is present in the `options`

I was playing around with trying to do something similar to this but didnt get far. After stepping away for a bit, I figured it out, thanks!

jcreiff (approved):
LGTM!

PR #4832, opened 2023-07-18
Kushki: Enable 3ds2

jherreraa:
## Summary:

Enable 3ds version 2 on the gateway above

SER-625

## Remote Tests

Finished in 77.484256 seconds.
22 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

## Unit Tests
Finished in 0.019977 seconds.
17 tests, 109 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/kushki.rb
+++ b/lib/active_merchant/billing/gateways/kushki.rb
@@ -20,14 +20,14 @@ module ActiveMerchant #:nodoc:
       def purchase(amount, payment_method, options = {})
         MultiResponse.run() do |r|
           r.process { tokenize(amount, payment_method, options) }
-          r.process { charge(amount, r.authorization, options) }
+          r.process { charge(amount, r.authorization, options, payment_method) }
         end
       end
 
       def authorize(amount, payment_method, options = {})
         MultiResponse.run() do |r|
           r.process { tokenize(amount, payment_method, options) }
-          r.process { preauthorize(amount, r.authorization, options) }
+          r.process { preauthorize(amount, r.authorization, options, payment_method) }
         end
       end
 
@@ -89,7 +89,7 @@ module ActiveMerchant #:nodoc:
         commit(action, post)
       end
 
-      def charge(amount, authorization, options)
+      def charge(amount, authorization, options, payment_method = {})
         action = 'charge'
 
         post = {}
@@ -100,11 +100,12 @@ module ActiveMerchant #:nodoc:
         add_metadata(post, options)
         add_months(post, options)
         add_deferred(post, options)
+        add_three_d_secure(post, payment_method, options)
 
         commit(action, post)
       end
 
-      def preauthorize(amount, authorization, options)
+      def preauthorize(amount, authorization, options, payment_method = {})
         action = 'preAuthorization'
 
         post = {}
@@ -114,6 +115,7 @@ module ActiveMerchant #:nodoc:
         add_metadata(post, options)
         add_months(post, options)
         add_deferred(post, options)
+        add_three_d_secure(post, payment_method, options)
 
         commit(action, post)
       end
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -204,6 +206,33 @@ def add_deferred(post, options)
         }
       end
 
+      def add_three_d_secure(post, payment_method, options)
+        three_d_secure = options[:three_d_secure]
+        return unless three_d_secure
```
Heavyblade, line 211: 💬 Comment:

For extra security and standards ensure the guard with `three_d_secure.present?`

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -204,6 +206,33 @@ def add_deferred(post, options)
...
+
+        post[:threeDomainSecure] = {}
+        post[:threeDomainSecure][:eci] = three_d_secure[:eci]
+        post[:threeDomainSecure][:specificationVersion] = three_d_secure[:version]
```
Heavyblade, line 215: 💬 Comment:

Just matter of taste but initialization could be more declarative instead of procedural:

```ruby

post[:threeDomainSecure] = {
  eci: three_d_secure[:eci],
  specificationVersion: three_d_secure[:version]
}
```

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -204,6 +206,33 @@ def add_deferred(post, options)
...
+
+        if payment_method.brand == 'mastercard'
```
Heavyblade, line 217: ❓ Question:

Maybe my memory is failing me but Kushki only considers 3DS for Visa and MasterCard right?, it that is the case what should be the behavior with the rest of the supported card brands?

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -204,6 +206,33 @@ def add_deferred(post, options)
...
+          post[:threeDomainSecure][:acceptRisk] = three_d_secure[:eci] == '00'
+          post[:threeDomainSecure][:ucaf] = three_d_secure[:cavv]
+          post[:threeDomainSecure][:directoryServerTransactionID] = three_d_secure[:ds_transaction_id] if three_d_secure[:version].to_f > 2.0
```
Heavyblade, line 220: ❓ Question:

What should happen if the brand is Mastercard but 3DS version is 1 ?

💬 Comment:

maybe, just to be sure comparison should be `>=` just in case someone sends just 2 and not 2.1 or 2.2

Heavyblade (commented):
Looks pretty much OK @jherreraa leaved some comments and remember to add jira Id to the PR description and commit message.

lib/active_merchant/billing/gateways/kushki.rb
jherreraa: Thanks a lot for the suggestion.

lib/active_merchant/billing/gateways/kushki.rb
jherreraa: Noted! Thanks!

lib/active_merchant/billing/gateways/kushki.rb
jherreraa: Checking tests, 3ds version has no effect in the request or response, statement corrected.

lib/active_merchant/billing/gateways/kushki.rb
jherreraa: after tests... the behaviour is the same as failed request with code k001

Heavyblade: Mmm maybe should we consider asking what is the expected behavior ?. if we just let the request fail debugging for a merchant and support is going to be a nightmare, maybe we should raise an Argument exception here or return a failure response with the proper message telling that 3DS is not supported for that brand?

javierpedrozaing (commented):
LGTM, just I left a small comment

lib/active_merchant/billing/gateways/kushki.rb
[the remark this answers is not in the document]
jherreraa: in the most recent correction this was removed...

lib/active_merchant/billing/gateways/kushki.rb
jherreraa: I have checked the docs in order to find some other way to handle this but I could not find a more proper way so I add the ArgumentError into the implementation... Hope this is a better way to deal with this.
Thank you for the suggestion!

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -204,6 +206,36 @@ def add_deferred(post, options)
...
+            post[:threeDomainSecure][:collectionIndicator] = '1'
+          else
+            post[:threeDomainSecure][:collectionIndicator] = '2'
+          end
+        elsif payment_method.brand == 'visa'
+          post[:threeDomainSecure][:acceptRisk] = three_d_secure[:eci] == '07'
+          post[:threeDomainSecure][:cavv] = three_d_secure[:cavv]
+          post[:threeDomainSecure][:xid] = three_d_secure[:xid]
+        else
+          raise ArgumentError.new '3ds2 authentication only applies for Visa and Mastercard brands.'
```
curiousepic, line 235:
```suggestion
          raise ArgumentError.new 'Kushki supports 3ds2 authentication for only Visa and Mastercard brands.'
```

jherreraa: Thanks for the suggestion! Added 👍

curiousepic:
Merged in d9c926af1a3972c25a5ef67b572b3c30d5a1a2af

PR #4836, opened 2023-07-20
CyberSource: remove credentials from tests

bbraschi:


fmarois87 (approved):

lunataoo (approved):
extra spaces nit 👀 but should be ok

bbraschi:
TY both! I added the extra space in the a previous commit.

PR #4838, opened 2023-07-21
Rapyd: Add customer object and fix tests

aenand:
The Rapyd gateway requires the Customer object on multiple payment types now and is optional on all. This commit adds teh customer subhash to requests and updates the remote tests to pass

Remote:
32 tests, 76 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 78.125% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -63,7 +63,7 @@ module ActiveMerchant #:nodoc:
       def store(payment, options = {})
         post = {}
         add_payment(post, payment, options)
-        add_customer_object(post, payment, options)
+        add_customer_data(post, payment, options, 'store')
         add_metadata(post, options)
         add_ewallet(post, options)
         add_payment_fields(post, options)
@@ -100,14 +100,13 @@ module ActiveMerchant #:nodoc:
       def add_auth_purchase(post, money, payment, options)
         add_invoice(post, money, options)
         add_payment(post, payment, options)
-        add_customer_object(post, payment, options)
+        add_customer_data(post, payment, options)
         add_3ds(post, payment, options)
         add_address(post, payment, options)
         add_metadata(post, options)
         add_ewallet(post, options)
         add_payment_fields(post, options)
         add_payment_urls(post, options)
-        add_customer_id(post, options)
       end
 
       def add_address(post, creditcard, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -214,11 +218,29 @@ def add_payment_urls(post, options)
...
+        formatted_address[:city] = address[:city] if address[:city]
+        formatted_address[:state] = address[:state] if address[:state]
+        formatted_address[:country] = address[:country] if address[:country]
+        formatted_address[:zip] = address[:zip] if address[:zip]
+        formatted_address[:phone_number] = address[:phone].gsub(/\D/, '') if address[:phone]
+
+        formatted_address
       end
 
       def add_customer_id(post, options)
```
BritneyS, line 246: `add_customer_id` sets `post[:customer] = options[:customer_id] if options[:customer_id]`. So in places where it's called after `add_customer_object` is called, like `add_auth_purchase`, it would currently override the customer object built in `add_customer_object`. Is that expected? If so, I would recommend adding a guard to `add_customer_object` to not bother building the customer object unless there's a `customer_id` option present.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -63,7 +63,11 @@ def verify(credit_card, options = {})
       def store(payment, options = {})
         post = {}
         add_payment(post, payment, options)
-        add_customer_object(post, payment, options)
+        phone = options.dig(:billing_address, :phone) .gsub(/\D/, '') unless options[:billing_address].nil?
```
BritneyS, line 66: Can this live in a method for adding top level contact info and called where appropriate, including `add_customer_object` ? It looks like the only difference is where the `name` is passed, but it looks like everything else is the same.

BritneyS (changes requested):
Nice! Just had some concerns about field state and some code repetition

lib/active_merchant/billing/gateways/rapyd.rb
aenand: I'll have to think about where to put this. The issue is that in the `store` call it's top level on the `post` hash but in other calls it's in `post[:customer]` subhash

lib/active_merchant/billing/gateways/rapyd.rb
aenand: Thanks for calling this out! It seems that `customer` and `customer_id` usage has changed. We should not expect a `customer` in options but instead a possible `customer_id` in options.

lib/active_merchant/billing/gateways/rapyd.rb
aenand: Attempted a refactor to make it more clean and split it out into another commit. Let me know what you think!

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -217,13 +212,25 @@ def add_payment_urls(post, options)
...
       end
 
-      def add_customer_object(post, payment, options)
-        phone = options.dig(:billing_address, :phone) .gsub(/\D/, '') unless options[:billing_address].nil?
-        post[:phone_number] = phone || options.dig(:customer, :phone_number)
-        post[:email] = options[:email] || options.dig(:customer, :email)
-        post[:customer] = {}
-        post[:customer][:name] = "#{payment.first_name} #{payment.last_name}" unless payment.is_a?(String)
-        post[:customer][:addresses] = [address(options)]
+      def add_customer_object(post, payment, options, action = '')
```
BritneyS, line 215: Since now we may or may not be creating a `customer` hash in this method, I think this method can be renamed to something like `add_customer_data`.

BritneyS (commented):
I like this, since it calls out the different use-cases directly, all in one place. I would just reconsider the naming with these changes.

test/unit/gateways/rapyd_test.rb
```diff
@@ -226,15 +219,15 @@ def test_failed_purchase_without_customer_object
   end
 
   def test_successful_purchase_with_customer_object
+    require 'pry'
```
BritneyS, line 222: Pesky `pry` line here

BritneyS (approved):
Refactor looks good 🎉 Just take care of the `pry` line before merging.

test/unit/gateways/rapyd_test.rb
aenand: good catch!

PR #4843, opened 2023-07-25
Adyen: Support raw refusal reason

aenand:
ECS-3082

A change was made two months ago to provide the MerchantAdviceCode as the error message if available before falling back to the refusalReasonRaw value. Some merchants do not want this change in messaging so this gives them a way to elect which error message they want

Test Summary
Remote:
11 failing tests on master
134 tests, 447 assertions, 11 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 91.791% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@
 
 == HEAD
 * PaymentExpress: Correct endpoints [steveh] #4827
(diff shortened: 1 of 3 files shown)


almalee24 (approved):
Lgtm!

PR #4844, opened 2023-07-28
Cybersource: Add merchant_id

almalee24:
Set the merchantId filed to options[:merchant_id] instead of @options[:login] if available.

Unit: 136 tests, 641 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -556,7 +556,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_merchant_data(xml, options)
-        xml.tag! 'merchantID', @options[:login]
+        xml.tag! 'merchantID', options[:merchant_id] || @options[:login]
         xml.tag! 'merchantReferenceCode', options[:order_id] || generate_unique_id
         xml.tag! 'clientLibrary', 'Ruby Active Merchant'
         xml.tag! 'clientLibraryVersion', VERSION
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -330,7 +330,7 @@ module ActiveMerchant #:nodoc:
         add_reconciliation_id(post, options)
         add_sec_code(post, options)
         add_invoice_number(post, options)
-        response = parse(ssl_post(url(action), post.to_json, auth_headers(action, post)))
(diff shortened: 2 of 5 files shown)


almalee24:
> This looks good! Can you also make this change to the CyberSource REST gateway?

Yes, I'll make those changes. I just realized that [@options[:merchant_id]](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source_rest.rb#L383) was different than this new field

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -556,7 +556,7 @@ def add_line_item_data(xml, options)
       end
 
       def add_merchant_data(xml, options)
-        xml.tag! 'merchantID', @options[:login]
+        xml.tag! 'merchantID', options[:merchant_id] || @options[:login]
```
BritneyS, line 559: `options` vs `@options` here is interesting. This makes sense, since it looks like `@options` is a hash for "gateway options" and `options` is a hash for "request options", so `merchant_id` is passed after the gateway was originally created.

BritneyS (approved):
Looks good!

aenand (approved):
Nice work!

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -374,13 +374,13 @@ def error_code_from(response)
...
       # https://developer.cybersource.com/docs/cybs/en-us/payments/developer/all/rest/payments/GenerateHeader/httpSignatureAuthentication.html
-      def get_http_signature(resource, digest, http_method = 'post', gmtdatetime = Time.now.httpdate)
+      def get_http_signature(resource, digest, options, http_method = 'post', gmtdatetime = Time.now.httpdate)
         string_to_sign = {
           host: host,
           date: gmtdatetime,
           "(request-target)": "#{http_method} /pts/v2/#{resource}",
           digest: digest,
-          "v-c-merchant-id": @options[:merchant_id]
+          "v-c-merchant-id": options[:merchant_id] || @options[:merchant_id]
```
aenand, line 383: After reviewing the [REST docs](https://developer.cybersource.com/docs/cybs/en-us/security-keys/user/all/ada/security-keys/keys-meta-intro/keys-meta-api-intro.html) I made a mistake in the requirements. I believe we should only override it for the HTTP header, not the Signature.

aenand (changes requested):

aenand (approved):
Nice work!

PR #4851, opened 2023-08-11

almalee24:

### DIFF
--- a/lib/active_merchant/billing/gateways/airwallex.rb
+++ b/lib/active_merchant/billing/gateways/airwallex.rb
@@ -32,7 +32,7 @@ module ActiveMerchant #:nodoc:
         @client_id = options[:client_id]
         @client_api_key = options[:client_api_key]
         super
-        @access_token = setup_access_token
+        @access_token = options[:access_token] || setup_access_token
       end
 
       def purchase(money, card, options = {})
@@ -133,8 +133,14 @@ module ActiveMerchant #:nodoc:
           'x-client-id' => @client_id,
           'x-api-key' => @client_api_key
         }
-        response = ssl_post(build_request_url(:login), nil, token_headers)
-        JSON.parse(response)['token']
+        raw_response = ssl_post(build_request_url(:login), nil, token_headers)
+        response = JSON.parse(raw_response)
+        if (token = response['token'])
+          token
+        else
+          oauth_response = Response.new(false, response['message'])
+          raise OAuthResponseError.new(oauth_response)
+        end
       end
 
       def build_request_url(action, id = nil)
--- a/lib/active_merchant/billing/gateways/alelo.rb
+++ b/lib/active_merchant/billing/gateways/alelo.rb
@@ -110,8 +110,18 @@ module ActiveMerchant #:nodoc:
           'Content-Type' => 'application/x-www-form-urlencoded'
         }
 
-        parsed = parse(ssl_post(url('captura-oauth-provider/oauth/token'), post_data(params), headers))
-        Response.new(true, parsed[:access_token], parsed)
+        begin
+          raw_response = ssl_post(url('captura-oauth-provider/oauth/token'), post_data(params), headers)
+        rescue ResponseError => e
+          raise OAuthResponseError.new(e)
+        else
+          response = parse(raw_response)
+          if (access_token = response[:access_token])
+            Response.new(true, access_token, response)
+          else
+            raise OAuthResponseError.new(response)
+          end
+        end
       end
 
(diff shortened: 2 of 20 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -341,8 +341,14 @@ def access_token_url
...
         request = 'grant_type=client_credentials'
-        response = parse(ssl_post(access_token_url, request, access_token_header))
-        response['access_token']
+        begin
+          raw_response = ssl_post(access_token_url, request, access_token_header)
+        rescue ResponseError => e
+          raise OAuthResponseError.new(e)
+        else
+          response = parse(raw_response)
+          response['access_token']
```
aenand, line 350: Should we raise the `OAuthResponseError` if `response['access_token']` is not present?

lib/active_merchant/billing/gateways/alelo.rb
```diff
@@ -110,8 +110,14 @@ def fetch_access_token
...
 
-        parsed = parse(ssl_post(url('captura-oauth-provider/oauth/token'), post_data(params), headers))
-        Response.new(true, parsed[:access_token], parsed)
+        begin
+          raw_response = ssl_post(url('captura-oauth-provider/oauth/token'), post_data(params), headers)
+        rescue ResponseError => e
+          raise OAuthResponseError.new(e)
+        else
+          parsed = parse(raw_response)
+          Response.new(true, parsed[:access_token], parsed)
```
aenand, line 119: Should we check that `parsed[:access_token` is defined and if it isn't raise an `OAuthResponseError`?

lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -187,10 +187,15 @@ def acquire_access_token
           'Content-Type'      => 'application/x-www-form-urlencoded'
         }
         response = ssl_post(url, data, oauth_headers)
-        json_response = JSON.parse(response)
+        json_response = parse(response)
```
aenand, line 190: Why remove the `JSON.`?

aenand (changes requested):
Would you be able to make some unit tests for this change?

lib/active_merchant/billing/gateways/checkout_v2.rb
almalee24: Yeah that's a good idea, will change it.

lib/active_merchant/billing/gateways/pay_trace.rb
almalee24: Just cause parse does the same thing lol

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -1,3 +1,4 @@
+require 'pry'
```
aenand, line 1: Pesky pry statement!

lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -32,7 +33,7 @@ def initialize(options = {})
         @client_id = options[:client_id]
         @client_api_key = options[:client_api_key]
         super
-        @access_token = setup_access_token
+        @access_token = options[:access_token] || setup_access_token
```
aenand, line 36: Why make this change?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -18,13 +19,13 @@ class CheckoutV2Gateway < Gateway
 
       def initialize(options = {})
         @options = options
-        @access_token = nil
+        @access_token = options[:access_token] || nil
```
aenand, line 22: When is `options[:access_token]` set?

aenand (commented):
Great test coverage! I think this looks good just one question about the `options[:access_token]`. Was that done to make testing easier?

lib/active_merchant/billing/gateways/airwallex.rb
almalee24: Good catch!

almalee24:
> Great test coverage! I think this looks good just one question about the `options[:access_token]`. Was that done to make testing easier?

Yes, to add unit testing. Without it I can't skip the access token method for all the other units tests and it causes them to fail. It couldn't find a better way to do it, the way it was done before was calling the method within the unit test file, https://github.com/activemerchant/active_merchant/pull/4851/files#diff-f3fb17a2b3e2cdf9d21bc45cc072a7f6f740a76e2b362b2168892dc94a1904b0L3

PR #4852, opened 2023-08-11
Authorize.NET: Update network token method

almalee24:
Update network token payment method to clarify what is being sent.

Remote:
85 tests, 304 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
122 tests, 684 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -89,7 +89,6 @@ module ActiveMerchant
         2 => /\A;(?<pan>[\d]{1,19}+)=(?<expiration>[\d]{0,4}|=)(?<service_code>[\d]{0,3}|=)(?<discretionary_data>.*)\?\Z/
       }.freeze
 
-      APPLE_PAY_DATA_DESCRIPTOR = 'COMMON.APPLE.INAPP.PAYMENT'
       PAYMENT_METHOD_NOT_SUPPORTED_ERROR = '155'
       INELIGIBLE_FOR_ISSUING_CREDIT_ERROR = '54'
 
@@ -165,7 +164,7 @@ module ActiveMerchant
             xml.transactionType('refundTransaction')
             xml.amount(amount(amount))
 
-            add_payment_source(xml, payment, options, :credit)
+            add_payment_method(xml, payment, options, :credit)
             xml.refTransId(transaction_id_from(options[:transaction_id])) if options[:transaction_id]
             add_invoice(xml, 'refundTransaction', options)
             add_customer_data(xml, payment, options)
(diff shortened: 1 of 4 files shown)


aenand (commented):
This PR seems to remove any `apple_pay` flagging for AuthNet. Are we sure AuthNet does not care about any flagging to let them know if the payment_method is `apple_pay` or not?

almalee24:
> This PR seems to remove any `apple_pay` flagging for AuthNet. Are we sure AuthNet does not care about any flagging to let them know if the payment_method is `apple_pay` or not?

Yes, I've been talking to them directly about this.  The `apple_pay` flagging is only for the Accept Mobile SDK and customers have enabled Apple Pay from within the gateway that it’s tied to. Also by the way it's currently implement that part of the code isn't actually being used.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -407,17 +406,20 @@ def normal_void(authorization, options)
...
-      def add_payment_source(xml, source, options, action = nil)
-        return unless source
+      def add_payment_method(xml, payment_method, options, action = nil)
+        return unless payment_method
 
-        if source.is_a?(String)
-          add_token_payment_method(xml, source, options)
-        elsif card_brand(source) == 'check'
-          add_check(xml, source)
-        elsif card_brand(source) == 'apple_pay'
```
BritneyS, line 417: Auth.net may not care about the `apple_pay` flagging, but I'm wondering if it's still worth it to keep this check in our decision tree. It will make a difference to us if we ever need to add GooglePay or network tokenization outside of digital wallets 🤔

almalee24: This doesn't really work because it won't ever return `apple_pay` but the returns the card brand for the PM. In my understanding this works for any Tokenized Credit Card because their docs doesn't say that this is only for ApplePay.

lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -526,17 +527,19 @@ def add_token_payment_method(xml, token, options)
...
+            xml.expirationDate(format(payment_method.month, :two_digits) + '/' + format(payment_method.year, :four_digits))
+            xml.isPaymentToken(true)
+            xml.cryptogram(payment_method.payment_cryptogram)
           end
         end
       end
 
       def add_market_type_device_type(xml, payment, options)
-        return if payment.is_a?(String) || card_brand(payment) == 'check' || card_brand(payment) == 'apple_pay'
+        return unless payment.is_a?(CreditCard) && !payment.is_a?(NetworkTokenizationCreditCard)
```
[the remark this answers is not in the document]
almalee24: I split them in two different lines

BritneyS (approved):
Looks good to me! Thanks for diving deep into this one 🚀

PR #4853, opened 2023-08-14
CommerceHub: Update headers

javierpedrozaing:
Description
-------------------------
This commit add new headers identifiers: 'x-originator', 'user-agent'. Additionally include a small test to verify the headers presence.

[SER-621](https://spreedly.atlassian.net/browse/SER-621)

Unit test
-------------------------
Finished in 0.033772 seconds.

25 tests, 175 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

740.26 tests/s, 5181.81 assertions/s

Remote test
-------------------------
Finished in 57.0542 seconds.

26 tests, 62 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 80.7692% passed

0.46 tests/s, 1.09 assertions/s

Rubocop
-------------------------
763 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/commerce_hub.rb
+++ b/lib/active_merchant/billing/gateways/commerce_hub.rb
@@ -282,7 +282,7 @@ module ActiveMerchant #:nodoc:
         raw_signature = @options[:api_key] + client_request_id.to_s + time + request
         hmac = OpenSSL::HMAC.digest('sha256', @options[:api_secret], raw_signature)
         signature = Base64.strict_encode64(hmac.to_s).to_s
-
+        custom_headers = options.fetch(:headers_identifiers, {})
         {
           'Client-Request-Id' => client_request_id,
           'Api-Key' => @options[:api_key],
@@ -292,7 +292,7 @@ module ActiveMerchant #:nodoc:
           'Content-Type' => 'application/json',
           'Accept' => 'application/json',
           'Authorization' => signature
-        }
+        }.merge!(custom_headers)
       end
 
       def add_merchant_details(post)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -291,7 +291,9 @@ def headers(request, options)
           'Auth-Token-Type' => 'HMAC',
           'Content-Type' => 'application/json',
           'Accept' => 'application/json',
-          'Authorization' => signature
+          'Authorization' => signature,
+          'x-originator' => 'CommerceHub-Partners-Spreedly',
+          'user-agent' => 'CommerceHub-Partners-Spreedly-V1.00'
```
Heavyblade, line 296: ⚠️ Warning ⚠️ 

There is an issue with this approach => See, Spreedly is not the only user of this gem so any other protect/library that depends on this adapter is going to send traffic is if it was Spreedly:

Alternative approach:

1) just add and extra option that will have a hash with the custom headers, something like:

```ruby
options.fetch(:custom_headers, {}).merge({
  'Client-Request-Id' => client_request_id,
  'Api-Key' => @options[:api_key],
  'Timestamp' => time,
  'Accept-Language' => 'application/json',
  'Auth-Token-Type' => 'HMAC',
  'Accept' => 'application/json',
  'Authorization' => signature
})
```

2) in core you only need to change the CommerceHubGateway#auth_purchase_params to include those headers in the list of params to AM.

Heavyblade (commented):
@javierpedrozaing done with the review, leave you something big to change.

Heavyblade (approved):
Looks 👍

naashton (approved):

PR #4854, opened 2023-08-14
IPG: Refactor Credentials

kylene-spreedly:
Refactor IPG credentials to accept a combined user ID/store ID string since this is what is provided to IPG merchants as their user ID.

ECS-2839

Unit: 31 tests, 135 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications, 100% passed

Remote: 18 tests, 42 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications, 66.6667% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -17,7 +17,7 @@
 * Rapyd: Add merchant_reference_id [jcreiff] #4858
 * Braintree: Return error for ACH on credit [jcreiff] #4859
 * Rapyd: Update handling of ewallet and billing address phone [jcreiff] #4863
-
+* IPG: Change credentials inputs to use a combined store and user ID string as the user ID input [kylene-spreedly] #4854
 
 == Version 1.134.0 (July 25, 2023)
 * Update required Ruby version [almalee24] #4823
--- a/lib/active_merchant/billing/gateways/ipg.rb
+++ b/lib/active_merchant/billing/gateways/ipg.rb
@@ -18,8 +18,8 @@ module ActiveMerchant #:nodoc:
       ACTION_REQUEST_ITEMS = %w(vault unstore)
 
       def initialize(options = {})
-        requires!(options, :store_id, :user_id, :password, :pem, :pem_password)
-        @credentials = options
+        requires!(options, :user_id, :password, :pem, :pem_password)
+        @credentials = options.merge(store_and_user_id_from(options[:user_id]))
         @hosted_data_id = nil
         super
       end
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/ipg.rb
```diff
@@ -395,6 +395,11 @@ def parse_element(reply, node)
         return reply
       end
 
+      def store_and_user_id_from(user_id)
+        split_credentials = user_id.split('._.')
```
kylene-spreedly, line 399: I didn't add any validation in here to make sure the delimiter is present. I wasn't sure if we typically validate user input to that degree, but happy to add it in if necessary.

lib/active_merchant/billing/gateways/ipg.rb
```diff
@@ -395,6 +395,11 @@ def parse_element(reply, node)
...
+        return { store_id: split_credentials[0].sub(/^WS/, ''), user_id: split_credentials[1] }
```
curiousepic, line 400: FYI `return` is implicit in most cases
```suggestion
        { store_id: split_credentials[0].sub(/^WS/, ''), user_id: split_credentials[1] }
```

CHANGELOG
```diff
@@ -14,7 +14,7 @@
 * Deepstack: Add Deepstack Gateway [khoinguyendeepstack] #4830
 * Braintree: Additional tests for credit transactions [jcreiff] #4848
 * Rapyd: Change nesting of description, statement_descriptor, complete_payment_url, and error_payment_url [jcreiff] #4849
-
+* IPG: Refactor Credentials [kylene-spreedly] #4854
```
curiousepic, line 17: Nitpick, but this is not technically a refactor since it changes functionality

lib/active_merchant/billing/gateways/ipg.rb
curiousepic: Nice thinking on accepting a string without `WS`, but we should make sure its only removed if it is the first two characters - if that's already what it does, awesome... I can't sightread regex 😄

kylene-spreedly: Removed the return! Also yes, the `^` in the regex string is a check to make sure it only matches at the beginning of the string. I included an additional test though to make sure!

CHANGELOG
kylene-spreedly: I updated this too. Lemme know if that makes more sense!

curiousepic (approved):
👍  Looks good, thanks for the changes.

PR #4855, opened 2023-08-14
VisaNet Peru: Update generate_purchase_number_stamp

almalee24:
Update generate_purchase_number_stamp to always produce unique 12-digit alphanumberic value even if multiple transactions occur at the same time.

Unit:
16 tests, 100 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/visanet_peru.rb
+++ b/lib/active_merchant/billing/gateways/visanet_peru.rb
@@ -143,7 +143,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def generate_purchase_number_stamp
-        Time.now.to_f.to_s.delete('.')[1..10] + rand(99).to_s
+        rand(('9' * 12).to_i).to_s.center(12, rand(9).to_s)
       end
 
       def commit(action, params, options = {})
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/visanet_peru.rb
```diff
@@ -143,7 +143,7 @@ def split_authorization(authorization)
       end
 
       def generate_purchase_number_stamp
-        Time.now.to_f.to_s.delete('.')[1..10] + rand(99).to_s
+        rand(('9' * 12).to_i).to_s.center(12, rand(9).to_s)
```
aenand, line 146: What does `center(12, rand(9).to_s)` do here?

almalee24: So the first part `rand(('9' * 12).to_i).to_s` doesn't always returns a 12-digit string so center checks if the length is 12 if not it returns a new string with added random numbers

aenand (approved):
LGTM 🚀

PR #4862, opened 2023-08-17
Checkout: Add support for `sender`, `destination` and `instruction` o…

rachelkirk:
…bjects

CER-757

These changes will enhance functionality for Credits/Payouts. The fields that are added with this logic are all opt-in, no existing logic has been changed.

There are several fields that should be added conditionally, for this it would be best to reference the [Checkout docs](https://api-reference.checkout.com/#operation/requestAPaymentOrPayout).

Remote Tests:
95 tests, 231 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit Tests:
60 tests, 360 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Local Tests:
5551 tests, 77653 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -57,12 +57,13 @@ module ActiveMerchant #:nodoc:
 
       def credit(amount, payment, options = {})
         post = {}
-        post[:instruction] = {}
-        post[:instruction][:funds_transfer_type] = options[:funds_transfer_type] || 'FD'
         add_processing_channel(post, options)
         add_invoice(post, amount, options)
         add_payment_method(post, payment, options, :destination)
         add_source(post, options)
+        add_instruction_data(post, options)
+        add_payout_sender_data(post, options)
+        add_payout_destination_data(post, options)
 
         commit(:credit, post, options)
       end
@@ -173,6 +174,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_payment_method(post, payment_method, options, key = :source)
+        # the key = :destination when this method is called in def credit
         post[key] = {}
         case payment_method
         when NetworkTokenizationCreditCard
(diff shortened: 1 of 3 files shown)


jcreiff (changes requested):
One small note about a test; other than that, clean up the comments and I think this is good to go 😅

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -57,12 +57,13 @@ def capture(amount, authorization, options = {})
...
         post = {}
-        post[:instruction] = {}
-        post[:instruction][:funds_transfer_type] = options[:funds_transfer_type] || 'FD'
         add_processing_channel(post, options)
         add_invoice(post, amount, options)
         add_payment_method(post, payment, options, :destination)
         add_source(post, options)
+        add_instruction_data(post, options)
+        add_payout_sender_data(post, options)
+        add_payout_destination_data(post, options)
```
aenand, line 66: Holy cow that is a lot of new fields for payouts. Are these fields reserved only for payouts?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -321,6 +334,80 @@ def add_processing_channel(post, options)
...
+        post[:sender][:middle_name] = options[:sender][:middle_name] if options[:sender][:middle_name]
+        post[:sender][:last_name] = options[:sender][:last_name] if options[:sender][:last_name] # must be <= 35 char
+        # company name required if sent by type: corporate or government
+        post[:sender][:company_name] = options[:sender][:company_name] if options[:sender][:company_name] # must be <= 35 char
+        # these are required fields for payout, may not work if address is blank or different than cardholder(option for sender to be a company or government).
+        # may need to still include in GSF hash.
+
+        post[:sender][:address] = {}
+        post[:sender][:address][:address_line1] = options.dig(:sender, :address, :address1) if options.dig(:sender, :address, :address1)
+        post[:sender][:address][:address_line2] = options.dig(:sender, :address, :address2) if options.dig(:sender, :address, :address2)
```
almalee24, line 360: Instead of checking if param exist could we just do `compact` to make this more clean?

lib/active_merchant/billing/gateways/checkout_v2.rb
almalee24: So by adding these fields it makes the credit transaction a payout? I asking cause I also that the endpoint doesn't change for payouts

yunnydang (approved):
lots of new fields! Aside from the comments from the others, I think this is good. Great work!

yunnydang:
Also you CI checks are failing, so I'd get those changes in as well, thanks!

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -190,13 +192,23 @@ def add_payment_method(post, payment_method, options, key = :source)
           post[key][:type] = 'card'
           post[key][:name] = payment_method.name
           post[key][:number] = payment_method.number
-          post[key][:cvv] = payment_method.verification_value
+          post[key][:cvv] = payment_method.verification_value unless options[:funds_transfer_type]
```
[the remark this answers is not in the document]
rachelkirk: options[:payout] will trigger one kind of payout by the extra fields being sent but funds_transfer_type is already being passed for the first type of credit/payout

PR #4864, opened 2023-08-18
Braintree: Add sca_exemption

almalee24:
To specifiy the SCA exemption that a trasaction will be claming.

Remote:
105 tests, 559 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
94 tests, 207 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -678,6 +678,8 @@ module ActiveMerchant #:nodoc:
 
         add_3ds_info(parameters, options[:three_d_secure])
 
+        parameters[:sca_exemption] = options[:three_ds_exemption_type] if options[:three_ds_exemption_type]
+
         if options[:payment_method_nonce].is_a?(String)
           parameters.delete(:customer)
           parameters[:payment_method_nonce] = options[:payment_method_nonce]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -673,6 +673,8 @@ def create_transaction_parameters(money, credit_card_or_vault_id, options)
 
         add_3ds_info(parameters, options[:three_d_secure])
 
+        parameters[:sca_exemption] = options[:three_ds_exemption_type] if options[:three_ds_exemption_type]
```
aenand, line 676: There will likely be a set amount of standard exemption types across gateways. What if we standardize them inside the `options[:three_d_secure]` [hash](https://github.com/activemerchant/active_merchant/wiki/Standardized-3DS-Fields) and then map them to the gateway's specific values kind of like `STANDARD_ERROR_CODE`
```
THREE_DS_EXEMPTIONS = {
  'low_value': 'low_value',
  'another_exemption': 'exemption_another'
}
```

aenand: Looking around AM quite a few gateways look for something like `options[:sca_exemption]` so maybe we standardize on that field?

almalee24: `:sca_exemption` is what we map for Adyen & Barclaycard and for Redsys is what the field the customer sends us. `three_ds_exemption_type` is the field that customer sends us but we map it on our side to be specific to the gateway.

aenand: the more you know!

PR #4866, opened 2023-08-21
Ebanx: Update Verify

almalee24:
Update Verify to use verifycard endpoint.

Unit:
19 tests, 85 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
32 tests, 88 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.875% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -20,7 +20,8 @@ module ActiveMerchant #:nodoc:
         refund: 'refund',
         void: 'cancel',
         store: 'token',
-        inquire: 'query'
+        inquire: 'query',
+        verify: 'verifycard'
       }
 
       HTTP_METHOD = {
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -192,13 +189,17 @@ def add_invoice(post, money, options)
...
         payment = payment.split('|')[0] if payment.is_a?(String)
-        post[:payment][:payment_type_code] = 'creditcard'
+        add_payment_type(post[:payment])
         post[:payment][:creditcard] = payment_details(payment)
         post[:payment][:creditcard][:soft_descriptor] = options[:soft_descriptor] if options[:soft_descriptor]
       end
 
-      def add_payment_details(post, payment)
+      def add_payment_type(post)
         post[:payment_type_code] = 'creditcard'
```
aenand, line 198: If the `payment_type` is only ever `creditcard` should we add this to `add_payment_details`?

almalee24: True, will do that!

almalee24: Actually there's a reason I did that because because `Verify` uses `post[:card]` and the other methods use `post[:creditcard]` which is why I move `payment_type` out. I'm actually going to update that and remove `add_payment_details` so it makes a bit more sense.

aenand (approved):
LGTM 🚀

PR #4869, opened 2023-08-22

almalee24:

### DIFF
--- a/lib/active_merchant/errors.rb
+++ b/lib/active_merchant/errors.rb
@@ -23,6 +23,18 @@ module ActiveMerchant #:nodoc:
     end
 
     def to_s
+      if response.kind_of?(String)
+        if response.start_with?('Failed with')
+          return response
+        else
+          return "Failed with #{response}"
+        end
+      end
+
+      if response.respond_to?(:message)
+        return response.message if response.message.start_with?('Failed with')
+      end
+
       "Failed with #{response.code if response.respond_to?(:code)} #{response.message if response.respond_to?(:message)}"
     end
   end
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -20,6 +20,7 @@
 * Decidir: Add support for network tokens [almalee24] #4870
 * Element: Fix credit card name bug [almalee24] #4898
 * Adyen: Add payout endpoint [almalee24] #4885
+* Adding Oauth Response for access tokens [almalee24] #4851
 
 == Version 1.135.0 (August 24, 2023)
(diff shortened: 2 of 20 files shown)


BritneyS (commented):
Can you share a Test Summary? I'm unable to get passing remote tests for these gateways on this branch so I wanted to see if it's just something wrong in my environment.

lib/active_merchant/errors.rb
```diff
@@ -23,6 +23,12 @@ def initialize(response, message = nil)
     end
 
     def to_s
+      if response.kind_of?(String)
+        return response if response.start_with?('Failed with')
```
BritneyS, line 27: As-is, an arbitrary `String` response that doesn't start with `Failed with` would say `ActiveMerchant::OAuthResponseError: Failed with`. Is that desirable?

BritneyS (commented):
Did some scattershot testing with other gateways that raise `ActiveMerchant::OAuthResponseError` and that looks good. Just had a general question.

lib/active_merchant/errors.rb
almalee24: Yes exactly. I add this check because [some gateways like Alelo check access_token method in `commit`](https://github.com/activemerchant/active_merchant/pull/4869/files#diff-9ab1fec4c8a427ccb1bbff06670245bf4f9adf54e536926b01fc8aeba6151634R221) so it would return `ActiveMerchant::OAuthResponseError: Failed with Failed with`.

BritneyS: But wouldn't returning the the `response` string in that case be more informative?

almalee24: In the case I described the OAuthResponseError or maybe I'm not understanding what you trying to say

BritneyS: I mean that error strings like `An error happened` would return `ActiveMerchant::OAuthResponseError: Failed with` and not fall to the general case `ActiveMerchant::OAuthResponseError: Failed with An error happened`. So I'm wondering if the `if response.start_with?('Failed with')` check is necessary, when it could read as `ActiveMerchant::OAuthResponseError: An error happened` without that check.

almalee24: Oh, if response is a string then it would be a `ActiveMerchant::OAuthResponseError`. If response was `ActiveMerchant::OAuthResponseError` the it would fall into the second statement, `elsif response.respond_to?(:message)`.

BritneyS (approved):
Looks good 👍🏾

PR #4870, opened 2023-08-22
Decidir: Add support for network tokens

almalee24:
This PR adds support for network tokens to the Decidir gateway.

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir.rb
+++ b/lib/active_merchant/billing/gateways/decidir.rb
@@ -167,29 +167,55 @@ module ActiveMerchant #:nodoc:
         post[:amount] = localized_amount(money, currency).to_i
       end
 
-      def add_payment(post, credit_card, options)
-        card_data = {}
+      def add_payment(post, payment_method, options)
+        add_common_payment_data(post, payment_method, options)
+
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/decidir.rb
```diff
@@ -167,29 +167,55 @@ def add_amount(post, money, options)
...
+
+      def add_network_token(post, payment_method, options)
+        post[:is_tokenized_payment] = true
+        post[:fraud_detection] ||= {}
+        post[:fraud_detection][:sent_to_cs] = false
+        post[:card_data][:last_four_digits] = payment_method.last_digits
+
+        post[:token_card_data] = {
+          token: payment_method.number,
+          eci: payment_method.eci,
```
BritneyS, line 203: Is the `eci` value required for Decidir? It's not required for network tokenization for all gateways. If not, maybe there should be a check for the presence of `eci` value on the `payment_method` before populating the `eci` field.

almalee24: Yes, it's required.

kylene-spreedly (approved):

PR #4874, opened 2023-08-24
CYBS: Recurring Apple Pay

aenand:
Cybersource's legacy gateway supports recurring transactions for Network Tokens. The way to accomplish this is to not send the `cryptogram` since that is one time use and mark the `commerce_indicator` as `internet`.

Remote:
123 tests, 619 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 95.935% passed
5 tests failing on master

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -807,26 +807,35 @@ module ActiveMerchant #:nodoc:
         payment_method.is_a?(NetworkTokenizationCreditCard)
       end
 
+      def subsequent_nt_apple_pay_auth(source, options)
+        return unless options[:stored_credential] || options[:stored_credential_overrides]
+        return unless @@payment_solution[source]
+
+        options.dig(:stored_credential_overrides, :subsequent_auth) || options.dig(:stored_credential, :initiator) == 'merchant'
+      end
+
       def add_auth_network_tokenization(xml, payment_method, options)
         return unless network_tokenization?(payment_method)
+
+        commerce_indicator = 'internet' if subsequent_nt_apple_pay_auth(payment_method.source, options)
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -807,26 +807,35 @@ def network_tokenization?(payment_method)
         payment_method.is_a?(NetworkTokenizationCreditCard)
       end
 
+      def subsequent_nt_apple_pay_auth(source, options)
+        return unless options[:stored_credential] || options[:stored_credential_overrides]
+        return unless @@payment_solution[source]
```
aenand, line 812: This change was reverted previously because Network Tokens were falling down the code path to update the `commerce_indicator` to `internet` causing an increase in `General Decline`. This change now verifies the payment method is AP or GP before updating the `commerce_indicator`

test/unit/gateways/cyber_source_test.rb
```diff
@@ -472,9 +467,69 @@ def test_successful_credit_cart_purchase_single_request_without_ignore_ccv
...
+    })
+    assert response = @gateway.purchase(@amount, credit_card, options)
+    assert_success response
+  end
+
+  def test_successful_network_token_purchase_subsequent_auth_visa
+    @gateway.expects(:ssl_post).with do |_host, request_body|
+      assert_match %r'<cavv>111111111100cryptogram</cavv>', request_body
+      assert_match %r'<commerceIndicator>vbv</commerceIndicator>', request_body
+      assert_not_match %r'<commerceIndicator>internet</commerceIndicator>', request_body
```
aenand, line 521: Test to verify we do not omit certain data for recurring NT like we do for recurring AP

BritneyS (approved):
Looks good 👍🏾

PR #4877, opened 2023-08-29
Rapyd: Add fields and update stored credential method

yunnydang:
This change is allow users to manage NTID themselves by adding an option for them to pass in the fields network_reference_id and initiation_type directly.

Local:
5588 tests, 77761 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications 99.66% passed

Unit:
27 tests, 137 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
34 tests, 95 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -141,10 +141,22 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_credential(post, options)
-        return unless stored_credential = options[:stored_credential]
+        add_network_reference_id(post, options)
+        add_initiation_type(post, options)
+      end
 
-        post[:payment_method][:fields][:network_reference_id] = stored_credential[:network_transaction_id] if stored_credential[:network_transaction_id]
-        post[:initiation_type] = stored_credential[:reason_type] if stored_credential[:reason_type]
+      def add_network_reference_id(post, options)
+        return unless options[:stored_credential] || options[:network_transaction_id]
+
+        network_transaction_id = options[:network_transaction_id] || options[:stored_credential][:network_transaction_id]
+        post[:payment_method][:fields][:network_reference_id] = network_transaction_id if network_transaction_id
+      end
+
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -141,10 +141,32 @@ def add_payment(post, payment, options)
...
+
+      def add_network_reference_id(post, options)
+        return unless options[:stored_credential] || options[:network_reference_id]
+
+        if options[:network_reference_id]
+          network_reference_id = options[:network_reference_id]
+        else
+          network_reference_id = options[:stored_credential][:network_transaction_id]
+        end
+
```
rachelkirk, line 156: Two things here: 
1. I would create this NTID override as options[:network_transaction_id] to keep it consistent across other gateways that have this override (even though Rapyd calls it something else)
2. I would use the ternary operator here. Something like 
`network_reference_id = options[:network_transaction_id] ? options[:network_transaction_id] : options[:stored_credential][:network_transaction_id]`.   That might be a little cleaner. You could also use this in the add_initiation_type method below.

yunnydang: Thanks for the suggestions, it totally looks cleaner this way!

yunnydang: I was running into rubocop errors when writing it `network_reference_id = options[:network_transaction_id] ? options[:network_transaction_id] : options[:stored_credential][:network_transaction_id]` as it's stating it was an unneeded conditional. The new way I wrote is the same as it'll prioritize the GSF if it's present and then default to the stored_credential hash. Also updated the tests and variable names to reflect network_transaction_id like you suggested and also added another assertion to the remote test when sending both stored credentials and options, thanks a bunch!

rachelkirk (approved):
Looks good! Nice refactoring!

PR #4878, opened 2023-08-30
Adyen: Add the store field

yunnydang:
This change is to add the store field

Local:
5589 tests, 77761 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications 99.66% passed

Unit:
111 tests, 583 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
134 tests, 447 assertions, 11 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 91.791% passed

Note: 
The reason why I didn't pursue a remote test for this change is because it requires an actual store reference to be sent, otherwise it'll return an invalid store error message and code. Not sure if the sandbox supports it, but we'd have to create a store, then grab the reference, and send that in our remote test in order to not have a failing test.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@
 * Cybersource: Support recurring apple pay [aenand] #4874
 * Verve BIN ranges and add card type to Rapyd gateway [jherreraa] #4875
 * Rapyd: Add network_reference_id, initiation_type, and update stored credential method [yunnydang] #4877
+* Adyen: Add the store field [yunnydang] #4878
 
 == Version 1.135.0 (August 24, 2023)
 * PaymentExpress: Correct endpoints [steveh] #4827
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -245,6 +245,7 @@ module ActiveMerchant #:nodoc:
         post[:additionalData][:industryUsage] = options[:industry_usage] if options[:industry_usage]
         post[:additionalData][:RequestedTestAcquirerResponseCode] = options[:requested_test_acquirer_response_code] if options[:requested_test_acquirer_response_code] && test?
         post[:deviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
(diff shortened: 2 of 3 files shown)


rachelkirk (changes requested):
I am hesitant to approve if we can't get a passing remote test. This may put us in the situation of having to do a quick revert which is never fun. 

In looking into this further, the docs provided in the ticket reference the [Checkout API](https://docs.adyen.com/api-explorer/Checkout/latest/overview) but I believe we are integrated with the [Classic API](https://docs.adyen.com/api-explorer/Payment/68/post/authorise) which you can verify by looking at the endpoints we're hitting in the transcripts. We hit /authorise and /capture and in the newer Checkout API the main endpoint is /payments. All this to say, we need to verify if `store` is even supported in the Classic API before moving forward.

yunnydang:
> I am hesitant to approve if we can't get a passing remote test. This may put us in the situation of having to do a quick revert which is never fun.
> 
> In looking into this further, the docs provided in the ticket reference the [Checkout API](https://docs.adyen.com/api-explorer/Checkout/latest/overview) but I believe we are integrated with the [Classic API](https://docs.adyen.com/api-explorer/Payment/68/post/authorise) which you can verify by looking at the endpoints we're hitting in the transcripts. We hit /authorise and /capture and in the newer Checkout API the main endpoint is /payments. All this to say, we need to verify if `store` is even supported in the Classic API before moving forward.

The classic version does seem to support the store field found [here](https://docs.adyen.com/api-explorer/Payment/68/post/authorise#request-store)

rachelkirk:
> The classic version does seem to support the store field found [here](https://docs.adyen.com/api-explorer/Payment/68/post/authorise#request-store)

Thanks for verifying this! Were you able to connect with Adyen on how we could get a passing remote test, or perhaps attempt to test via Postman? I don't know that your results would be any different but it's an option to try it with the whole flow.

aenand (approved):
LGTM 🚀

PR #4880, opened 2023-08-30
TNS: Use the customer specified order_id in the request

yunnydang:
Local:
5588 tests, 77757 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications 99.66% passed

Unit:
15 tests, 68 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
13 tests, 42 assertions, 2 failures, 1 errors, 0 pendings, 0 omissions, 0 notifications 76.9231% passed

Note: I removed the test_verify_credentials because even the 'unknown' fixture creds were returning 400 response codes which is the same as the actual fixures


rachelkirk (approved):
Looks good, thanks for cleaning up the tests. It looks like that `test_verify_credentials` is pretty inconsistently implemented across gateways so I guess it's ok to remove. Maybe commenting out with a note would be safer? Up to you!

PR #4881, opened 2023-09-06
CyberSource (SOAP): Added support for 3DS exemption request fields

BritneyS:
CyberSource (SOAP): Added support for 3DS exemption request fields

Test Summary
Local: 5591 tests, 77934 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Unit at test/unit/gateways/cyber_source_test.rb: 138 tests, 762 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Remote at test/remote/gateways/remote_cyber_source_test.rb: 125 tests, 627 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96% passed (same as `master`)

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -33,6 +33,15 @@ module ActiveMerchant #:nodoc:
         discover: 'pb',
         diners_club: 'pb'
       }.freeze
+      THREEDS_EXEMPTIONS = {
+        authentication_outage: 'authenticationOutageExemptionIndicator',
+        corporate_card: 'secureCorporatePaymentIndicator',
+        delegated_authentication: 'delegatedAuthenticationExemptionIndicator',
+        low_risk: 'riskAnalysisExemptionIndicator',
+        low_value: 'lowValueExemptionIndicator',
+        stored_credential: 'stored_credential',
+        trusted_merchant: 'trustedMerchantExemptionIndicator'
+      }
       DEFAULT_COLLECTION_INDICATOR = 2
 
       self.supported_cardtypes = %i[visa master american_express discover diners_club jcb dankort maestro elo]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -33,6 +33,15 @@ class CyberSourceGateway < Gateway
         discover: 'pb',
         diners_club: 'pb'
       }.freeze
+      THREEDS_EXEMPTIONS = {
+        authentication_outage: 'authenticationOutageExemptionIndicator',
+        corporate_card: 'secureCorporatePaymentIndicator',
+        delegated_authentication: 'delegatedAuthenticationExemptionIndicator',
+        low_risk: 'riskAnalysisExemptionIndicator',
+        low_value: 'lowValueExemptionIndicator',
+        stored_credential: 'stored_credential',
```
aenand, line 42: Fascinated that this is a 3DS exemption

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -746,6 +756,15 @@ def add_auth_service(xml, payment_method, options)
         end
       end
 
+      def add_threeds_exemption_data(xml, options)
+        exemption = options[:three_ds_exemption_type].to_sym
```
aenand, line 760: If `options[:three_ds_exemption_type]` is nil, will calling `.to_sym` result in a error? Along the lines of "undefined method to_sym for nil:NilClass?"

BritneyS: There's a check for a non-nil `options[:three_ds_exemption_type]` value before the method is called. I'll add an extra guard statement to that method to be on the safe side, though 👍🏾

aenand (approved):
LGTM :shipit:

lib/active_merchant/billing/gateways/cyber_source.rb
BritneyS: Same!

PR #4882, opened 2023-09-08
Stripe PI: Expand balance txns for off session transactions

yunnydang:
This is to expand the balance transaction object when either a setup_future_usage or off_session field is being sent.

Local:
5595 tests, 77795 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6604% passed

Unit:
53 tests, 268 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
89 tests, 422 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -541,6 +541,7 @@ def add_external_three_d_secure_auth_data(post, options = {})
       def setup_future_usage(post, options = {})
         post[:setup_future_usage] = options[:setup_future_usage] if %w(on_session off_session).include?(options[:setup_future_usage])
         post[:off_session] = options[:off_session] if off_session_request?(options)
+        post[:expand] = ['charges.data.balance_transaction'] if post[:setup_future_usage] || post[:off_session]
```
almalee24, line 544: It also works for regular transactions, just tested without `if post[:setup_future_usage] || post[:off_session]`.

almalee24: Idk why it errored out when you tried it last time.

almalee24: Also for unit test maybe you can just check that `post[:expand] = ['charges.data.balance_transaction']` gets added to params

yunnydang: Thanks for the review and suggestions @almalee24 ! I've re-requested you for review after making the changes as well. Unsure myself why I was getting errors last time

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -541,6 +541,7 @@ def add_external_three_d_secure_auth_data(post, options = {})
       def setup_future_usage(post, options = {})
         post[:setup_future_usage] = options[:setup_future_usage] if %w(on_session off_session).include?(options[:setup_future_usage])
         post[:off_session] = options[:off_session] if off_session_request?(options)
+        post[:expand] = ['charges.data.balance_transaction']
```
almalee24, line 544: I would also recommend moving this out of here and directly into `create_intent` since this is no longer related to just offsite transactions.

yunnydang: @almalee24 so when i try to pass it in the create_setup_intent method, I receive the error "This property cannot be expanded (charges)." Not sure if that is supported for that endpoint

almalee24: Okay cool, thanks for checking that!

almalee24: I noticed you removed the remote test but I think if you just add `assert purchase.params.dig('charges', 'data')[0]['balance_transaction']` to like `test_successful_purchase` it would be good enough!

almalee24 (approved):
Great work!

PR #4884, opened 2023-09-12
Braintree: return global_id in response

rachelkirk:
CER-868

This field needs to be added to the transaction_hash method to be returned in the response. The Braintree gem required an update to make the `payment_receipt` object available

I did not include a unit test because I'm trying to get something returned in the response and according to the [Braintree docs](https://developer.paypal.com/braintree/docs/reference/response/transaction/ruby#global_id), it does not appear that anything should be passed in the request to trigger this coming back. 

Remote
107 tests, 212 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
31.7757% passed

Unit
95 tests, 110 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
41.0526% passed

Local
5599 tests, 77770 assertions, 0 failures, 56 errors, 0 pendings, 0 omissions, 0 notifications
98.9998% passed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'rubocop', '~> 0.72.0', require: false
 
 group :test, :remote_test do
   # gateway-specific dependencies, keeping these gems out of the gemspec
-  gem 'braintree', '>= 4.12.0'
+  gem 'braintree', '>= 4.14.0'
   gem 'jose', '~> 1.1.3'
   gem 'jwe'
   gem 'mechanize'
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -618,6 +618,14 @@ module ActiveMerchant #:nodoc:
           risk_data = nil
         end
 
+        if transaction.payment_receipt
+          payment_receipt = {
+            'global_id' => transaction.payment_receipt.global_id
+          }
+        else
+          payment_receipt = nil
+        end
+
         {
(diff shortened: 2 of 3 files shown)


yunnydang (commented):
All looks good! However i noticed there wasn't any unit tests? Maybe we can write a simple one similar to the risk_data object on line 248 in the unit spec file?

almalee24 (approved):

yunnydang (approved):

rachelkirk:
> All looks good! However i noticed there wasn't any unit tests? Maybe we can write a simple one similar to the risk_data object on line 248 in the unit spec file?

The unit tests are testing what is going into the requests, since there's nothing about global_id in the request and it's only returned in the response I can't really write a unit test for it. Those risk data tests have things that are included in the request headers

yunnydang:
> > All looks good! However i noticed there wasn't any unit tests? Maybe we can write a simple one similar to the risk_data object on line 248 in the unit spec file?
> 
> The unit tests are testing what is going into the requests, since there's nothing about global_id in the request and it's only returned in the response I can't really write a unit test for it. Those risk data tests have things that are included in the request headers



> > All looks good! However i noticed there wasn't any unit tests? Maybe we can write a simple one similar to the risk_data object on line 248 in the unit spec file?
> 
> The unit tests are testing what is going into the requests, since there's nothing about global_id in the request and it's only returned in the response I can't really write a unit test for it. Those risk data tests have things that are included in the request headers

Nice! Thanks for clarifying I just peeked super quick at the unit tests, awesome work!

PR #4885, opened 2023-09-12
Adyen: Add payout endpoint

almalee24:
Remote:
137 tests, 453 assertions, 11 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 91.9708% passed

Unit:
111 tests, 585 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -93,12 +93,27 @@ module ActiveMerchant #:nodoc:
       end
 
       def credit(money, payment, options = {})
-        action = 'refundWithData'
+        action = options[:payout] ? 'payout' : 'refundWithData'
         post = init_post(options)
         add_invoice(post, money, options)
         add_payment(post, payment, options, action)
         add_shopper_reference(post, options)
         add_network_transaction_reference(post, options)
+
+        if action == 'payout'
+          add_shopper_interaction(post, payment, options)
+          add_fraud_offset(post, options)
+          add_fund_source(post, options)
+          add_recurring_contract(post, options)
+
+          if (address = options[:billing_address] || options[:address]) && address[:country]
+            add_billing_address(post, address)
+          end
+
+          post[:dateOfBirth] = options[:date_of_birth] if options[:date_of_birth]
+          post[:nationality] = options[:nationality] if options[:nationality]
+        end
+
         commit(action, post, options)
       end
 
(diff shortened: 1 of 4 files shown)


almalee24:
A few of these test are also failing in master because of `card_not_stored?`

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -688,6 +720,23 @@ def add_3ds2_authenticated_data(post, options)
         }
       end
 
+      def add_fund_source(post, options)
+        return unless fund_source = options[:fund_source]
```
almalee24, line 724: The also have a `card` object but wasn't sure that we want customers to send us card info like this

curiousepic (approved):
Nice - I am a fan of the restructing and cleanup in here too.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -229,13 +244,27 @@ def scrub(transcript)
...
-        post[:telephoneNumber] = (options[:billing_address][:phone_number] if options.dig(:billing_address, :phone_number)) || (options[:billing_address][:phone] if options.dig(:billing_address, :phone)) || ''
-        post[:fraudOffset] = options[:fraud_offset] if options[:fraud_offset]
         post[:selectedBrand] = options[:selected_brand] if options[:selected_brand]
         post[:selectedBrand] ||= NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
         post[:deliveryDate] = options[:delivery_date] if options[:delivery_date]
         post[:merchantOrderReference] = options[:merchant_order_reference] if options[:merchant_order_reference]
         post[:captureDelayHours] = options[:capture_delay_hours] if options[:capture_delay_hours]
+        post[:deviceFingerprint] = options[:device_fingerprint] if options[:device_fingerprint]
+        post[:shopperIP] = options[:ip] if options[:ip]
+        post[:shopperIP] = options[:shopper_ip] if options[:shopper_ip]
```
kylene-spreedly, line 254: Are both of these lines meant to assign `shopperIP`? If so, is there a way to make this more concise?

almalee24: Yeah, idk why it was originally done like that. I'll update it

kylene-spreedly: Nice! I didn't realize that was existing code that had moved. My bad 🙃

kylene-spreedly (approved):

PR #4890, opened 2023-09-14
SumUp - Setup, Scrub and Purchase build

sinourain:
Description
-------------------------
Setup, scrub and purchase method to SumUp Gateway adapter with the basic information needed.

This are the relevant links to review the initial implementation:

- [SumUp REST API](https://developer.sumup.com/docs/api/sum-up-rest-api/) 
- [Make a payment with a card entered by a customer](https://developer.sumup.com/docs/online-payments/guides/single-payment/) 
- [Checkouts](https://developer.sumup.com/docs/api/checkouts/)

Tickets for Spreedly reference
SER-764
SER-712
SER-711

Unit test
-------------------------
Finished in 37.342409 seconds.
5606 tests, 78027 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

Remote test
-------------------------
100% passed
150.12 tests/s, 2089.50 assertions/s

Rubocop
-------------------------
Inspecting 769 files
769 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/sum_up.rb
+++ b/lib/active_merchant/billing/gateways/sum_up.rb
@@ -1,0 +1,183 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class SumUpGateway < Gateway
+      self.live_url = 'https://api.sumup.com/v0.1/'
+
+      self.supported_countries = %w(AT BE BG BR CH CL CO CY CZ DE DK EE ES FI FR
+                                    GB GR HR HU IE IT LT LU LV MT NL NO PL PT RO
+                                    SE SI SK US)
+      self.currencies_with_three_decimal_places = %w(EUR BGN BRL CHF CZK DKK GBP
+                                                     HUF NOK PLN SEK USD)
+      self.default_currency = 'USD'
+
+      self.homepage_url = 'https://www.sumup.com/'
+      self.display_name = 'SumUp'
+
+      STANDARD_ERROR_CODE_MAPPING = {
+        multiple_invalid_parameters: 'MULTIPLE_INVALID_PARAMETERS'
+      }
+
+      def initialize(options = {})
+        requires!(options, :access_token, :pay_to_email)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        MultiResponse.run do |r|
+          r.process { create_checkout(money, payment, options) } unless options[:checkout_id]
+          r.process { complete_checkout(options[:checkout_id] || r.params['id'], payment, options) }
+        end
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Bearer )\w+), '\1[FILTERED]').
+          gsub(%r(("pay_to_email\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("number\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("cvv\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
+      end
+
+      private
+
+      def create_checkout(money, payment, options)
+        post = {}
+
+        add_merchant_data(post, options)
+        add_invoice(post, money, options)
+        add_address(post, options)
+        add_customer_data(post, payment, options)
+
+        commit('checkouts', post)
+      end
+
+      def complete_checkout(checkout_id, payment, options = {})
+        post = {}
+
+        add_payment(post, payment, options)
+
+        commit('checkouts/' + checkout_id, post, :put)
+      end
+
+      def add_customer_data(post, payment, options)
+        post[:customer_id]      = options[:customer_id]
+        post[:personal_details] = {
+          email:      options[:email],
+          first_name: payment&.first_name,
+          last_name:  payment&.last_name,
+          tax_id:     options[:tax_id]
+        }
+      end
+
+      def add_merchant_data(post, options)
+        # Required field: pay_to_email
+        # Description: Email address of the merchant to whom the payment is made.
+        post[:pay_to_email] = @options[:pay_to_email]
+      end
+
+      def add_address(post, options)
+        post[:personal_details] ||= {}
+        if address = (options[:billing_address] || options[:shipping_address] || options[:address])
+          post[:personal_details][:address] = {
+            city:        address[:city],
+            state:       address[:state],
+            country:     address[:country],
+            line_1:      address[:address1],
+            postal_code: address[:zip]
+          }
+        end
+      end
+
+      def add_invoice(post, money, options)
+        payment_currency = options[:currency] || currency(money)
+        post[:checkout_reference] = options[:order_id]
+        post[:amount]             = localized_amount(money, payment_currency)
+        post[:currency]           = payment_currency
+        post[:description]        = options[:description]
+      end
+
+      def add_payment(post, payment, options)
+        post[:payment_type] = options[:payment_type] || 'card'
+
+        post[:card] = {
+          name:         payment.name,
+          number:       payment.number,
+          expiry_month: format(payment.month, :two_digits),
+          expiry_year:  payment.year,
+          cvv:          payment.verification_value
+        }
+      end
+
+      def commit(action, post, method = :post)
+        response = api_request(action, post.compact, method)
+
+        Response.new(
+          success_from(response),
+          message_from(response),
+          response,
+          authorization: authorization_from(response),
+          test: test?,
+          error_code: error_code_from(response)
+        )
+      end
+
+      def api_request(action, post, method)
+        begin
+          raw_response = ssl_request(method, live_url + action, post.to_json, auth_headers)
+        rescue ResponseError => e
+          raw_response = e.response.body
+        end
+
+        response = parse(raw_response)
+        # Multiple invalid parameters
+        response = format_multiple_errors(response) if raw_response.include?('error_code') && response.is_a?(Array)
+
+        return response.symbolize_keys
+      end
+
+      def parse(body)
+        JSON.parse(body)
+      end
+
+      def success_from(response)
+        response[:status] == 'PENDING'
+      end
+
+      def message_from(response)
+        return response[:status] if success_from(response)
+
+        response[:message] || response[:error_message]
+      end
+
+      def authorization_from(response)
+        response[:id]
+      end
+
+      def auth_headers
+        {
+          'Content-Type' => 'application/json',
+          'Authorization' => "Bearer #{options[:access_token]}"
+        }
+      end
+
+      def error_code_from(response)
+        response[:error_code] unless success_from(response)
+      end
+
+      def format_multiple_errors(responses)
+        errors = responses.map do |response|
+          { error_code: response['error_code'], param: response['param'] }
+        end
+
+        {
+          error_code: STANDARD_ERROR_CODE_MAPPING[:multiple_invalid_parameters],
+          message: 'Validation error',
+          errors: errors
+        }
+      end
+    end
+  end
+end
--- a/test/remote/gateways/remote_sum_up_test.rb
+++ b/test/remote/gateways/remote_sum_up_test.rb
@@ -1,0 +1,112 @@
+require 'test_helper'
+
+class RemoteSumUpTest < Test::Unit::TestCase
+  def setup
+    @gateway = SumUpGateway.new(fixtures(:sum_up))
+
+    @amount = 100
+    @credit_card = credit_card('4000100011112224')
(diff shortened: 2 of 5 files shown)


lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,192 @@
...
+          gsub(%r(("pay_to_email\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("number\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("cvv\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
+      end
+
+      private
+
+      def upsert_checkout(money, payment, options = {})
+        if options[:checkout_id].present?
+          find_checkout(options[:checkout_id])
```
Heavyblade, line 50: ❓ Question:

What is the point of finding a checkout to extract it's ID if you already have the ID on the `options[:checkout_id]` ?

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,192 @@
...
+
+      def add_invoice(post, money, options)
+        post[:checkout_reference] = options[:checkout_reference]
+        post[:amount]             = amount(money)
+        post[:currency]           = (options[:currency] || currency(money))
+        post[:description]        = options[:description]
+      end
+
+      def add_payment(post, payment, options)
+        post[:payment_type] = options[:payment_type] if PAYMENT_TYPES.include?(options[:payment_type])
```
Heavyblade, line 115: ❓ Question:

Wouldn't be possible to identify the 'payment_type' from the card brand in conjunction with the payment method class ?

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,192 @@
...
+        card                = post[:card] = {}
+        card[:name]         = [payment.first_name, payment.last_name].join(' ')
```
Heavyblade, line 117: 💬 Comment:

Maybe just using the `ActiveMerchant::Billing::CreditCard#name` method?

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,192 @@
...
+
+      def trigger_request(action, post, method)
+        begin
+          response = parse(ssl_request(method, live_url + action, post.to_json, auth_headers))
+        rescue ResponseError => e
+          response = JSON.parse(e.response.body)
+        end
+
+        # Multiple invalid parameters
+        response = format_multiple_errors(response) if response.is_a?(Array)
```
Heavyblade, line 145: ❓ Question

`response.is_a?(Array)` is the best way to check if the response was successful or not ?

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,192 @@
...
+      end
+
+      def format_multiple_errors(responses)
+        errors = responses.map do |response|
+          { error_code: response['error_code'], param: response['param'] }
+        end
+
+        {
+          error_code: STANDARD_ERROR_CODE_MAPPING[:multiple_invalid_parameters],
+          message: 'Validation error',
```
Heavyblade, line 186: ❓ Question:

Any reason to set here explicitly that all errors are going to be validation errors?

Heavyblade (commented):
@sinourain first round Review is done, I'll leave the test review for the last one.

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: Good catch, you're right

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: I looked in the code and I don't see a way to do it, maybe in another instance it could work, not in this gem

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: You're right, I'm going to change this behavior.

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: Yes, in the [SumUp documentation](https://developer.sumup.com/docs/api/process-a-checkout/) the error 400 (Bad Request) have a response for Multiple_Invalid_Parameters all of them are `Vaidation error`.

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,192 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class SumUpGateway < Gateway
+      self.live_url = 'https://api.sumup.com/v0.1/'
```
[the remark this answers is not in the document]
sinourain: Yes, only one

lib/active_merchant/billing/gateways/sum_up.rb
Heavyblade: One idea => 

Well we could confirm with Charlotte/Nick but instead of thinking that the merchant is going to send us a GSF with the proper value for `options[:payment_type]` we could just start with the type 'card' of have it as a fallback if no value is available on `options[:payment_type]`

lib/active_merchant/billing/gateways/sum_up.rb
Heavyblade: I mean => let's say that you are going to have 401 or 500 errors are they to be shown as "validation errors" any way ?

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      PAYMENT_TYPES = %(card boleto ideal sofort bancontact)
+
+      def initialize(options = {})
+        requires!(options, :token, :pay_to_email)
+        super
+      end
+
+      def purchase(money, payment, options = {})
```
Heavyblade, line 27: 💬 Comment/Refactor:

suggestion if you want:

```ruby
      def purchase(money, payment, options = {})
        MultiResponse.run do |r|
          r.process { create_checkout(money, payment, options) } unless options[:checkout_id].present?
          r.process { complete_checkout(options[:checkout_id] || r.params['id'], payment, options) }
        end
      end

```

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -0,0 +1,198 @@
...
+        post[:card] = {
+          name:         credit_card.name,
+          number:       credit_card.number,
+          expiry_month: format(credit_card.month, :two_digits),
+          expiry_year:  credit_card.year,
+          cvv:          credit_card.verification_value
+        }
+      end
+
+      def add_credit_card(payment)
```
Heavyblade, line 118: ❓ Question:

What is the difference between the `add_credit_card` method a just calling `clone` on the payment object?

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    assert_success response
+    assert_equal 'PENDING', response.message
+    assert_equal @options[:order_id], response.params['checkout_reference']
+    refute_empty response.params['id']
+    refute_empty response.params['transactions']
+    refute_empty response.params['transactions'].first['id']
+    assert_equal 'PENDING', response.params['transactions'].first['status']
+  end
+
+  def test_successful_purchase_with_existing_checkout
```
Heavyblade, line 43: ❓ Question:

Ohhhh this is super interesting my assumption was that for every authorize/purchase transaction would exist just one Checkout (given the checkout definition for SumUp), but it is allowing you to use an "already" used reference for another purchase.

Is that a reasonable use case ?

test/unit/gateways/sum_up_test.rb
```diff
@@ -0,0 +1,361 @@
...
+    response = @gateway.purchase(@amount, @credit_card, @options)
+
+    assert_success response
+
+    assert_equal 'PENDING', response.message
+    assert response.test?
+  end
+
+  def test_failed_purchase
+    @credit_card.year = 2020
```
Heavyblade, line 31: 💬 Comment:

GIven that `failed_complete_checkout_response` is fixed changing the year is not going to have any effect

test/unit/gateways/sum_up_test.rb
```diff
@@ -0,0 +1,361 @@
...
+    @amount = 100
+
+    @options = {
+      payment_type: 'card',
+      billing_address: address,
+      description: 'Store Purchase'
+    }
+  end
+
+  def test_successful_purchase
```
Heavyblade, line 19: 💬 Comment:

given that response is going to be fixed the only point of having a unit test here is to actually check how the request is being built, so you can confirm that values are in the proper place and format.

test/unit/gateways/sum_up_test.rb
```diff
@@ -0,0 +1,361 @@
...
+    response = @gateway.purchase(@amount, @credit_card, @options)
+
+    assert_failure response
+    assert_equal SumUpGateway::STANDARD_ERROR_CODE_MAPPING[:multiple_invalid_parameters], response.error_code
+  end
+
+  def test_scrub
+    assert @gateway.supports_scrubbing?
+    assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
+  end
```
Heavyblade, line 42: 💬 Comment:

Maybe so small test for some private methods like would ensure that adapter behavior is declared.

- success_from
- message_from
- authorization_from
- format_multiple_errors
- error_code_from

Heavyblade (commented):
@sinourain done with my final review, thanks for the quick response

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: Perfect, I'm going to ask and for now I'll leave `card` as the default value

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: No, according to the documentation there is only one type of Multiple_Invalid_Parameters (Array) so the scope is only for these errors (400 Bad Request)

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: It makes sense, I hadn't realized xD

test/remote/gateways/remote_sum_up_test.rb
sinourain: Yes, in SumUp when a checkout is created it can have several transactions, in fact when the checkout is created a payment method is not placed. I will ask...

naashton (changes requested):
This all looks really good. I just had a few minor suggestions.

sinourain:
> This all looks really good. I just had a few minor suggestions.

@naashton and @Heavyblade thanks for your feedback, I updated the requested changes and also I add an extra change related to the credential: `token`, I changed it to `access_token` in order to prevent collide with a reserved word

naashton:
> > This all looks really good. I just had a few minor suggestions.
> 
> @naashton and @Heavyblade thanks for your feedback, I updated the requested changes and also I add an extra change related to the credential: `token`, I changed it to `access_token` in order to prevent collide with a reserved word

Thanks @sinourain - this actually made me curious, why aren't we using the `client_id` and `client_secret` for authentication as the [SumUp docs](https://developer.sumup.com/docs/api/sum-up-rest-api/) describe? Can you share the documentation where using the `pay_to_email` authentication is explained?

sinourain:
> > > This all looks really good. I just had a few minor suggestions.
> > 
> > 
> > @naashton and @Heavyblade thanks for your feedback, I updated the requested changes and also I add an extra change related to the credential: `token`, I changed it to `access_token` in order to prevent collide with a reserved word
> 
> Thanks @sinourain, this actually made me curious - Why aren't we using the `client_id` and `client_secret` for authentication as the [SumUp docs](https://developer.sumup.com/docs/api/sum-up-rest-api/) describe? And can you share the documentation where using the `pay_to_email` authentication is explained?

Sure, we discuss it in the Endava team, basically [SumUp have two ways for Authorization, API keys and OAuth 2.0](https://developer.sumup.com/docs/online-payments/introduction/authorization/), We decided to implement API Keys because it made the most sense given that if we implemented OAuth 2.0 it required additional development and requesting permissions to be able to make a purchase.

About the field `pay_to_email` in the [documentation mention it in the Request Body](https://developer.sumup.com/docs/api/create-a-checkout/#create-a-checkout). Curiously, it recommends that we use `merchant_code` instead of the `pay_to_email` field but according to tests we did, it is not possible to omit the `pay_to_email` field.

naashton (approved):
Awesome work!

PR #4896, opened 2023-09-21
Rapyd: Fix cvv validation

javierpedrozaing:
Description
-------------------------
This is a super small commit to update the validations to not send cvv value

Unit test
-------------------------
Finished in 0.132965 seconds.

34 tests, 163 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

100% passed

Remote test
-------------------------
Finished in 79.856446 seconds.

40 tests, 115 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

97.5% passed

Rubocop
-------------------------

766 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -169,12 +169,8 @@ module ActiveMerchant #:nodoc:
         pm_fields[:expiration_month] = payment.month.to_s
         pm_fields[:expiration_year] = payment.year.to_s
         pm_fields[:name] = "#{payment.first_name} #{payment.last_name}"
-        pm_fields[:cvv] = payment.verification_value.to_s unless send_cvv_value?(payment, options) || valid_network_transaction_id?(options)
+        pm_fields[:cvv] = payment.verification_value.to_s unless valid_network_transaction_id?(options) || payment.verification_value.blank?
         add_stored_credential(post, options)
-      end
-
-      def send_cvv_value?(payment, options)
-        return (payment.verification_value.nil? || payment.verification_value&.empty?)
       end
 
       def send_customer_object?(options)
@@ -182,8 +178,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def valid_network_transaction_id?(options)
-        network_transaction_id = options[:network_tansaction_id] || options.dig(:stored_credential_options, :network_transaction_id)
-        return false if network_transaction_id.nil? || network_transaction_id.empty?
+        network_transaction_id = options[:network_tansaction_id] || options.dig(:stored_credential_options, :network_transaction_id) || options.dig(:stored_credential, :network_transaction_id)
+        return network_transaction_id.present?
       end
 
       def add_ach(post, payment, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -174,7 +174,7 @@ def add_creditcard(post, payment, options)
       end
 
       def send_cvv_value?(payment, options)
-        payment.verification_value.nil? || payment.verification_value&.empty? || (options[:stored_credential] && options[:stored_credential][:reason_type] == 'recurring')
+        (payment.verification_value.nil? || payment.verification_value&.empty?) && (options[:stored_credential] && options[:stored_credential][:reason_type] == 'recurring')
```
naashton, line 177: Rethinking this because this condition reverts us back to a state where we send an empty `cvv`. I actually think each of these conditions should be validated independently, so maybe something like this:

We might not even need to check if the reason type is `recurring` here. Thoughts?

naashton (changes requested):

lib/active_merchant/billing/gateways/rapyd.rb
naashton: I actually think this method can just validate the first half of the condition (cvv is not nil and not empty) and then add an `||` condition on line 172 that also checks if the `options[:network_transaction_id]` exists, in which case we also don't send `cvv`.

So, we don't want to rely on `recurring` to tell us to not send `cvv`, since the initial txn can be `recurring` and contain a `cvv` that we need to send because there is no `network_transaction_id`.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -169,18 +169,23 @@ def add_creditcard(post, payment, options)
...
         pm_fields[:expiration_year] = payment.year.to_s
         pm_fields[:name] = "#{payment.first_name} #{payment.last_name}"
-        pm_fields[:cvv] = payment.verification_value.to_s unless send_cvv_value?(payment, options)
+        pm_fields[:cvv] = payment.verification_value.to_s unless send_cvv_value?(payment, options) || valid_network_transaction_id?(options)
         add_stored_credential(post, options)
       end
 
       def send_cvv_value?(payment, options)
-        payment.verification_value.nil? || payment.verification_value&.empty? || (options[:stored_credential] && options[:stored_credential][:reason_type] == 'recurring')
+        return (payment.verification_value.nil? || payment.verification_value&.empty?)
```
javierpedrozaing, line 177: it means that not matter if is a recurring transaction ?

lib/active_merchant/billing/gateways/rapyd.rb
javierpedrozaing: oh I can see, your changes make sense

lib/active_merchant/billing/gateways/rapyd.rb
naashton: Sort of, I think we can just simplify our condition to not be concerned with whether or not it's `recurring` since the main values we are checking for are `verification_value` and `network_transaction_id`.

naashton (approved):

PR #4897, opened 2023-09-21
Braintree create credit card nonce

gasb150:
In order to enhance the TokenNonce class and enable the token noce generation for CrediCards, this PR adds the graphql query and inputs necessary for CreditCards


Heavyblade (commented):
@gasb150 nice implementation, there is nothing to say in the code present, but would be nice to have unit tests that cover the `build_nonce_request`, `token_from`, and the replacement to address keys logic.

naashton (approved):

naashton:
Open sesame

PR #4898, opened 2023-09-22
Element: Fix credit card name bug

almalee24:
Update bug that causes error is first_name and last_name are nil

Remote:
31 tests, 86 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
27 tests, 136 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


naashton (approved):
Nice easy change!

PR #4903, opened 2023-10-02
Litle: Update stored credentials

almalee24:
Update how stored credential transactions are handle.

Unit:
58 tests, 257 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
57 tests, 251 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -310,7 +310,15 @@ module ActiveMerchant #:nodoc:
       def add_auth_purchase_params(doc, money, payment_method, options)
         doc.orderId(truncate(options[:order_id], 24))
         doc.amount(money)
-        add_order_source(doc, payment_method, options)
+
+        if options.dig(:stored_credential, :initial_transaction) == false
+          # orderSource needs to be added at the top of doc and
+          # processingType near the end
+          source_for_subsequent_stored_credential_txns(doc, options)
+        else
+          add_order_source(doc, payment_method, options)
+        end
+
         add_billing_address(doc, payment_method, options)
         add_shipping_address(doc, payment_method, options)
         add_payment_method(doc, payment_method, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -409,16 +417,17 @@ def add_payment_method(doc, payment_method, options)
...
+        return unless stored_credential = options[:stored_credential]
 
-        if options[:stored_credential][:initial_transaction]
-          add_stored_credential_params_initial(doc, options)
+        if stored_credential[:initial_transaction]
+          add_stored_credential_for_initial_txn(doc, options)
         else
-          add_stored_credential_params_used(doc, options)
+          doc.processingType("#{stored_credential[:initiator]}InitiatedCOF") if stored_credential[:reason_type] == 'unscheduled'
+          doc.originalNetworkTransactionId(stored_credential[:network_transaction_id]) unless stored_credential[:initiator] == 'cardholder'
```
aenand, line 426:
```suggestion
          doc.originalNetworkTransactionId(stored_credential[:network_transaction_id]) if stored_credential[:initiator] == 'merchant'
```

[pedantic] Ask to use if instead of unless for read-ability

lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -310,7 +310,15 @@ def add_authentication(doc)
...
         doc.orderId(truncate(options[:order_id], 24))
         doc.amount(money)
-        add_order_source(doc, payment_method, options)
+
+        if options.dig(:stored_credential, :initial_transaction) == false
+          # orderSource needs to be added at the top of doc and
+          # processingType near the end
+          source_for_subsequent_stored_credential_txns(doc, options)
+        else
+          add_order_source(doc, payment_method, options)
```
aenand, line 319: order_source is a GSF, would this change make it so that a merchant would no longer be able to override our stored creds framework with the GSF for non-initial transactions now?

almalee24: Yes, that's correct and this has also been approved by Litle.

aenand: Is it possible to preserve `order_source ` being preferred? I'm worried this might be a breaking change since we are now preventing the merchant from overriding if they engage in our stored creds flow and it's the initial_transaction

almalee24: that makes sense, I'll update this part.

almalee24: Actually looking at this code again we don't allow customer to override order_source currently for stored_credentials, https://github.com/activemerchant/active_merchant/pull/4903/files#diff-05db1124b764e8e57c13d2fca5b757cfde7c92a9839255e4a829ef1f90f3d7fcL497

aenand: oh well that's good!

aenand (approved):
LGTM :shipit:

PR #4910, opened 2023-10-05
Close stale issues/PRs

aenand:
The ActiveMerchant repository often has old issues and PRs that have gone stale. This makes it hard
to keep track of new requests. This commit adds the `stale` Github Action to ActiveMerchant.

If an issue or PR goes more than 60 days without activity it will apply the `stale` label. Any issue or PR with a `stale` label for more than 7 days will be closed. Draft PRs are exempt from this. 

Docs: https://github.com/marketplace/actions/close-stale-issues

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8
+        with:
+          stale-issue-message: 'To provide a cleaner slate for the maintenance of the library, this PR/Issue is being labeled stale after 60 days without activity. It will be closed in 14 days unless you comment with an update regarding its applicability to the current build. Thank you!'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,18 @@
...
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8
+        with:
+          stale-issue-message: 'Closing issue due to lack of activity!'
```
bradbroge, line 16: Could we add `days-before-stale` and `days-before-close` as specific times, instead of relying on the hidden defaults of `60` and `7` by default?

bradbroge (approved):
LGTM!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,20 @@
...
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8
+        with:
+          stale-issue-message: 'Message to comment on stale issues. If none provided, will not mark issues stale'
```
bradbroge, line 16: On further reflection, we should probably update this and the `stale-pr-message` with something more AM specific.

curiousepic:
Merged in https://github.com/activemerchant/active_merchant/commit/1caae49693fe2bb8364e52182754b9e78b8f8c7d

PR #4917, opened 2023-10-16
GlobalCollect: Add support for 3DS exemptions

almalee24:
Remote
53 tests, 121 assertions, 10 failures, 1 errors, 0 pendings, 0 omissions, 0 notifications 79.2453% passed

Unit
46 tests, 234 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -40,6 +40,7 @@ module ActiveMerchant #:nodoc:
         add_creator_info(post, options)
         add_fraud_fields(post, options)
         add_external_cardholder_authentication_data(post, options)
+        add_threeds_exemption_data(post, options)
         commit(:post, :authorize, post, options: options)
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -19,6 +19,13 @@ class GlobalCollectGateway < Gateway
       self.money_format = :cents
       self.supported_cardtypes = %i[visa master american_express discover naranja cabal]
 
+      THREEDS_EXEMPTIONS = {
+        moto: 'MOTO',
+        mail: 'MAIL',
+        telephone: 'TELEPHONE',
+        ecommerce: 'ECOMMERCE'
```
aenand, line 26: I'm surprised there's `moto`, `telephone`, and `mail` since MOTO stands for `mail order telephone order`. Does the gateway specify when to use `moto` vs the other two?

almalee24: I know but it doesn't specify when to use which. Now I'm actually thinking about removing `ecommerce` since it says [that's the default value](https://apireference.connect.worldline-solutions.com/s2sapi/v1/en_US/dotnet/payments/create.html?paymentPlatform=GLOBALCOLLECT).

aenand: > MAIL - The transaction is a Mail Order.
MOTO - The transaction is a Mail Order/Telephone Order.
TELEPHONE - The transaction is a Telephone Order.

Did the merchant specifically reqest to add `MAIL` and `TELEPHONE`? If not then I'd suggest we just add MOTO

almalee24: We want to support all available exemption types which is why I added `mail` and `telephone` but it does seem redundant with them also supporting MOTO

aenand (approved):
Code looks good to me :shipit: nice work! I'm wary about adding `mail` and `telephone` as exemption types when `moto` is supported but I'll defer to you 🚀

almalee24:
> Code looks good to me :shipit: nice work! I'm wary about adding `mail` and `telephone` as exemption types when `moto` is supported but I'll defer to you 🚀

I think you are right so I will remove mail and telephone

PR #4924, opened 2023-10-20
SumUp Gateway: Fix refund method

sinourain:
Description
-------------------------
Fix refund method to SumUp Gateway adapter.

This are the relevant links to review the implementation:

- [Refund a transaction](https://developer.sumup.com/docs/api/refund-transaction/)

Tickets for Spreedly reference
SER-836

Note: SumUp has shared with us an account to test with which we can refund a purchase

Unit test
-------------------------
Finished in 32.469516 seconds.
5638 tests, 78183 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote test
-------------------------
173.64 tests/s, 2407.89 assertions/s

Rubocop
-------------------------
773 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/sum_up.rb
+++ b/lib/active_merchant/billing/gateways/sum_up.rb
@@ -35,9 +35,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def refund(money, authorization, options = {})
-        transaction_id = authorization.split('#')[-1]
-        payment_currency = options[:currency] || currency(money)
-        post = money ? { amount: localized_amount(money, payment_currency) } : {}
+        transaction_id = authorization.split('#').last
+        post = money ? { amount: amount(money) } : {}
         add_merchant_data(post, options)
 
         commit('me/refund/' + transaction_id, post)
@@ -106,10 +105,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_invoice(post, money, options)
-        payment_currency = options[:currency] || currency(money)
         post[:checkout_reference] = options[:order_id]
-        post[:amount]             = localized_amount(money, payment_currency)
-        post[:currency]           = payment_currency
+        post[:amount]             = amount(money)
+        post[:currency]           = options[:currency] || currency(money)
         post[:description]        = options[:description]
       end
 
@@ -127,29 +125,31 @@ module ActiveMerchant #:nodoc:
 
       def commit(action, post, method = :post)
         response = api_request(action, post.compact, method)
+        succeeded = success_from(response)
 
         Response.new(
-          success_from(response),
-          message_from(response),
-          response,
+          succeeded,
+          message_from(succeeded, response),
+          action.include?('refund') ? { response_code: response.to_s } : response,
           authorization: authorization_from(response),
           test: test?,
-          error_code: error_code_from(response)
+          error_code: error_code_from(succeeded, response)
         )
       end
 
       def api_request(action, post, method)
-        begin
-          raw_response = ssl_request(method, live_url + action, post.to_json, auth_headers)
-        rescue ResponseError => e
-          raw_response = e.response.body
-        end
-
+        raw_response =
+          begin
+            ssl_request(method, live_url + action, post.to_json, auth_headers)
+          rescue ResponseError => e
+            e.response.body
+          end
         response = parse(raw_response)
-        # Multiple invalid parameters
-        response = format_multiple_errors(response) if raw_response.include?('error_code') && response.is_a?(Array)
+        response = response.is_a?(Hash) ? response.symbolize_keys : response
 
-        return response.symbolize_keys
+        return format_errors(response) if raw_response.include?('error_code') && response.is_a?(Array)
+
+        response
       end
 
       def parse(body)
@@ -157,6 +157,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(response)
+        return true if response == 204
+
         return false unless %w(PENDING EXPIRED PAID).include?(response[:status])
 
         response[:transactions].each do |transaction|
@@ -166,13 +168,19 @@ module ActiveMerchant #:nodoc:
         true
       end
 
-      def message_from(response)
-        return response[:status] if success_from(response)
+      def message_from(succeeded, response)
+        if succeeded
+          return 'Succeeded' if response.is_a?(Integer)
+
+          return response[:status]
+        end
 
         response[:message] || response[:error_message]
       end
 
       def authorization_from(response)
+        return nil if response.is_a?(Integer)
+
         return response[:id] unless response[:transaction_id]
 
         [response[:id], response[:transaction_id]].join('#')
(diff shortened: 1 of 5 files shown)


test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -123,6 +123,26 @@ def test_failed_void_invalid_checkout_id
     assert_equal 'Resource not found', response.message
   end
 
+  def test_successful_refund
+    gateway = SumUpGateway.new(fixtures(:sum_up_testing))
```
javierpedrozaing, line 127: Maybe we can find a better name and include it in the fixtures.yml file, to keep in mind that we need to use this credentials as well to run remote test.

javierpedrozaing (commented):
LGTM just I left one comment for you consideration. 👍🏻

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -157,7 +159,9 @@ def parse(body)
       end
 
       def success_from(response)
-        return false unless %w(PENDING EXPIRED PAID).include?(response[:status])
+        return false unless %w(PENDING EXPIRED PAID REFUND).include?(response[:status])
+
+        return true if response[:transactions].blank?
```
javierpedrozaing, line 164: since you are returning a proper response for `refund`, I think this `return true `is not needed, unless for other transaction we have empty responses.

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -139,11 +137,15 @@ def commit(action, post, method = :post)
       end
 
       def api_request(action, post, method)
-        begin
-          raw_response = ssl_request(method, live_url + action, post.to_json, auth_headers)
-        rescue ResponseError => e
-          raw_response = e.response.body
-        end
+        raw_response =
```
Heavyblade, line 140: 💬 Style/comment:

A couple of things about this method:

1) `build_response` sounds like an action-method seems me that doesn't represent the actual content of the variable.
2) Is not super bad, but I'm not sure if is a good idea to be creative and invent a response that was not delivered by the gateway itself, maybe at some point the customer is going to go to support with a transcript and this text/message was never part of the gateway's response.

3) `format_multiple_errors` returns a structured error, if the gateway returns just 1 error, that is structure fits the one returned by `format_multiple_errors` something like this => 

```ruby
{
  error_code: "xxxx",
  message: 'xxxxxxxx,
  errors: errors
}
```

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -123,6 +123,26 @@ def test_failed_void_invalid_checkout_id
...
+    purchase = gateway.purchase(@amount, @credit_card, @options)
+    assert_success purchase
+    assert_equal 'PAID', purchase.message
```
Heavyblade, line 130: 💬 Comment:

Do we need all those assertions on the part of the code that is just the context/preparation for the actual test that is going to be `refund`?, there are 7 assertions before the refund request is actually fired and that request only has 2 assertions, are we testing `purchase` or `refund`?

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -123,6 +123,26 @@ def test_failed_void_invalid_checkout_id
     assert_equal 'Resource not found', response.message
   end
 
+  def test_successful_refund
```
Heavyblade, line 126: 💬 Comment:

Would be great to actually a test that covers a partial refund.

Heavyblade (commented):
@sinourain left some complementary PR comments

naashton (changes requested):
Just a few comments, generally looks good though 👍

naashton:
@sinourain - just a nudge for you to respond to my review

sinourain:
> @sinourain - just a nudge for you to respond to my review

thank you.  With so many things I had not seen your comments.  I'm going to review them.

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -166,13 +168,19 @@ def success_from(response)
...
+          return 'SUCCESSFUL' if response.is_a?(Integer)
+
+          return response[:status]
+        end
 
         response[:message] || response[:error_message]
       end
 
       def authorization_from(response)
+        return nil if response.is_a?(Integer)
```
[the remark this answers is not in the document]
sinourain: Yes, we have a 201 without body response when refund a transaction

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -123,6 +124,30 @@ def test_failed_void_invalid_checkout_id
     assert_equal 'Resource not found', response.message
   end
 
+  # In Sum Up the account can only return checkout/purchase in pending or success status,
+  # to obtain a successful refund we will need an account that returns the checkout/purchase in successful status
+  #
+  # For this example configure in the fixtures => :sum_up_account_for_successful_purchases
+  def test_successful_refund
+    gateway = SumUpGateway.new(fixtures(:sum_up_account_for_successful_purchases))
```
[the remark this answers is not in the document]
sinourain: Done!

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -166,13 +168,19 @@ def success_from(response)
         true
       end
 
-      def message_from(response)
-        return response[:status] if success_from(response)
+      def message_from(succeeded, response)
+        if succeeded
+          return 'SUCCESSFUL' if response.is_a?(Integer)
```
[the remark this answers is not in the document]
sinourain: Done!

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -1,3 +1,4 @@
+require 'pry'
```
naashton, line 1: You can remove this `require` statement

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -123,6 +124,30 @@ def test_failed_void_invalid_checkout_id
...
   end
 
+  # In Sum Up the account can only return checkout/purchase in pending or success status,
+  # to obtain a successful refund we will need an account that returns the checkout/purchase in successful status
+  #
+  # For this example configure in the fixtures => :sum_up_account_for_successful_purchases
+  def test_successful_refund
+    gateway = SumUpGateway.new(fixtures(:sum_up_account_for_successful_purchases))
+    purchase = gateway.purchase(@amount, @credit_card, @options)
+    transaction_id = purchase.params['transaction_id']
```
naashton, line 134: We should `assert` the transaction_id here, otherwise the test throws an error if the value is `nil`

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -123,6 +124,30 @@ def test_failed_void_invalid_checkout_id
...
+
+    response = gateway.refund(@amount, transaction_id, {})
+    assert_success response
+    assert_equal 'Succeeded', response.message
+  end
+
+  def test_successful_partial_refund
+    gateway = SumUpGateway.new(fixtures(:sum_up_account_for_successful_purchases))
+    purchase = gateway.purchase(@amount * 10, @credit_card, @options)
+    transaction_id = purchase.params['transaction_id']
```
naashton, line 144: Same as above

naashton (changes requested):
Just a few requests to change in the tests, but everything else looks good!

test/remote/gateways/remote_sum_up_test.rb
sinourain: Done!

test/remote/gateways/remote_sum_up_test.rb
sinourain: Done!

PR #4928, opened 2023-10-23
Orbital: Enable Third Party Vaulting

javierpedrozaing:
Description
-------------------------
[SER-782](https://spreedly.atlassian.net/browse/SER-782)

This commit enable third party vaulting for Orbital in this case Orbital needs a flag to enable the store process and it doesn't support unstore function.

Orbital does not have a store and unstore endpoint. We are enabling the store process through a Spreedly store endpoint, but behind the scenes, we are performing a $0 authorization and sending the tokenTxType."

Unit test
-------------------------
Finished in 1.849371 seconds.

147 tests, 831 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

79.49 tests/s, 449.34 assertions/s

Remote test
-------------------------
Finished in 101.551157 seconds.

126 tests, 509 assertions, 9 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 92.8571% passed

1.24 tests/s, 5.01 assertions/s

Rubocop
-------------------------
773 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -193,6 +193,10 @@ module ActiveMerchant #:nodoc:
         'checking' => 'C'
       }
 
+      # safetech token flags
+      GET_TOKEN = 'GT'
+      USE_TOKEN = 'UT'
+
       def initialize(options = {})
         requires!(options, :merchant_id)
         requires!(options, :login, :password) unless options[:ip_authentication]
@@ -251,6 +255,11 @@ module ActiveMerchant #:nodoc:
         end
 
         commit(order, :refund, options[:retry_logic], options[:trace_number])
+      end
+
+      # Orbital save a payment method if the TokenTxnType is 'GT', that's why we use this as the default value for store
+      def store(creditcard, options = {})
+        authorize(0, creditcard, options.merge({ token_txn_type: GET_TOKEN }))
       end
 
       def void(authorization, options = {}, deprecated = {})
@@ -536,9 +545,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def billing_name(payment_source, options)
-        if payment_source&.name.present?
+        if !payment_source.is_a?(String) && payment_source&.name.present?
           payment_source.name[0..29]
-        elsif options[:billing_address][:name].present?
+        elsif options[:billing_address] && options[:billing_address][:name].present?
           options[:billing_address][:name][0..29]
         end
       end
@@ -555,10 +564,16 @@ module ActiveMerchant #:nodoc:
         options[:billing_address] || options[:address]
       end
 
+      def add_safetech_token_data(xml, payment_source, options)
+        xml.tag! :CardBrand, options[:card_brand]
+        xml.tag! :AccountNum, payment_source
+      end
+
       #=====PAYMENT SOURCE FIELDS=====
 
       # Payment can be done through either Credit Card or Electronic Check
       def add_payment_source(xml, payment_source, options = {})
+        add_safetech_token_data(xml, payment_source, options) if options[:token_txn_type] == USE_TOKEN
         payment_source.is_a?(Check) ? add_echeck(xml, payment_source, options) : add_credit_card(xml, payment_source, options)
       end
 
@@ -576,10 +591,10 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_credit_card(xml, credit_card, options)
-        xml.tag! :AccountNum, credit_card.number if credit_card
-        xml.tag! :Exp, expiry_date(credit_card) if credit_card
+        xml.tag! :AccountNum, credit_card.number if credit_card.is_a?(CreditCard)
+        xml.tag! :Exp, expiry_date(credit_card) if credit_card.is_a?(CreditCard)
         add_currency_fields(xml, options[:currency])
-        add_verification_value(xml, credit_card) if credit_card
+        add_verification_value(xml, credit_card) if credit_card.is_a?(CreditCard)
       end
 
       def add_verification_value(xml, credit_card)
@@ -886,12 +901,14 @@ module ActiveMerchant #:nodoc:
             request.call(remote_url(:secondary))
           end
 
+        authorization = order.include?('<TokenTxnType>GT</TokenTxnType>') ? response[:safetech_token] : authorization_string(response[:tx_ref_num], response[:order_id])
+
         Response.new(
           success?(response, message_type),
           message_from(response),
           response,
           {
-            authorization: authorization_string(response[:tx_ref_num], response[:order_id]),
+            authorization: authorization,
             test: self.test?,
             avs_result: OrbitalGateway::AVSResult.new(response[:avs_resp_code]),
             cvv_result: OrbitalGateway::CVVResult.new(response[:cvv2_resp_code])
@@ -1014,6 +1031,7 @@ module ActiveMerchant #:nodoc:
             add_stored_credentials(xml, parameters)
             add_pymt_brand_program_code(xml, payment_source, three_d_secure)
             add_mastercard_fields(xml, payment_source, parameters, three_d_secure) if mastercard?(payment_source)
+            xml.tag! :TokenTxnType, parameters[:token_txn_type] if parameters[:token_txn_type]
           end
         end
         xml.target!
--- a/test/fixtures.yml
+++ b/test/fixtures.yml
@@ -727,6 +727,11 @@ orbital_tandem_gateway:
   password: PASSWORD
   merchant_id: MERCHANTID
 
+orbital_tpv_gateway:
+  login: LOGIN
+  password: PASSWORD
+  merchant_id: MERCHANTID
+
 # Working credentials, no need to replace
 pagarme:
     api_key: 'ak_test_e1QGU2gL98MDCHZxHLJ9sofPUFJ7tH'
--- a/test/remote/gateways/remote_orbital_test.rb
+++ b/test/remote/gateways/remote_orbital_test.rb
@@ -6,10 +6,11 @@ class RemoteOrbitalGatewayTest < Test::Unit::TestCase
     @gateway = ActiveMerchant::Billing::OrbitalGateway.new(fixtures(:orbital_gateway))
     @echeck_gateway = ActiveMerchant::Billing::OrbitalGateway.new(fixtures(:orbital_asv_aoa_gateway))
     @three_ds_gateway = ActiveMerchant::Billing::OrbitalGateway.new(fixtures(:orbital_3ds_gateway))
-
+    @tpv_orbital_gateway = ActiveMerchant::Billing::OrbitalGateway.new(fixtures(:orbital_tpv_gateway))
     @amount = 100
     @google_pay_amount = 10000
     @credit_card = credit_card('4556761029983886')
+    @mastercard_card_tpv = credit_card('2521000000000006')
     @declined_card = credit_card('4000300011112220')
     # Electronic Check object with test credentials of saving account
     @echeck = check(account_number: '072403004', account_type: 'savings', routing_number: '072403004')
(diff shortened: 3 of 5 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -682,6 +688,10 @@ def add_mastercard_fields(xml, credit_card, parameters, three_d_secure)
         add_mc_ucafind(xml, credit_card, three_d_secure)
       end
 
+      def add_third_party_token(xml, parameters)
```
gasb150, line 691: Same here, looks like is not used in the method the parameter `parameters`.

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -1014,6 +1024,8 @@ def build_new_order_xml(action, money, payment_source, parameters = {})
             add_stored_credentials(xml, parameters)
             add_pymt_brand_program_code(xml, payment_source, three_d_secure)
             add_mastercard_fields(xml, payment_source, parameters, three_d_secure) if mastercard?(payment_source)
+            add_third_party_token(xml, parameters) if parameters[:store]
+            add_token_transaction_type(xml, parameters) if parameters[:transaction_token]
```
gasb150, line 1028: Is it possible to have parameters with `:store` and `:transaction_token`? 

What happens in TokenTxnType in that case?

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -118,7 +118,7 @@ def setup
     ]
   end
 
-  def test_successful_purchase
+  def test_successful_purchase_1
```
gasb150, line 121: why rename this test?

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -555,10 +555,16 @@ def get_address(options)
         options[:billing_address] || options[:address]
       end
 
+      def add_token_transaction_type(xml, options)
```
gasb150, line 558: 🤔 opinion:
It could be good to be consistent using parameters like the other methods.

gasb150, line 558: Is it necessary to use options?
Looks like you are using by default 'UT' as a TokenTxnType.

gasb150 (commented):
@javierpedrozaing, Looks good the implementation, great job.

I just left a few comments for you to take into consideration.

lib/active_merchant/billing/gateways/orbital.rb
gasb150: Looks like `options` is not necessary for the method.

javierpedrozaing: the name is because we are setting the `TokenTxnType` value, and you are right the options is not neccesary, I will update it, thanks

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: yes I will update it, parameters is not needed, thanks

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: the `store` would be a GSF to indicate that we are going to prcoess with a store, for Orbita is set `TokenTxnType` to 'GT'
the `transaction_token` is to send the token stored.

test/remote/gateways/remote_orbital_test.rb
javierpedrozaing: oh It's my bad, 🤦🏻‍♂️  doing testing, becuase there is another `test_successful_purchase` function on the line 1238, for `TandemOrbitalTests`. thanks

lib/active_merchant/billing/gateways/orbital.rb
gasb150: What happens if both values come in the parameters hash?

jherreraa (commented):
No additional comments besides the @gasb150.
Except that, everything else looks good to me. 👍

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -1014,6 +1024,8 @@ def build_new_order_xml(action, money, payment_source, parameters = {})
             add_stored_credentials(xml, parameters)
             add_pymt_brand_program_code(xml, payment_source, three_d_secure)
             add_mastercard_fields(xml, payment_source, parameters, three_d_secure) if mastercard?(payment_source)
+            add_third_party_token(xml, parameters) if parameters[:store]
```
gasb150, line 1027: ❓ Question: 
Why are you using a GSF store? `Core` currently allows you to process without that.
💭  Suggestion:
You could create a store method, where this `add_third_party_token` is called, instead of verifying a store GSF.

lib/active_merchant/billing/gateways/orbital.rb
gasb150, line 558: 💭 Suggestion.

Instead of defining a method to add a single parameter,  you could add it directly.

test/fixtures.yml
```diff
@@ -726,6 +726,11 @@ orbital_tandem_gateway:
   password: PASSWORD
   merchant_id: MERCHANTID
 
+orbital_tpv_gateway:
```
gasb150, line 729: ❓ question:
Is it necessary to add an extra entry in fixtures to test TPV? or if you update the current entry `orbital_gateway` with the TPV credentials the other non-TPV test will fail?

lib/active_merchant/billing/gateways/orbital.rb
gasb150: Same suggestion here, related to defining a method for a single parameter.

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -829,6 +829,37 @@ def test_successful_verify
     assert_equal 'No reason to decline', response.message
   end
 
+  def test_successful_store
+    options = @options.merge(store: true)
```
gasb150, line 833: considering the previous comments, this test looks like a purchase with an option to store instead of a normal store. 

If you add use a store method instead of a purchase I think this test will change.

gasb150 (commented):
Hey @javierpedrozaing,  I'm including more comments for you to consider.

I also have a doubt. If the idea is to be able to store pm (TPV), why is it not a store method or an authorization_from modification?

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: yes, you're right, since the gateway doesn't have a store endpoint, we can have two approach,  in this case in order to align us with the way Orbital handle store, we can send it as a GSF, from the Spreedly perspective we can handle the store method, and thinking about the documentation could be make sense use a store method 🤔 , I will check it

test/fixtures.yml
javierpedrozaing: Yes thanks, I was wondering about that, because the TPV test works if I update the `orbital_gateway` but is not clear for me,  I don't sure, because anyway if we want to run all the remote test I would need to include the `orbital_tpv_gateway` credentials, then I think that is good to have this credentials. What do you think ?

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -886,12 +896,14 @@ def commit(order, message_type, retry_logic = nil, trace_number = nil)
             request.call(remote_url(:secondary))
           end
 
+        authorization = order.include?('<TokenTxnType>GT</TokenTxnType>') ? response[:safetech_token] : authorization_string(response[:tx_ref_num], response[:order_id])
```
edgarv09, line 899: 👗  Style:
This is fine but looks a bit complex due to the length of the line. Using a conventional if-else block might enhance readability but is up to you. 

```ruby
token_required = order.include?('<TokenTxnType>GT</TokenTxnType>')
authorization = if token_required
  response[:safetech_token]
else
  authorization_string(response[:tx_ref_num], response[:order_id])
end
```

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -829,6 +830,96 @@ def test_successful_verify
...
+  def test_successful_store
+    response = @tpv_orbital_gateway.store(@mastercard_card_tpv, @options)
+    assert_success response
+    assert_false response.params['safetech_token'].blank?
+  end
+
+  def test_successful_purchase_stored_token
+    store = @tpv_orbital_gateway.store(@credit_card, @options)
+    assert_success store
+    response = @tpv_orbital_gateway.purchase(@amount, store.authorization, @options.merge(card_brand: 'VI', token_txn_type: 'UT'))
```
edgarv09, line 842: it would be helpful to include more descriptive arguments in tests to provide guidance for future developers or even to us. Why?, same as above, it is not clear what "UT" means and the effects that value has.

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -253,6 +253,10 @@ def credit(money, payment_method, options = {})
         commit(order, :refund, options[:retry_logic], options[:trace_number])
       end
 
+      def store(creditcard, options = {})
+        authorize(0, creditcard, options.merge({ token_txn_type: 'GT' }))
+      end
```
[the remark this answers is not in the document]
javierpedrozaing: since the store process for Orbital happen if the token_txn_type is 'GT' I have added 'GT' as the default value because in this case it's a store transaction.
this is sended via options because is a GSF. and this should be setted by the user for a authorization transaction with 'UT' .
it's a good idea use a constant for this value, thanks

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -555,10 +559,16 @@ def get_address(options)
...
+      def add_safetech_token_data(xml, payment_source, options)
+        xml.tag! :CardBrand, options[:card_brand]
+        xml.tag! :AccountNum, payment_source
+      end
+
       #=====PAYMENT SOURCE FIELDS=====
 
       # Payment can be done through either Credit Card or Electronic Check
       def add_payment_source(xml, payment_source, options = {})
+        add_safetech_token_data(xml, payment_source, options) if options[:token_txn_type] == 'UT'
```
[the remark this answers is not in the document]
javierpedrozaing: in this case we need to send the `token_txn_type` value as `UT` in order to indicate to Orbital we want to use the safetech token stored durgin the authorize o purchase transaction. 
I will add two constant to represent 'UT' and 'GT' values, thanks

test/remote/gateways/remote_orbital_test.rb
javierpedrozaing: yes, I will add some comment to clarify

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -886,12 +901,14 @@ def commit(order, message_type, retry_logic = nil, trace_number = nil)
             request.call(remote_url(:secondary))
           end
 
+        authorization = order.include?('<TokenTxnType>GT</TokenTxnType>') ? response[:safetech_token] : authorization_string(response[:tx_ref_num], response[:order_id])
```
naashton, line 904: @gasb150 and @edgarv09's questions got me thinking about this - should we just append the `safetech_token` to the `authorization_string`? I'm thinking something like what we do in [Adyen](https://github.com/activemerchant/active_merchant/blob/6580e97c12b8b629303656e42c70c0d6d799808e/lib/active_merchant/billing/gateways/adyen.rb#L855-L863).

If the gateway is returning `tx_ref_num` and `order_id`, I feel we should capture those values and append the `safetech_token` to the `;` delimited string, then parse it out when we need it.

naashton (changes requested):
I have a question about the implementation, but all of the changes here look good.

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: Hi @naashton I think this is a special case, because of authorization value should match with the `safetech_token` stored by Orbital, I mean, when we send an authorization/purchase the `safetech_token` value is sent int the `cardNumber` field, then if we use this token along with `tx_ref_num` and `order_id`, we would get the next error.
`"Error. The Orbital Gateway has received a badly formatted message. Field Card Number exceeded max length of 19"`

naashton (approved):

PR #4929, opened 2023-10-24
Worldpay: Update 3ds logic to accept df_reference_id directly

DustinHaefele:
unit tests:
-------------------------------------------------------------------------------------------------------------------------
113 tests, 645 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
--------------------------------------------------------------------------------------------------------------------------426.14 tests/s, 2432.37 assertions/s

remote tests:
--------------------------------------------------------------------------------------------------------------------------
100 tests, 412 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
--------------------------------------------------------------------------------------------------------------------------

ticket for spreedly tracking: ECS-3125

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -443,7 +443,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_additional_3ds_data(xml, options)
-        additional_data = { 'dfReferenceId' => options[:session_id] }
+        additional_data = { 'dfReferenceId' => options[:df_reference_id] }
         additional_data['challengeWindowSize'] = options[:browser_size] if options[:browser_size]
 
         xml.additional3DSData additional_data
--- a/test/remote/gateways/remote_worldpay_test.rb
+++ b/test/remote/gateways/remote_worldpay_test.rb
@@ -455,9 +455,8 @@ class RemoteWorldpayTest < Test::Unit::TestCase
     )
     assert first_message = @gateway.authorize(@amount, @threeDS_card, options)
     assert first_message.test?
+    assert first_message.success?
     refute first_message.authorization.blank?
-    refute first_message.params['issuer_url'].blank?
-    refute first_message.params['pa_request'].blank?
     refute first_message.params['cookie'].blank?
     refute first_message.params['session_id'].blank?
   end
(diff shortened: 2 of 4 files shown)


test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -455,9 +455,8 @@ def test_successful_authorize_with_3ds
     )
     assert first_message = @gateway.authorize(@amount, @threeDS_card, options)
     assert first_message.test?
+    assert first_message.success?
     refute first_message.authorization.blank?
-    refute first_message.params['issuer_url'].blank?
```
DustinHaefele, line 459: worldpay does not seem to send these two parameters back in their responses, nor can I find reference to them in their docs (though I'm sure they used to send them back?) That being said the purpose of the test is to prove successful transactions which I believe adding the `.success?` line above covers.

test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -869,7 +865,7 @@ def test_successful_fast_fund_credit_with_token_on_cft_gateway
 
   def test_failed_fast_fund_credit_on_cft_gateway
     options = @options.merge({ fast_fund_credit: true })
-    refused_card = credit_card('4917300800000000', name: 'REFUSED') # 'magic' value for testing failures, provided by Worldpay
+    refused_card = credit_card('4444333322221111', name: 'REFUSED') # 'magic' value for testing failures, provided by Worldpay
```
DustinHaefele, line 868: This updated test card number is the one listed in the worldpay docs for fast_fund testing. Test was failing with the previous number.

aenand (approved):
LGTM :shipit: nice work!

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -443,7 +443,7 @@ def add_token_for_ff_credit(xml, payment_method, options)
       end
 
       def add_additional_3ds_data(xml, options)
-        additional_data = { 'dfReferenceId' => options[:session_id] }
+        additional_data = { 'dfReferenceId' => options[:df_reference_id] }
```
aenand, line 446: Should we fall back to `session_id` if `df_reference_id` is not sent in the options?

DustinHaefele: The problem is the session_id option will be a different value then we are sending today once these PRs are merged. So we can certainly do it, but it won't be the id we are currently passing here.

aenand (approved):
LGTM 🚀

PR #4934, opened 2023-10-27
CyberSource: surface the reconciliationID2 field if present

yunnydang:
This is to help us parse out the second reconciliationID since it is sometimes returned in the second response.

Local:
5651 tests, 78200 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6638% passed

Unit:
142 tests, 794 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
128 tests, 639 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.0938% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -1161,6 +1161,7 @@ module ActiveMerchant #:nodoc:
             parent += '_' + node.parent.attributes['id'] if node.parent.attributes['id']
             parent += '_'
           end
+          reply[:reconciliationID2] = node.text if node.name == 'reconciliationID' && reply[:reconciliationID]
           reply["#{parent}#{node.name}".to_sym] ||= node.text
         end
         return reply
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -493,6 +493,12 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
     assert_successful_response(response)
   end
 
+  def test_successful_purchase_with_reconciliation_id_2
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_successful_response(response)
+    assert response.params['reconciliationID2']
+  end
+
   def test_successful_authorize_with_customer_id
     options = @options.merge(customer_id: '7500BB199B4270EFE05348D0AFCAD')
     assert response = @gateway.authorize(@amount, @credit_card, options)
--- a/test/unit/gateways/cyber_source_test.rb
+++ b/test/unit/gateways/cyber_source_test.rb
@@ -568,6 +568,24 @@ class CyberSourceTest < Test::Unit::TestCase
     assert response.test?
   end
 
+  def test_successful_reconciliation_id_2
+    @gateway.stubs(:ssl_post).returns(successful_purchase_and_capture_response)
+    assert response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_equal response.params['reconciliationID'], 'abcdf'
+    assert_equal response.params['reconciliationID2'], '31159291T3XM2B13'
+    assert response.success?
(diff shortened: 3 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -1161,6 +1161,9 @@ def parse_element(reply, node)
             parent += '_' + node.parent.attributes['id'] if node.parent.attributes['id']
             parent += '_'
           end
+          if node.name == 'reconciliationID' && reply[:reconciliationID]
+            reply[:reconciliationID2] = node.text unless reply[:reconciliationID] == node.text
```
jcreiff, line 1165: This `unless` clause is probably a smart guard to include; have you seen any transactions that suggest the same ID would be included twice?

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -493,6 +493,12 @@ def test_successful_purchase_with_reconciliation_id
     assert_successful_response(response)
   end
 
+  def test_successful_purchase_with_reconciliation_id_2
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_successful_response(response)
+    assert_not_equal response.params['reconciliationID'], response.params['reconciliationID2']
```
jcreiff, line 499: Can you also add an assertion here for the presence of `reconciliationID2`?

test/unit/gateways/cyber_source_test.rb
```diff
@@ -568,6 +568,15 @@ def test_successful_auth_request
     assert response.test?
   end
 
+  def test_successful_reconciliation_id_2
```
jcreiff, line 571: Out of an abundance of caution to prove that we are not breaking anything, can you also add a test that shows in the case there is only one reconciliationID present, it is still returned, and reconciliationID2 is not added to the response 😅 

I think the `successful_authorization_response` should work for this use case.

test/remote/gateways/remote_cyber_source_test.rb
jcreiff: (This would still pass if it was `nil`)

jcreiff (changes requested):
Overall looks good, just a couple of small testing addition suggestions!

lib/active_merchant/billing/gateways/cyber_source.rb
yunnydang: Hey @jcreiff yes I did notice some transactions did return the same reconciliationID twice in the authReply and captureReply. I thought it'd be a good idea to not double up if that were the case!

jcreiff: Interesting! Did you find those transactions "in the wild" or were they in sandbox/testing?

Thinking about it some more, I wonder if we actually want to surface both even if they are the same. If the expectation of an application consuming the response is that a purchase transaction returns two values for auth and capture reply, then the absence of the second ID in the case where they are identical will "feel like" a bug (much like what we have now 😄).

Whereas if the duplicate ID is returned in two places, that's just more information that indicates that both IDs were in fact returned, but they were the same. What do you think?

yunnydang: Thanks @jcreiff yep, just by quickly checking out some transaction I did notice a handful of them had duplicates. I was thinking of this as well but I erred on the less noisy side of things. 

I was thinking that reconciliationID_2 will 9 times out of 10 probably be a duplicate given the small sample size of transactions I peeked at. No one's made a fuss about it not being returned prior so I think just surfacing it only when we have the off chance case of it not being a duplicate would be more useful to others. That's my reasoning behind it, but I totally get the 'bug' example you mentioned. I'm totally okay with implementing it to surface all the second ID every time if you feel strongly about it though, let me know!

jcreiff (approved):
👍🏻

PR #4935, opened 2023-10-27
Quickbooks: Remove OAuth from commit method

almalee24:
The OAuth response in commit is preventing the transaction from being retried if they failed with AuthenticationFailed because of invalid credentials or expired access tokens.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -44,6 +44,7 @@
 * GlobalCollect: Added support for 3DS exemption request field [almalee24] #4917
 * NMI: Update supported countries list [jcreiff] #4931
 * Adyen: Add mcc field [jcreiff] #4926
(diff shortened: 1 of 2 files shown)


bradbroge (approved):
👍🏻

kylene-spreedly (approved):

naashton:
Is there a test we can/should add to validate this change?

curiousepic:
The commit title doesn't quite seem to match what this change does?

curiousepic (approved):
otherwise 👍

almalee24:
> The commit title doesn't quite seem to match what this change does?

True, I'll update the commit title

almalee24:
> Is there a test we can/should add to validate this change?

I don't think it's necessary to do a test for this but I'll look into it.

almalee24:
> Is there a test we can/should add to validate this change?

Decided not to add a test so I will merge as is

PR #4936, opened 2023-10-27
Kushki: Add support for Partial Refunds

rachelkirk:
This updates the logic to allow a request body to be sent as part of a refund request. Kushki uses the DELETE http method for refunds and voids, which meant that they were truly reference transactions only without a request body. The ssl_invoke method needed to be updated to allow for this, and still be able to send the ticketNumber through in post to be captured and added to the url.

This also updates several remote tests that had `PEN` currency to use credentials that are tied to Peru so they will pass.

Unit Tests:
18 tests, 117 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests:
25 tests, 74 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96% passed
*There are 3 tests failing due to credentials issues. Kushki has credentials tied to certain countries and we are trying to get that fixed

Local Tests:
5651 tests, 78252 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/kushki.rb
+++ b/lib/active_merchant/billing/gateways/kushki.rb
@@ -48,8 +48,9 @@ module ActiveMerchant #:nodoc:
         post = {}
         post[:ticketNumber] = authorization
         add_full_response(post, options)
+        add_invoice(action, post, amount, options)
 
-        commit(action, post)
+        commit(action, post, options)
       end
 
       def void(authorization, options = {})
@@ -185,7 +186,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_full_response(post, options)
-        post[:fullResponse] = options[:full_response].to_s.casecmp('true').zero? if options[:full_response]
+        # this is the only currently accepted value for this field, previously it was 'true'
+        post[:fullResponse] = 'v2' unless options[:full_response] == 'false' || options[:full_response].blank?
       end
 
       def add_metadata(post, options)
@@ -245,10 +247,10 @@ module ActiveMerchant #:nodoc:
         'capture' => 'capture'
       }
 
-      def commit(action, params)
+      def commit(action, params, options = {})
         response =
           begin
-            parse(ssl_invoke(action, params))
+            parse(ssl_invoke(action, params, options))
           rescue ResponseError => e
             parse(e.response.body)
           end
@@ -265,9 +267,11 @@ module ActiveMerchant #:nodoc:
         )
       end
 
-      def ssl_invoke(action, params)
+      def ssl_invoke(action, params, options)
         if %w[void refund].include?(action)
-          ssl_request(:delete, url(action, params), nil, headers(action))
+          # removes ticketNumber from request for partial refunds because gateway will reject if included in request body
+          data = options[:partial_refund] == true ? post_data(params.except(:ticketNumber)) : nil
+          ssl_request(:delete, url(action, params), data, headers(action))
         else
           ssl_post(url(action, params), post_data(params), headers(action))
         end
--- a/test/unit/gateways/kushki_test.rb
+++ b/test/unit/gateways/kushki_test.rb
@@ -280,6 +280,49 @@ class KushkiTest < Test::Unit::TestCase
     assert_equal 'K010', refund.error_code
   end
 
+  def test_partial_refund
+    @gateway.expects(:ssl_post).returns(successful_charge_response)
+    @gateway.expects(:ssl_post).returns(successful_token_response)
+
+    options = { currency: 'PEN' }
+
+    purchase = @gateway.purchase(100, @credit_card, options)
+
+    refund = stub_comms(@gateway, :ssl_request) do
+      refund_options = {
+        currency: 'PEN',
+        partial_refund: true,
+        full_response: true
+      }
+      @gateway.refund(50, purchase.authorization, refund_options)
+    end.check_request do |_method, _endpoint, data, _headers|
+      request = JSON.parse(data)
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -245,10 +247,10 @@ def add_three_d_secure(post, payment_method, options)
         'capture' => 'capture'
       }
 
-      def commit(action, params)
+      def commit(action, params, options)
```
jcreiff, line 250: Since `refund` is the only method that needs to pass any options, what if you made this method definition have a default? 
```suggestion
      def commit(action, params, options = {})
```

jcreiff: (And then you could remove the additional `options` arguments from the other transaction methods)

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -185,7 +186,8 @@ def add_contact_details(post, contact_details_options)
       end
 
       def add_full_response(post, options)
-        post[:fullResponse] = options[:full_response].to_s.casecmp('true').zero? if options[:full_response]
+        # this is the only currently accepted value for this field, previously it was 'true'
+        post[:fullResponse] = 'v2' unless options[:full_response].nil? || !options[:full_response]
```
jcreiff, line 190: Just to double-check: the idea here is still to send this field only if the request explicitly includes it in options, right?

test/unit/gateways/kushki_test.rb
```diff
@@ -280,6 +280,26 @@ def test_failed_refund
     assert_equal 'K010', refund.error_code
   end
 
+  def test_partial_refund
```
jcreiff, line 283: Should we also add a test that shows that we are not sending any body for a full refund?

jcreiff (commented):
Couple of questions for consideration

lib/active_merchant/billing/gateways/kushki.rb
rachelkirk: Yes, I had it written a different way and rubocop was being weird. This kinda looks weird to me so if you have a better idea that I'm too tired to see staring me in the face, fire away!

test/unit/gateways/kushki_test.rb
rachelkirk: ok will do

lib/active_merchant/billing/gateways/kushki.rb
jcreiff: Maybe `.blank?`

https://blog.arkency.com/2017/07/nil-empty-blank-ruby-rails-difference/#_code_blank___code_

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -265,8 +267,12 @@ def commit(action, params)
...
 
-      def ssl_invoke(action, params)
-        if %w[void refund].include?(action)
+      def ssl_invoke(action, params, options)
+        if %w[refund].include?(action) && options[:partial_refund] == true
+          # removes ticketNumber from request because gateway will reject if included in request body
+          post_params = params.except(:ticketNumber)
+          ssl_request(:delete, url(action, params), post_data(post_params), headers(action))
+        elsif %w[void refund].include?(action)
           ssl_request(:delete, url(action, params), nil, headers(action))
```
almalee24, line 276: I would suggest moving the top part to here. It could look something like this

elsif %w[void refund].include?(action)
   data = options[:partial_refund] == true ? post_data(params.except(:ticketNumber)) : nil
   ssl_request(:delete, url(action, params), data, headers(action))

lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -265,8 +267,12 @@ def commit(action, params)
...
       end
 
-      def ssl_invoke(action, params)
-        if %w[void refund].include?(action)
+      def ssl_invoke(action, params, options)
+        if %w[refund].include?(action) && options[:partial_refund] == true
+          # removes ticketNumber from request because gateway will reject if included in request body
+          post_params = params.except(:ticketNumber)
+          ssl_request(:delete, url(action, params), post_data(post_params), headers(action))
+        elsif %w[void refund].include?(action)
```
almalee24, line 275: I would suggest moving the top part to here. It could look something like this

```
elsif %w[void refund].include?(action)
    data = options[:partial_refund] == true ? post_data(params.except(:ticketNumber)) : nil
     ssl_request(:delete, url(action, params), data, headers(action))
```

lib/active_merchant/billing/gateways/kushki.rb
rachelkirk: Used `.blank?` and also allowed for if someone passed it as false. But since we stringify everything and won't be explicitly changing it to a boolean like we do for others left 'false' as a string in the check.

lib/active_merchant/billing/gateways/kushki.rb
rachelkirk: Nice, and done!

yunnydang (approved):
Thanks for addressing Joes and Almas comments with their recommended changes! I also pulled this branch down to run some tests and saw that the one you written had passed. I had 9 other test failures so I'll assume I dont have the most up to date creds that kushki recently gave you (same ones fail on master as well). Only thing I would suggest is to change/update the PR body description to reflect that your remote test is indeed passing now (feel free to run it again before merging to be double sure). Fantastic work on this 🚀!!

PR #4937, opened 2023-10-30
Braintree: Add new stored credential method

aenand:
ECS-3194

Braintree has informed us we are not handling stored credentials appropriately in certain cases. This commit adds a new method to handle this new behavior in a controlled manner.

The changes are that if it's the initial_transaction & recurring it is to be marked as `recurring_first`.

If there are any reason_types unrecognized we will mark transaction_source as nil

Test Summary
Remote:
111 tests, 585 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -827,15 +827,39 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_credential_data(parameters, credit_card_or_vault_id, options)
+        # Braintree has informed us that the stored_credential mapping may be incorrect
+        # In order to prevent possible breaking changes we will only apply the new logic if
+        # specifically requested. This will be the default behavior in a future release.
         return unless (stored_credential = options[:stored_credential])
 
-        parameters[:external_vault] = {}
-        if stored_credential[:initial_transaction]
-          parameters[:external_vault][:status] = 'will_vault'
+        add_external_vault(parameters, stored_credential)
+
+        if options[:stored_credentials_v2]
+          stored_credentials_v2(parameters, stored_credential)
         else
-          parameters[:external_vault][:status] = 'vaulted'
-          parameters[:external_vault][:previous_network_transaction_id] = stored_credential[:network_transaction_id]
+          stored_credentials_v1(parameters, stored_credential)
         end
+      end
+
+      def stored_credentials_v2(parameters, stored_credential)
+        # Differences between v1 and v2 are
+        # initial_transaction + recurring/installment should be labeled {{reason_type}}_first
+        # unscheduled in AM should map to '' at BT because unscheduled here means not on a fixed timeline or fixed amount
+        case stored_credential[:reason_type]
+        when 'recurring', 'installment'
+          if stored_credential[:initial_transaction]
+            parameters[:transaction_source] = "#{stored_credential[:reason_type]}_first"
+          else
+            parameters[:transaction_source] = stored_credential[:reason_type]
+          end
+        when 'recurring_first', 'moto'
+          parameters[:transaction_source] = stored_credential[:reason_type]
+        else
+          parameters[:transaction_source] = ''
+        end
+      end
+
+      def stored_credentials_v1(parameters, stored_credential)
         if stored_credential[:initiator] == 'merchant'
           if stored_credential[:reason_type] == 'installment'
             parameters[:transaction_source] = 'recurring'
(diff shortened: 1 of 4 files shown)


bradbroge (approved):
nice!

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -827,15 +827,38 @@ def xid_or_ds_trans_id(three_d_secure_opts)
...
         else
-          parameters[:external_vault][:status] = 'vaulted'
-          parameters[:external_vault][:previous_network_transaction_id] = stored_credential[:network_transaction_id]
+          classic_stored_credentials(parameters, stored_credential)
         end
+      end
+
+      def updated_stored_credentials(parameters, stored_credential)
+        case stored_credential[:reason_type]
+        when 'moto'
```
almalee24, line 846: How can `reason_type` be moto?

aenand: Good catch, it can't! It was there before so I wonder if non-spreedly people pass it in? Do you know if we have a ticket to add `moto` to BT?

almalee24: Yes but it's set on `:sca_exemption`

almalee24: So `sca_exemption` is already part of Braintree. Do we also want to add if for `transaction_source` if customer sends `three_ds_exemption_type` as `moto`. Maybe that's for another ticket?

aenand: Where do you see `sca_exemption` for Braintree? I don't see we support it now in AM

almalee24: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/braintree_blue.rb#L690

aenand: 🤦🏽 it helps if i look at a somewhat new branch. I'm not sure if we should add it here for MOTO, I'll look through their docs more

almalee24: I do see their docs shows it but I think we can remove MOTO from now. I think I would make more sense to assign `transaction_source: 'moto'` if 3ds exemption type MOTO is passed. But we should probably confirm first and this could possible be a separate ticket.

aenand: Cool I created a ticket and will remove it from here!

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -827,15 +827,39 @@ def xid_or_ds_trans_id(three_d_secure_opts)
...
+          updated_stored_credentials(parameters, stored_credential)
         else
-          parameters[:external_vault][:status] = 'vaulted'
-          parameters[:external_vault][:previous_network_transaction_id] = stored_credential[:network_transaction_id]
+          classic_stored_credentials(parameters, stored_credential)
+        end
+      end
+
+      def updated_stored_credentials(parameters, stored_credential)
+        # Differences between classic and updated are
```
curiousepic, line 845: Even with the comments, I think I'd suggestion something like `stored_credentials_v1` and `_v2` rather than "updated" and "classic", which are somewhat ambiguous when just reading code.

curiousepic (approved):
Small naming suggestion but looks good!

lib/active_merchant/billing/gateways/braintree_blue.rb
curiousepic: And in any case, the `options` field name should match

aenand: First part is done! what did you mean by the `options` field name should match?

curiousepic: You got it already 👍

curiousepic (approved):
Thanks, looks good.

lib/active_merchant/billing/gateways/braintree_blue.rb
curiousepic: oops, missed the change for the comment line this message is on.

aenand: nice catch!

curiousepic (approved):
👍

PR #4938, opened 2023-10-31
Braintree: 3DS2

gasb150:
Fix token nonce to support nil billing address
[SER-954](https://spreedly.atlassian.net/browse/SER-954)


naashton (approved):
Nice and simple!

PR #4939, opened 2023-10-31
GlobalCollect: Fix bug in success_from logic

DustinHaefele:
Some apple_pay transactions that were through global collect direct (ogone) were being marked as failed despite success. The response from direct was slightly different than the standard global collect so the logic wasn't seeing them as successful.

Spreedly reference ticket: ECS-3197
remote tests:
54 tests, 141 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

unit tests:
45 tests, 230 assertions, 0 failures, 0 errors, 0 pending, 0 omissions, 0 notifications 100% passed
1497.80 tests/s, 7655.44 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/global_collect.rb
+++ b/lib/active_merchant/billing/gateways/global_collect.rb
@@ -522,8 +522,6 @@ module ActiveMerchant #:nodoc:
       def success_from(action, response)
         return false if response['errorId'] || response['error_message']
 
-        return %w(CAPTURED CAPTURE_REQUESTED).include?(response.dig('payment', 'status')) if response.dig('payment', 'paymentOutput', 'paymentMethod') == 'mobile'
-
         case action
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/global_collect.rb
```diff
@@ -522,8 +522,6 @@ def content_type
       def success_from(action, response)
         return false if response['errorId'] || response['error_message']
 
-        return %w(CAPTURED CAPTURE_REQUESTED).include?(response.dig('payment', 'status')) if response.dig('payment', 'paymentOutput', 'paymentMethod') == 'mobile'
```
almalee24, line 525: Great catch, I was so confused about this one.

almalee24 (approved):

PR #4948, opened 2023-11-07
SafeCharge: Support tokens

almalee24:
Remote:
34 tests, 96 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
29 tests, 160 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/safe_charge.rb
+++ b/lib/active_merchant/billing/gateways/safe_charge.rb
@@ -149,26 +149,46 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_payment(post, payment, options = {})
-        post[:sg_ExpMonth] = format(payment.month, :two_digits)
-        post[:sg_ExpYear] = format(payment.year, :two_digits)
-        post[:sg_CardNumber] = payment.number
+        case payment
+        when String
+          add_token(post, payment)
+        when CreditCard
+          post[:sg_ExpMonth] = format(payment.month, :two_digits)
+          post[:sg_ExpYear] = format(payment.year, :two_digits)
+          post[:sg_CardNumber] = payment.number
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/safe_charge.rb
```diff
@@ -149,26 +149,46 @@ def add_transaction_data(trans_type, post, money, options)
...
+
+      def add_credit_card(post, payment, options)
+        post[:sg_CVV2] = payment.verification_value
+        post[:sg_NameOnCard] = payment.name
+        post[:sg_StoredCredentialMode] = (options[:stored_credential_mode] == true ? 1 : 0)
+      end
+
+      def add_network_token(post, payment, options)
+        post[:sg_CAVV] = payment.payment_cryptogram
+        post[:sg_ECI] = options[:three_d_secure] && options[:three_d_secure][:eci] || '05'
```
aenand, line 183: Should this be the ECI for the network token or the 3DS ECI value?

aenand (changes requested):
Looks good! Can you update the commit message with a description of what you are adding and is this possible to test with some unit tests?

lib/active_merchant/billing/gateways/safe_charge.rb
almalee24: Oh that's a good question, probably the network token. Since this is not related to the ticket and it will need some additional testing I'll create a ticket for it.

almalee24:
> Looks good! Can you update the commit message with a description of what you are adding and is this possible to test with some unit tests?

Yes, I'll do that.

aenand (approved):
LGTM :shipit:

PR #4950, opened 2023-11-08
Worldpay: Update stored credentials logic

DustinHaefele:
Remote tests:
99 tests, 429 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit tests:
116 tests, 658 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Spreedly ticket: ECS-3238

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -684,30 +684,27 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_credential_using_normalized_fields(xml, options)
-        if options[:stored_credential][:initial_transaction]
-          xml.storedCredentials 'usage' => 'FIRST'
-        else
-          reason = case options[:stored_credential][:reason_type]
-                   when 'installment' then 'INSTALMENT'
-                   when 'recurring' then 'RECURRING'
-                   when 'unscheduled' then 'UNSCHEDULED'
-                   end
+        reason = case options[:stored_credential][:reason_type]
+                 when 'installment' then 'INSTALMENT'
+                 when 'recurring' then 'RECURRING'
+                 when 'unscheduled' then 'UNSCHEDULED'
+                 end
+        is_initial_transaction = options[:stored_credential][:initial_transaction]
+        stored_credential_params = generate_stored_credential_params(is_initial_transaction, reason)
 
-          xml.storedCredentials 'usage' => 'USED', 'merchantInitiatedReason' => reason do
-            xml.schemeTransactionIdentifier options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id]
-          end
+        xml.storedCredentials stored_credential_params do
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -684,30 +684,27 @@ def add_stored_credential_options(xml, options = {})
...
-          reason = case options[:stored_credential][:reason_type]
-                   when 'installment' then 'INSTALMENT'
-                   when 'recurring' then 'RECURRING'
-                   when 'unscheduled' then 'UNSCHEDULED'
-                   end
-
-          xml.storedCredentials 'usage' => 'USED', 'merchantInitiatedReason' => reason do
-            xml.schemeTransactionIdentifier options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id]
-          end
+        reason = case options[:stored_credential][:reason_type]
```
aenand, line 687: The previous code did not include this if it was the `initial_transaction`, did we want to preserve that logic?

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -1027,6 +1024,15 @@ def eligible_for_0_auth?(payment_method, options = {})
       def card_holder_name(payment_method, options)
         test? && options[:execute_threed] && !options[:three_ds_version]&.start_with?('2') ? '3D' : payment_method.name
       end
+
+      def generate_stored_credential_params(is_initial_transaction, reason = nil)
+        customer_or_merchant = reason == 'RECURRING' && is_initial_transaction ? 'customerInitiatedReason' : 'merchantInitiatedReason'
```
aenand, line 1029: Should we use `stored_credential[:initiator]` for customer_or_merchant?

test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -521,6 +521,40 @@ def test_successful_auth_and_capture_with_normalized_stored_credential
     assert_success capture
   end
 
+  def test_successful_auth_and_capture_with_normalized_recurring_stored_credential
+    stored_credential_params = {
```
aenand, line 525: Fun fact, there is a way to invoke standardized stored creds params in active_merchant by calling `stored_credential`. You can see the options for it in `test_helper.rb` and the `braintree_blue_test` file has some examples of how to invoke it.

lib/active_merchant/billing/gateways/worldpay.rb
DustinHaefele: No, Worldpay's docs are pretty specific about how to send the standing instruction reason on the first transaction, and this difference between their docs and our implementation is what spawned this ticket in the first place.

test/remote/gateways/remote_worldpay_test.rb
DustinHaefele: I updated all the stored credential params in this test file to use the helper 🎉 , thanks for pointing it out!

PR #4951, opened 2023-11-09
Redsys REST

aenand:
ECS-3219

This ticket adds a new gateway that serves as an update over the legacy
Redsys integration by moving to the Redsys Rest integration. This commit
adds support for basic functionality including authorize, purchase, scrubbing,
and standardizes responses.

This commit does not include support for:
* Stored Credentials (Ticket not yet created to add this support)
* External 3DS Support (Ticket not yet created to add this support)
* Redsys 3DS Support (Ticket created to add this support)

Co-authored-by: Amit Enand <aenand@spreedly.com>
Co-authored-by: Piers Chambers <piers@varyonic.com>

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys_rest.rb
+++ b/lib/active_merchant/billing/gateways/redsys_rest.rb
@@ -1,0 +1,433 @@
+# coding: utf-8
+
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    # = Redsys Merchant Gateway
+    #
+    # Gateway support for the Spanish "Redsys" payment gateway system. This is
+    # used by many banks in Spain and is particularly well supported by
+    # Catalunya Caixa's ecommerce department.
+    #
+    # Redsys requires an order_id be provided with each transaction and it must
+    # follow a specific format. The rules are as follows:
+    #
+    #  * First 4 digits must be numerical
+    #  * Remaining 8 digits may be alphanumeric
+    #  * Max length: 12
+    #
+    #  If an invalid order_id is provided, we do our best to clean it up.
+    #
+    # Written by Piers Chambers (Varyonic.com)
+    #
+    # *** SHA256 Authentication Update ***
+    #
+    # Redsys has dropped support for the SHA1 authentication method.
+    # Developer documentation: https://pagosonline.redsys.es/desarrolladores.html
+    class RedsysRestGateway < Gateway
+      self.test_url = 'https://sis-t.redsys.es:25443/sis/rest/'
+      self.live_url = 'https://sis.redsys.es/sis/rest/'
+
+      self.supported_countries = ['ES']
+      self.default_currency    = 'EUR'
+      self.money_format        = :cents
+      # Not all card types may be activated by the bank!
+      self.supported_cardtypes = %i[visa master american_express jcb diners_club unionpay]
+      self.homepage_url        = 'http://www.redsys.es/'
+      self.display_name        = 'Redsys (REST)'
+
+      CURRENCY_CODES = {
+        'AED' => '784',
+        'ARS' => '32',
+        'AUD' => '36',
+        'BRL' => '986',
+        'BOB' => '68',
+        'CAD' => '124',
+        'CHF' => '756',
+        'CLP' => '152',
+        'CNY' => '156',
+        'COP' => '170',
+        'CRC' => '188',
+        'CZK' => '203',
+        'DKK' => '208',
+        'DOP' => '214',
+        'EUR' => '978',
+        'GBP' => '826',
+        'GTQ' => '320',
+        'HUF' => '348',
+        'IDR' => '360',
+        'INR' => '356',
+        'JPY' => '392',
+        'KRW' => '410',
+        'MYR' => '458',
+        'MXN' => '484',
+        'NOK' => '578',
+        'NZD' => '554',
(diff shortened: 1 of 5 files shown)


test/remote/gateways/remote_redsys_rest_test.rb
```diff
@@ -0,0 +1,206 @@
...
+  end
+
+  def test_successful_purchase
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal 'Transaction Approved', response.message
+  end
+
+  def test_purchase_with_invalid_order_id
+    response = @gateway.purchase(@amount, @credit_card, order_id: "a%4#{generate_order_id}")
```
DustinHaefele, line 25: Is this invalid order Id test testing that we can clean up the order id as noted in the gateway, or why do we expect success here?

test/remote/gateways/remote_redsys_rest_test.rb
```diff
@@ -0,0 +1,206 @@
...
+
+  def test_purchase_and_failed_refund
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success purchase
+    refund = @gateway.refund(@amount + 100, purchase.authorization, @options)
+    assert_failure refund
+    assert_equal 'SIS0057 ERROR', refund.message
+  end
+
+  def test_purchase_and_refund_with_currency
```
DustinHaefele, line 51: Nitpick: but maybe this test name should be something like `failed_purchase_with_unsupported_currency` since we are testing failure and not doing a refund.

test/remote/gateways/remote_redsys_rest_test.rb
```diff
@@ -0,0 +1,206 @@
...
+    clean_transcript = @gateway.scrub(transcript)
+
+    assert_scrubbed(@gateway.options[:secret_key], clean_transcript)
+    assert_scrubbed(@credit_card.number, clean_transcript)
+    assert_scrubbed(@credit_card.verification_value.to_s, clean_transcript)
+  end
+
+  def test_encrypt_handles_url_safe_character_in_secret_key_without_error
+    gateway = RedsysRestGateway.new({
+      login: '091952713',
```
DustinHaefele, line 144: Is there a reason we are using different hard coded creds here instead of the creds from the fixtures?

test/unit/gateways/redsys_rest_test.rb
```diff
@@ -0,0 +1,269 @@
+require 'test_helper'
+
+class RedsysRestTest < Test::Unit::TestCase
+  include CommStub
+
+  def setup
+    @credentials = {
+      login: '091952713',
```
DustinHaefele, line 8: Could we use fixtures for these creds too?

lib/active_merchant/billing/gateways/redsys_rest.rb
```diff
@@ -0,0 +1,433 @@
...
+      def add_direct_payment(post, options)
+        # Direct payment skips 3DS authentication. We should only apply this if execute_threed is false
+        # or authentication data is not present. Authentication data support to be added in the future.
+        return if options[:execute_threed] || options[:authentication_data]
+
+        post[:DS_MERCHANT_DIRECTPAYMENT] = true
+      end
+
+      def add_threeds(post, options)
+        post[:DS_MERCHANT_EMV3DS] = { threeDSInfo: 'CardData' } if options[:execute_threed]
```
DustinHaefele, line 279: Since the remote tests say that 3DS isn't supported yet, what is this `add_threeds` function doing here?

DustinHaefele (commented):
Had some questions, but overall this looks like it will work.

test/remote/gateways/remote_redsys_rest_test.rb
aenand: Yep! The external contributor added this in their original PR to ensure we clean the order_id

test/remote/gateways/remote_redsys_rest_test.rb
aenand: good idea!

test/remote/gateways/remote_redsys_rest_test.rb
aenand: Yep! This test is making sure url unsafe characters are fine to be used in the `login`/`authentication`

test/unit/gateways/redsys_rest_test.rb
aenand: I'll move these to the fixture creds!

lib/active_merchant/billing/gateways/redsys_rest.rb
aenand: Great question! This provides functionality for the initial setup to go to the 3DS server or the payment server. The logic to have this was already in place so rather than scrap it, I kept it and will make a note of it when we get to adding 3DS support here (the following ticket).

test/remote/gateways/remote_redsys_rest_test.rb
aenand: Done!

test/unit/gateways/redsys_rest_test.rb
aenand: Done!

DustinHaefele (approved):
:shipit:

PR #4956, opened 2023-11-13
Vantiv Express: New Xml gateway

DustinHaefele:
Unit test:
28 tests, 142 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote tests:
31 tests, 91 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/vantiv_express.rb
+++ b/lib/active_merchant/billing/gateways/vantiv_express.rb
@@ -64,6 +64,92 @@ module ActiveMerchant #:nodoc:
         'NoShow' => 6
       }
 
+      LODGING_PPC = {
+        'NonParticipant' => 0,
+        'DollarLimit500' => 1,
+        'DollarLimit1000' => 2,
+        'DollarLimit1500' => 3
+      }
+
+      LODGING_SPC = {
+        'Default' => 0,
+        'Sale' => 1,
+        'NoShow' => 2,
+        'AdvancedDeposit' => 3
+      }
+
+      LODGING_CHARGE_TYPE = {
+        'Default' => 0,
+        'Restaurant' => 1,
+        'GiftShop' => 2
+      }
+
+      TERMINAL_TYPE = {
+        'Unknown' => 0,
+        'PointOfSale' => 1,
+        'Ecommerce' => 2,
+        'MOTO' => 3,
+        'FuelPump' => 4,
+        'ATM' => 5,
+        'Voice' => 6,
+        'Mobile' => 7,
+        'WebSiteGiftCard' => 8
+      }
+
+      CARD_HOLDER_PRESENT_CODE = {
+        'Default' => 0,
+        'Unknown' => 1,
+        'Present' => 2,
+        'NotPresent' => 3,
+        'MailOrder' => 4,
+        'PhoneOrder' => 5,
+        'StandingAuth' => 6,
+        'ECommerce' => 7
+      }
+
+      CARD_INPUT_CODE = {
+        'Default' => 0,
+        'Unknown' => 1,
+        'MagstripeRead' => 2,
+        'ContactlessMagstripeRead' => 3,
+        'ManualKeyed' => 4,
+        'ManualKeyedMagstripeFailure' => 5,
+        'ChipRead' => 6,
+        'ContactlessChipRead' => 7,
+        'ManualKeyedChipReadFailure' => 8,
+        'MagstripeReadChipReadFailure' => 9,
+        'MagstripeReadNonTechnicalFallback' => 10
+      }
+
+      CVV_PRESENCE_CODE = {
+        'UseDefault' => 0,
+        'NotProvided' => 1,
+        'Provided' => 2,
+        'Illegible' => 3,
+        'CustomerIllegible' => 4
+      }
+
+      TERMINAL_CAPABILITY_CODE = {
+        'Default' => 0,
+        'Unknown' => 1,
+        'NoTerminal' => 2,
+        'MagstripeReader' => 3,
+        'ContactlessMagstripeReader' => 4,
+        'KeyEntered' => 5,
+        'ChipReader' => 6,
+        'ContactlessChipReader' => 7
+      }
+
+      TERMINAL_ENVIRONMENT_CODE = {
+        'Default' => 0,
+        'NoTerminal' => 1,
+        'LocalAttended' => 2,
+        'LocalUnattended' => 3,
+        'RemoteAttended' => 4,
+        'RemoteUnattended' => 5,
+        'ECommerce' => 6
+      }
+
       def initialize(options = {})
         requires!(options, :account_id, :account_token, :application_id, :acceptor_id, :application_name, :application_version)
         super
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/vantiv_express.rb
```diff
@@ -0,0 +1,475 @@
...
+        'NoShow' => 6
+      }
+
+      def initialize(options = {})
+        requires!(options, :account_id, :account_token, :application_id, :acceptor_id, :application_name, :application_version)
+        super
+      end
+
+      def purchase(money, payment, options = {})
+        action = payment.is_a?(Check) ? 'CheckSale' : 'CreditCardSale'
```
aenand, line 73: Can you not do a `Check` transaction against `authorize`?

lib/active_merchant/billing/gateways/vantiv_express.rb
```diff
@@ -0,0 +1,475 @@
...
+        eci = payment.is_a?(NetworkTokenizationCreditCard) ? parse_eci(payment) : nil
+
+        request = build_xml_request do |xml|
+          xml.send(action, xmlns: live_url) do
```
aenand, line 77: Is the `xmlns` always the `live_url` even in sandbox tests?

lib/active_merchant/billing/gateways/vantiv_express.rb
```diff
@@ -0,0 +1,475 @@
...
+            add_credentials(xml)
```
aenand, line 78: Should we move `add_credentials` to `def commit` since it's used on all requests?

lib/active_merchant/billing/gateways/vantiv_express.rb
```diff
@@ -0,0 +1,475 @@
...
+            add_terminal(xml, options, eci)
+          end
+        end
+
+        commit(request, trans_amount)
+      end
+
+      def store(payment, options = {})
+        request = build_xml_request do |xml|
+          xml.PaymentAccountCreate(xmlns: SERVICE_LIVE_URL) do
```
aenand, line 169: How is the `service_live_url` different than `live_url` and is there a reason for `store` to be the only one referencing `service_live_url`?

lib/active_merchant/billing/gateways/vantiv_express.rb
```diff
@@ -0,0 +1,475 @@
...
+        xml.Terminal do
+          xml.TerminalID options[:terminal_id] || '01'
+          xml.CardPresentCode network_token_eci ? 3 : card_present
+          xml.CardholderPresentCode network_token_eci ? 7 : 0
+          xml.CardInputCode network_token_eci ? 4 : 0
+          xml.CVVPresenceCode 0
+          xml.TerminalCapabilityCode network_token_eci ? 5 : 0
+          xml.TerminalEnvironmentCode network_token_eci ? 6 : 0
+          xml.MotoECICode network_token_eci || 7
+          xml.TerminalType 1 if network_token_eci
```
aenand, line 310: Many of these seem to be based around `network_token_eci`, are all of these meant to be set differently only for network token? The name of some suggest they would be set by other values (ex: `CardholderPresentCode` I would expect to match a stored creds or cvv presence check and `MotoECICode` would match a `sca_exemption` check)

lib/active_merchant/billing/gateways/vantiv_express.rb
```diff
@@ -0,0 +1,475 @@
...
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover diners_club jcb]
+
+      self.homepage_url = 'http://www.elementps.com'
+      self.display_name = 'Element'
+
+      SERVICE_TEST_URL = 'https://certservices.elementexpress.com'
+      SERVICE_LIVE_URL = 'https://services.elementexpress.com'
+
+      NETWORK_TOKEN_TYPE = {
```
aenand, line 20: Does this gateway support third party Network Tokens?

test/unit/gateways/vantiv_express_test.rb
```diff
@@ -0,0 +1,490 @@
+require 'test_helper'
+
+class VantivExpressTest < Test::Unit::TestCase
+  include CommStub
+
+  def setup
+    @gateway = VantivExpressGateway.new(account_id: '', account_token: '', application_id: '', acceptor_id: '', application_name: '', application_version: '')
```
aenand, line 7: Should we be using the values in the `test/fixtures.yml` file instead of blank values?

aenand (changes requested):
Excellent test coverage! I also really appreciate how uniform this is with other AM gateways.

DustinHaefele:
Good questions @aenand, a lot of this work was complete before I grabbed the ticket so it may take me a second to get answers but I'm looking into it now.

lib/active_merchant/billing/gateways/vantiv_express.rb
DustinHaefele: the service_live_url is `services.elementexpress.com` and the live_url is `transaction.elementexpress.com` They have a bunch of service methods and transaction methods for their api, and store is the only method that we have that hits their services api.

DustinHaefele: This is also how the element (non-xml) gateway works.

lib/active_merchant/billing/gateways/vantiv_express.rb
DustinHaefele: According to the `wallet_type` enumeration that this is being used to fill they only support apple_pay, google_pay, and samsung_pay.

lib/active_merchant/billing/gateways/vantiv_express.rb
DustinHaefele: Yes, I didn't notice this when I picked up the ticket. I'm updating this now based on their docs and the GSFs that we allow on this gateway.

lib/active_merchant/billing/gateways/vantiv_express.rb
aenand: Makes sense!

lib/active_merchant/billing/gateways/vantiv_express.rb
DustinHaefele: That does make sense, but I decided to just pull them from the GSFs to keep it in line with the current element integration. That deduction could be added in the future though, if desired.

aenand: Alma mentioned that one big change from the existing integration to the new one is that the old ones used `String` enums but the new integration uses codes. Since you are pulling these values from the GSFs, does this mean the merchant needs to change logic to pass in `codes` instead of the `Strings`?

aenand: amazing thank you 🙇🏽 this will be massive for moving people over seamlessly and hopefully deleting the old version

DustinHaefele: Okie doke. This should be ready for another review now!

aenand (approved):
LGTM :shipit: Amazing work 👏🏽

PR #4959, opened 2023-11-15
Orbital:  Remove needless GSF for TPV

javierpedrozaing:
Description
-------------------------
[SER-969](https://spreedly.atlassian.net/browse/SER-969) 
[SER-970](https://spreedly.atlassian.net/browse/SER-970) 
[SER-971](https://spreedly.atlassian.net/browse/SER-971)

This commit remove a couple of needless GSF during a TPV flow we hard-coded the values required for token_tx_type and card_brand instead of sending these as GSF

Unit test
-------------------------
Finished in 1.903239 seconds.

148 tests, 839 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

77.76 tests/s, 440.83 assertions/s

Remote test
-------------------------
Finished in 118.453815 seconds.

133 tests, 526 assertions, 13 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 90.2256% passed

1.12 tests/s, 4.44 assertions/s

Rubocop
-------------------------
781 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -565,15 +565,16 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_safetech_token_data(xml, payment_source, options)
+        payment_source_token = split_authorization(payment_source).values_at(2).first
         xml.tag! :CardBrand, options[:card_brand]
-        xml.tag! :AccountNum, payment_source
+        xml.tag! :AccountNum, payment_source_token
       end
 
       #=====PAYMENT SOURCE FIELDS=====
 
       # Payment can be done through either Credit Card or Electronic Check
       def add_payment_source(xml, payment_source, options = {})
-        add_safetech_token_data(xml, payment_source, options) if options[:token_txn_type] == USE_TOKEN
+        add_safetech_token_data(xml, payment_source, options) if payment_source.is_a?(String)
         payment_source.is_a?(Check) ? add_echeck(xml, payment_source, options) : add_credit_card(xml, payment_source, options)
       end
 
@@ -901,7 +902,7 @@ module ActiveMerchant #:nodoc:
             request.call(remote_url(:secondary))
           end
 
-        authorization = order.include?('<TokenTxnType>GT</TokenTxnType>') ? response[:safetech_token] : authorization_string(response[:tx_ref_num], response[:order_id])
+        authorization = authorization_string(response[:tx_ref_num], response[:order_id], response[:safetech_token], response[:card_brand])
 
         Response.new(
           success?(response, message_type),
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -74,6 +74,7 @@
 * RedsysRest: Update to $0 verify [almalee24] #4973
 * CommerceHub: Add credit transaction [sinourain] #4965
 * PayTrace: Send CSC value on gateway request. [DustinHaefele] #4974
+* Orbital: Remove needless GSF for TPV [javierpedrozaing] #4959
 
 == Version 1.135.0 (August 24, 2023)
 * PaymentExpress: Correct endpoints [steveh] #4827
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -565,15 +565,15 @@ def get_address(options)
       end
 
       def add_safetech_token_data(xml, payment_source, options)
-        xml.tag! :CardBrand, options[:card_brand]
-        xml.tag! :AccountNum, payment_source
+        xml.tag! :CardBrand, payment_source.split(';')[1]
```
Heavyblade, line 568: 💬 Style:

maybe is just matter of taste on semantics => `payment_source.split(';').first`

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -1031,7 +1031,7 @@ def build_new_order_xml(action, money, payment_source, parameters = {})
             add_stored_credentials(xml, parameters)
             add_pymt_brand_program_code(xml, payment_source, three_d_secure)
             add_mastercard_fields(xml, payment_source, parameters, three_d_secure) if mastercard?(payment_source)
-            xml.tag! :TokenTxnType, parameters[:token_txn_type] if parameters[:token_txn_type]
+            xml.tag! :TokenTxnType, parameters[:token_txn_type]
```
Heavyblade, line 1034: ❓ Question:

Given how strict is Orbital XSD definition and that this line will always include `TokenTxnType`, is this working properly with `refund` and `credit` transactions, because those methods are currently using the `build_new_order_xml`.

Heavyblade (commented):
Looking pretty much 👍  @javierpedrozaing , left a couple of comments and a question.

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: you 're right updated

lib/active_merchant/billing/gateways/orbital.rb
javierpedrozaing: you 're right updated, adding a validation

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -901,7 +901,7 @@ def commit(order, message_type, retry_logic = nil, trace_number = nil)
             request.call(remote_url(:secondary))
           end
 
-        authorization = order.include?('<TokenTxnType>GT</TokenTxnType>') ? response[:safetech_token] : authorization_string(response[:tx_ref_num], response[:order_id])
+        authorization = order.include?('<TokenTxnType>GT</TokenTxnType>') ? authorization_string(response[:safetech_token], response[:card_brand]) : authorization_string(response[:tx_ref_num], response[:order_id])
```
naashton, line 904: I think we need to rethink this `authorization` string we are building. There is a loose pattern for these things that we follow, which is not well documented, but it helps standardize these strings at the individual gateway level.

TLDR; We don't want to conditionally build the string to be different, depending on the transaction. Regardless of the values returned in the response from the gateway, we want to maintain a consistent structure.

Example:

```
authorization = authorization_string(r[field1], r[field2], r[field3])

# authorization = 'val1;val2;val3'
# authorization = 'val1;;val3'
# authorization = 'val1;;'
```

This way, wherever we split this string, we can reliably retrieve the index of the authorization and _know_ that it should be value we expect.

For this, we should take the existing `authorization_string` call and append the values we would like to include in this string, so that the length of this string (after it is split into the individual values) is always the same.

naashton (changes requested):

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -901,7 +902,7 @@ def commit(order, message_type, retry_logic = nil, trace_number = nil)
             request.call(remote_url(:secondary))
           end
 
-        authorization = order.include?('<TokenTxnType>GT</TokenTxnType>') ? response[:safetech_token] : authorization_string(response[:tx_ref_num], response[:order_id])
+        authorization = authorization_string(response[:tx_ref_num], response[:order_id], response[:safetech_token], response[:card_brand])
```
naashton, line 905: One final question regarding this line. Should we more safely fetch the values from the hash? My concern is if the gateway does not return one of these keys in the response that we'll fail on an index error.

naashton (approved):

PR #4961, opened 2023-11-16
VisaNet Peru: Pass the purchaseNumber in response

yunnydang:
This passes along the generated purchaseNumber in the response as Visanet peru does not echo this back to us.

Note: remote test for this gateway are un-available due to outdated credentials and issues with receiving new ones.

Local:
5732 tests, 78643 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6685% passed

Unit:
16 tests, 101 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


jcreiff (approved):
I think this looks good and after squinting through the various transaction paths it looks like there isn't any functional logic that will be affected by stashing this field in the `response` object prior to returning the `Response`

PR #4963, opened 2023-11-20
Ogone: Fix signature calulcation for blank fields

Heavyblade:
### Summary:
Add changes to deal with ogone signature when there are empty fields on the passed params

[SER-965](https://spreedly.atlassian.net/browse/SER-965)

### Remote Test:
Finished in 135.715931 seconds.
33 tests, 138 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.9697% passed

### Unit Tests:
Finished in 38.938916 seconds.
5733 tests, 78693 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/ogone.rb
+++ b/lib/active_merchant/billing/gateways/ogone.rb
@@ -460,7 +460,7 @@ module ActiveMerchant #:nodoc:
             raise "Unknown signature algorithm #{algorithm}"
           end
 
-        filtered_params = signed_parameters.compact
+        filtered_params = signed_parameters.reject { |_k, v| v.nil? || v == '' }
         sha_encryptor.hexdigest(
           filtered_params.sort_by { |k, _v| k.upcase }.map { |k, v| "#{k.upcase}=#{v}#{secret}" }.join('')
         ).upcase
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/ogone.rb
```diff
@@ -460,7 +460,7 @@ def calculate_signature(signed_parameters, algorithm, secret)
             raise "Unknown signature algorithm #{algorithm}"
           end
 
-        filtered_params = signed_parameters.compact
+        filtered_params = signed_parameters.reject { |_k, v| v.nil? || v == '' }
```
sinourain, line 463: I recommend `v.blank?` because it is true when the value is `nil?` or it is an empty string

sinourain (commented):
LGTM just leave a comment

lib/active_merchant/billing/gateways/ogone.rb
Heavyblade: I thought the same but it doesn't work with booleans:

Screenshot 2023-11-20 at 8 43 32 PM

naashton (approved):

lib/active_merchant/billing/gateways/ogone.rb
naashton, line 463: Nice! This `reject` method is new to me, but this is pretty clever.

naashton (approved):

PR #4964, opened 2023-11-21
Stripe PI: add the card brand field

yunnydang:
This will add the card brand field for the payment and setup intents

Local:
5733 tests, 78645 assertions, 0 failures, 19 errors, 0 pendings, 0 omissions, 0 notifications
99.6686% passed

Unit
58 tests, 302 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
95 tests, 448 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -44,6 +44,7 @@ module ActiveMerchant #:nodoc:
             add_fulfillment_date(post, options)
             request_three_d_secure(post, options)
             add_level_three(post, options)
+            add_card_brand(post, options)
             post[:expand] = ['charges.data.balance_transaction']
 
             CREATE_INTENT_ATTRIBUTES.each do |attribute|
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -323,6 +325,17 @@ def add_metadata(post, options = {})
...
 
+      def add_card_brand(post, options)
+        return unless options[:card_brand]
+
+        card_network = {}
+        card_network[:network] = options[:card_brand]
+
+        post[:payment_method_options] ||= {}
+        post[:payment_method_options][:card] ||= {}
+        post[:payment_method_options][:card] = card_network
```
jcreiff, line 336: 🔍 Shouldn't this be `post[:payment_method_options][:card][:network] = options[:card_brand]` ?

I think the idea here is correct, and it mirrors the implementation elsewhere in this gateway model which tries not to clear out any existing fields as we build out the hash for various scenarios. But as this is currently written, the memoization you're doing in line 335 becomes unnecessary, because whatever is currently nested at `post[:payment_method_options][:card]` (assuming there is something there other than a newly instantiated `{}`) will be overwritten by `card_network`

jcreiff (changes requested):

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
yunnydang: heya @jcreiff thanks for looking at this! As for your question if this should be "post[:payment_method_options][:card][:network] = options[:card_brand]" I originally tried to write it that way but ran into issues with create_setup_intent because that path does not currently have any ":payment_method_options" by default in the post. So it throws a NoMethodError. 
Looking back I do agree that line 335 is not needed as your explanation is correct. Thanks for catching that!

PR #4966, opened 2023-11-22
IPG: remove scrubbing of store id

aenand:
Store ID and User ID are not sensitive values so they should not be scrubbed.


bradbroge (approved):
LGTM

PR #4969, opened 2023-11-28
NexiXpay:  Add basic operation through 3ds

jherreraa:
Description
-------------------------
[SER-703](https://spreedly.atlassian.net/browse/SER-703)

This commit add NexiXpay gateway with its basic operations

### DIFF
--- a/lib/active_merchant/billing/gateways/xpay.rb
+++ b/lib/active_merchant/billing/gateways/xpay.rb
@@ -15,7 +15,7 @@ module ActiveMerchant #:nodoc:
 
       ENDPOINTS_MAPPING = {
         purchase: 'orders/2steps/payment',
-        authorize: 'orders/2steps/init',
+        preauth: 'orders/2steps/init',
         capture: 'operations/{%s}/captures',
         verify: 'orders/card_verification',
         void: 'operations/{%s}/cancels',
@@ -30,79 +30,161 @@ module ActiveMerchant #:nodoc:
 
       def purchase(amount, payment_method, options = {})
         post = {}
-        commit('purchase', post, options)
+        add_auth_purchase_params(post, amount, payment_method, options)
+        action = options[:operation_id] ? :purchase : :preauth
+        commit(action, post, options)
       end
 
       def authorize(amount, payment_method, options = {})
         post = {}
-        add_auth_purchase(post, amount, payment_method, options)
-        commit('authorize', post, options)
+        add_auth_purchase_params(post, amount, payment_method, options)
+        commit(:preauth, post, options)
       end
 
       def capture(amount, authorization, options = {})
         post = {}
-        commit('capture', post, options)
+        commit(:capture, post, options)
       end
 
       def void(authorization, options = {})
         post = {}
-        commit('void', post, options)
+        commit(:void, post, options)
       end
 
       def refund(amount, authorization, options = {})
         post = {}
-        commit('refund', post, options)
+        commit(:refund, post, options)
       end
 
       def verify(credit_card, options = {})
         post = {}
-        commit('verify', post, options)
+        add_invoice(post, 0, options)
+        add_credit_card(post, credit_card)
+        commit(:verify, post, options)
       end
 
       def supports_scrubbing?
         true
       end
 
-      def scrub(transcript) end
+      def scrub(transcript)
+        transcript.
+          gsub(%r((X-Api-Key: )(\w|-)+), '\1[FILTERED]').
+          gsub(%r(("pan\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("cvv\\?":\\?")\d+), '\1[FILTERED]')
+      end
 
       private
 
-      def add_invoice(post, money, options) end
+      def add_invoice(post, amount, options)
+        currency = options[:currency] || currency(amount)
+        post[:order] = {
+          orderId: options[:order_id],
+          amount: localized_amount(amount, currency),
+          currency: currency
+        }.compact
+      end
 
-      def add_payment_method(post, payment_method) end
+      def add_credit_card(post, payment_method)
+        post[:card] = {
+          pan: payment_method.number,
+          expiryDate: expdate(payment_method),
+          cvv: payment_method.verification_value
+        }
+      end
+
+      def add_payment_method(post, payment_method)
+        add_credit_card(post, payment_method) if payment_method.is_a?(CreditCard)
+      end
+
+      def add_customer_data(post, payment_method, options)
+        post[:order][:customerInfo] = {
+          cardHolderName: payment_method.name,
+          cardHolderEmail: options[:email]
+        }.compact
+      end
+
+      def add_address(post, options)
+        if address = options[:billing_address] || options[:address]
+          post[:order][:customerInfo][:billingAddress] = {
+            name: address[:name],
+            street: address[:address1],
+            additionalInfo: address[:address2],
+            city: address[:city],
+            postCode: address[:zip],
+            country: address[:country]
+          }.compact
+        end
+        if address = options[:shipping_address]
+          post[:order][:customerInfo][:shippingAddress] = {
+            name: address[:name],
+            street: address[:address1],
+            additionalInfo: address[:address2],
+            city: address[:city],
+            postCode: address[:zip],
+            country: address[:country]
+          }.compact
+        end
+      end
+
+      def add_recurrence(post, options)
+        post[:recurrence] = { action: options[:recurrence] || 'NO_RECURRING' }
+      end
+
+      def add_exemptions(post, options)
+        post[:exemptions] = options[:exemptions] || 'NO_PREFERENCE'
+      end
+
+      def add_3ds_params(post, options)
+        post[:threeDSAuthData] = { threeDSAuthResponse: options[:three_ds_auth_response] }.compact
+        post[:operationId] = options[:operation_id] if options[:operation_id]
+      end
+
+      def add_auth_purchase_params(post, amount, payment_method, options)
+        add_invoice(post, amount, options)
+        add_payment_method(post, payment_method)
+        add_customer_data(post, payment_method, options)
+        add_address(post, options)
+        add_recurrence(post, options) unless options[:operation_id]
+        add_exemptions(post, options)
+        add_3ds_params(post, options)
+      end
 
       def add_reference(post, authorization) end
 
       def add_auth_purchase(post, money, payment, options) end
 
-      def commit(action, params, options)
-        url = build_request_url(action)
-        response = ssl_post(url, params.to_json, request_headers(options))
+      def parse(body = {})
+        JSON.parse(body)
+      end
 
-        unless response
-          Response.new(
-            success_from(response),
-            message_from(success_from(response), response),
-            response,
-            authorization: authorization_from(response),
-            avs_result: AVSResult.new(code: response['some_avs_result_key']),
-            cvv_result: CVVResult.new(response['some_cvv_result_key']),
-            test: test?,
-            error_code: error_code_from(response)
-          )
+      def commit(action, params, options)
+        transaction_id = params.dig(:operation_id) unless action != 'capture'
+        begin
+          url = build_request_url(action, transaction_id)
+          raw_response = ssl_post(url, params.to_json, request_headers(options))
+          response = parse(raw_response)
+        rescue ResponseError => e
+          response = e.response.body
+          response = parse(response)
         end
-      rescue ResponseError => e
-        response = e.response.body
-        JSON.parse(response)
+
+        Response.new(
+          success_from(response),
+          message_from(response),
+          response,
+          authorization: authorization_from(response),
+          test: test?,
+          error_code: error_code_from(response)
+        )
       end
 
       def request_headers(options)
-        headers = {
+        {
           'Content-Type' => 'application/json',
           'X-Api-Key' => @api_key,
-          'Correlation-Id' => options.dig(:order, :order_id) || SecureRandom.uuid
+          'Correlation-Id' => options[:order_id] || SecureRandom.uuid
         }
-        headers
       end
 
       def build_request_url(action, id = nil)
@@ -112,23 +194,19 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(response)
-        response == 'SUCCESS'
+        response.dig('operation', 'operationResult') == 'PENDING' || response.dig('operation', 'operationResult') == 'FAILED' || response.dig('operation', 'operationResult') == 'AUTHORIZED'
       end
 
-      def message_from(succeeded, response)
-        if succeeded
-          'Succeeded'
-        else
-          response.dig('errors') unless response
-        end
+      def message_from(response)
+        response['errors'] || response.dig('operation', 'operationResult')
       end
 
       def authorization_from(response)
-        response.dig('latest_payment_attempt', 'payment_intent_id') unless response
+        response.dig('operation', 'operationId') unless response
       end
 
       def error_code_from(response)
-        response['provider_original_response_code'] || response['code'] unless success_from(response)
+        response.dig('errors', 0, 'code')
       end
     end
   end
--- a/test/remote/gateways/remote_xpay_test.rb
+++ b/test/remote/gateways/remote_xpay_test.rb
@@ -2,14 +2,64 @@ require 'test_helper'
 
 class RemoteRapydTest < Test::Unit::TestCase
   def setup
-    @gateway = XpayGateway.new(fixtures(:x_pay))
-    @amount = 200
-    @credit_card = credit_card('4111111111111111')
-    @options = {}
+    @gateway = XpayGateway.new(fixtures(:xpay))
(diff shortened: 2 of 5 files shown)


lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,13 +30,15 @@ def initialize(options = {})
 
       def purchase(amount, payment_method, options = {})
         post = {}
-        commit('purchase', post, options)
+        add_auth_purchase_params(post, amount, payment_method, options)
+        action = options[:operation_id] ? 'purchase' : 'preauth'
```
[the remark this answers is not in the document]
jherreraa: Thanks for the suggestion. Approach corrected.

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -56,51 +58,133 @@ def refund(amount, authorization, options = {})
...
+        post[:order][:amount] = localized_amount(amount, currency)
+        post[:order][:currency] = currency
+      end
+
+      def add_credit_card(post, payment_method)
+        post[:card] = {}
+        post[:card][:pan] = payment_method.number
+        post[:card][:expiryDate] = "#{format(payment_method.month, :two_digits)}#{format(payment_method.year, :two_digits)}"
+        post[:card][:cvv] = payment_method.verification_value
+      end
```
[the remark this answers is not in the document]
jherreraa: Suggestion noted! Thanks!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -56,51 +58,133 @@ def refund(amount, authorization, options = {})
...
+      def add_payment_method(post, payment_method)
+        add_credit_card(post, payment_method) if payment_method.is_a?(CreditCard)
+      end
+
+      def add_customer_data(post, payment_method, options)
+        post[:order][:customerInfo] = {}
+        card_holder_name = "#{payment_method.try(:first_name)} #{payment_method.try(:last_name)}"
+        post[:order][:customerInfo][:cardHolderName] = card_holder_name
+        post[:order][:customerInfo][:cardHolderEmail] = options[:email] if options[:email]
+      end
```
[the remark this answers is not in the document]
jherreraa: Thanks a lot for the corrections!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -56,51 +58,133 @@ def refund(amount, authorization, options = {})
...
+          post[:order][:customerInfo][:shippingAddress] = {}
+          post[:order][:customerInfo][:shippingAddress][:name] = address[:name] if address[:name]
+          post[:order][:customerInfo][:shippingAddress][:street] = address[:address1] if address[:address1]
+          post[:order][:customerInfo][:shippingAddress][:additionalInfo] = address[:address2] if address[:address2]
+          post[:order][:customerInfo][:shippingAddress][:city] = address[:city] if address[:city]
+          post[:order][:customerInfo][:shippingAddress][:postCode] = address[:zip] if address[:zip]
+          post[:order][:customerInfo][:shippingAddress][:country] = address[:country] if address[:country]
+        end
+        post
+      end
```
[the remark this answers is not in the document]
jherreraa: Thanks for the suggestions! Changes added!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -56,51 +58,133 @@ def refund(amount, authorization, options = {})
...
+
+      def add_exemptions(post, options)
+        post[:exemptions] = options[:exemptions] || 'NO_PREFERENCE'
+      end
+
+      def add_3ds_params(post, options)
+        post[:threeDSAuthData] = {}
+        post[:threeDSAuthData][:threeDSAuthResponse] = options[:three_ds_auth_response] if options[:three_ds_auth_response]
+        post[:operationId] = options[:operation_id] if options[:operation_id]
+      end
```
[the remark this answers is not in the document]
jherreraa: Suggestion added! Thanks!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,77 +30,162 @@ def initialize(options = {})
...
+        post[:order][:customerInfo] = {
+          cardHolderName: payment_method.name,
+          cardHolderEmail: options[:email]
+        }.compact
+      end
+
+      def add_address(post, options)
+        if options[:billing_address] || options[:address]
+          post[:order][:customerInfo][:billingAddress] = {
+            name: address[:name],
```
Heavyblade, line 110: 🔴 Warning 🔴 

Code is failing because `address` is not defined

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,77 +30,162 @@ def initialize(options = {})
...
+            street: address[:address1],
+            additionalInfo: address[:address2],
+            city: address[:city],
+            postCode: address[:zip],
+            country: address[:country]
+          }.compact
+        end
+        if options[:shipping_address]
+          post[:order][:customerInfo][:shippingAddress] = {
+            name: address[:name],
```
Heavyblade, line 120: 🔴 Warning 🔴 
 Same here

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,77 +30,162 @@ def initialize(options = {})
...
+            street: address[:address1],
+            additionalInfo: address[:address2],
+            city: address[:city],
+            postCode: address[:zip],
+            country: address[:country]
+          }.compact
+        end
+      end
+
```
Heavyblade, line 129: 💬 Comment:

Let's reduce the implementation scope and leave `add_recurrence` and `add_exemptions` to the time those features get requested

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,77 +30,162 @@ def initialize(options = {})
...
+        add_recurrence(post, options) unless options[:operation_id]
+        add_exemptions(post, options)
+        add_3ds_params(post, options)
+      end
 
       def add_reference(post, authorization) end
 
       def add_auth_purchase(post, money, payment, options) end
 
+      def parse(body)
```
Heavyblade, line 157: 💬 Style/Comment:

Maybe a Dry approach for the same result;

```ruby
      def parse(body = '{}')
        JSON.parse(body)
      end

```

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,77 +30,162 @@ def initialize(options = {})
...
+          success_from(response),
+          message_from(response),
+          response,
+          authorization: authorization_from(response),
+          test: test?,
+          error_code: error_code_from(response)
+        )
       end
 
       def request_headers(options)
```
Heavyblade, line 184: 💬 Style/ Comment:

Just to reduce unnecessary noise:

```Ruby
      def request_headers(options)
        {
          'X-Api-Key' => @api_key,
          'Correlation-Id' => options[:order_id] || SecureRandom.uuid
        }
      end

```

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,77 +30,162 @@ def initialize(options = {})
...
         headers = {
           'Content-Type' => 'application/json',
           'X-Api-Key' => @api_key,
-          'Correlation-Id' => options.dig(:order, :order_id) || SecureRandom.uuid
+          'Correlation-Id' => options.dig(:order_id) || SecureRandom.uuid
```
Heavyblade, line 188: ❓ Question:

Any particular reason to use one level `dig` instead of the simpler `[]` ?

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -112,23 +197,23 @@ def build_request_url(action, id = nil)
       end
 
       def success_from(response)
-        response == 'SUCCESS'
+        response.dig('operation', 'operationResult') == 'PENDING' || response.dig('operation', 'operationResult') == 'FAILED' || response.dig('operation', 'operationResult') == 'AUTHORIZED'
```
Heavyblade, line 200: ❓ Question:

How does a 'FAILED' `operationResult` become a successful response ?

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -112,23 +197,23 @@ def build_request_url(action, id = nil)
...
       end
 
-      def message_from(succeeded, response)
-        if succeeded
-          'Succeeded'
+      def message_from(response)
```
Heavyblade, line 203: 💬 Style/Comment:

perhaps Something DRYer:

```ruby
def message_from(response)
  response['errors'] || response.dig('operation', 'operationResult')
end
```

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -112,23 +197,23 @@ def build_request_url(action, id = nil)
...
       end
 
       def authorization_from(response)
-        response.dig('latest_payment_attempt', 'payment_intent_id') unless response
+        response.dig('operation', 'operationId') unless response
       end
 
       def error_code_from(response)
-        response['provider_original_response_code'] || response['code'] unless success_from(response)
+        response.dig('errors')&.first&.dig('code') unless success_from(response)
```
Heavyblade, line 216: 💬 Comment ❓ Question:

Comment => dig can handle this logic directly:

```ruby
def error_code_from(response)
  response.dig('errors', 0, 'code')
end
```

Question => checking just the first error `0` is because of the API response restructure or we are just going to show just the first message ?

test/remote/gateways/remote_xpay_test.rb
```diff
@@ -2,14 +2,47 @@
...
   end
 
   def test_successful_purchase
     response = @gateway.purchase(@amount, @credit_card, @options)
-    assert response
+    assert_success response
+    assert_match 'PENDING', response.message
+  end
+
+  def test_failed_purchase
```
Heavyblade, line 37: 💬 Comment:

The only thing that we can control is the initial request, may be can at least check if the 3DS ACS URL and the other params are present ?

Heavyblade (commented):
tons of work @jherreraa we are really close now, left some comments for your consideration

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -56,51 +58,133 @@ def refund(amount, authorization, options = {})
...
       private
 
-      def add_invoice(post, money, options) end
+      def add_invoice(post, amount, options)
+        currency = options[:currency] || currency(amount)
+        post[:order] = {}
+        post[:order][:orderId] = options[:order_id] if options[:order_id]
+        post[:order][:amount] = localized_amount(amount, currency)
+        post[:order][:currency] = currency
+      end
```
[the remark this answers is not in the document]
jherreraa: Thanks for the suggestion. Solved!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: Noted... Corrected!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: Noted! Thanks!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: The status of the gateway response for failed transactions is 200 so that why this were included this way

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: Same as the other one!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: None, suggestion taken!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: Just the first message, so far the tests only shows one error.

test/remote/gateways/remote_xpay_test.rb
jherreraa: Noted, additional asserts added!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: added! thanks!

lib/active_merchant/billing/gateways/xpay.rb
Heavyblade: 🤔 mmm:

That answer re-inforces the issue, if the `operationResult` returns 'FAILED' the method is going to return true which is incorrect.

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -15,7 +15,7 @@ class XpayGateway < Gateway
 
       ENDPOINTS_MAPPING = {
         purchase: 'orders/2steps/payment',
-        authorize: 'orders/2steps/init',
+        preauth: 'orders/2steps/init',
```
naashton, line 18: What's the reason behind calling this `preauth` instead of just `authorize` as it was?

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,79 +30,161 @@ def initialize(options = {})
...
+
+      def add_recurrence(post, options)
+        post[:recurrence] = { action: options[:recurrence] || 'NO_RECURRING' }
+      end
+
+      def add_exemptions(post, options)
+        post[:exemptions] = options[:exemptions] || 'NO_PREFERENCE'
+      end
+
+      def add_3ds_params(post, options)
```
naashton, line 138: Any case where `add_auth_purchase_params` will also `add_3ds_params`, so we might want to `return` from this method if we aren't performing a 3DS transaction.

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -30,79 +30,161 @@ def initialize(options = {})
...
+            additionalInfo: address[:address2],
+            city: address[:city],
+            postCode: address[:zip],
+            country: address[:country]
+          }.compact
+        end
+      end
+
+      def add_recurrence(post, options)
+        post[:recurrence] = { action: options[:recurrence] || 'NO_RECURRING' }
```
naashton, line 131: If the txn is not recurring, do we have to always default this value to `NO_RECURRING`? Same question for `add_exemptions`.

naashton (changes requested):
A few questions, but generally looks pretty good. Only other thing, make sure the GHA checks pass.

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: the endpoint do not correspond to an authorize transaction, according to documentation this is the first step to the purchase and authorize transactions which are described with some other fields included in the request body.

lib/active_merchant/billing/gateways/xpay.rb
naashton: I think @Heavyblade is correct here. The gateway returning `200` just tells us the request was successful, but the concern for `success_from` is to determine whether or not the transaction was successful.

test/remote/gateways/remote_xpay_test.rb
```diff
@@ -2,14 +2,70 @@
...
+    @amount = 100
+    @credit_card = credit_card(
+      '5186151650005008',
+      month: 12,
+      year: 2026,
+      first_name: 'Jim',
+      last_name: 'Smith',
+      verification_value: '123',
+      brand: 'master'
+    )
```
naashton, line 15: Nit: Unless the gateway test card checks all of these fields, you can just call the `credit_card` method and give it the `options` you want to change, so: 

```ruby
credit_card({card_number}, {brand: 'master', verification_value: '456'})
```

test/remote/gateways/remote_xpay_test.rb
```diff
@@ -2,14 +2,70 @@
...
+
+    @options = {
+      order_id: SecureRandom.alphanumeric(10),
+      billing_address: address,
+      order: {
+        currency: 'EUR',
+        amount: @amount,
+        customer_info: {
+          card_holder_name: 'Jim Smith',
+          card_holder_email: nil
```
naashton, line 25: `email` and `name` are common fields so you shouldn't need these

naashton (changes requested):
One change for the `success_from` and a few small nits from unused test values.

If you make the requested change and include a `CHANGELOG` I can approve and merge this up.

test/remote/gateways/remote_xpay_test.rb
jherreraa: Noted! Changed!

test/remote/gateways/remote_xpay_test.rb
jherreraa: not needed values removed! Thanks for the review!

naashton (approved):
Nice work! Thanks for your diligent work on this.

PR #4970, opened 2023-11-29
Rapyd: Enable new auth mode payment_redirect

javierpedrozaing:
Description
-------------------------
[SER-970](https://spreedly.atlassian.net/browse/SER-974)

This commit enable a new mode of authentication for Rapyd. 
This new mode only apply for create payments

Unit test
-------------------------
Finished in 0.189728 seconds.

41 tests, 188 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

100% passed

216.10 tests/s, 990.89 assertions/s

Remote test
-------------------------
Finished in 208.640892 seconds.

47 tests, 123 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 87.234% passed

0.23 tests/s, 0.59 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -1,8 +1,13 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class RapydGateway < Gateway
+      class_attribute :payment_redirect_test, :payment_redirect_live
+
       self.test_url = 'https://sandboxapi.rapyd.net/v1/'
       self.live_url = 'https://api.rapyd.net/v1/'
+
+      self.payment_redirect_test = 'https://sandboxpayment-redirect.rapyd.net/v1/'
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -294,10 +299,17 @@ def parse(body)
         JSON.parse(body)
       end
 
+      def url(action, mode = :default)
+        if mode == 'payment_redirect' && action == 'payments'
```
naashton, line 303: The default argument is using the symbol `:` but your conditional checks against a string. Will the conditional still succeed if the value is `:payment_redirect`?

javierpedrozaing: Hi @naashton thanks good point, I've added an update if for some reason the mode value is a symbol.

naashton (approved):
Awesome work, thanks @javierpedrozaing!

PR #4971, opened 2023-11-29
Adyen: Provide ZZ as default country code

jcreiff:
According to Adyen's documentation, ZZ should be provided as the country code if the country is not known or is not collected from the cardholder:

https://docs.adyen.com/api-explorer/Payment/latest/post/authorise#request-billingAddress-country https://docs.adyen.com/api-explorer/Payment/latest/post/authorise#request-deliveryAddress-country

Sending an empty string or nil value results in a validation error from Adyen's API

CER-979
Local
5741 tests, 78708 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit
115 tests, 604 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote
142 tests, 461 assertions, 11 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 92.2535% passed
*These same 11 tests (mostly related to store/unstore functionality) also fail on master

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -499,7 +499,7 @@ module ActiveMerchant #:nodoc:
           post[:deliveryAddress][:postalCode] = address[:zip] if address[:zip]
           post[:deliveryAddress][:city] = address[:city] || 'NA'
           post[:deliveryAddress][:stateOrProvince] = get_state(address)
-          post[:deliveryAddress][:country] = address[:country] if address[:country]
+          post[:deliveryAddress][:country] = get_country(address)
         end
         return unless post[:bankAccount]&.kind_of?(Hash) || post[:card]&.kind_of?(Hash)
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -515,14 +515,18 @@ def add_billing_address(post, options, address)
...
+        post[:billingAddress][:country] = get_country(address)
         post[:telephoneNumber] = address[:phone_number] || address[:phone] || ''
       end
 
       def get_state(address)
         address[:state] && !address[:state].blank? ? address[:state] : 'NA'
       end
 
+      def get_country(address)
+        address[:country] && !address[:country].blank? ? address[:country] : 'ZZ'
```
DustinHaefele, line 527: Nitpick so feel no obligation to change: but could this be simplified to the following? 
```suggestion
        address[:country].present? ? address[:country] : 'ZZ'
```

DustinHaefele (approved):
Looks good to me! left a little question but certainly not blocking.

lib/active_merchant/billing/gateways/adyen.rb
jcreiff: Much simpler, I like it! Updated 🎉

PR #4972, opened 2023-11-30
RedsysRest: Update to $0 verify

almalee24:
Unit:
21 tests, 91 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
16 tests, 45 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys_rest.rb
+++ b/lib/active_merchant/billing/gateways/redsys_rest.rb
@@ -82,7 +82,8 @@ module ActiveMerchant #:nodoc:
         authorize:  '1',
         capture:    '2',
         refund:     '3',
-        cancel:     '9'
+        cancel:     '9',
+        verify:     '7'
       }
 
       # These are the text meanings sent back by the acquirer when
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/redsys_rest.rb
```diff
@@ -82,7 +82,8 @@ class RedsysRestGateway < Gateway
         authorize:  '1',
         capture:    '2',
         refund:     '3',
-        cancel:     '9'
+        cancel:     '9',
+        verify:     '7'
```
aenand, line 86: 🤔  I was torn on whether to make `verify` a transaction type of 7 or not, what led you to use 7?

aenand (commented):
Everything looks good! Just curious about one thing

lib/active_merchant/billing/gateways/redsys_rest.rb
almalee24: Because in [their docs](https://pagosonline.redsys.es/funcionalidades-preautenticacion.html) it points out that for Rest it should be transaction type of 7.

aenand: When I read that I interpreted it as like a pre-authorization because you can pass in an amount and then you can "complete" the charge by doing a follow up with transaction 8. We can go with a `7` type though!

aenand (approved):
LGTM :shipit:

PR #4975, opened 2023-12-01
Checkout: Update reponse message for 3DS transactions

almalee24:
After a redirect from 3DS transactions the verify_payment method is usually called which returns the response_summary inside actions.

Unit:
64 tests, 378 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
98 tests, 236 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -88,7 +88,7 @@ module ActiveMerchant #:nodoc:
         authorize(0, credit_card, options)
       end
 
-      def verify_payment(authorization, option = {})
+      def verify_payment(authorization, options = {})
         commit(:verify_payment, nil, options, authorization, :get)
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -552,13 +552,16 @@ def success_from(action, response)
...
         if succeeded
           'Succeeded'
         elsif response['error_type']
           response['error_type'] + ': ' + response['error_codes'].first
+        elsif options[:threeds_response_message]
+          response_summary = response['response_summary'] || response.dig('actions', 0, 'response_summary')
+          response_summary || response['status'] || response['message'] || 'Unable to read error message'
         else
-          response['response_summary'] || response['response_code'] || response['status'] || response['message'] || 'Unable to read error message'
+          response['response_summary'] || response['status'] || response['message'] || 'Unable to read error message'
```
aenand, line 564: Why drop `response_code` from possible message options?

aenand (changes requested):

lib/active_merchant/billing/gateways/checkout_v2.rb
almalee24: Great catch, didn't realize I had removed it. Will update!

aenand (approved):
LGTM 🚀

PR #4978, opened 2023-12-04
SER-705 Nexi Xpay Void, Refund, Verify

jherreraa:
Description
-------------------------
[SER-705](https://spreedly.atlassian.net/browse/SER-705)

This commit add NexiXpay operations, void, refund, verify

### DIFF
--- a/lib/active_merchant/billing/gateways/xpay.rb
+++ b/lib/active_merchant/billing/gateways/xpay.rb
@@ -15,11 +15,12 @@ module ActiveMerchant #:nodoc:
 
       ENDPOINTS_MAPPING = {
         purchase: 'orders/2steps/payment',
+        authorize: 'orders/2steps/payment',
         preauth: 'orders/2steps/init',
-        capture: 'operations/{%s}/captures',
+        capture: 'operations/%s/captures',
         verify: 'orders/card_verification',
-        void: 'operations/{%s}/cancels',
-        refund: 'operations/{%s}/refunds'
+        void: 'operations/%s/cancels',
+        refund: 'operations/%s/refunds'
       }
 
       def initialize(options = {})
@@ -28,31 +29,35 @@ module ActiveMerchant #:nodoc:
         super
       end
 
+      def preauth(amount, payment_method, options = {})
+        post = {}
+        add_transaction_params_commit(:preauth, amount, post, payment_method, options)
+      end
+
       def purchase(amount, payment_method, options = {})
         post = {}
-        add_auth_purchase_params(post, amount, payment_method, options)
-        action = options[:operation_id] ? :purchase : :preauth
-        commit(action, post, options)
+        add_transaction_params_commit(:purchase, amount, post, payment_method, options)
       end
 
       def authorize(amount, payment_method, options = {})
         post = {}
-        add_auth_purchase_params(post, amount, payment_method, options)
-        commit(:preauth, post, options)
+        add_transaction_params_commit(:authorize, amount, post, payment_method, options)
       end
 
       def capture(amount, authorization, options = {})
         post = {}
+        add_refund_capture_params(amount, post, options)
         commit(:capture, post, options)
       end
 
       def void(authorization, options = {})
-        post = {}
+        post = { description: options[:description] }
         commit(:void, post, options)
       end
 
       def refund(amount, authorization, options = {})
         post = {}
+        add_refund_capture_params(amount, post, options)
         commit(:refund, post, options)
       end
 
@@ -75,6 +80,27 @@ module ActiveMerchant #:nodoc:
       end
 
       private
+
+      def add_transaction_params_commit(action, amount, post, payment_method, options = {})
+        add_capture_type(post, options, action)
+        add_auth_purchase_params(post, amount, payment_method, options)
+        commit(action, post, options)
+      end
+
+      def add_capture_type(post, options, action)
+        case action
+        when :purchase
+          post[:captureType] = 'IMPLICIT'
+        when :authorize
+          post[:captureType] = 'EXPLICIT'
+        end
+      end
+
+      def add_refund_capture_params(amount, post, options)
+        post[:amount] = amount
+        post[:currency] = options[:order][:currency]
+        post[:description] = options[:order][:description]
+      end
 
       def add_invoice(post, amount, options)
         currency = options[:currency] || currency(amount)
@@ -151,19 +177,15 @@ module ActiveMerchant #:nodoc:
         add_3ds_params(post, options)
       end
 
-      def add_reference(post, authorization) end
-
-      def add_auth_purchase(post, money, payment, options) end
-
       def parse(body = {})
         JSON.parse(body)
       end
 
       def commit(action, params, options)
-        transaction_id = params.dig(:operation_id) unless action != 'capture'
+        transaction_id = transaction_id_from(params, options, action)
         begin
           url = build_request_url(action, transaction_id)
-          raw_response = ssl_post(url, params.to_json, request_headers(options))
+          raw_response = ssl_post(url, params.to_json, request_headers(options, action))
           response = parse(raw_response)
         rescue ResponseError => e
           response = e.response.body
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -29,30 +30,35 @@ def initialize(options = {})
...
-        add_auth_purchase_params(post, amount, payment_method, options)
         action = options[:operation_id] ? :purchase : :preauth
+        post = add_capture_type(options, action)
+        add_auth_purchase_params(post, amount, payment_method, options)
         commit(action, post, options)
       end
 
       def authorize(amount, payment_method, options = {})
-        post = {}
+        action = options[:operation_id] ? :authorize : :preauth
```
edgarv09, line 40: @jherreraa parce, could you explain to me this if. Why do you set authorize as action when operation_id is a truethy value? 


The same question applies to all operations.

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -29,30 +30,35 @@ def initialize(options = {})
...
         post = {}
+        add_refund_capture_params(amount, post, options)
         commit(:capture, post, options)
       end
 
       def void(authorization, options = {})
-        post = {}
+        post = {
+          description: options[:description]
+        }
```
edgarv09, line 55: one line

sinourain (changes requested):
I made some small comments, and I am left wondering why there are no remote tests for new transactions?

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -76,6 +82,23 @@ def scrub(transcript)
...
+      def add_capture_type(options, action)
+        post = {}
+        case action
+        when :purchase
+          post[:captureType] = 'IMPLICIT'
+        when :authorize
+          post[:captureType] = 'EXPLICIT'
+        end
+        post
+      end
```
edgarv09, line 94: The function's name isn't inherently wrong, but it implies adding the capture type to someone, as is the case with `add_refund_capture_params`. 
However, the function returns a new hash with or without values. You might want to consider a name that better reflects the function's behavior.

edgarv09: What do you think if you use constants, 

```ruby
   when PaymentActions::PURCHASE
    post[:captureType] = 'IMPLICIT'
  when PaymentActions::AUTHORIZE
    post[:captureType] = 'EXPLICIT'
  end
```
consider it also for the :operation_id setters.

test/remote/gateways/remote_xpay_test.rb
```diff
@@ -47,4 +47,25 @@ def test_failed_purchase
...
+  # end
+
+  # def test_successful_void ## test requires set up (purchase or auth through 3ds)
+  #   options = {
+  #     description: 'void operation message',
+  #     operation_id: '168467730273233329'
+  #   }
+  #   response = @gateway.void(@amount, options)
+  #   assert_success response
+  # end
```
edgarv09, line 70: why is this commented on?

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -29,30 +30,35 @@ def initialize(options = {})
       end
 
       def purchase(amount, payment_method, options = {})
-        post = {}
-        add_auth_purchase_params(post, amount, payment_method, options)
         action = options[:operation_id] ? :purchase : :preauth
```
edgarv09, line 33: why a purchase request could be converted into a preauth? 
What is the difference between authorize and preauth?

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -29,30 +30,35 @@ def initialize(options = {})
...
         commit(action, post, options)
       end
 
       def authorize(amount, payment_method, options = {})
-        post = {}
+        action = options[:operation_id] ? :authorize : :preauth
+        post = add_capture_type(options, action)
         add_auth_purchase_params(post, amount, payment_method, options)
-        commit(:preauth, post, options)
+        commit(action, post, options)
```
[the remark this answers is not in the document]
jherreraa: Understood! Modified! Thanks

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -179,12 +198,28 @@ def commit(action, params, options)
...
+        when :refund, :capture
+          headers.merge!('Idempotency-Key' => SecureRandom.uuid)
+        end
+        headers
+      end
+
+      def transaction_id_from(params, options, action = nil)
+        case action
+        when :refund, :capture, :void
+          transaction_id = options[:operation_id]
```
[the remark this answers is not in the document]
jherreraa: Noted! suggestion added!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -179,12 +198,28 @@ def commit(action, params, options)
...
+        else
+          transaction_id = params.dig(:operation_id)
+        end
+        transaction_id
```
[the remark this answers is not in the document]
jherreraa: Noted! suggestion added!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: This is because 3ds flow is required for every authorize and purchase transaction and the flow works according to the documentation [here](https://developer.nexi.it/en/modalita-di-integrazione/server-to-server/pagamento-2-steps) requiring to start with `init` (preauth) and then with `payment` (authorize and purchase)

test/remote/gateways/remote_xpay_test.rb
jherreraa: in order to perform those kind of transactions a successful 3ds transaction should be executed and then void, refund or capture nevertheless this flow requires the previous setup done.

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: hey! thanks for the question, the logic works this way because of the [2 steps flow](https://developer.nexi.it/en/modalita-di-integrazione/server-to-server/pagamento-2-steps) that implies in the payment step send the operation id.
Let me know if you need any further explanation.

edgarv09: we discussed this via huddle and I expected some changes.

Checking the flow preauth is an independent action for the gateway so we need to add it as a function instead of the fuzzy logic of changing the action if the operation_id comes or not.

edgarv09 (changes requested):

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: Thanks for the suggestion! Changed!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -15,44 +15,59 @@ class XpayGateway < Gateway
...
       end
 
       def void(authorization, options = {})
-        post = {}
+        post = { description: options[:description] }
         commit(:void, post, options)
       end
 
-      def refund(amount, authorization, options = {})
+      def refund(amount, options = {})
```
naashton, line 68: `authorization` is a part of the `refund` method signature, so it's a required argument

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -15,44 +15,59 @@ class XpayGateway < Gateway
...
 
+      module PaymentActions
+        PREAUTH = :preauth
+        AUTHORIZE = :authorize
+        PURCHASE = :purchase
+        CAPTURE = :capture
+        VERIFY = :verify
+        VOID = :void
+        REFUND = :refund
+      end
```
naashton, line 34: This seems like overkill for this scenario. I think for your `case/when` statement, you can just check the `action` type and make your decision based on that `action` key to set the `captureType`

naashton (changes requested):
I think there might be some overlap between this and the other PR. Once we finish up the other PR, we can rebase this and re-review this.

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: Understood! Corrected!

lib/active_merchant/billing/gateways/xpay.rb
jherreraa: Understood! Corrected!

PR #4980, opened 2023-12-07
Rapyd: Enable idempotent request support

javierpedrozaing:
Description
-------------------------
[SER-1023](https://spreedly.atlassian.net/browse/SER-1023)

This commit adds an idempotency attribute to headers in order to support idempotent requests, Rapyd manage it, taking into account the idempotency value and the amount to determine if a transaction is idempotent or no.

Unit test
-------------------------
Finished in 0.504217 seconds.

42 tests, 196 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

83.30 tests/s, 388.72 assertions/s

Remote test
-------------------------
49 tests, 142 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 97.9592% passed

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/rapyd.rb
+++ b/lib/active_merchant/billing/gateways/rapyd.rb
@@ -112,6 +112,11 @@ module ActiveMerchant #:nodoc:
         add_ewallet(post, options)
         add_payment_fields(post, options)
         add_payment_urls(post, options)
+        add_idempotency(options)
+      end
+
+      def add_idempotency(options)
+        @options[:idempotency] = options[:idempotency_key] if options[:idempotency_key]
       end
 
       def add_address(post, creditcard, options)
@@ -347,8 +352,9 @@ module ActiveMerchant #:nodoc:
           'access_key' => @options[:access_key],
           'salt' => salt,
           'timestamp' => timestamp,
-          'signature' => generate_hmac(rel_path, salt, timestamp, payload)
-        }
+          'signature' => generate_hmac(rel_path, salt, timestamp, payload),
+          'idempotency' => @options[:idempotency]
+        }.delete_if { |_, value| value.nil? }
       end
 
       def generate_hmac(rel_path, salt, timestamp, payload)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -81,6 +81,7 @@
 * IPG: Allow for Merchant Aggregator credential usage [DustinHaefele] #4986
 * Adyen: Add support for `metadata` object [rachelkirk] #4987
 * Xpay: New adapter basic operations added [jherreraa] #4669
+* Rapyd: Enable idempotent request support [javierpedrozaing] #4980
 
 == Version 1.135.0 (August 24, 2023)
 * PaymentExpress: Correct endpoints [steveh] #4827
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -336,18 +341,19 @@ def revert_json_html_encoding!(string)
       def api_request(method, url, rel_path, params)
         params == {} ? body = '' : body = params.to_json
         revert_json_html_encoding!(body) if defined?(ActiveSupport::JSON::Encoding) && ActiveSupport::JSON::Encoding.escape_html_entities_in_json
-        parse(ssl_request(method, url, body, headers(rel_path, body)))
+        parse(ssl_request(method, url, body, headers(rel_path, body, params)))
```
sinourain, line 344: Looks like params is added but never used

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -112,6 +112,11 @@ def add_auth_purchase(post, money, payment, options)
         add_ewallet(post, options)
         add_payment_fields(post, options)
         add_payment_urls(post, options)
+        add_idempotency(options)
+      end
+
+      def add_idempotency(options)
+        @options[:idempotency] = options[:idempotency] if options[:idempotency]
```
sinourain, line 119: Looks like the if statement is not needed

sinourain (commented):
LGTM

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -336,18 +341,19 @@ def revert_json_html_encoding!(string)
...
         salt = SecureRandom.base64(12)
         timestamp = Time.new.to_i.to_s
         {
           'Content-Type' => 'application/json',
           'access_key' => @options[:access_key],
           'salt' => salt,
           'timestamp' => timestamp,
-          'signature' => generate_hmac(rel_path, salt, timestamp, payload)
+          'signature' => generate_hmac(rel_path, salt, timestamp, payload),
+          'idempotency' => @options[:idempotency]
```
edgarv09, line 356: Is this value allowed to be nil?

lib/active_merchant/billing/gateways/rapyd.rb
edgarv09: why idempotency is an instance variable?

lib/active_merchant/billing/gateways/rapyd.rb
javierpedrozaing: oh you're right, thanks, I'll update it

lib/active_merchant/billing/gateways/rapyd.rb
javierpedrozaing: good point, if we don't set a value for idempotency we should not send it, tahnks, I'll update it

javierpedrozaing:
Hi @naashton I will update the idempotency key name according a suggestion from Cristian on the CORE PR. I let you know when it is ready

javierpedrozaing:
> Hi @naashton I will update the idempotency key name according a suggestion from Cristian on the CORE PR. I let you know when it is ready

Update ready. thanks

naashton (approved):
Nice work!

PR #4982, opened 2023-12-08
VantiveExpress: Fix eci bug

almalee24:
If a ApplePay or GooglePay does not have an eci value then it would be considered not secure for Visa/MC/Discover so we need to send 6 and for Amex send 9. MotoECICode is considered a required value for ApplePay and GooglePay.

### DIFF
--- a/lib/active_merchant/billing/gateways/vantiv_express.rb
+++ b/lib/active_merchant/billing/gateways/vantiv_express.rb
@@ -1,6 +1,6 @@
 require 'nokogiri'
 require 'securerandom'
-
+require 'pry'
 module ActiveMerchant #:nodoc:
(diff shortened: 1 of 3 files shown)


aenand:
Can you link the docs where you saw when to send 6 and 9 as default values?

almalee24:
> Can you link the docs where you saw when to send 6 and 9 as default values?

https://developerengine.fisglobal.com/apis/express/features/in-app

aenand (approved):
LGTM :shipit:

aenand:
Can you also post the results of the remote tests?

almalee24:
> Can you also post the results of the remote tests?

Oh yes, forgot to add that. I'll update the commit message

PR #4983, opened 2023-12-08
Wompi: Add support for `tip_in_cents`

rachelkirk:
CER-1062

Remote Tests:
14 tests, 27 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 71.4286% passed
Some remote tests are failing due to a missing email, will address in a different ticket.

Unit Tests:
13 tests, 65 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Local Tests:
5740 tests, 78712 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/wompi.rb
+++ b/lib/active_merchant/billing/gateways/wompi.rb
@@ -34,6 +34,7 @@ module ActiveMerchant #:nodoc:
           public_key: public_key
         }
         add_invoice(post, money, options)
+        add_tip_in_cents(post, options)
         add_card(post, payment, options)
 
         commit('sale', post, '/transactions_sync')
@@ -139,6 +140,10 @@ module ActiveMerchant #:nodoc:
         post[:installments] = installments
         post[:card_holder] = card.name
         post[:cvc] = cvc if cvc && !cvc.empty?
+      end
+
+      def add_tip_in_cents(post, options)
+        post[:tip_in_cents] = options[:tip_in_cents].to_i if options[:tip_in_cents]
       end
 
       def parse(body)
--- a/test/remote/gateways/remote_wompi_test.rb
+++ b/test/remote/gateways/remote_wompi_test.rb
@@ -34,6 +34,11 @@ class RemoteWompiTest < Test::Unit::TestCase
     assert_success response
   end
 
+  def test_successful_purchase_with_tip_in_cents
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(tip_in_cents: 300))
+    assert_success response
+  end
+
   def test_failed_purchase
     response = @gateway.purchase(@amount, @declined_card, @options)
     assert_failure response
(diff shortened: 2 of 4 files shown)


test/remote/gateways/remote_wompi_test.rb
```diff
@@ -34,6 +34,13 @@ def test_successful_purchase_without_cvv
     assert_success response
   end
 
+  def test_successful_purchase_with_tip_in_cents
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(tip_in_cents: 300))
+    assert_success response
+    # response_data = response.params['data']
+    # assert_equal response_data.dig('tip_in_cents'), 300
+  end
```
rachelkirk, line 42: this test feels kind of lame and anti-climactic, left the commented portions so it would be easy to uncomment and run if desired. It will cause test to fail because it comes back as null in the response which is a bit mystifying. I tried playing around with putting the `add_tip_in_cents` method within the `add_invoice` method to exactly mimic their doc example attached in Jira, but it didn't make a difference with it coming back in the response as a non null value. Thought it best to keep 'em separated. If these ramblings don't make sense, I'll be happy to clarify!

lib/active_merchant/billing/gateways/wompi.rb
```diff
@@ -141,6 +142,10 @@ def add_basic_card_info(post, card, options)
         post[:cvc] = cvc if cvc && !cvc.empty?
       end
 
+      def add_tip_in_cents(post, options)
+        post[:tip_in_cents] = options[:tip_in_cents] if options[:tip_in_cents]
+      end
+
       def parse(body)
```
rachelkirk, line 149: Reminder to myself to not forget a Changelog entry!

test/remote/gateways/remote_wompi_test.rb
jcreiff: Curious! I wonder if we can confirm with Wompi that this value should be echoed back in the response 🤔

lib/active_merchant/billing/gateways/wompi.rb
```diff
@@ -141,6 +142,10 @@ def add_basic_card_info(post, card, options)
         post[:cvc] = cvc if cvc && !cvc.empty?
       end
 
+      def add_tip_in_cents(post, options)
+        post[:tip_in_cents] = options[:tip_in_cents] if options[:tip_in_cents]
```
jcreiff, line 146: I notice that the other places where `amount_in_cents` is being set, there's a `to_i` being called to ensure that we're passing an integer. I wonder if we need to do the same here for `tip_in_cents`, in the event that an amount is passed in as a string?

test/remote/gateways/remote_wompi_test.rb
rachelkirk: They confirmed it's expected to be null in sandbox but will show amount in production. Yay.

lib/active_merchant/billing/gateways/wompi.rb
rachelkirk: I added the `to_i` just in case. Tested it by changing amount in remote test to a String and it behaved as expected. Did not keep that change permanent or add a separate test because it seemed a bit overkill but can do that if you think it would be useful.

jcreiff (approved):
Looks good to me!

PR #4990, opened 2023-12-18
SagePay: Support v4

aenand:
ECS-3323

To de-risk the update of the SagePay gateway integration to move from V3.00 to V4.00 this commit adds the ability to specify an override protocol version via the transaction while also preserving the ability to override
via a gateway class

Remote:
Commented out test also failing on master
37 tests, 108 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/sage_pay.rb
+++ b/lib/active_merchant/billing/gateways/sage_pay.rb
@@ -78,6 +78,7 @@ module ActiveMerchant #:nodoc:
 
       def initialize(options = {})
         requires!(options, :login)
+        @protocol_version = options.fetch(:protocol_version, '3.00')
         super
       end
 
@@ -86,6 +87,7 @@ module ActiveMerchant #:nodoc:
 
         post = {}
 
+        add_override_protocol_version(options)
         add_amount(post, money, options)
         add_invoice(post, options)
         add_payment_method(post, payment_method, options)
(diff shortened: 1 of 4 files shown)


aenand:
Converting this to draft so that it is a feature branch PR for the basis of the v4 upgrade

lib/active_merchant/billing/gateways/sage_pay.rb
```diff
@@ -405,14 +442,20 @@ def post_data(action, parameters = {})
         parameters.update(
           Vendor: @options[:login],
           TxType: TRANSACTIONS[action],
-          VPSProtocol: @options.fetch(:protocol_version, '3.00')
+          VPSProtocol: @protocol_version
```
almalee24, line 445: I'm just curious why you decided to use add_override_protocol_version to update the version instead of just updating this line to be something like options[:protocol_version] || @protocol_version?

aenand: `post_data` does not get the `options` hash, nor does the method that calls it (`commit`) so rather than pass in options to two additional methods and update all the `commit` invocations it felt a little simpler to add the method. I'm open to either way though!

almalee24: Gotcha, makes sense!

almalee24 (approved):

PR #4992, opened 2023-12-18
Braintree Blue: Add more payment detail objects

yunnydang:
This adds the apple pay, google pay, and network token details object. Will also surface the payment instrument type so we can know which one was used

Local:
5754 tests, 78658 assertions, 0 failures, 26 errors, 0 pendings, 0 omissions, 0 notifications
99.5481% passed

Unit:
102 tests, 216 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
114 tests, 548 assertions, 6 failures, 4 errors, 0 pendings, 0 omissions, 0 notifications
91.2281% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -607,6 +607,23 @@ module ActiveMerchant #:nodoc:
           'issuing_bank'        => transaction.credit_card_details.issuing_bank
         }
 
+        network_token_details = {
+          'debit'               => transaction.network_token_details.debit,
+          'prepaid'             => transaction.network_token_details.prepaid,
+          'issuing_bank'        => transaction.network_token_details.issuing_bank
+        }
+
+        google_pay_details = {
+          'debit'               => transaction.google_pay_details.debit,
+          'prepaid'             => transaction.google_pay_details.prepaid
+        }
+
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -607,6 +607,23 @@ def transaction_hash(result)
           'issuing_bank'        => transaction.credit_card_details.issuing_bank
         }
 
+        network_token_details = {
+          'debit'               => transaction.network_token_details.debit,
+          'prepaid'             => transaction.network_token_details.prepaid,
+          'issuing_bank'        => transaction.network_token_details.issuing_bank
+        }
+
+        google_pay_details = {
```
jcreiff, line 616: @yunnydang Any reason why no `issuing_bank` in this block? I notice it's omitted from the test assertions too.

yunnydang: Hye @jcreiff ! According to their [docs](https://developer.paypal.com/braintree/docs/reference/response/transaction/ruby#:~:text=of%20the%20request.-,google_pay_details,-obj), BT does not return the issuing_bank field for google pay transactions, hoping this helps!

jcreiff (approved):
👍🏻

PR #4995, opened 2023-12-20
HiPay: Scrub/Refund/Void

gasb150:
SER-720
SER-721
----
Summary:
----
Adding to the HiPay gateway support for scrub, refund, and void.

HiPay allows partial capture and partial refunds, this commit include the amount to use it.

Tests
----
Remote Test:
Finished in 6.627757 seconds.
15 tests, 62 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit Tests:
21 tests, 61 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

RuboCop:
787 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/hi_pay.rb
+++ b/lib/active_merchant/billing/gateways/hi_pay.rb
@@ -51,22 +51,41 @@ module ActiveMerchant #:nodoc:
       end
 
       def capture(money, authorization, options)
-        post = {}
-        post[:operation] = 'capture'
-        post[:currency] = (options[:currency] || currency(money))
-        transaction_ref, _card_token, _payment_product = authorization.split('|')
-        commit('capture', post, { transaction_reference: transaction_ref })
+        reference_operation(money, authorization, options.merge({ operation: 'capture' }))
       end
 
       def store(payment_method, options = {})
         tokenize(payment_method, options.merge({ multiuse: '1' }))
       end
 
-      def scrub(transcrip)
-        # code
+      def refund(money, authorization, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/hi_pay.rb
```diff
@@ -51,23 +51,42 @@ def authorize(money, payment_method, options = {})
...
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Basic )[\w =]+), '\1[FILTERED]').
+          gsub(%r((card_number=)\w+), '\1[FILTERED]\2').
+          gsub(%r((cvc=)\w+), '\1[FILTERED]\2')
       end
 
       private
 
+      def maintenance_operation(money, authorization, options)
```
aenand, line 82: Can we rename this to be something a bit more specific. Maybe `reference_operation`?

aenand (commented):
This looks good to me! Just one request for re-naming

lib/active_merchant/billing/gateways/hi_pay.rb
gasb150: In this case, the name was according to those operations' names in the documentation. https://developer.hipay.com/api-explorer/api-online-payments#/maintenance/performMaintenanceOperation.

But yes, I can rename it.

aenand (approved):
LGTM :shipit:

naashton (approved):

PR #5002, opened 2024-01-05
Shift4: update response mapping

rachelkirk:
CER_1094

This change updates a few response methods to return more accurate codes and messages instead of the currently existing standard error code mapping. The previously implemented standard error code mapping was not great. It was returning messages not direct from the gateway, which was perhaps an old standard but there have been requests to make the messages returned aligned with the gateway's actual response.

Failed sandbox transactions do not return the `hostResponse`, which is from the card networks, but they do return `responseCode: D`. I've made some extensive updates to test cases and responses to more accurately reflect the gateway's responses in production transactions.

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4.rb
+++ b/lib/active_merchant/billing/gateways/shift4.rb
@@ -20,22 +20,6 @@ module ActiveMerchant #:nodoc:
         'add' => 'tokens',
         'verify' => 'cards'
       }
-      STANDARD_ERROR_CODE_MAPPING = {
-        'incorrect_number' => STANDARD_ERROR_CODE[:incorrect_number],
-        'invalid_number' => STANDARD_ERROR_CODE[:invalid_number],
-        'invalid_expiry_month' => STANDARD_ERROR_CODE[:invalid_expiry_date],
-        'invalid_expiry_year' => STANDARD_ERROR_CODE[:invalid_expiry_date],
-        'invalid_cvc' => STANDARD_ERROR_CODE[:invalid_cvc],
-        'expired_card' => STANDARD_ERROR_CODE[:expired_card],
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -284,13 +268,21 @@ def parse(body)
...
       def message_from(action, response)
-        success_from(action, response) ? 'Transaction successful' : (error(response)&.dig('longText') || 'Transaction declined')
+        success_from(action, response) ? 'Transaction successful' : (error(response)&.dig('longText') || response['result'].first&.dig('transaction', 'hostResponse', 'reasonDescription') || 'Transaction declined')
       end
 
       def error_code_from(action, response)
-        return unless success_from(action, response)
-
-        STANDARD_ERROR_CODE_MAPPING[response['primaryCode']]
+        code = response['result'].first&.dig('transaction', 'responseCode')
```
yunnydang, line 275: good work! Looks adequately guarded

yunnydang (approved):
Approving but highly recommended a member from Pathfinder to also grant their blessings

DustinHaefele (approved):
This looks good to me. I like pulling the error codes from the actual response. The only question I have is there any need to notify customers since some of the error messages are changing to a more generic "Transaction declined"? ... I'm approving since that is more in your team's wheel house to make that decision but I wanted to mention it.

yunnydang:
> This looks good to me. I like pulling the error codes from the actual response. The only question I have is there any need to notify customers since some of the error messages are changing to a more generic "Transaction declined"? ... I'm approving since that is more in your team's wheel house to make that decision but I wanted to mention it.

Heya @DustinHaefele just responding on Rachels behalf since shes out today, but here's what she said: "Transaction Declined will actually be more specific unless a hostResponse is not returned."

PR #5005, opened 2024-01-10
Update Rubocop to 1.14.0

almalee24:


curiousepic:
Nice work, this is a huge undertaking. Let's add a note in the commit that a lot of code was updated, thought; otherwise it just sounds like a gem bump!

almalee24:
> Nice work, this is a huge undertaking. Let's add a note in the commit that a lot of code was updated, though; otherwise it just sounds like a gem bump!

I've made the update

curiousepic (approved):

curiousepic:
Approving - as a note, the CI tests fail due to a Ruby versioning issue (which will be resolved soon) but pass locally.

PR #4888, opened 2023-09-12
Add payment_data to network_tokenization_credit_card

almalee24:
Some gateways required use to send the encrypted payment_data for ApplePay and GooglePay instead of the decrypted data.

Unit:
4 tests, 13 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

PR #5015, opened 2024-01-23
GlobalCollect: Add support for encrypted payment data

almalee24:


DustinHaefele (commented):
I think the code looks good. Can you add a test case for this though?

almalee24:
> I think the code looks good. Can you add a test case for this though?

Oh yes, I'll add a unit test. Can't do a remote test atm

almalee24:
> I think the code looks good. Can you add a test case for this though?

I've added the unit test but it's currently failing because I still need to merge [this](https://github.com/activemerchant/active_merchant/pull/4888) which will hopefully happen today

DustinHaefele (approved):

PR #5027, opened 2024-01-29
Kushki: add the product_details field

yunnydang:
This will add the product_details field

Local:
5807 tests, 78885 assertions, 0 failures, 26 errors, 0 pendings, 0 omissions, 0 notifications
99.5523% passed

Unit:
19 tests, 122 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
25 tests, 68 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
84% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/kushki.rb
+++ b/lib/active_merchant/billing/gateways/kushki.rb
@@ -102,6 +102,7 @@ module ActiveMerchant #:nodoc:
         add_months(post, options)
         add_deferred(post, options)
         add_three_d_secure(post, payment_method, options)
+        add_product_details(post, options)
 
         commit(action, post)
       end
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/kushki.rb
```diff
@@ -208,6 +211,29 @@ def add_deferred(post, options)
...
+
+        product_items_array = []
+        options[:product_details].each do |item|
+          product_items_obj = {}
+
+          product_items_obj[:id] = item[:id] if item[:id]
+          product_items_obj[:title] = item[:title] if item[:title]
+          product_items_obj[:price] = item[:price].to_i if item[:price]
+          product_items_obj[:sku] = item[:sku] if item[:sku]
+          product_items_obj[:qauntity] = item[:qauntity].to_i if item[:qauntity]
```
jcreiff, line 225: This should be `quantity` instead of `qauntity`, both here and in the tests

jcreiff (changes requested):

lib/active_merchant/billing/gateways/kushki.rb
yunnydang: thanks for catching this!

jcreiff (approved):

PR #5030, opened 2024-01-29
SumUp: Add 3DS fields

sinourain:
Description
-------------------------
[SER-798](https://spreedly.atlassian.net/browse/SER-798)

This commit enable the 3DS to retrieve the next_step object.

Unit test
-------------------------
Finished in 27.706081 seconds.
5807 tests, 78983 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
209.59 tests/s, 2850.75 assertions/s

Remote test
-------------------------
Finished in 3.182983 seconds.
1 tests, 3 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
0.31 tests/s, 0.94 assertions/s

Rubocop
-------------------------
787 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/sum_up.rb
+++ b/lib/active_merchant/billing/gateways/sum_up.rb
@@ -63,6 +63,7 @@ module ActiveMerchant #:nodoc:
         add_invoice(post, money, options)
         add_address(post, options)
         add_customer_data(post, payment, options)
+        add_3ds_data(post, options)
 
         commit('checkouts', post)
       end
@@ -123,6 +124,10 @@ module ActiveMerchant #:nodoc:
         }
       end
 
+      def add_3ds_data(post, options)
+        post[:redirect_url] = options[:redirect_url] if options[:redirect_url]
+      end
+
       def commit(action, post, method = :post)
         response = api_request(action, post.compact, method)
         succeeded = success_from(response)
@@ -157,7 +162,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(response)
-        return true if response == 204
+        return true if (response.is_a?(Hash) && response[:next_step]) || response == 204
 
         return false unless %w(PENDING EXPIRED PAID).include?(response[:status])
 
@@ -170,7 +175,7 @@ module ActiveMerchant #:nodoc:
 
       def message_from(succeeded, response)
         if succeeded
-          return 'Succeeded' if response.is_a?(Integer)
+          return 'Succeeded' if (response.is_a?(Hash) && response[:next_step]) || response == 204
 
           return response[:status]
         end
--- a/test/remote/gateways/remote_sum_up_test.rb
+++ b/test/remote/gateways/remote_sum_up_test.rb
@@ -126,9 +126,9 @@ class RemoteSumUpTest < Test::Unit::TestCase
   # In Sum Up the account can only return checkout/purchase in pending or success status,
   # to obtain a successful refund we will need an account that returns the checkout/purchase in successful status
   #
-  # For this example configure in the fixtures => :sum_up_account_for_successful_purchases
+  # For the following refund tests configure in the fixtures => :sum_up_successful_purchase
   def test_successful_refund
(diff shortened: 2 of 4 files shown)


test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -139,7 +139,7 @@ def test_successful_refund
   end
 
   def test_successful_partial_refund
-    gateway = SumUpGateway.new(fixtures(:sum_up_account_for_successful_purchases))
+    gateway = SumUpGateway.new(fixtures(:sum_up_successful_purchase))
```
gasb150, line 142: ❓ Question
I know you need a 3ds account coming from Europe. Is it possible to use the same 3ds account credentials here instead of this one?

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -126,9 +126,9 @@ def test_failed_void_invalid_checkout_id
   # In Sum Up the account can only return checkout/purchase in pending or success status,
   # to obtain a successful refund we will need an account that returns the checkout/purchase in successful status
   #
-  # For this example configure in the fixtures => :sum_up_account_for_successful_purchases
+  # For this example configure in the fixtures => :sum_up_successful_purchase
```
gasb150, line 129: 💭  Opinion:
I'm not sure if this comment is required (the fixtures method seems to be enough to inform that the sum_up_successful_purchase needs to be configured).

If you consider conserving it, maybe `test` instead of `example` could be better.

test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -168,6 +168,21 @@ def test_failed_refund_for_pending_checkout
     assert_equal 'The transaction is not refundable in its current state', response.message
   end
 
+  # In Sum Up to trigger the 3DS flow (next_step object) you need to an European account
+  #
+  # For this example configure in the fixtures => :sum_up_3ds
+  def test_trigger_3ds_flow
+    gateway = SumUpGateway.new(fixtures(:sum_up_3ds))
```
gasb150, line 175: ❗ Modify
This :sum_up_3ds entry is missing in the test/fixtures.yml file.

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -157,7 +162,7 @@ def parse(body)
       end
 
       def success_from(response)
-        return true if response == 204
+        return true if (response.is_a?(Hash) && response[:next_step]) || response == 204
```
gasb150, line 165: ❓  Question:
why it's required to validate if the response is a Hash? what are the other possible values?

gasb150 (commented):
It looks great, I left a couple of comments for your consideration.

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: Yes, this is because the refund transaction return an Integer, If we don't verify the Hash then it breaks when doing a refund

test/remote/gateways/remote_sum_up_test.rb
sinourain: No, it is because the account for `sum_up_successful_purchase` return a successful purchase, `sum_up_3ds` return a pending purchase and we can't `refund` a pending purchase.

lib/active_merchant/billing/gateways/sum_up.rb
naashton: Could this possibly lead to false positives? Would the presence of a `Hash` always mean a successful txn or are there failed transactions that return a hash with error messages?

naashton (commented):
LGTM, just one question following Gustavo's about the `success_from` method change.

sinourain:
> Could this possibly lead to false positives? Would the presence of a `Hash` always mean a successful txn or are there failed transactions that return a hash with error messages?

I think no because the second condition is the `next_step` object presence.

naashton (approved):
LGTM!

PR #5031, opened 2024-01-30
Elavon: Update Stored Credential behavior

curiousepic:
To satisfy new Elavon API requirements, including recurring_flag, approval_code for non-tokenized PMs, installment_count and _number, and situational par_value and association_token_data fields.

This also now allows Authorize transactions with a token/stored card.

Remote (Same 5 Failures on Master):
39 tests, 162 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 87.1795% passed

Unit:
53 tests, 298 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/elavon.rb
+++ b/lib/active_merchant/billing/gateways/elavon.rb
@@ -42,13 +42,7 @@ module ActiveMerchant #:nodoc:
           xml.ssl_vendor_id         @options[:ssl_vendor_id] || options[:ssl_vendor_id]
           xml.ssl_transaction_type  self.actions[:purchase]
           xml.ssl_amount            amount(money)
-
-          if payment_method.is_a?(String)
-            add_token(xml, payment_method)
-          else
-            add_creditcard(xml, payment_method)
-          end
-
+          add_token_or_card(xml, payment_method, options)
           add_invoice(xml, options)
           add_salestax(xml, options)
           add_currency(xml, money, options)
@@ -56,27 +50,26 @@ module ActiveMerchant #:nodoc:
           add_customer_email(xml, options)
           add_test_mode(xml, options)
           add_ip(xml, options)
-          add_auth_purchase_params(xml, options)
+          add_auth_purchase_params(xml, payment_method, options)
           add_level_3_fields(xml, options) if options[:level_3_data]
         end
         commit(request)
       end
 
-      def authorize(money, creditcard, options = {})
+      def authorize(money, payment_method, options = {})
         request = build_xml_request do |xml|
           xml.ssl_vendor_id         @options[:ssl_vendor_id] || options[:ssl_vendor_id]
           xml.ssl_transaction_type  self.actions[:authorize]
           xml.ssl_amount            amount(money)
-
-          add_salestax(xml, options)
+          add_token_or_card(xml, payment_method, options)
           add_invoice(xml, options)
-          add_creditcard(xml, creditcard)
+          add_salestax(xml, options)
           add_currency(xml, money, options)
           add_address(xml, options)
           add_customer_email(xml, options)
           add_test_mode(xml, options)
           add_ip(xml, options)
-          add_auth_purchase_params(xml, options)
+          add_auth_purchase_params(xml, payment_method, options)
           add_level_3_fields(xml, options) if options[:level_3_data]
         end
         commit(request)
@@ -92,7 +85,7 @@ module ActiveMerchant #:nodoc:
             add_salestax(xml, options)
             add_approval_code(xml, authorization)
             add_invoice(xml, options)
-            add_creditcard(xml, options[:credit_card])
+            add_creditcard(xml, options[:credit_card], options)
             add_currency(xml, money, options)
             add_address(xml, options)
             add_customer_email(xml, options)
@@ -139,7 +132,7 @@ module ActiveMerchant #:nodoc:
           xml.ssl_transaction_type  self.actions[:credit]
           xml.ssl_amount            amount(money)
           add_invoice(xml, options)
-          add_creditcard(xml, creditcard)
+          add_creditcard(xml, creditcard, options)
           add_currency(xml, money, options)
           add_address(xml, options)
           add_customer_email(xml, options)
@@ -152,7 +145,7 @@ module ActiveMerchant #:nodoc:
         request = build_xml_request do |xml|
           xml.ssl_vendor_id         @options[:ssl_vendor_id] || options[:ssl_vendor_id]
           xml.ssl_transaction_type  self.actions[:verify]
-          add_creditcard(xml, credit_card)
+          add_creditcard(xml, credit_card, options)
           add_address(xml, options)
           add_test_mode(xml, options)
           add_ip(xml, options)
@@ -165,7 +158,7 @@ module ActiveMerchant #:nodoc:
           xml.ssl_vendor_id         @options[:ssl_vendor_id] || options[:ssl_vendor_id]
           xml.ssl_transaction_type  self.actions[:store]
           xml.ssl_add_token 'Y'
-          add_creditcard(xml, creditcard)
+          add_creditcard(xml, creditcard, options)
           add_address(xml, options)
           add_customer_email(xml, options)
           add_test_mode(xml, options)
@@ -174,12 +167,12 @@ module ActiveMerchant #:nodoc:
         commit(request)
       end
 
-      def update(token, creditcard, options = {})
+      def update(token, payment_method, options = {})
         request = build_xml_request do |xml|
           xml.ssl_vendor_id         @options[:ssl_vendor_id] || options[:ssl_vendor_id]
           xml.ssl_transaction_type  self.actions[:update]
-          add_token(xml, token)
-          add_creditcard(xml, creditcard)
+          xml.ssl_token token
+          add_creditcard(xml, payment_method, options)
           add_address(xml, options)
           add_customer_email(xml, options)
           add_test_mode(xml, options)
@@ -200,6 +193,14 @@ module ActiveMerchant #:nodoc:
 
       private
 
+      def add_token_or_card(xml, payment_method, options)
+        if payment_method.is_a?(String) || options[:ssl_token]
+          xml.ssl_token options[:ssl_token] || payment_method
+        else
+          add_creditcard(xml, payment_method, options)
+        end
+      end
+
       def add_invoice(xml, options)
         xml.ssl_invoice_number    url_encode_truncate((options[:order_id] || options[:invoice]), 25)
         xml.ssl_description       url_encode_truncate(options[:description], 255)
@@ -213,11 +214,11 @@ module ActiveMerchant #:nodoc:
         xml.ssl_txn_id authorization.split(';').last
       end
 
-      def add_creditcard(xml, creditcard)
+      def add_creditcard(xml, creditcard, options)
         xml.ssl_card_number   creditcard.number
         xml.ssl_exp_date      expdate(creditcard)
 
-        add_verification_value(xml, creditcard) if creditcard.verification_value?
+        add_verification_value(xml, creditcard, options)
 
         xml.ssl_first_name    url_encode_truncate(creditcard.first_name, 20)
         xml.ssl_last_name     url_encode_truncate(creditcard.last_name, 30)
@@ -230,12 +231,12 @@ module ActiveMerchant #:nodoc:
         xml.ssl_transaction_currency currency
       end
 
-      def add_token(xml, token)
-        xml.ssl_token token
-      end
+      def add_verification_value(xml, credit_card, options)
+        return unless credit_card.verification_value?
+        # Don't add cvv if this is a non-initial stored credential transaction
+        return if options[:stored_credential] && !options[:stored_credential][:initial_transaction]
 
-      def add_verification_value(xml, creditcard)
-        xml.ssl_cvv2cvc2            creditcard.verification_value
+        xml.ssl_cvv2cvc2            credit_card.verification_value
         xml.ssl_cvv2cvc2_indicator  1
       end
 
@@ -294,16 +295,16 @@ module ActiveMerchant #:nodoc:
       end
 
       # add_recurring_token is a field that can be sent in to obtain a token from Elavon for use with their tokenization program
-      def add_auth_purchase_params(xml, options)
+      def add_auth_purchase_params(xml, payment_method, options)
         xml.ssl_dynamic_dba                     options[:dba] if options.has_key?(:dba)
         xml.ssl_merchant_initiated_unscheduled  merchant_initiated_unscheduled(options) if merchant_initiated_unscheduled(options)
         xml.ssl_add_token                       options[:add_recurring_token] if options.has_key?(:add_recurring_token)
-        xml.ssl_token                           options[:ssl_token] if options[:ssl_token]
         xml.ssl_customer_code                   options[:customer] if options.has_key?(:customer)
         xml.ssl_customer_number                 options[:customer_number] if options.has_key?(:customer_number)
-        xml.ssl_entry_mode                      entry_mode(options) if entry_mode(options)
+        xml.ssl_entry_mode                      add_entry_mode(payment_method, options) if add_entry_mode(payment_method, options)
         add_custom_fields(xml, options) if options[:custom_fields]
-        add_stored_credential(xml, options) if options[:stored_credential]
+        add_stored_credential(xml, payment_method, options) if options[:stored_credential]
+        add_installment_fields(xml, options) if options.dig(:stored_credential, :reason_type) == 'installment'
       end
 
       def add_custom_fields(xml, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -46,7 +46,7 @@ def purchase(money, payment_method, options = {})
           if payment_method.is_a?(String)
             add_token(xml, payment_method)
           else
-            add_creditcard(xml, payment_method)
+            add_creditcard(xml, payment_method) unless options[:ssl_token]
```
curiousepic, line 49: Elavon indicated that card number should not be added in addition to a token.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -56,27 +56,32 @@ def purchase(money, payment_method, options = {})
...
           xml.ssl_vendor_id         @options[:ssl_vendor_id] || options[:ssl_vendor_id]
           xml.ssl_transaction_type  self.actions[:authorize]
           xml.ssl_amount            amount(money)
 
-          add_salestax(xml, options)
+          if payment_method.is_a?(String)
+            add_token(xml, payment_method)
+          else
+            add_creditcard(xml, payment_method) unless options[:ssl_token]
+          end
```
curiousepic, line 75: Adds support for Authorize transctions with a token. I'm not sure why this wasn't added originally.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
+        xml.ssl_payment_number options[:payment_number]
+        xml.ssl_payment_count options[:installments]
       end
 
-      def entry_mode(options)
+      def add_entry_mode(payment_method, options)
         return options[:entry_mode] if options[:entry_mode]
-        return 12 if options[:stored_credential]
+        return if payment_method.is_a?(String)
+        return 12 if options.dig(:stored_credential, :reason_type) == 'unscheduled'
```
curiousepic, line 399: Elavon indicated that this should only be 12 for unscheduled reason types.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
         network_transaction_id = options.dig(:stored_credential, :network_transaction_id)
         case
-        when network_transaction_id.nil?
-          return
         when network_transaction_id.to_s.include?('|')
           oar_data, ps2000_data = options[:stored_credential][:network_transaction_id].split('|')
-          xml.ssl_oar_data oar_data unless oar_data.nil? || oar_data.empty?
-          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty?
+          xml.ssl_oar_data oar_data unless oar_data.nil? || oar_data.empty? || payment_method.is_a?(String)
+          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty? || payment_method.is_a?(String)
```
curiousepic, line 368: This prevents this data from being added for tokens, if it is also present.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
         when network_transaction_id.to_s.length > 22
-          xml.ssl_oar_data options.dig(:stored_credential, :network_transaction_id)
-        else
-          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
+          xml.ssl_oar_data network_transaction_id
+        else !network_transaction_id.blank?
+          xml.ssl_ps2000_data network_transaction_id
         end
+        xml.ssl_recurring_flag recurring_flag(options) if recurring_flag(options)
```
curiousepic, line 374: Primary new requirement for recurring txns.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
+        xml.ssl_approval_code options[:approval_code] if options[:approval_code] && !payment_method.is_a?(String)
+        xml.ssl_par_value options[:par_value] if options[:par_value]
+        xml.ssl_association_token_data options[:association_token_data] if options[:association_token_data]
```
curiousepic, line 377: Two new values that may be present on an initial transaction that should be passed in subsequent COF transactions.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
+      end
+
+      def recurring_flag(options)
+        return unless reason = options.dig(:stored_credential, :reason_type)
+        return 1 if reason == 'recurring'
+        return 2 if reason == 'installment'
```
curiousepic, line 383: This is the primary new requirement for recurring txns.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
       end
 
       def merchant_initiated_unscheduled(options)
         return options[:merchant_initiated_unscheduled] if options[:merchant_initiated_unscheduled]
-        return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled' || options.dig(:stored_credential, :reason_type) == 'recurring'
+        return 'Y' if options.dig(:stored_credential, :initiator) == 'merchant' && options.dig(:stored_credential, :reason_type) == 'unscheduled'
```
curiousepic, line 388: Should only be present for reason type unscheduled, not recurring.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
+        return unless options[:stored_credential]
         network_transaction_id = options.dig(:stored_credential, :network_transaction_id)
         case
-        when network_transaction_id.nil?
-          return
         when network_transaction_id.to_s.include?('|')
           oar_data, ps2000_data = options[:stored_credential][:network_transaction_id].split('|')
-          xml.ssl_oar_data oar_data unless oar_data.nil? || oar_data.empty?
-          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty?
+          xml.ssl_oar_data oar_data unless oar_data.nil? || oar_data.empty? || payment_method.is_a?(String)
```
almalee24, line 367: Could this be updated to `oar_data.blank?` instead of having `oar_data.nil? || oar_data.empty?` ?

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
         when network_transaction_id.to_s.length > 22
-          xml.ssl_oar_data options.dig(:stored_credential, :network_transaction_id)
-        else
-          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
+          xml.ssl_oar_data network_transaction_id
+        else !network_transaction_id.blank?
+          xml.ssl_ps2000_data network_transaction_id
         end
+        xml.ssl_recurring_flag recurring_flag(options) if recurring_flag(options)
+        xml.ssl_approval_code options[:approval_code] if options[:approval_code] && !payment_method.is_a?(String)
```
almalee24, line 375: Should this also check for ` options[:ssl_token]` since it's already checking for `payment_method.is_a?(String)`?

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
-          xml.ssl_oar_data oar_data unless oar_data.nil? || oar_data.empty?
-          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty?
+          xml.ssl_oar_data oar_data unless oar_data.nil? || oar_data.empty? || payment_method.is_a?(String)
+          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty? || payment_method.is_a?(String)
         when network_transaction_id.to_s.length > 22
-          xml.ssl_oar_data options.dig(:stored_credential, :network_transaction_id)
-        else
-          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
+          xml.ssl_oar_data network_transaction_id
+        else !network_transaction_id.blank?
```
almalee24, line 371: Should this be `when` instead of `else`?

curiousepic: Do you mean `when network_transaction_id.blank?` I assume?

lib/active_merchant/billing/gateways/elavon.rb
curiousepic: I'm not quite sure what you're suggesting, could you elaborate?

To be clear, `approval_code` is not needed if the PM is a token.

lib/active_merchant/billing/gateways/elavon.rb
curiousepic: I think you are right, good catch.

almalee24:
If you haven't already please don't forget to run `bundle exec rubocop`

curiousepic:
Updated with Rubocop fixes.

lib/active_merchant/billing/gateways/elavon.rb
almalee24: If a customer sends `options[:ssl_token]` then we will [send ssl_token and not add_creditcard](https://github.com/activemerchant/active_merchant/pull/5031/files#diff-9974d3a098a1e363ff26162f6bed4c26faf5fe0c79acc8c5ac18832752676e15R71). So in this case wouldn't the PM be a token then?

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
-          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty?
-        when network_transaction_id.to_s.length > 22
-          xml.ssl_oar_data options.dig(:stored_credential, :network_transaction_id)
-        else
-          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
+          xml.ssl_oar_data oar_data unless oar_data.blank? || payment_method.is_a?(String)
+          xml.ssl_ps2000_data ps2000_data unless ps2000_data.blank? || payment_method.is_a?(String)
+        elsif network_transaction_id.to_s.length > 22
+          xml.ssl_oar_data network_transaction_id
+        elsif !network_transaction_id.blank?
```
almalee24, line 371: This can be `network_transaction_id.present?`

lib/active_merchant/billing/gateways/elavon.rb
curiousepic: Yes... good catch, I think I second guessed myself. That brings up a good point, I should add tests for the combination of manual token and stored credential fields.

curiousepic:
Added tests for the combination of stored cred fields and using a manual token instead of a stored card authorization string.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,45 @@ def add_line_items(xml, level_3_data)
...
-          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
+          xml.ssl_oar_data oar_data unless oar_data.blank? || payment_method.is_a?(String)
+          xml.ssl_ps2000_data ps2000_data unless ps2000_data.blank? || payment_method.is_a?(String)
+        elsif network_transaction_id.to_s.length > 22
+          xml.ssl_oar_data network_transaction_id
+        elsif network_transaction_id.present?
+          xml.ssl_ps2000_data network_transaction_id
         end
+        xml.ssl_recurring_flag recurring_flag(options) if recurring_flag(options)
+        xml.ssl_approval_code options[:approval_code] if options[:approval_code] && !payment_method.is_a?(String) && !options[:ssl_token]
```
almalee24, line 375: Could we make this into a variable so it can also be used above? Where we check for `payment_method.is_a?(String)` in `ssl_oar_data` and `ssl_ps2000_data`. Maybe the the new variable could be called `used_payment_token`

curiousepic: Good suggestion, I've done some refactoring of this method to clean it up in general which should help.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +358,48 @@ def add_line_items(xml, level_3_data)
...
-          xml.ssl_ps2000_data ps2000_data unless ps2000_data.nil? || ps2000_data.empty?
-        when network_transaction_id.to_s.length > 22
-          xml.ssl_oar_data options.dig(:stored_credential, :network_transaction_id)
-        else
-          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
+        xml.ssl_recurring_flag recurring_flag(options) if recurring_flag(options)
+        xml.ssl_par_value options[:par_value] if options[:par_value]
+        xml.ssl_association_token_data options[:association_token_data] if options[:association_token_data]
+
+        if !payment_method.is_a?(String) && !options[:ssl_token].present?
```
almalee24, line 369: Could this to be `unless` instead of `if`

almalee24 (approved):
This looks great! Thanks for all the hard work you put in!

curiousepic:
@almalee24 Made some changes after feedback from Elavon - notably not sending cvv on non-initial stored cred txns, and ensuring a blank ps2000 data field was not passed in some situations, as well as being able to invoke real respones for the new par_value and association_token_data fields.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -234,8 +239,11 @@ def add_token(xml, token)
         xml.ssl_token token
```
almalee24, line 239: We probably need to update this method to either pass down options to grab token from there if payment is a credit card but has options[:ssl_token] or to return unless token.is_a?(String)

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +361,48 @@ def add_line_items(xml, level_3_data)
...
-        else
-          xml.ssl_ps2000_data options.dig(:stored_credential, :network_transaction_id)
+        xml.ssl_recurring_flag recurring_flag(options) if recurring_flag(options)
+        xml.ssl_par_value options[:par_value] if options[:par_value]
+        xml.ssl_association_token_data options[:association_token_data] if options[:association_token_data]
+
+        unless payment_method.is_a?(String) || options[:ssl_token].present?
+          xml.ssl_approval_code options[:approval_code] if options[:approval_code]
+          if network_transaction_id.to_s.include?('|')
+            oar_data, ps2000_data = options[:stored_credential][:network_transaction_id].split('|')
```
almalee24, line 375: could this just be network_transaction_id send you already set that variable

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -234,8 +239,11 @@ def add_token(xml, token)
         xml.ssl_token token
       end
 
-      def add_verification_value(xml, creditcard)
-        xml.ssl_cvv2cvc2            creditcard.verification_value
+      def add_verification_value(xml, credit_card, options)
+        return unless credit_card.verification_value?
+        return if options[:stored_credential] && options[:stored_credential][:initial_transaction] != 'Y'
```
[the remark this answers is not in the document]
curiousepic: I think we should rely on a strict positive value than a specific type of negative value to catch cases where it might be `N` for example.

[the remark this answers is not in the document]
curiousepic: I rewrote this condition a number of times from other options that caused incorrect behavior for one of many specific sitautions...

lib/active_merchant/billing/gateways/elavon.rb
curiousepic: Good catch... I thought I'd added enough test variations to cover that situation 😓

Oh wait, I think it didn't catch them because we have another place where we're adding `ssl_token` in `add_auth_purchase_params` 😅

lib/active_merchant/billing/gateways/elavon.rb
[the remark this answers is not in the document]
almalee24: The only thing is that `initial_transaction` is either true or false not Y or F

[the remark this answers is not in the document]
curiousepic: 🤦  Got it, thank you.

curiousepic:
Refactored all instances of adding card or token into a single helper method (update action needs both so adds them manually). Would appreciate another look at the diff.

lib/active_merchant/billing/gateways/elavon.rb
```diff
@@ -352,30 +353,48 @@ def add_line_items(xml, level_3_data)
         }
       end
 
-      def add_stored_credential(xml, options)
+      def add_stored_credential(xml, payment_method, options)
+        return unless options[:stored_credential]
```
almalee24, line 357: This is not necessary since you already did` if options[:stored_credential]` when calling the method

almalee24:
Looks good to me!

lib/active_merchant/billing/gateways/elavon.rb
curiousepic: True, though I'd lean toward overeager guard clauses than less; if it gets used in the future it's already there.

PR #5032, opened 2024-01-31
GlobalCollect: Remove decrypted payment data

almalee24:
Remove support for GooglePay decrypted payment data since they only support encrypted payment data.

Unit:
40 tests, 197 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


DustinHaefele (approved):
Looks good!

PR #5033, opened 2024-01-31
Quickbooks: Update scrubbing logic

almalee24:
Update quickbooks scrubbing to include card lengths from 13-19 digits. Update quickbooks scrubbing to include cvc codes from 3-4 digits long.

Remote:
19 tests, 48 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
23 tests, 120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -128,8 +128,8 @@ module ActiveMerchant #:nodoc:
           gsub(%r((oauth_nonce=\")\w+), '\1[FILTERED]').
           gsub(%r((oauth_signature=\")[a-zA-Z%0-9]+), '\1[FILTERED]').
           gsub(%r((oauth_token=\")\w+), '\1[FILTERED]').
-          gsub(%r((number\D+)\d{16}), '\1[FILTERED]').
-          gsub(%r((cvc\D+)\d{3}), '\1[FILTERED]').
+          gsub(%r((number\D+)\d+), '\1[FILTERED]').
+          gsub(%r((cvc\D+)\d+), '\1[FILTERED]').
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -128,8 +128,8 @@ def scrub(transcript)
           gsub(%r((oauth_nonce=\")\w+), '\1[FILTERED]').
           gsub(%r((oauth_signature=\")[a-zA-Z%0-9]+), '\1[FILTERED]').
           gsub(%r((oauth_token=\")\w+), '\1[FILTERED]').
-          gsub(%r((number\D+)\d{16}), '\1[FILTERED]').
-          gsub(%r((cvc\D+)\d{3}), '\1[FILTERED]').
+          gsub(%r((number\D+)\d{13,19}), '\1[FILTERED]').
+          gsub(%r((cvc\D+)\d{3,4}), '\1[FILTERED]').
           gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
```
aenand, line 133: Why use a specified range length of characters instead of `d+`?

DustinHaefele: Probably because that is what I wrote in the ticket. These should match specified ranges for these values currently, but `\d+` would keep us safe going forward, that's a good call.

almalee24: I've updated it to be` \d+`

DustinHaefele (approved):
Looks great! 🎉 Can't wait to have less alerts in our channel!

PR #5040, opened 2024-02-15
Quickbooks: Update scrub method

almalee24:
Update scrub method to accomdate all card types
including AMEX.

### DIFF
--- a/lib/active_merchant/billing/gateways/quickbooks.rb
+++ b/lib/active_merchant/billing/gateways/quickbooks.rb
@@ -128,8 +128,8 @@ module ActiveMerchant #:nodoc:
           gsub(%r((oauth_nonce=\")\w+), '\1[FILTERED]').
           gsub(%r((oauth_signature=\")[a-zA-Z%0-9]+), '\1[FILTERED]').
           gsub(%r((oauth_token=\")\w+), '\1[FILTERED]').
-          gsub(%r((number\D+)\d{16}), '\1[FILTERED]').
-          gsub(%r((cvc\D+)\d{3}), '\1[FILTERED]').
+          gsub(%r((number\\\":\\\")\d+), '\1[FILTERED]').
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/quickbooks.rb
```diff
@@ -128,8 +128,8 @@ def scrub(transcript)
           gsub(%r((oauth_nonce=\")\w+), '\1[FILTERED]').
           gsub(%r((oauth_signature=\")[a-zA-Z%0-9]+), '\1[FILTERED]').
           gsub(%r((oauth_token=\")\w+), '\1[FILTERED]').
-          gsub(%r((number\D+)\d{16}), '\1[FILTERED]').
-          gsub(%r((cvc\D+)\d{3}), '\1[FILTERED]').
+          gsub(%r((number\\\":\\\")\d+), '\1[FILTERED]').
+          gsub(%r((cvc\\\":\\\")\d+), '\1[FILTERED]').
```
DustinHaefele, line 132: Hmm, why did we change this to be so specific? Was the `\D+` that would include any non digit characters causing issues?

DustinHaefele: I'm just a little concerned that if there is white space or something that it will stop scubbing the digits.

almalee24: In remote test this is exactly how quickbooks returns it and it scrubs properly. The reason I change it was because `cvc\D+` was scrubbing other random ids in the response.

almalee24: Having it like `(cvc\D+)\d+` was scrubbing single digits

DustinHaefele: Ah that makes sense. I'm good with this!

DustinHaefele (approved):

PR #5041, opened 2024-02-16
Allow UCAFInd to be overridden in Orbital gateway

priceline-rosenfeld:
UCAFInd is always set to '4' in the Orbital gateway currently, but the documentation allows for other values too. I noticed @cgriego also created PR #3369 to support other values than '4' among other changes.

This third PR is the latest of the PRs I prepared for the Orbital gateway at the moment. If we can get those 3 PRs approved, we would be able to switch from our fork and use mainstream activemerchant.

Would you mind reviewing those PRs?

If you think we shouldn't default to '4' and break backwards compatibility, just let me know and I'll gladly update the PR.


cgriego:
What needs to happen so we can get this change in?

PR #5042, opened 2024-02-16
RedsysRest: Add support for 3DS

almalee24:
Unit:
23 tests, 101 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
18 tests, 38 assertions, 8 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 55.5556% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -952,6 +952,7 @@
 * PayU Latam: Improve error response [esmitperez] #3717
 * Vantiv: Vantiv Express - CardPresentCode, PaymentType, SubmissionType, DuplicateCheckDisableFlag [esmitperez] #3730,#3731
 * Cybersource: Ensure issueradditionaldata comes before partnerSolutionId [britth] #3733
+* RedsysRest: Add support for 3DS [almalee24] #5042
 
 == Version 1.111.0
 * Fat Zebra: standardized 3DS fields and card on file extra data for Visa scheme rules [montdidier] #3409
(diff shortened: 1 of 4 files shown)


almalee24:
> Why are so many remote tests failing?

Looks like it could be use needing a new test card but I can't find a non 3DS one on their website

aenand:
> > Why are so many remote tests failing?
> 
> Looks like it could be us needing a new test card but I can't find a non 3DS one on their website. I looks like maybe the test card that we're using is triggering a certain response but I'm not sure.

Hmmmm I think when I added these, if `DS_MERCHANT_DIRECTPAYMENT` was set to `true` it skipped 3DS. I wonder what the new test card value(s) are?

aenand:
I see an AMEX card in their test env, i wonder if that one trigger 3DS?

almalee24:
> I see an AMEX card in their test env, i wonder if that one trigger 3DS?

I'm not sure what AMEX card you're talking about

almalee24:
> > > Why are so many remote tests failing?
> > 
> > 
> > Looks like it could be us needing a new test card but I can't find a non 3DS one on their website. I looks like maybe the test card that we're using is triggering a certain response but I'm not sure.
> 
> Hmmmm I think when I added these, if `DS_MERCHANT_DIRECTPAYMENT` was set to `true` it skipped 3DS. I wonder what the new test card value(s) are?

I looks like DS_MERCHANT_DIRECTPAYMENT is not working anymore. I'm not sure, I can't see anything in their documentation. Or maybe they're requiring 3DS for all transactions now?

bradbroge (approved):
Verified that the failing remote tests are failing with 
```
   @message="Requires SCA authentication"
```
So it is still expecting a 3DS transaction, and I didn't see an easy way to resolve that in the Redsys documentation.

PR #5043, opened 2024-02-20
Braintree: Surface the paypal_details in response object

yunnydang:
This returns the paypal_details in the response object on paypal transaction so that we can surface the payer_id and payer_email fields for Braintree. 

Note: The remote test written requires a separate fixtures set.

Local:
5813 tests, 78910 assertions, 2 failures, 26 errors, 0 pendings, 0 omissions, 0 notifications
99.5183% passed

Unit:
100% passed

Remote:
116 tests, 607 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
98.2759% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -117,6 +117,7 @@
 * Rapyd: Adding 500 errors handling [Heavyblade] #5029
 * SumUp: Add 3DS fields [sinourain] #5030
 * Cecabank: Enable network_transaction_id as GSF [javierpedrozaing] #5034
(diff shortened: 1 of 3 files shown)


jcreiff (approved):
This looks good! I poked around on a local branch and confirmed that the `paypal_details` object is present on the transaction even when other payment types are used, so there shouldn't be a risk of NoMethodError when trying to dig for the `payer_id` and `payer_email`

PR #5045, opened 2024-02-22
Shift4: Add support for avs_result

rachelkirk:
CER-1128

This adds support for avs_result in the gateway's response. The standard ActiveMerchant AVS mapping is used, but varies slightly from Shift4's AVS message language in their docs. This should be taken into consideration during review. No remote test was added because the avs hash is not returned in sandbox transactions. It is returned consistently in the same format in the `transaction` hash in production transcripts.

Unit:
28 tests, 176 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
29 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Local:
5814 tests, 79016 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 99.9656% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4.rb
+++ b/lib/active_merchant/billing/gateways/shift4.rb
@@ -247,6 +247,7 @@ module ActiveMerchant #:nodoc:
           message_from(action, response),
           response,
           authorization: authorization_from(action, response),
+          avs_result: avs_result_from(response),
           test: test?,
           error_code: error_code_from(action, response)
         )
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/shift4.rb
```diff
@@ -247,6 +247,7 @@ def commit(action, parameters, option)
           message_from(action, response),
           response,
           authorization: authorization_from(action, response),
+          avs_result: AVSResult.new(code: response['result'].first&.dig('transaction', 'avs', 'result')),
```
jcreiff, line 250: Small suggestion: extract out the AVS parsing logic into an `avs_from` method and guard against possibility of no AVS result returned?

rachelkirk: Done!

jcreiff (approved):
LGTM!

PR #5046, opened 2024-02-22
Update Nexi Xpay to use 3DS 3steps API

sinourain:
This commit contains an update for Nexi Xpay to support 3steps API instead of 2steps API endpoints in order to complete 3DS transactions Documents for reference:

https://developer.nexi.it/en/modalita-di-integrazione/server-to-server/pagamento-3-steps

### DIFF
--- a/lib/active_merchant/billing/gateways/xpay.rb
+++ b/lib/active_merchant/billing/gateways/xpay.rb
@@ -14,14 +14,17 @@ module ActiveMerchant #:nodoc:
       self.supported_cardtypes = %i[visa master maestro american_express jcb]
 
       ENDPOINTS_MAPPING = {
-        purchase: 'orders/2steps/payment',
-        authorize: 'orders/2steps/payment',
-        preauth: 'orders/2steps/init',
+        validation: 'orders/3steps/validation',
+        purchase: 'orders/3steps/payment',
+        authorize: 'orders/3steps/payment',
+        preauth: 'orders/3steps/init',
         capture: 'operations/%s/captures',
         verify: 'orders/card_verification',
         void: 'operations/%s/cancels',
         refund: 'operations/%s/refunds'
       }
+
+      SUCCESS_MESSAGES = %w(PENDING AUTHORIZED THREEDS_VALIDATED EXECUTED).freeze
 
       def initialize(options = {})
         requires!(options, :api_key)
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -81,6 +82,19 @@ def scrub(transcript)
...
 
+      def validation(options = {})
+        post = {}
+        add_3ds_validation_params(post, options)
+        commit(:validation, post, options)
+      end
+
+      def complete_transaction(action, amount, payment_method, options = {})
+        MultiResponse.run do |r|
+          r.process { validation(options) }
```
Heavyblade, line 93: 💬 Comment:

Unless you know that server is always going to answer with a 200 on this end-point maybe a check here before calling the final step is a good idea.

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -81,6 +82,19 @@ def scrub(transcript)
...
+          r.process { add_transaction_params_commit(action, amount, {}, payment_method, options.merge!(validation: r.params)) }
```
Heavyblade, line 94: 💬 Comment:

Maybe the name `add_transaction_params_commit` is not entirely accurate because this method does way more than just that

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -162,9 +176,17 @@ def add_exemptions(post, options)
         post[:exemptions] = options[:exemptions] || 'NO_PREFERENCE'
       end
 
-      def add_3ds_params(post, options)
-        post[:threeDSAuthData] = { threeDSAuthResponse: options[:three_ds_auth_response] }.compact
-        post[:operationId] = options[:operation_id] if options[:operation_id]
+      def add_3ds_params(post, validation)
```
Heavyblade, line 179: 💬 Comment / Style:

Not a big deal just a little more Dryer:

```ruby
def add_3ds_params(post, validation)
  post[:threeDSAuthData] = {
    authenticationValue: validation['threeDSAuthResult']['authenticationValue'],
    eci: validation['threeDSAuthResult']['eci'],
    xid: validation['threeDSAuthResult']['xid']
  }
  post[:operationId] = validation['operation']['operationId']
end
```

Heavyblade (commented):
Done with the review @sinourain / @jherreraa couple of comments:

🔴 Warning 🔴 => Remember to fix the message method to handle the error case

lib/active_merchant/billing/gateways/xpay.rb
sinourain: Done!

lib/active_merchant/billing/gateways/xpay.rb
sinourain: When the transaction is not successfully the next process is not executed because the following unless statement: https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/response.rb#L62

Then the response retrieve a fail response in the validation part.

PR #5047, opened 2024-02-23
Stripe Payment Intents: Update expand latest_attempt on create_setup

DustinHaefele:
Stripe Payment Intents: Update expand latest_attempt on create_setup_intent

This expansion allows a user to get the avs and cvc check results from the create_setup_intent call

[ECS-3375](https://spreedly.atlassian.net/browse/ECS-3375)

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -696,7 +696,6 @@ module ActiveMerchant #:nodoc:
       end
 
       def commit(method, url, parameters = nil, options = {})
-
         add_expand_parameters(parameters, options) if parameters
 
         return Response.new(false, 'Invalid API Key provided') unless key_valid?(options)
@@ -706,7 +705,7 @@ module ActiveMerchant #:nodoc:
         success = success_from(response, options)
 
         card_checks = card_from_response(response)
-        avs_code = AVS_CODE_TRANSLATOR["line1: #{card_checks['address_line1_check']}, zip: #{card_checks['address_zip_check'] || card_checks['address_postal_code_check']}"]
+        avs_code = avs_code_from(card_checks)
         cvc_code = CVC_CODE_TRANSLATOR[card_checks['cvc_check']]
         Response.new(
           success,
@@ -789,7 +788,15 @@ module ActiveMerchant #:nodoc:
 
       def card_from_response(response)
         # StripePI puts the AVS and CVC check significantly deeper into the response object
-        response['card'] || response['active_card'] || response['source'] || response.dig('charges', 'data', 0, 'payment_method_details', 'card', 'checks') || {}
+        response['card'] || response['active_card'] || response['source'] ||
+          response.dig('charges', 'data', 0, 'payment_method_details', 'card', 'checks') ||
+          response.dig('latest_attempt', 'payment_method_details', 'card', 'checks') {}
+      end
+
+      def avs_code_from(card_checks)
+        return 'I' if card_checks['address_postal_code_check'].nil? && card_checks['address_line1_check'].nil? && card_checks['address_zip_check'].nil?
+
+        AVS_CODE_TRANSLATOR["line1: #{card_checks['address_line1_check']}, zip: #{card_checks['address_zip_check'] || card_checks['address_postal_code_check']}"]
       end
 
       def emv_authorization_from_response(response)
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -579,7 +579,7 @@ module ActiveMerchant #:nodoc:
 
       def add_source_owner(post, creditcard, options)
         post[:owner] = {}
-        post[:owner][:name] = creditcard.name if creditcard.name
+        post[:owner][:name] = creditcard.name if creditcard.respond_to?(:name) && creditcard.name
         post[:owner][:email] = options[:email] if options[:email]
 
         if address = options[:billing_address] || options[:address]
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -705,7 +705,7 @@ module ActiveMerchant #:nodoc:
         success = success_from(response, options)
 
         card_checks = card_from_response(response)
-        avs_code = avs_code_from(card_checks)
+        avs_code = AVS_CODE_TRANSLATOR["line1: #{card_checks['address_line1_check']}, zip: #{card_checks['address_zip_check'] || card_checks['address_postal_code_check']}"]
         cvc_code = CVC_CODE_TRANSLATOR[card_checks['cvc_check']]
         Response.new(
           success,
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -789,7 +788,15 @@ def quickchip_payment?(payment)
 
       def card_from_response(response)
         # StripePI puts the AVS and CVC check significantly deeper into the response object
-        response['card'] || response['active_card'] || response['source'] || response.dig('charges', 'data', 0, 'payment_method_details', 'card', 'checks') || {}
+        response['card'] || response['active_card'] || response['source'] ||
+          response.dig('charges', 'data', 0, 'payment_method_details', 'card', 'checks') ||
+          response.dig('latest_attempt', 'payment_method_details', 'card', 'checks') {}
```
aenand, line 793: Stripe nests this data in 5 spots. Insanity 🤯

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -145,6 +145,7 @@ def create_setup_intent(payment_method, options = {})
             post[:on_behalf_of] = options[:on_behalf_of] if options[:on_behalf_of]
             post[:usage] = options[:usage] if %w(on_session off_session).include?(options[:usage])
             post[:description] = options[:description] if options[:description]
+            post[:expand] = ['latest_attempt']
```
aenand, line 148: What is this parameter?

aenand (commented):
Nice work! One question

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: it expands certain parameters in Stripe we use it a lot but I just want to make sure it won't change any response fields on our side

almalee24: > it expands certain parameters in Stripe to get more information. We use it a lot but I just want to make sure it won't change any response fields on our side other than what we are trying to accomplish here

lib/active_merchant/billing/gateways/stripe.rb
bradbroge, line 793:
```suggestion
          response.dig('latest_attempt', 'payment_method_details', 'card', 'checks') || {}
```

bradbroge: some of those spots may be historical, but still.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
aenand: Where else do we use it @almalee24? Would we want to expand this for all requests, not just setup intents?

DustinHaefele: In stripe's response they generally send something like this. (Along with all the other parameters) `{"latest_attempt": "setatt_1Omn1KAWOtgoysoge1eoODtu"}` . The `expand` param pulls all the data from the latest_attempt object into the response so that you don't have to make a separate call to stripe to retrieve it.

lib/active_merchant/billing/gateways/stripe.rb
DustinHaefele: yeah it is wild.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: I looks like at least for setup_intent `latest_attempt` only returns a string but with this expand it not returns a object. The object is what we expect on our side because we use that to grab NTID and other stuff

almalee24: I think this is a safe and good addition to setup intent. I'm surprised that it currently just returns a string on setup intent but I wonder if it's only sometimes or all the time

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -789,7 +788,15 @@ def quickchip_payment?(payment)
...
       def card_from_response(response)
         # StripePI puts the AVS and CVC check significantly deeper into the response object
-        response['card'] || response['active_card'] || response['source'] || response.dig('charges', 'data', 0, 'payment_method_details', 'card', 'checks') || {}
+        response['card'] || response['active_card'] || response['source'] ||
+          response.dig('charges', 'data', 0, 'payment_method_details', 'card', 'checks') ||
+          response.dig('latest_attempt', 'payment_method_details', 'card', 'checks') || {}
+      end
+
+      def avs_code_from(card_checks)
+        return 'I' if card_checks['address_postal_code_check'].nil? && card_checks['address_line1_check'].nil? && card_checks['address_zip_check'].nil?
```
almalee24, line 797: Would 'I' be considered the a default value for Stripe or card brands?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
almalee24: Also `latest_attempt` is just a Setup Intent parameter so it should only be added here. Since we are do this change it would be nice to cleaning up Core a bit for the StripePI NTID since after this change we will be returning it on all SetupIntent transactions now

aenand (approved):
LGTM

almalee24 (approved):

lib/active_merchant/billing/gateways/stripe.rb
rachelkirk: It looks like 'I' is what is returned if the address line1 or zip are both unchecked. So I guess we can assume `nil` means unchecked or they are just not returning the results of the check.

lib/active_merchant/billing/gateways/stripe.rb
rachelkirk: Geez Louise

lib/active_merchant/billing/gateways/stripe.rb
DustinHaefele: That's a good question. the "I", just means that it wasn't checked, but I'm not sure who defined that code. I was just talking to @rachelkirk about this and I think I am going to remove these nil checks and only set a code if stripe is specifically sending us a code.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
DustinHaefele: `latest_attempt` is only a field for the `setup_intent` object so we should be good with having it only there.

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -579,7 +579,7 @@ def add_shipping_address(post, payment, options = {})
 
       def add_source_owner(post, creditcard, options)
         post[:owner] = {}
-        post[:owner][:name] = creditcard.name if creditcard.name
+        post[:owner][:name] = creditcard&.name if creditcard&.name
```
almalee24, line 582: The issue for this is that creditcard can be a string so this would still cause an error

rachelkirk: Does the &. operator not guard against it since everything in Ruby is an object? Or am I getting that wrong?

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -579,7 +579,7 @@ def add_shipping_address(post, payment, options = {})
 
       def add_source_owner(post, creditcard, options)
         post[:owner] = {}
-        post[:owner][:name] = creditcard.name if creditcard.name
+        post[:owner][:name] = creditcard.name if creditcard.respond_to?(:name) && creditcard.name
         post[:owner][:email] = options[:email] if options[:email]
```
rachelkirk, line 583: Noice

rachelkirk (approved):
The changes look good to me! Thanks for answering my questions about `nil`

lib/active_merchant/billing/gateways/stripe.rb
DustinHaefele: My bad, misunderstood that the safe access `&` only works on objects. This is fixed in the latest push!

PR #5048, opened 2024-02-26
StripePI: Update authorization_from

almalee24:
If the transaction fails with 'Received unexpected 3DS authentication response, but a 3DS initiation flag was not included in the request.' then authorization_from should return response['id']

Unit
59 tests, 308 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote
95 tests, 451 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


DustinHaefele (approved):
Looks great!

PR #5050, opened 2024-02-27
Braintree Blue: Refactor and add payment details to failed transactions

yunnydang:
This refactors a slight portion of the transaction hash regarding the payment details object that is being returned depending on which payment method is used. Also returns the these objects on failing transactions and added small fixes to previously failing tests. Also fixes some rubocop related issues.

Local:
5816 tests, 78913 assertions, 2 failures, 30 errors, 0 pendings, 0 omissions, 0 notifications
99.4498% passed

Unit:
100% passed

Remote:
120 tests, 634 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -499,6 +499,54 @@ module ActiveMerchant #:nodoc:
         result.transaction&.additional_processor_response
       end
 
+      def payment_instrument_type(result)
+        result&.payment_instrument_type
+      end
+
+      def credit_card_details(result)
+        if result
+          {
+            'masked_number'       => result.credit_card_details&.masked_number,
+            'bin'                 => result.credit_card_details&.bin,
+            'last_4'              => result.credit_card_details&.last_4,
+            'card_type'           => result.credit_card_details&.card_type,
+            'token'               => result.credit_card_details&.token,
+            'debit'               => result.credit_card_details&.debit,
+            'prepaid'             => result.credit_card_details&.prepaid,
+            'issuing_bank'        => result.credit_card_details&.issuing_bank
+          }
+        end
+      end
+
+      def network_token_details(result)
+        if result
+          {
+            'debit'               => result.network_token_details&.debit,
+            'prepaid'             => result.network_token_details&.prepaid,
+            'issuing_bank'        => result.network_token_details&.issuing_bank
+          }
+        end
+      end
+
+      def google_pay_details(result)
+        if result
+          {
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -499,6 +499,50 @@ def additional_processor_response_from_result(result)
         result.transaction&.additional_processor_response
       end
 
+      def credit_card_details(result)
+        if result && result.credit_card_details
+          credit_card_details = {
```
jcreiff, line 504: Just a thought - it seems to me that all that is being returned from this method is the hash.

Is there any need to set a variable `credit_card_details` equal to the output of the hash? I think you get the same result without `credit_card_details = `

jcreiff (approved):
This looks good to me, although I left one comment for your consideration.

The remote and unit tests all run successfully for me, so it appears this minor refactor should be compatible with the way the gateway has been working to this point. We'll just be getting more data surfaced on failed transactions.

Out of curiosity, what did you learn in your testing that suggested the reason for the narrow scope of the original return block?

lib/active_merchant/billing/gateways/braintree_blue.rb
yunnydang: Yep that is true since Im setting credit_card_details =  in the transaction hash to equal what is returned from the helper methods! Thanks for this callout!

yunnydang:
hey @jcreiff I know you approved, but I just wanted some quick re-review on this as I tweaked a little bit of things rubocop related. Thanks!

jcreiff (approved):
Thanks for cleaning up the rubocop offenses - looks good!

yunnydang:
heya @jcreiff I added some more stuff to this as I remembered that we're also expecting the payment_instrument_type to be returned so it can dig into the correct object depending on payment methods. This way the response object can be actually surfaced, thanks!

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -499,6 +499,56 @@ def additional_processor_response_from_result(result)
         result.transaction&.additional_processor_response
       end
 
+      def payment_instrument_type(result)
+        if result
+          result&.payment_instrument_type
```
jcreiff, line 504: Rubocop doesn't like this `if` statement - I think you can remove it altogether because the safe operator should act as the guard against `nil` here

jcreiff (approved):
🚢

PR #5055, opened 2024-03-01
Adyen: Update failed for network token cryptogram

almalee24:
For v68 and later the network token cryptogram needs to be sent to mpi.tokenAuthenticationVerificationValue

Remote:
143 tests, 463 assertions, 11 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 92.3077% passed

Unit:
116 tests, 613 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -194,7 +194,8 @@ module ActiveMerchant #:nodoc:
           gsub(%r(("cavv\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
           gsub(%r(("bankLocationId\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
           gsub(%r(("iban\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
-          gsub(%r(("bankAccountNumber\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
+          gsub(%r(("bankAccountNumber\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("tokenAuthenticationVerificationValue\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]')
       end
 
       private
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,9 +623,11 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
 
         post[:mpiData] = {}
         post[:mpiData][:authenticationResponse] = 'Y'
-        post[:mpiData][:cavv] = payment.payment_cryptogram
         post[:mpiData][:directoryResponse] = 'Y'
         post[:mpiData][:eci] = payment.eci || '07'
```
aenand, line 627: Yikes. This default of 07 is only applicable to Visa but we also support NT for Mastercard. We should change the default depending on the card scheme. 

I think we have (had?) a ticket for that somewhere

almalee24: Oh really, let me look it up on their docs and also look for that ticket

aenand: It's not an Adyen thing/default it's an ECI value as a concept. IIRC all Mastercard NTs are failing right now on this gateway

aenand (commented):
@almalee24 do you have some Adyen docs to support this?

lib/active_merchant/billing/gateways/adyen.rb
almalee24: Do you know what the standard values are? I can't see anything in their docs for defaults and I still haven't heard back from Adyen yet

almalee24:
> @almalee24 do you have some Adyen docs to support this?

The docs that I have is for the [Checkout API](https://docs.adyen.com/online-payments/network-tokenization/) so I'm waiting for feedback on Adyen if this also involves the Classic API

lib/active_merchant/billing/gateways/adyen.rb
aenand: Just emailed you the message from Mastercard

aenand:
> > @almalee24 do you have some Adyen docs to support this?
> 
> The docs that I have is for the [Checkout API](https://docs.adyen.com/online-payments/network-tokenization/) so I'm waiting for feedback on Adyen if this also involves the Classic API

Nice! I'll review this PR when you get that feedback

almalee24:
> > > @almalee24 do you have some Adyen docs to support this?
> > 
> > 
> > The docs that I have is for the [Checkout API](https://docs.adyen.com/online-payments/network-tokenization/) so I'm waiting for feedback on Adyen if this also involves the Classic API
> 
> Nice! I'll review this PR when you get that feedback

@aenand I ended up finding documentation for this change https://docs.adyen.com/api-explorer/Payment/68/post/authorise#request-mpiData-tokenAuthenticationVerificationValue

aenand (approved):
nice work!

PR #5056, opened 2024-03-05
AuthorizeNet: Remove turn_on_nt_flow

almalee24:
Remove turn_on_nt_flow so that all network tokens
such as ApplePay fall done to dd_network_token method Unit:
122 tests, 688 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
85 tests, 304 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


DustinHaefele (approved):
This looks good to me. I like cleaning it up the properties in active merchant, and since we aren't continuing to pass the property.

PR #5058, opened 2024-03-08
Update Nexi Xpay basic transactions after implement 3DS 3steps API

sinourain:
This commit contains an update for Nexi Xpay basic transactions (capture, refund, verify)

Documents for reference:

https://developer.nexi.it/en/api/

For Spreedly reference:

[SER-1133](https://spreedly.atlassian.net/browse/SER-1133)

### DIFF
--- a/lib/active_merchant/billing/gateways/xpay.rb
+++ b/lib/active_merchant/billing/gateways/xpay.rb
@@ -20,7 +20,6 @@ module ActiveMerchant #:nodoc:
         preauth: 'orders/3steps/init',
         capture: 'operations/%s/captures',
         verify: 'orders/card_verification',
-        void: 'operations/%s/cancels',
         refund: 'operations/%s/refunds'
       }
 
@@ -32,39 +31,30 @@ module ActiveMerchant #:nodoc:
         super
       end
 
-      def preauth(amount, payment_method, options = {})
-        post = {}
-        payment_request(:preauth, amount, post, payment_method, options)
+      def preauth(amount, credit_card, options = {})
+        order_request(:preauth, amount, {}, credit_card, options)
       end
 
-      def purchase(amount, payment_method, options = {})
-        complete_transaction(:purchase, amount, payment_method, options)
+      def purchase(amount, credit_card, options = {})
+        complete_order_request(:purchase, amount, credit_card, options)
       end
 
-      def authorize(amount, payment_method, options = {})
-        complete_transaction(:authorize, amount, payment_method, options)
+      def authorize(amount, credit_card, options = {})
+        complete_order_request(:authorize, amount, credit_card, options)
       end
 
       def capture(amount, authorization, options = {})
-        post = {}
-        add_refund_capture_params(amount, post, options)
-        commit(:capture, post, options)
-      end
-
-      def void(authorization, options = {})
-        post = { description: options[:description] }
-        commit(:void, post, options)
+        operation_request(:capture, amount, authorization, options)
       end
 
       def refund(amount, authorization, options = {})
-        post = {}
-        add_refund_capture_params(amount, post, options)
-        commit(:refund, post, options)
+        operation_request(:refund, amount, authorization, options)
       end
 
       def verify(credit_card, options = {})
         post = {}
         add_invoice(post, 0, options)
+        add_customer_data(post, credit_card, options)
         add_credit_card(post, credit_card)
         commit(:verify, post, options)
       end
@@ -88,32 +78,28 @@ module ActiveMerchant #:nodoc:
         commit(:validation, post, options)
       end
 
-      def complete_transaction(action, amount, payment_method, options = {})
+      def complete_order_request(action, amount, credit_card, options = {})
         MultiResponse.run do |r|
           r.process { validation(options) }
-          r.process { payment_request(action, amount, {}, payment_method, options.merge!(validation: r.params)) }
+          r.process { order_request(action, amount, { captureType: (action == :authorize ? 'EXPLICIT' : 'IMPLICIT') }, credit_card, options.merge!(validation: r.params)) }
         end
       end
 
-      def payment_request(action, amount, post, payment_method, options = {})
-        add_capture_type(post, options, action)
-        add_auth_purchase_params(post, amount, payment_method, options)
+      def order_request(action, amount, post, credit_card, options = {})
+        add_invoice(post, amount, options)
+        add_credit_card(post, credit_card)
+        add_customer_data(post, credit_card, options)
+        add_address(post, options)
+        add_recurrence(post, options) unless options[:operation_id]
+        add_exemptions(post, options)
+        add_3ds_params(post, options[:validation]) if options[:validation]
+
         commit(action, post, options)
       end
 
-      def add_capture_type(post, options, action)
-        case action
-        when :purchase
-          post[:captureType] = 'IMPLICIT'
-        when :authorize
-          post[:captureType] = 'EXPLICIT'
-        end
-      end
-
-      def add_refund_capture_params(amount, post, options)
-        post[:amount] = amount
-        post[:currency] = options[:order][:currency]
-        post[:description] = options[:order][:description]
+      def operation_request(action, amount, authorization, options)
+        options[:correlation_id], options[:reference] = authorization.split('#')
+        commit(action, { amount: amount, currency: options[:currency] }, options)
       end
 
       def add_invoice(post, amount, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -112,8 +107,8 @@ def add_capture_type(post, options, action)
 
       def add_refund_capture_params(amount, post, options)
         post[:amount] = amount
-        post[:currency] = options[:order][:currency]
-        post[:description] = options[:order][:description]
+        post[:currency] = options[:currency]
```
Heavyblade, line 110: ❓ Question:

Isn't a problem the possibility that `description` or `currency` is not present?

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -253,16 +248,21 @@ def build_request_url(action, id = nil)
         base_url + endpoint
       end
 
-      def success_from(response)
-        SUCCESS_MESSAGES.include?(response.dig('operation', 'operationResult'))
+      def success_from(action, response)
```
Heavyblade, line 251: 💬 Comment/Style =>

Leaving a comment here for the `build_request_url` method, a simpler implementation:

```ruby
      def build_request_url(action, id = nil)
        "#{test? ? test_url : live_url}#{ENDPOINTS_MAPPING[action.to_sym] % id}"
      end
```

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -228,7 +223,7 @@ def commit(action, params, options)
       def request_headers(options, action = nil)
         headers = {
           'X-Api-Key' => @api_key,
-          'Correlation-Id' => options.dig(:order_id) || SecureRandom.uuid,
+          'Correlation-Id' => SecureRandom.uuid,
```
Heavyblade, line 226: ⚠️ Warning / ❓ Question:

Any reason to remove `order_id` as the idempotency key ?, that value is not something that should be Random for each request, implementation should be along the lines of:

order_id => authorize, purchase, verify
authorization_value => refund, capture

maybe we should confirm with product on this one ...

Heavyblade: 💬 Comment / Style:

Maybe a case for just one option is too much when we know that is unlikely to have more operations shortly:

```ruby
 headers.merge!('Idempotency-Key' => SecureRandom.uuid) if %i[refund capture].include?(action)
```

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -253,16 +248,21 @@ def build_request_url(action, id = nil)
...
+        else
+          SUCCESS_MESSAGES.include?(response.dig('operation', 'operationResult'))
+        end
       end
 
       def message_from(response)
         response.dig('operation', 'operationResult') || response.dig('errors', 0, 'description')
       end
 
       def authorization_from(response)
```
Heavyblade, line 264: 💬 Comment

I get the intention of the guard clause, what about setting the default value for `response` as {}, in that way you don't need the guard clause.

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -206,17 +201,17 @@ def parse(body = {})
 
       def commit(action, params, options)
         transaction_id = transaction_id_from(params, options, action)
-        begin
```
Heavyblade, line 209: ⚠️ Warning:

Leaving a comment here for the `parse` method, if we are going to provide a default value for it should be `"{}"` not `{}`

test/remote/gateways/remote_xpay_test.rb
```diff
@@ -22,44 +22,11 @@ def setup
     }
   end
 
-  def test_successful_purchase
-    response = @gateway.preauth(@amount, @credit_card, @options)
+  ## Test for authorization, capture, purchase and refund requires set up through 3ds
+  ## The only test that does not depend on a 3ds flow is verify
+  def test_successful_verify
```
Heavyblade, line 27: ❓ Question:

Just to confirm:

isn't worth adding a test that confirms that setting up the transaction works properly,? I mean, I know that is not possible to get a completely successful transaction here, but need to confirm any way that we are sending transaction data correctly, that we are handing errors properly and may that scrub code actually works.

Heavyblade (commented):
done with my review @sinourain, left some comments for your consideration.

lib/active_merchant/billing/gateways/xpay.rb
sinourain: Right now I cannot confirm it because although the documentation says that they are optional fields, the Nexi Xpay SandBox is not working and there is no way to verify it.

Reference: https://developer.nexi.it/en/api/post-operations-operationId-refunds

lib/active_merchant/billing/gateways/xpay.rb
sinourain: I will remove it, instead of that I will ensure that we will parse a valid json

lib/active_merchant/billing/gateways/xpay.rb
sinourain: Done!

test/remote/gateways/remote_xpay_test.rb
sinourain: good catch, as soon as the nexi xpay sandbox works again I will add it

lib/active_merchant/billing/gateways/xpay.rb
sinourain: After check with the SandBox working well we have:`currency` is mandatory, `description` is not mandatory, I will remove `description` field

lib/active_merchant/billing/gateways/xpay.rb
sinourain: Few days ago the Nexi Xpay SandBox works with the order_id in the place of the Correlation-Id, now we need to deal with separated fields, I will update the changes to support both

sinourain:
These changes were integrated into the following PR: https://github.com/spreedly/core/pull/8883

test/remote/gateways/remote_xpay_test.rb
sinourain: done!

test/remote/gateways/remote_xpay_test.rb
```diff
@@ -22,44 +23,17 @@ def setup
...
+    response = @gateway.verify(@credit_card, @options)
     assert_success response
-    assert_true response.params.has_key?('threeDSAuthUrl')
-    assert_true response.params.has_key?('threeDSAuthRequest')
-    assert_match 'PENDING', response.message
+    assert_match 'EXECUTED', response.message
   end
 
-  def test_failed_purchase
```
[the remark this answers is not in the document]
sinourain: Done!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -190,79 +172,66 @@ def add_3ds_validation_params(post, options)
...
-          'X-Api-Key' => @api_key,
-          'Correlation-Id' => options.dig(:order_id) || SecureRandom.uuid,
-          'Content-Type' => 'application/json'
-        }
-        case action
-        when :refund, :capture
-          headers.merge!('Idempotency-Key' => SecureRandom.uuid)
-        end
+        headers = { 'X-Api-Key' => @api_key, 'Content-Type' => 'application/json', 'Correlation-Id' => options[:correlation_id] }
+        headers.merge!('Idempotency-Key' => SecureRandom.uuid) if %i[capture refund].include?(action)
```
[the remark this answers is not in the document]
sinourain: Done!

lib/active_merchant/billing/gateways/xpay.rb
```diff
@@ -190,79 +172,66 @@ def add_3ds_validation_params(post, options)
...
-          response = parse(response)
-        end
+        raw_response =
+          begin
+            url = build_request_url(action, transaction_id)
+            ssl_post(url, params.to_json, request_headers(options, action))
+          rescue ResponseError => e
+            e.response.body
+          end
+        response = raw_response.blank? ? {} : parse(raw_response)
```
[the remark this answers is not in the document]
sinourain: Done!

test/unit/gateways/xpay_test.rb
```diff
@@ -51,18 +51,18 @@ def test_invalid_instance
...
   def test_check_request_headers
     stub_comms(@gateway, :ssl_post) do
-      @gateway.authorize(@amount, @credit_card, @options)
+      @gateway.preauth(@amount, @credit_card, @options)
     end.check_request do |_endpoint, _data, headers|
       assert_equal headers['Content-Type'], 'application/json'
       assert_equal headers['X-Api-Key'], 'some api key'
-    end.respond_with(successful_purchase_response)
+    end.respond_with(successful_preauth_response)
   end
```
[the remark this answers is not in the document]
sinourain: Done!

test/unit/gateways/xpay_test.rb
```diff
@@ -71,6 +71,51 @@ def test_scrub
     assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
   end
```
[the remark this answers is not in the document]
sinourain: Done!

naashton (approved):
LGTM!

PR #5060, opened 2024-03-11
SumUp - Remove Void method

sinourain:
After checking the docs we came to the conclusion that void should be removed because Sumup doesn’t support the Authorization-Caputure-Void flow with an important assumption:

Assumption:  The Standard behavior for a purchase operation is to return success in the way that is working today with our credentials and not in the way it was working some time ago returning pending


naashton (commented):
LGTM!

PR #5062, opened 2024-03-11
Authorize Net: add the surcharge field

yunnydang:
This adds the surcharge object as an optional field
Local:
5828 tests, 79061 assertions, 4 failures, 27 errors, 0 pendings, 0 omissions, 0 notifications
99.4681% passed

Unit:
123 tests, 694 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
87 tests, 310 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/authorize_net.rb
+++ b/lib/active_merchant/billing/gateways/authorize_net.rb
@@ -272,6 +272,7 @@ module ActiveMerchant
           add_market_type_device_type(xml, payment, options)
           add_settings(xml, payment, options)
           add_user_fields(xml, amount, options)
+          add_surcharge_fields(xml, options)
           add_ship_from_address(xml, options)
           add_processing_options(xml, options)
           add_subsequent_auth_information(xml, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/authorize_net.rb
```diff
@@ -710,6 +712,16 @@ def add_duty_fields(xml, options)
         end
       end
 
+      def add_surcharge_fields(xml, options)
+        surcharge = options[:surcharge] if options[:surcharge]
+        if surcharge.is_a?(Hash)
+          xml.surcharge do
+            xml.amount(amount(surcharge[:amount].to_i)) if surcharge[:amount]
```
jcreiff, line 719: I think the `to_i` might be unnecessary hear since the `money_format` of this gateway is `dollars` 🤔

jcreiff (commented):
Philosophical q: since the contents of the `surcharge` object are limited to two fields, could we/should we implement them separately as `surcharge_amount` and `surcharge_description` rather than opening up the object-level field?

lib/active_merchant/billing/gateways/authorize_net.rb
yunnydang: Hmm now that you mention it, I think that would be fine. Though if both fields are mandatory, we'll not that in the documentation

jcreiff (approved):
Discussed this with @yunnydang in person and it's good to go!

PR #5066, opened 2024-03-19
FatZebra: Adding third-party 3DS params

Heavyblade:
## Summary:

Adds the needed fields to support 3DS Global on FatZebra Gateway

[SER-1170](https://spreedly.atlassian.net/browse/SER-1170)

### Remote Test:
Finished in 87.765183 seconds.
29 tests, 101 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.1034% passed

*Note Failing Test*: We have to remote tests failing because seems to be a change on the sandbox behavior when a transaction doesn't include a cvv

### Unit Tests:
Finished in 68.736139 seconds.
5808 tests, 78988 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/fat_zebra.rb
+++ b/lib/active_merchant/billing/gateways/fat_zebra.rb
@@ -28,6 +28,7 @@ module ActiveMerchant #:nodoc:
         add_order_id(post, options)
         add_ip(post, options)
         add_metadata(post, options)
+        add_three_ds(post, options)
 
         commit(:post, 'purchases', post)
       end
@@ -41,6 +42,7 @@ module ActiveMerchant #:nodoc:
         add_order_id(post, options)
         add_ip(post, options)
         add_metadata(post, options)
+        add_three_ds(post, options)
 
         post[:capture] = false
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/fat_zebra.rb
```diff
@@ -125,16 +127,42 @@ def add_creditcard(post, creditcard, options = {})
...
         extra[:ecm] = '32' if options[:recurring]
-        extra[:cavv] = options[:cavv] || options.dig(:three_d_secure, :cavv) if options[:cavv] || options.dig(:three_d_secure, :cavv)
-        extra[:xid] = options[:xid] || options.dig(:three_d_secure, :xid) if options[:xid] || options.dig(:three_d_secure, :xid)
-        extra[:sli] = options[:sli] || options.dig(:three_d_secure, :eci) if options[:sli] || options.dig(:three_d_secure, :eci)
         extra[:name] = options[:merchant] if options[:merchant]
         extra[:location] = options[:merchant_location] if options[:merchant_location]
         extra[:card_on_file] = options.dig(:extra, :card_on_file) if options.dig(:extra, :card_on_file)
         extra[:auth_reason]  = options.dig(:extra, :auth_reason) if options.dig(:extra, :auth_reason)
+
+        unless options[:three_d_secure].present?
```
gasb150, line 135: This validation and those extra values are handled in the #add_three_ds method, the difference, is the place you are getting the field values.

❓ Question:

Why do you have two places to populate the sli, xid, and the cvv extra fields? (add_extra_options and add_three_ds methods), Is it possible to include the if options[:sli] conditional in the conditionals for add_3ds?

gasb150 (commented):
LGTM, I left a comment.

lib/active_merchant/billing/gateways/fat_zebra.rb
Heavyblade: @gasb150 I found that someone basically added some fields related to 3DS in the options hash, so to prevent damage on someonelse work I leaved there just in case.

naashton (approved):
LGTM!

PR #5068, opened 2024-03-25
CheckoutV2: add processing and recipient fields

yunnydang:
This change allows us to pass the processing and recipient object with their subfields. Since the processing field is going to be wide open, I added the complex objects (not all) to remote testing to ensure that we can simply pass it along. In order to use the processing field, you will have to adhere to the structure checkout had defined in their documentations. and example can be found here: https://api-reference.checkout.com/#operation/getPaymentDetails

Also fixed some rubocop errors in xpay gateway.

Local:
5838 tests, 79131 assertions, 2 failures, 26 errors, 0 pendings, 0 omissions, 0 notifications
99.5204% passed

Unit:
66 tests, 402 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
101 tests, 235 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.0495% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -148,6 +148,8 @@ module ActiveMerchant #:nodoc:
         add_metadata(post, options, payment_method)
         add_processing_channel(post, options)
         add_marketplace_data(post, options)
+        add_recipient_data(post, options)
+        add_processing_data(post, options)
       end
 
       def add_invoice(post, money, options)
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -163,6 +165,36 @@ def add_invoice(post, money, options)
         post[:metadata][:udf5] = application_id || 'ActiveMerchant'
       end
 
+      def add_recipient_data(post, options)
+        return unless recipient = options[:recipient]
+
+        post[:recipient] = {}
+        post[:recipient][:dob] = recipient[:dob] if recipient[:dob]
+        post[:recipient][:account_number] = recipient[:account_number] if recipient[:account_number]
+        post[:recipient][:zip] = recipient[:zip] if recipient[:zip]
```
jcreiff, line 174: The [checkout docs](https://api-reference.checkout.com/#operation/requestAPaymentOrPayout!path=0/recipient/zip&t=request) say this is deprecated in favor of the `zip` nested under `address` so I think it's ok to remove this

yunnydang: nice good catch!

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -163,6 +165,35 @@ def add_invoice(post, money, options)
         post[:metadata][:udf5] = application_id || 'ActiveMerchant'
       end
 
+      def add_recipient_data(post, options)
+        return unless recipient = options[:recipient]
```
jcreiff, line 169: We might want to add a check to make sure that options[:recipient] is a Hash -- otherwise the attempts to access data from within might throw NoMethodError if someone passes in a string

jcreiff (changes requested):
Still pending a conversation about `processing`, but I do think this lil change for a type check would be good 👍🏻

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -163,6 +165,35 @@ def add_invoice(post, money, options)
...
+        post[:recipient] = {}
+        post[:recipient][:dob] = recipient[:dob] if recipient[:dob]
+        post[:recipient][:zip] = recipient[:zip] if recipient[:zip]
+        post[:recipient][:account_number] = recipient[:account_number] if recipient[:account_number]
+        post[:recipient][:first_name] = recipient[:first_name] if recipient[:first_name]
+        post[:recipient][:last_name] = recipient[:last_name] if recipient[:last_name]
+
+        if address = recipient[:address]
+          post[:recipient][:address] = {}
+          post[:recipient][:address][:address_line1] = address[:address_line_1] if address[:address_line_1]
```
jcreiff, line 182: This is nitpicky, but I think I'd prefer for us to expect `address_line1` instead of `address_line_1`

Since we are more or less mapping this directly to Checkout's hash structure, sneaking in that additional underscore could be confusing; it would be more clear if we match the Checkout naming conventions exactly

jcreiff (changes requested):
One observation about naming convention, but also I would encourage you to run Rubocop and address the violations👮🏻‍♂️ 

But otherwise this looks good, and all the unit and remote tests pass for me!

jcreiff (approved):
Looks good to me!

PR #5075, opened 2024-03-29
StripePI: Update for network tokens

almalee24:
Remote using the Token endpoint for ApplePay/GooglePay and the PaymentMethod endpoint to have a more uniform and simple payment method.

Remote
93 tests, 442 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
59 tests, 300 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


almalee24:
> Does Stripe require this feature to be gated in?

Yes, it's gated so they would need to reach out to Stripe to use this new route

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -11,48 +11,43 @@ class StripePaymentIntentsGateway < StripeGateway
       CONFIRM_INTENT_ATTRIBUTES = %i[receipt_email return_url save_payment_method setup_future_usage off_session]
       UPDATE_INTENT_ATTRIBUTES = %i[description statement_descriptor_suffix statement_descriptor receipt_email setup_future_usage]
       DEFAULT_API_VERSION = '2020-08-27'
+      NETWORK_TOKENS = {
+        apple_pay: 'apple_pay',
+        google_pay: 'google_pay_dpan',
+        untokenized_google_pay: 'google_pay_ecommerce_token'
+      }
```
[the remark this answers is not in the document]
almalee24: Network Tokens doesn't need a tokenization_method and the fields we send now are up to date.

almalee24:
> > Does Stripe require this feature to be gated in?
> 
> Yes, it's gated so they would need to reach out to Stripe to use this new route

This makes me think that we might also need to gate this first before introducing such a big change

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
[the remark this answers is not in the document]
aenand: Can we rename this const then to something like `DIGITAL_WALLETS` or something? I'm worried about this getting mixed with regular NT

aenand:
> > > Does Stripe require this feature to be gated in?
> > 
> > 
> > Yes, it's gated so they would need to reach out to Stripe to use this new route
> 
> This makes me think that we might also need to gate this first before introducing such a big change

Yeah we definitely need to include a flag here to toggle using this

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -399,6 +432,36 @@ def add_network_token_cryptogram_and_eci(post, payment_method)
         post[:payment_method_options][:card][:network_token][:electronic_commerce_indicator] = payment_method.eci if payment_method.eci
       end
 
+      def add_network_token(post, payment_method, options)
```
aenand, line 435: Can we rename this to be `add_digital_wallet` or something else since regular NT doesn't flow through these rails?

almalee24: Yep I can! In the original one it did, I really hope we can clean up this gateway soon

aenand (approved):
This looks good to me, nice work! I've been in bizarro world for a bit so would appreciate another review :) \

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -83,6 +97,12 @@ def create_payment_method(payment_method, options = {})
         commit(:post, 'payment_methods', post_data, options)
       end
 
+      def new_apple_google_pay_flow(payment_method, options)
```
bradbroge, line 100: there will inevitably be a `newer` apple/goole pay flow.  Is there a different word we can use to distinguish this flow that isn't `new`

almalee24: We are hoping to remove this in a few months this is just for now. So I think new would apply since its temporary

PR #5076, opened 2024-04-01
WIP: Add FlexCharge gateway

sinourain:


lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+        url = build_request_url(:authenticate)
+        response = parse(ssl_post(url, params.to_json, headers))
+        @options[:access_token] = response[:accessToken]
+        @options[:expires] = response[:expires]
+      rescue ResponseError => e
+        raise OAuthResponseError.new(e)
+      end
+
+      def ensure_access_token_credential(try_again = true)
+        return if @options[:access_token] && @options[:expires] && ((@options[:expires] || 0) > DateTime.now.strftime('%Q').to_i)
```
Heavyblade, line 251: 💬 Comment  / Style:

Maybe the double-check is not needed, you can achieve the same results with a default value:

```ruby
return if @options[:access_token] && @options.fetch(:expires, 0) > DateTime.now.strftime('%Q').to_i
```

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+
+        fetch_access_token
+      rescue ActiveMerchant::OAuthResponseError => e
```
Heavyblade, line 254: 💬 Comment / Question:

Maybe I'm not seeing something here but what is the point of rescuing an exception to raise the same exception without doing any other action ?, it makes sense on the `fetch_access_token` method to change the error type but in this case, is just indirection.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+        raise e
+      rescue ResponseError => e
+        raise e unless retry?(try_again, e)
```
Heavyblade, line 257: ⚠️ Warning ⚠️ 

Unless I am missing something this looks like a copy-paste error from the Alelo code, here we don't need this retry, Alelo had an ephemeral encryption key but that's not the case with Flexcharge.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+
+        @options.delete(:access_token)
+        ensure_access_token_credential(false)
+      end
+
+      def retry?(try_again, error)
+        try_again && %w(401 404).include?(error.response.code) && @options[:access_token].present?
+      end
+
+      def generate_uuid
```
Heavyblade, line 267: 💬 Comment

This method is not being used anywhere.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+        SecureRandom.uuid
+      end
+
+      def build_request_url(action, id = nil)
+        "#{test? ? test_url : live_url}#{ENDPOINTS_MAPPING[action] % id}"
+      end
+
+      def headers
```
Heavyblade, line 275: 💬 Comment / Style:

Just a matter of taste but I found forcing that merge not that clear, maybe tap can do the work more explicitly:

```ruby
def headers
  { 'Content-Type' => 'application/json' }.tap do |headers|
    headers['Authorization'] = "Bearer #{@options[:access_token]}" if @options[:access_token]
  end
end
```

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+            verification_value: credit_card.verification_value,
+            month: credit_card.month,
+            year: credit_card.year
+          }
+        }
+      end
+
+      def add_post_data(post)
+        post[:mid] = @options[:mid]
+        post[:environment] = @options[:tokenization_key] || @options[:access_token]
```
Heavyblade, line 191: ❓ Question:

I can't find the `environment` parameter and I can't infer how the tokenization_key or access_token are related to it.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+      def parse(body)
+        JSON.parse(body).with_indifferent_access
+      end
+
+      def commit(action, post, authorization = nil)
+        ensure_access_token_credential
+        response =
+          begin
+            url = build_request_url(action, authorization)
+            add_post_data(post) if action == :tokenize
```
Heavyblade, line 290: 💬 Comment:

This is too specific to the `tokenize` action and should 'live' there, assuming we are going to keep it

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+          error_code: error_code_from(response)
+        )
+      end
+
+      def success_from(response)
+        return false unless response[:success]
+
+        return true if response.dig(:transaction, :payment_method, :token)
+
+        SUCCESS_MESSAGES.include?(response[:status])
```
Heavyblade, line 311: 💬  Comment/Suggestion:

Maybe a more direct approach:

```ruby
def success_from(response)
  response[:success] && SUCCESS_MESSAGES.include?(response[:status]) ||
  response.dig(:transaction, :payment_method, :token).present?
end
```

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+      end
+
+      def message_from(response)
+        response[:errors].present? ? response[:errors] : response[:status]
```
Heavyblade, line 315: ❓ Question:

Maybe the name is misleading but `errors` sounds like an array, are we going to return a string version of an array on the message response?

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+      end
+
+      def authorization_from(response)
+        response[:orderSessionKey]
+      end
+
+      def error_code_from(response)
+        response[:status] unless success_from(response)
```
Heavyblade, line 323: 💬 Comment:

Unless this API is weird an the status is not an error code, is better to have a blank error code than an incorrect one in the field.

test/fixtures.yml
```diff
@@ -426,6 +426,11 @@ firstdata_e4_v27:
   key_id: ANINTEGER
   hmac_key: AMAGICALKEY
 
+# Working credentials, no need to replace
```
Heavyblade, line 429: 💬 Comment:

Seems that :app_key, :app_secret, :site_id and :mid should be listed here

Heavyblade (commented):
first round with out checking test files is done @sinourain, couple general comment:

1) I'm seeing a lot of optional fields with a ton of possible GSFs, try to make the failure surface small and achieve successful transactions with the minimal data required.

2) as a consecuence of (1) and given that there is not cleaning `compact` on the parameters there are some fields that are being sent as `null`

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+          city: address[:city],
+          zipCode: address[:zip]
+        }
+      end
+
+      def add_invoice(post, money, options)
+        post[:transaction] = {
+          id: options[:order_id],
+          dynamicDescriptor: options[:dynamic_descriptor],
+          timestampUtc: options[:timestamp_utc] || Time.now,
```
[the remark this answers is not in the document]
sinourain: After doing some tests I removed it because it is optional. However, this suggestion helped me elsewhere that also uses UTC.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+            'Auth'
+          when :capture
+            'Capture'
+          when :void
+            'Void'
+          end
+      end
+
+      def add_payment(post, credit_card, address, options)
+        post[:paymentMethod] = {
```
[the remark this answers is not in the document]
sinourain: Yes, create a flag called `tokenize` to use the tokenize endpoint, when this happens the payment and token fields would be there and used, otherwise the credit card fields per se would be used.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+          schemeBrand: credit_card.brand,
+          interval: options[:subscription_interval],
+          price: money,
+          currency: options[:currency]
+        }
+      end
+
+      def add_customer_data(post, options)
+        post[:payer] = {
+          id: options[:order_id],
```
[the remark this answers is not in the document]
sinourain: After doing some tests I removed it. Thanks!

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+
+      def purchase(money, credit_card, options = {})
+        evaluate(:purchase, money, credit_card, options)
+      end
+
+      def authorize(money, credit_card, options = {})
+        evaluate(:authorize, money, credit_card, options)
+      end
+
+      def capture(money, authorization, options = {})
```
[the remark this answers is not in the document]
sinourain: Yeah that's right. I'm trying to complete an authorization with `challenge` status to see if I can do the `capture` and `void`

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,327 @@
...
+          cavvResultCode: options[:cavv_result_code] || 'NA',
+          cardNotPresent: options[:card_not_present]
+        }
+      end
+
+      def add_transaction_type(post, action)
+        return unless %i(authorize capture void).include?(action)
+
+        post[:transaction] ||= {}
+        post[:transaction][:transactionType] =
```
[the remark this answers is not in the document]
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Removed!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Fixed!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Yes, it is an array

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: Done!

test/fixtures.yml
sinourain: Done!

lib/active_merchant/billing/gateways/flex_charge.rb
sinourain: I will change it to string

PR #5078, opened 2024-04-02
Adyen: Update "unexpected 3DS authentication response" error message …

almalee24:
…logic

Update "unexpected 3DS authentication response" error message logic to look for !options[:execute_threed] to cover all instances where the request didn't expect to perform 3DS.

Unit:
116 tests, 611 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
143 tests, 463 assertions, 11 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 92.3077% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -827,7 +827,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(action, response, options)
-        if %w[RedirectShopper ChallengeShopper].include?(response.dig('resultCode')) && !options[:execute_threed] && !options[:threed_dynamic]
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -827,7 +827,7 @@ def request_headers(options)
       end
 
       def success_from(action, response, options)
-        if %w[RedirectShopper ChallengeShopper].include?(response.dig('resultCode')) && !options[:execute_threed] && !options[:threed_dynamic]
```
almalee24, line 830: decided to just remove  !options[:threed_dynamic] since !options[:execute_threed] is sufficient and there's a similar check in StripePI

DustinHaefele:
Do you have any links to adyen's docs about the `threed_dynamic` flag. I'm concerned that we have customers using it purposely with execute_threeds = false and `threed_dynamic = true` and expecting 3ds dynamic to take place.

bradbroge (approved):
👍🏻

PR #5082, opened 2024-04-03
ECS-3370 Paymentez: Update success_from

jherreraa:
[ECS-3370](https://spreedly.atlassian.net/browse/ECS-3370)

## Summary:

This PR updates success_from method on paymentez gateway in order to handle success status on a proper way

Unit tests
----------------
Finished in 0.043317 seconds.

30 tests, 127 assertions, 0 failures,
0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote tests
----------------
Finished in 140.746551 seconds.

34 tests, 85 assertions, 0 failures,
0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -34,7 +34,7 @@ module ActiveMerchant #:nodoc:
         28 => :card_declined
       }.freeze
 
-      SUCCESS_STATUS = ['success', 'pending', 1, 0]
+      SUCCESS_STATUS = ['APPROVED', 'PENDING', 'success', 1, 0]
 
       CARD_MAPPING = {
         'visa' => 'vi',
@@ -263,7 +263,7 @@ module ActiveMerchant #:nodoc:
       def commit_transaction(action, parameters)
         response = commit_raw('transaction', action, parameters)
         Response.new(
-          success_from(response),
+          success_from(response, action),
           message_from(response),
           response,
           authorization: authorization_from(response),
@@ -291,10 +291,12 @@ module ActiveMerchant #:nodoc:
         }
       end
 
-      def success_from(response)
-        return false if response.include?('error')
-
-        SUCCESS_STATUS.include?(response['status'] || response['transaction']['status'])
+      def success_from(response, action = nil)
+        if action == 'refund'
+          response.dig('transaction', 'status_detail') == 7 || SUCCESS_STATUS.include?(response.dig('transaction', 'current_status') || response['status'])
+        else
+          SUCCESS_STATUS.include?(response.dig('transaction', 'current_status') || response['status'])
+        end
       end
 
       def card_success_from(response)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -142,6 +142,7 @@
 * SumUp: Remove Void method [sinourain] #5060
 * StripePI: Add new ApplePay and GooglePay flow [almalee24] #5075
 * Braintree: Add merchant_account_id to Verify [almalee24] #5070
+* Paymentez: Update success_from [jherrera] #5082
 
 == Version 1.135.0 (August 24, 2023)
 * PaymentExpress: Correct endpoints [steveh] #4827
--- a/test/unit/gateways/paymentez_test.rb
+++ b/test/unit/gateways/paymentez_test.rb
@@ -91,7 +91,6 @@ class PaymentezTest < Test::Unit::TestCase
 
     response = @gateway.purchase(@amount, '123456789012345678901234567890', @options)
     assert_success response
-
     assert_equal 'PR-926', response.authorization
     assert response.test?
   end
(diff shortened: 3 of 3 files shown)


lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -292,9 +293,13 @@ def headers
...
       def success_from(response)
-        return false if response.include?('error')
+        return false if response.include?('error') || FAILURE_STATUS.include?(response['status'])
 
-        SUCCESS_STATUS.include?(response['status'] || response['transaction']['status'])
+        if response['transaction']
+          SUCCESS_STATUS.include?(response['transaction']['current_status']) || response['transaction']['status_detail'] == 7
+        else
+          response['status']
+        end
```
sinourain, line 302: I suggest you can check this other way::

```suggestion
        SUCCESS_STATUS.include?(response['status'] || response.dig('transaction', 'current_status')) ||
        response.dig('transaction', 'status_detail') == 7
```

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -34,7 +34,8 @@ class PaymentezGateway < Gateway #:nodoc:
         28 => :card_declined
       }.freeze
 
-      SUCCESS_STATUS = ['success', 'pending', 1, 0]
+      SUCCESS_STATUS = ['APPROVED', 'PENDING', 1, 0]
+      FAILURE_STATUS = ['failure', 'Internal server error']
```
sinourain, line 38: I did some tests but I couldn't get: 'Internal server error'. How can we obtain this result?

sinourain (approved):
Good job, just leave a couple of comments

lib/active_merchant/billing/gateways/paymentez.rb
jherreraa: The specific unit test ´test_paymentez_crashes_fail´ gives that result.

lib/active_merchant/billing/gateways/paymentez.rb
jherreraa: Noted!!! Thanks!!

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -292,9 +293,13 @@ def headers
       end
 
       def success_from(response)
-        return false if response.include?('error')
+        return false if response.include?('error') || FAILURE_STATUS.include?(response['status'])
```
almalee24, line 296: This should also be updated to look for current_status

jherreraa: Noted!! thanks!

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -292,9 +293,11 @@ def headers
       end
 
       def success_from(response)
-        return false if response.include?('error')
+        return false if response.include?('error') || FAILURE_STATUS.include?(response['status'] || response.dig('transaction', 'current_status'))
```
almalee24, line 296: Make sure that  response.dig('transaction', 'current_status') is CANCELLED, REJECTED and EXPIRED would be a failed status. And current_status should be before status.

almalee24: But maybe it's better just to get rid of response['status']

almalee24: Well to clean this up should we just check that 'current_status' is APPROVED or PENDING and if not we return false?

jherreraa: According to my understanding in documentation, when 'current_status' is CANCELLED and 'status_detail' is 7 the transaction is successful in the context of refund so I could not reject every CANCELLED transaction, please let me know if my understanding is wrong

almalee24: Oh okay, then I think we should pass action to this method. If the action is a refund then we should check for what you said if it's not then we should just check for APPROVED or PENDING.

jherreraa: Thanks a lot for the help! Corrected

lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -291,10 +291,12 @@ def headers
         }
       end
 
-      def success_from(response)
-        return false if response.include?('error')
-
-        SUCCESS_STATUS.include?(response['status'] || response['transaction']['status'])
+      def success_from(response, action = nil)
+        if action == 'refund'
+          return response.dig('transaction', 'status_detail') == 7 || SUCCESS_STATUS.include?(response['status'])
```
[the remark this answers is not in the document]
jherreraa: Hello! Here is the link where I found that after checking with the refund responses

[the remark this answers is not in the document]
almalee24: Okay, I did see that docs. Can you just change the SUCCESS_STATUS to be SUCCESS_STATUS.include?(response.dig('transaction', 'current_status') || response['status']). I also don't think it's necessary to have the returns in this method. But other than that everything else looks good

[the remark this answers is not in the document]
jherreraa: understood, noted!

almalee24 (approved):

PR #5083, opened 2024-04-04
Cybersource Rest: Stored Credential refactor

aenand:
COMP-78

Refactors the stored credential support for the Cybersource Rest
gateway to be in-line with their documentation. Also repairs test
suite for this gateway by eliminating certain tests and fixing
others.

Test summary:
Local:
5838 tests, 79156 assertions, 2 failures, 23 errors, 0 pendings, 0 omissions, 0 notifications
99.5718% passed
Unit:
30 tests, 144 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
43 tests, 143 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -322,11 +322,7 @@ module ActiveMerchant #:nodoc:
       def commerce_indicator(reason_type, country = '')
         case reason_type
         when 'recurring'
-          if country == 'US'
-            'recurring'
-          else
-            'recurring'
-          end
+          'recurring'
         when 'installment'
           'install'
         else
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -315,11 +315,11 @@ module ActiveMerchant #:nodoc:
       def add_stored_credentials(post, payment, options)
         return unless options[:stored_credential]
 
-        post[:processingInformation][:commerceIndicator] = commerce_indicator(options.dig(:stored_credential, :reason_type), lookup_country_code(options.dig(:billing_address, :country))&.value)
+        post[:processingInformation][:commerceIndicator] = commerce_indicator(options.dig(:stored_credential, :reason_type))
         add_authorization_options(post, payment, options)
       end
 
-      def commerce_indicator(reason_type, country = '')
+      def commerce_indicator(reason_type)
         case reason_type
         when 'recurring'
           'recurring'
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -315,14 +315,18 @@ module ActiveMerchant #:nodoc:
       def add_stored_credentials(post, payment, options)
         return unless options[:stored_credential]
 
-        post[:processingInformation][:commerceIndicator] = commerce_indicator(options.dig(:stored_credential, :reason_type))
-        add_authorization_options(post, options)
+        post[:processingInformation][:commerceIndicator] = commerce_indicator(options.dig(:stored_credential, :reason_type), lookup_country_code(options.dig(:billing_address, :country))&.value)
+        add_authorization_options(post, payment, options)
       end
 
-      def commerce_indicator(reason_type)
+      def commerce_indicator(reason_type, country = '')
         case reason_type
         when 'recurring'
-          'recurring'
+          if country == 'US'
+            'recurring'
+          else
+            'recurring'
+          end
         when 'installment'
           'install'
         else
@@ -330,7 +334,7 @@ module ActiveMerchant #:nodoc:
         end
       end
 
-      def add_authorization_options(post, options)
+      def add_authorization_options(post, payment, options)
         initiator = options.dig(:stored_credential, :initiator) == 'cardholder' ? 'customer' : 'merchant'
         authorization_options = {
           authorizationOptions: {
@@ -346,7 +350,7 @@ module ActiveMerchant #:nodoc:
         unless options.dig(:stored_credential, :initial_transaction)
           network_transaction_id = options[:network_transaction_id] || options.dig(:stored_credential, :network_transaction_id) || ''
           authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:previousTransactionID] = network_transaction_id
-          authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:originalAuthorizedAmount] = post.dig(:orderInformation, :amountDetails, :totalAmount)
+          authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:originalAuthorizedAmount] = post.dig(:orderInformation, :amountDetails, :totalAmount) if card_brand(payment) == 'discover'
         end
         authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:reason] = options[:reason_code] if options[:reason_code]
         post[:processingInformation].merge!(authorization_options)
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -335,12 +335,12 @@ module ActiveMerchant #:nodoc:
         authorization_options = {
           authorizationOptions: {
             initiator: {
-              type: initiator,
-              storedCredentialUsed: true,
+              type: initiator
             }
           }
         }.compact
 
+        authorization_options[:authorizationOptions][:initiator][:storedCredentialUsed] = true if initiator == 'merchant'
         authorization_options[:authorizationOptions][:initiator][:credentialStoredOnFile] = true if options.dig(:stored_credential, :initial_transaction)
         authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction] ||= {}
         unless options.dig(:stored_credential, :initial_transaction)
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -313,56 +313,43 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_stored_credentials(post, payment, options)
-        return unless stored_credential = options[:stored_credential]
+        return unless options[:stored_credential]
 
-        options = stored_credential_options(stored_credential, options.fetch(:reason_code, ''))
-        post[:processingInformation][:commerceIndicator] = options.fetch(:transaction_type, 'internet')
-        stored_credential[:initial_transaction] ? initial_transaction(post, options) : subsequent_transaction(post, options)
+        post[:processingInformation][:commerceIndicator] = commerce_indicator(options.dig(:stored_credential, :reason_type))
+        add_authorization_options(post, options)
       end
 
-      def stored_credential_options(options, reason_code)
-        transaction_type = options[:reason_type]
-        transaction_type = 'install' if transaction_type == 'installment'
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -258,56 +258,43 @@ def add_merchant_description(post, options)
...
-        {
+      def add_authorization_options(post, options)
+        initiator = options.dig(:stored_credential, :initiator) == 'cardholder' ? 'customer' : 'merchant'
+        authorization_options = {
           authorizationOptions: {
             initiator: {
               type: initiator,
-              merchantInitiatedTransaction: merchant_initiated_transaction_hash,
-              storedCredentialUsed: true
+              storedCredentialUsed: true,
```
almalee24, line 284: Is storedCredentialUsed supposed to be only for subsequent transactions?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -258,56 +258,43 @@ def add_merchant_description(post, options)
...
-        })
-        post[:processingInformation].merge!(processing_information)
+        authorization_options[:authorizationOptions][:initiator][:credentialStoredOnFile] = true if options.dig(:stored_credential, :initial_transaction)
+        authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction] ||= {}
+        unless options.dig(:stored_credential, :initial_transaction)
+          network_transaction_id = options[:network_transaction_id] || options.dig(:stored_credential, :network_transaction_id) || ''
+          authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:previousTransactionID] = network_transaction_id
+          authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:originalAuthorizedAmount] = post.dig(:orderInformation, :amountDetails, :totalAmount)
+        end
+        authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:reason] = options[:reason_code] if options[:reason_code]
```
almalee24, line 296: Can this also only be send if it's a subsequent transactions?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: Their docs indicate this:
> true means the merchant-initiated transaction came from a card that was already stored on file.
false means the merchant-initiated transaction came from a card that was not stored on file.

Re-reading it I think I'm going to add a check that only includes this if it's an MIT!

almalee24: Can you please share the link to the docs?

aenand: https://developer.cybersource.com/api-reference-assets/index.html#payments_payments_process-a-payment

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -258,56 +258,43 @@ def add_merchant_description(post, options)
...
-      def subsequent_transaction(post, options)
-        network_transaction_id = options[:network_transaction_id] || options.dig(:stored_credential, :network_transaction_id) || ''
-        processing_information = add_processing_information(options[:initiator], {
-          originalAuthorizedAmount: post.dig(:orderInformation, :amountDetails, :totalAmount),
-          previousTransactionID: network_transaction_id,
-          reason: options[:reason_code],
-          storedCredentialUsed: options[:stored_on_file]
-        })
-        post[:processingInformation].merge!(processing_information)
+        authorization_options[:authorizationOptions][:initiator][:storedCredentialUsed] = true if initiator == 'merchant'
```
almalee24, line 288: Looking at their documentation this can be send for any subsequent transactions as long as it's not the initial transaction. 

> When a customer-initiated transaction (CIT) uses stored credentials, the transaction must include the following information in the authorization request:
• processingInformation.authorizationOptions.initiator.storedCredentialUsed–Set the field to true to indicate that the previously stored customer credentials were used.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -258,56 +258,43 @@ def add_merchant_description(post, options)
...
+        authorization_options[:authorizationOptions][:initiator][:credentialStoredOnFile] = true if options.dig(:stored_credential, :initial_transaction)
+        authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction] ||= {}
+        unless options.dig(:stored_credential, :initial_transaction)
+          network_transaction_id = options[:network_transaction_id] || options.dig(:stored_credential, :network_transaction_id) || ''
+          authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:previousTransactionID] = network_transaction_id
+          authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction][:originalAuthorizedAmount] = post.dig(:orderInformation, :amountDetails, :totalAmount)
```
almalee24, line 294: It looks to me like this is only required by Discover. Could their be any adverse results if we send it for all card types?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -258,56 +258,43 @@ def add_merchant_description(post, options)
...
-        processing_information = add_processing_information(options[:initiator], {
-          originalAuthorizedAmount: post.dig(:orderInformation, :amountDetails, :totalAmount),
-          previousTransactionID: network_transaction_id,
-          reason: options[:reason_code],
-          storedCredentialUsed: options[:stored_on_file]
-        })
-        post[:processingInformation].merge!(processing_information)
+        authorization_options[:authorizationOptions][:initiator][:storedCredentialUsed] = true if initiator == 'merchant'
+        authorization_options[:authorizationOptions][:initiator][:credentialStoredOnFile] = true if options.dig(:stored_credential, :initial_transaction)
+        authorization_options[:authorizationOptions][:initiator][:merchantInitiatedTransaction] ||= {}
```
almalee24, line 290: Could anything happen if we just send a empty hash? For example, if it's an initial transaction and the reason_code wasn't passed in options.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -258,56 +258,43 @@ def add_merchant_description(post, options)
...
-        stored_on_file = options[:reason_type] == 'recurring'
-        options.merge({
-          transaction_type: transaction_type,
-          initiator: initiator,
-          reason_code: reason_code,
-          stored_on_file: stored_on_file
-        })
+      def commerce_indicator(reason_type)
+        case reason_type
+        when 'recurring'
```
almalee24, line 269: I'll forward you the latest reply from CyberSource but they basically send that it should only be recurring for non-US transaction if it's a subsequent transaction. If it's an initial transaction it should be recurring_internet for non-US transactions.

I did double check on the API fields PDF for CyberSource Rest and recurring_internet appears as a option for commerceIndicator

test/remote/gateways/remote_cyber_source_rest_test.rb
```diff
@@ -146,13 +146,13 @@ def test_successful_capture_with_partial_amount
     assert_equal 'PENDING', response.message
   end
 
-  def test_failure_capture_with_higher_amount
-    authorize = @gateway.authorize(@amount, @visa_card, @options)
-    response = @gateway.capture(@amount + 10, authorize.authorization, @options)
+  # def test_failure_capture_with_higher_amount
```
almalee24, line 149: Could you add test to ensure we are sending the appropriate commerceIndicator fields for NTs when using stored credentials?

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: Where did you see that? Their API reference only mentions MITs for this

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: Hmmm I don't see any indications either way. Their sandbox accepts it for all transactions but I can conditionally add it to be on the safe side!

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: I don't think so, this isn't an XML based integration where `nils` aren't allowed. Additionally the sandbox remote tests accept it

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: How can we tell non-US transactions though? Billing address country can come in all sorts of formats

test/remote/gateways/remote_cyber_source_rest_test.rb
aenand: I can't because NT is not merged into master. Once that is I can write tests

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: Wait I see an option

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: Also there is a .compact at the end of the build_auth_request which will filter out `nils`

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: Hmmmm when testing this out in sandbox they **only** want recurring

aenand: The docs from cybersource rest's email indicates only recurring is desired. I've emailed them for more information

lib/active_merchant/billing/gateways/cyber_source_rest.rb
almalee24: Page 13 of https://developer.cybersource.com/content/dam/docs/cybs/en-us/payments/developer/ctv/rest/credentials.pdf

lib/active_merchant/billing/gateways/cyber_source_rest.rb
almalee24: It the docs I shared it says this is only for MasterCard transactions but since this is a GSF I'm not sure if we should add that check

lib/active_merchant/billing/gateways/cyber_source_rest.rb
DustinHaefele: It actually seems to be even more niche than just discover 

> This field is supported only for Apple Pay, Google Pay, and Samsung Pay transactions with Discover on FDC Nashville Global and Chase Paymentech

aenand:
@almalee24 we heard back from Cybersource and have updated the stored credential fields. It is not the same values as the legacy integration. notably, `recurring_internet` is not in scope.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -294,56 +294,43 @@ def add_merchant_description(post, options)
...
-        initiator = options[:initiator] if  options[:initiator]
-        initiator = 'customer' if initiator == 'cardholder'
-        stored_on_file = options[:reason_type] == 'recurring'
-        options.merge({
-          transaction_type: transaction_type,
-          initiator: initiator,
-          reason_code: reason_code,
-          stored_on_file: stored_on_file
-        })
+      def commerce_indicator(reason_type, country = '')
```
almalee24, line 303: Why do we need country here?

aenand: Good catch! We don't need this anymore

PR #5086, opened 2024-04-05
PayTrace: Always send name in billing_address

almalee24:
If name is present in payment method send even if
billing address is nil.

Unit:
30 tests, 156 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
33 tests, 84 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/pay_trace.rb
+++ b/lib/active_merchant/billing/gateways/pay_trace.rb
@@ -265,15 +265,16 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_address(post, creditcard, options)
-        return unless options[:billing_address] || options[:address]
-
-        address = options[:billing_address] || options[:address]
         post[:billing_address] = {}
+
+        if (address = options[:billing_address] || options[:address])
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/pay_trace.rb
```diff
@@ -265,15 +265,16 @@ def add_customer_data(post, options)
       end
 
       def add_address(post, creditcard, options)
-        return unless options[:billing_address] || options[:address]
-
-        address = options[:billing_address] || options[:address]
         post[:billing_address] = {}
+
+        if (address = options[:billing_address] || options[:address])
```
bradbroge, line 270: What happens if the address isn't fully filled out?  Eg. `options[:address][:city]` is `nil`?

almalee24: At least in sandbox nothing happens, the transactions succeeds.

bradbroge (approved):
LGTM

PR #5087, opened 2024-04-08
SumUp - Improve success_from and message_from methods

sinourain:
Improve the way that we detect successful transactions and the way that we extract messages for SumUp gateway

Spreedly reference:

[SER-1188](https://spreedly.atlassian.net/browse/SER-1188)


lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -157,15 +157,10 @@ def parse(body)
...
-        return false unless %w(PENDING EXPIRED PAID).include?(response[:status])
-
-        response[:transactions].each do |transaction|
-          return false unless %w(PENDING CANCELLED SUCCESSFUL).include?(transaction.symbolize_keys[:status])
-        end
-
-        true
+        (response.is_a?(Hash) && response[:next_step]) ||
+          response == 204 ||
+          %w(PENDING EXPIRED PAID).include?(response[:status]) ||
```
gasb150, line 162: ❓ I want to be sure.
Mean this that, if the value is PENDING, EXPIRED, or PAID, or if all the response transactions have PENDING CANCELLED or SUCCESSFUL this will be a successful transaction?

EXPIRED, and CANCELLED are considered successful transactions?

gasb150 (commented):
LGTM!! Just a question.

lib/active_merchant/billing/gateways/sum_up.rb
sinourain: EXPIRED and in the transactions object PENDING and CANCELLED was removed, these states were necessary when canceling a transaction that had been left in a pending state, but this behavior is not supported by Spreedly

naashton (approved):
Just one question, otherwise looks good.

lib/active_merchant/billing/gateways/sum_up.rb
```diff
@@ -157,15 +157,10 @@ def parse(body)
...
-
-        return false unless %w(PENDING EXPIRED PAID).include?(response[:status])
-
-        response[:transactions].each do |transaction|
-          return false unless %w(PENDING CANCELLED SUCCESSFUL).include?(transaction.symbolize_keys[:status])
-        end
-
-        true
+        (response.is_a?(Hash) && response[:next_step]) ||
+          response == 204 ||
```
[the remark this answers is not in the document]
sinourain: I share the refund documentation where it mentions that the body comes empty and the code is just a code:

https://developer.sumup.com/online-payments/guides/refund#option-a-make-a-full-refund

PR #5088, opened 2024-04-09
Shift4v2: Adding bank account support

Heavyblade:
## Summary:
This PR enables bank account payment method on the Shift4v2

[SER-1143](https://spreedly.atlassian.net/browse/SER-1143)

### Remote Test:
Finished in 56.316066 seconds.
40 tests, 143 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### Unit Tests:
Finished in 36.406 seconds.
5842 tests, 79270 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4_v2.rb
+++ b/lib/active_merchant/billing/gateways/shift4_v2.rb
@@ -57,6 +57,37 @@ module ActiveMerchant #:nodoc:
         super
         post[:currency]&.upcase!
       end
+
+      def add_creditcard(post, payment_method, options)
+        return super unless payment_method.is_a?(Check)
+
+        post.merge!({
+          paymentMethod: {
+            type: :ach,
+            fraudCheckData: {
+              ipAddress: options[:ip],
+              email: options[:email]
+            }.compact_blank,
+            billing: {
+              name: payment_method.name,
+              address: { country: options.dig(:billing_address, :country) }
+            }.compact_blank,
+            ach: {
+              account: {
+                routingNumber: payment_method.routing_number,
+                accountNumber: payment_method.account_number,
+                accountType: get_account_type(payment_method)
+              },
+              verificationProvider: :external
+            }
+          }
+        })
+      end
+
+      def get_account_type(check)
+        holder = (check.account_holder_type || '').match(/business/i) ? :corporate : :personal
+        "#{holder}_#{check.account_type}"
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/shift4_v2.rb
```diff
@@ -57,6 +57,34 @@ def add_amount(post, money, options, include_currency = false)
...
         post[:currency]&.upcase!
       end
+
+      def add_creditcard(post, payment_method, options)
+        return super unless payment_method.is_a?(Check)
+
+        post.merge!({
+          paymentMethod: {
+            type: :ach,
+            fraudCheckData: { ipAddress: options[:ip] },
```
javierpedrozaing, line 67: based upon on the [documentation](https://dev.shift4.com/docs/api#payment-method-create_fraudCheckData) since we are sending `:external` in the `verificationProvider`, the `ipAddress` is not required, we can avoid to send it or add validation to ensure it is sent only if verificationProvider == 'none'.

javierpedrozaing (approved):
LGTM, just I left a small comment for your consideration.

lib/active_merchant/billing/gateways/shift4_v2.rb
Heavyblade: Yes, I tried, maybe is because the special change they made for us, if remove that param then I get this error:

Screenshot 2024-04-09 at 3 30 40 PM

gasb150: 💭 Opinion:
This doesn't mean it can accept another value, not necessarily the ipAddress, which could be the email.

gasb150 (commented):
LGTM, just complementing @javierpedrozaing comment.

lib/active_merchant/billing/gateways/shift4_v2.rb
```diff
@@ -57,6 +57,37 @@ def add_amount(post, money, options, include_currency = false)
         super
         post[:currency]&.upcase!
       end
+
+      def add_creditcard(post, payment_method, options)
```
naashton, line 61: If this method is for adding a bank account, can we change the method name from `add_creditcard` to `add_bank_account`?

naashton (approved):
LGTM!

lib/active_merchant/billing/gateways/shift4_v2.rb
Heavyblade: I had the same consideration, but I preferred to keep the method signature in the same way that SecurionPay does, so I can properly use the `super` keyword and be more explicit about the implementation

PR #5091, opened 2024-04-12
Cybersource: Update NT flow

almalee24:
Remote:
135 tests, 670 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.2963% passed

Unit:
142 tests, 782 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -141,9 +141,14 @@ module ActiveMerchant #:nodoc:
         r703: 'Export hostname_country/ip_country match'
       }
 
-      @@payment_solution = {
+      @@wallet_payment_solution = {
         apple_pay: '001',
         google_pay: '012'
+      }
+
+      NT_PAYMENT_SOLUTION = {
+        'master' => '014',
+        'visa' => '015'
       }
 
       # These are the options that can be used when creating a new CyberSource
@@ -281,6 +286,8 @@ module ActiveMerchant #:nodoc:
           gsub(%r((<cvNumber>)[^<]*(</cvNumber>))i, '\1[FILTERED]\2').
           gsub(%r((<cavv>)[^<]*(</cavv>))i, '\1[FILTERED]\2').
           gsub(%r((<xid>)[^<]*(</xid>))i, '\1[FILTERED]\2').
+          gsub(%r((<networkTokenCryptogram>)[^<]*(</networkTokenCryptogram>))i, '\1[FILTERED]\2').
+          gsub(%r((<requestorID>)[^<]*(</requestorID>))i, '\1[FILTERED]\2').
           gsub(%r((<authenticationData>)[^<]*(</authenticationData>))i, '\1[FILTERED]\2')
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -841,49 +840,18 @@ def subsequent_nt_apple_pay_auth(source, options)
...
+        when :visa, :master, :american_express
           xml.tag! 'ccAuthService', { 'run' => 'true' } do
-            xml.commerceIndicator commerce_indicator.nil? ? ECI_BRAND_MAPPING[brand] : commerce_indicator
-            xml.tag!('reconciliationID', options[:reconciliation_id]) if options[:reconciliation_id]
-          end
-        when :american_express
-          cryptogram = Base64.decode64(payment_method.payment_cryptogram)
-          xml.tag! 'ccAuthService', { 'run' => 'true' } do
-            xml.tag!('cavv', Base64.encode64(cryptogram[0...20]))
-            xml.tag!('commerceIndicator', ECI_BRAND_MAPPING[brand])
```
aenand, line 867: Why remove the commerce_indicator?

almalee24: Because it's optional and I decided just to remove it.

aenand: Hmmm my concern with removing it is that we needed to mark commerce_indicator as `internet` for recurring AP/GP recently. Do we know what the default commerce_indicator is?

almalee24: Okay, I'll add it back then.

aenand: Where did you see it was optional? We previously had a brand_mapping for this field so I'm not sure if that mapping was needed or if hard coding `internet` is correct

aenand (approved):
Thanks for the explanation of commerce indicator! Let's watch this one closely 👀

aenand (commented):
This looks great!! Is there any remote test(s) that can cover our expected change or do we already have remote tests for NT/AP/GP so it was more ensuring everything passes?

sinourain (approved):
LGTM, I agree with Amit and Dustin's request, the remote tests we have do not go through the `network_token` [condition](https://github.com/activemerchant/active_merchant/pull/5091/files#diff-ec0894a572ca922f933a687ef23008524682ccea87eec57f3175e7e1c4013bb2R701)

almalee24:
I've updated the NTs payment methods to include `source: :network_token` and I've also added @mastercard_network_token and they're all passing. I didn't add any ApplePay or GooglePay tests because this code doesn't affect those PMs and they logic for that will all stay the same.

aenand (approved):
Nice :shipit:

sinourain:
> I've updated the NTs payment methods to include `source: :network_token` and I've also added @mastercard_network_token and they're all passing. I didn't add any ApplePay or GooglePay tests because this code doesn't affect those PMs and the logic for that will all stay the same.

I ran the tests and was noticing that `payment_method.brand` returns `master` instead of `mastercard` as mapped, this can cause a problem

almalee24:
@aenand @sinourain @DustinHaefele I made a small update to make transactionType 3 for NTs https://github.com/activemerchant/active_merchant/pull/5091/files#diff-ec0894a572ca922f933a687ef23008524682ccea87eec57f3175e7e1c4013bb2L887

sinourain (approved):
LGTM!

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -835,14 +851,26 @@ def network_tokenization?(payment_method)
...
+        return unless @@wallet_payment_solution[source]
 
         options.dig(:stored_credential_overrides, :subsequent_auth) || options.dig(:stored_credential, :initiator) == 'merchant'
       end
 
       def add_auth_network_tokenization(xml, payment_method, options)
-        return unless network_tokenization?(payment_method)
+        brand = card_brand(payment_method).to_sym
+        case brand
+        when :visa, :master, :american_express
```
JesseBowling, line 862: The documentation calls out additional supported brand types [here](https://developer.cybersource.com/docs/cybs/en-us/google-pay/developer/fdiglobal/rest/googlepay/googpay-intro/googpay-processors.html). Is there a reason we're not also adding support for Diners Club, JCB, or Discover here?

PR #5093, opened 2024-04-15
Worldpay: Encyrpted ApplePay and GooglePay

almalee24:
Add support for encrypted ApplePay and GooglePay.

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -573,6 +573,8 @@ module ActiveMerchant #:nodoc:
           add_amount_for_pay_as_order(xml, amount, payment_method, options)
         when :network_token
           add_network_tokenization_card(xml, payment_method, options)
+        when :encrypted_wallet
+          add_encrypted_wallet(xml, payment_method, options)
         else
           add_card_or_token(xml, payment_method, options)
         end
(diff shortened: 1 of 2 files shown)


aenand (commented):
Is there a way to add a remote test?

almalee24:
> Is there a way to add a remote test?

No unfortunately

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -981,7 +1000,15 @@ def payment_details(payment_method, options = {})
         when String
           token_type_and_details(payment_method)
         else
-          type = network_token?(payment_method) || options[:wallet_type] == :google_pay ? :network_token : :credit
+          type = if network_token?(payment_method) || options[:wallet_type] == :google_pay
```
aenand, line 1003: Is there a way to not have nested if statements here? Can we put another case for `NetworkTokenizationCreditCard`?

almalee24: Yeah let me fix this

almalee24: I'll update it a bit but I can't change it to NetworkTokenizationCreditCard because network_token?(payment_method) was added by Shopify and I don't want to break an integration

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -981,7 +1000,11 @@ def payment_details(payment_method, options = {})
         when String
           token_type_and_details(payment_method)
         else
-          type = network_token?(payment_method) || options[:wallet_type] == :google_pay ? :network_token : :credit
+          type = if network_token?(payment_method)
+                   payment_method.payment_data ? :encrypted_wallet : :network_token
+                 else
+                   options[:wallet_type] == :google_pay ? :network_token : :credit
+                 end
```
aenand, line 1007: Do payment methods with options[:wallet_type] == :google_pay have payment_data? would this work?
```suggestion
          type = if network_token?(payment_method) || options[:wallet_type] == :google_pay 
                   payment_method.payment_data ? :encrypted_wallet : :network_token
                 else
                   :credit
                 end
```

almalee24: It doesn't. options[:wallet_type] == :google_pay is for a non-tokenized GooglePay which we treat as a CreditCard

PR #5094, opened 2024-04-16
Paymentez: Update success_from for refunds

almalee24:
Update success_from for refunds to consider
current_status CANCELLED and status_detail 34
to be successful

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -292,10 +292,12 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(response, action = nil)
+        successful_transaction = SUCCESS_STATUS.include?(response.dig('transaction', 'current_status') || response['status'])
+
         if action == 'refund'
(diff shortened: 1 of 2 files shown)


DustinHaefele (approved):
This looks good, and I appreciate the good test coverage. 

Do we have anyway to ensure there aren't other strange status combinations that should be marked successful?

almalee24:
> This looks good, and I appreciate the good test coverage.
> 
> Do we have anyway to ensure there aren't other strange status combinations that should be marked successful?
Thanks!

Not at the moment but we are in the process of setting up a call with Paymentez to figure out if there's any other combination that would be considered successful.

curiousepic:
Looks good - I do wonder if `cancelled` could always be considered a success, in which case it could just be added to the `SUCCESS_STATUS` constant.

almalee24:
> Looks good - I do wonder if `cancelled` would always be considered a success? In which case it could just be added to the `SUCCESS_STATUS` constant.

Yeah that could be a possibility for refunds. We are checking with Paymentez to make sure

PR #5097, opened 2024-04-16
StripePI: Update eci format

almalee24:
Update eci format to always be two digits.

Unit:
60 tests, 311 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
93 tests, 444 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -458,8 +458,18 @@ module ActiveMerchant #:nodoc:
         post[:payment_method_options][:card][:network_token] ||= {}
         post[:payment_method_options][:card][:network_token] = {
           cryptogram: payment_method.respond_to?(:payment_cryptogram) ? payment_method.payment_cryptogram : options[:cryptogram],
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -458,10 +458,20 @@ def add_cryptogram_and_eci(post, payment_method, options)
...
           cryptogram: payment_method.respond_to?(:payment_cryptogram) ? payment_method.payment_cryptogram : options[:cryptogram],
-          electronic_commerce_indicator: payment_method.respond_to?(:eci) ? payment_method.eci : options[:eci]
+          electronic_commerce_indicator: format_eci(payment_method, options)
         }.compact
       end
 
+      def format_eci(payment_method, options)
+        eci_value = payment_method.respond_to?(:eci) ? payment_method.eci : options[:eci]
+
+        if eci_value.length == 1
```
DustinHaefele, line 468:
```suggestion
        if eci_value&.length == 1
```

DustinHaefele (approved):
This looks good. I left one change to ensure we don't throw an error if `eci_value` is nil, but if you put that change in I approve!

PR #5098, opened 2024-04-16
Checkout V2: Retain and refresh OAuth access token

sinourain:
This changes contains:
- Retain OAuth access token to be used on subsequent transactions and refreshed when it expires

Spreedly reference:
[ECS-2996](https://spreedly.atlassian.net/browse/ECS-2996)

Unit:
66 tests, 403 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
2902.89 tests/s, 17725.19 assertions/s

Remote:
103 tests, 254 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
0.63 tests/s, 1.56 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -17,15 +17,7 @@ module ActiveMerchant #:nodoc:
       TEST_ACCESS_TOKEN_URL = 'https://access.sandbox.checkout.com/connect/token'
 
       def initialize(options = {})
-        @options = options
-        @access_token = options[:access_token] || nil
-
-        if options.has_key?(:secret_key)
-          requires!(options, :secret_key)
-        else
-          requires!(options, :client_id, :client_secret)
-          @access_token ||= setup_access_token
-        end
+        options.has_key?(:secret_key) ? requires!(options, :secret_key) : requires!(options, :client_id, :client_secret)
 
         super
       end
@@ -103,7 +95,8 @@ module ActiveMerchant #:nodoc:
           gsub(/("cvv\\":\\")\d+/, '\1[FILTERED]').
           gsub(/("cryptogram\\":\\")\w+/, '\1[FILTERED]').
           gsub(/(source\\":\{.*\\"token\\":\\")\d+/, '\1[FILTERED]').
-          gsub(/("token\\":\\")\w+/, '\1[FILTERED]')
+          gsub(/("token\\":\\")\w+/, '\1[FILTERED]').
+          gsub(/("access_token\\?"\s*:\s*\\?")[^"]*\w+/, '\1[FILTERED]')
       end
 
       def store(payment_method, options = {})
@@ -458,30 +451,43 @@ module ActiveMerchant #:nodoc:
         test? ? TEST_ACCESS_TOKEN_URL : LIVE_ACCESS_TOKEN_URL
       end
 
-      def setup_access_token
-        request = 'grant_type=client_credentials'
-        begin
-          raw_response = ssl_post(access_token_url, request, access_token_header)
-        rescue ResponseError => e
-          raise OAuthResponseError.new(e)
-        else
-          response = parse(raw_response)
-
-          if (access_token = response['access_token'])
-            access_token
-          else
-            raise OAuthResponseError.new(response)
-          end
-        end
+      def expires_date_with_extra_range(expires_in)
+        # Two minutes are subtracted from the expires_in time to generate the expires date
+        # in order to prevent any transaction from failing due to using an access_token
+        # that is very close to expiring.
+        # e.g. the access_token has one second left to expire and the lag when the transaction
+        # use an already expired access_token
+        (DateTime.now + (expires_in - 120).seconds).strftime('%Q').to_i
       end
 
-      def commit(action, post, options, authorization = nil, method = :post)
+      def setup_access_token
+        response = parse(ssl_post(access_token_url, 'grant_type=client_credentials', access_token_header))
+        @options[:access_token] = response['access_token']
+        @options[:expires] = expires_date_with_extra_range(response['expires_in']) if response['expires_in'] && response['expires_in'] > 0
+
+        Response.new(
+          access_token_valid?,
+          message_from(access_token_valid?, response, {}),
+          response.merge({ expires: @options[:expires] }),
+          test: test?,
+          error_code: error_code_from(access_token_valid?, response, {})
+        )
+      rescue ResponseError => e
+        raise OAuthResponseError.new(e)
+      end
+
+      def access_token_valid?
+        @options[:access_token].present? && @options[:expires].to_i > DateTime.now.strftime('%Q').to_i
+      end
+
+      def perform_request(action, post, options, authorization = nil, method = :post)
         begin
           raw_response = ssl_request(method, url(action, authorization), post.nil? || post.empty? ? nil : post.to_json, headers(action, options))
           response = parse(raw_response)
           response['id'] = response['_links']['payment']['href'].split('/')[-1] if action == :capture && response.key?('_links')
-          source_id = authorization if action == :unstore
         rescue ResponseError => e
+          @options[:access_token] = '' if e.response.code == '401' && !@options[:secret_key]
+
           raise unless e.response.code.to_s =~ /4\d\d/
 
           response = parse(e.response.body, error: e.response)
@@ -489,7 +495,14 @@ module ActiveMerchant #:nodoc:
 
         succeeded = success_from(action, response)
 
-        response(action, succeeded, response, options, source_id)
+        response(action, succeeded, response, options)
+      end
+
+      def commit(action, post, options, authorization = nil, method = :post)
+        MultiResponse.run do |r|
+          r.process { setup_access_token } unless @options[:secret_key] || access_token_valid?
+          r.process { perform_request(action, post, options, authorization, method) }
+        end
       end
 
       def response(action, succeeded, response, options = {}, source_id = nil)
@@ -508,7 +521,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def headers(action, options)
-        auth_token = @access_token ? "Bearer #{@access_token}" : @options[:secret_key]
+        auth_token = @options[:access_token] ? "Bearer #{@options[:access_token]}" : @options[:secret_key]
         auth_token = @options[:public_key] if action == :tokens
         headers = {
           'Authorization' => auth_token,
--- a/test/remote/gateways/remote_checkout_v2_test.rb
+++ b/test/remote/gateways/remote_checkout_v2_test.rb
@@ -1,3 +1,4 @@
+require 'timecop'
 require 'test_helper'
 
 class RemoteCheckoutV2Test < Test::Unit::TestCase
@@ -167,7 +168,7 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
     end
   end
 
-  def test_failed_purchase_with_failed_access_token
+  def test_failed_purchase_with_failed_oauth_credentials
     error = assert_raises(ActiveMerchant::OAuthResponseError) do
       gateway = CheckoutV2Gateway.new({ client_id: 'YOUR_CLIENT_ID', client_secret: 'YOUR_CLIENT_SECRET' })
       gateway.purchase(@amount, @credit_card, @options)
@@ -197,6 +198,7 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
     assert_scrubbed(declined_card.verification_value, transcript)
     assert_scrubbed(@gateway_oauth.options[:client_id], transcript)
     assert_scrubbed(@gateway_oauth.options[:client_secret], transcript)
+    assert_scrubbed(@gateway_oauth.options[:access_token], transcript)
   end
 
   def test_network_transaction_scrubbing
@@ -229,6 +231,53 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
     response = @gateway_oauth.purchase(@amount, @credit_card, @options)
     assert_success response
     assert_equal 'Succeeded', response.message
+  end
+
+  def test_successful_purchase_via_oauth_with_access_token
+    assert_nil @gateway_oauth.options[:access_token]
+    assert_nil @gateway_oauth.options[:expires]
+    purchase = @gateway_oauth.purchase(@amount, @credit_card, @options)
+    assert_success purchase
+    access_token = @gateway_oauth.options[:access_token]
+    expires = @gateway_oauth.options[:expires]
+    response = @gateway_oauth.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_equal @gateway_oauth.options[:access_token], access_token
+    assert_equal @gateway_oauth.options[:expires], expires
+  end
+
+  def test_failure_purchase_via_oauth_with_invalid_access_token_without_expires
+    @gateway_oauth.options[:access_token] = 'ABC123'
+    @gateway_oauth.options[:expires] = DateTime.now.strftime('%Q').to_i + 3600.seconds
+
+    response = @gateway_oauth.purchase(@amount, @credit_card, @options)
+    assert_failure response
+    assert_equal '401: Unauthorized', response.message
+    assert_equal @gateway_oauth.options[:access_token], ''
+  end
+
+  def test_successful_purchase_via_oauth_with_invalid_access_token_with_correct_expires
+    @gateway_oauth.options[:access_token] = 'ABC123'
+    response = @gateway_oauth.purchase(@amount, @credit_card, @options)
+    assert_success response
+    assert_not_equal 'ABC123', @gateway_oauth.options[:access_token]
+  end
+
+  def test_successful_purchase_with_an_expired_access_token
+    initial_access_token = @gateway_oauth.options[:access_token] = SecureRandom.alphanumeric(10)
+    initial_expires = @gateway_oauth.options[:expires] = DateTime.now.strftime('%Q').to_i
+
+    Timecop.freeze(DateTime.now + 1.hour) do
+      purchase = @gateway_oauth.purchase(@amount, @credit_card, @options)
+      assert_success purchase
+
+      assert_equal 2, purchase.responses.size
+      assert_not_equal initial_access_token, @gateway_oauth.options[:access_token]
+      assert_not_equal initial_expires, @gateway.options[:expires]
+
+      assert_not_nil purchase.responses.first.params['access_token']
+      assert_not_nil purchase.responses.first.params['expires']
+    end
   end
 
   def test_successful_purchase_with_vts_network_token
@@ -1005,21 +1054,16 @@ class RemoteCheckoutV2Test < Test::Unit::TestCase
 
   def test_successful_verify
     response = @gateway.verify(@credit_card, @options)
-    # this should only be a Response and not a MultiResponse
-    # as we are passing in a 0 amount and there should be
-    # no void call
-    assert_instance_of(Response, response)
-    refute_instance_of(MultiResponse, response)
     assert_success response
     assert_match %r{Succeeded}, response.message
   end
 
   def test_successful_verify_via_oauth
     response = @gateway_oauth.verify(@credit_card, @options)
-    assert_instance_of(Response, response)
-    refute_instance_of(MultiResponse, response)
     assert_success response
     assert_match %r{Succeeded}, response.message
+    assert_not_nil response.responses.first.params['access_token']
+    assert_not_nil response.responses.first.params['expires']
   end
 
   def test_failed_verify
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -508,7 +509,7 @@ def response(action, succeeded, response, options = {}, source_id = nil)
       end
 
       def headers(action, options)
-        auth_token = @access_token ? "Bearer #{@access_token}" : @options[:secret_key]
+        auth_token = "Bearer #{@options[:access_token] || @options[:secret_key]}"
```
aenand, line 512: This could be a breaking change as this is altering the authentication from `Basic` to `Bearer` auth for gateway tokens that only have a secret_key. Why are we making this change? Is it compatible to use Bearer auth with a secret_key?

sinourain: This change is because the following [guide](https://www.checkout.com/docs/previous/upgrade-to-our-new-platform/payments/test-new-platform-payments/full-card-api) and [this ticket](https://spreedly.atlassian.net/browse/ECS-3487) that described it, to continue supporting the `secret_key` in current merchants, I could add a condition that when the `secret_key` includes `sk_sbox_` add `Bearer` and otherwise do not include the `Bearer` way

sinourain: I have an assumption that the previous `secret_key` ends with a uuid v4, so I am going to upload my changes so that all the tests in CORE work and once I have confirmation if my assumption is correct or not I will update the code (if it is necessary) and mention it to continue with the review :)

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,46 +457,50 @@ def access_token_url
...
-          if (access_token = response['access_token'])
-            access_token
-          else
-            raise OAuthResponseError.new(response)
-          end
-        end
+        response = parse(ssl_post(access_token_url, 'grant_type=client_credentials', access_token_header))
+        @options[:access_token] = response['access_token']
+      rescue ResponseError => e
+        @multiresponse = MultiResponse.new
```
almalee24, line 465: Why add a MultiResponse here?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -508,8 +510,17 @@ def response(action, succeeded, response, options = {}, source_id = nil)
       end
 
       def headers(action, options)
-        auth_token = @access_token ? "Bearer #{@access_token}" : @options[:secret_key]
-        auth_token = @options[:public_key] if action == :tokens
+        auth_token =
```
almalee24, line 513: I would prefer if these was it's own PR like it was before. It would make it a lot easier to diagnose a problem if something goes wrong when it's deployed.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,46 +457,50 @@ def access_token_url
...
 
       def commit(action, post, options, authorization = nil, method = :post)
+        setup_access_token
         begin
           raw_response = ssl_request(method, url(action, authorization), post.nil? || post.empty? ? nil : post.to_json, headers(action, options))
           response = parse(raw_response)
           response['id'] = response['_links']['payment']['href'].split('/')[-1] if action == :capture && response.key?('_links')
-          source_id = authorization if action == :unstore
         rescue ResponseError => e
+          if e.response.to_s.match?(/Unauthorized/) && @options[:access_token] && action != :tokens
```
almalee24, line 476: This needs to be expanded to include other response errors such as Failed with 429 Too Many Requests. This is the main error we're trying to fix

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,46 +457,50 @@ def access_token_url
...
+            @options[:access_token] = nil
+            return commit(action, post, options, authorization, method)
```
almalee24, line 478: I'm a bit worried this could cause a constant loop if a transaction fails and it going through commit again and again.

almalee24: I also don't think Unauthorized is a good error message to retry since I think this is a clear indication that the credentials are incorrect. The error that brought this ticket is Failed with 429 Too Many Requests so I'm not sure if this is a better one

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: The initial idea was to obtain the response when the `access_token` is created so that we can debug better in case something fails and know what is related to the `access_token`, I moved this logic elsewhere and I think it looks better :)

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -1,11 +1,13 @@
+require 'pry'
```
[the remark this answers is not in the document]
sinourain: Yes, thanks!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: Sure, I will

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: I changed the implementation of this part, to do this I added a new field in the gateway options called `expires`, now we prevent the loop and we know better what the error is in case it occurs

lib/active_merchant/billing/gateways/checkout_v2.rb
almalee24: Please don't forget to make this into it's on PR

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,28 +457,30 @@ def access_token_url
...
+          test: test?,
+          error_code: error_code_from(false, response, {})
+        )
+      rescue ResponseError => e
+        Response.new(false, e.response.body.to_s, {}, test: test?)
       end
 
-      def commit(action, post, options, authorization = nil, method = :post)
+      def access_token_valid?
+        @options[:access_token] && @options.fetch(:expires, 0) > DateTime.now.strftime('%Q').to_i
```
almalee24, line 476: I would suggest making sure the the default value for `expires` is 0. Currently `@options.fetch(:expires, 0)` could be nil which cause a NoMethodError when checking if it's less than

almalee24: Oh it think that's what you are trying to do with fetch and make 0 the default but for some reason it's not working

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,28 +457,30 @@ def access_token_url
...
-        request = 'grant_type=client_credentials'
-        begin
-          raw_response = ssl_post(access_token_url, request, access_token_header)
-        rescue ResponseError => e
-          raise OAuthResponseError.new(e)
-        else
-          response = parse(raw_response)
+        response = parse(ssl_post(access_token_url, 'grant_type=client_credentials', access_token_header))
+        @options[:access_token] = response['access_token']
+        @options[:expires] = (DateTime.now + response.fetch('expires_in', 0).seconds).strftime('%Q').to_i
```
almalee24, line 462: Some thing for this fetch it needs to be updated since it's not going to default to 0

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,28 +457,30 @@ def access_token_url
...
-        end
+        Response.new(
+          response['access_token'].present?,
+          message_from(true, response, {}),
+          response,
+          test: test?,
+          error_code: error_code_from(false, response, {})
+        )
+      rescue ResponseError => e
+        Response.new(false, e.response.body.to_s, {}, test: test?)
```
almalee24, line 472: Considering the way that y'all wrote this PR think this could be an OAuthResponseError instead

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: I did it, it's just that in this same PR I added the changes from the ticket: ECS-3487 about Update Authorization from Basic to Bearer for Checkout gateway

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -489,16 +489,28 @@ def commit(action, post, options, authorization = nil, method = :post)
...
+
+      def response(action, succeeded, response, options = {})
         authorization = authorization_from(response) unless action == :unstore
         body = action == :unstore ? { response_code: response.to_s } : response
+
         Response.new(
           succeeded,
           message_from(succeeded, response, options),
-          body,
+          body.merge(access_token: @options[:access_token]),
```
almalee24, line 513: I think you'll also want to add expires to this so that we can access it in Core

lib/active_merchant/billing/gateways/checkout_v2.rb
almalee24: I know but it's better if these are in two separate PRs. One for retaining and refreshing the access_token and another for updating the authorization. Keeping this two changes separate will make it easier to diagnose if something goes wrong when we merge it.

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: Yes, I will :)

almalee24: Actually since you're doing a MultiResponse you don't need to this merge since you'll be able to access the access_token and expires_in in Core

curiousepic:
@almalee24 @sinourain I'd like to explore options other than tracking an expiration date before we merge this - will chat offline.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -489,16 +489,28 @@ def commit(action, post, options, authorization = nil, method = :post)
...
         succeeded = success_from(action, response)
 
-        response(action, succeeded, response, options, source_id)
+        response(action, succeeded, response, options)
       end
 
-      def response(action, succeeded, response, options = {}, source_id = nil)
+      def commit(action, post, options, authorization = nil, method = :post)
+        if @options[:secret_key] || access_token_valid?
+          peform_payin(action, post, options, authorization, method)
```
curiousepic, line 497: What will happen if the token is valid when this is executed, but due to communications lag, it expires before the gateway receives it? It will not attempt to refresh the token, and will just fail with an auth failure, right?

almalee24: The token is valid for 60mins so if there's a communication lag it would be a few minutes at most but yes if it expires it would fail with an auth failure. But if the customer is retries the transaction we would refresh the token.

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: ok, I will

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: I will change it to:

`@options[:access_token] && @options[:expires] && @options[:expires] > DateTime.now.strftime('%Q').to_i`

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: ok

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: Sure, I will

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: I removed these changes from this PR, they caused a lot of confusion and they will be in a new PR

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: I removed these changes from this PR, they caused a lot of confusion and they will be in a new PR

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: done!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: done!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: done!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: good catch, done!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: To prevent transactions from being made very close to the expiration and therefore failing due to lag, I was thinking that we could, for example, subtract a minute or two from the expiration time (`expires`) and with this we prevent there from being a lower failure rate. What do you think?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
+      rescue ResponseError => e
+        raise OAuthResponseError.new(e)
       end
 
-      def commit(action, post, options, authorization = nil, method = :post)
+      def access_token_valid?
+        @options[:access_token].present? && @options[:expires] && @options[:expires].to_i > DateTime.now.strftime('%Q').to_i
+      end
+
+      def peform_payin(action, post, options, authorization = nil, method = :post)
```
almalee24, line 474: I think this is a small typo but it should be perform_payin

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
         succeeded = success_from(action, response)
 
-        response(action, succeeded, response, options, source_id)
+        response(action, succeeded, response, options)
+      end
+
+      def commit(action, post, options, authorization = nil, method = :post)
+        MultiResponse.run do |r|
+          r.process { setup_access_token } unless @options[:secret_key] || access_token_valid?
+          r.process { peform_payin(action, post, options, authorization, method) }
```
almalee24, line 495: Same typo here

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
+          test: test?,
+          error_code: error_code_from(false, response, {})
+        )
+      rescue ResponseError => e
+        raise OAuthResponseError.new(e)
       end
 
-      def commit(action, post, options, authorization = nil, method = :post)
+      def access_token_valid?
+        @options[:access_token].present? && @options[:expires] && @options[:expires].to_i > DateTime.now.strftime('%Q').to_i
```
almalee24, line 471: Since you've defaulted expires to be 0 i don't think the extract @options[:expires] is necessary

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
+      end
+
+      def peform_payin(action, post, options, authorization = nil, method = :post)
         begin
           raw_response = ssl_request(method, url(action, authorization), post.nil? || post.empty? ? nil : post.to_json, headers(action, options))
           response = parse(raw_response)
           response['id'] = response['_links']['payment']['href'].split('/')[-1] if action == :capture && response.key?('_links')
-          source_id = authorization if action == :unstore
         rescue ResponseError => e
+          @options[:access_token] = '' if e.response.code == '401'
```
almalee24, line 480: I would suggest just to added check here for access_token in case the the transaction fails when using the secret_key

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: thanks!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: thanks!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: make sense

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: good catch!

sinourain: @almalee24 I'm adding a test for this but it doesn't make sense since the `access_token` takes precedence over `secret_key` according to this [line of code](https://github.com/activemerchant/active_merchant/pull/5098/files#diff-ac1e1ae60a6cc04111367e448e705b6e74104fd908e64d1ffe6c7237b5deb65dR515)

almalee24: Yeah I guess it's not a big issue but if secret_key transaction fails with 401 then we do `@options[:access_token] = ''` which we don't need to do because we didn't use access_token. It's not breaking anything so it's not something you need to change but a nice improvement I guess.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
 
-          if (access_token = response['access_token'])
-            access_token
-          else
-            raise OAuthResponseError.new(response)
-          end
-        end
+        Response.new(
+          response['access_token'].present?,
+          message_from(true, response, {}),
```
almalee24, line 461: Just another small issue is that this should only be true if  response['access_token'].present?

sinourain: Will I use `access_token_valid?` as a verifier since we assign parts of the `response` that allow us to validate the presence of the `access_token` and the `expires`

almalee24: Ok, you just also have to update `message_from(true, response, {})` to be `message_from(access_token_valid?, response, {})`

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
-          else
-            raise OAuthResponseError.new(response)
-          end
-        end
+        Response.new(
+          access_token_valid?,
+          message_from(true, response, {}),
+          response.merge({ expires: @options[:expires] }),
+          test: test?,
+          error_code: error_code_from(false, response, {})
```
almalee24, line 464: It would be the same here, it should be `error_code_from(access_token_valid?, response, {})`

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: done!

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: done!

almalee24 (approved):
Looks good to me. Thanks for all the hard work!

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
-        request = 'grant_type=client_credentials'
-        begin
-          raw_response = ssl_post(access_token_url, request, access_token_header)
-        rescue ResponseError => e
-          raise OAuthResponseError.new(e)
-        else
-          response = parse(raw_response)
+        response = parse(ssl_post(access_token_url, 'grant_type=client_credentials', access_token_header))
+        @options[:access_token] = response['access_token']
+        @options[:expires] = response['expires_in'] && response['expires_in'] > 0 && (DateTime.now + (response['expires_in'] - 120).seconds).strftime('%Q').to_i
```
curiousepic, line 457: This is confusing to me with the `&&` behavior with integers. Could you add a comment that explains the purpose?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -459,37 +452,48 @@ def access_token_url
...
+      rescue ResponseError => e
+        raise OAuthResponseError.new(e)
       end
 
-      def commit(action, post, options, authorization = nil, method = :post)
+      def access_token_valid?
+        @options[:access_token].present? && @options[:expires].to_i > DateTime.now.strftime('%Q').to_i
+      end
+
+      def perform_payin(action, post, options, authorization = nil, method = :post)
```
curiousepic, line 474: `payin` is an unusual term for AM - I'd suggest just `perform_request`

sinourain: ok, thanks!

curiousepic (approved):
Thanks for the tweaks!

PR #5103, opened 2024-04-22
Worldpay: Remove default ECI for NT

aenand:
ONE-76

Removes the default ECI value for NT on Worldpay. Per Worldpay's docs on ECI they state
"If you receive an eci value with the token we suggest including in the request". AM should not add a default as that goes against Worldpay's documentation

Docs: https://developerengine.fisglobal.com/apis/wpg/tokenisation/network-payment-tokens

Test Summary
Local:
5840 tests, 79184 assertions, 1 failures, 23 errors, 0 pendings, 0 omissions, 0 notifications 99.589% passed

Tests failing unrelated to this gateway
Remote:
103 tests, 444 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 98.0583% passed

2 errors also failing on master

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -153,6 +153,10 @@ module ActiveMerchant #:nodoc:
 
       private
 
+      def eci_value(payment_method)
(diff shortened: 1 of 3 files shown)


DustinHaefele (approved):
This makes sense to me, and is a pretty straight forward code change. And with remote testing passing I'm 👍.

PR #5105, opened 2024-04-23
Cybersource Rest: Support normalized 3DS data

aenand:
COMP-80

Adds support for 3DS2 data to the Cybersource Rest gateway. This follows the standard pattern in AM to add three_d_secure data and takes inspriation from the legacy Cybersource integration for 3DS.

Docs:https://developer.cybersource.com/api-reference-assets/index.html#payments

Test Summary
Local:
5848 tests, 79233 assertions, 0 failures, 23 errors, 0 pendings, 0 omissions, 0 notifications 99.6067% passed
Tests failing are unrelated to this change
Unit:
32 tests, 165 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Remote:
49 tests, 150 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 85.7143% passed
Failing on master too


lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -100,6 +100,30 @@ def scrub(transcript)
 
       private
 
+      def add_three_ds(post, payment_method, options)
+        return unless three_d_secure = options[:three_d_secure]
+
+        post[:consumerAuthenticationInformation] ||= {}
+        if payment_method.brand == 'master'
+          post[:consumerAuthenticationInformation][:ucafAuthenticationData] = three_d_secure[:cavv]
+          post[:consumerAuthenticationInformation][:ucafCollectionIndicator] = '2'
```
DustinHaefele, line 109: Why are we defaulting this field to `2` ?

DustinHaefele (commented):
I left one question inline, but I feel pretty good about this. 

Also, `consumerAuthenticationInformation` has a lot of possible properties that we aren't mapping. Do we feel pretty confident that the ones that we are are sufficient? It seems like this object contains properties for requesting 3ds2 as well as passing in authentication values, so a lot definitely don't apply.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
aenand: Great questions! This was taken from our legacy Cybersource integration as the default. 

In terms of the other values under `consumerAuthenticationInformation` Cybersource's docs don't indicate which are required/optional for our purposes. I tried to take the approach of mapping as many of our standard 3DS fields as possible and then cross-referencing with our legacy Cybersource integration to see what data we sent there.

DustinHaefele: Cool, with the passing remote tests I feel pretty good about it. I'll go ahead and 👍

DustinHaefele (approved):

PR #5106, opened 2024-04-23
Cybersource: Update NT flow

almalee24:
Update NT flow to send cryptogram in networkTokenCryptogram. If it's
a regular transaction commerceIndicator should be internet. If it's a
CIT or MIT commerceIndicator could be installment, recurring or internet
for unscheduled. subsequentAuthStoredCredential should be sent for all
subsequent transactions.

Remote:
136 tests, 676 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.8529% passed

Unit:
158 tests, 941 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -157,6 +157,7 @@
 * Braintree: Add additional data to response [aenand] #5084
 * CheckoutV2: Retain and refresh OAuth access token [sinourain] #5098
 * Worldpay: Remove default ECI value [aenand] #5103
+* CyberSource: Update NT flow [almalee24] #5106
 
 
 == Version 1.135.0 (August 24, 2023)
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -141,9 +141,14 @@ module ActiveMerchant #:nodoc:
         r703: 'Export hostname_country/ip_country match'
       }
 
-      @@payment_solution = {
+      @@wallet_payment_solution = {
         apple_pay: '001',
         google_pay: '012'
+      }
+
+      NT_PAYMENT_SOLUTION = {
+        'master' => '014',
+        'visa' => '015'
       }
 
       # These are the options that can be used when creating a new CyberSource
@@ -170,9 +175,15 @@ module ActiveMerchant #:nodoc:
         super
       end
 
-      def authorize(money, creditcard_or_reference, options = {})
-        setup_address_hash(options)
-        commit(build_auth_request(money, creditcard_or_reference, options), :authorize, money, options)
+      def authorize(money, payment_method, options = {})
+        if valid_payment_method?(payment_method)
+          setup_address_hash(options)
+          commit(build_auth_request(money, payment_method, options), :authorize, money, options)
+        else
+          # this is for NetworkToken, ApplePay or GooglePay brands that aren't supported at CyberSource
+          payment_type = payment_method.source.to_s.gsub('_', ' ').titleize.gsub(' ', '')
+          Response.new(false, "#{card_brand(payment_method).capitalize} is not supported by #{payment_type} at CyberSource, check https://developer.cybersource.com/docs/cybs/en-us/payments/developer/all/rest/payments/CreatingOnlineAuth/CreatingAuthReqPNT.html")
+        end
       end
 
       def capture(money, authorization, options = {})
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -821,11 +859,11 @@ def add_threeds_2_ucaf_data(xml, payment_method, options)
       def stored_credential_commerce_indicator(options)
         return unless options[:stored_credential]
 
-        return if options[:stored_credential][:initial_transaction]
-
         case options[:stored_credential][:reason_type]
-        when 'installment' then 'install'
```
aenand, line 827: I don't see anywhere in the docs indicating the keyword has changed from `install` to `installment`

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -884,9 +929,13 @@ def add_mastercard_network_tokenization_ucaf_data(xml, payment_method, options)
...
       end
 
-      def add_payment_network_token(xml)
+      def add_payment_network_token(xml, payment_method, options)
+        return unless network_tokenization?(payment_method)
+
+        transaction_type = payment_method.source == :network_token ? '3' : '1'
         xml.tag! 'paymentNetworkToken' do
-          xml.tag!('transactionType', '1')
+          xml.tag!('requestorID', options[:trid]) if transaction_type == '3' && options[:trid]
```
aenand, line 937: Is the `transaction_type == '3'` guard in case an AP/GP has a trid?

CHANGELOG
```diff
@@ -153,6 +153,7 @@
 * PayTrace: Always send name in billing_address [almalee24] #5086
 * StripePI: Update eci format [almalee24] #5097
 * Paymentez: Remove reference_id flag [almalee24] #5081
+* CheckoutV2: Update NT flow [almalee24] #5091
```
aenand, line 156: This changelog seems incorrect

aenand (changes requested):

lib/active_merchant/billing/gateways/cyber_source.rb
almalee24: Yes, just to be extra cautious but I guess they wouldn't have a trid to begin with

CHANGELOG
almalee24: Oh yes, I need to update

lib/active_merchant/billing/gateways/cyber_source.rb
almalee24: That's what I got back from the CyberSource contact. I'll forward you the email

aenand: ugh i hate when what they say does not match their docs.

almalee24: I emailed them again just to confirm because like you said it says install in docs. I'll keep you updated.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -170,19 +175,29 @@ def initialize(options = {})
...
-      def authorize(money, creditcard_or_reference, options = {})
-        setup_address_hash(options)
-        commit(build_auth_request(money, creditcard_or_reference, options), :authorize, money, options)
+      def authorize(money, payment_method, options = {})
+        if valid_payment_method?(payment_method)
+          setup_address_hash(options)
+          commit(build_auth_request(money, payment_method, options), :authorize, money, options)
+        else
+          # this is for NetworkToken brands that aren't supported at CyberSource
+          Response.new(false, "Payment method #{card_brand(payment_method)} is not supported, check https://developer.cybersource.com/docs/cybs/en-us/payments/developer/all/rest/payments/CreatingOnlineAuth/CreatingAuthReqPNT.html")
```
DustinHaefele, line 184: Should we clarify in this response message that it is for network tokens? Or will the return value from `card_brand(payment_method)` make that clear?

almalee24: That's a good idea, I'll update the error message since I've added this to include ApplePay and GooglePay. I decided to fix this since currently we're getting `The payment gateway settings are most likely invalid.` which is very misleading.

lib/active_merchant/billing/gateways/cyber_source.rb
almalee24: @aenand It's actually ` install` but I've decided to remove stored credentials from here and update it in a separate ticket to keep this as simple as possible.

DustinHaefele (approved):
LGTM, thanks for that messaging change!

PR #5108, opened 2024-04-25
FlexCharge: Adding support fot FlexCharge gateway

Heavyblade:
## Summary:
This PR adds support for FlexCharge gateway adding the evaluate and refund operations

[SER-1130](https://spreedly.atlassian.net/browse/SER-1130)

### Remote Test:
Finished in 39.651458 seconds.
12 tests, 40 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.6667% passed

*Note:*
Failing test happens because of account limit on refunds

### Unit Tests:
Finished in 33.453236 seconds.
5855 tests, 79334 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:
795 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/flex_charge.rb
+++ b/lib/active_merchant/billing/gateways/flex_charge.rb
@@ -1,0 +1,244 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class FlexChargeGateway < Gateway
+      self.test_url = 'https://api-sandbox.flex-charge.com/v1/'
+      self.live_url = 'https://api.flex-charge.com/v1/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover]
+      self.money_format = :cents
+      self.homepage_url = 'https://www.flex-charge.com/'
+      self.display_name = 'FlexCharge'
+
+      ENDPOINTS_MAPPING = {
+        authenticate: 'oauth2/token',
+        purchase: 'evaluate',
+        sync: 'outcome',
+        refund: 'orders/%s/refund'
+      }
+
+      SUCCESS_MESSAGES = %w(APPROVED CHALLENGE SUBMITTED SUCCESS).freeze
+
+      def initialize(options = {})
+        requires!(options, :app_key, :app_secret, :site_id, :mid)
+        super
+      end
+
+      def purchase(money, credit_card, options = {})
+        post = {}
+        address = options[:billing_address] || options[:address]
+        add_merchant_data(post, options)
+        add_base_data(post, options)
+        add_invoice(post, money, credit_card, options)
+        add_mit_data(post, options)
+        add_payment_method(post, credit_card, address, options)
+        add_address(post, credit_card, address)
+        add_customer_data(post, options)
+
+        commit(:purchase, post)
+      end
+
+      def refund(money, authorization, options = {})
+        commit(:refund, { amountToRefund: (money.to_f / 100).round(2) }, authorization)
+      end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Bearer )[a-zA-Z0-9._-]+)i, '\1[FILTERED]').
+          gsub(%r(("AppKey\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("AppSecret\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("accessToken\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("mid\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("siteId\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("environment\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("number\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,244 @@
...
+          cardBinNumber: credit_card.number[0..5],
+          cardLast4Digits: credit_card.number[-4..-1],
+          cardNumber: credit_card.number,
+          Token: false
+        }.compact
+      end
+
+      def address_names(address_name, payment_method)
+        split_names(address_name).tap do |names|
+          names[0] = payment_method&.first_name unless names[0].present?
```
gasb150, line 139: this can be simplified with something like this no?

```ruby
split_names(address_name).tap do |names|
  names[0] ||= payment_method&.first_name
  names[1] ||= payment_method&.last_name
end
```

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,244 @@
...
+          names[1] = payment_method&.last_name unless names[0].present?
```
gasb150, line 140: Why names[0] ant not names[1] ?

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,244 @@
...
+        end
+      end
+
+      def phone_from(options)
+        options[:phone] || options.dig(:billing_address, :phone_number)
+      end
+
+      def access_token_valid?
+        @options[:access_token].present? && @options.fetch(:expires, 0) > DateTime.now.strftime('%Q').to_i
```
gasb150, line 149: 💭 Opinion: this `expires` name could be ambitious because it's not clear, maybe expires_access_token could be an optional name.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,244 @@
...
+      end
+
+      def headers
+        { 'Content-Type' => 'application/json' }.tap do |headers|
+          headers['Authorization'] = "Bearer #{@options[:access_token]}" if @options[:access_token]
+        end
+      end
+
+      def parse(body)
+        JSON.parse(body).with_indifferent_access
```
gasb150, line 182: Why do you use `with_indifferent_access`, it's because you want to use `response[:symbol]` or some methods need to use strings in some cases and symbols in others.

gasb150 (commented):
@Heavyblade good job! I just left a few questions to clarify my understanding.

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,244 @@
+module ActiveMerchant #:nodoc:
+  module Billing #:nodoc:
+    class FlexChargeGateway < Gateway
+      self.test_url = 'https://api-sandbox.flex-charge.com/v1/'
+      self.live_url = 'https://api.flex-charge.com/v1/'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
```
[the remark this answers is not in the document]
Heavyblade: Ahh good point 👍

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,244 @@
...
+      private
+
+      def add_merchant_data(post, options)
+        post[:siteId] = @options[:site_id]
+        post[:mid] = @options[:mid]
+      end
+
+      def add_base_data(post, options)
+        post[:isDeclined] = cast_bool(options[:is_declined])
```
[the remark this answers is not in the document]
Heavyblade: `cast_bool` is meant to boleans that come from GSFS, so I wanted to go full defensive covering different cases.

lib/active_merchant/billing/gateways/flex_charge.rb
Heavyblade: Yup, that was my initial intention, but sadly with just on name for example "Doe", what `split_names`gives you is a ["", "Doe"]

lib/active_merchant/billing/gateways/flex_charge.rb
Heavyblade: I see your point, but this `expires` only lives on the @options hash, that represents the credentials and the access_token so given the context is not hard to infer, any way `token_expires` is a better name

lib/active_merchant/billing/gateways/flex_charge.rb
Heavyblade: Yes, I wanted to use symbols on my accessors but I know that as an standar CORE access the params with strings

lib/active_merchant/billing/gateways/flex_charge.rb
Heavyblade: Ohhh !!! a bad copy-paste, thanks

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -0,0 +1,244 @@
...
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Bearer )[a-zA-Z0-9._-]+)i, '\1[FILTERED]').
+          gsub(%r(("AppKey\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("AppSecret\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("accessToken\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("mid\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("siteId\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
+          gsub(%r(("environment\\?"\s*:\s*\\?")[^"]*)i, '\1[FILTERED]').
```
naashton, line 58: Do we need to scrub these?

PR #5109, opened 2024-04-26
Checkout V2: Update authorization from Basic to Bearer

sinourain:
Start sending requests using http Auth:Bearer

Spreedly reference:
[ECS-3487](https://spreedly.atlassian.net/browse/ECS-3487)

Unit:
66 tests, 403 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
2821.48 tests/s, 17228.11 assertions/s

Remote:
103 tests, 254 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
0.63 tests/s, 1.56 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -580,14 +580,16 @@ module ActiveMerchant #:nodoc:
       end
 
       def headers(action, options)
-        auth_token = @options[:access_token] ? "Bearer #{@options[:access_token]}" : @options[:secret_key]
-        auth_token = @options[:public_key] if action == :tokens
-        headers = {
-          'Authorization' => auth_token,
-          'Content-Type' => 'application/json;charset=UTF-8'
-        }
+        headers = { 'Authorization' => auth_token(action), 'Content-Type' => 'application/json;charset=UTF-8' }
         headers['Cko-Idempotency-Key'] = options[:idempotency_key] if options[:idempotency_key]
+
         headers
+      end
+
+      def auth_token(action)
+        return @options[:public_key] if action == :tokens
+
+        "Bearer #{@options[:access_token] || @options[:secret_key]}"
       end
 
       def tokenize(payment_method, options = {})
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -16,16 +16,10 @@ class CheckoutV2Gateway < Gateway
       LIVE_ACCESS_TOKEN_URL = 'https://access.checkout.com/connect/token'
       TEST_ACCESS_TOKEN_URL = 'https://access.sandbox.checkout.com/connect/token'
 
-      def initialize(options = {})
-        @options = options
-        @access_token = options[:access_token] || nil
+      BASIC_SECRET_KEY_FORMAT = /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i
```
almalee24, line 19: Did you find this key format in their docs?

almalee24, line 19: Did you get this format from their docs?

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -16,16 +16,10 @@ class CheckoutV2Gateway < Gateway
...
 
-        if options.has_key?(:secret_key)
-          requires!(options, :secret_key)
-        else
-          requires!(options, :client_id, :client_secret)
-          @access_token ||= setup_access_token
-        end
+      def initialize(options = {})
```
almalee24, line 21: I think you might have accidentally added all the work from the other PR.

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: No, this is an assumption that I have, I am waiting for confirmation from the Checkout.com team, I am going to ask if there is any progress on this topic

lib/active_merchant/billing/gateways/checkout_v2.rb
sinourain: Yes, I already updated this PR

lib/active_merchant/billing/gateways/checkout_v2.rb
almalee24: Okay, then let's waiting on this until we hear back. I'll move the ticket to waiting

sinourain: This format will no longer be necessary, I did some tests and now Checkout supports both formats with the 'Bearer' form

PR #5110, opened 2024-04-26
Datatrans: NT, AP, GP support

gasb150:
Summary:
-----
This PR includes for Datatrans the
NetworkToken, ApplePay and GooglePay code
to perform transactions with that PM
[SER-1195](https://spreedly.atlassian.net/browse/SER-1195)

Tests
-----
Remote Test:
Finished in 23.393965 seconds.
19 tests, 51 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications 100% passed

Unit Tests:
23 tests, 111 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

RuboCop:
795 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/datatrans.rb
+++ b/lib/active_merchant/billing/gateways/datatrans.rb
@@ -15,6 +15,16 @@ module ActiveMerchant #:nodoc:
       self.homepage_url = 'https://www.datatrans.ch/'
       self.display_name = 'Datatrans'
 
+      CREDIT_CARD_SOURCE = {
+        visa: 'VISA',
+        master: 'MASTERCARD'
+      }.with_indifferent_access
+
+      DEVICE_SOURCE = {
+        apple_pay: 'APPLE_PAY',
+        google_pay: 'GOOGLE_PAY'
+      }.with_indifferent_access
+
       def initialize(options = {})
         requires!(options, :merchant_id, :password)
         @merchant_id, @password = options.values_at(:merchant_id, :password)
@@ -26,8 +36,8 @@ module ActiveMerchant #:nodoc:
       end
 
       def authorize(money, payment, options = {})
-        post = add_payment_method(payment)
-        post[:refno] = options[:order_id].to_s if options[:order_id]
+        post = { refno: options.fetch(:order_id, '') }
+        add_payment_method(post, payment)
         add_currency_amount(post, money, options)
         add_billing_address(post, options)
         post[:autoSettle] = options[:auto_settle] if options[:auto_settle]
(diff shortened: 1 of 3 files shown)


test/unit/gateways/datatrans_test.rb
```diff
@@ -13,125 +13,148 @@ def setup
...
-        anything
-      ).
-      returns(successful_authorize_response)
-
-    response = @gateway.authorize(@amount, @credit_card, @options)
+    response = stub_comms(@gateway, :ssl_request) do
+      @gateway.authorize(@amount, @credit_card, @options)
+    end.check_request do |_action, endpoint, data, _headers|
+      assert_match('authorize', endpoint)
+      assert_match(/\"number\":\"(\d+{12})\"/, data)
```
Heavyblade, line 44: 💬 Comment:

`data` is actually a JSON, why not  just parse it and check the data you need instead of dealing with regex.

test/unit/gateways/datatrans_test.rb
```diff
@@ -13,125 +13,148 @@ def setup
...
+      assert_match(/\"refno\":(\d+)/, data)
+      assert_match('"currency":"CHF"', data)
+      assert_match('"amount":"100"', data)
+      assert_match('"name":"Jim Smith"', data)
+      assert_match('"street":"456 My Street"', data)
+      assert_match('"street2":"Apt 1"', data)
+      assert_match('"city":"Ottawa"', data)
```
Heavyblade, line 65: 💬 Comment

Same here

Heavyblade (commented):
@gasb150 looks pretty much ok, there some comments for your consideration

lib/active_merchant/billing/gateways/datatrans.rb
```diff
@@ -67,21 +77,38 @@ def scrub(transcript)
...
+        post[:card] = {
+          expiryMonth: format(payment_method.month, :two_digits),
+          expiryYear: format(payment_method.year, :two_digits)
+        }.merge(card)
+      end
+
+      def build_card(payment_method)
+        if payment_method.is_a?(NetworkTokenizationCreditCard)
+          type_nt, token_type = if DEVICE_SOURCE[payment_method.source]
+                                  ['DEVICE_TOKEN', DEVICE_SOURCE[payment_method.source]]
```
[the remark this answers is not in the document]
gasb150: It could be something different than the listed, I want to be sure there's no option to pass values set by default.

test/unit/gateways/datatrans_test.rb
gasb150: Oh right, I'll update all of the tests.

lib/active_merchant/billing/gateways/datatrans.rb
```diff
@@ -26,32 +36,32 @@ def purchase(money, payment, options = {})
...
         add_currency_amount(post, money, options)
         commit('credit', post, { transaction_id: transaction_id })
       end
 
       def void(authorization, options = {})
         post = {}
-        transaction_id, = authorization.split('|')
-        commit('cancel', post, { transaction_id: transaction_id, authorization: authorization })
+        transaction_id = authorization.split('|').first
+        commit('cancel', post, { transaction_id: transaction_id })
```
edgarv09, line 64: 💬 
Just wondering why authorization was be deleted from the commit method arguments.
Are why sure it should be removed?

edgarv09 (commented):
🦝

lib/active_merchant/billing/gateways/datatrans.rb
gasb150: Previous PR @Heavyblade clarify me, that I should not pass it.

The idea was: If Datatrans is not providing me an authorization value (or any related parameter), I want to have a value so, I will pass it from the previous authorization to the void and use it as a reference. 

The thing is we must not modify the gateway behavior if the gateway doesn't provide it we must not. This should be a reference to that specific transaction, (a void transaction), if the gateway is not returning one it's something the merchant must deal with.

We pass the transaction_id (a part of the authorization value) because it's required in the capture, and refund and void transactions.

edgarv09 (approved):
Nice work, LGTM 🐰

PR #5111, opened 2024-04-29
Shift4V2: Adding store for bank account and cc

Heavyblade:
## Summary:

This PR enables Shift4v2 to implement TPV on Bank accounts and adds support to store CreditCards directly on the 'cards' end-point

[SER-1219](https://spreedly.atlassian.net/browse/SER-1219)

### Remote Test:
Finished in 52.459288 seconds.
47 tests, 166 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### Unit Tests:
Finished in 40.844376 seconds.
5848 tests, 79304 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/shift4_v2.rb
+++ b/lib/active_merchant/billing/gateways/shift4_v2.rb
@@ -11,6 +11,28 @@ module ActiveMerchant #:nodoc:
         commit('credits', post, options)
       end
 
+      def store(payment_method, options = {})
+        post = case payment_method
+               when CreditCard
+                 cc = {}.tap { |card| add_creditcard(card, payment_method, options) }[:card]
+                 options[:customer_id].blank? ? { email: options[:email], card: cc } : cc
+               when Check
+                 bank_account_object(payment_method, options)
+               else
+                 raise ArgumentError.new("Unhandled payment method #{payment_method.class}.")
+               end
+
+        commit url_for_store(payment_method, options), post, options
+      end
+
+      def url_for_store(payment_method, options = {})
+        case payment_method
+        when CreditCard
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/shift4_v2.rb
```diff
@@ -11,6 +11,28 @@ def credit(money, payment, options = {})
...
+                 bank_account_object(payment_method, options)
+               else
+                 raise ArgumentError.new("Unhandled payment method #{payment_method.class}.")
+               end
+
+        commit url_for_store(payment_method, options), post, options
+      end
+
+      def url_for_store(payment_method, options = {})
+        case payment_method
```
javierpedrozaing, line 29: matter of taste maybe, but since we have a simple binary condition for payment_method we can use if.
```suggestion
   if payment_method.is_a?(CreditCard)
       options[:customer_id].blank?...
    elsif payment_method.is_a?(Check)
        'payment-methods'
    end
```

lib/active_merchant/billing/gateways/shift4_v2.rb
```diff
@@ -11,6 +11,28 @@ def credit(money, payment, options = {})
         commit('credits', post, options)
       end
 
+      def store(payment_method, options = {})
+        post = case payment_method
+               when CreditCard
+                 cc = {}.tap { |card| add_creditcard(card, payment_method, options) }[:card]
```
javierpedrozaing, line 17: this line is a little confuse, we could simplify it something like this
```suggestion
                 cc = add_creditcard({}, payment_method, options)[:card]
```

javierpedrozaing (commented):
LGTM, I left some minor comment for consideration

PR #5112, opened 2024-04-30
Update Stored Credentials for dlocal gateway

sinourain:
Spreedly reference:
[ECS-3492](https://spreedly.atlassian.net/browse/ECS-3492)

Unit tests

Finished in 27.620662 seconds.
5849 tests, 79313 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

RuboCop


lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -175,6 +175,7 @@ def add_card(post, card, action, options = {})
           end
           # required for MC debit recurrent in BR 'USED'(subsecuence Payments) . 'FIRST' an inital payment
           post[:card][:stored_credential_usage] = (options[:stored_credential][:initial_transaction] ? 'FIRST' : 'USED') if options[:stored_credential]
+          post[:card][:network_payment_reference] = options.dig(:stored_credential, :initiator) == 'merchant' && options[:stored_credential][:network_transaction_id]
```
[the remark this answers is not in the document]
almalee24: These stored credentials fields should also be moved outside of this if statement since we want to expand this to CreditCards as well.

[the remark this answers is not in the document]
sinourain: thanks!

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -189,6 +189,7 @@ def add_card(post, card, action, options = {})
         post[:card][:installments_id] = options[:installments_id] if options[:installments_id]
         post[:card][:force_type] = options[:force_type].to_s.upcase if options[:force_type]
         post[:card][:save] = options[:save] if options[:save]
+        post[:card][:network_payment_reference] =  options[:stored_credential][:network_transaction_id] if options.dig(:stored_credential, :initiator) == 'merchant'
```
almalee24, line 192: Please move out all those [stored credential fields that are inside the if statement for NetworkTokenization ](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/d_local.rb#L166) to here

almalee24: Could you also update to only send NTID if it's present

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -163,6 +163,7 @@ def add_card(post, card, action, options = {})
           post[:card][:network_token] = card.number
           post[:card][:cryptogram] = card.payment_cryptogram
           post[:card][:eci] = card.eci
+
           # used case of Network Token: 'CARD_ON_FILE', 'SUBSCRIPTION', 'UNSCHEDULED_CARD_ON_FILE'
```
almalee24, line 167: Sorry, I think I missed a few lines but we also need to move 167-176 to be inside like 182.

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -163,23 +163,28 @@ def add_card(post, card, action, options = {})
...
-            post[:card][:stored_credential_type] = 'SUBSCRIPTION'
-          end
-          # required for MC debit recurrent in BR 'USED'(subsecuence Payments) . 'FIRST' an inital payment
-          post[:card][:stored_credential_usage] = (options[:stored_credential][:initial_transaction] ? 'FIRST' : 'USED') if options[:stored_credential]
         else
           post[:card][:number] = card.number
           post[:card][:cvv] = card.verification_value
         end
 
+        # used case of Network Token: 'CARD_ON_FILE', 'SUBSCRIPTION', 'UNSCHEDULED_CARD_ON_FILE'
```
almalee24, line 171: Could you add

almalee24, line 171: Could you add lines 171-180 to be inside `if options[:stored_credential]`? Because we only want to be sending those fields is `options[:stored_credential]` is present

sinourain: Looks like `stored_credential_type` is an option [here](https://github.com/activemerchant/active_merchant/pull/5112/files#diff-794e2c4393901725c7a286d8cd01de14e51e7ad9eb80370e34d91a879ddff305L71) If I make this change I will no longer get this [result](https://github.com/activemerchant/active_merchant/pull/5112/files#diff-794e2c4393901725c7a286d8cd01de14e51e7ad9eb80370e34d91a879ddff305L73) it is ok?

almalee24: That's looks like an incorrect test to me because in DLocal `stored_credential_type` is a field we look for in options if it makes sense but it's a DLocal field we send in a request. 'SUBSCRIPTION' looks be like a default value for any transaction that uses NT which could be an issue

sinourain: Ok, thanks!

almalee24: Good thing you were able to catch that we were always sending  SUBSCRIPTION for all NT transactions even if they aren't using stored credentials. Please do move lines 171-180 to be inside if `options[:stored_credential]` because we don't want to have any issues with us accidentally marking transactions as subscription even when they aren't

sinourain: Done!

almalee24 (approved):

kylene-spreedly (approved):

PR #5114, opened 2024-05-01
Worldpay: Update stored creds

aenand:
COMP-42

Adds tests to ensure stored credentials are passed for network tokens and adds the ability to override the NTID from the standard framework

Test Summary
Local:
5867 tests, 79281 assertions, 0 failures, 23 errors, 0 pendings, 0 omissions, 0 notifications 99.608% passed
Unit:
117 tests, 662 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Remote:
104 tests, 447 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 98.0769% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -613,6 +613,7 @@ module ActiveMerchant #:nodoc:
             eci = eci_value(payment_method)
             xml.eciIndicator eci if eci.present?
           end
+          add_stored_credential_options(xml, options)
         end
       end
 
@@ -698,7 +699,7 @@ module ActiveMerchant #:nodoc:
         stored_credential_params = generate_stored_credential_params(is_initial_transaction, reason)
 
         xml.storedCredentials stored_credential_params do
-          xml.schemeTransactionIdentifier options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id] && !is_initial_transaction
+          xml.schemeTransactionIdentifier network_transaction_id(options) if network_transaction_id(options) && !is_initial_transaction
         end
       end
 
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -699,7 +699,7 @@ module ActiveMerchant #:nodoc:
         stored_credential_params = generate_stored_credential_params(is_initial_transaction, reason)
 
         xml.storedCredentials stored_credential_params do
-          xml.schemeTransactionIdentifier network_transaction_id(options) if network_transaction_id(options) && !is_initial_transaction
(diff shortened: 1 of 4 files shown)


aenand:
2 test failures also exist on master

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -698,7 +699,7 @@ def add_stored_credential_using_normalized_fields(xml, options)
         stored_credential_params = generate_stored_credential_params(is_initial_transaction, reason)
 
         xml.storedCredentials stored_credential_params do
-          xml.schemeTransactionIdentifier options[:stored_credential][:network_transaction_id] if options[:stored_credential][:network_transaction_id] && !is_initial_transaction
+          xml.schemeTransactionIdentifier network_transaction_id(options) if network_transaction_id(options) && !is_initial_transaction
```
almalee24, line 702: Considering this method is for fields in stored_credentials I think it's best to keep as it was before. There could be instance where the customer passed stored_credential_transaction_id and also has stored_credential.

aenand: So I updated this because if a merchant passes in a GSF for NTID it should take priority over the one we have on file. That's how we handle it on other gateways.

aenand: Are you worried this would be a breaking change?

almalee24: That's what also came to mind but I noticed we don't do that in WorldPay for Core. I think it's better to change it core to keep uniformity and I also noticed that other stored credentials GSF are available but we aren't updating those like stored_credential_usage and stored_credential_initiated_reason

aenand: In core? I don't think we should hide that logic somewhere else. I didn't add the other fields because if you're using the stored credential framework you wouldn't want to override the mappings but you may want to override the NTID if you brought one from somewhere else though I'm second guessing that now. 

Stripe PI does this in AM by first setting the `stored_credential` value then overriding with the `options` value if present. fwiw it seems we do this already for TPV on worldpay (not sure why TPV has ntids 🤷🏽 )

aenand: I removed this and just left the stored credential for NT change

test/unit/gateways/worldpay_test.rb
```diff
@@ -330,6 +330,21 @@ def test_authorize_passes_stored_credential_options
     assert_success response
   end
 
+  def test_authorize_with_nt_passes_stored_credential_options
```
almalee24, line 333: Could you add another test were these GSF are present but also the hash `stored_credential`? I know that the fields in stored_credential would be send in the request but just for posterity

aenand: Done!

almalee24 (approved):

PR #5116, opened 2024-05-06
Paymentez: Updates success_from method

almalee24:
Update success_from method to take current_status for the first message to evaluate.
If the transaction type is refund the  successful current_status base
on CANCELLED if status is also success

Remote
34 tests, 85 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit
33 tests, 137 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paymentez.rb
+++ b/lib/active_merchant/billing/gateways/paymentez.rb
@@ -291,10 +291,20 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(response, action = nil)
-        if action == 'refund'
-          response.dig('transaction', 'status_detail') == 7 || SUCCESS_STATUS.include?(response.dig('transaction', 'current_status') || response['status'])
+        transaction_current_status = response.dig('transaction', 'current_status')
+        request_status = response['status']
+        transaction_status = response.dig('transaction', 'status')
+        default_response = SUCCESS_STATUS.include?(transaction_current_status || request_status || transaction_status)
+
+        case action
+        when 'refund'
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/paymentez.rb
```diff
@@ -291,10 +291,21 @@ def headers
...
-          response.dig('transaction', 'status_detail') == 7 || SUCCESS_STATUS.include?(response.dig('transaction', 'current_status') || response['status'])
+        transaction_current_status = response.dig('transaction', 'current_status')
+        request_status = response['status']
+        transaction_status = response.dig('transaction', 'status')
+        default_response = SUCCESS_STATUS.include?(transaction_current_status || request_status || transaction_status)
+
+        case action
+        when 'refund'
+          successful_current_status = transaction_current_status&.upcase == 'CANCELLED' && request_status&.downcase == 'success'
+          successful_current_status || default_response
```
DustinHaefele, line 302: If `current_status = "APPROVED"` and `request_status = "failure"` it looks to me like this would be marked as successful because on line 297 success status would inclued "APPROVED" but my understanding is for a refund this combo should be failed. Is that right?

almalee24: Ah that's a good point, let me add some validations. I'm don't think that response should be returned but if it's is it's good to add some validations to prevent some false positives

DustinHaefele (approved):
LGTM

PR #5117, opened 2024-05-07
Add L2/L3 data for cybersource rest and worldpay

aenand:
Worldpay:
Refactor L2/L3 data for Worldpay to be more in line with how active merchant gateways expect this data. It also lowers the burden for what merchants must provide to use L2/L3 data
    
Test Summary
Local:
5883 tests, 79441 assertions, 0 failures, 23 errors, 0 pendings, 0 omissions, 0 notifications
99.609% passed
Unit:
117 tests, 668 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
103 tests, 444 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
98.0583% passed

Cybersource Rest
Adds support for L2 and L3 data to the Cybersource Rest
    
Test Summary
Local:
5882 tests, 79430 assertions, 0 failures, 23 errors, 0 pendings, 0 omissions, 0 notifications
99.609% passed
Unit:
36 tests, 189 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
Remote:
54 tests, 168 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
87.037% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -275,10 +275,10 @@ module ActiveMerchant #:nodoc:
 
         add_optional_data_level_two_and_three(xml, data)
 
-        data[:item].each { |item| add_items_into_level_three_data(xml, item.symbolize_keys, data) } if data.include?(:item)
+        data[:line_items].each { |item| add_line_items_into_level_three_data(xml, item.symbolize_keys, data) } if data.include?(:line_items)
       end
 
-      def add_items_into_level_three_data(xml, item, data)
+      def add_line_items_into_level_three_data(xml, item, data)
         xml.item do
           xml.description item[:description] if item[:description]
           xml.productCode item[:product_code] if item[:product_code]
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -306,8 +306,7 @@ module ActiveMerchant #:nodoc:
             add_amount(xml, sub_total_amount, data)
           end
           xml.itemTotalWithTax do
-            total_amount = item[:quantity].to_i * (item[:unit_cost].to_i - item[:discount_amount].to_i + item[:tax_amount].to_i)
-            add_amount(xml, total_amount, data)
+            add_amount(xml, item[:total_amount], data)
           end
           xml.itemDiscountAmount do
             add_amount(xml, item[:discount_amount], data)
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -260,9 +260,8 @@ module ActiveMerchant #:nodoc:
         xml.invoiceReferenceNumber data[:invoice_reference_number] if data.include?(:invoice_reference_number)
         xml.customerReference data[:customer_reference] if data.include?(:customer_reference)
         xml.cardAcceptorTaxId data[:card_acceptor_tax_id] if data.include?(:card_acceptor_tax_id)
-
         {
-          sales_tax: 'salesTax',
+          tax_amount: 'salesTax',
           discount_amount: 'discountAmount',
           shipping_amount: 'shippingAmount',
           duty_amount: 'dutyAmount'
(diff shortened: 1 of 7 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -256,63 +256,60 @@ def add_level_two_and_three_data(xml, amount, data)
...
-            data[:shipping_courier][:name]
-          )
-        end
+        # if data.include?(:shipping_courier)
+        #   xml.shippingCourier(
+        #     data[:shipping_courier][:priority],
+        #     data[:shipping_courier][:tracking_number],
+        #     data[:shipping_courier][:name]
+        #   )
+        # end
```
DustinHaefele, line 283: What is the reason that these are commented out?

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -256,63 +256,60 @@ def add_level_two_and_three_data(xml, amount, data)
         xml.invoiceReferenceNumber data[:invoice_reference_number] if data.include?(:invoice_reference_number)
         xml.customerReference data[:customer_reference] if data.include?(:customer_reference)
         xml.cardAcceptorTaxId data[:card_acceptor_tax_id] if data.include?(:card_acceptor_tax_id)
-
         {
-          sales_tax: 'salesTax',
+          tax_amount: 'salesTax',
```
DustinHaefele, line 260: Has this field been misnamed the whole time?

DustinHaefele (commented):
Couple questions.

lib/active_merchant/billing/gateways/worldpay.rb
aenand: I was testing to see if these are actually a thing. They're not in WP's docs so I think I'm going to delete them.

lib/active_merchant/billing/gateways/worldpay.rb
aenand: So these fields made it into AM but never into core. With that in mind, I'm taking the time to normalize their approach a bit better so that it's similar to other gateways.

lib/active_merchant/billing/gateways/worldpay.rb
curiousepic: @aenand were you going to go ahead and remove these?

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -256,63 +256,59 @@ def add_level_two_and_three_data(xml, amount, data)
...
+        # end
 
         add_optional_data_level_two_and_three(xml, data)
 
-        if data.include?(:item) && data[:item].kind_of?(Array)
-          data[:item].each { |item| add_items_into_level_three_data(xml, item.symbolize_keys) }
-        elsif data.include?(:item)
-          add_items_into_level_three_data(xml, data[:item].symbolize_keys)
-        end
+        data[:item].each { |item| add_items_into_level_three_data(xml, item.symbolize_keys, data) } if data.include?(:item)
```
travismorrison, line 274: This implies that where `data.item` was once an optional hash OR array, it is now a required array. Is it so?
Also, does this possibly mean the key can be changed to `:items`? I've been this way before and have sad memories of the time I spent staring at `data[:item].kind_of?(Array)` like _derrrr it's just item, not items_

lib/active_merchant/billing/gateways/worldpay.rb
curiousepic: Actually I'm half of the mind to leave them as-is; it wouldn't be the first time an undocumented field was functional and utilized. It's a bit of a gray area, but I'm on board with either.

aenand: Ah I forgot to. Yeah I'm going to delete them

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -256,63 +256,59 @@ def add_level_two_and_three_data(xml, amount, data)
...
           xml.quantity item[:quantity] if item[:quantity]
-
-          {
-            unit_cost: 'unitCost',
-            item_total: 'itemTotal',
-            item_total_with_tax: 'itemTotalWithTax',
-            item_discount_amount: 'itemDiscountAmount',
-            tax_amount: 'taxAmount'
-          }.each do |key, tag|
-            next unless item.include?(key)
```
travismorrison, line 310: The old "each on a hardcoded hash" strategy checked to see if the param existed in `item`, at this `next` clause. That check appears to have been removed? is that cool?

Not for nothing but--while I see why you flattened the each out (there are two XML nodes that require special code) it's a little verbose. What would you think of keeping the original implementation for unitCost, itemTotalWithTax, itemDiscountAmount, and taxAmount... with less functionality, like I like how you moved `to_i` to the function that needs it... and then having special ones for `itemTotal` and `unitOfMeasure`

travismorrison (commented):
little bit of a stranger round here... comment/Qs

lib/active_merchant/billing/gateways/worldpay.rb
aenand: Yeah I don't think it makes sense to treat line_items as a hash if there's only one item. I can update it to expect `line_items` to be more clear

lib/active_merchant/billing/gateways/worldpay.rb
aenand: I had to change this because Worldpay is particular about the order in which the xml attributes are added and i kept having to interrupt the hash to get the order "right"

DustinHaefele (approved):
This looks good to me.

PR #5120, opened 2024-05-14
Flex Charge: Add support for TPV store

edgarv09:
Flex Charge: Add support for the TPV store

Test summary:
Local:
5898 tests, 79569 assertions, 0 failures, 17 errors, 0 pendings, 0 omissions, 0 notifications 99.7118% passed
Unit:
12 tests, 58 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
Remote:
13 tests, 34 assertions, 1 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications 91.6667% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,7 +2,6 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
-* FlexCharge: Add support for TPV store [edgarv09] #5120
 * Shift4V2: Add new gateway based on SecurionPay adapter [heavyblade] #4860
 * TNS: Use the specified order_id in request if available [yunnydang] #4880
 * Cybersource: Support recurring apple pay [aenand] #4874
@@ -164,7 +163,7 @@
 * Cybersource Rest: Add support for network tokens [aenand] #5107
 * Decidir: Add support for customer object [rachelkirk] #5071
 * Worldpay: Add support for stored credentials with network tokens [aenand] #5114
-
+* FlexCharge: Add support for TPV store [edgarv09] #5120
 
 == Version 1.135.0 (August 24, 2023)
 * PaymentExpress: Correct endpoints [steveh] #4827
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* FlexCharge: Add support for TPV store [edgarv09] #5120
 * Shift4V2: Add new gateway based on SecurionPay adapter [heavyblade] #4860
 * TNS: Use the specified order_id in request if available [yunnydang] #4880
 * Cybersource: Support recurring apple pay [aenand] #4874
--- a/lib/active_merchant/billing/gateways/flex_charge.rb
+++ b/lib/active_merchant/billing/gateways/flex_charge.rb
@@ -15,7 +15,8 @@ module ActiveMerchant #:nodoc:
         authenticate: 'oauth2/token',
         purchase: 'evaluate',
         sync: 'outcome',
-        refund: 'orders/%s/refund'
+        refund: 'orders/%s/refund',
+        store: 'tokenize'
       }
 
       SUCCESS_MESSAGES = %w(APPROVED CHALLENGE SUBMITTED SUCCESS).freeze
@@ -41,6 +42,25 @@ module ActiveMerchant #:nodoc:
 
       def refund(money, authorization, options = {})
         commit(:refund, { amountToRefund: (money.to_f / 100).round(2) }, authorization)
+      end
+
+      def store(credit_card, options = {})
+        address = options[:billing_address] || options[:address] || {}
+        first_name, last_name = address_names(address[:name], credit_card)
+
+        post = {
+          payment_method: {
+            credit_card: {
+              first_name: first_name,
+              last_name: last_name,
+              month: credit_card.month,
+              year: credit_card.year,
+              number: credit_card.number,
+              verification_value: credit_card.verification_value
+            }.compact
+          }
+        }
+        commit(:store, post)
       end
 
       def supports_scrubbing?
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -115,23 +135,29 @@ def add_invoice(post, money, credit_card, options)
           avsResultCode: options[:avs_result_code],
           cvvResultCode: options[:cvv_result_code],
           cavvResultCode: options[:cavv_result_code],
-          cardNotPresent: credit_card.verification_value.blank?
+          cardNotPresent: credit_card.is_a?(String) ? false : credit_card.verification_value.blank?
         }.compact
       end
 
       def add_payment_method(post, credit_card, address, options)
```
gasb150, line 142: The credit_card name could be something more generic, like payment_method considering this is not always a credit_card

test/unit/gateways/flex_charge_test.rb
```diff
@@ -153,6 +153,15 @@ def test_address_names_from_credit_card
     assert_equal 'Doe', names.last
   end
 
+  def test_successful_store
+    response = stub_comms do
+      @gateway.store(@credit_card, @options)
```
gasb150, line 158: You can check the request construction with the check_request.

something like 
```ruby
 response = stub_comms do
      @gateway.store(@credit_card, @options)
      end.check_request do |endpoint, data, _headers|
      if endpoint == endpoint_for_access_token_response
       #some assertions
       elsif endpoint == enpoint_for_store_response
       # some assertions
       end
    end.respond_with(successful_access_token_response, successful_store_response)
```

Or even ignore the responses if this approach with two responses is creating some issues.

test/remote/gateways/remote_flex_charge_test.rb
```diff
@@ -147,6 +147,14 @@ def test_failed_fetch_access_token
     assert_match(/400/, error.message)
   end
 
+  def test_successful_purchase_with_token
+    store = @gateway.store(@credit_card_cit, {})
+    assert_success store
+
```
gasb150, line 153: You could include assertions representing the store transaction (like a typical response field). (If it's something possible)

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -115,23 +135,29 @@ def add_invoice(post, money, credit_card, options)
           avsResultCode: options[:avs_result_code],
           cvvResultCode: options[:cvv_result_code],
           cavvResultCode: options[:cavv_result_code],
-          cardNotPresent: credit_card.verification_value.blank?
+          cardNotPresent: credit_card.is_a?(String) ? false : credit_card.verification_value.blank?
```
gasb150, line 138: Just confirming:  for TPV the cardNotPresent value will always be false.

But for credit cards, will it be something true if no verification value is provided?

gasb150 (commented):
@edgarv09 It's pretty good. I left a few comments to your consideration.

test/unit/gateways/flex_charge_test.rb
edgarv09: The validations you propose are already carried out in other tests, from my point of view in this case they do not add extra value.

test/remote/gateways/remote_flex_charge_test.rb
edgarv09: The only thing in the response that allows us to identify is a store. It's the token and that would be used to set the authorization parameter of the response. So, if this field is not present, blank, or wrong the purchase will fail.

lib/active_merchant/billing/gateways/flex_charge.rb
edgarv09: Flex charge Documentation is not clear on this topic but the messages during the testing led me to this conclusion.

lib/active_merchant/billing/gateways/flex_charge.rb
edgarv09: I saw that, but @Heavyblade is still working on flex charge so changing this argument led to merging problems that we are not willing to deal with at this moment.

test/unit/gateways/flex_charge_test.rb
gasb150: So the main purpose of this test is the authorization value generated?

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -228,8 +254,8 @@ def message_from(response)
         response[:title] || response[:responseMessage] || response[:status]
       end
 
-      def authorization_from(response)
-        response[:orderSessionKey]
+      def authorization_from(action, response)
+        action == :store ? response[:transaction][:payment_method][:token] : response[:orderSessionKey]
```
Heavyblade, line 258: 💬 Comment:

Just in case of unexpected scenarios `dig` is a safer way to get the token

Heavyblade (commented):
@edgarv09 looks pretty much ok, left a couple comments for your consideration

lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -115,23 +135,29 @@ def add_invoice(post, money, credit_card, options)
...
-          expirationMonth: credit_card.month,
-          expirationYear: credit_card.year,
-          cardBinNumber: credit_card.number[0..5],
-          cardLast4Digits: credit_card.number[-4..-1],
-          cardNumber: credit_card.number,
-          Token: false
-        }.compact
+        payment_method = case credit_card
+                         when String
+                           { token: true, cardNumber: credit_card }
```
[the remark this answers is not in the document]
edgarv09: It seems to work with lower/upper case.

CHANGELOG
```diff
@@ -2,6 +2,7 @@
 = ActiveMerchant CHANGELOG
 
 == HEAD
+* FlexCharge: Add support for TPV store [edgarv09] #5120
```
Heavyblade, line 5: 💬 Comment, 

This should be a the bottom of the block

Heavyblade (approved):

PR #5123, opened 2024-05-17
Orbital: Add support for L2 and L3 fields, update XSD file

rachelkirk:
A bit of refactoring was needed in the gateway file to support these fields and get things in better shape with adhering to the XSD. Unit tests are now more comprehensive to check schema. I added a remote test and left comments to indicate which fields are restricted to Canadian merchants.

Unit Tests:
148 tests, 850 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests:
134 tests, 504 assertions, 8 failures, 11 errors, 0 pendings, 0 omissions, 0 notifications 85.8209% passed
*Same number of failures and errors on master.

Local Tests:
5899 tests, 79647 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


test/unit/gateways/orbital_test.rb
```diff
@@ -1953,4 +1971,11 @@ def post_scrubbed_echeck
       Conn close
     REQUEST
   end
+
+  def assert_xml_valid_to_xsd(data)
```
jcreiff, line 1975: nice bit of refactoring here 🌟

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -1301,6 +1321,13 @@ def test_successful_purchase_with_level_2_data
     assert_equal 'Approved', response.message
   end
 
+  def test_successful_purchase_with_level_2_data_canadian_currency
+    response = @tandem_gateway.purchase(@amount, @credit_card, @options.merge(currency: 'CAD', merchant_vat_reg_number: '987654321', national_tax: '625', level_2_data: @level_2_options))
```
jcreiff, line 1325: Small q: should the `merchant_vat_reg_number` and `national_tax` be nested in the `level_2_data`? It doesn't look like we access them from top level `options`

It looks like that's what you did up in lines 54-62; although I'm not quite clear on the difference between that version of `level_2_options` and the one used in this section 🤔

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -1232,7 +1243,13 @@ def setup
       tax_indicator: '1',
       tax: '75',
       purchase_order: '123abc',
-      zip: address[:zip]
+      zip: address[:zip],
```
rachelkirk, line 1246: It's not super clear by looking at the diff, but these are in a separate Tandem Class that uses different credentials. When adding all fields to level_2_options, I got errors about fields being restricted to Canadian merchants which is the reason for the new remote test with that currency and those fields are only added in that test.

jcreiff (commented):
😮‍💨 great work parsing through the XSD; I tried to squint at it and find anything that might have been overlooked, but I think it's looking good!

Left one small comment about one of the tests.

I think my larger question is about whether or not updating the XSD that we are testing against means that we also need to update the `API_VERSION` declared at the top of the gateway model. I notice that in the previous commits that changed the XSD, there was also a change to that line.

https://github.com/activemerchant/active_merchant/pull/3850/files
https://github.com/activemerchant/active_merchant/pull/3117/files

test/remote/gateways/remote_orbital_test.rb
rachelkirk: These fields should only be used by Canadian merchants so they are not included in the main options hash.

DeeMeyers (approved):
LGTM!

test/unit/gateways/orbital_test.rb
```diff
@@ -269,6 +291,9 @@ def test_network_tokenization_credit_card_data
 
   def test_schema_for_soft_descriptors_with_network_tokenization_credit_card_data
```
rachelkirk, line 292: @jknipp I updated this unit test to bring in the fields with L2/L3 data. Refactored the assertion to ensure it is valid to XSD instead of just matching specific fields.

rachelkirk: This test could probably be renamed since it's no longer specific to network tokenization and soft descriptors

test/unit/gateways/orbital_test.rb
```diff
@@ -1953,4 +2000,11 @@ def post_scrubbed_echeck
       Conn close
     REQUEST
   end
+
+  def assert_xml_valid_to_xsd(data)
+    schema_file = File.read("#{File.dirname(__FILE__)}/../../schema/orbital/Request_PTI95.xsd")
+    doc = Nokogiri::XML(data)
+    xsd = Nokogiri::XML::Schema(schema_file)
```
jknipp, line 2007: Test improvement suggestion here, can we memoize the reading/loading of the XSD so that it is only done once for the whole test file? This will speed up tests.

rachelkirk:
> @rachelkirk I think this looks fine? I'd prefer another set of 👀 from a member of Pathfinder to be doubly sure on the specific field modifications.

@DeeMeyers reviewed it earlier but I can reach out to their team again if you meant you wanted more than one Pathfinder team member to review.

test/unit/gateways/orbital_test.rb
rachelkirk: Done!

jcreiff (approved):
Thanks for adding the additional xsd assertions!

PR #5127, opened 2024-05-20
WorldPay: Add support for encrypted ApplePay and GooglePay

almalee24:


aenand (commented):
Is it possible to add remote tests? Also can you post the results of running remote tests?

almalee24:
> Is it possible to add remote tests? Also can you post the results of running remote tests?

Nope it's not possible since we can't test this remotely. The customer will be testing this for us

PR #5133, opened 2024-05-23
Update the error_code_from method to grab and alpha_numeric characters

DustinHaefele:
Spreedly ref: ECS-3536


aenand (approved):
Works on Rubular :shipit:

PR #5134, opened 2024-05-24
NMI: add NTID override

yunnydang:
This adds an optional NTID ovveride to the stored credential options

Local:
5914 tests, 79717 assertions, 0 failures, 7 errors, 0 pendings, 0 omissions, 0 notifications
99.8816% passed

Unit:
57 tests, 461 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
52 tests, 189 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.2308% passed


test/remote/gateways/remote_nmi_test.rb
```diff
@@ -388,6 +388,18 @@ def test_purchase_using_stored_credential_recurring_mit
     assert_success purchase
   end
 
+  def test_purchase_using_ntid_override_mit
+    initial_options = stored_credential_options(:merchant, :recurring, :initial)
```
jcreiff, line 392: I think to make this test more true to a typical sequence, this initial txn should have `:cardholder` instead of `:merchant`

test/remote/gateways/remote_nmi_test.rb
```diff
@@ -388,6 +388,18 @@ def test_purchase_using_stored_credential_recurring_mit
...
+    assert purchase = @gateway.purchase(@amount, @credit_card, initial_options)
+    assert_success purchase
+    assert network_transaction_id = purchase.params['transactionid']
+
+    @options[:network_transaction_id] = purchase.params['transactionid']
```
jcreiff, line 397: You could use the `network_transaction_id` variable you just set a couple lines above this 😁

jcreiff (commented):
Just a couple of small suggestions! Going to pull this down and run the tests as well 🧪

test/remote/gateways/remote_nmi_test.rb
yunnydang: Thanks @jcreiff I agree with your suggestions and pushed the updates, thanks!

PR #5135, opened 2024-05-24
Litle: Add 141 and 142 as successful responses

almalee24:
141 and 142 are successful responses for prepaid cards.

Unit:
61 tests, 274 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
57 tests, 250 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 98.2456% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/litle.rb
+++ b/lib/active_merchant/billing/gateways/litle.rb
@@ -561,7 +561,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def success_from(kind, parsed)
-        return %w(000 001 010).any?(parsed[:response]) unless kind == :registerToken
+        return %w(000 001 010 141 142).any?(parsed[:response]) unless kind == :registerToken
 
         %w(000 801 802).include?(parsed[:response])
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/litle.rb
```diff
@@ -561,7 +561,7 @@ def commit(kind, request, money = nil)
       end
 
       def success_from(kind, parsed)
-        return %w(000 001 010).any?(parsed[:response]) unless kind == :registerToken
+        return %w(000 001 010 141 142).any?(parsed[:response]) unless kind == :registerToken
```
aenand, line 564: Do you have some documentation to support this?

almalee24: Nope, the documentation we have doesn't mention these two response codes. We do have example transactions and they literally do say Approved at the end of the response message. If they are failed it says Soft Decline at the end and the response code for that is 138.

aenand: I remember some gateway as mentioning one message as "Approved" but the transaction code indicated it failed, do we think this gateway might be the same?

almalee24: Not in this case, the customer did reach out to Litle and they confirmed that those codes would be considered successful.

aenand (approved):
LGTM

curiousepic (approved):

PR #5136, opened 2024-05-29
WIP => CyberSource: Update Stored Credential flow

sinourain:
Spreedly reference:
[ECS-3532](https://spreedly.atlassian.net/browse/ECS-3532)

Unit tests

Finished in 26.745895 seconds.
5912 tests, 79756 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

RuboCop

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -354,11 +354,13 @@ module ActiveMerchant #:nodoc:
       end
 
       def build_auth_request(money, creditcard_or_reference, options)
+        card_brand = card_brand(creditcard_or_reference).to_sym unless creditcard_or_reference.is_a?(String)
+
         xml = Builder::XmlMarkup.new indent: 2
         add_customer_id(xml, options)
         add_payment_method_or_subscription(xml, money, creditcard_or_reference, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -862,11 +866,21 @@ def add_threeds_2_ucaf_data(xml, payment_method, options)
       def stored_credential_commerce_indicator(options)
         return unless options[:stored_credential]
 
-        return if options[:stored_credential][:initial_transaction]
+        if options[:stored_credential][:initial_transaction]
```
almalee24, line 869: I think it's best just to remove lines 869-875  since it doesn't work and we haven't heard back from Cybersource.

bradbroge (approved):
Looks good. 
I do see something in the ticket about `recurring_internet` as a commerce_indicator, but I don't see it in the code.  Is that something that we need to support?

almalee24:
> Looks good. I do see something in the ticket about `recurring_internet` as a commerce_indicator, but I don't see it in the code. Is that something that we need to support?

I ended up removing that because it wasn't working when testing remotely and I didn't hear back from the CyberSource on why this was failing.

PR #5141, opened 2024-06-04
EXAMPLE Cybersource discover NT/Apple Pay

klumsy:
to support Discover. as it seems the requirements
per cyber_source api docs (to be verified) seem to match those of VISA.

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -323,7 +323,7 @@ module ActiveMerchant #:nodoc:
       def valid_payment_method?(payment_method)
         return true unless payment_method.is_a?(NetworkTokenizationCreditCard)
 
-        %w(visa master american_express).include?(card_brand(payment_method))
+        %w(visa master american_express discover).include?(card_brand(payment_method))
       end
 
       # Create all required address hash key value pairs
@@ -895,7 +895,7 @@ module ActiveMerchant #:nodoc:
         brand = card_brand(payment_method).to_sym
 
         case brand
-        when :visa
+        when :visa, :discover
           xml.tag! 'ccAuthService', { 'run' => 'true' } do
             xml.tag!('cavv', payment_method.payment_cryptogram) unless commerce_indicator
             xml.commerceIndicator commerce_indicator.nil? ? ECI_BRAND_MAPPING[brand] : commerce_indicator


aenand:
Hi @klumsy do you have any documentation to support this?

klumsy:
@aenand , only by comparing the cybersource SOAP api docs , and seeing that the instructions were the same. Its not something i can test myself however.

DustinHaefele:
Hey @klumsy, could you link the SOAP docs that you are referring to that give the discover details? I'm looking [here](https://developer.cybersource.com/docs/cybs/en-us/apple-pay/developer/vantivcnp/so/applepay/applepay-txns-intro.html) and only seeing visa and mastercard mentioned

DustinHaefele:
Also this list of supported SOAP apple pay payment methods does not include discover https://developer.cybersource.com/docs/cybs/en-us/apple-pay/developer/ctv/so/applepay/applepay-intro/applepay-processors.html

klumsy:
@DustinHaefele that link is for a different underlying processor than we use (we use cybersource integrations to first data processors, Fdiglobal, FDC Nashville Global etc.. Cybersource docs are hard to find, as they have a PDF for each combination. When we did see it (i can;’t find the link again now) it said it was supported, additionally their support said it was supported, however we can’t verify until the code no longer blocks attempting a NT payment with a cryptogram for it.. if this gets updated so that we can send it the same as with Visa, maybe it will work, or maybe we’ll get a different error from the gateway and processor, and then we can follow up with Cybersource Support again with more information.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -895,7 +895,7 @@ def add_auth_wallet(xml, payment_method, options)
         brand = card_brand(payment_method).to_sym
 
         case brand
-        when :visa
+        when :visa, :discover
           xml.tag! 'ccAuthService', { 'run' => 'true' } do
             xml.tag!('cavv', payment_method.payment_cryptogram) unless commerce_indicator
             xml.commerceIndicator commerce_indicator.nil? ? ECI_BRAND_MAPPING[brand] : commerce_indicator
```
klumsy, line 901: the XID thing, which im not sure happens in our circumstance here might be an only VISA thing, so maybe the correct discover code might be the same code, but without the XID line

klumsy:
looking further it seems the XID isn't needed for discover so likely the logic would be duplicated but with out that such as 

```
when :discover
      xml.commerceIndicator commerce_indicator.nil? ? ECI_BRAND_MAPPING[brand] : commerce_indicator
      xml.tag!('cavv', payment_method.payment_cryptogram) unless commerce_indicator
      xml.tag!('reconciliationID', options[:reconciliation_id]) if options[:reconciliation_id]
    end
```

lib/active_merchant/billing/gateways/cyber_source.rb
klumsy, line 1: regarding the ECI_brand_mapping... for discover what is the different between `pb` here (like 33) and DIPB as per the cybersource docs... https://developer.cybersource.com/content/dam/docs/cybs/en-us/apple-pay/developer/smartpay/so/applepay.pdf

klumsy:
this has now been completed with another PR, so closing

PR #5142, opened 2024-06-04
Braintree Blue: Update card verfification payload billing address fields are not present

yunnydang:
This lets us gracefully return an error code or message from Braintree if the postal zip code is missing.

Local:
5923 tests, 79804 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
100% passed

Remote:
121 tests, 649 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -144,16 +144,21 @@ module ActiveMerchant #:nodoc:
           exp_month = creditcard.month.to_s
           exp_year = creditcard.year.to_s
           expiration = "#{exp_month}/#{exp_year}"
+          zip = options[:billing_address].try(:[], :zip)
+          address1 = options[:billing_address].try(:[], :address1)
           payload = {
             credit_card: {
               number: creditcard.number,
               expiration_date: expiration,
-              cvv: creditcard.verification_value,
-              billing_address: {
-                postal_code: options[:billing_address][:zip]
-              }
+              cvv: creditcard.verification_value
             }
           }
+          if zip || address1
+            payload[:credit_card][:billing_address] = {}
+            payload[:credit_card][:billing_address][:postal_code] = zip if zip
+            payload[:credit_card][:billing_address][:street_address] = address1 if address1
+          end
+
           if merchant_account_id = (options[:merchant_account_id] || @merchant_account_id)
             payload[:options] = { merchant_account_id: merchant_account_id }
           end
--- a/test/remote/gateways/remote_braintree_blue_test.rb
+++ b/test/remote/gateways/remote_braintree_blue_test.rb
@@ -271,6 +271,54 @@ class RemoteBraintreeBlueTest < Test::Unit::TestCase
 
     assert_match 'OK', response.message
     assert_equal 'M', response.cvv_result['code']
+    assert_equal 'M', response.avs_result['code']
+  end
+
+  def test_successful_credit_card_verification_without_billing_address
+    options = {
+      order_ID: '1',
+      description: 'store purchase'
+    }
+    card = credit_card('4111111111111111')
+    assert response = @gateway.verify(card, options.merge({ allow_card_verification: true, merchant_account_id: fixtures(:braintree_blue)[:merchant_account_id] }))
+    assert_success response
+
+    assert_match 'OK', response.message
+    assert_equal 'M', response.cvv_result['code']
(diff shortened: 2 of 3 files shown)


test/remote/gateways/remote_braintree_blue_test.rb
```diff
@@ -274,6 +274,30 @@ def test_successful_credit_card_verification
     assert_equal 'P', response.avs_result['code']
   end
 
+  def test_graceful_failure_credit_card_verification_without_zip_code
```
jcreiff, line 277: This test includes the `billing_address` hash, but I think the error scenario is what occurs when there is no `billing_address` present in `options`

That's why the NoMethodError occurs when we try to access `[:zip]`

We need a test scenario that shows the transaction fails gracefully when `billing_address` returns `nil`

jcreiff (changes requested):

lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -144,13 +144,14 @@ def verify(creditcard, options = {})
...
           expiration = "#{exp_month}/#{exp_year}"
+          zip = options[:billing_address][:zip] || nil
           payload = {
             credit_card: {
               number: creditcard.number,
               expiration_date: expiration,
               cvv: creditcard.verification_value,
               billing_address: {
-                postal_code: options[:billing_address][:zip]
+                postal_code: zip
```
jcreiff, line 154: I think you might be able to achieve what we need here without instantiating `zip` above - you could instead do `options[:billing_address].try(:[], :zip)`

test/remote/gateways/remote_braintree_blue_test.rb
yunnydang: ahhh gotcha, ill get it resorted!

jcreiff (commented):
We discussed this IRL and think maybe we need further investigation:

- Why are we only sending `zip` in the `billing_address` hash?
- Is `billing_address` optional? ie, if it's empty in `options` should we just skip adding it to the payload?

The nature of the error message (`Addresses must have at least one field filled in.`) led us to wonder about this, because it doesn't seem like the Braintree API is rejecting an empty `zip` per se; rather, it seems like a `billing_address` object with no details included is the problem

That error message would also be a strange one to receive (as an end user) if there were other details in the `billing_address` hash, but not `zip`

yunnydang:
> We discussed this IRL and think maybe we need further investigation:
> 
> * Why are we only sending `zip` in the `billing_address` hash?
> * Is `billing_address` optional? ie, if it's empty in `options` should we just skip adding it to the payload?
> 
> The nature of the error message (`Addresses must have at least one field filled in.`) led us to wonder about this, because it doesn't seem like the Braintree API is rejecting an empty `zip` per se; rather, it seems like a `billing_address` object with no details included is the problem
> 
> That error message would also be a strange one to receive (as an end user) if there were other details in the `billing_address` hash, but not `zip`

Braintrees card verification documentation requires:
Card number
Card expiration date
Street address and postal code (however im not sure this is actually true, more on that later)
CVV
I ran some remote tests in AM to mess around with this and if we send the street_address: options[:billing_address].try(:[], :address1) but not the zip/postal code, the verification is successful, at least according to the remote test. Same with just sending the zip/postal code by itself. I think the reason why Braintree doesnt like us sending an empty billing_address is because its one of the parameters according to this [doc](https://developer.paypal.com/braintree/docs/reference/request/credit-card-verification/create/ruby) and they expect to have some address related field within it. For example I think billing_address may actually just be optional as I've commented out that field/hash entirely and was able to get a successful verification. So in conclusion I think its like this:
For card verification, its optional to send the billing_address field
If we do send the billing_addressBraintree will expect at least one address related field (zip code, address1, or etc) to be present within it

jcreiff (changes requested):
I think the changes to the gateway model look good; my only request is for some updates to the remote test suite:

- `test_successful_credit_card_verification` now returns a failure because the AVS Result Code comes back as `M` instead of `P`. This is expected because we are sending more information to Braintree; as I understand it, `M` means street and postal code match, while `P` means there's a postal code match but street can't be confirmed. `P` was accurate for the previous zip-only logic, but now we are sending more than that.
- Can you add assertions for your new tests that check the `response.avs_result['code']`, similar to the above failing test?

jcreiff (approved):
🚢

PR #5143, opened 2024-06-06
DLocal: update the phone and ip fields

yunnydang:
Updates the top level `ip` and `phone` fields to correctly be mapped.
Note: the `onboarding_ip_address` is actually supposed to be within the `addtional_risk_data` object according to [DLocals documention](https://docs.dlocal.com/docs/risk-data-documentation#example-payment-with-additional-risk-data-object). Unsure why we were trying to set it as the `ip` field prior, but it wasn't being used correctly. 

Local:
5932 tests, 79858 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
46 tests, 202 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
40 tests, 110 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@
 * Cybersource Rest: Update support for stored credentials [aenand] #5083
 * Plexo: Add support to NetworkToken payments [euribe09] #5130
 * Braintree: Update card verfification payload if billing address fields are not present [yunnydang] #5142
+* DLocal: Update the phone and ip fields [yunnydang] #5143
 
 == Version 1.136.0 (June 3, 2024)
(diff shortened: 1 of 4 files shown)


jcreiff (changes requested):
I'm not sure it will make any difference in the outcome of the tests, but just in case, can we send something that looks like an actual ip address as the `ip`? We could just use the default `127.0.0.1`, for example

jcreiff (approved):
Looks good!

PR #5146, opened 2024-06-14
NMI: Adding GooglePay and ApplePay

javierpedrozaing:
Description
-------------------------
This commit add support to create transaction with GooglePay and ApplePay.
This payment methods are working for nmi_secure.

[SER-1295](https://spreedly.atlassian.net/browse/SER-1295)

Unit test
-------------------------
Finished in 11.283174 seconds.

59 tests, 475 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

100% passed

5.23 tests/s, 42.10 assertions/s

Remote test
-------------------------
Finished in 115.513346 seconds.

55 tests, 199 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications

78.1818% passed

0.48 tests/s, 1.72 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/nmi.rb
+++ b/lib/active_merchant/billing/gateways/nmi.rb
@@ -134,6 +134,7 @@ module ActiveMerchant #:nodoc:
           gsub(%r((cvv=)\d+), '\1[FILTERED]').
           gsub(%r((checkaba=)\d+), '\1[FILTERED]').
           gsub(%r((checkaccount=)\d+), '\1[FILTERED]').
+          gsub(%r((cavv=)[^&\n]*), '\1[FILTERED]').
           gsub(%r((cryptogram=)[^&]+(&?)), '\1[FILTERED]\2')
       end
 
@@ -166,7 +167,7 @@ module ActiveMerchant #:nodoc:
         elsif payment_method.is_a?(NetworkTokenizationCreditCard)
           post[:ccnumber] = payment_method.number
           post[:ccexp] = exp_date(payment_method)
-          post[:token_cryptogram] = payment_method.payment_cryptogram
+          add_network_token_fields(post, payment_method)
         elsif card_brand(payment_method) == 'check'
           post[:payment] = 'check'
           post[:firstname] = payment_method.first_name
@@ -184,6 +185,17 @@ module ActiveMerchant #:nodoc:
           post[:ccnumber] = payment_method.number
           post[:cvv] = payment_method.verification_value unless empty?(payment_method.verification_value)
           post[:ccexp] = exp_date(payment_method)
+        end
+      end
+
+      def add_network_token_fields(post, payment_method)
+        if payment_method.source == :apple_pay || payment_method.source == :google_pay
+          post[:cavv] = payment_method.payment_cryptogram
+          post[:eci] = payment_method.eci
+          post[:decrypted_applepay_data] = 1
+          post[:decrypted_googlepay_data] = 1
+        else
+          post[:token_cryptogram] = payment_method.payment_cryptogram
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/nmi.rb
```diff
@@ -166,7 +166,10 @@ def add_payment_method(post, payment_method, options)
         elsif payment_method.is_a?(NetworkTokenizationCreditCard)
           post[:ccnumber] = payment_method.number
           post[:ccexp] = exp_date(payment_method)
-          post[:token_cryptogram] = payment_method.payment_cryptogram
+          post[:cavv] = payment_method.payment_cryptogram
+          post[:eci] = payment_method.eci
+          post[:decrypted_applepay_data] = 1 if payment_method.source.to_s == 'apple_pay'
```
sinourain, line 171: If `payment_method.source` is a symbol you can compare it with `:apple_pay` and remove the to_s method (same with `'google_pay'`)

sinourain (approved):
LGTM just a small comment

lib/active_merchant/billing/gateways/nmi.rb
javierpedrozaing: thanks, updated

lib/active_merchant/billing/gateways/nmi.rb
```diff
@@ -166,7 +166,10 @@ def add_payment_method(post, payment_method, options)
         elsif payment_method.is_a?(NetworkTokenizationCreditCard)
           post[:ccnumber] = payment_method.number
           post[:ccexp] = exp_date(payment_method)
-          post[:token_cryptogram] = payment_method.payment_cryptogram
```
Heavyblade, line 169: ❓ Question:

with this change you mean that `token_cryptogram` is wrong for general EMVco Tokens ? and all apple pay / Google pay and VISA/MC tokens should use `cavv` ?

I'm asking just because backwards compatibility

Heavyblade (commented):
Looks pretty much ok @javierpedrozaing, left a comment just prevent possible issues with current customers (NMI have tone of transactions)

lib/active_merchant/billing/gateways/nmi.rb
Heavyblade: 🤔 🤔 sounds like a CAB process if do that move, what if just in case we leave that `token_cryptogram` field there for Emco tokens and use `cavv` for the ones with apple/google on it's source.

javierpedrozaing: I agree, sound more reasonable, I will do the change

lib/active_merchant/billing/gateways/nmi.rb
```diff
@@ -187,6 +188,17 @@ def add_payment_method(post, payment_method, options)
         end
       end
 
+      def add_network_token_fields(post, payment_method)
+        if payment_method.source == :apple_pay || payment_method.source == :google_pay
+          post[:cavv] = payment_method.payment_cryptogram
+          post[:eci] = payment_method.eci
+          post[:decrypted_applepay_data] = 1
+          post[:decrypted_googlepay_data] = 1
```
naashton, line 196: Do both of these need to be set to `1` regardless of which payment method type is being used?

naashton (approved):
LGTM! Just a question about the `decrypted_*` fields?

lib/active_merchant/billing/gateways/nmi.rb
javierpedrozaing: Hi @naashton base on the NMI docs yes, for both cases set 1 to indicate decrypted data is being sent.

PR #5150, opened 2024-06-24
Orbital: Update to accept UCAF Indicator GSF

almalee24:
If alternate_ucaf_flow is true and the eci value is 4, 6, or 7 then send the ucaf passed in by customer. If alternate_ucaf_flow is not passed then only send ucaf if provided by customer if not pass default of 4.

Remote:
134 tests, 543 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Unit:
152 tests, 892 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -724,7 +724,7 @@ module ActiveMerchant #:nodoc:
         add_mc_sca_recurring(xml, credit_card, parameters, three_d_secure)
         add_mc_program_protocol(xml, credit_card, three_d_secure)
         add_mc_directory_trans_id(xml, credit_card, three_d_secure)
-        add_mc_ucafind(xml, credit_card, three_d_secure)
+        add_mc_ucafind(xml, credit_card, three_d_secure, parameters)
       end
 
       def add_mc_sca_merchant_initiated(xml, credit_card, parameters, three_d_secure)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -753,10 +753,16 @@ def add_mc_directory_trans_id(xml, credit_card, three_d_secure)
         xml.tag!(:MCDirectoryTransID, three_d_secure[:ds_transaction_id]) if three_d_secure[:ds_transaction_id]
       end
 
-      def add_mc_ucafind(xml, credit_card, three_d_secure)
+      def add_mc_ucafind(xml, credit_card, three_d_secure, options)
         return unless three_d_secure
 
-        xml.tag! :UCAFInd, '4'
+        if options[:alternate_ucaf_flow]
```
DustinHaefele, line 759: A few questions on this if else block. 

1. If `alternate_ucaf_flow = true` are we ok if `options[:ucaf] = nil`?
2. Why have we added the eci check? 
3. Without the `alternate_ucaf_flow` feature flag enabled (so in the else block), we are still allowing a merchant to set options[:ucaf] and forwarding that to the gateway. I don't have a problem with this, but now the `eci` isn't being checked like it is in the block above, which seems like the functionality differs a bit.

DustinHaefele (commented):
I left a few questions in the main functionality portion.

lib/active_merchant/billing/gateways/orbital.rb
almalee24: If options[:ucaf] is nil then we won't send :UCAFInd. But looking at the ticket again it looks like I miss read it and we should send it as nil if not was provided

almalee24: To answer your third question, it was mentioned in the ticket that we should send the GSF if provided by the customer but it didn't mention eci check.

DustinHaefele: Ah, I should have read the ticket. This all makes sense then! I'll approve with the current code. Thanks for explaining things to me.

DustinHaefele (approved):

kylene-spreedly (approved):
This looks good to me! Nice work!

PR #5151, opened 2024-06-25
CheckoutV2: truncate the reference id for amex transactions

yunnydang:
This is to truncate the reference id for Amex auth and purchase transactions. I left the the original add_invoice(post, amount, options) method alone because other types of endpoints utilize it but do not pass the payment_method object. If calling purchase or authorization, and you happen to pass an amex payment method, then the reference id will be overwritten/shortened

Local:
5937 tests, 79881 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
68 tests, 425 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
111 tests, 269 assertions, 4 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.3964% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,6 +13,7 @@
 * CheckoutV2: Add support for risk data fields [yunnydang] #5147
 * Pin Payments: Add new 3DS params mentioned in Pin Payments docs [hudakh] #4720
 * RedsysRest: Add support for stored credentials & 3DS exemptions [jherreraa] #5132
+* CheckoutV2: Truncate the reference id for amex transactions [yunnydang] #5151
 
 == Version 1.136.0 (June 3, 2024)
 * Shift4V2: Add new gateway based on SecurionPay adapter [heavyblade] #4860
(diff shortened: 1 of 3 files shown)


jcreiff:
🤔 Is there any reason we would need to truncate for transactions that are not purchase or authorize? Could an order_id/reference be passed in for `capture` `refund` or `credit`? And would the error be thrown from Checkout if the reference was too long in those transactions?

yunnydang:
> 🤔 Is there any reason we would need to truncate for transactions that are not purchase or authorize? Could an order_id/reference be passed in for `capture` `refund` or `credit`? And would the error be thrown from Checkout if the reference was too long in those transactions?

hmm a quick peek at the docs for these other endpoints indicate so. Ill get this update to reflect in these other mentioned endpoints!

jcreiff:
We looked at this further and decided it was not worth the extra complexity to add this logic to reference transactions, where the payment method (and therefore brand) are not typically available in `options`

jcreiff (approved):

PR #5155, opened 2024-06-25
ECS-3530 Adyen Format error fix

jherreraa:
[ECS-3530](https://spreedly.atlassian.net/browse/ECS-3530)

## Summary:

This PR fix format error using NT

Unit tests
----------------
Finished in 0.126432 seconds.

121 tests, 641 assertions, 0 failures,
0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote tests
----------------
Finished in 609.957316 seconds.

143 tests, 464 assertions, 11 failures,
0 errors, 0 pendings, 0 omissions, 0 notifications 92.3077% passed

0.23 tests/s, 0.76 assertions/s

-> failures not related to change

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -63,7 +63,7 @@ module ActiveMerchant #:nodoc:
         add_3ds(post, options)
         add_3ds_authenticated_data(post, options)
         add_splits(post, options)
-        add_recurring_contract(post, options)
+        add_recurring_contract(post, options, payment)
         add_network_transaction_reference(post, options)
         add_application_info(post, options)
         add_level_2_data(post, options)
@@ -625,20 +625,31 @@ module ActiveMerchant #:nodoc:
 
         post[:mpiData] = {}
         post[:mpiData][:authenticationResponse] = 'Y'
-        post[:mpiData][:cavv] = payment.payment_cryptogram
+        if NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s].nil? && options[:switch_cryptogram_mapping_nt]
+          post[:mpiData][:tokenAuthenticationVerificationValue] = payment.payment_cryptogram
+        else
+          post[:mpiData][:cavv] = payment.payment_cryptogram
+        end
         post[:mpiData][:directoryResponse] = 'Y'
         post[:mpiData][:eci] = payment.eci || '07'
       end
 
-      def add_recurring_contract(post, options = {})
-        return unless options[:recurring_contract_type]
+      def add_recurring_contract(post, options = {}, payment = nil)
+        return unless options[:recurring_contract_type] || (payment.try(:source) == :network_token && options[:switch_cryptogram_mapping_nt])
 
-        post[:recurring] = {}
-        post[:recurring][:contract] = options[:recurring_contract_type]
+        post[:recurring] ||= {}
+        post[:recurring][:contract] = options[:recurring_contract_type] if options[:recurring_contract_type]
         post[:recurring][:recurringDetailName] = options[:recurring_detail_name] if options[:recurring_detail_name]
         post[:recurring][:recurringExpiry] = options[:recurring_expiry] if options[:recurring_expiry]
         post[:recurring][:recurringFrequency] = options[:recurring_frequency] if options[:recurring_frequency]
         post[:recurring][:tokenService] = options[:token_service] if options[:token_service]
+        if payment.try(:source) == :network_token && options[:switch_cryptogram_mapping_nt]
+          post[:recurring][:contract] = 'EXTERNAL'
+          post[:recurring][:tokenService] = case payment.brand
+                                            when 'visa' then 'VISATOKENSERVICE'
+                                            else 'MCTOKENSERVICE'
+                                            end
+        end
       end
 
       def add_application_info(post, options)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,7 @@
 * Braintree: Updates to Paypal Integration [almalee24] #5190
 * CommerceHub: Update test url [DustinHaefele] #5211
 * Elavon: Update sending CVV for MIT transactions  [almalee24] #5210
+* Adyen: Fix NT integration [jherreraa] #5155
 
 == Version 1.137.0 (August 2, 2024)
 * Unlock dependency on `rexml` to allow fixing a CVE (#5181).
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -287,7 +287,7 @@ class RemoteAdyenTest < Test::Unit::TestCase
   end
 
   def test_successful_authorize_with_network_token
-    response = @gateway.authorize(@amount, @nt_credit_card, @options)
+    response = @gateway.authorize(@amount, @nt_credit_card, @options.merge(switch_cryptogram_mapping_nt: true))
     assert_success response
     assert_equal 'Authorised', response.message
   end
@@ -562,7 +562,13 @@ class RemoteAdyenTest < Test::Unit::TestCase
   end
 
   def test_successful_purchase_with_apple_pay
-    response = @gateway.purchase(@amount, @apple_pay_card, @options)
+    response = @gateway.purchase(@amount, @apple_pay_card, @options.merge(switch_cryptogram_mapping_nt: true))
+    assert_success response
+    assert_equal '[capture-received]', response.message
+  end
+
+  def test_successful_purchase_with_apple_pay_with_ld_flag_false
+    response = @gateway.purchase(@amount, @apple_pay_card, @options.merge(switch_cryptogram_mapping_nt: false))
     assert_success response
     assert_equal '[capture-received]', response.message
   end
@@ -580,7 +586,7 @@ class RemoteAdyenTest < Test::Unit::TestCase
   end
 
   def test_successful_purchase_with_google_pay
-    response = @gateway.purchase(@amount, @google_pay_card, @options)
+    response = @gateway.purchase(@amount, @google_pay_card, @options.merge(switch_cryptogram_mapping_nt: true))
     assert_success response
     assert_equal '[capture-received]', response.message
   end
@@ -610,7 +616,7 @@ class RemoteAdyenTest < Test::Unit::TestCase
   end
 
   def test_successful_purchase_with_google_pay_and_truncate_order_id
-    response = @gateway.purchase(@amount, @google_pay_card, @options.merge(order_id: @long_order_id))
+    response = @gateway.purchase(@amount, @google_pay_card, @options.merge(order_id: @long_order_id, switch_cryptogram_mapping_nt: true))
     assert_success response
     assert_equal '[capture-received]', response.message
   end
@@ -634,7 +640,7 @@ class RemoteAdyenTest < Test::Unit::TestCase
   end
 
   def test_successful_purchase_with_network_token
-    response = @gateway.purchase(@amount, @nt_credit_card, @options)
+    response = @gateway.purchase(@amount, @nt_credit_card, @options.merge(switch_cryptogram_mapping_nt: true))
     assert_success response
     assert_equal '[capture-received]', response.message
   end
@@ -1450,7 +1456,8 @@ class RemoteAdyenTest < Test::Unit::TestCase
     first_options = options.merge(
       order_id: generate_unique_id,
       shopper_interaction: 'Ecommerce',
-      recurring_processing_model: 'Subscription'
+      recurring_processing_model: 'Subscription',
+      switch_cryptogram_mapping_nt: true
     )
     assert auth = @gateway.authorize(@amount, @apple_pay_card, first_options)
     assert_success auth
@@ -1465,7 +1472,8 @@ class RemoteAdyenTest < Test::Unit::TestCase
       skip_mpi_data: 'Y',
       shopper_interaction: 'ContAuth',
       recurring_processing_model: 'Subscription',
-      network_transaction_id: auth.network_transaction_id
+      network_transaction_id: auth.network_transaction_id,
+      switch_cryptogram_mapping_nt: true
     )
 
     assert purchase = @gateway.purchase(@amount, @apple_pay_card, used_options)
--- a/test/unit/gateways/adyen_test.rb
+++ b/test/unit/gateways/adyen_test.rb
@@ -1232,12 +1232,38 @@ class AdyenTest < Test::Unit::TestCase
 
   def test_authorize_with_network_tokenization_credit_card
     response = stub_comms do
-      @gateway.authorize(@amount, @apple_pay_card, @options)
+      @gateway.authorize(@amount, @apple_pay_card, @options.merge(switch_cryptogram_mapping_nt: false))
     end.check_request do |_endpoint, data, _headers|
       parsed = JSON.parse(data)
       assert_equal 'YwAAAAAABaYcCMX/OhNRQAAAAAA=', parsed['mpiData']['cavv']
       assert_equal '07', parsed['mpiData']['eci']
       assert_equal 'applepay', parsed['additionalData']['paymentdatasource.type']
+    end.respond_with(successful_authorize_response)
+    assert_success response
+  end
+
+  def test_authorize_with_network_tokenization_credit_card_using_ld_option
+    response = stub_comms do
+      @gateway.authorize(@amount, @apple_pay_card, @options.merge(switch_cryptogram_mapping_nt: true))
+    end.check_request do |_endpoint, data, _headers|
+      parsed = JSON.parse(data)
+      assert_equal 'YwAAAAAABaYcCMX/OhNRQAAAAAA=', parsed['mpiData']['cavv']
+      assert_equal '07', parsed['mpiData']['eci']
+      assert_equal 'applepay', parsed['additionalData']['paymentdatasource.type']
+    end.respond_with(successful_authorize_response)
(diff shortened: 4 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,34 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
+      def add_recurring_contract(post, options = {}, payment = nil)
+        return unless options[:recurring_contract_type] || payment.is_a?(NetworkTokenizationCreditCard)
 
         post[:recurring] = {}
-        post[:recurring][:contract] = options[:recurring_contract_type]
-        post[:recurring][:recurringDetailName] = options[:recurring_detail_name] if options[:recurring_detail_name]
-        post[:recurring][:recurringExpiry] = options[:recurring_expiry] if options[:recurring_expiry]
-        post[:recurring][:recurringFrequency] = options[:recurring_frequency] if options[:recurring_frequency]
-        post[:recurring][:tokenService] = options[:token_service] if options[:token_service]
+        if payment.is_a?(NetworkTokenizationCreditCard)
```
almalee24, line 638: I would also suggest adding `options[:switch_cryptogram_mapping_nt]` here since this is also part of the overall changes

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,34 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
+          post[:recurring][:contract] = 'EXTERNAL'
+          post[:recurring][:tokenService] = case payment.brand
+          when 'visa'
+            'VISATOKENSERVICE'
+          else
+            'MCTOKENSERVICE'
+          end
+        else
+          post[:recurring][:contract] = options[:recurring_contract_type]
+          post[:recurring][:recurringDetailName] = options[:recurring_detail_name] if options[:recurring_detail_name]
```
almalee24, line 648: I think it's okay to move lines 648-651 outside of the else statement.

almalee24:
There's also a Rubocop error which is why the GitHub actions tests are failing

lib/active_merchant/billing/gateways/adyen.rb
jherreraa: I think that according the behavior on the tests, what you suggests is not needed because these fields are neeeded no matter what the cryptogram mapping is

lib/active_merchant/billing/gateways/adyen.rb
jherreraa: Noted! Done!

lib/active_merchant/billing/gateways/adyen.rb
almalee24: Yes, I understand but they are changes failed last time so we just want to be extra careful which is why we want to use the LD flag in general.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,31 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
 
         post[:mpiData] = {}
         post[:mpiData][:authenticationResponse] = 'Y'
-        post[:mpiData][:cavv] = payment.payment_cryptogram
+        if options[:switch_cryptogram_mapping_nt]
+          post[:mpiData][:tokenAuthenticationVerificationValue] = payment.payment_cryptogram
```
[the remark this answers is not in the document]
DustinHaefele: I'm not against putting it behind a flag as well, but at some point we will want to remove the flag, and I think we will need both mappings at that point.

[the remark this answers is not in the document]
jherreraa: On previous tests we got some error some in order to avoid a double deploy while we assure the proper behavior we are going to use the LD approach. @almalee24 could give you some additional info on that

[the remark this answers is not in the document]
almalee24: Yeah last time the told us to change this field but it failed when it went out which is why we are doing it with a LD flag. Just to be extra cautious and it'll be easier to turn off it sometimes goes wrong.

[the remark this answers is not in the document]
almalee24: But we should update it to be `options[:switch_cryptogram_mapping_nt] && payment.is_a?(NetworkTokenizationCreditCard)`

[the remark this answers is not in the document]
DustinHaefele: Yeah that sounds good, if we use the flag and also check the payment type, then if it works we can remove the flag, and keep the payment type to have the same behavior.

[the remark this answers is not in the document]
jherreraa: after running tests, the switch in cryptogram mapping causes an error on mpidata so this latest update keeps the adapter correct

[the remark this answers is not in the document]
DustinHaefele: We should probably talk to adyen about this. This issue started because there were failures that adyen said was because the cryptogram was sent in the `cavv` instead of the `tokenAuthenticationVerificationValue` field. 

Need to make sure we understand what exactly the failures are in the remote tests and how we should proceed with these fields long term.

DustinHaefele:
I added the conditional back into the `app_mpi_data_for_network_tokenization_card` method updating the `cavv` to `tokenAuthenticationVerificationValue`. It seems to me that the issue in the tests is with ApplePay and GooglePay. We need to be able to differentiate these from the standard network token flow, and continue to pass that value as `cavv` for apple pay and google pay. 

Then we also need to continue to pass the existing recurring_contract info for apple_pay and google pay and only update that when it is a pure network token. 

cc: @almalee24 @jherreraa

DustinHaefele (changes requested):
I left a couple of individual comments on this one. 

We may not need to reach out to Adyen if we handle the situations we left in my second comment.

lib/active_merchant/billing/gateways/adyen.rb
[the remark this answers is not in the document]
jherreraa: So is there a way to contact adyen support? as far as I know is not easy to get in touch with then.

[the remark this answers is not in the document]
DustinHaefele: I'm not sure. Did you see my comment about trying to have these updates only apply to NetworkTokens, and not ApplePay and GooglePay? It seems that if we can keep the apple pay and google pay integration the same, and only update the fields we are sending for NetworkTokens it may solve the problems you are seeing in the tests.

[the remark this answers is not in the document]
jherreraa: Noted! Done!

DustinHaefele (dismissed):

DustinHaefele:
Great work on this one Johan, I know this framework can be a bit tricky!

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,31 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
+          post[:mpiData][:cavv] = payment.payment_cryptogram
+        end
         post[:mpiData][:directoryResponse] = 'Y'
         post[:mpiData][:eci] = payment.eci || '07'
       end
 
-      def add_recurring_contract(post, options = {})
-        return unless options[:recurring_contract_type]
+      def add_recurring_contract(post, options = {}, payment = nil)
+        return unless options[:recurring_contract_type] || payment.is_a?(NetworkTokenizationCreditCard)
```
DustinHaefele, line 635: I ran remote tests real quick on this for if `options[:switch_cryptogram_mapping_nt]=false` during an apple pay or google pay transaction. In that case we get an error from Adyen that says `"message"=>"Invalid recurring contract specified"` This is because the body has the following in it, and contract can't be nil if it is passed.  So we need to check for apple and google pay on the recurring contract logic as well.
```
{
  :recurring=>{ :contract=>nil }
}
```

```suggestion
        return unless options[:recurring_contract_type] || ( payment.is_a?(NetworkTokenizationCreditCard) && NETWORK_TOKENIZATION_CARD_SOURCE[payment&.source.to_s].nil? )
```

DustinHaefele (changes requested):
@jherreraa, Hey, I found one more issue that we need to resolve before merging this. It should be a quick fix though. 

Also maybe add at least one test on apple pay and google pay where the new flag is set to false so we make sure that it passes in both cases.

lib/active_merchant/billing/gateways/adyen.rb
jherreraa: After check the tests and verify the request contents I have tried a different approach because in my findings the fields `post[:recurring][:tokenService]` & `post[:recurring][:contract]` were not properly populated.
Please take a look and let me know what you think.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,33 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
-        post[:recurring][:tokenService] = options[:token_service] if options[:token_service]
+        post[:recurring] ||= {}
+
+        if options[:recurring_contract_type]
+          post[:recurring][:contract] ||= options[:recurring_contract_type]
+          post[:recurring][:recurringDetailName] = options[:recurring_detail_name]
+          post[:recurring][:recurringExpiry] = options[:recurring_expiry]
+          post[:recurring][:recurringFrequency] = options[:recurring_frequency]
+          post[:recurring][:tokenService] ||= options[:token_service]
+        elsif payment.is_a?(NetworkTokenizationCreditCard) && options[:switch_cryptogram_mapping_nt]
```
DustinHaefele, line 645: In this new `elsif` scenario we aren't setting the `recurringDetailName` `recurringExpiry` or `recurringFrequency` Is this intentional? If so why do we not want to set them in that case?

DustinHaefele: Also I still don't know if these fields are applicable to apple_pay and google_pay since those aren't automatically recurring payments. I think we should only set them automatically for true NetworkTokens.

DustinHaefele (changes requested):
I think this is the last thing that we need to solidify and then we are good to go on this. I know this has been a long one, thanks for sticking with it!

lib/active_merchant/billing/gateways/adyen.rb
almalee24: I agree that this should only be for NetworkTokens and that the NetworkToken check should be outside the if statment

```
return unless options[:recurring_contract_type] || (payment.try(:source) == :network_token && options[:switch_cryptogram_mapping_nt])

        post[:recurring] ||= {}

        if options[:recurring_contract_type]
          post[:recurring][:contract] ||= options[:recurring_contract_type]
          post[:recurring][:recurringDetailName] = options[:recurring_detail_name]
          post[:recurring][:recurringExpiry] = options[:recurring_expiry]
          post[:recurring][:recurringFrequency] = options[:recurring_frequency]
          post[:recurring][:tokenService] ||= options[:token_service]
        end

if payment.try(:source) == :network_token && options[:switch_cryptogram_mapping_nt]
          post[:recurring][:contract] = 'EXTERNAL'
          post[:recurring][:tokenService] = case payment.brand
                                            when 'visa' then 'VISATOKENSERVICE'
                                            else 'MCTOKENSERVICE'
                                            end
        end
```

jherreraa: I missed the other recurring fields in the previous approach trying to avoid the `recurring=>{nil}` ocurrence.
I think the current one solves that.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,31 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
+          post[:mpiData][:cavv] = payment.payment_cryptogram
+        end
         post[:mpiData][:directoryResponse] = 'Y'
         post[:mpiData][:eci] = payment.eci || '07'
       end
 
-      def add_recurring_contract(post, options = {})
-        return unless options[:recurring_contract_type]
+      def add_recurring_contract(post, options = {}, payment = nil)
+        return unless options[:recurring_contract_type] || (payment.is_a?(NetworkTokenizationCreditCard) && options[:switch_cryptogram_mapping_nt])
```
almalee24, line 635: NetworkTokenizationCreditCard is for ApplePay, GooglePay and NetworkToken and `payment.source` is what tells you what exactly it is. This is why because of this it's why here you have to updated it to be just like you did in [line 625](https://github.com/activemerchant/active_merchant/pull/5155/files#diff-022d82d7d09d8eb787e16961826d7f88f536d9eb2a9ef947432dcf26250258e0R625) or you can try `payment.try(:source) == :network_token`

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,31 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
 
-        post[:recurring] = {}
-        post[:recurring][:contract] = options[:recurring_contract_type]
+        post[:recurring] ||= {}
+        post[:recurring][:contract] = options[:recurring_contract_type] if options[:recurring_contract_type]
         post[:recurring][:recurringDetailName] = options[:recurring_detail_name] if options[:recurring_detail_name]
         post[:recurring][:recurringExpiry] = options[:recurring_expiry] if options[:recurring_expiry]
         post[:recurring][:recurringFrequency] = options[:recurring_frequency] if options[:recurring_frequency]
         post[:recurring][:tokenService] = options[:token_service] if options[:token_service]
+        if payment.is_a?(NetworkTokenizationCreditCard) && options[:switch_cryptogram_mapping_nt]
```
almalee24, line 643: Same as what I commented above

jherreraa: Understood, updated!

lib/active_merchant/billing/gateways/adyen.rb
jherreraa: Updated!

lib/active_merchant/billing/gateways/adyen.rb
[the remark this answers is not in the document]
almalee24: @DustinHaefele can you take a look at this PR again since you have requested changes?

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,31 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
+          post[:mpiData][:cavv] = payment.payment_cryptogram
+        end
         post[:mpiData][:directoryResponse] = 'Y'
         post[:mpiData][:eci] = payment.eci || '07'
       end
 
-      def add_recurring_contract(post, options = {})
-        return unless options[:recurring_contract_type]
+      def add_recurring_contract(post, options = {}, payment = nil)
+        return unless options[:recurring_contract_type] || (payment.try(:source) && options[:switch_cryptogram_mapping_nt])
```
almalee24, line 635: This needs to be `payment.try(:source) == :network_token`

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -622,20 +622,31 @@ def add_mpi_data_for_network_tokenization_card(post, payment, options)
...
 
-        post[:recurring] = {}
-        post[:recurring][:contract] = options[:recurring_contract_type]
+        post[:recurring] ||= {}
+        post[:recurring][:contract] = options[:recurring_contract_type] if options[:recurring_contract_type]
         post[:recurring][:recurringDetailName] = options[:recurring_detail_name] if options[:recurring_detail_name]
         post[:recurring][:recurringExpiry] = options[:recurring_expiry] if options[:recurring_expiry]
         post[:recurring][:recurringFrequency] = options[:recurring_frequency] if options[:recurring_frequency]
         post[:recurring][:tokenService] = options[:token_service] if options[:token_service]
+        if payment.try(:source) && options[:switch_cryptogram_mapping_nt]
```
almalee24, line 643: This needs to be `payment.try(:source) == :network_token && options[:switch_cryptogram_mapping_nt]`

DustinHaefele, line 643: The payment will have a source even if it isn't apple_pay or google_pay. We still need to check that it isn't those types just like on line 625 above.
```suggestion
        if NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s].nil? && options[:switch_cryptogram_mapping_nt]
```

DustinHaefele: Once these are fixed I'll pull my needs updates tag off.

jherreraa: @DustinHaefele, @almalee24 In case of googlepay  and applepay what would happen in 643 to 648?
for network tokens the `post[:recurring][:contract]` & `post[:recurring][:tokenService]` would be missed in both cases and according to my understanding of the info in https://spreedly.slack.com/archives/CH7B6NWA0/p1709310814097569 those are required.
Please check and let me know what you think

DustinHaefele: For googlepay and applepay we don't want the values set in 643-648. This is because lines 643-648 are for true network_tokens that are issued by visa and mastercard. In the case of applepay and googlepay the customer would have to pass in the `options[:token_service]` and `options[:recurring_contract_type]` values if they wanted to add them here.

jherreraa: Understood, fix added. Thanks!

DustinHaefele (approved):

almalee24 (approved):

PR #5157, opened 2024-06-27
CommerceHub: Add billing address name overide

yunnydang:
This allows us to grab the name from billing address if card holder name is not present. Also fixed apiKey scrubbing and a few tests messages. 2 of the 3 remote test failures were due to "The transaction limit was exceeded. Please try again!" even though i ran the test suite with a sleep of 10 mentioned in the comment.

Local:
5942 tests, 79901 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
28 tests, 194 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
34 tests, 86 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.1765% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/commerce_hub.rb
+++ b/lib/active_merchant/billing/gateways/commerce_hub.rb
@@ -112,6 +112,7 @@ module ActiveMerchant #:nodoc:
         transcript.
           gsub(%r((Authorization: )[a-zA-Z0-9+./=]+), '\1[FILTERED]').
           gsub(%r((Api-Key: )\w+), '\1[FILTERED]').
+          gsub(%r(("apiKey\\?":\\?")\w+), '\1[FILTERED]').
           gsub(%r(("cardData\\?":\\?")\d+), '\1[FILTERED]').
           gsub(%r(("securityCode\\?":\\?")\d+), '\1[FILTERED]').
           gsub(%r(("cavv\\?":\\?")\w+), '\1[FILTERED]')
@@ -171,10 +172,7 @@ module ActiveMerchant #:nodoc:
         return unless billing = options[:billing_address]
 
         billing_address = {}
-        if payment.is_a?(CreditCard)
-          billing_address[:firstName] = payment.first_name if payment.first_name
-          billing_address[:lastName] = payment.last_name if payment.last_name
-        end
+        name_from_address(billing_address, billing) || name_from_payment(billing_address, payment)
         address = {}
         address[:street] = billing[:address1] if billing[:address1]
         address[:houseNumberOrName] = billing[:address2] if billing[:address2]
@@ -190,6 +188,22 @@ module ActiveMerchant #:nodoc:
           billing_address[:phone][:phoneNumber] = billing[:phone_number]
         end
         post[:billingAddress] = billing_address
+      end
+
+      def name_from_payment(billing_address, payment)
+        return unless payment.respond_to?(:first_name) && payment.respond_to?(:last_name)
+
+        billing_address[:firstName] = payment.first_name if payment.first_name
+        billing_address[:lastName] = payment.last_name if payment.last_name
+      end
+
+      def name_from_address(billing_address, billing)
+        return unless address = billing
+
+        first_name, last_name = split_names(address[:name]) if address[:name]
+
+        billing_address[:firstName] = first_name if first_name
+        billing_address[:lastName] = last_name if last_name
       end
 
       def add_shipping_address(post, options)
--- a/test/remote/gateways/remote_commerce_hub_test.rb
+++ b/test/remote/gateways/remote_commerce_hub_test.rb
@@ -70,6 +70,49 @@ class RemoteCommerceHubTest < Test::Unit::TestCase
     assert_equal 'Approved', response.message
   end
 
+  def test_successful_purchase_with_payment_name_override
+    billing_address = {
+      address1: 'Infinite Loop',
+      address2: 1,
+      country: 'US',
+      city: 'Cupertino',
+      state: 'CA',
+      zip: '95014'
+    }
+
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(billing_address: billing_address))
+    assert_success response
+    assert_equal 'Approved', response.message
+    assert_equal 'John', response.params['billingAddress']['firstName']
+    assert_equal 'Doe', response.params['billingAddress']['lastName']
+  end
+
+  def test_successful_purchase_with_name_override_on_alternative_payment_methods
+    billing_address = {
+      address1: 'Infinite Loop',
+      address2: 1,
+      country: 'US',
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -171,10 +172,7 @@ def add_billing_address(post, payment, options)
         return unless billing = options[:billing_address]
 
         billing_address = {}
-        if payment.is_a?(CreditCard)
-          billing_address[:firstName] = payment.first_name if payment.first_name
-          billing_address[:lastName] = payment.last_name if payment.last_name
-        end
+        name_from_payment(billing_address, payment) || name_from_address(billing_address, options) if payment.is_a?(CreditCard)
```
jcreiff, line 175: I think the order of this `or` statement needs to be swapped - if the  billing address is provided, the idea is that those values override what might be stored on the card

jcreiff: I also wonder if the `if payment.is_a?(CreditCard)` is necessary here or if it should be a guard clause on `name_from_payment` instead?

Or better yet, should `name_from_payment` just check if `payment.respond_to?(:first_name)` (and last name)? This gateway also supports ApplePay and GooglePay, so that solution might be more flexible and ensure that we aren't skipping logic that could apply to non-credit card payments

lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -192,6 +190,22 @@ def add_billing_address(post, payment, options)
         post[:billingAddress] = billing_address
       end
 
+      def name_from_payment(billing_address, payment)
+        return unless payment.first_name && payment.last_name
```
jcreiff, line 194: The other reason I think we might want to use `respond_to?` here is that the previous logic would in theory allow for us to send just a first name or just a last name, if by chance only one was included.

test/remote/gateways/remote_commerce_hub_test.rb
```diff
@@ -70,6 +70,27 @@ def test_successful_purchase
     assert_equal 'Approved', response.message
   end
 
+  def test_successful_purchase_with_payment_name_overide
```
jcreiff, line 73: Small nitpick - `override` instead of `overide` for both of these test names

lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -192,6 +190,22 @@ def add_billing_address(post, payment, options)
...
+
+        billing_address[:firstName] = payment.first_name
+        billing_address[:lastName] = payment.last_name
+      end
+
+      def name_from_address(billing_address, options)
+        return unless address = options[:billing_address]
+
+        first_name, last_name = address[:name].split(' ') if address[:name]
```
jcreiff, line 203: An edge case that would break this logic would be if there were two spaces in the name field - for example `Michael J. Fox` would become `first_name = Michael` and `last_name = J.`, because the shotgun assignment would only take the first two items in the array.

There is actually a helper method for this type of operation in `gateway.rb` called `split_names` !

lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -112,6 +112,7 @@ def scrub(transcript)
         transcript.
           gsub(%r((Authorization: )[a-zA-Z0-9+./=]+), '\1[FILTERED]').
           gsub(%r((Api-Key: )\w+), '\1[FILTERED]').
+          gsub(%r(("apiKey\\?":\\?")\w+), '\1[FILTERED]').
```
jcreiff, line 115: Interesting - is this field being echoed back to us in the response?

jcreiff (changes requested):

lib/active_merchant/billing/gateways/commerce_hub.rb
yunnydang: ahh i see! I didnt catch that we were prioritizing the billing_address first! Also good point on the googlepay/apple pay thought!

lib/active_merchant/billing/gateways/commerce_hub.rb
yunnydang: good to know!

lib/active_merchant/billing/gateways/commerce_hub.rb
yunnydang: yes the remote scrub test was failing because this apiKey field was not being scrubbed out!

lib/active_merchant/billing/gateways/commerce_hub.rb
yunnydang: heya @jcreiff a little confused about this but from what im gathering, we're trying to not send any name fields if both do not exist correct? In that case, I always thought respond_to? was to see if the attribute/class method exists (even if its nil). I did some testing for this and if i replace that line with something like return unless payment.respond_to?(:first_name) && payment.respond_to?(:last_name) and set the credit_card.last_name for example to nil, it will still proceed and try to map the both names. If we're trying to skip this method if only one name was sent, wouldn't the return unless payment.first_name && payment.last_name cover that? 🧐

jcreiff: > we're trying to not send any name fields if both do not exist correct?

The way it is originally written on lines 175-76, it doesn't look like that is an expectation. (although...should it be? 🤷🏻 )

What I was thinking was this:
- remove the `if payment.is_a?(CreditCard)` check from 175 so that we can add names for non-CC payment methods
- add the `respond_to` guard clause at the top of `name_from_payment` so that we don't try to call those methods on something like a third party token, for example
- then if you keep the `if payment.first_name` and `if payment.last_name` nil checks at the ends of lines 196-97, you will still skip adding empty fields

lib/active_merchant/billing/gateways/commerce_hub.rb
jcreiff, line 175: One other thing I just noticed - you could pass `billing` to `name_from_address` instead of the entire `options` hash, since it's already whittled down to just `options[:billing_address]`

lib/active_merchant/billing/gateways/commerce_hub.rb
yunnydang: ahh gotcha, that makes more sense! so in theory, we will still send either the first_name or last_name fields even if only one of them exists

yunnydang: with the return unless payment.first_name && payment.last_name at the top originally, it would skip the entire method and return out if both name fields do not exist

jcreiff (approved):
Looks good! 🚢

PR #5159, opened 2024-07-01
Credorax: Update 3DS version mapping

almalee24:
Update so that if a 3DS version starts with 2 it doesn't default to 2.0 but on what is passed in for the three_ds_version field.

Unit:
82 tests, 394 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
51 tests, 173 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 88.2353% passed


bradbroge (approved):
Looks good

PR #5160, opened 2024-07-01
Stripe PI: add optional ability for 3DS exemption on verify calls

yunnydang:
This allows us to send the optional moto and confirm on create_setup_intent transactions.

Local:
5945 tests, 79911 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
61 tests, 315 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
94 tests, 425 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.8085% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@
 * RedsysRest: Add support for stored credentials & 3DS exemptions [jherreraa] #5132
 * CheckoutV2: Truncate the reference id for amex transactions [yunnydang] #5151
 * CommerceHub: Add billing address name override [yunnydang] #5157
+* StripePI: Add optional ability for 3DS exemption on verify calls [yunnydang] #5160
 
 == Version 1.136.0 (June 3, 2024)
 * Shift4V2: Add new gateway based on SecurionPay adapter [heavyblade] #4860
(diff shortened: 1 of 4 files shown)


jcreiff (approved):
I think this looks good; the `add_exemption` logic is fairly minimal and only a small addition to the existing `verify` operation. The one edge case I thought we might need to be concerned about would be if `moto` _isn't_ sent, in case the new result of a `create_setup_intent` would be that we add `payment_method_options.card` as an empty object. I commented out the `moto: true` option in the remote test that was added, and the test still passes in that case, so perhaps it's nothing to be concerned about.

It doesn't seem like Stripe minds if we send `payment_method_options.card` as an empty object, but just in case, it might be nice to have 👀 from @activemerchant/marketplace-services for a second opinion

DustinHaefele:
I agree with @jcreiff, though it may not be an issue I think it would be easy to account for. 

On line 535 if you update it from `return unless options[:confirm]` => `return unless options[:confirm] && options[:moto]` that should cover any edge case that StripePI might not like. Or find the stripePI docs and ensure that the object can be blank without any issues.

jcreiff (approved):
Looks good!

PR #5161, opened 2024-07-01
CyberSource: Add addtional invoiceHeader fields

yunnydang:
Adds the merchant_descriptor_city and submerchant_id fields

Local:
5944 tests, 79910 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
159 tests, 945 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
136 tests, 675 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.8529% passed

### DIFF
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -96,6 +96,10 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
       ignore_cvv: 'true',
       commerce_indicator: 'internet',
       user_po: 'ABC123',
+      merchant_descriptor_country: 'US',
+      merchant_descriptor_state: 'NY',
+      merchant_descriptor_city: 'test123',
+      submerchant_id: 'AVSBSGDHJMNGFR',
       taxable: true,
       sales_slip_number: '456',
       airline_agent_code: '7Q',
@@ -129,6 +133,8 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
     + '1111111115555555222233101abcdefghijkl7777777777777777777777777promotionCde'
   end
 
+  # Scrubbing is working but may fail at the @credit_card.verification_value assertion
+  # if the the 3 digits are showing up in the Cybersource requestID
   def test_transcript_scrubbing
     transcript = capture_transcript(@gateway) do
       @gateway.purchase(@amount, @credit_card, @options)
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -129,14 +131,16 @@ def setup
     + '1111111115555555222233101abcdefghijkl7777777777777777777777777promotionCde'
   end
 
+  # Scrubbing is working but will fail at the @credit_card.verification_value assertion
```
jcreiff, line 134: Hey @yunnydang I think these are probably intermittent failures depending on what ID the Cybersource API returns. I just ran this test a couple of times with the CVV assertion and it passed. I think it's ok to leave your comment above to clarify that the test _may_ intermittently fail if the request ID contains the same substring, but let's not comment out the assertion itself.

We definitely want to know if CVV scrubbing isn't working 😅

jcreiff (changes requested):
Would like to re-assert the scrubbing line.

Also left a comment elsewhere about the potential need for `merchantDescriptorCountry` instead of (or in addition to?) `City`

jcreiff:
But otherwise I think this is looking good - the fields are added in the correct order to conform to the XSD!

test/remote/gateways/remote_cyber_source_test.rb
yunnydang: gotcha! I was going nuts with this since i ran it 6 times and it hit this error every single time LOL. thanks for checking!

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -129,6 +133,8 @@ def setup
...
```
jcreiff, line 136:
```suggestion
  # Scrubbing is working but may fail at the @credit_card.verification_value assertion
```

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -129,6 +133,8 @@ def setup
...
+  # becasue the the 3 digits are showing up in the rquestID
```
jcreiff, line 137:
```suggestion
  # if the 3 digits are showing up in the Cybersource requestID
```

jcreiff (approved):
Just a minor suggestion for the test comment, but other than that looks good to go!

PR #5166, opened 2024-07-08
Braintree: Pass overridden mid into client token for GS 3DS

sinourain:
Summary:
Add merchant_account_id for the Client Token Generate and returns a string which contains all authorization and configuration information that the client needs to initialize the client SDK to communicate with Braintree

[ECS-3617](https://spreedly.atlassian.net/browse/ECS-3617) [ECS-3607](https://spreedly.atlassian.net/browse/ECS-3607)

Remote Test:
Finished in 6.262003 seconds.
7 tests, 15 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
1.12 tests/s, 2.40 assertions/s

Unit Tests:
Finished in 38.744354 seconds.
5956 tests, 79952 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed
153.73 tests/s, 2063.58 assertions/s

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree/token_nonce.rb
+++ b/lib/active_merchant/billing/gateways/braintree/token_nonce.rb
@@ -18,10 +18,10 @@ module ActiveMerchant #:nodoc:
         "https://payments#{'.sandbox' if sandbox}.braintree-api.com/graphql"
       end
 
-      def create_token_nonce_for_payment_method(payment_method)
+      def create_token_nonce_for_payment_method(payment_method, options = {})
         headers = {
           'Accept' => 'application/json',
-          'Authorization' => "Bearer #{client_token}",
+          'Authorization' => "Bearer #{client_token(options)['authorizationFingerprint']}",
           'Content-Type' => 'application/json',
           'Braintree-Version' => '2018-05-10'
         }
@@ -34,9 +34,9 @@ module ActiveMerchant #:nodoc:
         return token, message
       end
 
-      def client_token
-        base64_token = @braintree_gateway.client_token.generate
-        JSON.parse(Base64.decode64(base64_token))['authorizationFingerprint']
+      def client_token(options = {})
+        base64_token = @braintree_gateway.client_token.generate({ merchant_account_id: options[:merchant_account_id] || @options[:merchant_account_id] }.compact)
+        JSON.parse(Base64.decode64(base64_token))
       end
 
       private
--- a/test/remote/gateways/remote_braintree_token_nonce_test.rb
+++ b/test/remote/gateways/remote_braintree_token_nonce_test.rb
@@ -26,8 +26,25 @@ class RemoteBraintreeTokenNonceTest < Test::Unit::TestCase
 
   def test_client_token_generation
     generator = TokenNonce.new(@braintree_backend)
-    token = generator.client_token
-    assert_not_nil token
+    client_token = generator.client_token
+    assert_not_nil client_token
+    assert_not_nil client_token['authorizationFingerprint']
+  end
+
+  def test_client_token_generation_with_mid
+    @options[:merchant_account_id] = '1234'
+    generator = TokenNonce.new(@braintree_backend, @options)
+    client_token = generator.client_token
+    assert_not_nil client_token
+    assert_equal client_token['merchantAccountId'], '1234'
+  end
+
+  def test_client_token_generation_with_a_new_mid
+    @options[:merchant_account_id] = '1234'
+    generator = TokenNonce.new(@braintree_backend, @options)
+    client_token = generator.client_token({ merchant_account_id: '5678' })
+    assert_not_nil client_token
+    assert_equal client_token['merchantAccountId'], '5678'
   end
 
   def test_successfully_create_token_nonce_for_bank_account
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/braintree/token_nonce.rb
```diff
@@ -35,8 +35,13 @@ def create_token_nonce_for_payment_method(payment_method)
       end
 
       def client_token
-        base64_token = @braintree_gateway.client_token.generate
-        JSON.parse(Base64.decode64(base64_token))['authorizationFingerprint']
+        parameters = options[:merchant_account_id] ? { merchant_account_id: options[:merchant_account_id] } : {}
```
Heavyblade, line 38: 💬 Comment/Style:

nothing big:

```ruby
base64_token = @braintree_gateway.client_token.generate({ merchant_account_id: options[:merchant_account_id] }.compact)
```

lib/active_merchant/billing/gateways/braintree/token_nonce.rb
```diff
@@ -35,8 +35,13 @@ def create_token_nonce_for_payment_method(payment_method)
...
+        base64_token = @braintree_gateway.client_token.generate(parameters)
+        JSON.parse(Base64.decode64(base64_token))
+      end
+
+      def authorization_fingerprint
```
Heavyblade, line 43: ❓  Question:

I get the point of calling it `authorization_fingerprint` instead of `client_token` but a method just get a key from a hash ?

test/remote/gateways/remote_braintree_token_nonce_test.rb
```diff
@@ -24,12 +24,20 @@ def setup
...
 
-  def test_client_token_generation
+  def test_authorization_fingerprint_generation
     generator = TokenNonce.new(@braintree_backend)
-    token = generator.client_token
+    token = generator.authorization_fingerprint
     assert_not_nil token
   end
 
+  def test_client_token_generation_with_mid
```
Heavyblade, line 33: ❓ Question:

so Nraintree Sandbox doesn't care about a valid mid ?, you can just send a '1234' and they are going to provide you with a client token any way ?

Heavyblade (commented):
Looks pretty much OK, @sinourain left some comments

Heavyblade (commented):
one general comment about the approach => 

Shouldn't the approach be to overwrite the `mid` that is being sent from CORE to initialize the library so you don't have to think on modifing the mid on the client token creation ?, because I'm seeing the library being initialized with the original `@options[:merchant_id]` so is possible to have the the instance intialized with mid 1234 and the create_token uses the mid 5678

lib/active_merchant/billing/gateways/braintree/token_nonce.rb
```diff
@@ -21,7 +21,7 @@ def url
       def create_token_nonce_for_payment_method(payment_method)
```
almalee24, line 21: I'm pretty sure this payment method needs to be updated to take in options because where would we be getting options[:merchant_account_id] from

lib/active_merchant/billing/gateways/braintree/token_nonce.rb
sinourain: Make sense, I will remove `authorization_fingerprint`

test/remote/gateways/remote_braintree_token_nonce_test.rb
sinourain: Correct, I tried passing multiple values ​​and got a successful result with every field sent including the known field

lib/active_merchant/billing/gateways/braintree/token_nonce.rb
sinourain: good catch, I added the changes :)

lib/active_merchant/billing/gateways/braintree/token_nonce.rb
```diff
@@ -34,9 +34,9 @@ def create_token_nonce_for_payment_method(payment_method)
         return token, message
       end
 
-      def client_token
-        base64_token = @braintree_gateway.client_token.generate
-        JSON.parse(Base64.decode64(base64_token))['authorizationFingerprint']
+      def client_token(options = {})
+        base64_token = @braintree_gateway.client_token.generate({ merchant_account_id: options[:merchant_account_id] || @options[:merchant_account_id] }.compact)
```
DustinHaefele, line 38: What is this generate method that is being called here?

DustinHaefele (commented):
I think this looks good. I got stuck in the middle of the review trying to figure out what is happening on a certain method so I left a question.

lib/active_merchant/billing/gateways/braintree/token_nonce.rb
sinourain: This is a call to the SDK gateway, here the documentation: https://developer.paypal.com/braintree/docs/reference/request/client-token/generate/ruby#client-token:-generate

DustinHaefele: Thank you that is super helpful! I'll finish my review now!

DustinHaefele (approved):
Thanks for pointing me to the SDK docs. This looks good to me!

PR #5168, opened 2024-07-09
SER-1327 enabling void call

Heavyblade:
## Summary:

Changes FlexCharge to add support for the void operation.
[SER-1327](https://spreedly.atlassian.net/browse/SER-1327)

### Remote Test:
Finished in 70.477035 seconds.
19 tests, 54 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications
100% passed

### Unit Tests:
Finished in 60.782586 seconds.
5956 tests, 79948 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/flex_charge.rb
+++ b/lib/active_merchant/billing/gateways/flex_charge.rb
@@ -18,7 +18,8 @@ module ActiveMerchant #:nodoc:
         refund: 'orders/%s/refund',
         store: 'tokenize',
         inquire: 'orders/%s',
-        capture: 'capture'
+        capture: 'capture',
+        void: 'orders/%s/cancel'
       }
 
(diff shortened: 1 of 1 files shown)


lib/active_merchant/billing/gateways/flex_charge.rb
```diff
@@ -68,7 +69,9 @@ def refund(money, authorization, options = {})
       end
 
       def void(money, authorization, options = {})
-        refund(money, authorization, options)
+        order_id, _currency = authorization.split('#')
+        self.money_format = :dollars
```
naashton, line 73: Do you want to set this at the class level?

naashton (approved):
Just one comment on the `self.money_format` placement, but otherwise LGTM

lib/active_merchant/billing/gateways/flex_charge.rb
Heavyblade: is no needed, didn't commit my latest change

PR #5173, opened 2024-07-12
FlexCharge fix `add_payment_method` bug

naashton:
Fix bug where `add_payment_method` was incorrectly returned early if the payment value was a token


Heavyblade (approved):
Looks 👍 @naashton

javierpedrozaing (approved):
Sweet and clean change LGTM

PR #5178, opened 2024-07-15
HPs: Update NetworkTokenizationCreditCard flow

almalee24:
Update NetworkTokenizationCreditCArd flow to now be under WalletData.

Remote
58 tests, 155 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 96.5517% passed

Unit
61 tests, 295 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/hps.rb
+++ b/lib/active_merchant/billing/gateways/hps.rb
@@ -17,11 +17,6 @@ module ActiveMerchant #:nodoc:
 
       PAYMENT_DATA_SOURCE_MAPPING = {
         apple_pay:        'ApplePay',
-        master:           'MasterCard 3DSecure',
-        visa:             'Visa 3DSecure',
-        american_express: 'AMEX 3DSecure',
-        discover:         'Discover 3DSecure',
-        android_pay:      'GooglePayApp',
         google_pay:       'GooglePayApp'
       }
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -254,27 +258,27 @@ def add_descriptor_name(xml, options)
...
         # PaymentDataSource is required when supplying the SecureECommerce data group,
         # and the gateway currently only allows the values within the mapping
-        return unless PAYMENT_DATA_SOURCE_MAPPING[three_d_secure[:source].to_sym]
+        return unless (three_d_secure = options[:three_d_secure])
+
+        three_d_secure[:source] ||= card_brand(card_or_token)
+
+        return unless (source = PAYMENT_DATA_SOURCE_MAPPING[three_d_secure[:source].to_sym])
 
         xml.hps :SecureECommerce do
```
DustinHaefele, line 280: I think this line needs to be the `:Secure3D` now according to the docs in the ticket.

DustinHaefele: Or will there be a separate ticket for the 3dsecure updates?

almalee24: Ah thanks for pointing that out, I definitely missed that bullet point.

DustinHaefele:
@almalee24 Have you rerun the remote tests with the latest updates?

almalee24:
> @almalee24 Have you rerun the remote tests with the latest updates?

Just saw this but yes I did

lib/active_merchant/billing/gateways/hps.rb
```diff
@@ -254,32 +253,24 @@ def add_descriptor_name(xml, options)
...
-        elsif options[:three_d_secure]
-          options[:three_d_secure][:source] ||= card_brand(card_or_token)
-          build_three_d_secure(xml, options[:three_d_secure])
+      def add_wallet_data(xml, payment_method, options)
+        return unless payment_method.is_a?(NetworkTokenizationCreditCard)
+
+        xml.hps :WalletData do
+          xml.hps :PaymentSource, PAYMENT_DATA_SOURCE_MAPPING[payment_method.source]
+          xml.hps :Cryptogram, payment_method.payment_cryptogram
+          xml.hps :ECI, payment_method.eci&.to_s&.sub!(/^0/, '') if payment_method.eci
```
aenand, line 262: Why not use `strip_leading_zero`?

DustinHaefele (approved):

lib/active_merchant/billing/gateways/hps.rb
almalee24: I can't even remember why but I'll update it do use strip_leading_zero

lib/active_merchant/billing/gateways/hps.rb
almalee24: Even though we have 3DS fields here it doesn't look like we support 3DS on our side for this gateway

PR #5182, opened 2024-07-17
Nuvei: Base Gateway Layout

javierpedrozaing:
Description
-------------------------
[SER-1351](https://spreedly.atlassian.net/browse/SER-1351)

This commit add a the basic structure for the new Nuvei Gateway

Unit test
-------------------------
Finished in 0.124334 seconds.

1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

8.04 tests/s, 16.09 assertions/s

Remote test
-------------------------
Finished in 0.602895 seconds.

1 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

1.66 tests/s, 3.32 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/nuvei.rb
+++ b/lib/active_merchant/billing/gateways/nuvei.rb
@@ -1,0 +1,238 @@
+module ActiveMerchant
+  module Billing
+    class NuveiGateway < Gateway
+      self.test_url = 'https://ppp-test.nuvei.com/ppp/api/v1'
+      self.live_url = 'https://secure.safecharge.com/ppp/api/v1'
+
+      self.supported_countries = %w[US CA IN NZ GB AU US]
+      self.default_currency = 'USD'
+      self.money_format = :cents
+      self.supported_cardtypes = %i[visa master american_express discover union_pay]
+      self.currencies_without_fractions = %w[CLP KRW JPY ISK MMK PYG UGX VND XAF XOF]
+      self.homepage_url = 'https://www.nuvei.com/'
+      self.display_name = 'Nuvei'
+
+      ENDPOINTS_MAPPING = {
+        authenticate: '/getSessionToken',
+        purchase: '/payment', # /authorize with transactionType: "Auth"
+        capture: '/settleTransaction',
+        refund: '/refundTransaction',
+        void: '/voidTransaction',
+        general_credit: '/payout'
+      }
+
+      def initialize(options = {})
+        requires!(options, :merchant_id, :merchant_site_id, :secret_key)
+        super
+        fetch_session_token unless session_token_valid?
+      end
+
+      def authorize(money, payment, options = {})
+        post = { transactionType: 'Auth' }
+
+        build_post_data(post, :authorize)
+        add_amount(post, money, options)
+        add_payment_method(post, payment)
+        add_address(post, payment, options)
+        add_customer_ip(post, options)
+
+        commit(:purchase, post)
+      end
+
+      def purchase(money, payment, options = {}); end
+
+      def capture(money, authorization, options = {})
+        post = { relatedTransactionId: authorization }
+
+        build_post_data(post, :capture)
+        add_amount(post, money, options)
+
+        commit(:capture, post)
+      end
+
+      def refund(money, authorization, options = {}); end
+
+      def void(authorization, options = {}); end
+
+      def credit(money, payment, options = {}); end
+
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
+          gsub(%r(("cardNumber\\?":\\?")[^"\\]*)i, '\1[FILTERED]').
+          gsub(%r(("cardCvv\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("merchantId\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("merchantSiteId\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("merchantKey\\?":\\?")\d+), '\1[FILTERED]')
+      end
+
+      private
+
+      def add_customer_ip(post, options)
+        return unless options[:ip]
+
+        post[:deviceDetails] = { ipAddress: options[:ip] }
+      end
+
+      def add_amount(post, money, options)
+        post[:amount] = amount(money)
+        post[:currency] = (options[:currency] || currency(money))
+      end
+
+      def credit_card_hash(payment)
+        {
+          cardNumber: payment.number,
+          cardHolderName: payment.name,
+          expirationMonth: format(payment.month, :two_digits),
+          expirationYear: format(payment.year, :four_digits),
+          CVV: payment.verification_value
+        }
+      end
+
+      def add_payment_method(post, payment)
+        if payment.is_a?(CreditCard)
+          post[:paymentOption] = { card: credit_card_hash(payment) }
+        else
+          post[:paymentOption] = { card: { cardToken: payment } }
+        end
+      end
+
+      def add_customer_names(full_name, payment_method)
+        split_names(full_name).tap do |names|
+          names[0] = payment_method&.first_name unless names[0].present? || payment_method.is_a?(String)
+          names[1] = payment_method&.last_name unless names[1].present? || payment_method.is_a?(String)
+        end
+      end
+
+      def add_address(post, payment, options)
+        return unless address = options[:billing_address] || options[:address]
+
+        first_name, last_name = add_customer_names(address[:name], payment)
+
+        post[:billingAddress] = {
+          email: options[:email],
+          country: address[:country],
+          phone: options[:phone] || address[:phone],
+          firstName: first_name,
+          lastName: last_name
+        }.compact
+      end
+
+      def current_timestamp
+        Time.now.utc.strftime('%Y%m%d%H%M%S')
+      end
+
+      def build_post_data(post, action)
+        post[:merchantId] = @options[:merchant_id]
+        post[:merchantSiteId] = @options[:merchant_site_id]
+        post[:timeStamp] = current_timestamp.to_i
+        post[:clientRequestId] = SecureRandom.uuid
+        post[:clientUniqueId] = SecureRandom.hex(16)
+      end
+
+      def calculate_checksum(post, action)
+        common_keys = %i[merchantId merchantSiteId clientRequestId]
+        keys = case action
+               when :authenticate
+                 [:timeStamp]
+               when :capture
+                 %i[clientUniqueId amount currency relatedTransactionId timeStamp]
+               else
+                 %i[amount currency timeStamp]
+               end
+
+        to_sha = post.values_at(*common_keys.concat(keys)).push(@options[:secret_key]).join
+        Digest::SHA256.hexdigest(to_sha)
+      end
+
+      def send_session_request(post)
+        post[:checksum] = calculate_checksum(post, 'authenticate')
+        response = parse(ssl_post(url(:authenticate), post.to_json, headers)).with_indifferent_access
+        expiration_time = post[:timeStamp]
+        @options[:session_token] = response.dig('sessionToken')
+        @options[:token_expires] = expiration_time
+
+        Response.new(
+          response[:sessionToken].present?,
+          message_from(response),
+          response,
+          test: test?,
+          error_code: response[:errCode]
+        )
+      end
+
+      def fetch_session_token(post = {})
+        build_post_data(post, :authenticate)
+        send_session_request(post)
+      end
+
+      def session_token_valid?
+        return false unless @options[:session_token] && @options[:token_expires]
+
+        (Time.now.utc.to_i - @options[:token_expires].to_i) < 900 # 15 minutes
+      end
+
+      def commit(action, post, authorization = nil, method = :post)
+        post[:sessionToken] = @options[:session_token] unless action == :capture
+        post[:checksum] = calculate_checksum(post, action)
+
+        response = parse(ssl_request(method, url(action, authorization), post.to_json, headers))
+
+        Response.new(
+          success_from(response),
+          message_from(response),
+          response,
+          authorization: authorization_from(action, response, post),
+          test: test?,
+          error_code: error_code_from(action, response)
+        )
+      rescue ResponseError => e
+        response = parse(e.response.body)
+        @options[:session_token] = '' if e.response.code == '401'
+
+        Response.new(false, message_from(response), response, test: test?)
+      end
+
+      def url(action, id = nil)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,137 @@
...
+      def fetch_session_token
+        post = {
+          'merchantId' => @options[:merchant_id],
+          'merchantSiteId' => @options[:merchant_site_id],
+          'merchantKey' => @options[:secret_key],
+          'clientRequestId' => SecureRandom.uuid,
+          'timeStamp' => Time.now.utc.strftime('%Y%m%d%H%M%S')
+        }
+
+        post.tap { |p| p['checksum'] = calculate_checksum(p) }
```
Heavyblade, line 77: ❓ Question:

what is the reason to use the tap method instead of a direct assignation like `p['checksum'] = calculate_checksum(p)`, something weird happens if you do ?

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,137 @@
...
+
+        response = JSON.parse(ssl_post(url(:authenticate), post.to_json, headers)).with_indifferent_access
```
Heavyblade, line 79: 💬 Comment:

You already have a `parse` method that deals with possible exception, I would be wise to make use of it here.

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,137 @@
...
+        current_time = Time.now
+        expiration_time = current_time + (15 * 60) # 15 minutes in seconds
```
Heavyblade, line 81: ⚠️ Warning:

What I can see on the intention is that current_time should be a Unix Epoch number but here you are storing a date, so in the future you are going to compare the string "2024-07-18 15:31:03.326644 -0500" witht epoch time that `DateTime.now.strftime('%Q').to_i` produces

tmux

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      def capture(money, authorization, options = {}); end
+
+      def refund(money, authorization, options = {}); end
+
+      def void(authorization, options = {}); end
+
+      def credit(money, payment, options = {}); end
+
+      def scrub(transcript)
```
Heavyblade, line 41: 💬 Comment:

maybe scrubbing merchantId, merchantSiteId and merchantKey from the the getSession call is a good idea.

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,137 @@
...
+            end
+          end
+        end
+      end
+
+      def session_token_valid?
+        @options[:session_token] && @options[:token_expires] && Time.now <= @options[:token_expires]
+      end
+
+      def fetch_session_token
```
Heavyblade, line 68: 💬 Comment:

Unlike FlexCharge this method needs to do more work to build the request, maybe a `build_xxxxx` method would be nice to separte concerns and also to add proper unit testing for the code that builds the request body.

Heavyblade (commented):
@javierpedrozaing, I did my review just focusing on the getSession call and left some comments for you

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: yes you are right, I will update it

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: that is because I am passing the post values to the `calculate_checksum` function, if I use a direct assignation without use a tap I cannot get the merchantId, merchantSiteId, clientRequestId and others one into that function.

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: yes, you are right I will update it

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: good Catch , thanks I will update it

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: yes

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: Good suggestion, I will take a look on it, thanks

test/remote/gateways/remote_nuvei_test.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def test_successful_authorize
+    response = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success response
+
+    assert_not_nil response.params[:transactionId]
+    assert_match 'SUCCESS', response.message
+    assert_match 'APPROVED', response.params['transactionStatus']
+  end
```
Heavyblade, line 40: 💬 Comment:

Same here, we need to confirm how request with errors are handled, and also confirm that token renovation works as expected.

test/unit/gateways/nuvei_test.rb
```diff
@@ -0,0 +1,115 @@
...
+    action = :authenticate
+    assert_equal @gateway.send(:url, action), "#{@gateway.test_url}/getSessionToken"
+  end
+
+  def test_scrub
+    assert @gateway.supports_scrubbing?
+    assert_equal @gateway.scrub(pre_scrubbed), post_scrubbed
+  end
+
+  def test_successful_authorize
```
Heavyblade, line 40: 🔴 Warning 🔴 

Even when this test is working, is almost a 'false positive', you are using the same stub for the token request and the authorize, is working basically because the token request just checks for the presense of the `sessionToken` attribute.

Heavyblade (commented):
@javierpedrozaing tons of nice work here, left you some comments and questions.

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,265 @@
...
+
+      def initialize(options = {})
+        requires!(options, :merchant_id, :merchant_site_id, :secret_key)
+        super
+      end
+
+      def authorize(money, payment, options = {})
+        post = {}
+        post[:transactionType] = 'Auth'
+        post.merge!(build_post_data(:authorize))
```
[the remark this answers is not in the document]
javierpedrozaing: There is not a special reason for this, I will update it to be consistent with the patern, thanks

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,265 @@
...
+        add_customer_names(billing_address, address[:name], payment) if address[:name]
+        add_customer_phone(billing_address, options)
+        post[:billingAddress] = billing_address
+      end
+
+      def current_timestamp
+        Time.now.utc.strftime('%Y%m%d%H%M%S')
+      end
+
+      def build_post_data(action)
```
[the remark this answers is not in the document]
javierpedrozaing: yes, less code, I will update it

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,265 @@
...
+        first_name, last_name = split_names(full_name).tap do |names|
+          names[0] = payment_method&.first_name unless names[0].present? || payment_method.is_a?(String)
+          names[1] = payment_method&.last_name unless names[1].present? || payment_method.is_a?(String)
+        end
+
+        billing_address[:firstName] = first_name
+        billing_address[:lastName] = last_name
+      end
+
+      def add_customer_phone(billing_address, options)
```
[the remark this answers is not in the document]
javierpedrozaing: you are right, let me update it

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,265 @@
...
+          error_code: response[:errCode]
+        )
+      end
+
+      def fetch_session_token
+        post = build_post_data(:authenticate)
+        send_request(post)
+      end
+
+      def session_token_valid?
```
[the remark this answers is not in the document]
javierpedrozaing: Good point, I will check it

test/remote/gateways/remote_nuvei_test.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.authorize(@amount, @credit_card, @options)
+    end
+
+    @gateway.scrub(transcript)
+  end
+
+  def test_successful_session_token_generation
```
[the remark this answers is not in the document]
javierpedrozaing: yes, thanks

test/remote/gateways/remote_nuvei_test.rb
javierpedrozaing: I will add some failure tests

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,265 @@
...
+        {
+          merchantId: @options[:merchant_id],
+          merchantSiteId: @options[:merchant_site_id],
+          timeStamp: current_timestamp,
+          clientRequestId:  client_request_id,
+          clientUniqueId: client_unique_id
+        }.compact
+      end
+
+      def calculate_checksum(post, action)
```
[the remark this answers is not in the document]
javierpedrozaing: Thanks, I will add it

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,255 @@
...
+        when :authenticate
+          Digest::SHA256.hexdigest("#{post[:merchantId]}#{post[:merchantSiteId]}#{post[:clientRequestId]}#{post[:timeStamp]}#{@options[:secret_key]}")
+        when :capture
+          Digest::SHA256.hexdigest("#{post[:merchantId]}#{post[:merchantSiteId]}#{post[:clientRequestId]}#{post[:clientUniqueId]}#{post[:amount]}#{post[:currency]}#{post[:relatedTransactionId]}#{post[:timeStamp]}#{@options[:secret_key]}")
+        else
+          Digest::SHA256.hexdigest("#{post[:merchantId]}#{post[:merchantSiteId]}#{post[:clientRequestId]}#{post[:amount]}#{post[:currency]}#{post[:timeStamp]}#{@options[:secret_key]}")
+        end
+      end
+
+      def send_request(post)
```
naashton, line 150: What is the difference between `send_request` and `api_request`?

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,255 @@
...
+
+      def commit(action, post, authorization = nil, method = :post)
+        MultiResponse.run do |r|
+          r.process { fetch_session_token } unless session_token_valid?
+          r.process do
+            api_request(action, post, authorization, method).tap do |response|
+              response.params.merge!(@options.slice(:session_token, :token_expires)) if @options[:valid_credentials]
+            end
+          end
+        end
```
naashton, line 189: I think we should try to keep the established pattern for the `commit` method if we can. Typically, it handles the request, parses the response, and builds the standardized response. I wonder if the logic to refresh the session token would be better served in the `init` for the gateway class.

naashton (changes requested):
A lot of work in here, so great job with this initial buildout. I had have a few design related concerns, specifically around the request and commit's.

I think, if we can do the following, we'll be in good shape: 

1. Consolidate the various `Response.new` to a helper method, this would be helpful for debugging in the future.
2. Perform the session token request/refresh in the `initialize` of the gateway. This will help us in 3DSGS implementation where the code stack may not hit those request/commit methods built in the AM adapter.

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,255 @@
...
+      def supports_scrubbing?
+        true
+      end
+
+      def scrub(transcript)
+        transcript.
+          gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
+          gsub(%r(("cardNumber\\?":\\?")[^"\\]*)i, '\1[FILTERED]').
+          gsub(%r(("cardCvv\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("cardHolderName\\?":\\?")\d+), '\1[FILTERED]').
```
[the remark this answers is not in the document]
javierpedrozaing: thanks, I will update it

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: `send_request` is to send the session token request, but I will do some updates

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: yes, you're right we can add the session token request in the init, let me update it, thanks

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -0,0 +1,255 @@
...
+      end
+
+      def send_request(post)
+        post[:checksum] = calculate_checksum(post, 'authenticate')
+
+        response = parse(ssl_post(url(:authenticate), post.to_json, headers)).with_indifferent_access
+        expiration_time = post[:timeStamp]
+        @options[:session_token] = response.dig('sessionToken')
+        @options[:token_expires] = expiration_time
+        @options[:valid_credentials] = true
```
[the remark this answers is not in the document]
javierpedrozaing: yes, it s to ensure a valid session token

javierpedrozaing:
> A lot of work in here, so great job with this initial buildout. I had have a few design related concerns, specifically around the request and commit's.
> 
> I think, if we can do the following, we'll be in good shape:
> 
> 1. Consolidate the various `Response.new` to a helper method, this would be helpful for debugging in the future.
> 2. Perform the session token request/refresh in the `initialize` of the gateway. This will help us in 3DSGS implementation where the code stack may not hit those request/commit methods built in the AM adapter.

Thanks Nick I did the updates to avoid use multi response on the commit method and request the session token on the init method.

lib/active_merchant/billing/gateways/nuvei.rb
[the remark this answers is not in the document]
javierpedrozaing: I have decided remove, since I am no using it. only I am checking precense of `session_token` and `token_expires`

naashton (approved):
This looks good! Make sure to add a changelog for this one 🚢

PR #5183, opened 2024-07-17
Datatrans: Add TPV

gasb150:
Summary:
-------------------------
Include Store and Unstore Methods in Datatrans
gateway to support Third Party Token.

[SER-1395](https://spreedly.atlassian.net/browse/SER-1395)

Tests
-------------------------

Remote Test:
-------------------------
Finished in 31.477035 seconds.
25 tests, 72 assertions, 0 failures, 0 errors, 0 pendings, 1 omissions, 0 notifications
100% passed

Unit Tests:
-------------------------
Finished in 0.115603 seconds.
29 tests, 165 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/datatrans.rb
+++ b/lib/active_merchant/billing/gateways/datatrans.rb
@@ -1,8 +1,8 @@
 module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class DatatransGateway < Gateway
-      self.test_url = 'https://api.sandbox.datatrans.com/v1/transactions/'
-      self.live_url = 'https://api.datatrans.com/v1/transactions/'
+      self.test_url = 'https://api.sandbox.datatrans.com/v1/'
+      self.live_url = 'https://api.datatrans.com/v1/'
 
       self.supported_countries = %w(CH GR US) # to confirm the countries supported.
       self.default_currency = 'CHF'
@@ -72,6 +72,28 @@ module ActiveMerchant #:nodoc:
         commit('cancel', post, { transaction_id: transaction_id })
       end
 
+      def store(payment_method, options = {})
+        exp_year = format(payment_method.year, :two_digits)
+        exp_month = format(payment_method.month, :two_digits)
+
+        post = {
+          requests: [
+            {
+              type: 'CARD',
+              pan: payment_method.number,
+              expiryMonth: exp_month,
+              expiryYear: exp_year
+            }
+          ]
+        }
+        commit('tokenize', post, { expiry_month: exp_month, expiry_year: exp_year })
+      end
+
+      def unstore(authorization, options = {})
+        data_alias = authorization.split('|')[2].split('-')[0]
+        commit('delete_alias', {}, { alias_id: data_alias }, :delete)
+      end
+
       def supports_scrubbing?
         true
       end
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/datatrans.rb
```diff
@@ -196,26 +231,36 @@ def headers
...
       def success_from(action, response)
         case action
         when 'authorize', 'credit'
-          true if response.include?('transactionId') && response.include?('acquirerAuthorizationCode')
+          response.include?('transactionId') && response.include?('acquirerAuthorizationCode')
         when 'settle', 'cancel'
-          true if response.dig('response_code') == 204
+          response.dig('response_code') == 204
+        when 'tokenize'
+          response['responses'][0].include?('alias') && response['overview']['failed'] == 0
```
Heavyblade, line 251: 💬 Comment/Readability:

just matter of taste but `has_key?` is more explicit for hashes than `include?` which is normally asociated with checking arrays elements.

Heavyblade: maybe more robust conditional:

```ruby
response.dig('responses', 0, 'alias') && response.dig('overview', 'failed') == 0
```

lib/active_merchant/billing/gateways/datatrans.rb
```diff
@@ -72,6 +72,28 @@ def void(authorization, options = {})
...
+          ]
+        }
+        commit('tokenize', post, { expiry_month: exp_month, expiry_year: exp_year })
+      end
+
+      def unstore(authorization, options = {})
+        data_alias = authorization.split('|')[2]
+        commit('delete_alias', {}, { alias_id: data_alias }, :delete)
+      end
+
```
Heavyblade, line 96: ❓ Question:

Just asking should we scrub the alias ?

lib/active_merchant/billing/gateways/datatrans.rb
```diff
@@ -196,26 +231,36 @@ def headers
...
         else
           false
         end
       end
 
-      def authorization_from(response)
-        auth = [response['transactionId'], response['acquirerAuthorizationCode']].join('|')
-        return auth unless auth == '|'
+      def authorization_from(response, action, options)
+        token_array = [response.dig('responses', 0, 'alias'), options[:expiry_month], options[:expiry_year]] if action == 'tokenize'
```
Heavyblade, line 260: 💬 👃 Comment/Code smell:

Event when implementation seems correct, you can see that token_array should be just one value maybe something like "<token_xxxxxx>-01-2025" so when you see the "xxxxxx|xxxxx|<token_xxxxxx>-01-2025" you can know that those 3 values are independent from each other, an no interpretation is needed.

lib/active_merchant/billing/gateways/datatrans.rb
```diff
@@ -86,10 +108,23 @@ def scrub(transcript)
       private
 
       def add_payment_method(post, payment_method)
-        card = build_card(payment_method)
+        card = {}
```
Heavyblade, line 111: 💬 Comment:

I see this as an opportunity to refactor `add_payment_method` to be more explicit/declarative on the payment method construction, something like:

```ruby
def add_payment_method(post, payment_method)
  case payment_method
  when String
    token, exp_month, exp_year = payment_method.split('|')[2..4]
    card = {
      type: 'ALIAS',
      alias: token,
      expiryMonth: exp_month,
      expiryYear: exp_year
    }
  when NetworkTokenizationCreditCard
    card = {
      type: DEVICE_SOURCE[payment_method.source] ? 'DEVICE_TOKEN' : 'NETWORK_TOKEN',
      tokenType: DEVICE_SOURCE[payment_method.source] || CREDIT_CARD_SOURCE[card_brand(payment_method)],
      token: payment_method.number,
      cryptogram: payment_method.payment_cryptogram,
      expiryMonth: format(payment_method.month, :two_digits),
      expiryYear: format(payment_method.year, :two_digits)
    }
  when CreditCard
    card = {
      type: 'PLAIN',
      number: payment_method.number,
      cvv: payment_method.verification_value.to_s,
      expiryMonth: format(payment_method.month, :two_digits),
      expiryYear: format(payment_method.year, :two_digits)
    }
  end

  post[:card] = card
end
```

test/remote/gateways/remote_datatrans_test.rb
```diff
@@ -183,6 +183,34 @@ def test_successful_void
     assert_equal response.authorization, nil
   end
 
+  def test_successful_store_purchase_unstore_flow
```
Heavyblade, line 186: 💬 Comment:

Event when I really really like the confidence that this approach produces 👍 , having separted tests for store and unstore would be great to have more control on how and why a test could be failing.

test/unit/gateways/datatrans_test.rb
```diff
@@ -283,10 +320,14 @@ def test_scrub
   end
 
   def test_authorization_from
-    assert_equal '1234|9248', @gateway.send(:authorization_from, { 'transactionId' => '1234', 'acquirerAuthorizationCode' => '9248' })
-    assert_equal '1234|', @gateway.send(:authorization_from, { 'transactionId' => '1234' })
-    assert_equal '|9248', @gateway.send(:authorization_from, { 'acquirerAuthorizationCode' => '9248' })
-    assert_equal nil, @gateway.send(:authorization_from, {})
+    assert_equal '1234|9248|', @gateway.send(:authorization_from, { 'transactionId' => '1234', 'acquirerAuthorizationCode' => '9248' }, '', {})
```
Heavyblade, line 323: 💬 Comment

Niiiice coverture here 💪

Heavyblade (approved):
Overall functionality looks good @gasb150 , left some style/structure comments

lib/active_merchant/billing/gateways/datatrans.rb
gasb150: I was dubbing about it hehe

lib/active_merchant/billing/gateways/datatrans.rb
gasb150: I'm not sure, some gateways don't do it. Considering we are also passing expiration info. Filter the alias could be better.

PR #5184, opened 2024-07-17
Adyen: Add header fields to response body

yunnydang:
This adds the header fields and values to the response body object in the parse method. By using the handle_response override, we can set an instance variable of the headers and merge it into the response body. The 12 remote test failures are also the same ones that fail on the master branch. 

Local:
5961 tests, 79980 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
120 tests, 630 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
144 tests, 471 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.6667% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -755,10 +755,34 @@ module ActiveMerchant #:nodoc:
         post[:metadata].merge!(options[:metadata]) if options[:metadata]
       end
 
+      def add_header_fields(response)
+        return unless @response_headers.present?
+
+        headers = {}
+        headers['response_headers'] = {}
+        headers['response_headers']['transient_error'] = @response_headers['transient-error'] if @response_headers['transient-error']
+
+        response.merge!(headers)
+      end
+
       def parse(body)
         return {} if body.blank?
 
-        JSON.parse(body)
+        response = JSON.parse(body)
+        add_header_fields(response)
+        response
+      end
+
+      # Override the regular handle response so we can access the headers
+      # set header fields and values so we can add them to the response body
+      def handle_response(response)
+        @response_headers = response.each_header.to_h if response.respond_to?(:header)
+        case response.code.to_i
+        when 200...300
+          response.body
+        else
+          raise ResponseError.new(response)
+        end
       end
 
       def commit(action, parameters, options)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,6 +28,7 @@
 * CheckoutV2: Update 3DS message & error code [almalee24] #5177
 * DecicirPlus: Update error_message to add safety navigator [almalee24] #5187
 * Elavon: Add updated stored credential version [almalee24] #5170
+* Adyen: Add header fields to response body [yunnydang] #5184
 
 == Version 1.136.0 (June 3, 2024)
 * Shift4V2: Add new gateway based on SecurionPay adapter [heavyblade] #4860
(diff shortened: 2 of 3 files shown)


aenand:
FYI @yunnydang you don't need to run local tests locally, the GHA will do it :)

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -755,10 +755,22 @@ def add_metadata(post, options = {})
         post[:metadata].merge!(options[:metadata]) if options[:metadata]
       end
 
-      def parse(body)
-        return {} if body.blank?
+      def parse(response)
+        return {} if response.blank?
```
DustinHaefele, line 759: Couple thoughts here. I think that the parse call should still take body as the argument since that is what is being passed in, and we are adding the headers. Then I think that we should add the headers under their own tag in the response so that they aren't just mixed into the parsed body without any differentiation. 

I don't see any functional problems with the implementation though.

```suggestion
      def parse(body)
        return {} if body.blank?
        
        response = JSON.parse(body)
        response.merge!({response_headers: @response_headers}) if @response_headers.present?
        response
```

yunnydang: thanks @DustinHaefele for the reccomendation! I was wondering would it be better if we specifically return only the field that is being requested "transient-error" or return the and merge all header fields. Would there be any security related risk to returning the all of the header fields? i was mulling over this when working through this however i'd rather return the all the fields in case we need to grab them in the future.

DustinHaefele: Yeah that is a good point to. I don't think there is much risk, but since it would be pretty easy to add more headers in the future if others are requested maybe it is a good idea to only grab the one that we actually care about.

DustinHaefele (commented):
Any thoughts on ways that we could test this? I know it is hard because we can't force that header in remote tests, but maybe stubbing it somehow in unit tests? I don't know if this is the best because it doesn't actually get the variable from actual response headers but even something like this might be better than no added testing. 

```
  def test_successful_authorize_with_transient_error
    @gateway.instance_variable_set(:@response_headers, {'transient-error' => 'error_will_robinson'})
    @gateway.expects(:ssl_post).returns(successful_authorize_response)

    response = @gateway.authorize(@amount, @credit_card, @options)
    assert_success response

    assert response.params['response_headers']['transient_error'], 'error_will_robinson'
    assert response.test?
  end
```

jcreiff (commented):
I left a similar comment [here](https://github.com/activemerchant/active_merchant/pull/5185) on the Stripe version of this change, but I think we may need to be able to pull these headers for transactions that are returned with 400-level HTTP codes too.

The thing that makes it a little less clear for Adyen (when compared to Stripe) is that this gateway seems to return HTTP 200 responses for most garden-variety transaction failures. However, the way that their [docs](https://docs.adyen.com/development-resources/api-idempotency/#transient-errors) talk about `transient-error`, it makes me wonder if the rare use case where the field is returned in headers, we might be more likely to see a non-200 status code

jcreiff:
I think this looks good - I am trying to poke holes in the logic to see if there's ever a situation where we might unexpectedly end up with a `NoMethodError` but since we're only setting `@response_headers` if the response responds to `:header`, that means we'll always either have a Hash to access in `add_header_fields`, or we'll just skip that method entirely. I'll ✅ but also re-request review from the others who have taken a peek, since these changes will come into play on every transaction.

jcreiff (approved):

DustinHaefele (approved):
This looks good to me. 

The only thought I have is when we deploy this code in Spreedly we should be monitoring Adyen transactions very closely for failure since this does touch every transaction. We could also put it behind a flag for instant revert-ability but I don't think it it is necessary.

PR #5185, opened 2024-07-17
Stripe and Stripe PI: add headers to response body

yunnydang:
This adds some of the header fields to the response body object using the handle_response override.

Local:
5962 tests, 79990 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Stripe unit:
146 tests, 769 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Stripe remote:
77 tests, 328 assertions, 10 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
87.013% passed

Stripe PI Unit:
61 tests, 315 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Stripe PI Remote:
95 tests, 427 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.8421% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -29,6 +29,7 @@
 * DecicirPlus: Update error_message to add safety navigator [almalee24] #5187
 * Elavon: Add updated stored credential version [almalee24] #5170
 * Adyen: Add header fields to response body [yunnydang] #5184
+* Stripe and Stripe PI: Add header fields to response body [yunnydang] #5185
 
 == Version 1.136.0 (June 3, 2024)
 * Shift4V2: Add new gateway based on SecurionPay adapter [heavyblade] #4860
--- a/lib/active_merchant/billing/gateways/stripe.rb
+++ b/lib/active_merchant/billing/gateways/stripe.rb
@@ -617,8 +617,21 @@ module ActiveMerchant #:nodoc:
         post[:radar_options] = radar_options unless radar_options.empty?
       end
 
+      def add_header_fields(response)
+        return unless @response_headers.present?
+
+        headers = {}
+        headers['response_headers'] = {}
+        headers['response_headers']['idempotent_replayed'] = @response_headers['idempotent-replayed'] if @response_headers['idempotent-replayed']
+        headers['response_headers']['stripe_should_retry'] = @response_headers['stripe-should-retry'] if @response_headers['stripe-should-retry']
+
+        response.merge!(headers)
+      end
+
       def parse(body)
-        JSON.parse(body)
+        response = JSON.parse(body)
+        add_header_fields(response)
+        response
       end
 
       def post_data(params)
@@ -750,6 +763,18 @@ module ActiveMerchant #:nodoc:
 
       def success_from(response, options)
         !response.key?('error') && response['status'] != 'failed'
+      end
+
+      # Override the regular handle response so we can access the headers
+      # set header fields and values so we can add them to the response body
+      def handle_response(response)
+        @response_headers = response.each_header.to_h if response.respond_to?(:header)
+        case response.code.to_i
+        when 200...300
+          response.body
+        else
+          raise ResponseError.new(response)
+        end
       end
 
       def response_error(raw_response)
--- a/test/remote/gateways/remote_stripe_payment_intents_test.rb
+++ b/test/remote/gateways/remote_stripe_payment_intents_test.rb
@@ -380,6 +380,17 @@ class RemoteStripeIntentsTest < Test::Unit::TestCase
     refute purchase.params.dig('error', 'payment_intent', 'charges', 'data')[0]['captured']
   end
 
+  def test_unsuccessful_purchase_returns_header_response
+    options = {
+      currency: 'GBP',
+      customer: @customer
+    }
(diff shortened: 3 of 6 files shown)


jcreiff (changes requested):
Hey @yunnydang this is looking promising! I think it may need a little bit of tweaking though:

The way it is currently written, we will only set `@response_headers` when the response is successful (in the HTTP 200 range). Otherwise, we skip the `response.each_header.to_h` logic.

My understanding of these fields we want to surface is that they will be more useful for failed transactions (ie, I don't think anyone is concerned about whether or not they should retry after a successful purchase 😁), so I think we need to be able to grab the headers when we hit the `else raise ResponseError.new` path too 🤔 

I pulled down your branch and did some poking around with `byebug` to confirm my suspicion about the above, and sure enough, if I change the payment method to `@declined_payment_method` in your remote test, then we do skip over the `when 200...300` block of the case statement in `handle_response`, but **only** on the POST to `/payment_intents`. 

However! The test still passes in that scenario because that purchase request is preceded by a POST to `/payment_methods` (or `/tokens`), and when we handle that initial response, we are storing the headers in the `@response_headers` instance variable. This means those headers are available the second time we go through `add_header_fields`, even though we haven't updated `@response_headers` with info from the second response.  A randomly selected txn id that illustrates this kind of  sequence is `8Q3Hpae1p6HNz7hLLzB1CJcFkAR`, but I think this would happen to any transaction where we receive a `200` followed by a `402`

jcreiff:
One more thought - should we also add test coverage for Stripe (not PI) since `handle_response` and `parse` are shared logic between the two gateways?

yunnydang:
> Hey @yunnydang this is looking promising! I think it may need a little bit of tweaking though:
> 
> The way it is currently written, we will only set `@response_headers` when the response is successful (in the HTTP 200 range). Otherwise, we skip the `response.each_header.to_h` logic.
> 
> My understanding of these fields we want to surface is that they will be more useful for failed transactions (ie, I don't think anyone is concerned about whether or not they should retry after a successful purchase 😁), so I think we need to be able to grab the headers when we hit the `else raise ResponseError.new` path too 🤔
> 
> I pulled down your branch and did some poking around with `byebug` to confirm my suspicion about the above, and sure enough, if I change the payment method to `@declined_payment_method` in your remote test, then we do skip over the `when 200...300` block of the case statement in `handle_response`, but **only** on the POST to `/payment_intents`.
> 
> However! The test still passes in that scenario because that purchase request is preceded by a POST to `/payment_methods` (or `/tokens`), and when we handle that initial response, we are storing the headers in the `@response_headers` instance variable. This means those headers are available the second time we go through `add_header_fields`, even though we haven't updated `@response_headers` with info from the second response. A randomly selected txn id that illustrates this kind of sequence is `8Q3Hpae1p6HNz7hLLzB1CJcFkAR`, but I think this would happen to any transaction where we receive a `200` followed by a `402`

Thanks for pointing out this insight @jcreiff ! Im thinking that with the use case like you mentioned where its most useful for failing transactions, do you think it might be better to move the logic to grab headers only for failing transactions? I think that would be something we can explore instead of adding headers to both fail/successful

yunnydang:
> > Hey @yunnydang this is looking promising! I think it may need a little bit of tweaking though:
> > The way it is currently written, we will only set `@response_headers` when the response is successful (in the HTTP 200 range). Otherwise, we skip the `response.each_header.to_h` logic.
> > My understanding of these fields we want to surface is that they will be more useful for failed transactions (ie, I don't think anyone is concerned about whether or not they should retry after a successful purchase 😁), so I think we need to be able to grab the headers when we hit the `else raise ResponseError.new` path too 🤔
> > I pulled down your branch and did some poking around with `byebug` to confirm my suspicion about the above, and sure enough, if I change the payment method to `@declined_payment_method` in your remote test, then we do skip over the `when 200...300` block of the case statement in `handle_response`, but **only** on the POST to `/payment_intents`.
> > However! The test still passes in that scenario because that purchase request is preceded by a POST to `/payment_methods` (or `/tokens`), and when we handle that initial response, we are storing the headers in the `@response_headers` instance variable. This means those headers are available the second time we go through `add_header_fields`, even though we haven't updated `@response_headers` with info from the second response. A randomly selected txn id that illustrates this kind of sequence is `8Q3Hpae1p6HNz7hLLzB1CJcFkAR`, but I think this would happen to any transaction where we receive a `200` followed by a `402`
> 
> Thanks for pointing out this insight @jcreiff ! Im thinking that with the use case like you mentioned where its most useful for failing transactions, do you think it might be better to move the logic to grab headers only for failing transactions? I think that would be something we can explore instead of adding headers to both fail/successful

thinking more on it, i can just return it on both successful and failures as we are currently digging for a specific header field. I dont think it'll hurt if we implement it this way

jcreiff (approved):
Thanks for adding the additional tests!

yunnydang:
thanks Joe! @DustinHaefele would you mind peeking when you have the chance, much appreciated!

DustinHaefele (approved):

PR #5187, opened 2024-07-18
DecidirPlus: Update error_message to add safety navigator

almalee24:
Add safety navigator to error_message to prevent NoMethod errors and create a validation_error_message to construct a new error message.

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir_plus.rb
+++ b/lib/active_merchant/billing/gateways/decidir_plus.rb
@@ -321,8 +321,11 @@ module ActiveMerchant #:nodoc:
         return error_code_from(response) unless validation_errors = response.dig('validation_errors')
 
         validation_errors = validation_errors[0]
+        message = "#{validation_errors&.dig('code')}: #{validation_errors&.dig('param')}"
+        return message unless message == ': '
 
-        "#{validation_errors.dig('code')}: #{validation_errors.dig('param')}"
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -321,8 +321,15 @@ def error_message(response)
         return error_code_from(response) unless validation_errors = response.dig('validation_errors')
 
         validation_errors = validation_errors[0]
+        message = "#{validation_errors&.dig('code')}: #{validation_errors&.dig('param')}"
+        message == ': ' ? nil : message
```
DustinHaefele, line 325: Do we need lines 323-325 since on line 286 we are now calling `validation_error_message`. Or is error message called somewhere else as well?

DustinHaefele (commented):
I think it looks pretty good, but I left one question.

lib/active_merchant/billing/gateways/decidir_plus.rb
almalee24: I could update error_message do be like validation_error_message. That would make a lot more sense

almalee24: I didn't want to change how we format the current error message because that will probably need customer comm.

DustinHaefele (approved):
Looks great. The latest update is clean!

PR #5192, opened 2024-07-24
Rapyd: Add support for save_payment_method field

rachelkirk:
CER-1613

Top level boolean field that will trigger payment method to be saved.

Remote Tests:
54 tests, 152 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 94.4444% passed
* I fixed a few failing tests, so less failing on master.

Also, more of them seem fixable but I started to gown down a docs rabbit hole that is not in the scope of this ticket.

### DIFF
--- a/test/remote/gateways/remote_rapyd_test.rb
+++ b/test/remote/gateways/remote_rapyd_test.rb
@@ -145,6 +145,14 @@ class RemoteRapydTest < Test::Unit::TestCase
     assert_equal 'SUCCESS', response.message
   end
 
+  def test_successful_purchase_with_save_payment_method
+    @options[:pm_type] = 'gb_visa_mo_card'
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(save_payment_method: true))
+    assert_success response
+    assert_equal 'SUCCESS', response.message
+    assert_equal true, response.params['data']['save_payment_method']
+  end
+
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_rapyd_test.rb
```diff
@@ -145,6 +145,13 @@ def test_successful_purchase_with_reccurence_type
     assert_equal 'SUCCESS', response.message
   end
 
+  def test_successful_purchase_with_save_payment_method
+    @options[:pm_type] = 'gb_visa_mo_card'
+    response = @gateway.purchase(@amount, @credit_card, @options.merge(save_payment_method: true))
+    assert_success response
+    assert_equal 'SUCCESS', response.message
```
jcreiff, line 152: Is there any additional assertion we can add to validate that Rapyd is receiving the field? Not sure if or how their response differs when the field is present or absent in the request, but if there's something we could check related to that, it would make the test a little stronger

jcreiff (approved):
Just one minor suggestion about the remote test but not necessarily a blocker, LGTM!

test/remote/gateways/remote_rapyd_test.rb
rachelkirk: Good idea, will look at their docs and our responses.

PR #5194, opened 2024-07-26
Braintree: Support override_application_id

aenand:
COMP-266

ActiveMerchant currently uses Class.application_id or @options[:channel] to let merchants pass platform BN codes to gateways to signify that a transaction may be going to one merchant's accounts but it originates from a platform or aggregator.

This poses a problem in that the Class value is assigned at compile time and the @options[:channel] refers to a static value used when initializing the class.

This commit adds support for `override_application_id` which is a way for aggregators to pass in a new BN code at transaction time alongside transaction parameters.

Test Summary
Remote:
123 tests, 661 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Cannot add a remote test because there is no generic `channel` value

### DIFF
--- a/test/unit/gateways/braintree_blue_test.rb
+++ b/test/unit/gateways/braintree_blue_test.rb
@@ -955,8 +955,6 @@ class BraintreeBlueTest < Test::Unit::TestCase
     assert_equal gateway.send(:create_transaction_parameters, 100, credit_card('41111111111111111111'), {})[:channel], 'overidden-channel'
 
     assert_equal gateway.send(:create_transaction_parameters, 100, credit_card('41111111111111111111'), { override_application_id: 'override-application-id' })[:channel], 'override-application-id'
-  ensure
-    ActiveMerchant::Billing::BraintreeBlueGateway.application_id = nil
   end
 
   def test_successful_purchase_with_descriptor
--- a/test/unit/gateways/braintree_blue_test.rb
+++ b/test/unit/gateways/braintree_blue_test.rb
@@ -952,9 +952,11 @@ class BraintreeBlueTest < Test::Unit::TestCase
 
   def test_override_application_id_is_sent_to_channel
     gateway = BraintreeBlueGateway.new(merchant_id: 'test', public_key: 'test', private_key: 'test', channel: 'overidden-channel')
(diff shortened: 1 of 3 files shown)


test/unit/gateways/braintree_blue_test.rb
```diff
@@ -921,7 +921,7 @@ def test_that_setting_a_wiredump_device_on_the_gateway_sets_the_braintree_logger
     end
   end
 
-  def test_solution_id_is_added_to_create_transaction_parameters
+  def test_channel_is_added_to_create_transaction_parameters
```
travismorrison, line 924: how come this is renamed?

test/unit/gateways/braintree_blue_test.rb
```diff
@@ -932,6 +932,13 @@ def test_solution_id_is_added_to_create_transaction_parameters
     ActiveMerchant::Billing::BraintreeBlueGateway.application_id = nil
   end
 
+  def test_override_application_id_is_sent_to_channel
+    gateway = BraintreeBlueGateway.new(merchant_id: 'test', public_key: 'test', private_key: 'test', channel: 'overidden-channel')
+    assert_equal gateway.send(:create_transaction_parameters, 100, credit_card('41111111111111111111'), {})[:channel], 'overidden-channel'
+
+    assert_equal gateway.send(:create_transaction_parameters, 100, credit_card('41111111111111111111'), { override_application_id: 'override-application-id' })[:channel], 'override-application-id'
```
travismorrison, line 939: Non blocking suggestion: sweeten this test a little bit. The addition of the options hash seems to push it into overstuffed-one-liner territory.

I'm talking as simple as
```
    gateway_response = gateway.send(:create_transaction_parameters, 100, credit_card('41111111111111111111'), { override_application_id: 'override-application-id' })
    assert_equal gateway_response[:channel], 'override-application-id'
```

... and then change up the first one to mirror that pattern. The first one, on its own, I wouldn't quite say seems overstuffed, but no point in spelling them different.

travismorrison (approved):
Non-blocking style suggestions, otherwise looks good

test/unit/gateways/braintree_blue_test.rb
aenand: `solution_id` was the old name for the field at BT. It's now called `channel` so for future people reading tests this will make more sense

PR #5201, opened 2024-08-01
Iveri: Add AuthReversal for Authorizations

almalee24:
If the transaction to be voided is an Authorization then use AuthReversal instead of Void.

Remote
22 tests, 56 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 95.4545% passed

Unit
16 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


bradbroge (approved):
👍🏻

PR #5205, opened 2024-08-06
Decidir: Pass CVV for NT

almalee24:
This changes will allow CVV, card number, exp month and exp year.

Remote
27 tests, 97 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


aenand:
Decidir wants the Network Token data and the raw pan? What's the point of sending a NT then

almalee24:
> Decidir wants the Network Token data and the raw pan? What's the point of sending a NT then

It's not Decidir but a Visa mandate that we need to send CVV for NT transactions. This is meant to just send CVV but I was think we might as well just send everything.

bradbroge (approved):
LGTM

PR #5206, opened 2024-08-07
Repair Mecury and TransFirst

aenand:
A previous commit broke these gateways after updating REXML. This commit updates the two gateways so that they can process the XML like it used to.

Test Summary
Remote:
No credentials available for Mercury
Could not reach the test server for TransFirst


JesseBowling (approved):
Thanks for fixing these!

PR #5207, opened 2024-08-08
Repair to mercury

aenand:
Remote
20 tests, 66 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95% passed

### DIFF
--- a/test/remote/gateways/remote_mercury_test.rb
+++ b/test/remote/gateways/remote_mercury_test.rb
@@ -8,11 +8,12 @@ class RemoteMercuryTest < Test::Unit::TestCase
     @gateway = MercuryGateway.new(fixtures(:mercury))
 
     @amount = 100
+    @decline_amount = 257
 
-    @credit_card = credit_card('4003000123456781', brand: 'visa', month: '12', year: '18')
+    @credit_card = credit_card('4895281000000006', brand: 'visa', month: '12', year: Time.now.year)
 
-    @track_1_data = '%B4003000123456781^LONGSEN/L. ^18121200000000000000**123******?*'
-    @track_2_data = ';5413330089010608=2512101097750213?'
+    @track_1_data = "%B#{@credit_card.number}^LONGSEN/L. ^18121200000000000000**111******?*"
+    @track_2_data = ";#{@credit_card.number}=18121200000000000000?"
 
     @options = {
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_mercury_test.rb
```diff
@@ -9,10 +9,10 @@ def setup
 
     @amount = 100
 
-    @credit_card = credit_card('4003000123456781', brand: 'visa', month: '12', year: '18')
+    @credit_card = credit_card('4895281000000006', brand: 'visa', month: '12', year: '25')
```
bradbroge, line 12: can we set the year to `Time.now.year` or something further into the future?

test/remote/gateways/remote_mercury_test.rb
```diff
@@ -46,13 +46,13 @@ def test_successful_authorize_and_capture
   end
 
   def test_failed_authorize
-    response = @gateway.authorize(1100, @credit_card, @options)
+    response = @gateway.authorize(257, @credit_card, @options)
```
bradbroge, line 49: is this a `magic number` or can it be `@amount`?  If it is a magic number, can we set it as a variable instead of using the magic number here?

bradbroge (commented):
Looks great, a couple of minor comments in the test file.

test/remote/gateways/remote_mercury_test.rb
almalee24: Let me test that because the year 25 is in their testing guide

test/remote/gateways/remote_mercury_test.rb
almalee24: It's magic number it triggers the Decline message but I'll create a variable for it

test/unit/gateways/mercury_test.rb
```diff
@@ -126,41 +126,47 @@ def test_transcript_scrubbing
...
     RESPONSE
   end
 
+  # def successful_purchase_response
```
almalee24, line 133: @aenand should I remove this?

aenand: Yeah I think so. I replaced it with a more accurate successful response

almalee24 (approved):

PR #5208, opened 2024-08-09
Adyen: Fix billing address empty string error

yunnydang:
This change allows us to return "NA" instead of nil for street address and house or number address if address_override is set to true.

Local:
5985 tests, 80165 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
121 tests, 635 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
144 tests, 465 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.6667% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -510,9 +510,12 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_billing_address(post, options, address)
+        address[:address1] = 'NA' if address[:address1].blank?
+        address[:address2] = 'NA' if address[:address2].blank?
+
         post[:billingAddress] = {}
-        post[:billingAddress][:street] = options[:address_override] == true ? address[:address2] : address[:address1] || 'NA'
-        post[:billingAddress][:houseNumberOrName] = options[:address_override] == true ? address[:address1] : address[:address2] || 'NA'
+        post[:billingAddress][:street] = options[:address_override] == true ? address[:address2] : address[:address1]
+        post[:billingAddress][:houseNumberOrName] = options[:address_override] == true ? address[:address1] : address[:address2]
         post[:billingAddress][:postalCode] = address[:zip] if address[:zip]
         post[:billingAddress][:city] = address[:city] || 'NA'
         post[:billingAddress][:stateOrProvince] = get_state(address)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -39,6 +39,7 @@
 * Adyen: Add header fields to response body [yunnydang] #5184
 * Stripe and Stripe PI: Add header fields to response body [yunnydang] #5185
 * Stripe and Stripe PI: Add metadata and order_id for refund and void [yunnydang] #5204
+* Adyen: Fix billing address empty string error [yunnydang] #5208
 
 == Version 1.136.0 (June 3, 2024)
 * Shift4V2: Add new gateway based on SecurionPay adapter [heavyblade] #4860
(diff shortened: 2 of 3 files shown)


jcreiff (changes requested):
From the way I read the request for this change, it's not a `nil` address field that is the issue, it's an empty string `''`. If you add `address2: ""` to your remote test, you'll see the error thrown.

The `.nil?` check won't insert the `'NA'` in that circumstance because the empty string is `truthy`. Also, there's already an OR statement tacked onto the ends of the lines where `:street` and `houseNumberOrName` are applied to `post`, to try and fallback to `NA`

I think we'll need to use `.blank?` instead. This could also potentially be extracted into a method similar to `get_state` or `get_country`, since those fields also have expected default values for Adyen.

yunnydang:
> From the way I read the request for this change, it's not a `nil` address field that is the issue, it's an empty string `''`. If you add `address2: ""` to your remote test, you'll see the error thrown.
> 
> The `.nil?` check won't insert the `'NA'` in that circumstance because the empty string is `truthy`. Also, there's already an OR statement tacked onto the ends of the lines where `:street` and `houseNumberOrName` are applied to `post`, to try and fallback to `NA`
> 
> I think we'll need to use `.blank?` instead. This could also potentially be extracted into a method similar to `get_state` or `get_country`, since those fields also have expected default values for Adyen.

ahhh interesting! Idk why I was fixated on the nil aspect of it and didnt include the empty string case. I think i went down the nil rabbit hole in this method but when we send options[:address_override] == true, this puts us through that ternary block. You mentioned that the OR statement would guard it but from my testing and understanding, when not sending the address1 and address2, it will set the post[:billingAddress][:street] and post[:billingAddress][:houseNumberOrName] to nil. Thus causing us a failing response from the gateway. This is why i put in the nil check in case we get a request that doesnt have both of the address1 and address2. 

You can double check the statement above, but i dont think we will gracefully fallback to the OR "NA" statement if we send the options[:address_override] but dont have the address1 or address2.

The weird thing with the empty string  is that when i pass them in as a single quote '' with no spaces, the tests still pass. However! When I pass in a double quote with no space "", it doesnt pass! so im not sure if this is a sandbox issue or what but I'll add the guard for empty string as well! I still think my nil guard is necessary though as mentioned above

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -510,6 +510,9 @@ def add_address(post, options)
       end
 
       def add_billing_address(post, options, address)
+        address[:address1] = 'NA' if address[:address1].nil? || address[:address1].blank?
+        address[:address2] = 'NA' if address[:address2].nil? || address[:address2].blank?
```
jcreiff, line 514: Based on [this description](https://stackoverflow.com/a/888877) of the behavior of `.blank?`, I think the `.nil?` is extraneous here - `.blank?` should return true for both `nil` and empty strings.

I think you could simply swap the `nil?` for `blank?` and then also remove the `|| 'NA'` fallback at the end of the `address_override` ternary statements, because it won't be needed anymore (and removing it will make the intention of that logic a little easier to parse)

jcreiff (changes requested):

lib/active_merchant/billing/gateways/adyen.rb
yunnydang: sounds good!

yunnydang: good to know .blank? also accounts for nil values!

jcreiff (approved):

PR #5212, opened 2024-08-13
StripePI: Skip add_network_token_cryptogram_and_eci method to accept ApplePay recurring payments

sinourain:
Summary:
StripePI: Skip add_network_token_cryptogram_and_eci method to accept ApplePay recurring payments and add unit/remote tests

[SER-3618](https://spreedly.atlassian.net/browse/ECS-3618)

Remote Test:
Loaded suite test/remote/gateways/remote_stripe_payment_intents_test
Started
Finished in 255.69074 seconds.
97 tests, 460 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
0.38 tests/s, 1.80 assertions/s

Unit Tests:
Finished in 29.539052 seconds.
5993 tests, 80199 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
202.88 tests/s, 2715.02 assertions/s

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -38,7 +38,7 @@ module ActiveMerchant #:nodoc:
               return result if result.is_a?(ActiveMerchant::Billing::Response)
             end
 
-            add_network_token_cryptogram_and_eci(post, payment_method)
+            add_network_token_cryptogram_and_eci(post, payment_method, options)
             add_external_three_d_secure_auth_data(post, options)
             add_metadata(post, options)
             add_return_url(post, options)
@@ -423,17 +423,22 @@ module ActiveMerchant #:nodoc:
         post_data
       end
 
-      def add_network_token_cryptogram_and_eci(post, payment_method)
-        return unless adding_network_token_card_data?(payment_method)
+      def add_network_token_cryptogram_and_eci(post, payment_method, options)
+        return unless payment_method.is_a?(NetworkTokenizationCreditCard) && options.dig(:stored_credential, :initiator) != 'merchant'
+        return if digital_wallet_payment_method?(payment_method) && options[:new_ap_gp_route] != true
 
         post[:payment_method_options] ||= {}
         post[:payment_method_options][:card] ||= {}
         post[:payment_method_options][:card][:network_token] ||= {}
-        post[:payment_method_options][:card][:network_token][:cryptogram] = payment_method.payment_cryptogram if payment_method.payment_cryptogram
-        post[:payment_method_options][:card][:network_token][:electronic_commerce_indicator] = payment_method.eci if payment_method.eci
+        post[:payment_method_options][:card][:network_token].merge!({
+          cryptogram: payment_method.respond_to?(:payment_cryptogram) ? payment_method.payment_cryptogram : options[:cryptogram],
+          electronic_commerce_indicator: format_eci(payment_method, options)
+        }.compact)
       end
 
       def add_digital_wallet(post, payment_method, options)
+        source = payment_method.respond_to?(:source) ? payment_method.source : options[:wallet_type]
+
         post[:payment_method_data] = {
           type: 'card',
           card: {
@@ -443,24 +448,11 @@ module ActiveMerchant #:nodoc:
             network_token: {
               number: payment_method.number,
               exp_month: payment_method.month,
-              exp_year: payment_method.year
+              exp_year: payment_method.year,
+              tokenization_method: DIGITAL_WALLETS[source]
             }
           }
         }
-
-        add_cryptogram_and_eci(post, payment_method, options) unless options[:wallet_type]
-        source = payment_method.respond_to?(:source) ? payment_method.source : options[:wallet_type]
-        post[:payment_method_data][:card][:network_token][:tokenization_method] = DIGITAL_WALLETS[source]
-      end
-
-      def add_cryptogram_and_eci(post, payment_method, options)
-        post[:payment_method_options] ||= {}
-        post[:payment_method_options][:card] ||= {}
-        post[:payment_method_options][:card][:network_token] ||= {}
-        post[:payment_method_options][:card][:network_token] = {
-          cryptogram: payment_method.respond_to?(:payment_cryptogram) ? payment_method.payment_cryptogram : options[:cryptogram],
-          electronic_commerce_indicator: format_eci(payment_method, options)
-        }.compact
       end
 
       def format_eci(payment_method, options)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,6 +26,7 @@
 * PayTrace: Update MultiResponse for Capture [almalee24] #5203
 * Adyen: Add support for Pan Only GooglePay [almalee24] #5221
 * Decidir: Send extra fields for tokenized NT transactions [sinourain] #5224
+* StripePI: Skip add_network_token_cryptogram_and_eci method to accept ApplePay recurring payments [sinourain] #5212
 
 == Version 1.137.0 (August 2, 2024)
 * Unlock dependency on `rexml` to allow fixing a CVE (#5181).
--- a/test/remote/gateways/remote_stripe_payment_intents_test.rb
+++ b/test/remote/gateways/remote_stripe_payment_intents_test.rb
@@ -305,6 +305,57 @@ class RemoteStripeIntentsTest < Test::Unit::TestCase
     assert_match('apple_pay', purchase.params.dig('charges', 'data')[0]['payment_method_details']['card']['wallet']['type'])
   end
 
+  def test_successful_purchase_with_apple_pay_and_cit
+    options = {
+      currency: 'GBP',
+      new_ap_gp_route: true,
+      stored_credential_transaction_type: true,
+      stored_credential: {
+        initiator: 'cardholder',
+        reason_type: 'unscheduled',
+        initial_transaction: true
+      }
+    }
+
+    purchase = @gateway.purchase(@amount, @apple_pay, options)
+    assert purchase.success?
+    assert_match('apple_pay', purchase.params.dig('charges', 'data')[0]['payment_method_details']['card']['wallet']['type'])
+  end
+
+  def test_succeeds_apple_pay_ntid_and_passes_it_to_mit
+    options = {
+      currency: 'GBP',
+      new_ap_gp_route: true,
+      stored_credential_transaction_type: true,
+      stored_credential: {
+        initiator: 'cardholder',
+        reason_type: 'unscheduled',
+        initial_transaction: true
+      }
+    }
+
+    cit_purchase = @gateway.purchase(@amount, @apple_pay, options)
+    assert cit_purchase.success?
+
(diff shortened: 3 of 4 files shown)


aenand (changes requested):
We need to prevent sending cryptogram and eci if the `stored_credential[:initiator] == 'merchant'`

sinourain:
> We need to prevent sending cryptogram and eci if the `stored_credential[:initiator] == 'merchant'`

Good catch, done!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -38,7 +38,7 @@ def create_intent(money, payment_method, options = {})
               return result if result.is_a?(ActiveMerchant::Billing::Response)
             end
 
-            add_network_token_cryptogram_and_eci(post, payment_method)
+            add_network_token_cryptogram_and_eci(post, payment_method, options)
```
aenand, line 41: Does Apple Pay and Google Pay also add these values in `add_digital_wallet`? It seems a bit confusing if the new AP route hits this method twice.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -316,8 +316,11 @@ def digital_wallet_payment_method?(payment_method)
         payment_method.source == :google_pay || payment_method.source == :apple_pay
       end
 
-      def adding_network_token_card_data?(payment_method)
-        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) && payment_method.source == :network_token
+      def adding_network_token_card_data?(payment_method, options = {})
+        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) &&
+                       payment_method.source == :network_token &&
+                       payment_method.source != :apple_pay &&
+                       options.dig(:stored_credential, :initiator) != 'merchant'
```
aenand, line 323:
```suggestion
      def add_network_token_card_data?(payment_method, options = {})
        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) &&
                       payment_method.source == :network_token &&
                       options.dig(:stored_credential, :initiator) != 'merchant'
```

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: Good catch! Fixed!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -316,8 +316,10 @@ def digital_wallet_payment_method?(payment_method)
         payment_method.source == :google_pay || payment_method.source == :apple_pay
       end
 
-      def adding_network_token_card_data?(payment_method)
-        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) && payment_method.source == :network_token
+      def adding_network_token_card_data?(payment_method, options = {})
+        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) &&
```
almalee24, line 320: Can we just leave this method as it was before? I rather remove it from `add_network_token_cryptogram_and_eci`. `adding_network_token_card_data?` is used in other methods and I wouldn't want us to break anything else.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -423,17 +425,21 @@ def add_network_token_data(post_data, payment_method, options)
         post_data
       end
 
-      def add_network_token_cryptogram_and_eci(post, payment_method)
-        return unless adding_network_token_card_data?(payment_method)
+      def add_network_token_cryptogram_and_eci(post, payment_method, options)
+        return unless !options[:wallet_type] || adding_network_token_card_data?(payment_method, options)
```
almalee24, line 429: I think it would be better to update this to be something like 
`return unless payment_method.is_a?(NetworkTokenizationCreditCard) && options.dig(:stored_credential, :initiator) != 'merchant'`

almalee24: We only pass options[:wallet_type] when the payment method is a non tokenized GooglePay and in AM the payment method model for this would be CreditCard

almalee24: > I think it would be better to update this to be something like 
```
return unless payment_method.is_a(NetworkTokenizationCreditCard)
return if options[:stored_credential] && options.dig(:stored_credential, :initiator) == 'merchant'
```

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -423,17 +425,21 @@ def add_network_token_data(post_data, payment_method, options)
...
 
         post[:payment_method_options] ||= {}
         post[:payment_method_options][:card] ||= {}
         post[:payment_method_options][:card][:network_token] ||= {}
-        post[:payment_method_options][:card][:network_token][:cryptogram] = payment_method.payment_cryptogram if payment_method.payment_cryptogram
-        post[:payment_method_options][:card][:network_token][:electronic_commerce_indicator] = payment_method.eci if payment_method.eci
+        post[:payment_method_options][:card][:network_token] = {
```
almalee24, line 434: Also if `post[:payment_method_options][:card][:network_token]` already exist we would have to use `merge!` because here we are resetting the object.

almalee24: So you could do something like

```
post[:payment_method_options][:card][:network_token].merge!({
  cryptogram: payment_method.respond_to?(:payment_cryptogram) ? payment_method.payment_cryptogram : options[:cryptogram],
          electronic_commerce_indicator: format_eci(payment_method, options)
 }.compact)
```

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: I agree, done!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: Yes, thanks!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: Done!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -317,7 +317,8 @@ def digital_wallet_payment_method?(payment_method)
       end
 
       def adding_network_token_card_data?(payment_method)
-        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) && payment_method.source == :network_token
+        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) &&
+                       payment_method.source == :network_token
```
aenand, line 321: Do we need this check for source?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -317,7 +317,7 @@ def digital_wallet_payment_method?(payment_method)
       end
 
       def adding_network_token_card_data?(payment_method)
-        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) && payment_method.source == :network_token
+        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard)
 
         false
       end
```
aenand, line 323: What do you think about updating this method to have the check for `options.dig(:stored_credential, :initiator) != 'merchant'` as well so that the `add_network_token_cryptogram_and_eci` method can just rely on this one?

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: I checked and it is not necessary

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: Yes, done!

aenand (approved):
LGTM :shipit: Please make sure all of Alma's feedback is addressed!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -316,8 +316,8 @@ def digital_wallet_payment_method?(payment_method)
         payment_method.source == :google_pay || payment_method.source == :apple_pay
       end
 
-      def adding_network_token_card_data?(payment_method)
-        return true if payment_method.is_a?(ActiveMerchant::Billing::NetworkTokenizationCreditCard) && payment_method.source == :network_token
+      def adding_network_token_card_data?(payment_method, options = {})
```
almalee24, line 319: I would prefer this stay as it was originally was. `adding_network_token_card_data?` is used in 4 other places and I do not want to alter how that is working at the moment.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -423,17 +423,21 @@ def add_network_token_data(post_data, payment_method, options)
         post_data
       end
 
-      def add_network_token_cryptogram_and_eci(post, payment_method)
-        return unless adding_network_token_card_data?(payment_method)
+      def add_network_token_cryptogram_and_eci(post, payment_method, options)
+        return unless adding_network_token_card_data?(payment_method, options)
```
almalee24, line 427: This should be updated to `return unless payment_method.is_a?(NetworkTokenizationCreditCard) && options.dig(:stored_credential, :initiator) != 'merchant'`

almalee24: Only AP/GP transactions on the new flow would reach here since if it was on the old flow the [payment_method would be a token](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe_payment_intents.rb#L24)

almalee24: If we want to just be safe that it's only AP/GP new flow on this we could do something like

```
return unless payment_method.is_a?(NetworkTokenizationCreditCard) && options.dig(:stored_credential, :initiator) != 'merchant'
return if digital_wallet_payment_method?(payment_method) && options[:new_ap_gp_route] != true
```

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: Sure, no problem :)

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
sinourain: Great, done!

almalee24 (approved):

PR #5213, opened 2024-08-13
Cybersource: Add apple_pay params for discover if flag passed

DustinHaefele:


almalee24 (approved):

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -931,6 +933,13 @@ def add_auth_wallet(xml, payment_method, options)
             xml.tag!('xid', Base64.encode64(cryptogram[20...40])) if cryptogram.bytes.count > 20
             xml.tag!('reconciliationID', options[:reconciliation_id]) if options[:reconciliation_id]
           end
+        when :discover
+          return unless options[:enable_cybs_discover_apple_pay]
+
+          xml.tag! 'ccAuthService', { 'run' => 'true' } do
+            xml.tag!('cavv', payment_method.payment_cryptogram) unless commerce_indicator
+            xml.commerceIndicator commerce_indicator.nil? ? ECI_BRAND_MAPPING[brand] : commerce_indicator
```
almalee24, line 941: Do we also want to pass `options[:reconciliation_id]`?

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -931,6 +933,13 @@ def add_auth_wallet(xml, payment_method, options)
             xml.tag!('xid', Base64.encode64(cryptogram[20...40])) if cryptogram.bytes.count > 20
             xml.tag!('reconciliationID', options[:reconciliation_id]) if options[:reconciliation_id]
           end
+        when :discover
+          return unless options[:enable_cybs_discover_apple_pay]
+
+          xml.tag! 'ccAuthService', { 'run' => 'true' } do
+            xml.tag!('cavv', payment_method.payment_cryptogram) unless commerce_indicator
```
klumsy, line 940: with the underlying data, what will the commerce indicator actually be.. the non Network tokenization PB or the DIPB ?

DustinHaefele: Based on line 911 the commerce_indicator will be `internet` in the case of network tokenization and it will be `pb` for non-network tokenization

lib/active_merchant/billing/gateways/cyber_source.rb
DustinHaefele: No I don't think so. That is noted in the visa docs, but not the discover docs.

klumsy: in the cybersource docs, it seems like in this context DIPB is what is required. https://developer.cybersource.com/content/dam/docs/cybs/en-us/apple-pay/developer/smartpay/so/applepay.pdf
image

lib/active_merchant/billing/gateways/cyber_source.rb
klumsy: in the cybersource docs, its seems DIPB is what is required in this context
https://developer.cybersource.com/content/dam/docs/cybs/en-us/apple-pay/developer/smartpay/so/applepay.pdf
image

klumsy: >  Based on line 911 the commerce_indicator will be internet in the case of network tokenization and it will be pb for non-network tokenization

i think that code is only for SUBSEQUENT, when its merchant initiated follow up. as a way to do recurring.. i think the other code earlier has the logic the other way 'internet' for regular, and 'pb' for network tokenization

DustinHaefele: @klumsy 

Yes the 902 code is for pure network tokens, and not related to apple_pay, that is differentiated in `add_auth_service` around line 795. 

This is also a little confusing because in the example, as you posted, it shows `dipb`. But in the actual definition of required elements it says it should be `internet`.  I think I am going to go with `internet` in this case since it is written more explicitly in the docs.

image

PR #5216, opened 2024-08-14
Updated the Adyen NT and stored credential flow.

DustinHaefele:
unit:
124 tests, 655 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

remote:
143 tests, 462 assertions, 13 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 90.9091% passed (Same store and unstore failures as master branch)

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -464,9 +464,7 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_shopper_interaction(post, payment, options = {})
-        if  (options.dig(:stored_credential, :initial_transaction) && options.dig(:stored_credential, :initiator) == 'cardholder') ||
-            (payment.respond_to?(:verification_value) && payment.verification_value && options.dig(:stored_credential, :initial_transaction).nil?) ||
-            payment.is_a?(NetworkTokenizationCreditCard)
+        if ecommerce_shopper_interaction?(payment, options)
           shopper_interaction = 'Ecommerce'
         else
           shopper_interaction = 'ContAuth'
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -982,6 +981,17 @@ def unsupported_failure_response(initial_response)
...
+
+      def skip_mpi_data?(options = {})
+        # Skips adding the NT mpi data if it is explicitly skipped in options, or if it is MIT and not the initial transaction.
+        options[:skip_mpi_data] == 'Y' || (!options.dig(:stored_credential, :initial_transaction) && options.dig(:stored_credential, :initiator) == 'merchant' && options[:switch_cryptogram_mapping_nt])
+      end
+
+      def ecommerce_shopper_interaction?(payment, options)
+        (options.dig(:stored_credential, :initial_transaction) && options.dig(:stored_credential, :initiator) == 'cardholder') ||
+          (payment.respond_to?(:verification_value) && payment.verification_value && options.dig(:stored_credential, :initial_transaction)) ||
+          (payment.is_a?(NetworkTokenizationCreditCard) && !options[:switch_cryptogram_mapping_nt])
```
almalee24, line 993: Does this mean you want to pass ecommerce if this a NetworkTokenizationCreditCard and switch_cryptogram_mapping_nt is nil?

DustinHaefele: Yeah if it is nil or false. 

I did this because when the flag is on I want to remove the payment.is_a?(NetworkTokenizationCreditCard) check that is there currently, and this allows the current functionality to stay the same while the flag is off.

almalee24 (approved):

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -982,6 +981,17 @@ def unsupported_failure_response(initial_response)
       def card_not_stored?(response)
         response.authorization ? response.authorization.split('#')[2].nil? : true
       end
+
+      def skip_mpi_data?(options = {})
+        # Skips adding the NT mpi data if it is explicitly skipped in options, or if it is MIT and not the initial transaction.
+        options[:skip_mpi_data] == 'Y' || (!options.dig(:stored_credential, :initial_transaction) && options.dig(:stored_credential, :initiator) == 'merchant' && options[:switch_cryptogram_mapping_nt])
```
almalee24, line 987: Is this just because you want to test that it works all payment methods types? Because switch_cryptogram_mapping_nt is passed for all payment methods not just NT.

DustinHaefele: Good question, let me look at their docs and make sure I'm not missing something with other payment method types when it comes to mpi_data. I'll ensure that is in a good place before I move this closer to deployment.

DustinHaefele: So the reason I am using the flag here is because I don't want to update the fields we are sending until we activate the flag. `(!options.dig(:stored_credential, :initial_transaction) && options.dig(:stored_credential, :initiator) == 'merchant'` <== this portion is because Adyen's docs tell us to skip it when it is a follow-up merchant initiated transaction. I am adding the flag there because I don't want to start skipping the mpiData in new situations until we full activate this flow.

PR #5218, opened 2024-08-15
Adyen: Update split refund method

yunnydang:
This change updates the add_split method to be more flexible in when building and sending the split object to Adyen.

Local:
5988 tests, 80175 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
123 tests, 649 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
145 tests, 467 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.7241% passed


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -435,17 +435,17 @@ def add_splits(post, options)
...
-          }
-          split_hash['account'] = split['account'] if split['account']
+          if split['amount']
+            amount = {}
+            amount[:value] = split['amount']['value'] if split['amount']['value']
+            amount[:currency] = split['amount']['currency'] if split['amount']['currency']
+          end
+
+          split_hash = {}
+          split_hash[:amount] = amount unless split['type'] == 'PaymentFee' || split['type'] == 'AdyenFees' || amount.nil?
```
jcreiff, line 445: Instead of specifying the types to omit, could we just check for the presence of `amount` and make sure that it's not an empty hash?

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -435,17 +435,17 @@ def add_splits(post, options)
...
+          split_hash[:type] = split['type'] if split['type']
+          split_hash[:reference] = split['reference'] if split['reference']
+          split_hash[:account] = split['account'] if split['account']
```
jcreiff, line 448: I notice this is the one place where you've changed the syntax from string to symbol 🧐

test/remote/gateways/remote_adyen_test.rb
```diff
@@ -814,6 +814,39 @@ def test_failed_refund
...
+          'type' => 'PaymentFee',
+          'account' => '123345566'
+        }
+      ]
+    }
+
+    purchase = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success purchase
+    assert refund = @gateway.refund(@amount, purchase.authorization, options)
+    assert_failure refund
```
jcreiff, line 847: If we can't get a passing remote test, I'm wondering if a unit test makes more sense to verify that we're still passing all the appropriate `splits` fields, regardless of the `amount` being present or not?

lib/active_merchant/billing/gateways/adyen.rb
yunnydang: hmm now that you mentioned it, i wonder why the split_hash['account'] wasnt a symbol to begin with but the other split_hash fields are?

lib/active_merchant/billing/gateways/adyen.rb
yunnydang: This is a simpler way to omit the amount object, I think I originally wrote it to omit just in case they send the amount alng with the type being 'PaymentFee' or 'AdyenFees'. Almost like an extra guard but i guess they can refer to the Adyens documentation

lib/active_merchant/billing/gateways/adyen.rb
yunnydang: Should I leave it as it is now (it'd be kinda weird) or just change it to symbol to keep these fields uniformed?

jcreiff: For posterity - we chatted about this and decided to leave as is since it's currently working

jcreiff (approved):
Nice work!

PR #5221, opened 2024-08-16
Adyen: Add support for Pan Only GooglePay

almalee24:
Remote
146 tests, 469 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 91.7808% passed

If it is a Pan Only GooglePay then pass 'paymentdatasource.type and selectedBrand as googlepay and paymentdatasource.tokenized as false

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -247,8 +247,7 @@ module ActiveMerchant #:nodoc:
 
       def add_extra_data(post, payment, options)
         post[:telephoneNumber] = (options[:billing_address][:phone_number] if options.dig(:billing_address, :phone_number)) || (options[:billing_address][:phone] if options.dig(:billing_address, :phone)) || ''
-        post[:selectedBrand] = options[:selected_brand] if options[:selected_brand]
-        post[:selectedBrand] ||= NETWORK_TOKENIZATION_CARD_SOURCE[payment.source.to_s] if payment.is_a?(NetworkTokenizationCreditCard)
+        post[:selectedBrand] = options[:selected_brand] if options[:selected_brand] && !post[:selectedBrand]
         post[:deliveryDate] = options[:delivery_date] if options[:delivery_date]
         post[:merchantOrderReference] = options[:merchant_order_reference] if options[:merchant_order_reference]
         post[:captureDelayHours] = options[:capture_delay_hours] if options[:capture_delay_hours]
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -1,3 +1,4 @@
+require 'pry'
```
bradbroge, line 1: this should be removed.

bradbroge (changes requested):
looks like there is a lingering `pry`, other than that it looks good.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -558,7 +557,7 @@ def add_payment(post, payment, options, action = nil)
         elsif payment.is_a?(Check)
           add_bank_account(post, payment, options, action)
         else
-          add_mpi_data_for_network_tokenization_card(post, payment, options) if payment.is_a?(NetworkTokenizationCreditCard)
+          add_network_tokenization_card(post, payment, options) if payment.is_a?(NetworkTokenizationCreditCard) || options[:wallet_type] == :google_pay
```
DustinHaefele, line 560: What is the situation where there will be an `options[:wallet_type]` but it won't be a `NetworkTokenizationCreditCard`?

DustinHaefele: Ah, nevermind, I'm following this now.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -620,18 +619,27 @@ def add_reference(post, authorization, options = {})
...
+      def add_network_tokenization_card(post, payment, options)
+        selected_brand = NETWORK_TOKENIZATION_CARD_SOURCE[options[:wallet_type]&.to_s || payment.source.to_s]
+        if selected_brand
+          post[:selectedBrand] = selected_brand
+          post[:additionalData] = {} unless post[:additionalData]
+          post[:additionalData]['paymentdatasource.type'] = selected_brand
+          post[:additionalData]['paymentdatasource.tokenized'] = options[:wallet_type] ? 'false' : 'true' if selected_brand == 'googlepay'
+        end
+
+        return if options[:skip_mpi_data] == 'Y' || options[:wallet_type]
```
DustinHaefele, line 631: Is wallet_type a gsf, or where is it coming from? If so is there a chance that a customer would send this with a `NetworkTokenizationCreditCard` and we would skip mpiData in a case where we might not want to?

almalee24: wallet_type get's passed by us for non-tokenized GooglePays which would be a CreditCard in AM. That's why we want to skip it because they would have cryptogram or eci.

bradbroge (approved):
👍🏻 Awesome work as always!

lib/active_merchant/billing/gateways/adyen.rb
DustinHaefele: Awesome! That is perfect, just making sure it couldn't be passed by mistake on regular CCs.

DustinHaefele (approved):

PR #5223, opened 2024-08-20
Ebanx: Add support for Stored Credentials

almalee24:


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -50,6 +50,7 @@ def purchase(money, payment, options = {})
         add_address(post, options)
         add_customer_responsible_person(post, payment, options)
         add_additional_data(post, options)
+        add_stored_credentials(post, options)
```
DustinHaefele, line 53: This is possibly just me still learning stored credentials, but should this be added to authorize or captures as well? Or does this only work on a full purchase? Particularly for initial transactions.

DustinHaefele (commented):
One question, but the code looks great, and the test coverage is excellent.

lib/active_merchant/billing/gateways/ebanx.rb
almalee24: Yeah probably not, I forgot to add it in authorize but let be double check if we need it for capture

almalee24: I've added it to Authorize since this is only for `ws/direct API` which is just Purchase & Authorize

DustinHaefele (approved):

PR #5225, opened 2024-08-21
Decidir: Update scrubbing cryptogram and token

almalee24:
Update scrubbing cryptogram in Prod since they add / in prod from cryptogram and we only want to scrub
the token in token_card_data


test/remote/gateways/remote_decidir_test.rb
```diff
@@ -360,14 +360,4 @@ def test_transcript_scrubbing
     assert_scrubbed(@credit_card.verification_value, transcript)
     assert_scrubbed(@gateway_for_purchase.options[:api_key], transcript)
   end
-
-  def test_transcript_scrubbing_network_token
```
almalee24, line 364: Removed this since cryptogram in Sandbox doesn't return the '/' in front of cryptogram value

bradbroge (approved):
Approved

PR #5226, opened 2024-08-21
Decidir: Remove pass_cvv_for_nt

almalee24:
This will allow all NT transactions to pass CVV if present.

Remote
27 tests, 97 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


almalee24:
The remote test that's failing is the Purchase that used NetworkToken. I've tried different NetworkToken info but it still fails. I have seen NT transactions passing in in Prod with security_code.

sinourain (approved):
LGTM! Nice job!

PR #5233, opened 2024-08-27
Iveri: Add AuthorisationReversal for Auth Void

almalee24:
If a Authorization transaction needs to be voided then AuthorisationReversal needs to be used instead of Void.

22 tests, 58 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


sinourain (approved):
LGTM! Goog job!

DustinHaefele (approved):

PR #5236, opened 2024-08-27
StripePI: Update Stored Credentials

almalee24:
Update Stored Credentials to only send NTID if
stored_credential_transaction_type is
setup_off_session_unscheduled or setup_off_session_recurring.

Remote:
96 tests, 457 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


sinourain (approved):
LGTM, I tried to move the code so I wouldn't have to delete the network_transaction_id option, but then I convinced myself that it's the best option we have, good job!

PR #5252, opened 2024-09-06
SafeCharge: add card holder verification fields

yunnydang:
This change adds the optional middle_name and card_holder_verification fields
Local:
6017 tests, 80326 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
31 tests, 176 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
35 tests, 93 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
94.2857% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -33,6 +33,7 @@
 * StripePI: Add metadata for GooglePay FPAN [almalee24] #5242
 * Paypal: Add inquire method [almalee24] #5231
 * Adyen: Enable multiple legs within airline data [jcreiff] #5249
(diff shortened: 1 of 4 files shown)


jcreiff (approved):
Looks good to me!

I was curious to see if sending the `card_holder_verification` field would change the response sent back from SafeCharge, and it does look like they add a `CardHolderNameVerification` object to the response. You could add an assertion to the remote test to show that it's present, but I will ✅ this either way

PR #5253, opened 2024-09-06
CyberSource Rest: Add the mcc field

yunnydang:
Adds the optional mcc field

Local:
6016 tests, 80320 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
40 tests, 217 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
54 tests, 177 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -1103,10 +1103,15 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_threeds_services(xml, options)
-        xml.tag! 'payerAuthEnrollService', { 'run' => 'true' } if options[:payer_auth_enroll_service]
+        if options[:payer_auth_enroll_service]
+          xml.tag! 'payerAuthEnrollService', { 'run' => 'true' } do
+            xml.tag! 'MCC', options[:mcc] if options[:mcc]
+          end
+        end
+
         if options[:payer_auth_validate_service]
           xml.tag! 'payerAuthValidateService', { 'run' => 'true' } do
-            xml.tag! 'signedPARes', options[:pares]
+            xml.tag! 'signedPARes', options[:pares] if options[:pares]
           end
         end
       end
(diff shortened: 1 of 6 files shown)


lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -138,6 +138,7 @@ def add_three_ds(post, payment_method, options)
         post[:consumerAuthenticationInformation][:paSpecificationVersion] = three_d_secure[:version] if three_d_secure[:version]
         post[:consumerAuthenticationInformation][:directoryServerTransactionID] = three_d_secure[:ds_transaction_id] if three_d_secure[:ds_transaction_id]
         post[:consumerAuthenticationInformation][:eciRaw] = three_d_secure[:eci] if three_d_secure[:eci]
+        post[:consumerAuthenticationInformation][:mcc] = options[:mcc] if options[:mcc]
```
yunnydang, line 141: hey @jcreiff after our discussion I changed it to be be pulling the mcc from the options object however would it pose any issue or concern if line 128 is still there? im thinking theres no point to send the mcc by itself if none of the rest of the threeds fields are there anyways but wanted to run it by you, thanks!

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -1103,10 +1103,15 @@ def add_installments(xml, options)
...
+        if options[:payer_auth_enroll_service]
+          xml.tag! 'payerAuthEnrollService', { 'run' => 'true' } do
+            xml.tag! 'MCC', options[:mcc] if options[:mcc]
+          end
+        end
+
         if options[:payer_auth_validate_service]
           xml.tag! 'payerAuthValidateService', { 'run' => 'true' } do
-            xml.tag! 'signedPARes', options[:pares]
+            xml.tag! 'signedPARes', options[:pares] if options[:pares]
```
jcreiff, line 1114: Any particular reason to add this? I think it should be reliably included in `options` based on the way we communicate these details from the Spreedly side

lib/active_merchant/billing/gateways/cyber_source_rest.rb
jcreiff: I think that guard clause should stay as is - the 3DS info is the main reason to enter this block of logic, and it doesn't seem like we want to entertain the option of only entering this method to add MCC.

However, this brings up a question about the requested use case that initiated this work, which I will follow up on in a separate thread

PR #5255, opened 2024-09-10
Rapdy: Add idempotency for all endpoints

almalee24:


lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -74,10 +77,12 @@ def store(payment, options = {})
         add_payment_fields(post, options)
         add_payment_urls(post, options, 'store')
         add_address(post, payment, options)
+        add_idempotency(options)
         commit(:post, 'customers', post)
       end
 
       def unstore(customer)
+        add_idempotency(options)
```
bradbroge, line 85: I don't think `options` exists in this context.

bradbroge (commented):
Question around the `unstore` action

lib/active_merchant/billing/gateways/rapyd.rb
almalee24: Right, why didn't the unit tests catch that. Let me look at that.

aenand:
Why would someone want idempotency for all endpoints?

lib/active_merchant/billing/gateways/rapyd.rb
almalee24: Fix it

almalee24:
> Why would someone want idempotency for all endpoints?

I'm not sure why they would want to but if they did pass it all these endpoints support it.

bradbroge (approved):
👍🏻

PR #5261, opened 2024-09-12
Shift4: Update response parsing to account for hostresponse

jcreiff:
The response object returned from the Shift4 gateway seems to have changed `hostResponse` to `hostresponse` around August 26th, which causes all failed auth/purchase transactions to default to the fallback message of "Transaction declined"

This commit corrects the parsing to account for this change but also leaves `hostResponse` as an option, in case the change is reverted by Shift4 in the future.

Assertions were added to show that the refactoring still provides correct values for both `message` and `error_code` for the various response structures we receive. Shift4 does not return a `host(r/R)esponse` object in sandbox testing, so we have to rely on unit tests here.

CER-1740

LOCAL
6018 tests, 80340 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

UNIT
29 tests, 190 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

REMOTE
29 tests, 67 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -38,6 +38,7 @@
 * Stripe PI: Update Stored Credentials [almalee24] #5236
 * Checkout V2: Update stored credential options function [jherreraa] #5239
 * Ebanx: Add support for Stored Credentials [almalee24] #5243
+* Shift4: Update response parsing to account for hostresponse [jcreiff] #5261
 
 == Version 1.137.0 (August 2, 2024)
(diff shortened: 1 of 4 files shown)


yunnydang (approved):
great work!

PR #5264, opened 2024-09-16
NMI: Add `industry_indicator` field

naashton:
Transactions on NMI could fail if this industry field is not provided. This field is not documented publicly on NMI.

Unit: 61 tests, 487 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 62 tests, 247 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
98.3871% passed


Heavyblade (approved):
Looks 👍 , 🚢  it

PR #5272, opened 2024-09-20
SumUp: Append partner_id to checkout_reference

naashton:
For merchants that need to pass an identifier that the source of the transaction originates from their platform, they can pass this value which will be prepended to the `order_id` and passed along as the `checkout_reference`

Unit: 10 tests, 32 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote: 11 tests, 20 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
54.5455% passed

Note: Some of the remote tests are failing, unrelated to changes in this PR. SumUP echo's back the `checkout_reference` so we can easily validate that the one we pass in the request by checking the response.

### DIFF
--- a/test/remote/gateways/remote_sum_up_test.rb
+++ b/test/remote/gateways/remote_sum_up_test.rb
@@ -51,6 +51,16 @@ class RemoteSumUpTest < Test::Unit::TestCase
     assert_equal 'PAID', response.message
   end
 
+  def test_successful_purchase_with_partner_id
+    options = {
+      partner_id: 'PartnerId',
+      order_id: SecureRandom.uuid
+    }
+
+    response = @gateway.purchase(@amount, @credit_card, options)
+    assert_equal "#{options[:partner_id]}-#{options[:order_id]}", response.params['checkout_reference']
+  end
+
   def test_failed_purchase
     response = @gateway.purchase(@amount, @declined_card, @options)
     assert_failure response
--- a/test/remote/gateways/remote_sum_up_test.rb
+++ b/test/remote/gateways/remote_sum_up_test.rb
@@ -58,6 +58,7 @@ class RemoteSumUpTest < Test::Unit::TestCase
     }
 
     response = @gateway.purchase(@amount, @credit_card, options)
+    assert_success response
     assert_equal "#{options[:partner_id]}-#{options[:order_id]}", response.params['checkout_reference']
   end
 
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_sum_up_test.rb
```diff
@@ -51,6 +51,16 @@ def test_successful_purchase_with_more_options
...
   end
 
+  def test_successful_purchase_with_partner_id
+    options = {
+      partner_id: 'PartnerId',
+      order_id: SecureRandom.uuid
+    }
+
+    response = @gateway.purchase(@amount, @credit_card, options)
+    assert_equal "#{options[:partner_id]}-#{options[:order_id]}", response.params['checkout_reference']
```
gasb150, line 61: Could you include the
 ```ruby 
assert_success response
```
 to ensure this will be a success transaction with this field?

File: test/remote/gateways/remote_sum_up_test.rb [hunk 2]
```diff
@@ -51,6 +51,16 @@ def test_successful_purchase_with_more_options
     assert_equal 'PAID', response.message
   end
 
+  def test_successful_purchase_with_partner_id
+    options = {
+      partner_id: 'PartnerId',
+      order_id: SecureRandom.uuid
+    }
+
+    response = @gateway.purchase(@amount, @credit_card, options)
```
javierpedrozaing, line 60: Since this is a remote test, could be good idea add a ` assert_success response` 💡

gasb150 (commented):
LGTM, just a small comment for you to consider.

javierpedrozaing (commented):
LGTM, just a small comment for your concideration.

test/remote/gateways/remote_sum_up_test.rb
naashton: I didn't because the gateway transactions are failing and it wasn't quite clear why. Since nothing has changed that would break the integration, I oped to ignore it for now.

test/remote/gateways/remote_sum_up_test.rb
naashton: Ah yes, Gustavo pointed this out as well. I opted to leave it out since the purchase transactions are failing.

gasb150 (approved):

javierpedrozaing (approved):
LGTM

PR #5275, opened 2024-09-23
Cybersource: Updated commerceIndicator on apple pay with discover

DustinHaefele:


bradbroge (approved):
It really feels like this change should reference the `ECI_BRAND_MAPPING` for discover, but I can see where that would cause other changes in the code >.>

So this feels like the least bad version of this change.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -938,7 +938,7 @@ def add_auth_wallet(xml, payment_method, options)
 
           xml.tag! 'ccAuthService', { 'run' => 'true' } do
             xml.tag!('cavv', payment_method.payment_cryptogram) unless commerce_indicator
-            xml.tag!('commerceIndicator', 'internet')
+            xml.tag!('commerceIndicator', 'dipb')
```
almalee24, line 941: Shouldn't the value change if they are using stored credentials?

DustinHaefele: In this case I think it is going to be handled more like amex than visa, where it is the same either way. It fails at the processor if it is internet which a customer found out while testing behind the flag.

PR #5276, opened 2024-09-23
CybersourceRest: Update message and error_code

almalee24:
Add dig to message_from and error_code_from to prevent NoMethod errors.

Remote
53 tests, 176 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source_rest.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source_rest.rb
@@ -415,21 +415,18 @@ module ActiveMerchant #:nodoc:
       def message_from(response)
         return response['status'] if success_from(response)
 
-        response['errorInformation']['message'] || response['message']
+        response.dig('errorInformation', 'message') || response['message']
       end
 
       def authorization_from(response)
         id = response['id']
-        has_amount = response['orderInformation'] && response['orderInformation']['amountDetails'] && response['orderInformation']['amountDetails']['authorizedAmount']
-        amount = response['orderInformation']['amountDetails']['authorizedAmount'].delete('.') if has_amount
+        amount = response.dig('orderInformation', 'amountDetails', 'authorizedAmount')&.delete('.')
 
-        return id if amount.blank?
-
-        [id, amount].join('|')
+        amount.present? ? [id, amount].join('|') : id
       end
 
       def error_code_from(response)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -398,21 +398,21 @@ def success_from(response)
       def message_from(response)
         return response['status'] if success_from(response)
 
-        response['errorInformation']['message'] || response['message']
+        response.dig('errorInformation', 'message') || response['message']
```
bradbroge, line 401:
```suggestion
        response.dig('errorInformation', 'message') || response.dig('message')
```

lib/active_merchant/billing/gateways/cyber_source_rest.rb
```diff
@@ -398,21 +398,21 @@ def success_from(response)
...
       end
 
       def authorization_from(response)
         id = response['id']
-        has_amount = response['orderInformation'] && response['orderInformation']['amountDetails'] && response['orderInformation']['amountDetails']['authorizedAmount']
-        amount = response['orderInformation']['amountDetails']['authorizedAmount'].delete('.') if has_amount
+        has_amount = response['orderInformation'] && response.dig('orderInformation', 'amountDetails') && response.dig('orderInformation', 'amountDetails', 'authorizedAmount')
```
bradbroge, line 406: dig makes this much less complex
```suggestion
        has_amount = response.dig('orderInformation', 'amountDetails', 'authorizedAmount')
```

bradbroge (commented):
mostly looks good, a couple of things that could be made better.

lib/active_merchant/billing/gateways/cyber_source_rest.rb
almalee24: I rather keep it as response['message'] since it's not looking it will just return nil without breaking if message is not present

lib/active_merchant/billing/gateways/cyber_source_rest.rb
almalee24: Sounds good! I wonder why it was even looking for amountDetails

bradbroge (approved):
Looks great 👍🏻

PR #5280, opened 2024-09-30
Adyen: remove unused ecommerce flag for NetworkTokenization cards

DustinHaefele:


lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -1006,7 +1006,7 @@ def skip_mpi_data?(options = {})
       end
 
       def ecommerce_shopper_interaction?(payment, options)
-        return true if payment.is_a?(NetworkTokenizationCreditCard) && !options[:switch_cryptogram_mapping_nt]
```
almalee24, line 1009: So this means that we always want to pass ecommerce for NT/AP/GP transactions?

DustinHaefele: Yeah That is how it was before my changes, and I would rather not update that as well when we are working through the GooglePay liability shift. I think for these types the liability shift doesn't depend on this flagging, but we may need to update this in the future.

almalee24 (approved):

PR #5281, opened 2024-09-30
DLocal: Add X-Dlocal-Payment-Source to header

almalee24:
Remote
40 tests, 110 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -299,6 +299,7 @@ module ActiveMerchant #:nodoc:
           'Authorization' => signature(post, timestamp)
         }
         headers['X-Idempotency-Key'] = options[:idempotency_key] if options[:idempotency_key]
+        headers['X-Dlocal-Payment-Source'] = application_id if application_id
         headers
       end
 
--- a/test/unit/gateways/d_local_test.rb
+++ b/test/unit/gateways/d_local_test.rb
@@ -37,10 +37,12 @@ class DLocalTest < Test::Unit::TestCase
   end
 
   def test_purchase_with_save
+    DLocalGateway.application_id = 'ActiveMerchant'
     stub_comms do
       @gateway.purchase(@amount, @credit_card, @options.merge(save: true))
-    end.check_request do |_endpoint, data, _headers|
+    end.check_request do |_endpoint, data, headers|
       assert_equal true, JSON.parse(data)['card']['save']
+      assert_equal 'ActiveMerchant', headers['X-Dlocal-Payment-Source']
     end.respond_with(successful_purchase_response)
   end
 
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -294,6 +294,7 @@ def headers(post, options = {})
           'Authorization' => signature(post, timestamp)
         }
         headers['X-Idempotency-Key'] = options[:idempotency_key] if options[:idempotency_key]
+        headers['X-Dlocal-Payment-Source'] = options[:payment_source] if options[:payment_source]
```
aenand, line 297: Is this static? If so you may want to use the existing pattern that `application_id` offers

almalee24: I'll update it, I new we had a field like that but I could remember the name

aenand: `application_id` is not a field in option, it's a class attribute. So in the gateway class you can access it by calling `application_id` and then it's set on the class (`ActiveMerchant::Billing::{{gateway}}.application_id = 'xyz'`). 

AuthorizeNet seems to follow a similar pattern to this and uses application_id if you'd like to copy it

lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -294,6 +294,7 @@ def headers(post, options = {})
           'Authorization' => signature(post, timestamp)
         }
         headers['X-Idempotency-Key'] = options[:idempotency_key] if options[:idempotency_key]
+        headers['X-Dlocal-Payment-Source'] = options[:application_id] if options[:application_id]
```
aenand, line 297: This is not how application_id gets passed in. 

application_id is a class variable so it would be passed in like

```suggestion
        headers['X-Dlocal-Payment-Source'] = application_id if application_id
```

test/unit/gateways/d_local_test.rb
```diff
@@ -38,9 +38,10 @@ def test_successful_purchase
 
   def test_purchase_with_save
     stub_comms do
-      @gateway.purchase(@amount, @credit_card, @options.merge(save: true))
-    end.check_request do |_endpoint, data, _headers|
+      @gateway.purchase(@amount, @credit_card, @options.merge({ save: true, application_id: 'ActiveMerchant' }))
```
aenand, line 41: `application_id` does not get passed in via options. It is a class variable so you should set it like `ActiveMerchant::Billing::DLocalGateway.application_id = 'ActiveMerchant'`

aenand (changes requested):

aenand (approved):

PR #5282, opened 2024-09-30
Orbital: Update alternate_ucaf_flow

almalee24:
Update UCAFInd to only be passed if the ECI value is 4, 6 or 7.

Remote
134 tests, 543 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -758,15 +758,9 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_mc_ucafind(xml, credit_card, three_d_secure, options)
-        return unless three_d_secure
+        return unless three_d_secure && %w(4 6 7).include?(three_d_secure&.dig(:eci))
 
-        if options[:alternate_ucaf_flow]
-          return unless %w(4 6 7).include?(three_d_secure[:eci])
-
-          xml.tag! :UCAFInd, options[:ucaf_collection_indicator] if options[:ucaf_collection_indicator]
-        else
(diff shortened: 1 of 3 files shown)


sinourain (approved):
LGTM!

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -758,15 +758,9 @@ def add_mc_directory_trans_id(xml, credit_card, three_d_secure)
...
-
-        if options[:alternate_ucaf_flow]
-          return unless %w(4 6 7).include?(three_d_secure[:eci])
+        return unless three_d_secure && %w(4 6 7).include?(three_d_secure[:eci])
 
-          xml.tag! :UCAFInd, options[:ucaf_collection_indicator] if options[:ucaf_collection_indicator]
-        else
-          xml.tag! :UCAFInd, options[:ucaf_collection_indicator] || '4'
-        end
+        xml.tag! :UCAFInd, options[:ucaf_collection_indicator] if options[:ucaf_collection_indicator]
```
DustinHaefele, line 763: What is the purpose of removing the `4` default? I hate random defaults like this so I like removing it, but just curious on the reasoning?

almalee24: That we should only be sending the filed if options[:ucaf_collection_indicator] at least that's how the original ticket was written for this.

DustinHaefele (approved):

DustinHaefele:
Also if you rebase this with master the tests should pass. I patched the testing issues last week.

PR #5287, opened 2024-10-02
update active support version for rails master test

DustinHaefele:
This is a temporary fix to allow for our gha tests to run while we work on updating the ruby version of AM to >= 3.2.0.

Hopefully that version update will be fairly simple and we can get back to truly testing rails master shortly, but it doesn't make sense to let not run gha tests on pull requests while we work on that.


DeeMeyers (approved):
lgtm!

sinourain (approved):
LGTM!

PR #5290, opened 2024-10-03
Redsys: update decoding to catch Json errors and try to urlsafe_decode

DustinHaefele:
Remote tests:
26 tests, 91 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.1538% passed

Unit Tests:
27 tests, 113 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

^^ Same remote failure is seen on master.

### DIFF
--- a/lib/active_merchant/billing/gateways/redsys_rest.rb
+++ b/lib/active_merchant/billing/gateways/redsys_rest.rb
@@ -361,7 +361,12 @@ module ActiveMerchant #:nodoc:
         payload = raw_response['Ds_MerchantParameters']
         return Response.new(false, "#{raw_response['errorCode']} ERROR") unless payload
 
-        response = JSON.parse(Base64.decode64(payload)).transform_keys!(&:downcase).with_indifferent_access
+        begin
+          response = JSON.parse(Base64.decode64(payload)).transform_keys!(&:downcase).with_indifferent_access
+        rescue JSON::ParserError
+          response = JSON.parse(Base64.urlsafe_decode64(payload)).transform_keys!(&:downcase).with_indifferent_access
+        end
+
         return Response.new(false, 'Unable to verify response') unless validate_signature(payload, raw_response['Ds_Signature'], response[:ds_order])
 
         succeeded = success_from(response, options)


almalee24:
There's a small rubocop error

lib/active_merchant/billing/gateways/redsys_rest.rb
```diff
@@ -361,7 +361,12 @@ def commit(post, options)
         payload = raw_response['Ds_MerchantParameters']
         return Response.new(false, "#{raw_response['errorCode']} ERROR") unless payload
 
-        response = JSON.parse(Base64.decode64(payload)).transform_keys!(&:downcase).with_indifferent_access
+        begin
+          response = JSON.parse(Base64.decode64(payload)).transform_keys!(&:downcase).with_indifferent_access
```
travismorrison, line 365: @DustinHaefele does the Base64 interface really ask that you catch an exception and then try another method in the interface? if you don't mind tell me a little more about it, although I know there is urgency. the thing that kinda bothers me is--what if `urlsafe_decode64` raises an error.

DustinHaefele: Great question.

The issue is with a standard base64 encoding as opposed to the urlsafe_encoding there are different character sets. In this case when we are trying to do a standard base64 decoding, if there is a `?` it will not decode correctly and it will break the JSON syntax because everything after the `?` will be something like. `\x0FY\x0ELM\x18\x8DL\xD8\xCBML\x8DL\xCBM\` and wont end up with the correct closing brackets. 

It's likely that we could always use the `urlsafe_decode64` and remove the first block but with an emergency change I wanted to be abundantly safe and only fix the specific error that we are seeing which we know the root cause of...That damn `?`... I didn't do it whole sale because I'm not sure if there are characters that will cause issues while using the urlsafe_decode, in the same way that we are seeing the `?` cause issues here. I would like more time to explore that and then put the change in whole sale with a ticket to address it. 

If the urlsafe_decode throws an error after the standard decode throws an error than it is a different issue, and I'm fine with an exception being raised.

travismorrison: ok cool I will recuse myself. makes sense

PR #5302, opened 2024-10-10
Cybersource: Update order of XML fields

almalee24:
Update order of XML fields for add_auth_service with this new order aggregatorID, reconciliationID, firstRecurringPayment and mobileRemotePaymentType can now be passed for NetworkTokenizationCreditCard.

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -39,7 +39,6 @@ module ActiveMerchant #:nodoc:
         delegated_authentication: 'delegatedAuthenticationExemptionIndicator',
         low_risk: 'riskAnalysisExemptionIndicator',
         low_value: 'lowValueExemptionIndicator',
-        stored_credential: 'stored_credential',
         trusted_merchant: 'trustedMerchantExemptionIndicator'
       }
       DEFAULT_COLLECTION_INDICATOR = 2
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -738,13 +737,8 @@ def add_decision_manager_fields(xml, options)
...
         return unless network_tokenization?(payment_method)
 
-        case payment_method.source
-        when :network_token
-          payment_solution = NT_PAYMENT_SOLUTION[payment_method.brand]
-          xml.tag! 'paymentSolution', payment_solution if payment_solution
-        when :apple_pay, :google_pay
-          xml.tag! 'paymentSolution', @@wallet_payment_solution[payment_method.source]
-        end
+        payment_solution = payment_method.try(:network_token?) ? NT_PAYMENT_SOLUTION[payment_method.brand] : @@wallet_payment_solution[payment_method.source]
```
Buitragox, line 740: Minor detail. I think we could skip the `.try` here, since the early return guarantees that it is a `NetworkTokenizationCreditCard`. Same thing in the `add_payment_network_token` method.

almalee24: NetworkTokenizationCreditCard can be Network Token, ApplePay or GooglePay and `payment_method.try(:network_token?) ` ensures that it is a Network Token.

Buitragox: Every `NetworkTokenizationCreditCard` should have the method `#network_token?` available. I think we could call `payment_method.network_token?` directly instead of using `try`.

almalee24: @Buitragox you're right, I've update it

PR #5307, opened 2024-10-15
Nuvei:  Adding account founding transaction

javierpedrozaing:
Description
-------------------------
[SER-1457](https://spreedly.atlassian.net/browse/SER-1457)

This commit add Account funding transaction (AFT) for Nuvei

Unit test
-------------------------
Finished in 0.090215 seconds.

19 tests, 104 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

210.61 tests/s, 1152.80 assertions/s

Remote test
-------------------------
Finished in 105.744939 seconds.

29 tests, 90 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

0.27 tests/s, 0.85 assertions/s

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/nuvei.rb
+++ b/lib/active_merchant/billing/gateways/nuvei.rb
@@ -43,6 +43,7 @@ module ActiveMerchant
         add_address(post, payment, options)
         add_customer_ip(post, options)
         add_stored_credentials(post, payment, options)
+        add_account_funding_transaction(post, payment, options)
         post[:userTokenId] = options[:user_token_id] if options[:user_token_id]
         post[:isPartialApproval] = options[:is_partial_approval] ? 1 : 0
         post[:authenticationOnlyType] = options[:authentication_only_type] if options[:authentication_only_type]
@@ -122,6 +123,18 @@ module ActiveMerchant
           storedCredentialsMode: stored_credentials_mode
         }
         post[:isRebilling] = stored_credentials_mode
+      end
+
+      def add_account_funding_transaction(post, payment, options = {})
+        return unless options[:is_aft]
+
+        recipient_details = {
+          firstName: payment.first_name,
+          lastName: payment.last_name,
+          country: options.dig(:billing_address, :country)
+        }.compact
+
+        post[:recipientDetails] = recipient_details unless recipient_details.empty?
       end
 
       def set_reason_type(post, options)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -123,6 +125,18 @@ def set_initiator_type(post, payment, options)
         post[:isRebilling] = stored_credentials_mode
       end
 
+      def add_account_funding_transaction(post, payment = {}, options = {})
+        return unless options[:is_aft]
+
+        recipient_details = {
+          firstName: options[:cardholder_first_name] || payment.first_name,
+          lastName: options[:cardholder_last_name] || payment.last_name,
+          country: options.dig(:cardholder_country) || options.dig(:billing_address, :country)
```
naashton, line 134: I don't think we want to use custom fields, when we can just use the standard fields. Just use the values we get from `payment` that's given and the `billing_address` options.

lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -123,6 +125,18 @@ def set_initiator_type(post, payment, options)
         post[:isRebilling] = stored_credentials_mode
       end
 
+      def add_account_funding_transaction(post, payment = {}, options = {})
```
naashton, line 128: Since we are executing the AFT via the `authorize` transaction, we should always have a payment method, so I don't think you need to give it a default value.

naashton (changes requested):
I think we don't need to implement any custom values, we should be able to get everything we need from standard fields for AFT.

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: Hi @naashton, thanks! I did this for two reasons. The first is that the ticket states, "if the GSFs recipientFirstName, recipientLastName, recipientCountry are present, use those values; otherwise, map from the card information." I interpreted this as referring to GSF. Additionally, I saw in the documentation that it's possible to send a refund request with AFT values, and in that refund request, we don't have a payment method.

javierpedrozaing: For the second option, I misinterpreted it. The refund only shows us in the response whether the initial purchase was AFT or not.

javierpedrozaing: finally I think that make sense only use the firstName and LastName coming from the Payment. Thanks

naashton (approved):

PR #5314, opened 2024-10-22
StripePI: Update version to 2022-11-15

almalee24:
Update version to 2022-11-15 and charges to be
latest_charge. latest_charge is no longer
automatically expanded andwill only be expanded
if the repsonse is succesful.

Remote
99 tests, 463 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 98.9899% passed


jherreraa (approved):
LGTM 👍

PR #5319, opened 2024-10-29
Normalize API versions

Buitragox:
Summary:
This change centralizes version management at the class level, creating a consistent and reusable way to define and access API versions.

Unit Tests:
6103 tests, 80806 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/adyen.rb
+++ b/lib/active_merchant/billing/gateways/adyen.rb
@@ -17,8 +17,9 @@ module ActiveMerchant # :nodoc:
       self.homepage_url = 'https://www.adyen.com/'
       self.display_name = 'Adyen'
 
-      PAYMENT_API_VERSION = 'v68'
-      RECURRING_API_VERSION = 'v68'
+      version 'v68', :payment_api
+      version 'v68', :payout_api
+      version 'v68', :recurring_api
 
       STANDARD_ERROR_CODE_MAPPING = {
         '0' => STANDARD_ERROR_CODE[:processing_error],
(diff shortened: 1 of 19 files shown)


almalee24:
The PR looks good so far could you add some unit tests to verify that the API version is being added correctly?

DustinHaefele:
Looks like some linting errors are causing the tests to fail here.

DustinHaefele:
> The same thing is happening in #5326. I don't get this error locally using the same rubocop version (1.14.0). I think #5325 should fix this.

That is super weird. It looks like a rule that we don't even enable in the config file `Style/AccessModifierDeclarations`... If that commit doesn't fix it you might need to disable that rule specifically.

lib/active_merchant/billing/gateways/adyen.rb
```diff
@@ -870,11 +871,11 @@ def cvv_result_from(response)
       def endpoint(action)
         case action
         when 'disable', 'storeToken'
-          "Recurring/#{RECURRING_API_VERSION}/#{action}"
```
almalee24, line 873: Can you still please add a adyen unit test to very the version is being properly passed? I think adding one unit test to all the gateways would be good

Buitragox: I added tests to check that the URLs are being properly build. For the gateways that don't use the API version (braintree and hps), I'm not sure if it's worth adding a test. Let me know what you think of this approach.

almalee24 (approved):

PR #5322, opened 2024-11-04
StripePI: Last4 From Payment Method

naashton:
For some transactions and payment methods, the `options[:last_4]` will not be prepopluated, so we need to look in the `payment_method` to slice the last 4 from the number

Remote: 100 tests, 472 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit: 68 tests, 357 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Rubocop:
nick-mbp:active_merchant nick$ bundle exec rubocop
Inspecting 804 files
....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

804 files inspected, no offenses detected

### DIFF
--- a/test/remote/gateways/remote_stripe_payment_intents_test.rb
+++ b/test/remote/gateways/remote_stripe_payment_intents_test.rb
@@ -275,6 +275,17 @@ class RemoteStripeIntentsTest < Test::Unit::TestCase
     assert_not_nil(purchase.params.dig('charges', 'data')[0]['payment_method_details']['card']['network_token'])
   end
 
+  def test_successful_purchase_with_network_token_cc
+    options = {
+      currency: 'USD'
+    }
+
+    purchase = @gateway.purchase(@amount, @network_token_credit_card, options)
+    assert_equal(nil, purchase.responses.first.params.dig('token', 'card', 'tokenization_method'))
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -275,6 +275,17 @@ def test_successful_purchase_with_tokenized_visa
     assert_not_nil(purchase.params.dig('charges', 'data')[0]['payment_method_details']['card']['network_token'])
   end
 
+  def test_successful_purchase_with_network_token_cc
```
almalee24, line 278: Can you also unit test were we confirm that we pass last_4 even when last_4 is not present in options? Similar to this [test](https://github.com/activemerchant/active_merchant/blob/3b28d8622f26121b0eff29d4f0b6bc9be6c817cd/test/unit/gateways/stripe_payment_intents_test.rb#L530)

naashton:
Also no idea why rubocop is failing - I don't get that same lint issue locally 🤷

almalee24 (approved):

PR #5325, opened 2024-11-05
Update Rubocop 1.26.0

almalee24:
Update done using bundle exec rubocop -a


.rubocop.yml
```diff
@@ -126,3 +126,47 @@ Style/OptionalBooleanParameter:
...
+Lint/EmptyInPattern: # new in 1.16
+  Enabled: true
+Lint/IncompatibleIoSelectWithFiberScheduler: # new in 1.21
+  Enabled: true
+Lint/RequireRelativeSelfPath: # new in 1.22
+  Enabled: true
+Lint/UselessRuby2Keywords: # new in 1.23
+  Enabled: true
+Naming/BlockForwarding: # new in 1.24
+  Enabled: true
```
DustinHaefele, line 146: Can we update this rule to the following? Or just disable it. Unless there are strong opinions otherwise I think it is easier to read a function that is passing `&block` instead of just `&`. It's updated it in a couple places with the autocorrect. I'll note the first one I come to.
```suggestion
   EnforcedStyle: explicit
```

lib/active_merchant/billing/compatibility.rb
```diff
@@ -90,8 +90,8 @@ def add_to_base(error)
             add(:base, error)
           end
 
-          def each_full(&block)
-            full_messages.each(&block)
+          def each_full(&)
```
DustinHaefele, line 93: Here is one place for what I referred to in the rule above. Here is a link to that rule if the docs help. https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/BlockForwarding

DustinHaefele (commented):
Left a couple comments, but overall looks good.

.rubocop.yml
almalee24: Yeah I'll disable

activemerchant.gemspec
```diff
@@ -34,4 +34,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rake')
   s.add_development_dependency('test-unit', '~> 3')
   s.add_development_dependency('thor')
+  s.metadata['rubygems_mfa_required'] = 'true'
```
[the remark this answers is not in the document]
almalee24: [rubygems_mfa_required](https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecrequiremfa)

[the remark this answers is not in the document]
almalee24: But I'm thinking I should also disable this one

DustinHaefele (approved):
Looks good. Thanks for the updates.

PR #5326, opened 2024-11-05
DLocal: add country override optional field

yunnydang:
Adds the optional country field override

Local:
6104 tests, 80829 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
8 tests, 209 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
42 tests, 118 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/d_local.rb
+++ b/lib/active_merchant/billing/gateways/d_local.rb
@@ -110,9 +110,10 @@ module ActiveMerchant #:nodoc:
       end
 
       def add_country(post, card, options)
-        return unless address = options[:billing_address] || options[:address]
+        return unless (address = options[:billing_address] || options[:address]) || options[:country]
 
-        post[:country] = lookup_country_code(address[:country])
+        country = options[:country] ? lookup_country_code(options[:country]) : lookup_country_code(address[:country])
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/d_local.rb
```diff
@@ -112,7 +112,8 @@ def add_additional_data(post, options)
       def add_country(post, card, options)
         return unless address = options[:billing_address] || options[:address]
```
jcreiff, line 113: We need to add a provision for `options[:country]` in this or statement, otherwise if it's passed through options but the address hashes empty, we'll return of the method

jcreiff (changes requested):

lib/active_merchant/billing/gateways/d_local.rb
yunnydang: ah right thanks for catching that!

jcreiff (approved):
Looks good!

ali-hassan (approved):

PR #5331, opened 2024-11-11
CommerceHub: Add Network Token support

almalee24:
Remote
36 tests, 96 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 94.4444% passed

Unit
29 tests, 205 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/commerce_hub.rb
+++ b/lib/active_merchant/billing/gateways/commerce_hub.rb
@@ -311,20 +311,31 @@ module ActiveMerchant #:nodoc:
         source[:walletType] = payment.source.to_s.upcase
       end
 
+      def add_network_token(source, payment, options)
+        source[:sourceType] = 'PaymentToken'
+        source[:tokenData] = payment.number
+        source[:tokenSource] = 'NETWORK_TOKEN'
+        source[:cryptogram] = payment.payment_cryptogram
+        source[:card] = {
+          expirationMonth: format(payment.month, :two_digits),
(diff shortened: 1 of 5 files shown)


lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -311,20 +311,31 @@ def add_decrypted_wallet(source, payment, options)
...
-            source[:sourceType] = 'PaymentCard'
-            source[:encryptionData] = options[:encryption_data]
-          else
-            add_credit_card(source, payment, options)
-          end
-        when String
+
+        if payment.is_a?(String)
           add_payment_token(source, payment, options)
+        elsif payment.mobile_wallet?
```
DustinHaefele, line 330: Can you define this method in the `Check` model? If we end up adding that payment method type in the future we will get a no method error here if it isn't added there. I should have added it there originally when I created the `mobile_wallet?` method but only added it to credit card.

```
def mobile_wallet?
  false
end
```

DustinHaefele (commented):
One request, but this looks good.

lib/active_merchant/billing/gateways/commerce_hub.rb
almalee24: Good idea!

DustinHaefele (approved):

PR #5332, opened 2024-11-11
Credorax: Add support for gateway card validation

rachelkirk:
This change allows the `cardholder_name_inquiry` flag to be sent. If set to true, it will trigger first and last name data to be sent for c22, and c23. $0 auth transactions to validate cards also require the value for a9 to be set to 5.

Remote Tests:
52 tests, 176 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 88.4615% passed
*6 failing tests also failing on master

Unit Tests
83 tests, 399 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -165,6 +165,7 @@ module ActiveMerchant # :nodoc:
         add_echo(post, options)
         add_submerchant_id(post, options)
         add_stored_credential(post, options)
+        add_account_name_inquiry(post, options)
         add_processor(post, options)
         add_authorization_details(post, options)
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -164,6 +164,7 @@ def authorize(amount, payment_method, options = {})
         add_3ds_2_optional_fields(post, options)
         add_echo(post, options)
         add_submerchant_id(post, options)
+        add_cardholder_name_inquiry(post, options)
```
jcreiff, line 167: I wonder if this should come after `add_stored_credential` in the order of sub-methods?

The reason being that `a9` is also set in that submethod, either by applying the `transaction_type` that is passed in, or defaulting to `9` if it's a CIT.

In this current order, I think there's a risk that the `a9` value we're setting would get overwritten 😬

jcreiff (approved):
🚀

PR #5334, opened 2024-11-13
Nuvei: Fix NTID stored credentials

javierpedrozaing:
Description
-------------------------
[SER-1510](https://spreedly.atlassian.net/browse/SER-1510)

Unit test
-------------------------
Finished in 0.910025 seconds.

26 tests, 132 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

28.57 tests/s, 145.05 assertions/s

Remote test
-------------------------
Finished in 112.665892 seconds.

38 tests, 120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

0.34 tests/s, 1.07 assertions/s

Rubocop
-------------------------
804 files inspected, no offenses detected


lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -178,7 +178,7 @@ def scrub(transcript)
       private
 
       def network_transaction_id_from(response)
-        response.dig('paymentOption', 'paymentAccountReference')
+        response.dig('transactionId')
```
naashton, line 181: This looks like the `transactionId` is just a generic id for the transaction, but can also be the NTID for a COF transaction?

naashton (approved):
No issues really, just looking to clarify the point on the `transactionId`. I did peak at Nuvei's docs and didn't see a specific NTID value.

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: Hi @naashton yes, according to the Nuvei documentation the proper value is `transactionId`, this always represents the previous or original transaction CIT and should be sent as the `relatedTransactionId`. The  `paymentAccountReference ` does not represent a transaction id, it represents a unique identifier associated with a specific cardholder PAN.

PR #5338, opened 2024-11-14
Fix the stripe PI add three d supported field logic

yunnydang:
This fix allows us to actually return a false boolean for the three d supported field

Local:
6109 tests, 80773 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
69 tests, 363 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
01 tests, 451 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.0297% passed


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -729,9 +729,9 @@ def add_name_only(post, payment_method)
 
       # This surfaces the three_d_secure_usage.supported field and saves it as an instance variable so that we can access it later on in the response
       def add_card_3d_secure_usage_supported(response)
-        return unless response.params['card'] && response.params['card']['three_d_secure_usage']
+        return unless response.params['card'] && response.params['card']['three_d_secure_usage'] && response.params['card']['three_d_secure_usage']['supported']
```
jcreiff, line 732: We should leave off this final `&&` because we need to be able to return a `false` value

jcreiff (changes requested):

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
yunnydang: ope! Youre totally right thanks for catching this!

jcreiff (approved):
Looks good!

PR #5340, opened 2024-11-18
CommerceHub: Update Production URL

almalee24:
Unit:
28 tests, 194 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote:
35 tests, 94 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 94.2857% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/commerce_hub.rb
+++ b/lib/active_merchant/billing/gateways/commerce_hub.rb
@@ -2,7 +2,7 @@ module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class CommerceHubGateway < Gateway
       self.test_url = 'https://connect-cert.fiservapps.com/ch'
-      self.live_url = 'https://prod.api.fiservapps.com/ch'
+      self.live_url = 'https://connect.fiservapis.com/ch'
 
       self.supported_countries = ['US']
       self.default_currency = 'USD'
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -2,7 +2,7 @@ module ActiveMerchant #:nodoc:
   module Billing #:nodoc:
     class CommerceHubGateway < Gateway
       self.test_url = 'https://connect-cert.fiservapps.com/ch'
-      self.live_url = 'https://prod.api.fiservapps.com/ch'
+      self.live_url = 'https://connect.fiservapis.com/ch/'
```
DustinHaefele, line 5: Two questions 

1. Is the `/` that was added to the end going to be an issue? 
2. How confident are we that this url is going to be good to go? I just worry since our remote tests all use the test_url.

almalee24: I've removed the '/' it was in the ticket but it doesn't need to be added

DustinHaefele: Perfect!

DustinHaefele (approved):

PR #5342, opened 2024-11-22
Ebanx: Fix processing_type header

sinourain:
Fix processing_type header

Spreedly reference: OPPS-220

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -52,7 +52,7 @@ module ActiveMerchant # :nodoc:
         add_additional_data(post, options)
         add_stored_credentials(post, options)
 
-        commit(:purchase, post)
+        commit(:purchase, post, options)
       end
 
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -147,6 +156,10 @@ def scrub(transcript)
 
       private
 
+      def add_processing_type(post, options)
+        post[:processing_type] = options[:processing_type] if options[:processing_type]
```
almalee24, line 160: do you know if processing_type is an actual field that Ebanx supports the request bode or is this just for setting the header? If it's just for the header then I would suggest passing options all the way down to the headers method

almalee24: By passing the field like this we are sending it in the request body and not just the header. If this is just a field that is passed in hear then we shouldn't be passing it in the request body

almalee24 (approved):

PR #5349, opened 2024-11-27
Fortis: Initial implementation

Heavyblade:
Summary:
Fortis Gateway initial implementation with support for:

- Authorize
- Purchase
- Void
- Capture
- Refund
- Credit
- Store
- Unstore

### Context:
There is an issue with the Fortis API because their back-end returns an "invalid credentials" error if the HTTP Request headers are not in lower-case format, but Ruby by default follows the [HTTP 1.1 RFC 2616 ](https://datatracker.ietf.org/doc/html/rfc2616#page-31:~:text=an%0A%20%20%20extra%20CRLF.-,4.2,-Message%20Headers%0A%0A%20%20%20HTTP) on how to handle headers; Ruby camellize all headers from something like 'user-api-key' to 'User-Api-Key', that is performed by [Net::HTTPHeader#capitalize](https://github.com/ruby/net-http/blob/master/lib/net/http/header.rb#L493), so to deal with that we had to find a way to force net/http to send lower-case headers.

### Enter Refinements:
So with Ruby Refinements, we can monkey patch net/http but only in the context of ActiveMerchant requests without side-effects on other areas of the system, two changes were introduced:

* The `CaseSensitiveHeaders` class, an alias for a hash, allows us to indicate in the Fortis adapter that we will need case-sensitive headers.
* The net/http refinement => that creates one HTTP::Request class per HTTP method, injecting a module that will overwrite the #capitalize method, it also overwrites the post, put, patch and delete methods to use the custom requests  only when the headers sent to the Connection class are `CaseSensitiveHeaders`.

Remote Test:
Finished in 114.375184 seconds.
22 tests, 53 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
0.19 tests/s, 0.46 assertions/s

Unit Tests:
Finished in 68.455283 seconds.
6138 tests, 80277 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

RuboCop:
807 files inspected, no offenses detected

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -31,6 +31,7 @@ RuboCop::RakeTask.new
 
 namespace :test do
   Rake::TestTask.new(:units) do |t|
+    ENV['RUNNING_UNIT_TESTS'] = 'true'
     t.pattern = 'test/unit/**/*_test.rb'
     t.libs << 'test'
     t.verbose = false
--- a/lib/active_merchant/billing/gateways/fortis.rb
+++ b/lib/active_merchant/billing/gateways/fortis.rb
@@ -1,0 +1,331 @@
+module ActiveMerchant # :nodoc:
+  module Billing # :nodoc:
+    class FortisGateway < Gateway
+      self.test_url = 'https://api.sandbox.fortis.tech/v1'
+      self.live_url = 'https://api.fortis.tech/v1'
+
+      self.supported_countries = %w{US CA}
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover jbc unionpay]
+      self.money_format = :cents
+      self.homepage_url = 'https://fortispay.com'
+      self.display_name = 'Fortis'
+
+      STATUS_MAPPING = {
+        101 => 'Sale cc Approved',
+        102 => 'Sale cc AuthOnly',
+        111 => 'Refund cc Refunded',
+        121 => 'Credit/Debit/Refund cc AvsOnly',
+        131 => 'Credit/Debit/Refund ach Pending Origination',
+        132 => 'Credit/Debit/Refund ach Originating',
+        133 => 'Credit/Debit/Refund ach Originated',
+        134 => 'Credit/Debit/Refund ach Settled',
+        191 => 'Settled (deprecated - batches are now settled on the /v2/transactionbatches endpoint)',
+        201 => 'All cc/ach Voided',
+        301 => 'All cc/ach Declined',
+        331 => 'Credit/Debit/Refund ach Charged Back'
+      }
+
+      REASON_MAPPING = {
+        0 => 'N/A',
+        1000 => 'CC - Approved / ACH - Accepted',
+        1001 => 'AuthCompleted',
+        1002 => 'Forced',
+        1003 => 'AuthOnly Declined',
+        1004 => 'Validation Failure (System Run Trx)',
+        1005 => 'Processor Response Invalid',
+        1200 => 'Voided',
+        1201 => 'Partial Approval',
+        1240 => 'Approved, optional fields are missing (Paya ACH only)',
+        1301 => 'Account Deactivated for Fraud',
+        1500 => 'Generic Decline',
+        1510 => 'Call',
+        1518 => 'Transaction Not Permitted - Terminal',
+        1520 => 'Pickup Card',
+        1530 => 'Retry Trx',
+        1531 => 'Communication Error',
+        1540 => 'Setup Issue, contact Support',
+        1541 => 'Device is not signature capable',
+        1588 => 'Data could not be de-tokenized',
+        1599 => 'Other Reason',
+        1601 => 'Generic Decline',
+        1602 => 'Call',
+        1603 => 'No Reply',
+        1604 => 'Pickup Card - No Fraud',
+        1605 => 'Pickup Card - Fraud',
+        1606 => 'Pickup Card - Lost',
+        1607 => 'Pickup Card - Stolen',
+        1608 => 'Account Error',
+        1609 => 'Already Reversed',
+        1610 => 'Bad PIN',
+        1611 => 'Cashback Exceeded',
+        1612 => 'Cashback Not Available',
+        1613 => 'CID Error',
+        1614 => 'Date Error',
+        1615 => 'Do Not Honor',
+        1616 => 'NSF',
+        1618 => 'Invalid Service Code',
+        1619 => 'Exceeded activity limit',
+        1620 => 'Violation',
+        1621 => 'Encryption Error',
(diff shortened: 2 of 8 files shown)


Rakefile
```diff
@@ -31,6 +31,7 @@ RuboCop::RakeTask.new
 
 namespace :test do
   Rake::TestTask.new(:units) do |t|
+    ENV['RUNNING_UNIT_TESTS'] = 'true'
```
javierpedrozaing, line 34: Why is this ENV needed ❓

lib/active_merchant/billing/gateways/fortis.rb
```diff
@@ -0,0 +1,330 @@
...
+        132 => 'Credit/Debit/Refund ach Originating',
+        133 => 'Credit/Debit/Refund ach Originated',
+        134 => 'Credit/Debit/Refund ach Settled',
+        191 => 'Settled (deprecated - batches are now settled on the /v2/transactionbatches endpoint)',
+        201 => 'All cc/ach Voided',
+        301 => 'All cc/ach Declined',
+        331 => 'Credit/Debit/Refund ach Charged Back'
+      }
+
+      REASON_MAPPING = {
```
javierpedrozaing, line 29: was the mapping for status and reason needed due the Fortis response does not return a readable message, or was it for another specific  reason ❓

lib/active_merchant/billing/gateways/fortis.rb
```diff
@@ -0,0 +1,330 @@
...
+          'developer-id' => @options[:developer_id]
+        })
+      end
+
+      def add_location_id(post)
+        post[:location_id] = @options[:location_id]
+      end
+
+      def commit(action, post, method = :post, options = {})
+        add_location_id(post) if post.present?
```
javierpedrozaing, line 261: what is the advantage of sending the `location_id` in the post ❓

lib/active_merchant/connection.rb
```diff
@@ -76,7 +76,7 @@ def request(method, body, headers = {})
               http.get(endpoint.request_uri, headers)
             when :post
               debug body
-              http.post(endpoint.request_uri, body, RUBY_184_POST_HEADERS.merge(headers))
+              http.post(endpoint.request_uri, body, headers.reverse_merge!(RUBY_184_POST_HEADERS))
```
javierpedrozaing, line 79: Is this change related to the HTTP version ❓

lib/active_merchant/net_http_ssl_connection.rb
```diff
@@ -1,11 +1,46 @@
 require 'net/http'
 
 module NetHttpSslConnection
+  module InnocuousCapitalize
+    def capitalize(name) = name
```
javierpedrozaing, line 5: ahh yes, funny thing with this gateway 👏🏼

lib/active_merchant/posts_data.rb
```diff
@@ -92,5 +92,16 @@ def handle_response(response)
         raise ResponseError.new(response)
       end
     end
+
+    # This class is needed to play along with the Refinement done for Net::HTTP
+    # class so it can have a way to detect if the hash that represent the headers
+    # should use the case sensitive version of the headers or not.
+    class CaseSensitiveHeaders < Hash
```
javierpedrozaing, line 99: Nice 🔥

javierpedrozaing (approved):
Awesome job 🚀

Rakefile
Heavyblade: Oh, this one is a workaround for preventing Net::HTTP refinement from having some effect on unit tests mocs on test/unit/connection_test.rb

lib/active_merchant/billing/gateways/fortis.rb
Heavyblade: Most of the times Fortis returns an status_mapping and a reason_mapping but not the actual message

lib/active_merchant/billing/gateways/fortis.rb
Heavyblade: Is an optional normally that is usefull to identify the request source (Fortis customer normally are shops with card machines or terminals) and for the store transaction is a must.

lib/active_merchant/connection.rb
naashton: Why using reverse merge? Is order of these headers also important?

lib/active_merchant/net_http_ssl_connection.rb
```diff
@@ -1,11 +1,46 @@
...
+    private :capitalize
+  end
+
+  class CaseSensitivePost < Net::HTTP::Post; prepend InnocuousCapitalize; end
```
naashton, line 9: Is the `prepend` intended to give precedence to the existing behavior, which is capitalizing header fields?

naashton (commented):
Most of my review is around the changes to the http modules to accommodate this gateways quirkiness. It seems like the implementation should leave the existing gateways unaffected, but did you run any of the remote tests against any other gateways to see if they yielded any different results?

lib/active_merchant/billing/gateways/fortis.rb
```diff
@@ -0,0 +1,330 @@
...
+        1800 => 'Incorrect CVV',
+        1801 => 'Duplicate Transaction',
+        1802 => 'MID/TID Not Registered',
+        1803 => 'Stop Recurring',
+        1804 => 'No Transactions in Batch',
+        1805 => 'Batch Does Not Exist'
+      }
+
+      def initialize(options = {})
+        requires!(options, :user_id, :user_api_key, :developer_id, :location_id)
```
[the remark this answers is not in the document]
Heavyblade: You are right 👍

lib/active_merchant/net_http_ssl_connection.rb
Heavyblade: reponses:

1) no because of order, but because of classes see => hash_object.merge!(case_sensitve_header_hash), en-up being a plain ruby `Hash`, but case_sensitve_header_hash.merge!(hash_object) updates `CaseSensitiveHeaders`  instant but is still that class, so I can differentiate in the Refinement.

2) yes exactly that

lib/active_merchant/billing/gateways/fortis.rb
```diff
@@ -0,0 +1,331 @@
...
+        super
+      end
+
+      def authorize(money, payment, options = {})
+        commit path(:authorize, payment_type(payment)), auth_purchase_request(money, payment, options)
```
Buitragox, line 118: Just a minor question. Any reason for not putting parenthesis in the commit method call? Is this personal preference or something else?

naashton (approved):

PR #5354, opened 2024-12-03
Add the wallet_id field for decidir gateways

yunnydang:
Local:
6119 tests, 80836 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Decidir Unit:
46 tests, 223 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Decidir Remote:
29 tests, 99 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.1034% passed

Decidir Plus Unit:
20 tests, 84 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Decidir Plus Remote:
28 tests, 85 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
82.1429% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/decidir_plus.rb
+++ b/lib/active_merchant/billing/gateways/decidir_plus.rb
@@ -93,6 +93,12 @@ module ActiveMerchant # :nodoc:
         authorization.split('|')[0]
       end
 
+      def add_wallet_id(post, options)
+        return unless options[:wallet_id]
+
+        post[:wallet_id] = options[:wallet_id]
+      end
+
       def add_payment(post, payment, options = {})
         if payment.is_a?(String)
           token, bin = payment.split('|')
@@ -133,6 +139,7 @@ module ActiveMerchant # :nodoc:
 
         add_aggregate_data(post, options) if options[:aggregate_data]
         add_sub_payments(post, options)
+        add_wallet_id(post, options)
       end
 
       def add_aggregate_data(post, options)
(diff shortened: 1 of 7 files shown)


lib/active_merchant/billing/gateways/decidir_plus.rb
```diff
@@ -85,6 +85,7 @@ def build_purchase_authorize_request(post, money, payment, options)
         add_payment(post, payment, options)
         add_purchase_data(post, money, payment, options)
         add_fraud_detection(post, options)
+        add_wallet_id(post, options)
```
jcreiff, line 88: I notice that most of the fields that `wallet_id` gets grouped with in the original `decidir` gateway are located in `add_purchase_data`. Any particular reason to add it via this separate method instead of including in the existing method? I don't necessarily have a preference either way, just curious

jcreiff (approved):
Looks good!

lib/active_merchant/billing/gateways/decidir_plus.rb
yunnydang: hmmm, there wasnt much reasoning behind it! I just saw that we add fraud detection separately and thought itd be fine to do so for the wallet_id but I can place it inside the add_purchase_data instead!

PR #5359, opened 2024-12-04
OPPS-225 Worldpay Idempotency key fix

jherreraa:
# Unit tests
Finished in 0.418613 seconds.

138 tests, 756 assertions, 0 failures, 0 errors, 
100% passed

# Remote Tests
Finished in 118.498554 seconds.

117 tests, 500 assertions, 0 failures, 0 errors, 
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -947,7 +947,7 @@ module ActiveMerchant # :nodoc:
       end
 
       def headers(options)
-        idempotency_key = options[:idempotency_key]
+        @idempotency_key ||= options[:idempotency_key]
 
         headers = {
           'Content-Type' => 'text/xml',
@@ -960,7 +960,12 @@ module ActiveMerchant # :nodoc:
         cookie = options[:cookie] || cookie
         headers['Cookie'] = cookie if cookie
 
-        headers['Idempotency-Key'] = idempotency_key if idempotency_key
+        # Required because Worldpay does not accept duplicate idempotency keys
+        # for different transactions, such as in the case of an authorize => capture flow.
+        if @idempotency_key
+          headers['Idempotency-Key'] = @idempotency_key
+          @idempotency_key = SecureRandom.uuid
+        end
         headers
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -960,7 +960,12 @@ def headers(options)
         cookie = options[:cookie] || cookie
         headers['Cookie'] = cookie if cookie
 
-        headers['Idempotency-Key'] = idempotency_key if idempotency_key
+        # Required because Worldpay does not accept duplicate idempotency keys
+        # for different transactions, such as in the case of an authorize => capture flow.
+        if @idempotency_key
+          headers['Idempotency-Key'] = @idempotency_key
+          options[:idempotency_key] = SecureRandom.uuid
```
Buitragox, line 967: I think this line should be changing the instance variable
```suggestion
          @idempotency_key = SecureRandom.uuid
```

test/unit/gateways/worldpay_test.rb
```diff
@@ -223,6 +223,12 @@ def test_payment_type_for_credit_card
     assert_equal payment, :credit
   end
 
+  def test_idempotency_header
+    options = { idempotency_key: 'test123' }
+    headers = @gateway.send(:headers, options)
+    assert_not_equal options[:idempotency_key], headers['Idempotency-Key']
```
Buitragox, line 229: Probably need to change this test too.

Buitragox (changes requested):
Couple of things to fix. I also think it may be worth it to make a test that checks that it's actually changing the idempotency key between requests. Maybe something similar to the test `test_optional_idempotency_key_header` (that test is also missing an `assert_equal`, it's just comparing strings with `==`).
https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/worldpay_test.rb#L1505

test/unit/gateways/worldpay_test.rb
jherreraa: fixed!

lib/active_merchant/billing/gateways/worldpay.rb
jherreraa: fixed!

Buitragox (approved):
I have retested and everything looks good. Unit tests and remote tests are good.

PR #5361, opened 2024-12-06
Fortis: Adding Verify + Reference Purchase

Heavyblade:
## Summary:

Add neded changes to support reference authorize/purchase + supporting location_id as GSF for store

### Remote Test:
Finished in 181.033884 seconds.
25 tests, 58 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### Unit Tests:
Finished in 49.78482 seconds.
6143 tests, 80932 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### RuboCop:

### DIFF
--- a/lib/active_merchant/billing/gateways/fortis.rb
+++ b/lib/active_merchant/billing/gateways/fortis.rb
@@ -143,11 +143,18 @@ module ActiveMerchant # :nodoc:
         add_payment(post, payment, include_cvv: false)
         add_address(post, payment, options)
 
-        commit path(:store), post
+        commit path(:store), post, :post, options
       end
 
       def unstore(authorization, options = {})
         commit path(:unstore, authorization), nil, :delete
+      end
+
+      def verify(credit_card, options = {})
+        MultiResponse.run(:use_first_response) do |r|
+          r.process { authorize(100, credit_card, options) }
+          r.process(:ignore_result) { void(r.authorization, options) }
+        end
       end
 
       def supports_scrubbing?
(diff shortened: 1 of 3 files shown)


naashton (approved):
LGTM!

lib/active_merchant/billing/gateways/fortis.rb
```diff
@@ -175,11 +182,15 @@ def path(action, value = '')
...
           store: '/tokens/cc',
           unstore: '/tokens/{placeholder}'
-        }[action]&.gsub('{placeholder}', value.to_s)
+        }[action]&.gsub('{placeholder}', value.to_s.split('|').first || '')
       end
 
       def payment_type(payment)
-        payment.is_a?(String) ? 'token' : 'keyed'
+        if payment.is_a?(String)
+          payment.split('|').last == 'txn' ? 'prev-trxn' : 'token'
```
javierpedrozaing, line 190: it took me a little bit to understand why the split is needed, but I can see that it is required for case like this. `"transaction/31e59fb2ec9d4bc5fb|txn/refund"` 👍🏼

test/remote/gateways/remote_fortis_test.rb
```diff
@@ -155,6 +179,14 @@ def test_storing_credit_card
     assert_success store
   end
 
+  def test_storing_credit_card_with_location_as_option
```
javierpedrozaing, line 182: is there another case where the `location_id` is required ? or maybe would be better to remove it from the fixtures ?

Heavyblade: You can have location_id, in your fixtures/gateway optionally, but if you don't want to, then fo the store operation yopu need to send it as aGSF

PR #5368, opened 2024-12-12
Remove  from list of required fields when instantiating gateway class

naashton:
Remove `developer_id` from list of required fields when creating gateway.


lib/active_merchant/billing/gateways/fortis.rb
```diff
@@ -110,7 +110,7 @@ class FortisGateway < Gateway
       }
 
       def initialize(options = {})
-        requires!(options, :user_id, :user_api_key, :developer_id)
+        requires!(options, :user_id, :user_api_key)
```
javierpedrozaing, line 113: Since `developer_id` is no longer a required field, should we also remove it from the fixtures? ❓

javierpedrozaing (commented):
LGTM, just  I left a small question

lib/active_merchant/billing/gateways/fortis.rb
naashton: It's still "required" just not for the merchant. For now, I think we can leave it because it's loaded on our side via an env var.

javierpedrozaing (approved):
LGTM

naashton:
No longer needed based on internal discussions

PR #5371, opened 2024-12-16
Normalize API versions

Buitragox:
Summary:
This change centralizes version management at the class level, creating a consistent and reusable way to define and access API versions.

Unit Tests:
6154 tests, 80974 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

RuboCop:

### DIFF
--- a/lib/active_merchant/versionable.rb
+++ b/lib/active_merchant/versionable.rb
@@ -1,0 +1,29 @@
+module ActiveMerchant
+  module Versionable
+    def self.included(base)
+      if base.respond_to?(:class_attribute)
+        base.class_attribute :versions, default: {}
+        base.extend(ClassMethods)
+      end
+    end
+
+    module ClassMethods
+      def inherited(subclass)
+        super
+        subclass.versions = {}
+      end
+
+      def version(version, feature = :default_api)
+        versions[feature] = version
+      end
+
+      def fetch_version(feature = :default_api)
+        versions[feature]
+      end
+    end
+
+    def fetch_version(feature = :default_api)
+      versions[feature]
+    end
+  end
+end
--- a/test/unit/versionable_test.rb
+++ b/test/unit/versionable_test.rb
@@ -1,0 +1,89 @@
+require 'test_helper'
+
+class VersionableTest < Test::Unit::TestCase
+  class ParentClass
+    include ActiveMerchant::Versionable
+  end
+
+  class DummyClass < ParentClass
+  end
+
+  class FakeClass < ParentClass
+  end
+
+  class DummyChildClass < DummyClass
+  end
+
+  def setup
+    @dummy_instance = DummyClass.new
+    @fake_instance = FakeClass.new
+    @dummy_child_instance = DummyChildClass.new
+  end
+
+  def test_class_can_set_and_fetch_default_version
+    DummyClass.version('1.0')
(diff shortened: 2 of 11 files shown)


test/unit/versionable_test.rb
```diff
@@ -0,0 +1,89 @@
...
+    DummyClass.version('v4', :some_feature)
+
+    assert_equal '1.0', @dummy_instance.fetch_version, 'Class should return the correct version'
+    assert_equal 'v4', @dummy_instance.fetch_version(:some_feature), 'Class should return the correct version'
+    assert_equal '2.0', @fake_instance.fetch_version, 'Class should return the correct version'
+    assert_equal 'v3', @fake_instance.fetch_version(:some_feature), 'Class should return the correct version'
+    assert_equal '3.0', @dummy_child_instance.fetch_version, 'Class should return the correct version'
+    assert_equal 'v5', @dummy_child_instance.fetch_version(:some_feature), 'Class should return the correct version'
+  end
+end
```
Buitragox, line 89: New tests to verify classes and instances don't share the same `versions` hash.

lib/active_merchant/versionable.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+    end
+
+    module ClassMethods
+      def inherited(subclass)
+        super
+        subclass.versions = {}
+      end
+
+      def version(version, feature = :default_api)
```
almalee24, line 16: How is this method used? Is it just for the `versionable_test.rb`?

lib/active_merchant/versionable.rb
```diff
@@ -0,0 +1,29 @@
...
+        versions[feature] = version
+      end
+
+      def fetch_version(feature = :default_api)
```
almalee24, line 20: Also this one, how is it used? What is the difference between this one and the one in line 25?

lib/active_merchant/versionable.rb
Buitragox: This method is inside the ClassMethods module. This module is extended to the base class in line 6, so any methods inside this module will be class methods.

The intended usage for this method is to be used at the top of any Gateway subclass.

```ruby
class MyTestGateway < Gateway
  version 'v3' # This sets the :default_api key in the versions hash.
  version 'v5', :recurring_api # This sets the :recurring_api key in the versions hash
end
```

You can see this method being used in this PR in `adyen.rb`, `braintree_blue.rb` and `global_collect.rb`.

lib/active_merchant/versionable.rb
Buitragox: The `fetch_version` method inside ClassMethods is intended to be used outside of instance methods.

For example, some gateways create `live_url` and `test_url` at the top of the file without creating an instance, so you can do something like this:

```ruby
module ActiveMerchant # :nodoc:
  module Billing # :nodoc:
    class ForteGateway < Gateway
      version 'v2'

      self.test_url = "https://sandbox.forte.net/api/#{fetch_version}"
      self.live_url = "https://api.forte.net/#{fetch_version}"

      # other methods ...
    end
  end
end
```

The `fetch_version` method outside ClassMethods is intended to be used inside of instance methods. There are a couple of examples of this in this PR, you can see it being used for the Adyen gateway and GlobalCollect gateway.

almalee24 (approved):
Looks good to me!

PR #5375, opened 2024-12-30
Orbital: Add XSD version testing to unit test

almalee24:
Add unit test to verify that the API verison used for testing is the same one used in the gateway.


DustinHaefele (approved):
Looks pretty straight forward.

PR #5384, opened 2025-01-20
CardTypes: Adding card type to gateways

javierpedrozaing:
Description
-------------------------
This commit is a script to add card types to gateways

The script update the gateway classes adding the new card type in `supported_cardtypes` variable, additionally update the gateway test file to validate card types supported

Update all gateways:
`/script/generate card_type new_card_type`

Update specific gateways:
`./script/generate card_type new_card_type --gateways stripe authorize_net paypal`

Update a single gateway:
`./script/generate card_type new_card_type --gateways stripe`

### DIFF
--- a/generators/gateway/card_type_generator.rb
+++ b/generators/gateway/card_type_generator.rb
@@ -1,0 +1,82 @@
+require 'thor/group'
+
+class CardTypeGenerator < Thor::Group
+  include Thor::Actions
+
+  argument :card_type, type: :string, desc: "The name of the card type to add"
+  class_option :gateways, type: :array, desc: "List of gateway names to update (optional)", default: nil
+
+  def self.source_root
+    File.dirname(__FILE__)
+  end
+
+  def add_card_type_to_gateways
+    gateway_dir = "lib/active_merchant/billing/gateways"
+    test_dir = "test/unit/gateways"
+
+    gateways_to_update = if options[:gateways]
+      options[:gateways].map { |g| "#{gateway_dir}/#{g}.rb" }
+    else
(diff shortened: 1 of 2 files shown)


generators/gateway/card_type_generator.rb
```diff
@@ -0,0 +1,82 @@
...
+    # $3 = "]"
+    updated_content = content.gsub(/(self\.supported_cardtypes\s*=\s*%i\[)(.*?)(\])/m) do
+      current_types = $2.strip
+      return if current_types.include?(new_card_type)
+      "#{$1}#{current_types.empty? ? new_card_type : "#{current_types} #{new_card_type}"}#{$3}"
+    end
+    File.write(path, updated_content)
+  end
+
+  def update_test_file(path, new_card_type)
```
Heavyblade, line 58: 💬 Comment

This is the world we live in now, a proper prompt will compete with our best Regex:

prompt => 
`I want add a new type of card called "dava" to the list of supported types for the 3 gateways listed, I also want to update the corresponding unit tests adding "dava" to the list of cards that the test checks`

Screenshot 2025-01-20 at 4 34 16 PM

Heavyblade: Screenshot 2025-01-20 at 4 37 09 PM

Heavyblade: Screenshot 2025-01-20 at 4 37 42 PM

javierpedrozaing: Yes, it would be great to figure out how we can automate this kind of update using scripts or prompts like this

PR #5385, opened 2025-01-20
Checkout v2: add l2/l3

gasb150:
Summary
----------------
This adds the L2/L3 fields required in purchase, authorize and capture transactions

[SER-1554](https://spreedly.atlassian.net/browse/SER-1554)

Unit Tests
----------------
Finished in 0.282324 seconds.
70 tests, 481 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests
----------------
Finished in 164.547288 seconds.
116 tests, 280 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 93.9655% passed

Rubocop
----------------


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -505,6 +507,69 @@ def add_marketplace_data(post, options)
...
+          }.compact
+        )
+      end
+
+      def add_shipping_data(post, options)
+        post[:shipping] ||= {}
+        post[:shipping][:from_address_zip] = options[:from_address_zip]
+      end
+
+      def build_items(line_items)
```
Heavyblade, line 553: 💬 Comment:

Just an idea => if you set line_items with a default value `[]` you will not need the really return

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -505,6 +507,69 @@ def add_marketplace_data(post, options)
...
+        return unless line_items
+
+        line_items.map do |item|
+          {
+            # for lvl 2 amex and lvl3 visa/master
+            name: item[:name],
+             quantity: item[:quantity],
```
javierpedrozaing, line 560: Looks like there is a indentation problem here.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -505,6 +507,69 @@ def add_marketplace_data(post, options)
...
+        post[:shipping] ||= {}
+        post[:shipping][:from_address_zip] = options[:from_address_zip]
+      end
+
+      def build_items(line_items)
+        return unless line_items
+
+        line_items.map do |item|
+          {
+            # for lvl 2 amex and lvl3 visa/master
```
Heavyblade, line 558: 💬 Comment:

two things here to comment:

1) the method has comments that let me know that seems to be the case that there are some params that are to visa/master for lvl3 and some for lvl2 amex, but there is not actually any logic that represents that, no matter the card brand all the fields on this hash are going to be included.

2) seems that perhaps there is a misaligned value from the quantity line to the bottom.

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -1186,4 +1191,51 @@ def test_non_truncate_id_for_non_amex_transactions
...
+      currency: 'USD',
+      line_items: [{ item_name: 'Paint', quantity: '1', unit_cost: '1270' }]
+    }
+    @amex_card = credit_card('345678901234564', brand: 'american_express', verification_value: '1000', month: '12', year: Time.now.year)
+
+    response = @gateway.purchase(1500, @amex_card, @minimun_lvl2_options.merge(amex_options))
+    assert_success response
+  end
+
+  def test_successful_purchase_with_minimun_lvl_3_data
```
javierpedrozaing, line 1211: if is possible, would be good idea add a capture request with l2/l3 data as well.

test/unit/gateways/checkout_v2_test.rb
```diff
@@ -1115,6 +1155,89 @@ def test_authorize_supports_alternate_credit_card_implementation
     end.respond_with(successful_authorize_response)
   end
 
+  def test_authorize_with_lvl_2_3_data
```
javierpedrozaing, line 1158: For small words like level personally I don't like using abbreviation and also it would be good idea to follow the level2 and level3 standards in this kind if test.

javierpedrozaing (commented):
Great Job, I left some comments for you consideration.

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -505,6 +507,69 @@ def add_marketplace_data(post, options)
...
+        items = build_items(options[:line_items])
+        post[:items] = items unless items.nil?
+      end
+
+      def add_level_two_data(post, options)
+        post[:customer][:tax_number] = options[:tax_number] # field no require for amex
+
+        post[:processing].merge!(
+          {
+            order_id: options.dig(:processing, :order_id),
```
Heavyblade, line 532: ❓ Question:

Is expected that lvl2 data to haave nested values ?

in this case you are asking for options[:processing][:order_id]

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -1186,4 +1191,51 @@ def test_non_truncate_id_for_non_amex_transactions
...
+    response = @gateway.purchase(@amount, @credit_card, @minimun_lvl2_options)
+    assert_success response
+  end
+
+  def test_successful_purchase_with_minimun_lvl_2_data_for_amex
+    amex_options = {
+      currency: 'USD',
+      line_items: [{ item_name: 'Paint', quantity: '1', unit_cost: '1270' }]
+    }
+    @amex_card = credit_card('345678901234564', brand: 'american_express', verification_value: '1000', month: '12', year: Time.now.year)
```
Heavyblade, line 1205: 💬 Comment

@amex_card doesn't have to an instance variable it could just be a local variable.

test/remote/gateways/remote_checkout_v2_test.rb
```diff
@@ -1186,4 +1191,51 @@ def test_non_truncate_id_for_non_amex_transactions
...
+
+    response = @gateway.purchase(1500, @amex_card, @minimun_lvl2_options.merge(amex_options))
+    assert_success response
+  end
+
+  def test_successful_purchase_with_minimun_lvl_3_data
+    order_data = {
+      processing: { order_id: '01234' },
```
Heavyblade, line 1213: ❓ Question:
This one relates to one above => how is order_data[:processing][:order_id] going to be sent in this structure if is a GSF, unless the entire lvl3 data fits in one big GSF

Heavyblade (commented):
is pretty much ok @gasb150, left some minor comments and question

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: 1. All of them works for Visa and Mastercard, but only the 3 first are exclusively for amex. This is something also we need to put in documentation, it's risky to made validations for this in the code if for some reason this change in the future we will maintaining this. Also this seams to be something the client should be aware.

If we add any brand validation also we will need to pass the brand in the authorization result for capture cases, according the documentation the level 2 and level 3 values need to be passed in capture.

lib/active_merchant/billing/gateways/checkout_v2.rb
gasb150: In this case is to avoid confusing with the order_id in the transaction, I can change it to a simple value called invoice_id and could work too.

test/remote/gateways/remote_checkout_v2_test.rb
gasb150: In this case only order_id will be passed in the structure, due confusing with the order_id in the transaction, but it can be passed in a different and not nested value like invoice_id.

Is not needed all the values to ActiveMerchant pass in the same structure. But maybe for clarity could be better modify the value name.

naashton (approved):
Nice work!

PR #5386, opened 2025-01-21
Add alternate spelling for Vietnam

jcreiff:
CER-2034

6169 tests, 81065 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -117,6 +117,7 @@
 * Ebanx: Add the notification_url field [yunnydang] #5388
 * Ebanx: Add optional payment_type_code override [jcreiff] #5370
 * Ebanx: Update list of supported countries [jcreiff] #5387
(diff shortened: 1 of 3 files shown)


jcreiff:
Just a note about this change - there is at least one precedent for duplicate country entries in the `COUNTRIES` array, to support more than one country code for Romania. Because of the way that `detect` works, this change should be purely additive: the correct country codes will be returned whether `Viet Nam` or `Vietnam` are supplied to `Country.find`, and if an alpha or numeric code is supplied instead, the original entry (`Viet Nam`) will still be returned, since it comes first in the array.

yunnydang (approved):
looks good! Not sure why the github actions failed but will approve this so its non blocking!

PR #5390, opened 2025-01-22
Add AFT fields for credorax

yunnydang:
Adds the s( sender) field for aft transactions while also re-using the j (recipient)fields for purchase and authorize.

Local:
6168 tests, 81069 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
84 tests, 417 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
55 tests, 195 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
89.0909% passed

The failing tests are 3ds related. I dont believe its due to my changes.

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -151,6 +151,11 @@ module ActiveMerchant # :nodoc:
         add_stored_credential(post, options)
         add_processor(post, options)
 
+        if options[:aft]
+          add_recipient(post, options)
+          add_sender(post, options)
+        end
+
         commit(:purchase, post)
       end
 
@@ -168,6 +173,11 @@ module ActiveMerchant # :nodoc:
         add_account_name_inquiry(post, options)
         add_processor(post, options)
         add_authorization_details(post, options)
+
+        if options[:aft]
+          add_recipient(post, options)
+          add_sender(post, options)
+        end
 
         commit(:authorize, post)
       end
@@ -341,14 +351,34 @@ module ActiveMerchant # :nodoc:
         post[:c3] = options[:email] || 'unspecified@example.com'
       end
 
+      def add_sender(post, options)
+        return unless options[:sender_ref_number] || options[:sender_fund_source] || options[:sender_country_code] || options[:sender_address] || options[:sender_street_address] || options[:sender_city] || options[:sender_state] || options[:sender_first_name] || options[:sender_last_name]
+
+        sender_country_code = options[:sender_country_code]&.length == 3 ? options[:sender_country_code] : Country.find(options[:sender_country_code]).code(:alpha3).value if options[:sender_country_code]
+        post[:s15] = sender_country_code
+        post[:s17] = options[:sender_ref_number] if options[:sender_ref_number]
+        post[:s18] = options[:sender_fund_source] if options[:sender_fund_source]
+        post[:s10] = options[:sender_first_name] if options[:sender_first_name]
+        post[:s11] = options[:sender_last_name] if options[:sender_last_name]
+        post[:s12] = options[:sender_street_address] if options[:sender_street_address]
+        post[:s13] = options[:sender_city] if options[:sender_city]
+        post[:s14] = options[:sender_state] if options[:sender_state]
+      end
+
       def add_recipient(post, options)
-        return unless options[:recipient_street_address] || options[:recipient_city] || options[:recipient_province_code] || options[:recipient_country_code]
+        return unless options[:recipient_street_address] || options[:recipient_city] || options[:recipient_province_code] || options[:recipient_country_code] || options[:recipient_first_name] || options[:recipient_last_name] || options[:recipient_postal_code]
 
         recipient_country_code = options[:recipient_country_code]&.length == 3 ? options[:recipient_country_code] : Country.find(options[:recipient_country_code]).code(:alpha3).value if options[:recipient_country_code]
         post[:j6] = options[:recipient_street_address] if options[:recipient_street_address]
         post[:j7] = options[:recipient_city] if options[:recipient_city]
         post[:j8] = options[:recipient_province_code] if options[:recipient_province_code]
+        post[:j12] = options[:recipient_postal_code] if options[:recipient_postal_code]
         post[:j9] = recipient_country_code
+
+        if options[:aft]
+          post[:j5] = options[:recipient_first_name] if options[:recipient_first_name]
+          post[:j13] = options[:recipient_last_name] if options[:recipient_last_name]
+        end
       end
 
       def add_customer_name(post, options)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -121,6 +121,7 @@
 * Checkout v2: add l2/l3 [gasb150] #5385
 * Worldpay: Update passing 3DS data for NT [almalee24] #5389
 * Ebanx: Add the merchant_payment_code override [yunnydang] #5394
+* Credorax: Add AFT fields [yunnydang] #5390
 
 == Version 1.137.0 (August 2, 2024)
 * Unlock dependency on `rexml` to allow fixing a CVE (#5181).
--- a/test/remote/gateways/remote_credorax_test.rb
+++ b/test/remote/gateways/remote_credorax_test.rb
@@ -154,6 +154,32 @@ class RemoteCredoraxTest < Test::Unit::TestCase
     assert_equal 'Succeeded', response.message
   end
 
+  def test_successful_purchase_with_aft_fields
+    aft_options = @options.merge(
+      aft: true,
+      sender_ref_number: 'test',
+      sender_fund_source: '01',
+      sender_country_code: 'USA',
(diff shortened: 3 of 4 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -341,6 +352,23 @@ def add_email(post, options)
         post[:c3] = options[:email] || 'unspecified@example.com'
       end
 
+      def add_sender(post, payment_method, options)
+        return unless options[:sender_ref_number] || options[:sender_fund_source] || options[:sender_country_code] || options[:billing_address]
+
+        sender_country_code = options[:sender_country_code]&.length == 3 ? options[:sender_country_code] : Country.find(options[:sender_country_code]).code(:alpha3).value if options[:sender_country_code]
```
jcreiff, line 358: I think the ternary statement and the `if` clause all smashed into one line makes this logic a little hard to parse

Could you instead do something like 

```
if options[:sender_country_code]
  post[:s15] = Country.find(options[:sender_country_code]).code(:alpha3).value
end
```

Do we need to do the check for `.length == 3` or can we just send the code in to `Country.find` without that?

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -341,6 +352,23 @@ def add_email(post, options)
         post[:c3] = options[:email] || 'unspecified@example.com'
       end
 
+      def add_sender(post, payment_method, options)
+        return unless options[:sender_ref_number] || options[:sender_fund_source] || options[:sender_country_code] || options[:billing_address]
```
jcreiff, line 356: Based on our earlier discussion, do we want these `[:billing_address]` fields to come from the top-level `options[:billing_address]`, or will they be pulled from `sender_` specific fields, similar to how the `add_recipient` logic works?

yunnydang: hmm now that i think about it, it would make sense to have it pulled from specific sender address fields like the recipient fields ill add later!

lib/active_merchant/billing/gateways/credorax.rb
yunnydang: I didnt give too much thought into this as it was mocked similarly to the recipient method for the country code

jcreiff: Ohhhh I see that now 🧐

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -341,6 +353,21 @@ def add_email(post, options)
...
 
+      def add_sender(post, payment_method, options)
+        return unless options[:sender_ref_number] || options[:sender_fund_source] || options[:sender_country_code] || options[:sender_address] || options[:sender_street_address] || options[:sender_city] || options[:sender_state]
+
+        sender_country_code = options[:sender_country_code]&.length == 3 ? options[:sender_country_code] : Country.find(options[:sender_country_code]).code(:alpha3).value if options[:sender_country_code]
+        post[:s15] = sender_country_code
+        post[:s16] = payment_method.number
+        post[:s17] = options[:sender_ref_number] if options[:sender_ref_number]
+        post[:s18] = options[:sender_fund_source] if options[:sender_fund_source]
+        post[:s10] = payment_method.first_name
```
jcreiff, line 364: Is there any chance we have a `payment_method` here that doesn't have a `first_name` attribute?

And if we are requiring explicit `sender_*` fields for the other address data, why not also for the sender name?

yunnydang: hmm thats a good point! Ill open up a new name field

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -151,6 +151,12 @@ def purchase(amount, payment_method, options = {})
         add_stored_credential(post, options)
         add_processor(post, options)
 
+        if options[:aft]
+          add_customer_name(post, options)
```
jcreiff, line 155: Sorry to keep dropping one-off questions instead of giving one comprehensive review, but additional things keep occurring to me each time I think about this 😓 

The `j5` and `j13` fields that are added via `add_customer_name` are looking for `options[:first_name]` and `options[:last_name]`, which on our side we are pulling from the payment method in use, but only when it's a GeneralCredit transaction or if the `account_name_inquiry` flag is passed

I think in this case, because the name here is related to the recipient, we can't pull that from the card and we may need to have an option to specify `recipient_first_name` and `recipient_last_name` as separate fields. Does that seem correct?

I'm not sure if the clearest way to do that is to modify the existing `add_customer_name` to allow for the new fields or to be a little less `DRY` and create a separate `add_recipient_name` submethod 🤔

rachelkirk: the `account_name_inquiry` was only added for authorize transaction types just FYI

rachelkirk: > create a separate add_recipient_name submethod

Can you just add the name fields to the existing add_recipient method?

yunnydang: I was thinking of doing this rachel since it'd be the most (in my mind) distinction between referal_cft path and the aft path

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -341,6 +351,21 @@ def add_email(post, options)
...
+        post[:s18] = options[:sender_fund_source] if options[:sender_fund_source]
+        post[:s10] = options[:sender_first_name] if options[:sender_first_name]
+        post[:s11] = options[:sender_last_name] if options[:sender_last_name]
+        post[:s12] = options[:sender_street_address] if options[:sender_street_address]
+        post[:s13] = options[:sender_city] if options[:sender_city]
+        post[:s14] = options[:sender_state] if options[:sender_state]
+      end
+
       def add_recipient(post, options)
         return unless options[:recipient_street_address] || options[:recipient_city] || options[:recipient_province_code] || options[:recipient_country_code]
```
jcreiff, line 370: Should we add the names fields to the guard clause? I see that they are always mandatory, so just in case someone only sends the names and none of the address data, I think I would prefer that we send those details along and let the gateway complain about missing address fields in the response, rather than us failing to map any of the `j` fields to the request at all

yunnydang: good point!

yunnydang: wait @jcreiff are you referring to the sender name or the recipient names?

yunnydang: if the recipient name, would we want to return out of the method though if the refund is also using it and doesnt require the recipient_first_name and recipient_last_name?

jcreiff: Recipient names, for sure. I was thinking just to add them to the string of `||` so that if _only_ the name fields were sent, we would still send them; not that we should bail out of the method if they're not present

yunnydang: ahhh thats what you were referring to, got it thanks!

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -341,14 +351,34 @@ def add_email(post, options)
         post[:c3] = options[:email] || 'unspecified@example.com'
       end
 
+      def add_sender(post, payment_method, options)
+        return unless options[:sender_ref_number] || options[:sender_fund_source] || options[:sender_country_code] || options[:sender_address] || options[:sender_street_address] || options[:sender_city] || options[:sender_state] || options[:sender_first_name] || options[:sender_last_name]
+
+        sender_country_code = options[:sender_country_code]&.length == 3 ? options[:sender_country_code] : Country.find(options[:sender_country_code]).code(:alpha3).value if options[:sender_country_code]
+        post[:s15] = sender_country_code
+        post[:s16] = payment_method.number
```
jcreiff, line 359: OK one more thing (the last, I hope 😓 )

It occurred to me that with this change we would also need to update the `scrub` method to make sure we filter out the PAN here.

Then it also occurred to me to double-check the docs to make sure this was actually a required field to send, and as of the December 2024 updated reference from Shift4/Credorax, it is `optional`

Here's the [latest version](https://epower.credorax.com/wp-content/uploads/2025/01/Shift4-AFT-Guide-version-1.6b.pdf)

Since it is optional, I think we should omit it - that's the simplest thing to do

jcreiff (changes requested):

lib/active_merchant/billing/gateways/credorax.rb
yunnydang: thanks for this call out! I see the excerpt: "The recipient's full name and address parameters are mandatory for AFTs, and they should be sent
on all AFT transactions. "

but now im wondering if recipient postal code (j12) will also need to be added as well?

jcreiff (approved):

PR #5391, opened 2025-01-23
Adyen: Update shopperInteraction

almalee24:
shopperInteraction should be Ecommerce if
shopper_interaction is in options and
initiator is cardholder.

Remote:
149 tests, 474 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 91.9463% passed


sinourain (approved):
LGTM!

PR #5394, opened 2025-01-29
Ebanx: Add the merchant_payment_code override

yunnydang:
Adds the optional merchant_payment_code override

Local:
6178 tests, 81154 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
33 tests, 163 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
45 tests, 102 assertions, 7 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
84.4444% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -213,7 +213,7 @@ module ActiveMerchant # :nodoc:
       def add_invoice(post, money, options)
         post[:payment][:amount_total] = amount(money)
         post[:payment][:currency_code] = (options[:currency] || currency(money))
-        post[:payment][:merchant_payment_code] = Digest::MD5.hexdigest(options[:order_id])
+        post[:payment][:merchant_payment_code] = Digest::MD5.hexdigest(order_id_override(options))
         post[:payment][:instalments] = options[:instalments] || 1
         post[:payment][:order_number] = options[:order_id][0..39] if options[:order_id]
       end
@@ -249,11 +249,16 @@ module ActiveMerchant # :nodoc:
         end
       end
 
+      # we will prefer the merchant_payment_code if both fields are provided
+      def order_id_override(options)
+        options[:merchant_payment_code] || options[:order_id]
+      end
+
       def add_additional_data(post, options)
         post[:device_id] = options[:device_id] if options[:device_id]
         post[:metadata] = options[:metadata] if options[:metadata]
         post[:metadata] = {} if post[:metadata].nil?
-        post[:metadata][:merchant_payment_code] = options[:order_id] if options[:order_id]
+        post[:metadata][:merchant_payment_code] = order_id_override(options)
         post[:payment][:tags] = TAGS
         post[:notification_url] = options[:notification_url] if options[:notification_url]
       end
--- a/test/unit/gateways/ebanx_test.rb
+++ b/test/unit/gateways/ebanx_test.rb
@@ -54,6 +54,32 @@ class EbanxTest < Test::Unit::TestCase
     assert_success response
   end
 
+  def test_successful_purchase_without_merchant_payment_code
+    # hexdigest of 1 is c4ca4238a0b923820dcc509a6f75849b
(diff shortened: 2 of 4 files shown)


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -249,11 +249,18 @@ def payment_details(payment)
         end
       end
 
+      def order_id_override(options)
+        return unless options[:merchant_payment_code] || options[:order_id]
```
jcreiff, line 253: Order ID should always be present; I don't think we need this guard clause

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -213,9 +213,9 @@ def add_address(post, options)
       def add_invoice(post, money, options)
         post[:payment][:amount_total] = amount(money)
         post[:payment][:currency_code] = (options[:currency] || currency(money))
-        post[:payment][:merchant_payment_code] = Digest::MD5.hexdigest(options[:order_id])
+        post[:payment][:merchant_payment_code] = Digest::MD5.hexdigest(order_id_override(options)) if order_id_override(options)
```
jcreiff, line 216: Since the override method should return something everytime, we shouldn't need the `if` here

test/unit/gateways/ebanx_test.rb
```diff
@@ -54,6 +54,17 @@ def test_successful_purchase_with_soft_descriptor
     assert_success response
   end
 
+  def test_successful_purchase_with_merchant_payment_code
```
jcreiff, line 57: For sanity sake can we also add a unit test that shows the order_id is still successfully mapped to merchant payment code if one is not provided in the options?

test/unit/gateways/ebanx_test.rb
```diff
@@ -54,6 +54,17 @@ def test_successful_purchase_with_soft_descriptor
...
+    response = stub_comms(@gateway, :ssl_request) do
+      @gateway.purchase(@amount, @credit_card, @options.merge(merchant_payment_code: '2'))
+    end.check_request do |_method, _endpoint, data, _headers|
+      assert_match %r{"merchant_payment_code\":\"2\"}, data
+      assert_match %r{"order_number\":\"2\"}, data
```
jcreiff, line 62: The `order_number` assertion will need to change to account for the fact that we are still mapping it to `order_id`

lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -213,9 +213,9 @@ def add_address(post, options)
...
         post[:payment][:instalments] = options[:instalments] || 1
-        post[:payment][:order_number] = options[:order_id][0..39] if options[:order_id]
+        post[:payment][:order_number] = order_id_override(options)[0..39] if order_id_override(options)
```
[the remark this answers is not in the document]
yunnydang: ah gotcha thanks for clarifying!

lib/active_merchant/billing/gateways/ebanx.rb
yunnydang: I was going back and forth on this whether we should guard it but i agree!

test/unit/gateways/ebanx_test.rb
```diff
@@ -54,6 +54,28 @@ def test_successful_purchase_with_soft_descriptor
     assert_success response
   end
 
+  def test_successful_purchase_without_merchant_payment_code
+    response = stub_comms(@gateway, :ssl_request) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end.check_request do |_method, _endpoint, data, _headers|
+      assert_match %r{"merchant_payment_code\":\"1\"}, data
```
jcreiff, line 61: I just realized that this assertion is being satisfied by the override in the metadata - can you also add an `assert_match` for both tests that shows we are sending the correct hexdigest value for the value that's nested under `payment`?

jcreiff (changes requested):

PR #5397, opened 2025-01-31
Adyen: Add support for localized shopper statement

rachelkirk:
CER-2023

This field is an object that expects data specific to Japanese branded cards and the ja-Kana character set. If localized shopper statement is not provided it will default to shopper statement in authorize transaction types.

For refund and capture, it is sent through as part of additional data and will likely not display on credit card statements for that reason.

Unit Tests
138 tests, 731 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests
152 tests, 481 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 92.1053% passed
*12 tests also failing on master

### DIFF
--- a/test/remote/gateways/remote_adyen_test.rb
+++ b/test/remote/gateways/remote_adyen_test.rb
@@ -1471,6 +1471,22 @@ class RemoteAdyenTest < Test::Unit::TestCase
     assert_success capture
   end
 
+  def test_successful_capture_with_localized_shopper_statement
+    auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_success auth
+
+    assert capture = @gateway.capture(@amount, auth.authorization, @options.merge(localized_shopper_statement: { 'ja-Kana' => 'ADYEN - セラーA' }))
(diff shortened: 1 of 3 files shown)


test/remote/gateways/remote_adyen_test.rb
```diff
@@ -158,7 +158,7 @@ def setup
       shopper_reference: 'John Smith',
       billing_address: @us_address,
       order_id: '123',
-      stored_credential: { reason_type: 'unscheduled' }
+      stored_credential: { reason_type: 'unscheduled' },
```
yunnydang, line 161: not sure why this comma was added but maybe copy pasta?

rachelkirk: Already fixed. I had added the field to the main options originally to see if it would trigger failures and it didn't but still opted to keep it out of main options hash

yunnydang (approved):
looks good once the checks pass!

PR #5398, opened 2025-02-03
SecurepayAU: New test for stored card

dacook:
Contributing a new test that we created when troubleshooting a certain case.

test/remote/gateways/remote_secure_pay_au_test.rb
19 tests, 69 assertions, 0 failures, 0 errors, 0 pendings, 2 omissions, 0 notifications 100% passed


dacook:
I did also do a refactor of this file at the time, but now I'm not sure it's worth it, so left it out.
https://github.com/activemerchant/active_merchant/commit/1ec5cd542a1d13c0a6d7d308a266859984e3b516
 Because it's the API call it's probably worth specifying the exact parameters each time.

dacook:
This remains applicable to the current build, but unfortunately nobody seems interested.

PR #5400, opened 2025-02-07
Update authorization_from method so that refunds will have the payment_intent_id

yunnydang:
Local:
6189 tests, 81209 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
36 tests, 196 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
29 tests, 67 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
93.1034% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -124,6 +124,7 @@
 * Credorax: Add AFT fields [yunnydang] #5390
 * Cashnet: Update max_retries to 1 [almalee24] #5393
 * Cybersource: Update 3DS fields [almalee24] #5396
+* Airwallex: Update authorization_from method to include payment intent ID on refunds [yunnydang] #5400
 
 == Version 1.137.0 (August 2, 2024)
 * Unlock dependency on `rexml` to allow fixing a CVE (#5181).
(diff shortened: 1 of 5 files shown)


jcreiff (approved):
This looks good! One non-blocking comment - can you add an assertion to the `test_successful_refund` unit test that asserts the `response.authorization` is equal to the `payment_intent_id` value found in the mocked `successful_refund_response`?

yunnydang:
> This looks good! One non-blocking comment - can you add an assertion to the `test_successful_refund` unit test that asserts the `response.authorization` is equal to the `payment_intent_id` value found in the mocked `successful_refund_response`?

Hey joe i pushed up an extra bit for the credorax gateway as i realize it was left over artifact prior to the finalized version. The  options[:sender_address] is not a valid field since the options[:sender_street_address] is the correct name

jcreiff (approved):

jcreiff:
Thanks for the heads up! I don't have a problem with tacking that on here

PR #5402, opened 2025-02-07
Worldpay: Add AFT support to Authorize and Purchase transaction types

rachelkirk:
Adds AFT fields to auth and purchase.

CER-2069

Unit Tests
140 tests, 771 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests
122 tests, 490 assertions, 11 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 90.9836% passed
*11 tests also failing on master

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -63,7 +63,11 @@ module ActiveMerchant # :nodoc:
       def authorize(money, payment_method, options = {})
         requires!(options, :order_id)
         payment_details = payment_details(payment_method, options)
-        authorize_request(money, payment_method, payment_details.merge(options))
+        if options[:account_funding_transaction]
+          aft_request(money, payment_method, payment_details.merge(**options))
+        else
+          authorize_request(money, payment_method, payment_details.merge(options))
+        end
       end
 
       def capture(money, authorization, options = {})
--- a/test/remote/gateways/remote_worldpay_test.rb
+++ b/test/remote/gateways/remote_worldpay_test.rb
@@ -1056,6 +1056,38 @@ class RemoteWorldpayTest < Test::Unit::TestCase
     assert_equal '20', credit.error_code
   end
 
+  def test_successful_authorize_visa_account_funding_transfer
+    auth = @gateway.authorize(@amount, @credit_card, @options.merge(@aft_options))
+    assert_success auth
+    assert_equal 'funding_transfer_transaction', auth.params['action']
+    assert_equal 'SUCCESS', auth.message
+  end
+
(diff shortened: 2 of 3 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -63,7 +63,11 @@ def purchase(money, payment_method, options = {})
       def authorize(money, payment_method, options = {})
         requires!(options, :order_id)
         payment_details = payment_details(payment_method, options)
-        authorize_request(money, payment_method, payment_details.merge(options))
+        if options[:account_funding_transaction]
+          aft_request(money, payment_method, payment_details.merge(**options))
+        else
+          authorize_request(money, payment_method, payment_details.merge(options))
+        end
       end
```
rachelkirk, line 71: As discussed in the ticket, this change was additive only. I left the AFT logic in the `credit` method until we determine it should be removed/updated according to Worldpay's docs

test/remote/gateways/remote_worldpay_test.rb
jcreiff, line 1: Question! Is there anything that comes back in the response from the gateway that indicates "Yep this was AFT"? I can't remember seeing either way in the docs but will scan them again. If so it might be good to add an assertion that checks for it.

rachelkirk: There is only one element in the response that seems non standard `fundingLinkId`. It only returns a value if a mastercard is used, otherwise it's just an empty element. The cards used for testing when this was previously added were all visa, so I'm assuming that's why no assertion was made but I can add one.

rachelkirk: Had to go another route, added some assertions based on the action.

test/remote/gateways/remote_worldpay_test.rb
```diff
@@ -1056,6 +1056,39 @@ def test_failed_visa_account_funding_transfer_acquirer_error
     assert_equal '20', credit.error_code
   end
 
+  def test_successful_authorize_visa_account_funding_transfer
+    auth = @gateway.authorize(@amount, @credit_card, @options.merge(@aft_options))
+    assert_success auth
+    # require 'pry'; binding.pry
```
jcreiff, line 1062: 🪓

jcreiff (approved):
Just one note about the `pry` comment but that can be removed with the commit to update the CHANGELOG 👍🏻

test/remote/gateways/remote_worldpay_test.rb
rachelkirk: @jcreiff removed pry statement, will add changelog when merging

PR #5411, opened 2025-02-20
Worldpay: update authorize request for AFT

rachelkirk:
CER-2099

Update authorize request to include AFT fields when 3DS is invoked.

### DIFF
--- a/lib/active_merchant/billing/gateways/worldpay.rb
+++ b/lib/active_merchant/billing/gateways/worldpay.rb
@@ -63,11 +63,7 @@ module ActiveMerchant # :nodoc:
       def authorize(money, payment_method, options = {})
         requires!(options, :order_id)
         payment_details = payment_details(payment_method, options)
-        if options[:account_funding_transaction]
-          aft_request(money, payment_method, payment_details.merge(**options))
-        else
-          authorize_request(money, payment_method, payment_details.merge(options))
-        end
+        authorize_request(money, payment_method, payment_details.merge(options))
       end
 
       def capture(money, authorization, options = {})
@@ -169,7 +165,11 @@ module ActiveMerchant # :nodoc:
       end
 
       def authorize_request(money, payment_method, options)
-        commit('authorize', build_authorization_request(money, payment_method, options), 'AUTHORISED', 'CAPTURED', options)
+        if options[:account_funding_transaction]
+          aft_request(money, payment_method, options)
+        else
+          commit('authorize', build_authorization_request(money, payment_method, options), 'AUTHORISED', 'CAPTURED', options)
+        end
       end
 
       def capture_request(money, authorization, options)
@@ -427,6 +427,8 @@ module ActiveMerchant # :nodoc:
               add_payment_method(xml, money, payment_method, options)
               add_shopper(xml, options)
               add_sub_merchant_data(xml, options[:sub_merchant_data]) if options[:sub_merchant_data]
+              add_additional_3ds_data(xml, options) if options[:execute_threed] && options[:three_ds_version] && options[:three_ds_version] =~ /^2/
+              add_3ds_exemption(xml, options) if options[:exemption_type]
               add_aft_data(xml, payment_method, options)
             end
           end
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -169,7 +165,11 @@ def eci_value(payment_method, options)
       end
 
       def authorize_request(money, payment_method, options)
-        commit('authorize', build_authorization_request(money, payment_method, options), 'AUTHORISED', 'CAPTURED', options)
+        if options[:account_funding_transaction]
+          commit('funding_transfer_transaction', build_aft_request(money, payment_method, options), :ok, 'AUTHORISED', options)
```
jcreiff, line 169: Can we point this to `aft_request(money, payment_method, options)` instead? That way if any changes need to be made to how the `build_aft_request` method is referenced, it's not repeated in more than one place

rachelkirk: Makes sense, done!

lib/active_merchant/billing/gateways/worldpay.rb
```diff
@@ -169,7 +165,11 @@ def eci_value(payment_method, options)
       end
 
       def authorize_request(money, payment_method, options)
-        commit('authorize', build_authorization_request(money, payment_method, options), 'AUTHORISED', 'CAPTURED', options)
+        if options[:account_funding_transaction]
+          aft_request(money, payment_method, payment_details.merge(**options))
```
jcreiff, line 169: Seeing an error in the tests now - I think we need to pass in just `options` as the third argument here, since the `.merge` has already been called in the middle of the `authorize` method before referencing this separate sub-method

jcreiff (changes requested):

lib/active_merchant/billing/gateways/worldpay.rb
rachelkirk: already fixed! saw the failing tests in the GHA. TY!

jcreiff (commented):
Hey @rachelkirk I was thinking about this change some more, related to 3DS scenario that led us here, and I'm wondering if we need to add some additional logic to the `aft_request` method to fully enable this to be compatible with the 3DS flow.

I notice that the `build_aft_request` method looks like a truncated version of `build_authorization_request` - outside of the `aft_data` bits, the authorize logic has all the same sub-methods, and then some. In particular, it seems relevant that `add_additional_3ds_data` and `add_3ds_exemption` are present in the authorize step but missing from the `build_aft_request` logic.

Is it naive to assume that simply adding those two sub-methods into the AFT logic would cover our bases for this AFT/SCA scenario? Could we write a remote test to validate that these request components will play nicely together?

jcreiff:
One other thing that came to me as I was tinkering with the existing implementation in Postman: I think the [WP docs](https://docs.worldpay.com/apis/wpg/industryschemeextras/funding-transactions) undersold how many fields they will enforce the "can't be blank or empty" validation that previously forced us to adjust the handling of `middle` and `address2`. I came to this realization because a sample request I sent was barking at me for excluding `state`

I tried to vet this theory by commenting out various lines of the `aft_options` in the existing remote test and in I think we also need to apply similar handling to omit the following fields if not supplied:

- address1
- postalCode
- city
- state
(all the above are in both the sender and recipient blocks)

also:
- telephoneNumber
- the entire `birthDate` object (if someone submits telephoneNumber but no `birthDate`, we'll hit a NoMethodError in the `add_date_element` logic

rachelkirk:
> Hey @rachelkirk I was thinking about this change some more, related to 3DS scenario that led us here, and I'm wondering if we need to add some additional logic to the `aft_request` method to fully enable this to be compatible with the 3DS flow.
> 
> I notice that the `build_aft_request` method looks like a truncated version of `build_authorization_request` - outside of the `aft_data` bits, the authorize logic has all the same sub-methods, and then some. In particular, it seems relevant that `add_additional_3ds_data` and `add_3ds_exemption` are present in the authorize step but missing from the `build_aft_request` logic.
> 
> Is it naive to assume that simply adding those two sub-methods into the AFT logic would cover our bases for this AFT/SCA scenario? Could we write a remote test to validate that these request components will play nicely together?

👍 I updated the `build_aft_request` method to include those two 3ds sub-methods. Added a test and a temporary addition to a challenge test in case you'd like to run it, will remove after review.

jcreiff (approved):
From my perspective this looks good! Will defer to the 3DS experts on whether or not there are other considerations for the full E2E authentication flow

PR #5416, opened 2025-02-26
Plexo: Add transaction Sync

javierpedrozaing:
Description
-------------------------
[SER-1592](https://spreedly.atlassian.net/browse/SER-1592)

This commit adds transaction sync for Plexo gateway, 
includes a small code refactoring on the commit method 
and updates the expiration year in the remote test.

Unit test
-------------------------
26 tests, 144 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote test
-------------------------
33 tests, 65 assertions, 3 failures, 0 errors, 0 pendings, 3 omissions, 0 notifications 90% passed

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/plexo.rb
+++ b/lib/active_merchant/billing/gateways/plexo.rb
@@ -44,6 +44,11 @@ module ActiveMerchant # :nodoc:
         commit("/#{authorization}/captures", post, options)
       end
 
+      def inquire(authorization, options = {})
+        post = { payment_id: authorization }
+        commit('status', post, options)
+      end
+
       def refund(money, authorization, options = {})
         post = {}
         post[:ReferenceId] = options[:reference_id] || generate_unique_id
@@ -264,8 +269,11 @@ module ActiveMerchant # :nodoc:
         JSON.parse(body)
       end
 
-      def build_url(action, base)
-        url = base
+      def build_url(action, options, parameters)
+        base_url = test? ? test_url : live_url
+        return "#{base_url}/#{parameters[:payment_id]}" if action == 'status' && parameters[:payment_id]
+
+        url = base_url
         url += action if APPENDED_URLS.any? { |key| action.include?(key) }
         url
       end
@@ -285,10 +293,12 @@ module ActiveMerchant # :nodoc:
       end
 
       def commit(action, parameters, options = {})
-        base_url = (test? ? test_url : live_url)
-        url = build_url(action, base_url)
-        response = parse(ssl_post(url, parameters.to_json, header(options)))
-        response = reorder_amount_fields(response) if AMOUNT_IN_RESPONSE.include?(action)
+        url = build_url(action, options, parameters)
+
+        raw = make_request(action, url, parameters, options)
+        response = parse(raw)
+
+        reorder_amount_fields(response) if AMOUNT_IN_RESPONSE.include?(action)
 
         Response.new(
           success_from(response),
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/plexo.rb
```diff
@@ -264,7 +269,9 @@ def parse(body)
         JSON.parse(body)
       end
 
-      def build_url(action, base)
+      def build_url(action, base, options, parameters)
+        return "#{base}/#{parameters[:payment_id]}" if action == 'status' && parameters[:payment_id]
```
Heavyblade, line 273: 💬 Comment:

Maybe a more "holistic" approach could be taken here to add clarity to the `build_url` method and maybe bring the entire url logic here by moving line 295 to here.

lib/active_merchant/billing/gateways/plexo.rb
```diff
@@ -285,10 +292,13 @@ def reorder_amount_fields(response)
...
 
       def commit(action, parameters, options = {})
-        base_url = (test? ? test_url : live_url)
-        url = build_url(action, base_url)
-        response = parse(ssl_post(url, parameters.to_json, header(options)))
-        response = reorder_amount_fields(response) if AMOUNT_IN_RESPONSE.include?(action)
+        base_url = test? ? test_url : live_url
+        url = build_url(action, base_url, options, parameters)
+
+        raw = make_request(action, url, parameters, options)
```
Heavyblade, line 298: 💬 Comment:

Maybe we could be more conservative here and follow a pattern (that's actually good) by changing `ssl_post` to `ssl_resquest` and setting another optional parameter on `commit` and setting 'post' as the default value, then in the inquiry that is the only one that deviates from the standard post send the optional parameter to specify a `get`

```ruby
      def commit(action, parameters, options = {}, method = :post)
        base_url = test? ? test_url : live_url
        url = build_url(action, base_url, options, parameters)

        response = parse(ssl_request(url, method, parameters.to_json, header(options)))
       
        # more code
        # ...
```

test/remote/gateways/remote_plexo_test.rb
```diff
@@ -62,6 +62,15 @@ def test_successful_purchase_with_network_token
...
 
+  def test_successful_inquire_with_payment_id
+    response = @gateway.purchase(@amount, @credit_card, @options)
+    assert_success response
+    auth = response.authorization
+    inquire = @gateway.inquire(auth, @options)
+    assert_success inquire
+    assert_match auth, response.params['id']
+  end
+
```
Heavyblade, line 73: 💬 Comment:

Maybe a test showing the difference (or the lack of) in the transaction structure between the original purchase and the inquire could be useful, so you can know how core is going to update the references and status.

Heavyblade (commented):
Looks good @javierpedrozaing , left some comments on code correctness

lib/active_merchant/billing/gateways/plexo.rb
javierpedrozaing: yes, you are right however I am not sure what is the pattern since I can see several gateways using ssl_post, and this feature was based on the previous implementation for d_local gateway

test/remote/gateways/remote_plexo_test.rb
javierpedrozaing: Yes I will include a test for it

naashton (approved):
LGTM! Thanks for fixing up the test card exp date. Can you also add a changelog for this?

javierpedrozaing:
Hey @naashton changelog updated, thanks

PR #5417, opened 2025-02-26
Stripe PI: add the request_extended_authorization field

yunnydang:
adds the optional request_extended_authorization field for authorization requests

local:
6206 tests, 81265 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
70 tests, 367 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
101 tests, 451 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.0297% passed

A note about tests: I had tried to write a remote test for this and it fails due to account related issues. Here is the error message: "This account is not eligible for the requested card features. See https://stripe.com/docs/payments/flexible-payments for more details."

Given the error message, I do think stripe is acknowledging that this field is valid, but its just that our test account is not set up for it. I tried poking around in the dashboard but havent made progress.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -56,6 +56,7 @@ module ActiveMerchant # :nodoc:
             add_card_brand(post, options)
             add_aft_recipient_details(post, options)
             add_aft_sender_details(post, options)
+            add_request_extended_authorization(post, options)
             post[:expand] = ['charges.data.balance_transaction']
 
             CREATE_INTENT_ATTRIBUTES.each do |attribute|
@@ -381,6 +382,14 @@ module ActiveMerchant # :nodoc:
         post[:payment_method_options] ||= {}
         post[:payment_method_options][:card] ||= {}
         post[:payment_method_options][:card][:network] = options[:card_brand] if options[:card_brand]
+      end
+
+      def add_request_extended_authorization(post, options)
+        return unless options[:request_extended_authorization]
+
+        post[:payment_method_options] ||= {}
+        post[:payment_method_options][:card] ||= {}
+        post[:payment_method_options][:card][:request_extended_authorization] = options[:request_extended_authorization] if options[:request_extended_authorization]
       end
 
       def add_level_three(post, options = {})
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -134,6 +134,7 @@
 * Stripe: Update tokenization method for google pay [yunnydang] #5414
 * StripePI: Add AFT fields [yunnydang] #5418
 * CheckoutV2: Improve payout handling with blank sender data [jcreiff] #5423
+* StripePI: Add the request_extended_authorization field [yunnydang] #5417
 
 == Version 1.137.0 (August 2, 2024)
 * Unlock dependency on `rexml` to allow fixing a CVE (#5181).
--- a/test/remote/gateways/remote_stripe_payment_intents_test.rb
+++ b/test/remote/gateways/remote_stripe_payment_intents_test.rb
@@ -612,6 +612,19 @@ class RemoteStripeIntentsTest < Test::Unit::TestCase
     refute authorization.params.dig('charges', 'data')[0]['captured']
   end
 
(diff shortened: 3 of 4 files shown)


jcreiff (commented):
@yunnydang I think it might be worth reaching out to Stripe Support to see if the feature can be enabled for our test account. The response you're getting suggests that the way you've updated the request is correct, but I also notice that the docs suggest you have to `expand['latest_charge']` in order to get back the details about whether the extension is enabled, and the `capture_before` timestamp. If we plan to return those details in the response I think we'll want to see if expanding the latest charge also needs to be added to this new submethod, and if that affects anything about the way we are interpreting the response

yunnydang:
> docs suggest you have to `expand['latest_charge']`

Stripe support had confirmed its a paid IC+ account feature so we cant really turn it on for testing. As for the expanding the latest charge fiasco, they did confirm that we do need to expand it to see the response for extend_authorization. Now looking at the code for latest_charge is on of the candidates in the GSRF method:

  def extract_charge_object(params)
    return unless params

    # there's 6 potential places where we could find it.
    candidate1 = params.dig('charges', 'data').try(:[], 0)
    candidate2 = params.dig('data', 'object', 'charges', 'data').try(:[], 0)
    candidate3 = params.dig('refunds', 'data').try(:[], 0)
    candidate4 = params.dig('error', 'payment_intent', 'charges', 'data').try(:[], 0)
    candidate5 = params.dig('error', 'payment_intent', 'latest_charge')
    candidate6 = params.dig('latest_charge')

    candidate1 || candidate2 || candidate3 || candidate4 || candidate5 || candidate6
  end

So i'd be wary of running an expand on it from the get go.

yunnydang:
> @yunnydang I think it might be worth reaching out to Stripe Support to see if the feature can be enabled for our test account. The response you're getting suggests that the way you've updated the request is correct, but I also notice that the docs suggest you have to `expand['latest_charge']` in order to get back the details about whether the extension is enabled, and the `capture_before` timestamp. If we plan to return those details in the response I think we'll want to see if expanding the latest charge also needs to be added to this new submethod, and if that affects anything about the way we are interpreting the response

Also i think given that we probably would want to expand the latest_charge object eventually, i think creating a separate commit/ticket for so that it can be de-coupled would make some more sense given the risky nature of the change. What do you think?

jcreiff:
If the `expand` is confined to the `add_request_extended_authorization` method you added, then I think it should be safe to include it with this change; if remote/unit testing look fine here with that addition, we can determine with some core end-to-end testing whether there are any issues with the response parsing on our side before calling it done

yunnydang:
> If the `expand` is confined to the `add_request_extended_authorization` method you added, then I think it should be safe to include it with this change; if remote/unit testing look fine here with that addition, we can determine with some core end-to-end testing whether there are any issues with the response parsing on our side before calling it done

Hmm that would make sense! Ill see what this would affect!

yunnydang:
Heya @jcreiff so now that we're unblocked on the remote testing and seeing a response, i can confirm that the response does return the "capture_before" and "extended_authorization" status. I dont believe we will need to send and expand request for this

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -57,6 +57,7 @@ def create_intent(money, payment_method, options = {})
             add_aft_recipient_details(post, options)
             add_aft_sender_details(post, options)
             post[:expand] = ['charges.data.balance_transaction']
+            add_request_extended_authorization(post, options)
```
jcreiff, line 60: This is entirely an aesthetic/taste comment but could you move this method above the `post[:expand]` line so that all the methods are grouped in a row 😁

jcreiff (approved):
Looks good!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
yunnydang: oh right i forgot to put that back when i was testing the expand latest_charge, will do!

PR #5418, opened 2025-02-28
Stripe PI: Add AFT fields

yunnydang:
This change adds the optional AFT fields, specifically the sender and recipient fields.

Local:
6205 tests, 81261 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
70 tests, 388 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
102 tests, 453 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
97.0588% passed

Note: I have this implemented to bail out of the helper method if none of the 4 conditions are met. Im thinking that this would make sense but can of course listen to others suggestions. Also im not entirely sure how much of a difference it makes for capture_method to be manual, but its only set that way on authorize calls but i tested a purchase call with this and it returns a successful transaction as well.

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -54,6 +54,8 @@ module ActiveMerchant # :nodoc:
             request_three_d_secure(post, options)
             add_level_three(post, options)
             add_card_brand(post, options)
+            add_aft_recipient_details(post, options)
+            add_aft_sender_details(post, options)
             post[:expand] = ['charges.data.balance_transaction']
 
             CREATE_INTENT_ATTRIBUTES.each do |attribute|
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_stripe_payment_intents_test.rb
```diff
@@ -531,6 +531,59 @@ def test_successful_purchase_with_skip_radar_rules
...
+          state: 'KA',
+          city: 'Preston'
+        },
+        account_details: {
+          card: {
+            first6: '123456',
+            last4: '1234'
+          },
+          bank: {
+            account_number: '1234567890'
```
jcreiff, line 557: Hey @yunnydang it occurs to me that if this is an option that customers might be sending to stripe, it's something we would need to scrub from the transcript (and also other places in the transaction 😓 ). I'm going to check with the AM about whether or not this is a need; if not, we may want to rethink leaving this wide open and instead be more prescriptive about how we map the fields 😞

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -394,6 +396,62 @@ def add_level_three(post, options = {})
...
+        address = recipient_details[:address] if recipient_details[:address]
+        account_details = recipient_details[:account_details] if recipient_details[:account_details]
+
+        post[:payment_method_options] ||= {}
+        post[:payment_method_options][:card] ||= {}
+        post[:payment_method_options][:card][:recipient_details] ||= {}
+        post[:payment_method_options][:card][:recipient_details][:address] ||= {}
+        post[:payment_method_options][:card][:recipient_details][:account_details] ||= {}
+        post[:payment_method_options][:card][:recipient_details][:account_details][:card] ||= {}
+        post[:payment_method_options][:card][:recipient_details][:account_details][:unique_identifier] ||= {}
```
jcreiff, line 412: I don't think you need to use the `||=` here for two reasons: 

1. There are no other places in the integration where `recipient_details` is being set. The reason the `||=` is useful for `[:payment_method_options][:card]` is that there are multiple sub-methods that try to add details in that block, so we can ensure that we are instantiating the `{}` if necessary but not overwriting something else if it already exists
2. Since so many of these fields are optional depending on the situation, I might recommend grouping each block a little differently, for example:
```
if address = recipient_details[:address]
  post[:payment_method_options][:card][:recipient_details][:address] = {}
  post[:payment_method_options][:card][:recipient_details][:address][:line1] = address[:line1] if address[:line1]
...etc
end

if account_details = recipient_details[:account_details]
  post[:payment_method_options][:card][:recipient_details][:account_details] = {}
  ...etc
end

```
I think this will make it a little easier to follow what's happening in the method, and it will also protect against potential occurrences of `NoMethodError`. Right now the way this is set up, if someone omits `unique_identifier` for example, we are still going to try to dig all the way to `account_details[:unique_identifier][:identifier]`, which will raise an exception

I would recommend a similar re-arrangement in the `sender_details` section too

yunnydang: thanks! I was going back and forth with the ||= after recipient details but i like your reasoning! will get these suggestions up shortly!

yunnydang: > I don't think you need to use the `||=` here for two reasons:
> 
> 1. There are no other places in the integration where `recipient_details` is being set. The reason the `||=` is useful for `[:payment_method_options][:card]` is that there are multiple sub-methods that try to add details in that block, so we can ensure that we are instantiating the `{}` if necessary but not overwriting something else if it already exists
> 2. Since so many of these fields are optional depending on the situation, I might recommend grouping each block a little differently, for example:
> 
> ```
> if address = recipient_details[:address]
>   post[:payment_method_options][:card][:recipient_details][:address] = {}
>   post[:payment_method_options][:card][:recipient_details][:address][:line1] = address[:line1] if address[:line1]
> ...etc
> end
> 
> if account_details = recipient_details[:account_details]
>   post[:payment_method_options][:card][:recipient_details][:account_details] = {}
>   ...etc
> end
> ```
> 
> I think this will make it a little easier to follow what's happening in the method, and it will also protect against potential occurrences of `NoMethodError`. Right now the way this is set up, if someone omits `unique_identifier` for example, we are still going to try to dig all the way to `account_details[:unique_identifier][:identifier]`, which will raise an exception
> 
> I would recommend a similar re-arrangement in the `sender_details` section too

Also good callout on the nomethodError! too many nest hashes and the suggested edits looks so much more readable

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -394,6 +396,72 @@ def add_level_three(post, options = {})
...
+        if account_details = recipient_details[:account_details]
+          post[:payment_method_options][:card][:recipient_details][:account_details] = {}
+
+          if card = account_details[:card]
+            post[:payment_method_options][:card][:recipient_details][:account_details][:card] = {}
+            post[:payment_method_options][:card][:recipient_details][:account_details][:card][:first6] = card[:first6] if card[:first6]
+            post[:payment_method_options][:card][:recipient_details][:account_details][:card][:last4] = card[:last4] if card[:last4]
+          end
+
+          if (unique_identifier = account_details[:unique_identifier]) && (unique_identifier[:identifier])
```
jcreiff, line 430: I think the `&&` here could return a NoMethodError if `unique_identifier` is passed in as a String instead of a Hash (not that we'd expect that to happen but you never know)

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -394,6 +396,72 @@ def add_level_three(post, options = {})
...
+      def add_aft_sender_details(post, options)
+        return unless options[:sender_details]&.is_a?(Hash)
+
+        sender_details = options[:sender_details]
+        post[:payment_method_options] ||= {}
+        post[:payment_method_options][:card] ||= {}
+        post[:payment_method_options][:card][:sender_details] = {}
+        post[:payment_method_options][:card][:sender_details][:first_name] = sender_details[:first_name] if sender_details[:first_name]
+        post[:payment_method_options][:card][:sender_details][:last_name] = sender_details[:last_name] if sender_details[:last_name]
+
```
jcreiff, line 446: Could we add the other optional `sender` fields here too just in case? I see `email`, `nationality`, `occupation`, and `birth_country` listed in the docs

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
yunnydang: ahh that was rubocops doing one moment!

yunnydang: alrighty I guess for better error handling, ill check if the objects were expecting are hashes before setting and adding fields. We'll send it through and let stripe return any error messages rather than us failing on our side

jcreiff (approved):

PR #5419, opened 2025-03-03
Nuvei: Fix Issues

javierpedrozaing:
Description
-------------------------
[SER-1624](https://spreedly.atlassian.net/browse/SER-1624)
[SER-1525](https://spreedly.atlassian.net/browse/SER-1625)
[SER-1627](https://spreedly.atlassian.net/browse/SER-1627)
    
This commit add the next fixes.
    

- Fix the scrubing of the CVV value
-  Fix when send browser_size in a 3ds GSF transaction
-  Fix get billing_address from payment method otherwise override it
    with transaction object billing address

Unit test
-------------------------
36 tests, 184 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote test
-------------------------
45 tests, 150 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Rubocop
-------------------------

### DIFF
--- a/lib/active_merchant/billing/gateways/nuvei.rb
+++ b/lib/active_merchant/billing/gateways/nuvei.rb
@@ -201,7 +201,7 @@ module ActiveMerchant
         transcript.
           gsub(%r((Authorization: Basic )\w+), '\1[FILTERED]').
           gsub(%r(("cardNumber\\?":\\?")[^"\\]*)i, '\1[FILTERED]').
-          gsub(%r(("cardCvv\\?":\\?")\d+), '\1[FILTERED]').
+          gsub(%r(("CVV\\?":\\?")\d+), '\1[FILTERED]').
           gsub(%r(("merchantId\\?":\\?")\d+), '\1[FILTERED]').
           gsub(%r(("merchantSiteId\\?":\\?")\d+), '\1[FILTERED]').
           gsub(%r(("merchantKey\\?":\\?")\d+), '\1[FILTERED]').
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/nuvei.rb
```diff
@@ -162,10 +162,10 @@ def add_account_funding_transaction(post, payment, options = {})
...
           firstName: payment.first_name,
           lastName: payment.last_name,
-          country: options.dig(:billing_address, :country),
-          address: options.dig(:billing_address, :address1),
-          city: options.dig(:billing_address, :city),
-          state: options.dig(:billing_address, :state)
+          country: payment.try(:country) || options.dig(:billing_address, :country),
+          address: payment.try(:address1) || payment.try(:address2) || options.dig(:billing_address, :address1),
+          city: payment.try(:city) || options.dig(:billing_address, :city),
+          state: payment.try(:state) || options.dig(:billing_address, :state)
```
naashton, line 168: I don't think this will work since address information is not a part of any of AM's classes related to payment methods.

naashton (changes requested):
Remove changes related to the billing address and we can address the other issues. I think we need to just clarify with the merchant how to use AFT's so they avoid issues in the future.

lib/active_merchant/billing/gateways/nuvei.rb
javierpedrozaing: Hi @naashton I thought that too, but I think if they create a payment method with spreedly API, we can get the billing informattion from there ,can't we ?

javierpedrozaing: Oh nevermind I just saw your comment on the ticket and it is already covered, thanks

naashton (approved):
LGTM!

PR #5425, opened 2025-03-11
Add american express credit card support to HiPay gateway

sinourain:
Spreedly reference: [OPPS-377](https://spreedly.atlassian.net/browse/OPPS-377)

Add American Express credit card support to HiPay gateway
Disclaimer: Added a frictionless test because we only found this test card in [this](https://support.hipay.com/hc/en-us/articles/213882649-How-can-I-test-payment-methods) guide


jherreraa (approved):
LGTM!

PR #5426, opened 2025-03-11
Remove AVSname accent for Orbital gateway

sinourain:
Orbital: Fix cardholder's name that contains special characters and accents is not being filtered out.


almalee24:
@sinourain looks like you have a test failing. Can you look into that failing test then I'll review?

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -619,7 +621,7 @@ def add_echeck(xml, check, options = {})
         add_bank_account_type(xml, check)
         xml.tag! :ECPAuthMethod, options[:auth_method] if options[:auth_method]
         xml.tag! :BankPmtDelv, options[:payment_delivery] || 'B'
-        xml.tag! :AVSname, (check&.name ? check.name[0..29] : nil) if get_address(options).blank?
+        xml.tag! :AVSname, (check&.name ? I18n.transliterate(check.name)[0..29] : nil) if get_address(options).blank?
```
almalee24, line 624: Can you also add a test for this?

almalee24: It would also be good to tests this remotely

sinourain: Done!

almalee24 (approved):

PR #5430, opened 2025-03-17
Adyen: Update shopperInteraction

almalee24:
If stored_credential[:reason_type] is unschedule we should only be setting shopperInteraction as Ecommerce if initial_transaction is true and stored_credential[:initiator] is cardholder.

Remote
152 tests, 481 assertions, 12 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 92.1053% passed

Unit
138 tests, 731 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed


jherreraa (approved):
Good work! Looks great!

PR #5442, opened 2025-04-02
CyberSource: Void Acepts nil values

gasb150:
Summary
-------------------------
Improve void request to handle cases where we
don't receive an authorization, (timeOuts cases)

[SER-1636](https://spreedly.atlassian.net/browse/SER-1636)

Unit Tests
----------------
Finished in 3.532173 seconds.
167 tests, 981 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests
----------------
Finished in 127.168569 seconds.
143 tests, 730 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 97.9021% passed

Rubocop
----------------

### DIFF
--- a/lib/active_merchant/billing/gateways/cyber_source.rb
+++ b/lib/active_merchant/billing/gateways/cyber_source.rb
@@ -489,8 +489,11 @@ module ActiveMerchant # :nodoc:
       end
 
       def build_void_request(identification, options)
-        order_id, request_id, request_token, action, money, currency = identification.split(';')
+        order_id, request_id, request_token, action, money, currency = identification&.split(';')
         options[:order_id] = order_id
+
+        money ||= options[:amount]
+        currency ||= options[:currency]
 
         xml = Builder::XmlMarkup.new indent: 2
         case action
@@ -1038,7 +1041,7 @@ module ActiveMerchant # :nodoc:
 
       def add_auth_reversal_service(xml, request_id, request_token)
         xml.tag! 'ccAuthReversalService', { 'run' => 'true' } do
-          xml.tag! 'authRequestID', request_id
+          xml.tag! 'authRequestID', request_id unless request_id.nil?
           xml.tag! 'authRequestToken', request_token
         end
       end
@@ -1075,6 +1078,7 @@ module ActiveMerchant # :nodoc:
       end
 
       def add_merchant_category_code(xml, options)
+        xml.tag! 'merchantTransactionIdentifier', options[:merchant_transaction_id] if options[:merchant_transaction_id]
         xml.tag! 'merchantCategoryCode', options[:merchant_category_code] if options[:merchant_category_code]
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -488,9 +488,13 @@ def reference_is_a_check?(payment_method_or_reference)
         payment_method_or_reference.is_a?(String) && payment_method_or_reference.split(';')[7] == 'check'
       end
 
-      def build_void_request(identification, options)
-        order_id, request_id, request_token, action, money, currency = identification.split(';')
-        options[:order_id] = order_id
+      def build_void_request(identification, options, money = nil)
+        if identification.nil?
```
javierpedrozaing, line 492: I am not sure about this validation, maybe we can use something like this.

```suggestion
        `order_id, request_id, request_token, action, money, currency = identification&.split(';')`
```

javierpedrozaing (commented):
I left a small comment, LGTM.

lib/active_merchant/billing/gateways/cyber_source.rb
```diff
@@ -221,8 +221,8 @@ def purchase(money, payment_method, options = {})
         end
       end
 
-      def void(identification, options = {})
-        commit(build_void_request(identification, options), :void, nil, options)
+      def void(identification, options = {}, money = nil)
```
Heavyblade, line 224: ⚠️ Warning - ❓ Question:

Core doesn't send money value, so what's the point of setting a default value in a way that will never be filled.

Heavyblade (commented):
@gasb150 left you some feedback

lib/active_merchant/billing/gateways/cyber_source.rb
gasb150: My idea is to modify the void method in the backend, 

As far as I tested, the money value is required, so I need the money value in void, as other methods (with partial void) handle it.

test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -412,6 +412,15 @@ def test_purchase_and_void
     assert_successful_response(void)
   end
 
+  def test_void_with_no_authorization_value
+    merchant_transaction_id = 'testTransaction131' + SecureRandom.hex(3)
+
+    @gateway.authorize(@amount, @credit_card, @options.merge(merchant_transaction_id:))
```
naashton, line 418: It looks like you are setting the key in the merge, but not giving it a value `.merge(merchant_transaction_id:))`

naashton (changes requested):
Looks pretty good, but there's one small issue in the remote test.

Can you also add a unit test here and include a validation that inspects the request for the `nil` value for the `authRequestID`. This is a good candidate for that type of test.

test/remote/gateways/remote_cyber_source_test.rb
naashton: Ah, thanks for the explantation!

naashton (approved):
Looks good! Thanks!

PR #5446, opened 2025-04-16
Payeezy: Update authorization when extra space

almalee24:
If transaction_id in response has a extra space then use delete will be used to remove the spaces. This will prevent issues such as URI::InvalidURIError.


adarsh-spreedly (approved):
LGTM

PR #5447, opened 2025-04-17
Include additional scrub on store call

naashton:
This adds an additional scrub for `pan` when the transaction is a `store`


rachelkirk (approved):
I see that the test covers credit card number and cvv and not pan specifically, but looked in store method to see how request is formatted. LGTM! 🚢

PR #5449, opened 2025-04-18
Checkout: Bug fix for AFT date_of_birth

rachelkirk:
Update field in sender object for AFT transactions to date_of_birth. This is what is now stated in their docs.

Unit Tests
75 tests, 510 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests
Hot Mess Express
Credentials issues, working to fix. The remote test I updated passes.

### DIFF
--- a/lib/active_merchant/billing/gateways/checkout_v2.rb
+++ b/lib/active_merchant/billing/gateways/checkout_v2.rb
@@ -245,11 +245,13 @@ module ActiveMerchant # :nodoc:
 
         sender = options[:sender]
 
+        birthdate = sender[:dob] || sender[:date_of_birth]
+
         post[:sender] = {}
         post[:sender][:type] = sender[:type] if sender[:type]
         post[:sender][:first_name] = sender[:first_name] if sender[:first_name]
         post[:sender][:last_name] = sender[:last_name] if sender[:last_name]
-        post[:sender][:dob] = sender[:dob] if sender[:dob]
+        post[:sender][:date_of_birth] = birthdate if birthdate
         post[:sender][:reference] = sender[:reference] if sender[:reference]
         post[:sender][:company_name] = sender[:company_name] if sender[:company_name]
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -249,7 +249,7 @@ def add_payment_sender_data(post, options)
         post[:sender][:type] = sender[:type] if sender[:type]
         post[:sender][:first_name] = sender[:first_name] if sender[:first_name]
         post[:sender][:last_name] = sender[:last_name] if sender[:last_name]
-        post[:sender][:dob] = sender[:dob] if sender[:dob]
+        post[:sender][:date_of_birth] = sender[:date_of_birth] if sender[:date_of_birth]
```
jcreiff, line 252:
```suggestion
        post[:sender][:date_of_birth] = birthday if birthday
```

jcreiff: This would also require a `birthday = sender[:dob] || sender[:date_of_birth]` added above, but I can't make suggestions up that high

lib/active_merchant/billing/gateways/checkout_v2.rb
```diff
@@ -245,11 +245,13 @@ def add_payment_sender_data(post, options)
...
         sender = options[:sender]
 
+        birthdate = sender[:dob] || sender[:date_of_birth]
+
         post[:sender] = {}
         post[:sender][:type] = sender[:type] if sender[:type]
         post[:sender][:first_name] = sender[:first_name] if sender[:first_name]
         post[:sender][:last_name] = sender[:last_name] if sender[:last_name]
-        post[:sender][:dob] = sender[:dob] if sender[:dob]
+        post[:sender][:date_of_birth] = birthdate
```
jcreiff, line 254:
```suggestion
        post[:sender][:date_of_birth] = birthdate if birthdate
```

In case the API is going to complain about a blank/empty value being sent, I think we should guard with `if` like the others

jcreiff (approved):
Thanks for fixing!

PR #5450, opened 2025-04-21
Summary: Normalize API version for SecurePayAu, Simetrik and StripePaymentIntents.

adarsh-spreedly:
Summary:
-------------------------------------------
Normalize API version for SecurePayAu, Simetrik and StripePaymentIntents.

Unit Tests:
-------------------------------------------
6247 tests, 81499 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 
100% passed

RuboCop:
-------------------------------------------
809 files inspected, no offenses detected.


ankurspreedly (approved):

test/unit/gateways/secure_pay_au_test.rb
```diff
@@ -27,6 +28,18 @@ def test_supported_card_types
...
   end
 
+  def test_endpoint
+    xml_request = @gateway.send(:build_request, :purchase, Builder::XmlMarkup.new.target!)
+    doc = Nokogiri::XML(xml_request)
+    assert_equal 'xml-4.2', doc.xpath('SecurePayMessage/MessageInfo/apiVersion').text
+  end
+
+  def test_periodic_endpoint
+    xml_request = @gateway.send(:build_periodic_request, Builder::XmlMarkup.new.target!)
```
almalee24, line 38: You don't really have to use nokogiri to check for the api version. For example if you go to [test_periodic_payment_submits_order_id](https://github.com/activemerchant/active_merchant/blob/master/test/unit/gateways/secure_pay_au_test.rb#L82) you can add `assert_match(/<apiVersion>spxml-3.0<\/apiVersion>/, data)` to check the Periodic API Version

test/unit/gateways/secure_pay_au_test.rb
```diff
@@ -27,6 +28,18 @@ def test_supported_card_types
     assert_equal %i[visa master american_express diners_club jcb], SecurePayAuGateway.supported_cardtypes
   end
 
+  def test_endpoint
+    xml_request = @gateway.send(:build_request, :purchase, Builder::XmlMarkup.new.target!)
```
almalee24, line 32: for this api version you can do the similar to what I mentioned in the below comment but in `test_localized_currency` for example. If we go to the [purchase](lib/active_merchant/billing/gateways/secure_pay_au.rb) method in looks like the Periodic API only gets used if the payment method doesn't respond to number.

adarsh-spreedly:
> Looks like you forgot to change the StripePaymentIntents version and you updated the wrong version for Stripe. So for Stripe/StripePaymentIntents its a little bit different than the other gateways. Their version lives in DEFAULT_API_VERSION and this gets passed in the headers through `Stripe-Version`.
> 
> For version for StripePaymentIntents is `DEFAULT_API_VERSION = '2020-08-27'` this is in lib/active_merchant/billing/gateways/stripe_payment_intents.rb. For Stripe is DEFAULT_API_VERSION = '2020-08-27'. StripePaymentIntents inherits from Stripe

Thanks for sharing
I'll update the version for stripe_payment_intents, but as you said stripe is the parent class and it also have the default version and both uses same method `api_version` to get the version so I am making changes in both file unless you want the changes in only stripe_payment_intents file.

almalee24 (approved):

lib/active_merchant/billing/gateways/stripe.rb
```diff
@@ -664,7 +664,7 @@ def stripe_client_user_agent(options)
       end
 
       def api_version(options)
```
almalee24, line 666: We need to add this method in StripePaymentIntents because currently for StripePaymentIntent transactions we are getting the version from Stripe and not StripePaymentIntents

almalee24 (changes requested):

lib/active_merchant/billing/gateways/stripe.rb
almalee24: @adarsh-spreedly when you get an opportunity to fix this

adarsh-spreedly: @almalee24 ok, currently the version are same for both, is this required because you are going to update the stripe version or it is just we want it to be separate for both?

almalee24: Yes they are the same but if they weren't it would cause an issue which is why it's a bug.

adarsh-spreedly: done!

almalee24 (approved):

PR #5452, opened 2025-04-21
MultiPay: Base Gateway adapter

javierpedrozaing:
Description
-------------------------
This commit add a the initial setup for MultiPay gateway 
with basic operations like authorize, purchase and capture

WIP

Unit test
-------------------------

Remote test
-------------------------

Rubocop
-------------------------


test/remote/gateways/remote_multi_pay_test.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def test_transcript_scrubbing
+    transcript = capture_transcript(@gateway) do
+      @gateway.purchase(@amount, @credit_card, @options)
+    end
+    transcript = @gateway.scrub(transcript)
+
+    assert_scrubbed(@credit_card.number, transcript)
+    assert_scrubbed(@credit_card.verification_value, transcript)
+    assert_scrubbed(@gateway.options[:api_key], transcript)
```
naashton, line 68: We should also make sure we are scrubbing the `password` that we have in the fixtures

test/remote/gateways/remote_multi_pay_test.rb
```diff
@@ -0,0 +1,76 @@
+require 'test_helper'
+
+class RemoteMultiPayTest < Test::Unit::TestCase
+  def setup
+    @gateway = MultiPayGateway.new(fixtures(:multi_pay))
+
+    @amount = 100
```
naashton, line 7: Might be helpful to include this value in a test to assert that it is what we expect it should be (`1.00`)

naashton (commented):
Just a first pass review.

lib/active_merchant/billing/gateways/multi_pay.rb
```diff
@@ -0,0 +1,240 @@
...
+          RequestType: request_type,
+          SystemIdentification: '1.0.0', # optional
+          CompanyIdentification: @company,
+          BranchIdentification: @branch,
+          POSIdentification: @pos,
+          CardNumber: payment.number,
+          CardExp: expdate(payment),
+          SecurityCode: payment.verification_value,
+          Security: build_security_data(options),
+          Amount: amount(money),
```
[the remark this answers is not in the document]
javierpedrozaing: Good question, I think yes and should be different depending on the currency code USD or CLP, I will add it. thanks

test/remote/gateways/remote_multi_pay_test.rb
javierpedrozaing: I will check, thanks

test/remote/gateways/remote_multi_pay_test.rb
javierpedrozaing: yes, I will include a test for amount

PR #5453, opened 2025-04-22
CyberSource: Add the optional merchant reference number for capture

yunnydang:
Adds the optional merchant reference number override on the capture endpoint.

Note: Im having issues running the remote tests for this since the introduction of line number 9 in the remote spec file:

    @gateway_certificate = CyberSourceGateway.new({ nexus: 'NC' }.merge(fixtures(:cyber_source_certificate)))

Though hen i comment it out my remote spec will pass. I checked that the fixture is properly in place but still run into an argument error.

### DIFF
--- a/test/remote/gateways/remote_cyber_source_test.rb
+++ b/test/remote/gateways/remote_cyber_source_test.rb
@@ -753,6 +753,27 @@ class RemoteCyberSourceTest < Test::Unit::TestCase
     assert !response.authorization.blank?
   end
 
+  def test_successful_capture_with_merchant_reference_code_override
+    assert auth = @gateway.authorize(@amount, @credit_card, @options)
+    assert_successful_response(auth)
+
+    assert response = @gateway.capture(@amount, auth.authorization, @capture_options.merge(merchant_reference_code: '1234566778'))
+    assert_successful_response(response)
+    assert !response.authorization.blank?
+    assert_equal '1234566778', response.params['merchantReferenceCode']
+  end
+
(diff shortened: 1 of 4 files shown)


test/remote/gateways/remote_cyber_source_test.rb
```diff
@@ -753,6 +753,16 @@ def test_successful_capture_with_issuer_additional_data
     assert !response.authorization.blank?
   end
 
+  def test_successful_capture_with_merchant_reference_number_override
```
jcreiff, line 756: While i think this override is straightforward and should do what we need, for the sake of being thorough I think it would be good to add a test that demonstrates that the order ID set from the previous authorization still goes through as expected, since there's not existing test coverage for that behavior as far as I can tell 🧐

jcreiff (changes requested):

jcreiff:
@yunnydang one other question - is there any compelling reason to name this `merchant_reference_number` instead of `merchant_reference_code`? I realize the ticket was written with `_number` but now I notice that the Cybersource field is `_code`

yunnydang:
> @yunnydang one other question - is there any compelling reason to name this `merchant_reference_number` instead of `merchant_reference_code`? I realize the ticket was written with `_number` but now I notice that the Cybersource field is `_code`

now that you mentioned it, i do think renaming it code would be better suited. Not sure why the ticket calls out for number specifically, but will have the changes pushed up!

jcreiff (approved):

PR #5455, opened 2025-04-22
Orbital keep only A-Za-z characters

sinourain:
Orbital keep only A-Za-z characters

### DIFF
--- a/lib/active_merchant/billing/gateway.rb
+++ b/lib/active_merchant/billing/gateway.rb
@@ -316,6 +316,14 @@ module ActiveMerchant # :nodoc:
         [first_name, last_name]
       end
 
+      def clean_emoji_emoticons_and_accent(str = '')
+        str = str.force_encoding('utf-8').encode
+        emoji_regex = /[\u{1f600}-\u{1f64f}\u{2702}-\u{27b0}\u{1f680}-\u{1f6ff}\u{24C2}-\u{1F251}\u{1f300}-\u{1f5ff}]/
+        emoticon_regex = /[:;=8xX]-?[)(DPpO3]/
+        str = str.gsub(emoji_regex, '').gsub(emoticon_regex, '').strip
+        I18n.transliterate(str)
+      end
+
       def split_address(full_address)
         address_parts = (full_address || '').split
         return [nil, nil] if address_parts.size == 0
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -538,7 +538,7 @@ module ActiveMerchant # :nodoc:
           elsif options[:billing_address] && options[:billing_address][:name].present?
             options[:billing_address][:name]
           end
-        name ? I18n.transliterate(name || '')[0..29] : nil
+        name ? clean_emoji_emoticons_and_accent(name)[0..29] : nil
       end
 
       def avs_supported?(address)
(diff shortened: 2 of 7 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -538,7 +538,7 @@ def billing_name(payment_source, options)
           elsif options[:billing_address] && options[:billing_address][:name].present?
             options[:billing_address][:name]
           end
-        name ? I18n.transliterate(name || '')[0..29] : nil
+        name ? I18n.transliterate(name).gsub(/[^a-zA-Z ]/, ' ').strip[0..29] : nil
```
almalee24, line 541: Certain special characters doesn't cause issues. I wonder if there's a different way to look for emojis

test/remote/gateways/remote_orbital_test.rb
```diff
@@ -746,7 +746,7 @@ def test_credit_purchase_with_address_responds_with_name
 
   def test_truncates_and_removes_accents_from_name
     truncated_name = 'Jose Maria Lopez Garc'
-    credit_card = credit_card('4556761029983886', first_name: 'José María', last_name: 'López García')
+    credit_card = credit_card('4556761029983886', first_name: ':) José María', last_name: 'López García')
```
almalee24, line 749: :) passes without issues even if you remove the `gsub(/[^a-zA-Z ]/, ' ')`

almalee24: Oh it looks like the sandbox doesn't work the same way that in production. Even sending emojies and accents doesn't cause any issues.

lib/active_merchant/billing/gateways/orbital.rb
almalee24, line 541: I need to test this through postman

test/remote/gateways/remote_orbital_test.rb
sinourain: I updated it to remove the emojis and emoticons in the consumer name

lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -538,7 +538,7 @@ def billing_name(payment_source, options)
           elsif options[:billing_address] && options[:billing_address][:name].present?
             options[:billing_address][:name]
           end
-        name ? I18n.transliterate(name || '')[0..29] : nil
+        name ? clean_emoji_emoticons_and_accent(name)[0..29] : nil
```
almalee24, line 541: I don't think this change is needed for this gateway

lib/active_merchant/billing/gateway.rb
```diff
@@ -316,6 +316,14 @@ def split_names(full_name)
         [first_name, last_name]
       end
 
+      def clean_emoji_emoticons_and_accent(str = '')
+        str = str.force_encoding('utf-8').encode
+        emoji_regex = /[\u{1f600}-\u{1f64f}\u{2702}-\u{27b0}\u{1f680}-\u{1f6ff}\u{24C2}-\u{1F251}\u{1f300}-\u{1f5ff}]/
+        emoticon_regex = /[:;=8xX]-?[)(DPpO3]/
+        str = str.gsub(emoji_regex, '').gsub(emoticon_regex, '').strip
+        I18n.transliterate(str)
```
almalee24, line 324: Looks like you changed it since yesterday. This definitely works. For MercadoPago just works `I18n.transliterate(name)` also works. It just ends up replacing the emoji with some question marks.

lib/active_merchant/billing/gateways/orbital.rb
almalee24, line 541: I guess it doesn't harm to do this for orbital since their sandbox is not the most accurate anyways

lib/active_merchant/billing/gateway.rb
sinourain: I think I prefer not to have the question mark appear because sometimes the gateway verifies the name printed on the card.

almalee24 (approved):

PR #5456, opened 2025-04-23
StripePI: Add inquire capability

adarsh-spreedly:
Summary:
----------------------------------
Add inquire capability to StripePaymentIntents.

Unit Test:
----------------------------------
6246 tests, 81506 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Test:
----------------------------------
107 tests, 491 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

RuboCop:
----------------------------------
809 files inspected, no offenses detected


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -314,6 +314,16 @@ def setup_purchase(money, options = {})
         commit(:post, 'payment_intents', post, options)
       end
 
+      def inquire(intent_id, options = {})
```
almalee24, line 317: Can you make it be `authorization` instead of `intent_id` just to follow the same patter we have for the other gateways

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -314,6 +314,16 @@ def setup_purchase(money, options = {})
...
+        if intent_id.start_with?('pi_')
+          show_intent(intent_id, options)
+        elsif intent_id.start_with?('seti_')
+          retrieve_setup_intent(intent_id, options)
+        else
+          unsupported_inquire_response
```
almalee24, line 323: This is not needed since it wouldn't be possible for a transactions to have a different intent_id. Only PaymentIntent and SetupIntent are supported by StripePI

almalee24 (changes requested):
It looks like `success_from` needs to be update since the response body is different for inquire request. It looks like when a create_intent has a failed response it has `error` but not `status` however a show_intent doesn't have error but it looks like it became `last_payment_error`. The show_intent response does has status but a failed response won't always return a status of failed.

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -829,8 +838,12 @@ def format_idempotency_key(options, suffix)
 
       def success_from(response, options)
         if response['status'] == 'requires_action' && !options[:execute_threed]
-          response['error'] = {}
-          response['error']['message'] = 'Received unexpected 3DS authentication response, but a 3DS initiation flag was not included in the request.'
+          response['error'] = { 'message' => 'Received unexpected 3DS authentication response, but a 3DS initiation flag was not included in the request.' }
+          return false
+        end
+
+        if options[:request_type] == :inquire && (error = response['last_setup_error'] || response['last_payment_error'])
```
adarsh-spreedly, line 845: @almalee24 I think assigning the `last_error` to `error` will preserve us from overriding other methods which are dependent on `error` field like `message_from`, `error_id`.

almalee24: This looks great, let me test it

almalee24 (approved):

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -829,8 +838,12 @@ def format_idempotency_key(options, suffix)
 
       def success_from(response, options)
         if response['status'] == 'requires_action' && !options[:execute_threed]
```
almalee24, line 840: When inquiring on pending transaction it fails here because the request type doesn't have execute_threed. If you add `options[:request_type] != :inquire` it should fix it.

almalee24 (changes requested):
success_from needs to be update to accommodate inquiring for pending transactions.

almalee24 (approved):

PR #5457, opened 2025-04-23
Credorax: support zero dollar verify

yunnydang:
This supports the optional zero dollar verify for the credorax gateway. Per documentation when doing a zero $0 verify, we should be sending the a9 field set to the value 5, while skipping the subsequent void request. This change will also require the additional optional field of zero_dollar_auth to trigger said verify request. Also updated the remote test file to fix a handful of the failing tests

Local:
6244 tests, 81498 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
85 tests, 422 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
56 tests, 197 assertions, 6 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
89.2857% passed

The 6 failing remote tests are 3ds related and the error message is "Format error". Not sure whats causing it.

### DIFF
--- a/lib/active_merchant/billing/gateways/credorax.rb
+++ b/lib/active_merchant/billing/gateways/credorax.rb
@@ -240,10 +240,11 @@ module ActiveMerchant # :nodoc:
         commit(:credit, post)
       end
 
-      def verify(credit_card, options = {})
+      def verify(payment_method, options = {})
+        amount = eligible_for_0_auth?(payment_method, options) ? 0 : 100
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
-          r.process(:ignore_result) { void(r.authorization, options) }
+          r.process { authorize(amount, payment_method, options) }
+          r.process(:ignore_result) { void(r.authorization, options) } unless eligible_for_0_auth?(payment_method, options)
         end
       end
 
@@ -302,6 +303,10 @@ module ActiveMerchant # :nodoc:
         post[:b3] = format(payment_method.month, :two_digits)
       end
 
+      def eligible_for_0_auth?(payment_method, options = {})
+        payment_method.is_a?(CreditCard) && %w(visa master).include?(payment_method.brand) && options[:zero_dollar_auth]
+      end
+
       def add_network_tokenization_card(post, payment_method, options)
         post[:b21] = NETWORK_TOKENIZATION_CARD_SOURCE[payment_method.source.to_s]
         post[:token_eci] = post[:b21] == 'vts_mdes_token' ? '07' : nil
@@ -330,12 +335,12 @@ module ActiveMerchant # :nodoc:
       def add_customer_data(post, options)
         post[:d1] = options[:ip] || '127.0.0.1'
         if (billing_address = options[:billing_address])
-          post[:c5]   = billing_address[:address1]  if billing_address[:address1]
-          post[:c7]   = billing_address[:city]      if billing_address[:city]
-          post[:c10]  = billing_address[:zip]       if billing_address[:zip]
-          post[:c8]   = billing_address[:state]     if billing_address[:state]
-          post[:c9]   = billing_address[:country]   if billing_address[:country]
-          post[:c2]   = billing_address[:phone]     if billing_address[:phone]
+          post[:c5]   = billing_address[:address1]      if billing_address[:address1]
+          post[:c7]   = billing_address[:city]          if billing_address[:city]
+          post[:c10]  = billing_address[:zip]           if billing_address[:zip]
+          post[:c8]   = billing_address[:state]         if billing_address[:state]
+          post[:c9]   = billing_address[:country]       if billing_address[:country]
+          post[:c2]   = billing_address[:phone] if billing_address[:phone]
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -240,10 +241,11 @@ def credit(amount, payment_method, options = {})
...
       end
 
-      def verify(credit_card, options = {})
+      def verify(payment_method, options = {})
+        amount = (eligible_for_0_auth?(payment_method, options) ? 0 : 100)
         MultiResponse.run(:use_first_response) do |r|
-          r.process { authorize(100, credit_card, options) }
-          r.process(:ignore_result) { void(r.authorization, options) }
+          r.process { authorize(amount, payment_method, options) }
+          r.process(:ignore_result) { void(r.authorization, options) unless eligible_for_0_auth?(payment_method, options) }
```
jcreiff, line 248: Should this `unless` be outside of the `{}`?

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -173,6 +173,7 @@ def authorize(amount, payment_method, options = {})
         add_account_name_inquiry(post, options)
         add_processor(post, options)
         add_authorization_details(post, options)
+        add_transaction_type(post, options)
```
jcreiff, line 176: Could we modify this submethod so that the `a9` value is automatically set to `5` if we the zero-value verify conditions are met? That would simplify the change for customers wanting to use it to passing in one new field instead of two

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -240,10 +241,11 @@ def credit(amount, payment_method, options = {})
         commit(:credit, post)
       end
 
-      def verify(credit_card, options = {})
+      def verify(payment_method, options = {})
+        amount = (eligible_for_0_auth?(payment_method, options) ? 0 : 100)
```
jcreiff, line 245: Small syntax nitpick but I don't think you need the `()` wrapped around this evaluation

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -302,6 +304,10 @@ def add_payment_method(post, payment_method, options)
         post[:b3] = format(payment_method.month, :two_digits)
       end
 
+      def eligible_for_0_auth?(payment_method, options = {})
+        payment_method.is_a?(CreditCard) && %w(visa master).include?(payment_method.brand) && options[:zero_dollar_auth]
```
jcreiff, line 308: I see that in other gateway implementations this is limited to visa/master - do we know that this restriction still applies, based on docs or some other conversation? Just curious.

lib/active_merchant/billing/gateways/credorax.rb
```diff
@@ -330,12 +336,12 @@ def add_stored_credential(post, options)
       def add_customer_data(post, options)
         post[:d1] = options[:ip] || '127.0.0.1'
         if (billing_address = options[:billing_address])
-          post[:c5]   = billing_address[:address1]  if billing_address[:address1]
-          post[:c7]   = billing_address[:city]      if billing_address[:city]
-          post[:c10]  = billing_address[:zip]       if billing_address[:zip]
-          post[:c8]   = billing_address[:state]     if billing_address[:state]
-          post[:c9]   = billing_address[:country]   if billing_address[:country]
-          post[:c2]   = billing_address[:phone]     if billing_address[:phone]
+          post[:c5]   = billing_address[:address1]      if billing_address[:address1]
```
jcreiff, line 339: Did rubocop complain?

test/unit/gateways/credorax_test.rb
```diff
@@ -289,6 +289,16 @@ def test_successful_verify
     assert_equal 'Succeeded', response.message
   end
 
+  def test_successful_verify_with_0_auth
+    response = stub_comms do
+      @gateway.verify(@credit_card, @options.merge(transaction_type: '5', zero_dollar_auth: true))
+    end.check_request do |_endpoint, data, _headers|
+      assert_match(/a9=5/, data)
```
jcreiff, line 296: Can we also assert that the amount is zero?

jcreiff (changes requested):

lib/active_merchant/billing/gateways/credorax.rb
yunnydang: oh woops! thanks for catching this, copy and pasted it too quickly

lib/active_merchant/billing/gateways/credorax.rb
yunnydang: great idea!

lib/active_merchant/billing/gateways/credorax.rb
yunnydang: hmm i dont see this being mentioned anywhere in the document they provided and the ZD thread. should i have shad reach out and confirm?

jcreiff (approved):
This looks good! Provisionally approving so that my previous request for change isn't a blocker while I am ooo, but let's hold off on merging until we determine if the brand specification is a required limitation

yunnydang:
> This looks good! Provisionally approving so that my previous request for change isn't a blocker while I am ooo, but let's hold off on merging until we determine if the brand specification is a required limitation

thats fine, i think this will have to go to CAB anyways!

lib/active_merchant/billing/gateways/credorax.rb
jcreiff, line 176: Just realized based on our conversation in the other Credorax PR that this `add_transaction_type` addition isn't necessary since we have the stored credential method above!

PR #5461, opened 2025-05-05
OPPS-414 - Mapped ecommerce_indicator value

mjdonga:
### Summary
Map ecommerce_indicator value from additionalData3DS to CommerceHub eciIndicator

[OPPS-414](https://spreedly.atlassian.net/browse/OPPS-414)

### Unit Tests
30 tests, 209 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### Remote Tests
37 tests, 100 assertions, 3 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
91.8919% passed

### Rubocop
809 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/commerce_hub.rb
+++ b/lib/active_merchant/billing/gateways/commerce_hub.rb
@@ -141,7 +141,7 @@ module ActiveMerchant # :nodoc:
       def add_transaction_interaction(post, options)
         post[:transactionInteraction] = {}
         post[:transactionInteraction][:origin] = options[:origin] || 'ECOM'
-        post[:transactionInteraction][:eciIndicator] = options[:eci_indicator] || 'CHANNEL_ENCRYPTED'
+        post[:transactionInteraction][:eciIndicator] = map_ecommerce_indicator(options)
         post[:transactionInteraction][:posConditionCode] = options[:pos_condition_code] || 'CARD_NOT_PRESENT_ECOM'
         post[:transactionInteraction][:posEntryMode] = (options[:pos_entry_mode] || 'MANUAL') unless options[:encryption_data].present?
         post[:transactionInteraction][:additionalPosInformation] = {}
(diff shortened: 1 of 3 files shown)


ankurspreedly (approved):
lgtm. can you pls add
- changelog
- details of spec file in PR description.

almalee24 (changes requested):
Let's not stop passing options[:eci_indicator] if it's present

lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -141,7 +141,7 @@ def add_three_d_secure(post, options)
       def add_transaction_interaction(post, options)
         post[:transactionInteraction] = {}
         post[:transactionInteraction][:origin] = options[:origin] || 'ECOM'
-        post[:transactionInteraction][:eciIndicator] = options[:eci_indicator] || 'CHANNEL_ENCRYPTED'
+        post[:transactionInteraction][:eciIndicator] = map_ecommerce_indicator(options[:three_d_secure][:eci]) if options[:three_d_secure].present?
```
almalee24, line 144: just pass options here so that we can still pass `options[:eci_indicator]` if present. We don't want to break people who are actually passing their own eci_indicator

lib/active_merchant/billing/gateways/commerce_hub.rb
```diff
@@ -430,6 +430,17 @@ def authorization_from(action, response, options)
       def error_code_from(response, action)
         response.dig('error', 0, 'code') unless success_from(response, action)
       end
+
+      def map_ecommerce_indicator(eci_indicator)
```
almalee24, line 434: The beginning of this should be updated to be something like 

```
def map_ecommerce_indicator(options)
    return options[:eci_indicator] if options[:eci_indicator]
    return 'CHANNEL_ENCRYPTED' unless options[:three_d_secure]
```

lib/active_merchant/billing/gateways/commerce_hub.rb
mjdonga: Yes, make sense

lib/active_merchant/billing/gateways/commerce_hub.rb
mjdonga: Made this change, please check

almalee24 (approved):

PR #5462, opened 2025-05-05
Pin: Add support for Apple/Google Pay

naashton:
Refactor `add_creditcard` method to support AP and GP payment method types at the Pin Payments gateway

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -161,6 +161,7 @@
 * Payflow: Send ds_trans_id as xid if xid is not present [adarsh-spreedly] #5465
 * Normalize the use of API_VERSION for some gateways (iVeri, JetypayV2 and Kushki) [ritesh-kapoor-spreedly] #5454
 * Normalize API versions for Payeezy, Paymentez, and Paymill [mjdonga] #5451
+* Pin Payments: Add support for Apple and Google payment method types [naashton] #5462
 
 == Version 1.137.0 (August 2, 2024)
 * Unlock dependency on `rexml` to allow fixing a CVE (#5181).


lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -149,23 +149,39 @@ def add_capture(post, options)
...
+          post[:card].merge!(
+            number: payment_method.number,
+            expiry_month: payment_method.month,
+            expiry_year: payment_method.year,
+            network_type: payment_method.source.to_s.gsub('_', ''),
+            cryptogram: payment_method.payment_cryptogram,
+            eci: payment_method.eci,
+            cvc: payment_method.verification_value,
+            name: payment_method.name
+          )
```
AMHOL, line 167: @naashton this won't work, in order to make a charge with a network tokenized credit card it will need to be stored with the [payment sources API endpoint](https://pinpayments.com/developers/api-reference/payment-sources#type__network_token) before posting to the [charges API](https://pinpayments.com/developers/api-reference/charges#post-charges) with the resulting `payment_source_token`.

AMHOL: @naashton the customers API doesn't currently support creating a customer with a payment source token, or network tokenized card, however, you can work around this with the following:

* Store a network tokenized card using the [payment sources API](https://pinpayments.com/developers/api-reference/payment-sources#type__network_token)
* Use the card token returned from the payment sources endpoint under `response.source.token` to create a customer
* You can then use the customer token to make a charge or activate a subscription

naashton: Excellent, thanks for this clarification!

lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -122,8 +122,8 @@ def add_customer_data(post, options)
         post[:ip_address] = options[:ip] if options[:ip]
       end
 
-      def add_address(post, creditcard, options)
-        return if creditcard.kind_of?(String)
+      def add_address(post, payment_method, options)
+        return if payment_method.kind_of?(String)
```
Heavyblade, line 126: 💬 Comment:

If the method already has a guard, maybe the check inside the `purchase` method, line 30 should be better placed here.

lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -149,23 +149,56 @@ def add_capture(post, options)
...
 
-      def add_creditcard(post, creditcard)
-        if creditcard.respond_to?(:number)
-          post[:card] ||= {}
+      def add_payment_method(post, payment_method)
+        return unless payment_method
 
+        case payment_method
+        when NetworkTokenizationCreditCard
+          token = get_single_use_token(payment_method, options)
```
Heavyblade, line 157: 💬 Comment/Style:

If the `token` variable is just a 'transition' variable, maybe a direct assignment to the post hash is better:

```ruby
   post[:payment_source_token] = get_single_use_token(payment_method, options)
```

lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -149,23 +149,56 @@ def add_capture(post, options)
...
+          raise ArgumentError, "Invalid payment method type: #{payment_method.class}"
+        end
+      end
+
+      # Get a single use token from Pin for NT based payment methods
+      # This is used to create a charge using a network tokenized card
+      def get_single_use_token(payment_method, options)
+        return unless payment_method.is_a?(NetworkTokenizationCreditCard)
+
+        # binding.pry
```
Heavyblade, line 184: 💬 Comment:

small typo here with the binding.pry

lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -149,23 +149,56 @@ def add_capture(post, options)
...
+            post[:customer_token] = payment_method
           end
+        else
+          raise ArgumentError, "Invalid payment method type: #{payment_method.class}"
+        end
+      end
+
+      # Get a single use token from Pin for NT based payment methods
+      # This is used to create a charge using a network tokenized card
+      def get_single_use_token(payment_method, options)
```
Heavyblade, line 181: 💬 ⚠️ Comment/Warning:

I see a couple of inconsistencies in this method:

1) The return 'signature' is not consistent, if everything goes OK the method returns an string to be used in the post hash but if there is an error, the rescue is Returning a `Response` object with error response and the `add_payment_method` method is not checking that to raise an exception or actually deal with the error.

2) Isn't harmful but this type check is maybe duplicated => `get_single_use_token` is a private method that is only used by the `add_payment_method` before ensuring the payment_method type.

lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -149,23 +149,56 @@ def add_capture(post, options)
...
+            number: payment_method.number,
+            expiry_month: payment_method.month,
+            expiry_year: payment_method.year,
+            network_type: payment_method.source.to_s.gsub('_', ''),
+            cryptogram: payment_method.payment_cryptogram,
+            eci: payment_method.eci
+          }
+        }
+        begin
+          result = commit(:post, 'payment_sources', post, options)
```
Heavyblade, line 197: ❓ Question:

I'm entirely sure, but perhaps this call could e wrapped into a multi-response object?

Heavyblade (approved):
@naashton done with my review left some comments for your consideration.

lib/active_merchant/billing/gateways/pin.rb
naashton: Thanks for this callout. I've improved the return of this method to be consistent and also removed the duplicate check.

lib/active_merchant/billing/gateways/pin.rb
```diff
@@ -36,18 +36,18 @@ def purchase(money, creditcard, options = {})
         commit(:post, 'charges', post, options)
       end
 
-      # Create a customer and associated credit card. The token that is returned
-      # can be used instead of a credit card parameter in the purchase method
-      def store(creditcard, options = {})
+      # Create a customer and associated payment method. The token that is returned
+      # can be used instead of a payment method parameter in the purchase method
+      def store(payment_method, options = {})
```
AMHOL, line 41: This will not work, the [POST /customers](https://pinpayments.com/developers/api-reference/customers#post-customers) endpoint does not support customer creation with a `payment_source_token`, only embedded `card` parameters or a `card_token`, until this is supported you'd need to send the card token returned under `response.source.token` from the [POST /payment_sources](https://pinpayments.com/developers/api-reference/payment-sources#post-payment_sources) endpoint.

naashton: Thanks, Andy. I'm going to create a ticket for our internal CI team to address this as well as some other issues I think I've identified. It looks like a lot of this integration is a bit old and may have experienced some technical drift, so it probably warrants some attention.

PR #5470, opened 2025-05-09
LoanPaymentPro: Initial implementation

Heavyblade:
## Summary:
This PR introduces the initial implementation of the AM adapter
for [LoanPaymentPro gateway](www.loanpaymentpro.com) with support
for:

  * Authorize
  * Purchase
  * Capture
  * Void
  * Verify
  * Refund

### Remote Test:
18 tests, 44 assertions, 0 failures, 0 errors, 0 pendings,  1 omissions, 0 notifications
100% passed

###   Unit Tests:
Finished in 43.722285 seconds.
6255 tests, 81546 assertions, 0 failures, 0 errors, 0 pending, 0 omissions, 0 notifications
100% passed

### RuboCop:
812 files inspected, no offenses detected

### DIFF
--- a/lib/active_merchant/billing/gateways/loan_payment_pro.rb
+++ b/lib/active_merchant/billing/gateways/loan_payment_pro.rb
@@ -1,0 +1,247 @@
+module ActiveMerchant # :nodoc:
+  module Billing # :nodoc:
+    class LoanPaymentProGateway < Gateway
+      self.live_url = 'https://gateway.loanpaymentpro.com'
+      self.test_url = 'https://gateway.loanpaymentpro.com'
+
+      self.supported_countries = ['US']
+      self.default_currency = 'USD'
+      self.supported_cardtypes = %i[visa master american_express discover]
+
+      self.homepage_url = 'https://loanpaymentpro.com/'
+      self.display_name = 'LoanPaymentPro'
+
+      RESPONSE_CODE_MAPPING = {
+        1 => 'Customer added successfully.',
+        3 => 'Customer removed successfully.',
+        4 => 'An error occurred while removing the customer.',
+        5 => 'Customer updated successfully.',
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/loan_payment_pro.rb
```diff
@@ -0,0 +1,244 @@
...
+          authorize: 'v2-3/paymentcards/authorize',
+          capture: 'v2-1/transactions/{placeholder}/capture',
+          void: 'v2/payments/{placeholder}/void',
+          refund: 'v2/payments/{placeholder}/refund',
+          verify: 'v2/paymentcards/validate/detailed',
+          store: 'v2/paymentcards/add',
+          unstore: 'v2/paymentcards/{placeholder}/delete',
+          authorize_with_token: 'v2-3/payments/paymentcards/{placeholder}/authorize',
+          purchase_with_token: 'v2-3/payments/paymentcards/{placeholder}/run'
+        }[action]&.gsub('{placeholder}', value.to_s.split('|').first || '')
```
naashton, line 185: Any particular advantage of using `gsub` over string interpolation with `#{placeholder}`?

Also, the `value` argument implies a string but it looks like the places where this method is called the 2nd argument?

Might be just be me that's getting a little confused by this method, but maybe this would be a good one to have a comment that explains what the expected behavior is here.

naashton (approved):
Overall, looks like a solid start. Just had some questions about the `path` method that could help clarify things for me.

lib/active_merchant/billing/gateways/loan_payment_pro.rb
Heavyblade: @naashton you are right, I was trying to avoid Rubocop warnings when using string formatting,  but just using string interpolation is way simpler.

PR #5472, opened 2025-05-13
Fix Rapyd error on billing address

jherreraa:


naashton (commented):
I think this change is a good candidate for a unit test that validates the presence or absence of fields, based on what we expect should happen.

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -279,10 +286,11 @@ def add_customer_data(post, payment, options, action = '')
       def customer_fields(payment, options)
         return if options[:customer_id]
 
-        customer_address = address(options)
+        customer_address = address(options) if valid_address?(options[:billing_address])
```
naashton, line 289: Similar to the other request, you could move the call of this method inside of the `address` or `add_address` method.

naashton (changes requested):
A couple of suggestions for you.

I also see now we have 2 different methods for handling the address, `def address` and `def add_address`. Do you think there is an opportunity to refactor this to consolidate the logic of adding address details in the requests?

lib/active_merchant/billing/gateways/rapyd.rb
jherreraa: Got it! Done!

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -138,6 +138,13 @@ def add_address(post, creditcard, options)
         post[:address][:phone_number] = address[:phone].gsub(/\D/, '') if address[:phone]
       end
 
+      def valid_address?(address)
+        required_fields = %i[name address1 city state country zip]
+        missing_fields = required_fields.select { |field| address[field].nil? || address[field].to_s.strip.empty? }
```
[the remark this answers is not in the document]
jherreraa: Corrected! Thanks!

jherreraa:
> I think this change is a good candidate for a unit test that validates the presence or absence of fields, based on what we expect should happen.

Understood! Unit test added

lib/active_merchant/billing/gateways/rapyd.rb
```diff
@@ -138,6 +139,15 @@ def add_address(post, creditcard, options)
         post[:address][:phone_number] = address[:phone].gsub(/\D/, '') if address[:phone]
       end
 
+      def valid_address?(address)
+        return false unless address
+
+        required_fields = %i[name address1 city state country zip]
+        missing_fields = required_fields.select { |field| address[field].nil? || address[field].to_s.strip.empty? }
```
adarsh-spreedly, line 146:
```suggestion
        missing_fields = required_fields.select { |field| address[field].to_s.strip.empty? }
```
As `nil` will be converted to empty string.

PR #5476, opened 2025-05-19
Orbital: Pass override_exp_date if provided

almalee24:
If override_exp_data is in options pass it for Exp instead of the CreditCard expiration date.

Remote:
139 tests, 557 assertions, 0 failures, 1 errors, 0 pendings, 0 omissions, 1 notifications 99.2806% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/orbital.rb
+++ b/lib/active_merchant/billing/gateways/orbital.rb
@@ -587,7 +587,7 @@ module ActiveMerchant # :nodoc:
 
       def add_credit_card(xml, credit_card, options)
         xml.tag! :AccountNum, credit_card.number if credit_card.is_a?(CreditCard)
-        xml.tag! :Exp, expiry_date(credit_card) if credit_card.is_a?(CreditCard)
+        xml.tag! :Exp, options[:override_exp_date] || expiry_date(credit_card) if credit_card.is_a?(CreditCard)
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/orbital.rb
```diff
@@ -587,7 +587,7 @@ def add_echeck(xml, check, options = {})
 
       def add_credit_card(xml, credit_card, options)
         xml.tag! :AccountNum, credit_card.number if credit_card.is_a?(CreditCard)
-        xml.tag! :Exp, expiry_date(credit_card) if credit_card.is_a?(CreditCard)
+        xml.tag! :Exp, options[:override_exp_date] || expiry_date(credit_card) if credit_card.is_a?(CreditCard)
```
sinourain, line 590: a group of lines (before and after this one) have the same if statement: `credit_card.is_a?(CreditCard)`
I suggest doing just one, is this possible?

sinourain (commented):
LGTM, just a small comment!

sinourain (approved):
LGTM, just a small comment!

lib/active_merchant/billing/gateways/orbital.rb
almalee24: I left lit like this because of currency, since it can be used by other payment method

PR #5479, opened 2025-05-22
StripePI: validate capture reference

jherreraa:
[OPPS-254](https://spreedly.atlassian.net/browse/OPPS-254)

## Summary:

Validate that the reference used for Captures starts with ‘pi_‘

Unit Tests:
---
Finished in 0.076833 seconds.
74 tests, 409 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
963.13 tests/s, 5323.23 assertions/s

Remote Tests:
---
Finished in 831.599938 seconds.
110 tests, 508 assertions, 0 failures, 1 errors, 0 pendings, 0 omissions, 0 notifications
99.0909% passed
0.13 tests/s, 0.61 assertions/s

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -213,16 +213,22 @@ module ActiveMerchant # :nodoc:
       end
 
       def capture(money, intent_id, options = {})
-        post = {}
-        currency = options[:currency] || currency(money)
-        post[:amount_to_capture] = localized_amount(money, currency)
-        if options[:transfer_amount]
-          post[:transfer_data] = {}
-          post[:transfer_data][:amount] = options[:transfer_amount]
+        if intent_id.include?('pi_')
+          post = {}
+          currency = options[:currency] || currency(money)
+          post[:amount_to_capture] = localized_amount(money, currency)
+          if options[:transfer_amount]
+            post[:transfer_data] = {}
+            post[:transfer_data][:amount] = options[:transfer_amount]
+          end
+          post[:application_fee_amount] = options[:application_fee] if options[:application_fee]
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -213,16 +213,22 @@ def purchase(money, payment_method, options = {})
       end
 
       def capture(money, intent_id, options = {})
-        post = {}
-        currency = options[:currency] || currency(money)
-        post[:amount_to_capture] = localized_amount(money, currency)
-        if options[:transfer_amount]
-          post[:transfer_data] = {}
-          post[:transfer_data][:amount] = options[:transfer_amount]
+        if intent_id.include?('pi_')
```
almalee24, line 216: It would be better just to do a `return Response.new(false, error_message) unless intent_id.include?('pi_')` here. Then the error message could be more generic like `"Only Authorizations performed via the Payment Intent API are  capturable”`

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -213,6 +213,8 @@ def purchase(money, payment_method, options = {})
       end
 
       def capture(money, intent_id, options = {})
+        return Response.new(false, 'Only Authorizations performed via the Payment Intent API are  capturable') unless intent_id.include?('pi_')
```
almalee24, line 216: If you could just remove that extra space between are and capturable

almalee24: Everything else looks good

jherreraa: done!

PR #5491, opened 2025-06-16
Add support for Stripe PI multicapture

mjdonga:
Stripe Payment Intents: Add support for Stripe PI multi-capture

Unit tests result:
76 tests, 415 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote tests result:
112 tests, 532 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -415,6 +418,15 @@ def add_statement_descriptor_suffix_kanji_kana(post, options)
         post[:payment_method_options][:card][:statement_descriptor_suffix_kana] = options[:statement_descriptor_suffix_kana] if options[:statement_descriptor_suffix_kana]
       end
 
+      def add_request_multicapture(post, options)
+        request_multicapture = options[:request_multicapture].to_s
```
adarsh-spreedly, line 422: @mjdonga I don't think we need `to_s` here if `request_multicapture` is not provided then it will be nil and below condition will handle it.

adarsh-spreedly:
@mjdonga please add unit spec for passing newly defined values in `stripe_payment_intents_test.rb`

test/unit/gateways/stripe_payment_intents_test.rb
```diff
@@ -740,6 +756,50 @@ def test_authorize_with_apple_pay_with_billing_address
...
+      request_multicapture: 'if_available'
+    }
+    stub_comms(@gateway, :ssl_request) do
+      @gateway.purchase(@amount, @visa_token, options)
+    end.check_request do |_method, _endpoint, data, _headers|
+      assert_match('payment_method_options[card][request_multicapture]=if_available', data)
+    end.respond_with(successful_purchase_with_multicapture)
+  end
+
+  def test_capture_with_final_capture_true
```
mjdonga, line 781: @almalee24 Added these two specs

PR #5494, opened 2025-06-17
Ebanx: fix field nesting

yunnydang:
This change fixes the proper field nesting for a handful of optional fields. Currently they're being ignored and this will correct  that.

Local:
6298 tests, 81751 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
35 tests, 172 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
46 tests, 113 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/ebanx.rb
+++ b/lib/active_merchant/billing/gateways/ebanx.rb
@@ -180,7 +180,7 @@ module ActiveMerchant # :nodoc:
       def add_stored_credentials(post, options)
         return unless (stored_creds = options[:stored_credential])
 
-        post[:cof_info] = {
+        post[:payment][:cof_info] = {
           cof_type: stored_creds[:initial_transaction] ? 'initial' : 'stored',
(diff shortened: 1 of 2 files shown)


lib/active_merchant/billing/gateways/ebanx.rb
```diff
@@ -116,7 +116,8 @@ def verify(credit_card, options = {})
         add_payment_type(post, options)
         customer_country(post, options)
         post[:card] = payment_details(credit_card)
-        post[:device_id] = options[:device_id] if options[:device_id]
+        post[:payment] = {}
```
jcreiff, line 119: Just a quick gut check on this - I notice that unlike the `purchase` and `authorize` methods, the `post` on `verify` doesn't get instantiated with `post[:payment] = {}`, which is why you had add that here.

Are we sure from the EBANX docs that the `payment` object should be part of the verify operation too? Just curious about this since nothing else in this transaction type is nested within 🤔

PR #5508, opened 2025-07-01
Braintree OffsiteSynchronous Transactions: update additional_processor_response appropriately

aaqibrashidmir:
Update additional-processor-response properly or remove if new complete response doesn’t have additional-processor-response

Remote Tests:

127 tests, 618 assertions, 5 failures, 5 errors, 0 pendings, 0 omissions, 0 notifications 92.126% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/braintree_blue.rb
+++ b/lib/active_merchant/billing/gateways/braintree_blue.rb
@@ -637,8 +637,7 @@ module ActiveMerchant # :nodoc:
 
         transaction = result.transaction
         if transaction.vault_customer
-          vault_customer = {
-          }
+          vault_customer = {}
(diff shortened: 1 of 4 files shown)


lib/active_merchant/billing/gateways/braintree_blue.rb
```diff
@@ -724,6 +723,7 @@ def transaction_hash(result)
           'risk_data'                    => risk_data,
           'network_transaction_id'       => transaction.network_transaction_id || nil,
           'processor_response_code'      => response_code_from_result(result),
+          'additional_processor_response' => nil,
```
almalee24, line 726: Why do we need this here when additional_processor_response is already in line 627?

almalee24: There's no need to do this because in our side additional_processor_response will already be nil for successful response.

PR #5513, opened 2025-07-03
worldpay: add refund reference field

yunnydang:
Adds the optional refund reference field.

Local:
6326 tests, 81841 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
147 tests, 802 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
128 tests, 550 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed


jcreiff (commented):
@yunnydang Are you able to add a unit test that shows the tag is updated with the value? Right now we aren't asserting anything in the new test other than "refunds still succeed"

PR #5527, opened 2025-07-18
Paysafe: Add support for lodging fields

rachelkirk:
CER-2412

These fields can only be used by Lodging Merchants (must be enabled at gateway).

Unit Tests:
25 tests, 139 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Remote Tests:
37 tests, 109 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

Local Tests:
6387 tests, 82115 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications 100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paysafe.rb
+++ b/lib/active_merchant/billing/gateways/paysafe.rb
@@ -21,6 +21,7 @@ module ActiveMerchant # :nodoc:
         post = {}
         add_auth_purchase_params(post, money, payment, options)
         add_airline_travel_details(post, options)
+        add_lodging_details(post, options)
         add_split_pay_details(post, options)
         post[:settleWithAuth] = true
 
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -292,6 +293,22 @@ def add_flight_details(details, obj, leg)
...
+        post[:lodgingDetails][:checkOutDate] = options[:lodging_details][:check_out_date] if options[:lodging_details][:check_out_date]
+        post[:lodgingDetails][:customerServicePhone] = options[:lodging_details][:customer_service_phone] if options[:lodging_details][:customer_service_phone]
+        post[:lodgingDetails][:propertyLocalPhone] = options[:lodging_details][:property_local_phone] if options[:lodging_details][:property_local_phone]
+        post[:lodgingDetails][:extraCharges] = options[:lodging_details][:extra_charges] if options[:lodging_details][:extra_charges]
+        post[:lodgingDetails][:roomRate] = options[:lodging_details][:room_rate].to_i if options[:lodging_details][:room_rate]
+        post[:lodgingDetails][:programCode] = options[:lodging_details][:program_code] if options[:lodging_details][:program_code]
+        post[:lodgingDetails][:numberOfNights] = options[:lodging_details][:number_of_nights].to_i if options[:lodging_details][:number_of_nights]
+        post[:lodgingDetails][:isFireSafetyActCompliant] = options[:lodging_details][:is_fire_safety_act_compliant] if options[:lodging_details][:is_fire_safety_act_compliant]
```
jcreiff, line 309:
```suggestion
        post[:lodgingDetails][:isFireSafetyActCompliant] = options[:lodging_details][:is_fire_safety_act_compliant] if options[:lodging_details][:is_fire_safety_act_compliant].to_s.present?
```
Since this is a boolean I think we can't use the same `if` structure here; if the value passed is `false`, this field will get omitted. This should be nil-safe because `nil.to_s` turns into `""`, which evaluates `false` for `.present?`

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -21,6 +21,7 @@ def purchase(money, payment, options = {})
         post = {}
         add_auth_purchase_params(post, money, payment, options)
         add_airline_travel_details(post, options)
+        add_lodging_details(post, options)
```
jcreiff, line 24: Looking at the Paysafe docs as we worked through some of the other changes, we determined that this (and also the airline method) should probably be moved to the `add_auth_purchase_params` method, since the fields appear to be available on both auth and purchase - the primary difference between auth/purchase on this gateway is the `settleWithAuth` tag that gets added below

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -292,6 +293,22 @@ def add_flight_details(details, obj, leg)
...
+        return unless options[:lodging_details]
+
+        post[:lodgingDetails] = {}
+        post[:lodgingDetails][:hotelFolioNumber] = options[:lodging_details][:hotel_folio_number] if options[:lodging_details][:hotel_folio_number]
+        post[:lodgingDetails][:checkInDate] = options[:lodging_details][:check_in_date] if options[:lodging_details][:check_in_date]
+        post[:lodgingDetails][:checkOutDate] = options[:lodging_details][:check_out_date] if options[:lodging_details][:check_out_date]
+        post[:lodgingDetails][:customerServicePhone] = options[:lodging_details][:customer_service_phone] if options[:lodging_details][:customer_service_phone]
+        post[:lodgingDetails][:propertyLocalPhone] = options[:lodging_details][:property_local_phone] if options[:lodging_details][:property_local_phone]
+        post[:lodgingDetails][:extraCharges] = options[:lodging_details][:extra_charges] if options[:lodging_details][:extra_charges]
+        post[:lodgingDetails][:roomRate] = options[:lodging_details][:room_rate].to_i if options[:lodging_details][:room_rate]
```
jcreiff, line 306: One other last-minute thought here - I think we should actually leave `to_i` off here, as it would strip out any decimals (and a room rate could in theory be 199.50, for example)

PR #5530, opened 2025-07-21
Airwallex: add optional products array field

yunnydang:
Adds the optional products array field. The expected structure is defined on airwallex's documentation, but we will simply pass it through.

Local:
6387 tests, 82120 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
7 tests, 216 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
30 tests, 71 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
96.6667% passed

Note: my remote test on master fails for the same one here


lib/active_merchant/billing/gateways/airwallex.rb
```diff
@@ -244,7 +244,12 @@ def add_order(post, options)
         shipping[:last_name] = last_name if last_name
         shipping[:phone_number] = shipping_address[:phone_number] if shipping_address[:phone_number]
         shipping[:address] = physical_address
-        post[:order] = { shipping: }
+
```
jcreiff, line 247: I think we may need to add `options[:products]` to the guard clause at the top of the method, and skip the shipping address logic if no shipping address is provided

jcreiff (approved):

PR #5532, opened 2025-07-22
Paysafe: Add cruisline details and update airline details

yunnydang:
Adds the optional cruisline details fields. Also cleaned up some boolean fields and added some additional airline data fields per documentation:

cruiseline: https://developer.paysafe.com/en/cards-api/#/schemas/cruiselineTravelDetails

airline: https://developer.paysafe.com/en/cards-api/#/schemas/airlineTravelDetails

Local:
6387 tests, 82108 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
25 tests, 132 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
37 tests, 113 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/paysafe.rb
+++ b/lib/active_merchant/billing/gateways/paysafe.rb
@@ -20,7 +20,6 @@ module ActiveMerchant # :nodoc:
       def purchase(money, payment, options = {})
         post = {}
         add_auth_purchase_params(post, money, payment, options)
-        add_airline_travel_details(post, options)
         add_split_pay_details(post, options)
         post[:settleWithAuth] = true
 
@@ -129,6 +128,8 @@ module ActiveMerchant # :nodoc:
         add_three_d_secure(post, payment, options) if options[:three_d_secure]
         add_stored_credential(post, options) if options[:stored_credential]
         add_funding_transaction(post, options)
+        add_airline_travel_details(post, options)
+        add_cruiseline_travel_details(post, options)
       end
 
       # Customer data can be included in transactions where the payment method is a credit card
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -230,55 +231,137 @@ def add_three_d_secure(post, payment, options)
...
         post[:airlineTravelDetails][:tripLegs] = trip_legs_hash
       end
 
-      def add_leg_details(obj, leg)
+      def add_airline_leg_details(obj, leg)
         details = {}
         add_flight_details(details, obj, leg)
         details[:serviceClass] = leg[:service_class] if leg[:service_class]
-        details[:isStopOverAllowed] = leg[:is_stop_over_allowed] if leg[:is_stop_over_allowed]
+        details[:isStopOverAllowed] = leg[:is_stop_over_allowed].to_s.present?
```
jcreiff, line 364:
```suggestion
        details[:isStopOverAllowed] = leg[:is_stop_over_allowed] if  leg[:is_stop_over_allowed].to_s.present?
```

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -230,55 +231,137 @@ def add_three_d_secure(post, payment, options)
...
         details[:destination] = leg[:destination] if leg[:destination]
         details[:fareBasis] = leg[:fare_basis] if leg[:fare_basis]
         details[:departureDate] = leg[:departure_date] if leg[:departure_date]
```
jcreiff, line 367: While we are here could you add `fare` and `departureAirport` too?

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -230,55 +231,137 @@ def add_three_d_secure(post, payment, options)
...
+
+        post[:cruiselineTravelDetails] = {}
+        post[:cruiselineTravelDetails][:cruiseShipName] = cruiseline_data[:cruise_ship_name] if cruiseline_data[:cruise_ship_name]
+        post[:cruiselineTravelDetails][:passengerName] = cruiseline_data[:passenger_name] if cruiseline_data[:passenger_name]
+        post[:cruiselineTravelDetails][:departureDate] = cruiseline_data[:departure_date] if cruiseline_data[:departure_date]
+        post[:cruiselineTravelDetails][:returnDate] = cruiseline_data[:return_date] if cruiseline_data[:return_date]
+        post[:cruiselineTravelDetails][:country] = cruiseline_data[:country] if cruiseline_data[:country]
+        post[:cruiselineTravelDetails][:state] = cruiseline_data[:state] if cruiseline_data[:state]
+        post[:cruiselineTravelDetails][:originCity] = cruiseline_data[:origin_city] if cruiseline_data[:origin_city]
+        post[:cruiselineTravelDetails][:roomRate] = cruiseline_data[:room_rate] if cruiseline_data[:room_rate]
```
jcreiff, line 245: Wondering if we should call a `.to_i` on the fields that appear to be integers in the Paysafe docs (room_rate, ticket_number, fare)

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -230,55 +231,137 @@ def add_three_d_secure(post, payment, options)
...
+        add_cruiseline_passenger_details(post, options)
+        add_cruisline_trip_leg_details(post, options)
+      end
+
+      def add_cruiseline_ticket_details(post, options)
+        return unless ticket = options[:cruiseline_travel_details][:ticket]
+
+        post[:cruiselineTravelDetails][:ticket] = {}
+        post[:cruiselineTravelDetails][:ticket][:ticketNumber] = ticket[:ticket_number] if ticket[:ticket_number]
+        post[:cruiselineTravelDetails][:ticket][:isRestrictedTicket] = ticket[:is_restricted_ticket].to_s.present?
```
jcreiff, line 258:
```suggestion
        post[:cruiselineTravelDetails][:ticket][:isRestrictedTicket] = ticket[:is_restricted_ticket] if ticket[:is_restricted_ticket].to_s.present?
```

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -230,55 +231,137 @@ def add_three_d_secure(post, payment, options)
...
+      end
+
+      def add_cruiseline_passenger_details(post, options)
+        return unless passengers = options[:cruiseline_travel_details][:passengers]
+
+        passengers_hash = {}
+        passengers.each.with_index(1) do |passenger, i|
+          my_passenger = "passenger#{i}".to_sym
+          details = add_passenger_details(my_passenger, passenger[1])
```
jcreiff, line 267: Do you need to pass `my_passenger` into the method? It doesn't look like that argument gets used, and the symbol is only applied in line 269.

Also, why the `[1]` index for `passenger` here? Is the passenger detail array getting wrapped in an array? What happens if you just pass in `passenger`?

jcreiff: I'm also curious about why we need to do the `each.with_index(1)` if the way we are writing this we are expecting a passengers hash instead of passengers array.

I notice in the way the test is written, the key for each passenger object is already the format that Paysafe expects. I think it is probably preferable to have these passed in as hash with nested hashes as opposed to array of hashes, but I'm curious what we plan to document as the expectation for how each individual passenger object is labeled. I suppose the way we have it now, it could work with either `passenger_1` or `passenger1`, in which case the logic we have that creates the symbol should stay.

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -230,55 +231,137 @@ def add_three_d_secure(post, payment, options)
...
+
+          passengers_hash[my_passenger] = details
+        end
+        post[:cruiselineTravelDetails][:passengers] = passengers_hash
+      end
+
+      def add_passenger_details(obj, passenger)
```
jcreiff, line 274: Could we rename this to `add_cruisline_passenger` for clarity?

lib/active_merchant/billing/gateways/paysafe.rb
```diff
@@ -230,55 +231,137 @@ def add_three_d_secure(post, payment, options)
...
+        trip_legs.each.with_index(1) do |leg, i|
+          my_leg = "leg#{i}".to_sym
+          details = add_cruiseline_leg_details(my_leg, leg[1])
+
+          trip_legs_hash[my_leg] = details
+        end
+        post[:cruiselineTravelDetails][:tripLegs] = trip_legs_hash
+      end
+
+      def add_cruiseline_leg_details(obj, leg)
```
jcreiff, line 300: And perhaps rename this `add_cruiseline_leg`?

jcreiff (changes requested):

lib/active_merchant/billing/gateways/paysafe.rb
yunnydang: heya @jcreiff ! Paysafe expects that we designate the objects in here with incremental indexes very similar to the airline data so i thought it be easier to keep it concise. Also yes, i agree that we dont need to pass the my_passenger to the add sub method. Ill clean it up for trip legs as well and the airline one, since its not being used. also yes ill document that they should pass it in as nested hashes and not an array of hashes.

PR #5533, opened 2025-07-22
Stripe add amount details line items

yunnydang:
Adds the new optional payment line item fields when a version override is also present

Local:
6388 tests, 82129 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Unit:
82 tests, 447 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
113 tests, 496 assertions, 5 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
95.5752% passed

### DIFF
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -449,8 +449,8 @@ module ActiveMerchant # :nodoc:
         post[:payment_method_options][:card][:request_multicapture] = request_multicapture
       end
 
-      # stripe's newer API is depracting the level3 field. It still works now but
-      # we will open up the new amount_details field for level 3 data as well
+      # stripe's newer API is depracting the level3 field. It still works now but we will open
+      # up the new amount_details field for level 3 data as well if an override is included.
       def add_level_three(post, options = {})
         level_three = {}
 
--- a/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
+++ b/lib/active_merchant/billing/gateways/stripe_payment_intents.rb
@@ -54,7 +54,7 @@ module ActiveMerchant # :nodoc:
             add_error_on_requires_action(post, options)
             add_fulfillment_date(post, options)
             request_three_d_secure(post, options)
-            add_level_three(post, options)
+            options[:version] == '2025-04-30.preview' ? add_level_3_payment_line_item_fields(post, options) : add_level_three(post, options)
             add_card_brand(post, options)
             add_aft_recipient_details(post, options)
             add_aft_sender_details(post, options)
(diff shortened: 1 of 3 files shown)


lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -449,6 +449,8 @@ def add_request_multicapture(post, options)
         post[:payment_method_options][:card][:request_multicapture] = request_multicapture
       end
 
+      # stripe's newer API is depracting the level3 field. It still works now but we will open
```
jcreiff, line 452: *deprecating :)

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -54,7 +54,7 @@ def create_intent(money, payment_method, options = {})
             add_error_on_requires_action(post, options)
             add_fulfillment_date(post, options)
             request_three_d_secure(post, options)
-            add_level_three(post, options)
+            options[:version] == '2025-04-30.preview' ? add_level_3_payment_line_item_fields(post, options) : add_level_three(post, options)
```
jcreiff, line 57: What would you think about putting this check for the `version` at the top of the `add_level_three` method instead, to divert that method to the newly created `add_level_3_payment_line_item_fields` when needed?

I know it should work the same either way, but I wonder if for the sake of readability/anyone going to investigate the level three fields, it might be clearer to show that redirection within the method itself? I could be persuaded either way

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
```diff
@@ -462,6 +464,18 @@ def add_level_three(post, options = {})
...
 
+      def add_level_3_payment_line_item_fields(post, options)
+        post[:amount_details] ||= {}
+        post[:payment_details] ||= {}
+
+        post[:payment_details][:customer_reference] = options[:customer_reference] if options[:customer_reference]
+        post[:payment_details][:order_reference] = options[:order_reference] if options[:order_reference]
+
+        post[:amount_details][:discount_amount] = options[:discount_amount] if options[:discount_amount]
+        post[:amount_details][:shipping] = options[:level_3_shipping] if options[:level_3_shipping]
```
jcreiff, line 475: Previously the 3 fields that were nested underneath this (which I see are changing their names) were individually specified; any reason not to keep it that way going forward?

jcreiff (commented):
One other question - I see that `tax[total_amount]` can also be specified within `amount_details` for L2 (whereas it's nested in `line_items` for L3). Do we need to be able to support that too? 🤔

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
yunnydang: I agree that would make it readable and clean!

lib/active_merchant/billing/gateways/stripe_payment_intents.rb
yunnydang: We could do so! I just thought it would be easier to bundle them together since its what stripe expected an object rather than adding more fields

PR #5540, opened 2025-10-20
Change Braintree gem requirement to >= 4.11.0

latortuga:
Updated the Braintree gem requirement to version 4.11.0 or higher. Version 4.11.0 introduced availability of the network_token_details API that is required by this gateway. Version 2 is not nearly enough to work with this gateway.


latortuga:
This PR is still relevant, any chance of getting it merged? IDK if y'all accept contributions but it's a very short diff!

PR #5541, opened 2025-10-23
PinPayments: Add transaction_search method to search charges

hudakh:
Needed a way to search transactions

Local:
31 tests, 119 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

Remote:
35 tests, 100 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -203,6 +203,7 @@
 * Stripe PI: Update API version [almalee24] #5360
 * Airwallex: Add optional products array field [yunnydang] #5530
 * Pin Payments: Add inquire method to get details of payment [hudakh] #5537
+* Pin Payments: Add transaction_search method and support paginated API responses [hudakh] #5541
 
 == Version 1.137.0 (August 2, 2024)
 * Unlock dependency on `rexml` to allow fixing a CVE (#5181).
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -374,6 +374,7 @@ Spreedly Core gateway (December 2012)
 Pin gateway (February 2013)
 
 * Myles Eftos (madpilot)
+* Portions of the Pin Payments Gateway by Huda Kharrufa (hudakh)
 
 Merchant Warrior (February 2013)
 


hudakh:
> Interesting, this introduces a new pattern for an action. Why do you need to support this?

As a business, we are often finding that payment callbacks aren't received on our end, possibly due to the customer ending the session or network issues. We need to periodically reconcile our records with Pin Payments to ensure no transactions are missed. We're implementing hourly tasks to search for transactions created in the last hour and verify they're all recorded in our system. The `transaction_search` method allows us to query Pin Payments by date range and other criteria to perform this reconciliation, similar to how PayPal's gateway already supports `transaction_search` for the same use case.

The implementation adds a paginated_response handler for endpoints that return arrays (like search/list endpoints), similar to how other gateways handle list operations. This pattern can be reused for other Pin Payments list endpoints in the future (like `GET /charges`, `GET /refunds`, etc.).

hudakh:
> > Interesting, this introduces a new pattern for an action. Why do you need to support this?
> 
> As a business, we are often finding that payment callbacks aren't received on our end, possibly due to the customer ending the session or network issues. We need to periodically reconcile our records with Pin Payments to ensure no transactions are missed. We're implementing hourly tasks to search for transactions created in the last hour and verify they're all recorded in our system. The transaction_search method allows us to query Pin Payments by date range and other criteria to perform this reconciliation, similar to how PayPal's gateway already supports transaction_search for the same use case.
> 
> The implementation adds a paginated_response handler for endpoints that return arrays (like search/list endpoints), similar to how other gateways handle list operations. This pattern can be reused for other Pin Payments list endpoints in the future (like GET /charges, GET /refunds, etc.).
> 
> 

Do you think this is reasonable @aenand ?

hudakh:
> > > > Interesting, this introduces a new pattern for an action. Why do you need to support this?
> > > 
> > > 
> > > As a business, we are often finding that payment callbacks aren't received on our end, possibly due to the customer ending the session or network issues. We need to periodically reconcile our records with Pin Payments to ensure no transactions are missed. We're implementing hourly tasks to search for transactions created in the last hour and verify they're all recorded in our system. The transaction_search method allows us to query Pin Payments by date range and other criteria to perform this reconciliation, similar to how PayPal's gateway already supports transaction_search for the same use case.
> > > The implementation adds a paginated_response handler for endpoints that return arrays (like search/list endpoints), similar to how other gateways handle list operations. This pattern can be reused for other Pin Payments list endpoints in the future (like GET /charges, GET /refunds, etc.).
> > 
> > 
> > Do you think this is reasonable @aenand ?
> 
> @hudakh yeah I think so. Sorry for the delayed response I'm not working on this project with as much regularity as I used to.
> 
> Please add a changelog and I will merge it.

Thanks. No worries about the delay. I'm glad someone is still trying to keep the project going! Much appreciated!

aenand (approved):

PR #5544, opened 2025-12-17
Fat Zebra: Add support for purchase with wallet

dacook:
May be Apple Pay or Google Pay, as per https://docs.fatzebra.com/reference/create-a-purchase-with-wallet

~~Can't test fully here because it requires a valid current token from the Apple SDK (for example) which requires Apple device.~~
I've since found that a recorded test token does seem to work; let me know if you think it would be worth adding a remote test with that, and if I should add the token (a json blob) to the `test/fixtures.yml` file.

I'll add a changelog comment after first review, to avoid conflicts with longrunning PRs.

```
ruby -Itest test/unit/gateways/fat_zebra_test.rb -n test_successful_purchase_with_wallet_apple_pay
1 tests, 7 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed

ruby -Itest test/remote/gateways/remote_fat_zebra_test.rb -n test_failed_purchase_with_wallet_apple_pay
1 tests, 3 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
```

### DIFF
--- a/lib/active_merchant/billing/gateways/fat_zebra.rb
+++ b/lib/active_merchant/billing/gateways/fat_zebra.rb
@@ -25,7 +25,7 @@ module ActiveMerchant # :nodoc:
         post = {}
 
         add_amount(post, money, options)
-        add_creditcard(post, creditcard, options)
+        add_creditcard(post, creditcard, options) unless creditcard.nil?
         add_extra_options(post, options)
         add_order_id(post, options)
         add_ip(post, options)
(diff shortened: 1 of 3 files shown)


dacook:
I don't understand the CI failure, but can see that it occurs on other PRs too so is pre-existing: https://github.com/activemerchant/active_merchant/actions/runs/19050980996/job/54410942014

dacook:
Hi @aenand , are you available to review this PR? It is currently in use in production with no known issues.

aenand:
HI @dacook I am no longer a member of this repository so I cannot assist with merging it!

dacook:
Thanks for the heads up both. 
I'm not sure who, if anyone is maintaining this repo, and what the priorities are. Sadly I don't think it will be worth us contributing any further so I think we'll be using a fork and/or moving away from this repo in the future.

PR #5552, opened 2026-04-07
CyberSource REST: Add commerce_indicator option override

romankhadka:
## Summary

The CyberSource REST API accepts many values for `processingInformation.commerceIndicator` (`moto`, `retail`, `recurring`, `install`, `internet`, etc.) but the gateway only computes three of them from `stored_credential[:reason_type]`.

The SOAP gateway (`CyberSourceGateway`) already supports `options[:commerce_indicator]` as a direct pass-through ([line 843](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/cyber_source.rb#L843)). This brings the same capability to the REST gateway for parity.

## Use Case

MOTO (Mail Order/Telephone Order) transactions require `commerceIndicator: "moto"`. Without this option, there is no way to set it via the REST gateway. The [CyberSource docs](https://github.com/CyberSource/cybersource-rest-client-node/blob/master/docs/Ptsv2paymentsProcessingInformation.md) list `moto` as one of many valid indicator values.

## Changes

- Add `add_commerce_indicator` private method that applies `options[:commerce_indicator]` as a final override after `add_stored_credentials` in `build_auth_request`
- When the option is absent, existing behavior is completely unchanged
- Follows the same `add_*` method convention used by every other line in `build_auth_request`

## Tests

- `test_commerce_indicator_override` — verifies override with `"moto"` on a regular credit card
- `test_commerce_indicator_override_with_stored_credentials` — verifies override takes precedence over stored credential computed value
- All 48 CyberSource REST tests pass (46 existing + 2 new)


romankhadka:
Update: this still applies to the current build. PR #5552 is mergeable against current `activemerchant:master`, and the affected CyberSource REST tests pass on the PR merge ref.

Verification run locally with Ruby 3.2.2 across the CI gemfiles (`rails50`, `rails51`, `rails52`, `rails60`, `rails_master`):

`BUNDLE_GEMFILE=<gemfile> bundle exec ruby -Itest test/unit/gateways/cyber_source_rest_test.rb`

Each gemfile run completed with 48 tests, 249 assertions, 0 failures, and 0 errors. Please keep this PR open.

